PR #73, opened 2017-07-02
Multiple dynamic panels only point to source used by the first panel #70

acjh:
Fixes #70

Issue #70, opened 2017-06-07
Multiple dynamic panels only point to source used by the first panel


rebekahlow-jy:
Issue:
Multiple dynamic panels seem to only load file from the first dynamic panel used (or previously used dynamic panels) and use that file for the rest of the dynamic panels. 

Replication Steps: 
1. Use dynamic panel in a file pointing to file_a.md
2. Use another dynamic panel pointing to file_b.md
3. Bug: Both dynamic panels will show file_a.md
4. Comment out dynamic panel pointing to file_a.md
5. Refresh page, dynamic panel pointing to file_b.md will be correctly showing the file.
6. Uncomment dynamic panel pointing to file_a.md
7. Bug: Both dynamic panels will show file_b.md which was shown previously
8. Uncomment dynamic panel pointing to file_b.md
9. Refresh page, dynamic panel pointing to file_a.md will be correctly showing the file.
6. Uncomment dynamic panel pointing to file_b.md
7. Bug: Both dynamic panels will show file_b.md which was shown previously

damithc:
@Gisonrg we'll probably need your help with this

acjh:
@rebekahlow-jy Can I check if the basename of the files (given different directories) were the same?

damithc:
> Can I check if the basename of the files (given different directories) were the same?

@acjh here's an example in https://github.com/MarkBind/se-book/edit/master/oop/associations/basics/index.md

In that page this part is supposed to be a dynamic include but we are using a static include because dynamic includes don't work.
```html
<panel header="UML: Object Diagrams: Object Structures">
  <include src="../../../uml/objectDiagrams/objectStructures/topicPanel.md" />
</panel>
```

acjh:
Yes, I narrowed the problem to the case where all the dynamic includes are for `topicPanel.md`.
This generates `topicPanel._include_.html` in the includer directory, based on the basename.
This is likely why both panels show the same file, despite pointing to files in different directories.

However, the issue mentions `file_a.md` and `file_b.md`. Are they actually both `topicPanel.md`?

If so, we can deconflict by:
1. storing the generated file in the original directory, or
2. generating a unique file name in the includer directory.

I prefer `1` because `2` may require a hashing algorithm, with almost no benefit in deployment.
`2` lets you to see a generated file in the same folder (in `_site`) and can hide the original path.

damithc:
@Gisonrg any inputs?

damithc:
[...]

PR #75, opened 2017-07-06
Multiple dynamic panels only point to source used by the first panel #70

acjh:
Fixes Markbind/se-book#13, #70, #73

I wrongly used [`path.resolve`](https://github.com/MarkBind/markbind/commit/73722d12462f53e107337ba1c70a4e39ebe4d545#diff-3fab227e34d65fe9bb2d6ea53eec41cfR195), which has a different behavior on Windows, for a url.
Instead, the relative path should be rebased on `baseUrl`.


Gisonrg:
Thanks! I will let @rebekahlow-jy to give it a try.

PR #113, opened 2018-01-07
Improve generation of radio-group id #106

### LINKED ISSUE #106: Improve generation of radio-group id
There's a huge diff just from the `radio-group` id, which is also used for `radio-list-item` names.
This makes the diff difficult to compare and practically useless, e.g. https://github.com/nus-cs2103/website/commit/2a264dde68713224c5050624207f274cb2e6793d

Currently, it's randomly generated:

https://github.com/MarkBind/markbind/blob/d25efcbcb0fc4f1469108bf88c09523342701245/lib/markdown-it/markdown-it-radio-button.js#L21

**Suggestion**: Explore *smart* deterministic generation based on a parameter.

*smart* → counting is deterministic generation, but isn't *smart* because removing a single file may affect the id of all other radio groups down the line. A possible example is relative file path — to be explored.

acjh:
Resolves #106


lib/markdown-it/markdown-it-radio-button.js
```diff
@@ -110,4 +110,7 @@ function isListItem(token) { return token.type === 'list_item_open'; }
...
...7],y,1735328473);X=f(X,W,V,Y,C[P+12],w,2368359562);Y=D(Y,X,W,V,C[P+5],o,4294588738);V=D(V,Y,X,W,C[P+8],m,2272392833);W=D(W,V,Y,X,C[P+11],l,1839030562);X=D(X,W,V,Y,C[P+14],j,4259657740);Y=D(Y,X,W,V,C[P+1],o,2763975236);V=D(V,Y,X,W,C[P+4],m,1272893353);W=D(W,V,Y,X,C[P+7],l,4139469664);X=D(X,W,V,Y,C[P+10],j,3200236656);Y=D(Y,X,W,V,C[P+13],o,681279174);V=D(V,Y,X,W,C[P+0],m,3936430074);W=D(W,V,Y,X,C[P+3],l,3572445317);X=D(X,W,V,Y,C[P+6],j,76029189);Y=D(Y,X,W,V,C[P+9],o,3654602809);V=D(V,Y,X,W,C[P+12],m,3873151461);W=D(W,V,Y,X,C[P+15],l,530742520);X=D(X,W,V,Y,C[P+2],j,3299628645);Y=t(Y,X,W,V,C[P+0],U,4096336452);V=t(V,Y,X,W,C[P+7],T,1126891415);W=t(W,V,Y,X,C[P+14],R,2878612391);X=t(X,W,V,Y,C[P+5],O,4237533241);Y=t(Y,X,W,V,C[P+12],U,1700485571);V=t(V,Y,X,W,C[P+3],T,2399980690);W=t(W,V,Y,X,C[P+10],R,4293915773);X=t(X,W,V,Y,C[P+1],O,2240044497);Y=t(Y,X,W,V,C[P+8],U,1873313359);V=t(V,Y,X,W,C[P+15],T,4264355552);W=t(W,V,Y,X,C[P+6],R,2734768916);X=t(X,W,V,Y,C[P+13],O,1309151649);Y=t(Y,X,W,V,C[P+4],U,4149444226);V=t(V,Y,X,W,C[P+11],T,3174756917);W=t(W,V,Y,X,C[P+2],R,718787259);X=t(X,W,V,Y,C[P+9],O,3951481745);Y=K(Y,h);X=K(X,E);W=K(W,v);V=K(V,g)}var i=B(Y)+B(X)+B(W)+B(V);return i.toLowerCase()};
```
Gisonrg: Since this is offline code, we could use Node's Crypto library. I think you can just use crypto.createHash('md5') instead of this function.

lib/markdown-it/markdown-it-radio-button.js
```diff
@@ -18,7 +18,7 @@ module.exports = function(md, options) {
         if (group) {
           group = group[1];
         } else {
-          group = Math.random().toString(36).substr(2, 5); // generate a new group id
+          group = md5(tokens[i-5].content + tokens[i-4].content + tokens[i].content).substr(2, 5); // generate a deterministic group id
```
Gisonrg: If you use Crypto library, you can do hash.update(tokens[i-5].content).update(tokens[i-4].content).update(tokens[i].content).digest('hex');

Gisonrg: .substr(2, 5) is only used for getting a short random id. We could use a longer one if we go with md5 way (or just getting rid of this).

acjh: Thanks for the `crypto` suggestion!

Oddly, sometimes it doesn't give the same value as the `md5` function, but it's fine for our use.

I think it's nice to keep it short. I doubt there would be a collision anyway. What's your thought?

Gisonrg: Let's keep it for now then. It should work fine for now and we could always change it later :)

Issue #95, opened 2017-08-09
MCQ options get bold style


damithc:
Syntax used `- ( ) a. Each developer should find a suitable coding standard and follow it in their coding.` etc.

![image](https://user-images.githubusercontent.com/1673303/29113974-5bec6ce2-7d25-11e7-8881-fb3cb247399d.png)

Tick boxes don't seem to have the same problem. i.e. `- [ ] a. Each developer ...`

PR #114, opened 2018-01-17
Remove label wrapper for radio list

nicholaschuayunzhi:
Would like to re-look into the issue #95 that was "resolved" recently. 

Initially I had noted that the label was causing the bold text and made changes to the css to rectify it in https://github.com/MarkBind/markbind-cli/pull/12

However I feel that the label wrapper should not be there to be consistent with task-list-items.

 `- [ ] item`
```html
<li class="task-list-item"><input class="task-list-item-checkbox" type="checkbox"> item</li>
```

This pr removes the wrapper from the markdown-it-radio-button plugin. MCQ options will not have the default bold stlye. https://github.com/MarkBind/markbind-cli/pull/12 can be reverted as it is extraneous.

Would like to know your thoughts @acjh. Thanks!

### DIFF
--- a/lib/markdown-it/markdown-it-radio-button.js
+++ b/lib/markdown-it/markdown-it-radio-button.js
@@ -1,7 +1,6 @@
 const crypto = require('crypto');
 
 var disableRadio = false;
-var useLabelWrapper = true;
 
 /**
  * Modified from https://github.com/revin/markdown-it-task-lists/blob/master/index.js
@@ -9,7 +8,6 @@ var useLabelWrapper = true;
 module.exports = function(md, options) {
   if (options) {
     disableRadio = !options.enabled;
-    useLabelWrapper = !!options.label;
   }
 
   md.core.ruler.after('inline', 'radio-lists', function(state) {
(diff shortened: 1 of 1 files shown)


acjh:
Labels are actually a [UX enhancement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Providing_a_bigger_hit_area_for_your_radio_buttons) :)

It's often done for radio buttons, and sometimes for checkboxes.

I couldn't find an authoritative guide on whether to use it for checkboxes, but here's my view:
The difference is that clicking twice on the label of a radio button shouldn't unselect an option.
Clicking twice on the label of a checkbox should be consistent — it shouldn't unselect an option.
So, I wouldn't add a label wrapper for the task-list-item just yet.

nicholaschuayunzhi:
> Labels are actually a UX enhancement :)

Thanks for sharing :) Did not consider this at all.

> Clicking twice on the label of a checkbox should be consistent — it shouldn't unselect an option.

Do you mean the opposite - it should unselect the option since its like a check list?

acjh:
I meant that I'm fine with a checkbox label not selecting (and not unselecting) an option.

That's the current behaviour.

I think we should leave it as-is. There are a range of other issues at hand ;)

PR #115, opened 2018-01-17
Rebase reference for static includes that target a div

### LINKED ISSUE #94: static includes that target a div: re-basing links doesn't work
Given this code `[link]({{baseUrl}}/common/print.html)` in `book/usage.md` which is included in `schedule\overview\about.md` 
image
image

`<include src="../../book/about/usage.md"/>` : link works
`<include src="../../book/about/usage.md#printing"/>` : produces a broken link
`<include src="../../book/about/usage.md#printing" dynamic/>` : link works

Looks like re-basing links doesn't work for static includes that targets a div (rather than entire page)

acjh:
Fixes #94

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -276,7 +276,7 @@ Parser.prototype._trimNodes = function (node) {
   }
 };
 
-Parser.prototype.includeFile = function (file, cb, config) {
+Parser.prototype.includeFile = function (file, config, cb) {
   cb = cb || function () {}; // create empty callback
 
   let context = {};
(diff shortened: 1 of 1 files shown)


lib/parser.js
```diff
@@ -275,7 +276,7 @@ Parser.prototype._trimNodes = function (node) {
   }
 };
 
-Parser.prototype.includeFile = function (file, cb) {
+Parser.prototype.includeFile = function (file, cb, config) {
```
Gisonrg: I am thinking about deprecating the callback for the API and use Promise only, and I don't think we are still using callback anywhere in the codebase, so I think it's safe to change to put `config` after the `file` parameter here (do a check before you do so) :)  
Let's create an issue and get rid of callback in a separate PR.

acjh: Checked and changed!

Issue #55, opened 2017-05-01
Fix wrong words being bolded in ```text blocks


damithc:
E.g. `As` `each`
![image](https://cloud.githubusercontent.com/assets/1673303/25578839/0ef73224-2ea5-11e7-8699-7678d4c7b795.png)

Should be:
```text
[Epic] As a lecturer, I can monitor student participation levels 
  * As a lecturer, I can view the forum post count of each student 
    so that I can identify the activity level of students in the forum
  * As a lecturer, I can view webcast view records of each student 
    so that I can identify the students who did not view webcasts
  * As a lecturer, I can view file download statistics of each student
    so that I can identify the students who do not download lecture materials
```

Gisonrg:
Fixed as auto language detection is disabled. Now you have to explicitly state the language of the code block.

PR #125, opened 2018-01-25
[WIP] Search for file in site/common directory if not found in current working folder #110

### LINKED ISSUE #110: Support reuse of boilerplate files
For example, https://github.com/se-edu/se-book has a lot of boilerplate files that are repeated 100s of times. e.g. many of the files in [this folder](https://github.com/se-edu/se-book/tree/master/reuse/apis/what) are repeated in many other folders.

It would be nice if we can put those files in a central location and reuse everywhere.

rachx:
Fixes #110 - Support reuse of boilerplate files by looking for missing files in a central location

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -19,6 +19,8 @@ const nunjucks = require('nunjucks');
 
 const ATTRIB_INCLUDE_PATH = 'include-path';
 const ATTRIB_CWF = 'cwf';
+
+const BOILERPLATE_DIRECTORY_NAME = '_boilerplates';
 /*
  * Utils
  */
@@ -74,8 +76,8 @@ Parser.prototype._preprocess = function (element, context, config) {
       self._fileCache[filePath] = self._fileCache[filePath] ?
         self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
     } catch (e) {
+      filePath = path.resolve(path.join(process.cwd(), BOILERPLATE_DIRECTORY_NAME), includeSrcPath);
       try {
-        filePath = path.resolve(path.join(process.cwd(), 'common'), includeSrcPath);
         self._fileCache[filePath] = fs.readFileSync(filePath, 'utf8');
       } catch (e) {
         // Read file fail
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -78,11 +78,11 @@ Parser.prototype._preprocess = function (element, context, config) {
         filePath = path.resolve(path.join(process.cwd(), 'common'), includeSrcPath);
         self._fileCache[filePath] = fs.readFileSync(filePath, 'utf8');
       } catch (e) {
-          // Read file fail
-          this._onError(e);
-          element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
-          return element;
-        }
+        // Read file fail
+        this._onError(e);
+        element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
+        return element;
+      }
     }
 
 
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -81,15 +81,10 @@ Parser.prototype._preprocess = function (element, context, config) {
       self._fileCache[filePath] = self._fileCache[filePath] ?
         self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
     } catch (e) {
-      filePath = path.resolve(path.join(process.cwd(), BOILERPLATE_DIRECTORY_NAME), includeSrcPath);
-      try {
-        self._fileCache[filePath] = fs.readFileSync(filePath, 'utf8');
-      } catch (e) {
-        // Read file fail
-        this._onError(e);
-        element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
-        return element;
-      }
+      // Read file fail
+      this._onError(e);
+      element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
+      return element;
     }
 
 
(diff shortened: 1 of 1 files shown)


acjh:
#55 seems unrelated.

lib/parser.js
```diff
@@ -74,10 +74,15 @@ Parser.prototype._preprocess = function (element, context, config) {
       self._fileCache[filePath] = self._fileCache[filePath] ?
         self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
     } catch (e) {
-      // Read file fail
-      this._onError(e);
-      element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
-      return element;
+      try {
+        filePath = path.resolve(path.join(process.cwd(), 'common'), includeSrcPath);
```
acjh: This line should be outside try-catch.

acjh:
Actually, the boilerplate files in "common" are addressable pages in [site.json](https://github.com/se-edu/se-book/blob/d1261b8d39a02bd8ba6fb5094edeec3dd29d8a28/site.json#L1836-L1851).

@damithc Since this concerns authoring, do you have a suitable "central location" in mind?
Both "common" and "reuse" are currently being used in https://github.com/se-edu/se-book.

How about a prefixed "_common" or "_templates"?

damithc:
> How about a prefixed "_common" or "_templates"?

`_templates` sound good but might already be used by other frameworks for other purposes.
`_boilerplates` ? It matches the purpose of the folder.

rachx:
Should the error for the missing file show the address of the original directory or boilerplate folder or both?

damithc:
> Should the error for the missing file show the address of the original directory or boilerplate folder or both?

Original directory. It's good if the error message can mention which file is referring to the missing file (because sometimes it is an error in the reference rather than a file actually missing)

lib/parser.js
```diff
@@ -195,7 +210,7 @@ Parser.prototype._parse = function (element, context) {
       element.attribs.isOpen = hasIsOpen ? hasIsOpen : false;
       try {
         if (fs.statSync(element.attribs.src).isFile()) {
-          let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
+          let resultDir = path.dirname(path.relative('{{hostBaseUrl}}', element.attribs.src));
```
acjh: Why is this change necessary?

acjh:
Does not seem to work when I simply rename "common" folder to "_boilerplates" (or copy it to root).

lib/parser.js
acjh: This will break when hosted on GitHub:

screenshot 2018-01-27 02 16 11

rachx:
Just to clarify: Is the _boilerplates folder supposed to replace the common folder?
I tested by retaining the original common folder and creating a new _boilerplates folder with other files (using a small test site and 2103 website). The files in the _boilerplates folder were included correctly.

lib/parser.js
rachx: If a dynamic panel includes a boilerplate file,
I will replace the `filePath` and `element.attribs.src` with the boilerplate filepath. However, it does not work correctly as it is unable to retrieve the content from source: `/\_boilerplates\filename._include_.html`.
 
I tried normalizing the filepath and replacing it with the above line. It works locally during testing but I didn't know it will break when hosted on Github.

acjh: The `{{baseUrl}}` is `/website` → https://nus-cs2103-ay1718s2.github.io/website/

Doing `../schedule` will give https://nus-cs2103-ay1718s2.github.io/schedule/
It's important that rendered files do not contain relative paths for `include` to work well.

damithc:
> Just to clarify: Is the _boilerplates folder supposed to replace the common folder?

No, `common` is just a regular folder that I use to put files that are used from many other places of the site. The files in that folder are not boilerplate code; rather, they are _widely included_ files. Those files don't contain any location-sensitive references, so it is fine to keep only one copy of them.
The new `_boilerplates` folder is different. It's supposed to contain repetitive files such as `full.md` that contains exactly the same text but needs to be repeated in many places because they contain location-sensitive references (e.g., if `full.md` contains a `<include src="text.md">`, it is location-sensitive because the which `text.md` file to include will depend on the folder `full.md` is in). Although the new `_boilerplates` folder will allow us to keep just one copy of the repetitive file, the final outcome should be as if the file was actually present in all the different locations. i.e., as if the file is first copied to the target folder before it is rendered. Hope there is no confusion about that.

acjh:
Thanks Prof Damith for the explanation. It was a confusing example on my part.

> i.e. as if the file is first copied to the target folder before it is rendered

@rachx Together with [your findings](https://github.com/MarkBind/markbind/pull/125#discussion_r164262832) and [my reply](https://github.com/MarkBind/markbind/pull/125#discussion_r164263332), one way I can think of is to read the file in `_boilerplates` when we read a target file that does not exist, rather than replace `filePath`.

rachx:
Adding a WIP label since it currently does not show the expected behavior (instead of being location-sensitive, it is searching with the `_boilerplates` folder)

rachx:
Do we have to handle dynamic includes of boilerplate files? I have not figured out how to resolve the file path correctly.
Will there be only one `_include_.html` file per boilerplate file?

acjh:
> Do we have to handle dynamic includes of boilerplate files?

Yes, it should behave like the target file was in the folder.

> Will there be only one `_include_.html` file per boilerplate file?

Yes.

rachx:
Done in https://github.com/MarkBind/markbind/pull/130/

PR #126, opened 2018-01-25
Update package.json to point to htmlparser2 fork

### LINKED ISSUE #101: Code snippets: text in angle brackets become lowercase
For example `<Person>` becomes `<person>` after rendering to html. 
The current workaround is to add extra spaces `< Person >` which stays as `< Person >` after rendering.

danielbrzn:
Change the dependency in package.json to point to Markbind's fork of htmlparser2 that includes fixes for Markdown-in-HTML parsing.

Follows MarkBind/htmlparser2#1

Resolves #101

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "bluebird": "^3.4.7",
     "cheerio": "^0.22.0",
     "highlight.js": "^9.10.0",
-    "htmlparser2": "^3.9.2",
+    "htmlparser2": "MarkBind/htmlparser2#v3.10.0-markbind.1",
     "lodash": "^4.17.2",
     "markdown-it": "^8.3.0",
     "markdown-it-anchor": "^4.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
     "anymatch": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-1.3.2.tgz",
-      "integrity": "sha512-0XNayC8lTHQ2OI8aljNCN3sSx6hsr/1+rlcDAotXJR7C1oZZHCNsfpbKwMjRA3Uqb5tF1Rae2oloTr4xpq+WjA==",
+      "integrity": "sha1-VT3Lj5HjyImEXf26NMd3IbkLnXo=",
       "optional": true,
       "requires": {
         "micromatch": "2.3.11",
@@ -95,7 +95,7 @@
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
-      "integrity": "sha512-L3hKV5R/p5o81R7O02IGnwpDmkp6E982XhtbuwSe3O4qOtMMMtodicASA1Cny2U+aCXcNpml+m4dPsvsJ3jatg==",
+      "integrity": "sha1-NgSLv/TntH4TZkQxbJlmnqWukfE=",
       "optional": true
     },
     "array-union": {
@@ -229,7 +229,7 @@
         "css-select": "1.2.0",
         "dom-serializer": "0.1.0",
         "entities": "1.1.1",
-        "htmlparser2": "3.9.2",
+        "htmlparser2": "github:MarkBind/htmlparser2#815b507c5268a178fa24741d6f74fcc0bc9be4cd",
         "lodash.assignin": "4.2.0",
         "lodash.bind": "4.2.1",
         "lodash.defaults": "4.2.0",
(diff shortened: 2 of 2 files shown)


acjh:
Can you share the link of the article that you showed me?

danielbrzn:
[Article] (https://coderwall.com/p/q_gh-w/fork-and-patch-npm-moduels-hosted-on-github)

acjh:
I'm not sure how versioning would work with `tarball/master`.

Maybe we should use commits (or tags; both mentioned in your link) to target specific versions.

@Gisonrg Your thoughts?

danielbrzn:
I think it would pull the latest version of the master instead of a specific version which I feel is more ideal as we don't have to keep updating the link in package.json.

acjh:
> which I feel is more ideal as we don't have to keep updating the link in package.json.

That's not ideal. It will be hard to debug errors if an earlier version of **MarkBind** is used with a later version of **htmlparser2** (or any other library). There's a reason why packages are always published with a **package.json** that points to specific known-working versions of other packages.

Gisonrg:
> I think it would pull the latest version of the master instead of a specific version which I feel is more ideal as we don't have to keep updating the link in package.json.

Even though it might work in some cases, this is not an ideal practice in general :)
If you always depends on the latest version (or snapshot) of a library, you may have issues regarding compatibility. Suppose you change the API in your library in one commit but does not update the usage for your users (e.g. MarkBind), all usages in user library might not work due to the change.
In production system, you won't release a version that's under active development. That's why they usually have "release" branches which are stable build after testing.
Also, if you have multiple engineers in the team use different version of a library but working in the same project, and they don't know about that (because they all pull from latest version at the time of their pull), one might expect inconsistent behaviour of the library due to different version.
That's exactly why we have to specific a version in the `package.json`, to ensure consistency across all users.

danielbrzn:
I see. Should we tag a version of our htmlparser2 fork and use it in our`package.json` instead?

acjh:
Sure. Let's use [pre-release versioning](https://semver.org/#spec-item-9) as per the Semantic Versioning guidelines.

Since the current version of htmlparser2 is [v3.9.2](https://github.com/MarkBind/htmlparser2/releases), I propose `v3.10.0-markbind.1`.

Note: We can consider versioning [MarkBind/vue-strap](https://github.com/MarkBind/vue-strap) too.

acjh:
Can you add "Follows MarkBind/htmlparser2#1" in the PR description?

acjh:
Can you add "Resolves #101" in the PR description?

danielbrzn:
Just to clarify, the version number in `package.json` should read as follows?

```
    "htmlparser2": "^v3.10.0-markbind.1",
```

acjh:
No, that's if you publish as htmlparser2 in the npm registry.
The website you linked suggests a form that it should take.

package.json
```diff
@@ -16,7 +16,7 @@
     "bluebird": "^3.4.7",
     "cheerio": "^0.22.0",
     "highlight.js": "^9.10.0",
-    "htmlparser2": "^3.9.2",
+    "htmlparser2": "https://github.com/MarkBind/htmlparser2/archive/v3.10.0-markbind.1.tar.gz",
```
Gisonrg: I believe if we push a new tag (like the v3.10 here) to the head of repo, we can do something like:
```
"htmlparser2": "MarkBind/htmlparser2#v3.10.0",
```

danielbrzn: Yep that works! The tag is `v3.10.0-markbind.1`, will be changing it.

package.json
```diff
@@ -16,7 +16,7 @@
     "bluebird": "^3.4.7",
     "cheerio": "^0.22.0",
     "highlight.js": "^9.10.0",
-    "htmlparser2": "^3.9.2",
+    "htmlparser2": "https://github.com/MarkBind/htmlparser2#v3.10.0-markbind.1",
```
Gisonrg: Can we get rid of the `https://github.com/` part?

The "author/repo" format should be supported by [npm](https://docs.npmjs.com/files/package.json#github-urls)

acjh:
Commit **package-lock.json** too.

acjh:
@danielbrzn Thanks for your contribution!

acjh:
@danielbrzn I just noticed the bad [integrity checksum changes](https://github.com/MarkBind/markbind/pull/126/files#diff-32607347f8126e6534ebc7ebaec4853d) ([fixed](https://github.com/MarkBind/markbind/commit/28095fb3e5b2c9ad2c44e6c884dfeaa82bb122c7#diff-32607347f8126e6534ebc7ebaec4853d)). You may need to upgrade npm.

danielbrzn:
Odd. What version of npm are you running? I have v5.6.0.

acjh:
v5.6.0 as well. See https://github.com/npm/npm/issues/16938#issuecomment-339863980.

Issue #127, opened 2018-01-26
Nested includes do not get live reloaded


nicholaschuayunzhi:
Stumbled upon a bug when trying to improve build time for live reloading. 

BUG: Anything nested in two levels of dynamic include do not get updated.

### Set up

`site.json`

```json
{ 
  "baseUrl": "",
  "pages": [
    {
      "src": "index.md",
      "title": "Hello World"
    }
  ],
  "ignore": [
    "_site/*",
    "site.json"
  ],
  "deploy": {
    "message": "Site Update."
  }
}

```

`index.md`

```md
<include src="include1.md" dynamic />
```

`include1.md`

```md
<include src="include2.md" dynamic />

Any changes here will be reloaded correctly
```

`include2.md`

``` md
Any changes here will NOT be reloaded correctly
```

### Steps
1. run `$ markbind serve`
2. make a change to text in `include2.md` and save
3. site reloads but content in the nested dynamic panel is unchanged

### Additional Notes
- after applying steps 1 - 3, a change to `include1.md` will update the include1's content correctly but the include2 content is still unchanged

nicholaschuayunzhi:
I found a case in `website-base` for CS2103.

### Steps to Reproduce
1. `$ markbind serve`
2. Make changes to `\website-base\book\codeQuality\practices\useNameExplain\text.md`
3. Wait for live reload
4. Week12 -> Todo ->  Admin >> Project v1.5rc -> Tips how to score high for code ... -> code quality guidelines -> Guideline: Namewell -> Intermediate -> Use name to explain well
5. Changes made not updated here

![image](https://user-images.githubusercontent.com/22221132/35520926-6aca1c26-0553-11e8-932f-c26536466ab1.png)
![image](https://user-images.githubusercontent.com/22221132/35520955-7e727692-0553-11e8-9a26-8c202d1f6c77.png)
![image](https://user-images.githubusercontent.com/22221132/35521005-945da10c-0553-11e8-85c4-4da00f8df0f3.png)

nicholaschuayunzhi:
Update: removing https://github.com/MarkBind/markbind-cli/blob/54697812e9ef554ef5d613f4df8210160de3467e/lib/Page.js#L138-L147 resolves this issue, but comes with longer build time. Have not timed it yet.

Issue #88, opened 2017-07-31
Speed up the build task


damithc:
Is there any way to speed up the `markbind build` task?

damithc:
This seems to be a case of website size. Small websites are built quite fast. May be figure out a way to do incremental builds? Probably not trivial though.

damithc:
Closing this as no longer a major problem

PR #128, opened 2018-01-26
Speed up live reload process

nicholaschuayunzhi:
Instead of rebuilding the entire site for a change to one file, rebuild relevant pages only.

If changed file was statically or dynamically included in page, regenerate the page.

Helps with build performance when editing sites. Related #88 

https://github.com/MarkBind/markbind-cli/pull/17

#### Steps
1. `$ markbind serve`
2. make changes to files.
3. save _site file
4. `$ markbind build`
5. compare old _site with new _site

##### Paths tried
- admin/project-v1.5rc
- schedule/week1/index.html
- book/codeQuality/practices/useNameExplain/text.md

##### Results
image
image

- These files were not rebuilt correctly in step 3 above, possibly related to #127

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -31,10 +31,15 @@ function Parser(options) {
   this._onError = this._options.errorHandler || console.error;
   this._fileCache = {};
   this.dynamicIncludeSrc = [];
+  this.staticIncludeSrc = [];
 }
 
 Parser.prototype.getDynamicIncludeSrc = function () {
   return _.clone(this.dynamicIncludeSrc);
+};
+
+Parser.prototype.getStaticIncludeSrc = function () {
+  return _.clone(this.staticIncludeSrc);
 };
 
 Parser.prototype._preprocess = function (element, context, config) {
@@ -67,6 +72,9 @@ Parser.prototype._preprocess = function (element, context, config) {
     if (isUrl) {
       return element; // only keep url path for dynamic
     }
+
+    // is static include file
+    this.staticIncludeSrc.push({from: context.cwf, to: filePath});
 
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 


lib/parser.js
```diff
@@ -197,6 +202,7 @@ Parser.prototype._parse = function (element, context) {
         if (fs.statSync(element.attribs.src).isFile()) {
           let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
           element.attribs.src = path.join(resultDir, utils.setExtension(path.basename(element.attribs.src), '._include_.html'));
+
```
acjh: Remove newline.

lib/parser.js
```diff
@@ -250,7 +256,6 @@ Parser.prototype._parse = function (element, context) {
         self._parse(child, context);
       });
     }
-
```
acjh: Restore newline.

nicholaschuayunzhi:
@acjh do you think it will be better if i put the filePath in the dynamicIncludeSrc as another object property?

acjh:
I'm not sure what you mean.

nicholaschuayunzhi:
Instead of maintaining a separate array `this.includedFiles`, i can add the file path to dynamic include src.
```js
this.dynamicIncludeSrc.push({from: context.cwf, to: element.attribs.src, filePath});
```

Then in `markbind-cli`, i will use the `dynamicIncludeSrc` to get the included filePath information

acjh:
It's referring to the same thing. Looks like no change is needed in this repo :)

nicholaschuayunzhi:
You're right! :O.
Was confused by https://github.com/nicholaschuayunzhi/markbind/blob/fda9cd4d986e07a153f049ab3b718e611fb16e46/lib/parser.js#L57-L68

Realised `element.attrib.src` was reassigned to `filePath` as well!

nicholaschuayunzhi:
Actually the above is not working correctly. Previously, I will account for both static and dynamic includes. However, if use the dynamicIncludeSrc, then it won't reload for static included files.

acjh:
How about having `staticIncludeSrc` array? Seems more meaningful.

nicholaschuayunzhi:
I agree too. I believe this can useful when doing build optimization. I've omitted the `from` property as it its not needed at the moment.

acjh:
> I've omitted the `from` property as it is not needed at the moment.

I disagree. Its "shape" should be consistent with `dynamicIncludeSrc` — an array of edges.

PR #129, opened 2018-01-26
Better error reporting for missing files #120

### LINKED ISSUE #120: Better error reporting for missing files
When a file is deleted/renamed, it can lead to broken file references elsewhere. These errors are reported during the rendering process. However, the error only states which file is missing, not which file has the broken reference; showing both can help the user locate the error faster.

rachx:
Fixes #120 

When an included file cannot be found, report the missing referencing file in the console and the site as well.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -75,8 +75,10 @@ Parser.prototype._preprocess = function (element, context, config) {
         self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
     } catch (e) {
       // Read file fail
+      let missingReferenceError = ` Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
       this._onError(e);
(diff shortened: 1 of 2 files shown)


lib/parser.js
```diff
@@ -75,8 +75,10 @@ Parser.prototype._preprocess = function (element, context, config) {
         self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
     } catch (e) {
       // Read file fail
+      let missingReferenceError = ` Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
```
acjh: - Remove space in <code>` M</code >
- Rename to `missingReferenceErrorMessage`
- How about <code>e.message += \`\n${missingReferenceErrorMessage}`;</code>?

PR #130, opened 2018-01-31
Support reuse of boilerplate files #110

### LINKED ISSUE #110: Support reuse of boilerplate files

rachx:
Fixes #110 

If a statically/dynamically included file cannot be found, look for the file into the _boilerplates folder. The end result is location sensitive.

This is a replacement for https://github.com/MarkBind/markbind/pull/125. I checked out one of my earlier commits and created a branch from there. Can I close that issue?

**How to test**:

1. Update the local repository with code from this PR and [markbind CLI PR](https://github.com/MarkBind/markbind-cli/pull/20) .
2. In CS2103 website, create a folder _boilerplates (on the same level as common).
3. Move the full.md file from `architecture/architecturalStyles/clientServer/what/full.md` to the `_boilerplates` folder. Other similar full.md files in other architecture folders can be deleted as well.
4. Run `markbind serve` and check the affected pages. For example, the architecture page should show the correct and different content for different architectural styles.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -76,9 +76,6 @@ Parser.prototype._preprocess = function (element, context, config) {
       return element; // only keep url path for dynamic
     }
 
-    // is static include file
-    this.staticIncludeSrc.push({from: context.cwf, to: filePath});
-
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
     let actualFilePath = filePath;
@@ -86,6 +83,9 @@ Parser.prototype._preprocess = function (element, context, config) {
       let boilerplateFilePath = calculateBoilerplateFilePath(filePath, config);
       actualFilePath = utils.fileExists(boilerplateFilePath) ? boilerplateFilePath : actualFilePath;
     }
+
+    // is static include file
+    this.staticIncludeSrc.push({from: context.cwf, to: actualFilePath});
 
     try {
       self._fileCache[actualFilePath] = self._fileCache[actualFilePath] ?
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -19,6 +19,9 @@ const nunjucks = require('nunjucks');
 
 const ATTRIB_INCLUDE_PATH = 'include-path';
 const ATTRIB_CWF = 'cwf';
+
+const BOILERPLATE_DIRECTORY_NAME = '_boilerplates';
+
 /*
  * Utils
  */
@@ -78,9 +81,14 @@ Parser.prototype._preprocess = function (element, context, config) {
 
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
+    let boilerplateFilePath = calculateBoilerplateFilePath(includeSrcPath);
+    let actualFilePath = (utils.fileExists(filePath) || !utils.fileExists(boilerplateFilePath))
+                       ? filePath
+                       : boilerplateFilePath;
+
     try {
-      self._fileCache[filePath] = self._fileCache[filePath] ?
-        self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
+      self._fileCache[actualFilePath] = self._fileCache[actualFilePath] ?
+        self._fileCache[actualFilePath] : fs.readFileSync(actualFilePath, 'utf8');
     } catch (e) {
       // Read file fail
       let missingReferenceErrorMessage = `Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
@@ -96,7 +104,7 @@ Parser.prototype._preprocess = function (element, context, config) {
       element.name = 'markdown';
     }
 
-    let fileContent = self._fileCache[filePath]; // cache the file contents to save some I/O
+    let fileContent = self._fileCache[actualFilePath]; // cache the file contents to save some I/O
     delete element.attribs.src;
     delete element.attribs.inline;
 
@@ -204,7 +212,9 @@ Parser.prototype._parse = function (element, context) {
     case 'dynamic-panel': {
       let hasIsOpen = _.hasIn(element.attribs, 'isOpen');
       element.attribs.isOpen = hasIsOpen ? hasIsOpen : false;
-      if (utils.fileExists(element.attribs.src)) {
+      let boilerplateFilePath = calculateBoilerplateFilePath(element.attribs.src);
+      let fileExists = utils.fileExists(element.attribs.src) || utils.fileExists(boilerplateFilePath);
+      if (fileExists) {
         let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
         element.attribs.src = path.join(resultDir, utils.setExtension(path.basename(element.attribs.src), '._include_.html'));
         if (element.attribs.fragment) {
@@ -220,7 +230,9 @@ Parser.prototype._parse = function (element, context) {
       if (!_.hasIn(element.attribs, 'src')) {
         break;
       }
-      if (utils.fileExists(element.attribs.src)) {
+      let boilerplateFilePath = calculateBoilerplateFilePath(element.attribs.src);
+      let fileExists = utils.fileExists(element.attribs.src) || utils.fileExists(boilerplateFilePath);
+      if (fileExists) {
         let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
         element.attribs.src = path.join(resultDir, utils.setExtension(path.basename(element.attribs.src), '._include_.html'));
         if (element.attribs.fragment) {
@@ -233,7 +245,9 @@ Parser.prototype._parse = function (element, context) {
       if (!_.hasIn(element.attribs, 'src')) { // dynamic panel
         break;
       }
-      if (utils.fileExists(element.attribs.src)) {
+      let boilerplateFilePath = calculateBoilerplateFilePath(element.attribs.src);
+      let fileExists = utils.fileExists(element.attribs.src) || utils.fileExists(boilerplateFilePath);
+      if (fileExists) {
         let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
         element.attribs.src = path.join(resultDir, utils.setExtension(path.basename(element.attribs.src), '._include_.html'));
         if (element.attribs.fragment) {
@@ -297,8 +311,13 @@ Parser.prototype.includeFile = function (file, config) {
       decodeEntities: true
     });
 
+    let boilerplateFilePath = calculateBoilerplateFilePath(file);
+    let actualFilePath = (utils.fileExists(file) || !utils.fileExists(boilerplateFilePath))
+                       ? file
+                       : boilerplateFilePath;
+
     // Read files
-    fs.readFile(file, (err, data) => {
+    fs.readFile(actualFilePath, (err, data) => {
       if (err) {
         reject(err);
         return;
@@ -475,6 +494,10 @@ Parser.prototype._rebaseReferenceForStaticIncludes = function (pageData, element
   let newBase = fileBase.relative;
   return nunjucks.renderString(pageData, { baseUrl: `{{hostBaseUrl}}/${newBase}` });
 };
+
+function calculateBoilerplateFilePath(filePath) {
+  return path.resolve(path.join(process.cwd(), BOILERPLATE_DIRECTORY_NAME), path.basename(filePath));
+}
 
 function calculateNewBaseUrl(filePath, root, lookUp) {
   function calculate(file, result) {
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -81,10 +81,11 @@ Parser.prototype._preprocess = function (element, context, config) {
 
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
-    let boilerplateFilePath = calculateBoilerplateFilePath(includeSrcPath);
-    let actualFilePath = (utils.fileExists(filePath) || !utils.fileExists(boilerplateFilePath))
-                       ? filePath
-                       : boilerplateFilePath;
+    let actualFilePath = filePath;
+    if (!utils.fileExists(filePath)) {
+      let boilerplateFilePath = calculateBoilerplateFilePath(filePath, config);
+      actualFilePath = utils.fileExists(boilerplateFilePath) ? boilerplateFilePath : actualFilePath;
+    }
 
     try {
       self._fileCache[actualFilePath] = self._fileCache[actualFilePath] ?
@@ -186,11 +187,11 @@ Parser.prototype._preprocess = function (element, context, config) {
   return element;
 };
 
-Parser.prototype._parse = function (element, context) {
+Parser.prototype._parse = function (element, context, config) {
   let self = this;
   if (_.isArray(element)) {
     return element.map((el) => {
-      return self._parse(el, context);
+      return self._parse(el, context, config);
     });
   }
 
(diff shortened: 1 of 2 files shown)


lib/parser.js
```diff
@@ -283,6 +322,10 @@ Parser.prototype.includeFile = function (file, config) {
   context.cwf = file; // current working file
   context.mode = 'include';
 
+  if (config) {
+    context.cwf = config.context || conext.cwf;
```
acjh: This is confusing (and `conext` is mispelled). The same value is sometimes called `actualFilePath`, `actual`, `alternateSource` and `context`. If we stick to `actualFilePath`, let's rename `filePath` to `targetFilePath`.

rachx: For consistency, how about
* rename `filePath` to `targetFilePath`
* remove  `alternateSource` and use `element.attribs[ATTRIB_BOILERPLATE_PATH]` directly
* `this.dynamicIncludeSrc.push({from: context.cwf, to: filePath, actual: actualFilePath});` ->  `this.dynamicIncludeSrc.push({from: context.cwf, target: targetFilePath, actualFilePath: actualFilePath});`
* Use `config.targetFilePath` instead

acjh: > For consistency, how about
> * rename `filePath` to `targetFilePath`

Let's try that.

> * remove  `alternateSource` and use `element.attribs[ATTRIB_BOILERPLATE_PATH]` directly

You can assign a variable named `boilerplateFilePath` in the `catch` block.

> * `this.dynamicIncludeSrc.push({from: context.cwf, to: filePath, actual: actualFilePath});` ->  `this.dynamicIncludeSrc.push({from: context.cwf, target: targetFilePath, actualFilePath: actualFilePath});`

Let's try that.

> * Use `config.targetFilePath` instead

```js
context.cwf = (config && config.targetFilePath) || file; // current working file
```

lib/parser.js
```diff
@@ -47,32 +49,37 @@ Parser.prototype._preprocess = function (element, context, config) {
     let isUrl = utils.isUrl(element.attribs.src);
     let includeSrc = url.parse(element.attribs.src);
     let includeSrcPath = includeSrc.path;
-    let filePath = isUrl ? element.attribs.src : path.resolve(path.dirname(context.cwf), includeSrcPath);
+    let targetFilePath = isUrl ? element.attribs.src : path.resolve(path.dirname(context.cwf), includeSrcPath);
+    let actualFilePath = targetFilePath;
+    let boilerplateFilePath = calculateBoilerplateFilePath(includeSrcPath);
+    if (!fs.existsSync(targetFilePath) && fs.existsSync(boilerplateFilePath)) {
+      actualFilePath = boilerplateFilePath;
+    }
```
acjh: Do we need `actualFilePath` here? See next comment.

lib/parser.js
```diff
@@ -47,32 +49,37 @@ Parser.prototype._preprocess = function (element, context, config) {
...
+    let isIncludeSrcMd = utils.getExtName(targetFilePath) === 'md';
 
     try {
-      self._fileCache[filePath] = self._fileCache[filePath] ?
-        self._fileCache[filePath] : fs.readFileSync(filePath, 'utf8');
+      self._fileCache[actualFilePath] = self._fileCache[actualFilePath] ?
+        self._fileCache[actualFilePath] : fs.readFileSync(actualFilePath, 'utf8');
     } catch (e) {
       // Read file fail
       let missingReferenceErrorMessage = `Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
```
acjh: How about:

```js
try {
  self._fileCache[targetFilePath] = self._fileCache[targetFilePath] || fs.readFileSync(targetFilePath, 'utf8');
} catch (e) {
  // Read file fail: targetFilePath
  let boilerplateFilePath = calculateBoilerplateFilePath(includeSrcPath);
  
  try {
    self._fileCache[targetFilePath] = fs.readFileSync(boilerplateFilePath, 'utf8');
  } catch (e) {
    // Read file fail: boilerplateFilePath
    let missingReferenceErrorMessage = `Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
      e.message += `\n${missingReferenceErrorMessage}`;
      this._onError(e);
      element.children = cheerio.parseHTML(utils.createErrorElement(e), true);
      return element;
  }
}
```

lib/parser.js
```diff
@@ -280,7 +298,7 @@ Parser.prototype._trimNodes = function (node) {
 
 Parser.prototype.includeFile = function (file, config) {
   let context = {};
-  context.cwf = file; // current working file
+  context.cwf = (config && config.targetFilePath) || file; // current working file
```
acjh: If we are prioritising `targetFilePath` (which always has value) and so never using `file` ([which is `actualFilePath`](https://github.com/MarkBind/markbind-cli/pull/20/commits/a05647fc5842ed5dd3491c803d6adaca576a0cf3)), then why are we passing `actualFilePath` around?

rachx: There is no config for https://github.com/MarkBind/markbind-cli/blob/master/lib/Page.js#L67. It uses `file`.

I was thinking if it is possible to not change markbind cli code and keep `this.dynamicIncludeSrc.push({from: context.cwf, to: filePath`. Then check for the boilerplate file only in `Parser.prototype.includeFile`. It doesn't seem to work for now

acjh: You modified markbind-cli to pass `file` as `config.targetFilePath` and `actualFilePath` as `file`, then we use `config.targetFilePath` (i.e. `file`) so we're never using `file` (i.e. `actualFilePath`).

Then, why are we passing `actualFilePath` to this function as `file`?

rachx: The `targetFilePath` is the local folder where the file is supposed to be. It is `file = source.target` in `resolveDependency` in markbind cli and is used to determine the `tempPath` and `resultPath` to save. It is also passed in `config.targetFilePath` so that the context of the children of dynamically included boilerplates files can be set to the local folder too.

The `actualFilePath` is where the file exists - it can be in the boilerplate folder. It is read in  `includeFile` in `fs.readFile(file, (err, data)`.  If we use `targetFilePath` and try to read from it in (unmodified) `includeFile` , it will throw an error.

acjh: Thanks for clarifying. Let's rename `file` to `targetFilePath` in `Page.prototype.resolveDependency`:

```diff
- const file = dependency.target;
+ const targetFilePath = dependency.target;
  const actualFilePath = dependency.actualFilePath;

  return markbinder.includeFile(actualFilePath, {
    baseUrlMap: this.baseUrlMap,
    rootPath: this.rootPath,
-   targetFilePath: file,
+   targetFilePath: targetFilePath
  })
```

lib/parser.js
acjh: Change `fs.existsSync` to `utils.fileExists`:

```js
let actualFilePath = (utils.fileExists(targetFilePath) || !utils.fileExists(boilerplateFilePath))
                   ? targetFilePath
                   : boilerplateFilePath;
```

lib/parser.js
```diff
@@ -496,4 +514,8 @@ function calculateNewBaseUrl(filePath, root, lookUp) {
   return calculate(filePath, []);
 }
 
+function calculateBoilerplateFilePath(targetFilePath) {
+  return path.resolve(path.join(process.cwd(), BOILERPLATE_DIRECTORY_NAME), path.basename(targetFilePath));
+}
+
```
acjh: Move this above `calculateNewBaseUrl` (alphabetical order).

nicholaschuayunzhi: > ```diff
> - const file = dependency.target;
> + const targetFilePath = dependency.target;
> const actualFilePath = dependency.actualFilePath;
>
>  return markbinder.includeFile(actualFilePath, {
>    baseUrlMap: this.baseUrlMap,
>    rootPath: this.rootPath,
> -   targetFilePath: file,
> +   targetFilePath: targetFilePath
>  })
> ```

can actually do

```diff
 return markbinder.includeFile(actualFilePath, {
   baseUrlMap: this.baseUrlMap,
    rootPath: this.rootPath,
-   targetFilePath: file,
+   targetFilePath
  })
```

acjh: @nicholaschuayunzhi You can move your comment [here](https://github.com/MarkBind/markbind/pull/130#discussion_r165357626) :)

lib/parser.js
nicholaschuayunzhi: > ```diff
> - const file = dependency.target;
> + const targetFilePath = dependency.target;
> const actualFilePath = dependency.actualFilePath;
>
>  return markbinder.includeFile(actualFilePath, {
>    baseUrlMap: this.baseUrlMap,
>    rootPath: this.rootPath,
> -   targetFilePath: file,
> +   targetFilePath: targetFilePath
>  })
> ```

can actually do

```diff
 return markbinder.includeFile(actualFilePath, {
   baseUrlMap: this.baseUrlMap,
    rootPath: this.rootPath,
-   targetFilePath: file,
+   targetFilePath
  })
```

rachx:
It now works without changing markbind cli. (So test with cli master instead). 😅

I replaced the `actualFilePath` in `includeFile`. This achieves the same effect as checking beforehand and passing the correct file to `includeFile`.
(I thought it didn't work but that was because I forgot to change `require(markbind)` after pulling changes)

If this is okay, I will close the other PR, refactor the commits and rebase on master.

lib/parser.js
rachx: I used `path.basename(targetFilePath)`. This means that it won't support having sub folders within the `_boilerplates` folder.

lib/parser.js
rachx: Should this still be changed?

I think it should be
`self._fileCache[boilerplateFilePath] = fs.readFileSync(boilerplateFilePath, 'utf8');`
and the error reported should be the first one

acjh: Hmm, since we're using the same pattern for `actualFilePath` in `includeFile`, let's not change this.

acjh:
Great :)

Yes, change `targetFilePath` back to `filePath` since `actualFilePath` is used locally.

lib/parser.js
```diff
@@ -195,17 +202,15 @@ Parser.prototype._parse = function (element, context) {
...
+      if (utils.fileExists(element.attribs.src)) {
+        let resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(process.cwd(), element.attribs.src)));
+        element.attribs.src = path.join(resultDir, utils.setExtension(path.basename(element.attribs.src), '._include_.html'));
+        if (element.attribs.fragment) {
+          element.attribs.src = `${element.attribs.src}#${element.attribs.fragment}`; // add hash back to path
         }
-      } catch (e) {
-        // Pass: src is not on filesystem (invalid or remote)
       }
+
```
acjh: No need for the newlines around this `if` block.

lib/parser.js
```diff
@@ -19,6 +19,8 @@ const nunjucks = require('nunjucks');
 
 const ATTRIB_INCLUDE_PATH = 'include-path';
 const ATTRIB_CWF = 'cwf';
+
+const BOILERPLATE_DIRECTORY_NAME = '_boilerplates';
```
acjh: Add a newline after this.

acjh:
Hmm, 2 problems.

1. Diff on Mac:


2. The `_boilerplates` should be in the site's root (i.e. `book/_boilerplates`), not the host's root.

damithc:
Also note that there can be multiple _boilerplates folders in case the website contains submodules which are also markbind sites (e.g., cs2103 website)

rachx:
>1.  Diff on Mac:

I think this is fixed. I shouldn't have removed ` let fileExists = utils.fileExists(element.attribs.src) || utils.fileExists(boilerplateFilePath);`

Gisonrg:
> the book site should use its own _boilerplates folder instead of the CS2103's website one?

Might be so, like the `site.json` case, the book site could itself be a standalone site but get included in the cs2103 website.

acjh:
>> 1.  Diff on Mac:
>
> I think this is fixed - the diff shouldn't occur. I shouldn't have removed ` let fileExists = utils.fileExists(element.attribs.src) || utils.fileExists(boilerplateFilePath);`

Verified.

>> 2. The _boilerplates should be in the site's root (i.e. book/_boilerplates), not the host's root.
>
> Is the `calculateBoilerplateFilePath` method incorrect?
Is this about 2103 website - the book site should use its own `_boilerplates` folder instead of the CS2103's website one?

Yes, as explained by @Gisonrg.

rachx:
> The _boilerplates should be in the site's root (i.e. book/_boilerplates), not the host's root.  

I have looked into it but I can't figure it out yet.

Any suggestions on how to get the site's root/base url of the `filePath` ? (e.g.  `book/testing/testingTypes/integrationTesting/what/index.md` -> `book` )? Is this related to the `calculateNewBaseUrl` function?

damithc:
Not sure if it helps but a site root folder has a file named `site.json`

rachx:
Follows https://github.com/MarkBind/markbind-cli/pull/25

It should correctly use `book/_boilerplates` now

lib/parser.js
```diff
@@ -70,9 +75,15 @@ Parser.prototype._preprocess = function (element, context, config) {
 
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
+    let actualFilePath = filePath;
+    if (!utils.fileExists(filePath)) {
```
rachx: This is to avoid unnecessary calls to calculateBoilerplateFilePath

lib/parser.js
```diff
@@ -487,7 +488,11 @@ Parser.prototype._rebaseReferenceForStaticIncludes = function (pageData, element
 };
 
 function calculateBoilerplateFilePath(filePath) {
-  return path.resolve(path.join(process.cwd(), BOILERPLATE_DIRECTORY_NAME), path.basename(filePath));
+  let possibleBases = Object.keys(Parser.baseUrlMap).filter(base => pathIsInside(filePath, base));
+  let longest = possibleBases.sort(function (a, b) {
+    return b.length - a.length;
+  })[0];
```
acjh: Seems related: `baseUrlFromRoot` in [Page.js#L35-L50](https://github.com/MarkBind/markbind-cli/blob/1ff54e85f239778aed8c3ba4567759cc8f3f3fca/lib/Page.js#L35-L50)

rachx: Is that the same as https://github.com/MarkBind/markbind/blob/master/lib/parser.js#L483?

acjh: Similar, but not the same. They should probably be renamed:
- markbind's parser.js: `calculateNewBaseUrl` → `calculateNewBaseUrlMap`
- markbind-cli's Page.js: `baseUrlFromRoot` → `calculateNewBaseUrl`

Not sure if the implementation can be improved, or if your algorithm is a comparable alternative.

rachx: Edited to use `calculateNewBaseUrl` from `parser.js` and it seems to work.
But there might be a simpler way to implement both functions

lib/parser.js
```diff
@@ -31,6 +34,8 @@ function Parser(options) {
   this._onError = this._options.errorHandler || console.error;
   this._fileCache = {};
   this.dynamicIncludeSrc = [];
+  Parser.baseUrlMap = this._options.baseUrlMap;
+  Parser.rootPath = this._options.rootPath;
```
acjh: These should be passed to `includeFile` and `renderFile` as `config` to better trace its use.
See `includeFile` in markbind-cli's `Page.prototype.resolveDependency`: [Page.js#L126-L129](https://github.com/MarkBind/markbind-cli/blob/1ff54e85f239778aed8c3ba4567759cc8f3f3fca/lib/Page.js#L126-L129)

acjh: @rachx You can remove this after commit 99e0ce2

rachx:
Force pushed

acjh:
Can you rebase these commits?

rachx:
Rebased

acjh:
The [third](https://github.com/rachx/markbind/commit/008a5eac26aef02ea491f191ab17175a4544ea45), [fourth](https://github.com/MarkBind/markbind/pull/130/commits/2ab86fa4a896243bc358f33788731dd663f59b44) and [fifth](https://github.com/MarkBind/markbind/pull/130/commits/463b57dcd67224ebc341fa4618a387c6af4ffd4f) commits can be squashed.

rachx:
Updated

lib/parser.js
```diff
@@ -78,9 +81,15 @@ Parser.prototype._preprocess = function (element, context, config) {
 
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
+    let actualFilePath = filePath;
+    if (!utils.fileExists(filePath)) {
+      let boilerplateFilePath = calculateBoilerplateFilePath(filePath, config);
+      actualFilePath = utils.fileExists(boilerplateFilePath) ? boilerplateFilePath : actualFilePath;
+    }
+
```
nicholaschuayunzhi: Can we bring line 77 
```js
 this.staticIncludeSrc.push({from: context.cwf, to: filePath});
```
and change to 
```js
 this.staticIncludeSrc.push({from: context.cwf, to: actualFilePath});
```
So that we capture both boilerplate and non boilerplate files?

rachx: Done

damithc:
🎉 Looking forward to using this. :-)

PR #142, opened 2018-02-09
Update parser to allow for in-line rendering of Markdown content

### LINKED ISSUE #132: Stop adding a line break for output when rendering <markdown> content
For example, this,
`some text <span id="def-client-code"><markdown>**bold** other text 
</markdown></span>`

should render as,

>some text **bold** other text 

and not like this:

>some text
>**bold** other text  

Also document the use of  `<markdown>` tag in the user guide.

danielbrzn:
Resolves #132 

Content enclosed within `<markdown>` tags were being rendered as a block element, which resulted in an unintended line break being added between content as shown below:
image

As such, this PR proposes to allow in-line rendering when the `<md>` tag is used, so that unnecessary line breaks will not be added unless the author specifically wants it. The output will then be as follows:
image

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -187,6 +187,11 @@ Parser.prototype._parse = function (element, context) {
 
     switch (element.name) {
     case 'md':
+      element.name = 'span';
+      cheerio.prototype.options.xmlMode = false;
+      element.children = cheerio.parseHTML(md.renderInline(cheerio.html(element.children)), true);
+      cheerio.prototype.options.xmlMode = true;
+      break;
     case 'markdown':
       element.name = 'div';
       cheerio.prototype.options.xmlMode = false;


lib/parser.js
```diff
@@ -187,6 +187,11 @@ Parser.prototype._parse = function (element, context) {
 
     switch (element.name) {
     case 'md':
+      element.name = 'span'
```
nicholaschuayunzhi: Remember to add semi colon

nicholaschuayunzhi:
> As such, this PR proposes to allow in-line rendering when the <md> tag is used, so that unnecessary line breaks will not be added unless the author specifically wants it.

To clarify, does it mean that if the author does not specify <md> or <markdown> tags, it will render inline?

lib/parser.js
danielbrzn: Thanks!

danielbrzn:
> To clarify, does it mean that if the author does not specify or tags, it will render inline?

If tags are not specified by the author, Markdown content will be rendered as block elements. This is because currently, Markbind wraps Markdown content within `markdown` tags which have the normal rendering behaviour instead of inline rendering.

PR #144, opened 2018-02-10
Record boilerplate includes separately

nicholaschuayunzhi:
This addresses the issue where boilerplate (bp) files that are dynamically included are not reloaded.

I have separated bp includes into a separate array for both static and dynamic.
Had to separate dynamic bps from `dynamicIncludeSrc` as `resolveDependency` relies on and iterates through it to handle dynamic includes.

Follows #110,  #128, https://github.com/MarkBind/markbind-cli/pull/17, https://github.com/MarkBind/markbind-cli/pull/25

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -88,8 +88,10 @@ Parser.prototype._preprocess = function (element, context, config) {
       this.staticIncludeSrc.push({from: context.cwf, to: actualFilePath});
     } else {
       let boilerplateFilePath = calculateBoilerplateFilePath(filePath, config);
-      actualFilePath = utils.fileExists(boilerplateFilePath) ? boilerplateFilePath : actualFilePath;
-      this.boilerplateIncludeSrc.push({from: context.cwf, to: actualFilePath});
+      if (utils.fileExists(boilerplateFilePath)) {
+        actualFilePath = boilerplateFilePath;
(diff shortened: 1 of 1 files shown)


lib/parser.js
```diff
@@ -79,14 +84,14 @@ Parser.prototype._preprocess = function (element, context, config) {
     let isIncludeSrcMd = utils.getExtName(filePath) === 'md';
 
     let actualFilePath = filePath;
-    if (!utils.fileExists(filePath)) {
+    if (utils.fileExists(filePath)) {
+      this.staticIncludeSrc.push({from: context.cwf, to: actualFilePath});
+    } else {
       let boilerplateFilePath = calculateBoilerplateFilePath(filePath, config);
       actualFilePath = utils.fileExists(boilerplateFilePath) ? boilerplateFilePath : actualFilePath;
+      this.boilerplateIncludeSrc.push({from: context.cwf, to: actualFilePath});
```
rachx: This will add the original file path if the boilerplate file does not exist.
Should it be the following?
```js
if (utils.fileExists(boilerplateFilePath)) {
  actualFilePath = boilerplateFilePath;
  this.boilerplateIncludeSrc.push({from: context.cwf, to: boilerplateFilePath});
 }
```

nicholaschuayunzhi: Thanks for pointing this out!

Issue #143, opened 2018-02-10
Support user-defined variables


damithc:
Currently we support a special variable `{{baseUrl}}`
It would be useful to allow the user to define more such variables where the value can be any markbind-compliant code fragment.

e.g., we can have a file called `_system\variables.md` which has a format like this:
```
<span id="year">2018</span>

<span id="options">
* yes
* no
* maybe
</span>
```
And the normal files can use `{{year}}` `{{options}}` etc.

A feature like this can help to reduce the verbosity of the code.

rachx:
I am working on this. My current idea is to use regex to match and replace the content before parsing is done.
How about
```
{{options}}
<span id="options">
<markdown>
* yes
* no
* maybe
</markdown>
</span>
{{/options}}
```

damithc:
If we enforce that all variables for a site will be in a known location `[root]\_system\variables.md`, we don't need the tags `{{options}}..{{/options}}`?. We can simply go by span/div IDs to identify the variables and their values. Would that work?

acjh:
> My current idea is to use regex to match and replace the content before parsing is done.

Use [`nunjucks.renderString`](https://mozilla.github.io/nunjucks/api.html#renderstring) instead of regex:

https://github.com/MarkBind/markbind/blob/921753440d8e38f77bbe0d183695b4cda5a7664e/lib/parser.js#L494

> We can simply go by span/div IDs to identify the variables and their values. Would that work?

Seems possible with [cheerio](https://github.com/cheeriojs/cheerio), which MarkBind is already using. More work, but definitely cleaner.

damithc:
Started used this feature. Defined this text that appears several times in [this website](https://nus-te3201.github.io/website/programming/index.html)
```html
<span style="float:right">[<span class="glyphicon glyphicon-home" aria-hidden="true"></span> ↑](../index.html)</span>
```
that shows this 'back to home' link
![image](https://user-images.githubusercontent.com/1673303/37272064-6faa2ac8-2610-11e8-8244-7ad4b56eb034.png)
as variable`{{home↑}}`
It worked despite the use of the arrow. Nice. 💯

PR #145, opened 2018-02-12
Support user-defined variables #143

rachx:
Fixes https://github.com/MarkBind/markbind/issues/143

### Testing
1. Create a folder in the root `_system`.
2. Create a file `variables.md` in the folder with the content
```
<span id="year">2018</span>

<span id="options">
* yes
* no
* maybe
</span>
```
(This example works fine but sometimes they need to be protected by the markdown tags)
3. In any file in the site, use {{options}} or {{year}}

I have tested and it seems to be working.

But in `userDefinedVariables`, there is no `baseUrl` (i.e. undefined).
When I use `nunjucks.renderString` in `includeFile` and `_preprocess`  to add code fragments, it may also replace `baseUrl` with nothing. Will this affect future reference rebasing?

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -21,7 +21,6 @@ const ATTRIB_INCLUDE_PATH = 'include-path';
 const ATTRIB_CWF = 'cwf';
 
 const BOILERPLATE_DIRECTORY_NAME = '_boilerplates';
-const USER_VARIABLES_PATH = '_system/variables.md';
 
 /*
  * Utils
@@ -51,7 +50,7 @@ Parser.prototype.getBoilerplateIncludeSrc = function () {
   return _.clone(this.boilerplateIncludeSrc);
 };
 
-Parser.prototype._preprocess = function (element, context, config, userDefinedVariables) {
+Parser.prototype._preprocess = function (element, context, config) {
   let self = this;
   element.attribs = element.attribs || {};
   element.attribs[ATTRIB_CWF] = path.resolve(context.cwf);
@@ -114,6 +113,8 @@ Parser.prototype._preprocess = function (element, context, config, userDefinedVa
     }
 
     let fileContent = self._fileCache[actualFilePath]; // cache the file contents to save some I/O
+    let fileBase = path.resolve(calculateNewBaseUrl(filePath, config.rootPath, config.baseUrlMap).relative);
+    let userDefinedVariables = config.userDefinedVariablesMap[fileBase];
     fileContent = nunjucks.renderString(fileContent, userDefinedVariables);
     delete element.attribs.src;
     delete element.attribs.inline;
@@ -150,7 +151,7 @@ Parser.prototype._preprocess = function (element, context, config, userDefinedVa
     childContext.source = isIncludeSrcMd ? 'md' : 'html';
     if (element.children && element.children.length > 0) {
       element.children = element.children.map((e) => {
-        return self._preprocess(e, childContext, config, userDefinedVariables);
+        return self._preprocess(e, childContext, config);
       });
     }
   } else if (element.name === 'dynamic-panel') {
@@ -188,7 +189,7 @@ Parser.prototype._preprocess = function (element, context, config, userDefinedVa
   } else {
     if (element.children && element.children.length > 0) {
       element.children = element.children.map((e) => {
-        return self._preprocess(e, context, config, userDefinedVariables);
+        return self._preprocess(e, context, config);
       });
     }
   }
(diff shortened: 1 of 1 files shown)


damithc:
> But in userDefinedVariables, there is no baseUrl (i.e. undefined).

From the user's POV, `baseUrl` is a pre-defined variable. In future, we might have more such variables.

Note that there is a complication when the same variable is defined in a sub-site(s) as well as host-site.

rachx:
> Note that there is a complication when the same variable is defined in a sub-site(s) as well as host-site.

Yes, I realized it is the same thing as the boilerplate files. Like the boilerplate version, it currently use the sub site's `variables.md`.


> From the user's POV, baseUrl is a pre-defined variable. In future, we might have more such variables.  

If the other variables are constant and can be replaced easily, we can just merge both dictionaries/objects to be used by `nunjucks`.
But the `baseUrl` is recomputed in the code.

lib/parser.js
```diff
@@ -299,19 +301,42 @@ Parser.prototype._trimNodes = function (node) {
   }
 };
 
+Parser.prototype._initializeUserVariables = function(filePath, config) {
+  const userDefinedVariables = {};
+  userDefinedVariables['baseUrl'] = '{{baseUrl}}';
```
acjh: 1. Does this correctly solve the `baseUrl` problem?
2. The next 2 lines are about `baseUrl` for a different purpose — is that why it's named `base`?
   Let's place this line (with a comment) just before `return userDefinedVariables;` instead.

rachx: 1. Yes. I tested printing {{ baseUrl }} in the site with and without this line.
2. The next 2 lines are for using the correct variables.md if there is a subsite. I will move the first line down

lib/parser.js
```diff
@@ -299,19 +301,47 @@ Parser.prototype._trimNodes = function (node) {
...
+Parser.prototype._initializeUserVariables = function(filePath, config) {
+  const userDefinedVariables = {};
+  const baseDirectoryForFile = calculateNewBaseUrl(filePath, config.rootPath, config.baseUrlMap).relative;
+  const definedVariablePath = path.resolve(baseDirectoryForFile, USER_VARIABLES_PATH);
+  let content;
+  try {
+    content = fs.readFileSync(definedVariablePath, 'utf8');
+  } catch (e) {
+    content = '';
+    this._onError(e);
```
acjh: Must this file always exist?



Currently, the terminal is spammed with the above error messages. I'd rather create a blank file after `_onError`.

rachx: How about creating the file in markbind cli or checking if the file does not exist and warn once in cli (then remove this error check)?

rachx:
I think I might still need to make some changes
* When a website includes a file X from a sub-site, we should use the sub-site's `variables.md` instead of the  website.
* Can try to read the website and sub-site's `variables.md` once per Parser instead of on every include

lib/parser.js
acjh: Sure.

> I think I might still need to make some changes
> * When a website includes a file X from a sub-site, we should use the sub-site's `variables.md` instead of the  website.
> * Can try to read the website and sub-site's `variables.md` once per Parser instead of on every include

You can do it similar to [`Site.prototype.collectBaseUrl`](https://github.com/MarkBind/markbind-cli/blob/8a0871d96f5ab12fb42afe352dd8b53691d04cd1/lib/Site.js#L173-L186).
Then pass `userDefinedVariables` in [`markbinder.includeFile`](https://github.com/MarkBind/markbind-cli/blob/6b04530a561b28015450beaf2af6decb1c4d4695/lib/Page.js#L81) as part of the [`config`](https://github.com/markbind/markbind/blob/215cfca4ee546bb32c95034b2c2f18d7cef3d703/lib/parser.js#L302) parameter.

lib/parser.js
```diff
@@ -113,6 +113,9 @@ Parser.prototype._preprocess = function (element, context, config) {
     }
 
     let fileContent = self._fileCache[actualFilePath]; // cache the file contents to save some I/O
+    let fileBase = path.resolve(calculateNewBaseUrl(filePath, config.rootPath, config.baseUrlMap).relative);
```
acjh: Why is `path.resolve` used here but not in `includeFile`?

lib/parser.js
```diff
@@ -113,6 +113,9 @@ Parser.prototype._preprocess = function (element, context, config) {
...
+    let userDefinedVariables = config.baseUrlToDefinedVariablesMap[fileBase];
```
acjh: Same as https://github.com/MarkBind/markbind-cli/pull/34#discussion_r173389028.

lib/parser.js
rachx: I forgot to add it in `includeFile`. Commit amended

damithc:
@rachx is this feature documented somewhere in user docs?

PR #146, opened 2018-02-12
Apply eslint --fix and refactor code

acjh:
Done:
- Sync ESLint with markbind-cli project
  - Restore eslint-plugin-lodash
  - Update eslint-plugin-lodash version to v2.6.1
- Add eslintignore for lib/markdown-it/*a
- Apply eslint --fix
- Fix ESLint errors manually:
  - "some"
    - `consistent-return`, `default-case`, `eqeqeq`, `no-shadow`, `no-unused-expressions`, `no-use-before-define`, `no-void`, `prefer-destructuring`
  - `lodash/import-scope`
  - `max-len`
  - `no-param-reassign`
- Refactor actualContent assignment in _preprocess
  - `no-nested-ternary`

Ignored:
- `no-underscore-dangle` in parser.js
- `no-console` in `this._onError = this._options.errorHandler || console.error;`
- `no-plusplus` in `for` loop


.eslintrc.js
```diff
@@ -3,30 +3,21 @@ module.exports = {
     "node": true,
     "es6": true
   },
-  "plugins": ["lodash"],
-  "extends": ["eslint:recommended", "plugin:lodash/recommended"],
```
nicholaschuayunzhi: I was wondering if you would want to keep the lodash plugin in this repository as it is used in this repository and the plugins will enforce us to use lodash functions.

acjh: Thanks for the suggestion! We can keep the lodash plugin, but we had (and have) `"lodash/prefer-lodash-method": [0]` so it doesn't force us to use lodash functions.

Gisonrg:
We might want to apply fix again when other PR are merged to avoid conflicts.

PR #156, opened 2018-02-24
Record missing include src

rachx:
File A includes File B which does not exist (e.g. due to a typo).
When File B is added, in the current markbind-cli, we will rebuild all files and this will update all files including A.
If we want to only rebuild file A when B is added, we should keep missing include file sources as well.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -36,6 +36,7 @@ function Parser(options) {
   this.dynamicIncludeSrc = [];
   this.staticIncludeSrc = [];
   this.boilerplateIncludeSrc = [];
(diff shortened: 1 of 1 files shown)


nicholaschuayunzhi:
https://github.com/rachx/markbind/blob/998b706dc0e928b6183ca6629017dded084a4339/lib/parser.js#L105-L115

Would we need to capture the file path here as well?

rachx:
We need to capture the missing target and boilerplate file source in either the try-catch block or when checking the file exists (No need for both)

I placed it in the else block when checking for the file exists since the other sources are collected there

PR #161, opened 2018-03-11
Use Bootstrap class for table

nicholaschuayunzhi:
Add bootstrap class for for striped and borded tables

Related https://github.com/MarkBind/markbind-cli/pull/33

### DIFF
--- a/lib/markdown-it/index.js
+++ b/lib/markdown-it/index.js
@@ -34,7 +34,7 @@ markdownIt.normalizeLink = require('./normalizeLink');
 
 // fix table style
 markdownIt.renderer.rules.table_open = (tokens, idx) => {
-  return '<table class="markbind-table table">';
+  return '<table class="markbind-table table table-bordered table-striped">';
 };
 
 // fix emoji numbers


lib/markdown-it/index.js
```diff
@@ -34,7 +34,7 @@ markdownIt.normalizeLink = require('./normalizeLink');
 
 // fix table style
 markdownIt.renderer.rules.table_open = (tokens, idx) => {
-  return '<table class="markbind-table table">';
+  return '<table class="markbind-table table-bordered table-striped table">';
```
acjh: Why did `table` become the last class?

nicholaschuayunzhi: My bad. Should it be ` markbind-table table table-bordered table-striped` instead? Based on alphabetical order.

acjh: Yes. Sometimes special classes like `markbind-table` can be excluded from the "alphabetical order" convention.

acjh:
Please squash the commits.

nicholaschuayunzhi:
Squashed! @acjh

Issue #159, opened 2018-03-10
boilerplates: allow multiple boilerplates with similar names


damithc:
Building on the current boilerplates mechanism, we can support a syntax like this:

`<include src="*/foo/bar.md" />`

`*` indicates that it is a boilerplate. The benefit of this is the reader of the code can immediately identify this as a reference to a boilerplate file rather than a normal file.

This also allows us to support multiple boilerplates with the same name. e.g., `*/full/index.md` vs `*/print/index.md`.

rachx:
How will you specify for/distinguish the above case with
```html
<panel type="info" header="`W8.3c` Can identify client-server architectural style :star::star::star:" expanded no-close>
  <include src="../../book/architecture/architecturalStyles/clientServer/what/full.md" />
<!-- TODO: add evidence -->
</panel>
```

where `full.md` is a boilerplate file and you want the location sensitive reference to be `"../../book/architecture/architecturalStyles/clientServer/what/full.md"`

damithc:
This is what I was thinking:
```html
<panel type="info" header="`W8.3c` some text" expanded no-close>
  <include src="../../book/architecture/architecturalStyles/clientServer/what/*/full.md" />
</panel>
```
In this case it will take `_boilerplates/full.md` file and act as if `full.md` was in the `....clientServer/what/` folder. i.e., same as current behavior but `*` makes it clear that we are referring to a boilerplate file.

```html
<panel type="info" header="`W8.3c` some text" expanded no-close>
  <include src="../../book/architecture/architecturalStyles/clientServer/what/*/print/full.md" />
</panel>
```
In this case it will take the `_boilerplates/print/full.md` file and act as if `full.md` was in the `....clientServer/what/` folder.

That is, folder structure inside the `_boilerplates` folder is just a way to distinguish between different boilerplates with the same file name. We could have used different files names instead too (e.g., `_boilerplates/full.md` and `_boilerplates/print_full.md` but that technique doesn't work for `index.md` which needs to be named `index`.

What do you think? Too much work for too little gain? We can also support `*` to indicate boilerplate files just to improve readability but not support sub folders inside it. In that case index.md files cannot benefit from boilerplate mechanism fully.

rachx:
I was trying to implement it.

While testing, I notice that it is hard to replace all the references in the 2103 website. Although we report missing references for static include, it doesn't report for dynamic include and it took me a while to find it. 
Also since this is a md file, everything after * might be italicized in some editors (Atom).

How about
[...]

PR #167, opened 2018-03-16
boilerplates: allow multiple boilerplates with similar names #159

rachx:
Fixes https://github.com/MarkBind/markbind/issues/159

To specify a boilerplate file, add the `boilerplate` in the attribute for include/dynamic panel etc
e.g.
```html
<include src="full.md" boilerplate />
<include src="book/architecture/architecturalStyles/clientServer/what/index.md" boilerplate />
```
This will use the file base name.

Alternatively, specify the path within boilerplate with the `boilerplate`  attribute
e.g.
```html
<include src="index.md" boilerplate="print-index.md" />
<include src="book/architecture/architecturalStyles/clientServer/what/index.md" boilerplate="print/index.md" />
```
Folders within boilerplate will work as well.

**However, it doesn't work for boilerplate files which are addressable pages in site.json. Will need to figure out how to make it work with front matter etc**

Testing
I tested moving full.md in clientServer/what into the `_boilerplates` folder. Then update the references in
* schedule/week8/outcome-architecturalStyle.md
* book/architecture/architecturalStyles/clientServer/what/embed.md
* book/architecture/architecturalStyles/clientServer/what/index.md
* book/common/outcomes.md (dynamic panel)
I removed a section of the code as it is not needed for live-reloading (checked as well)


lib/parser.js
```diff
@@ -109,26 +113,22 @@ Parser.prototype._preprocess = function (node, context, config) {
 
       element.attribs.header = element.attribs.name || '';
       delete element.attribs.dynamic;
-      this.dynamicIncludeSrc.push({ from: context.cwf, to: element.attribs.src });
+      this.dynamicIncludeSrc.push({ from: context.cwf, to: actualFilePath, asIfTo: element.attribs.src });
```
rachx: `to` refers to the actual file location (which might be in the boilerplate folder)

In a previous PR, we considered renaming it to `targetFilePath` and `actualFilePath` but that is inconsistent with the `staticIncludeSrc` etc and in cli's `collectIncludedFiles`, we use `dependency.to`

damithc:
I asked this before but forgot to follow up (or missed your reply): are we moving `_boilerplates` inside `_markbind` folder (e.g., `_markbind/boilerplates`) to match `_markbind/variables.md`?

rachx:
Oh yep. Mentioned in https://github.com/MarkBind/markbind-cli/pull/34
I think I will move it in this PR

acjh:
@rachx Can you include MarkBind/markbind-cli#37 and force-push to this branch?

rachx:
Done

nicholaschuayunzhi:
Remember to update the `expected site` in `test_site` @rachx

rachx:
I might add more tests for boilerplate files later

rachx:
Prof @damithc
Do you intend to use boilerplate files as addressable pages (those in site.json)?
I realized I might not have handled that case

lib/markbind/lib/parser.js
```diff
@@ -358,20 +376,8 @@ Parser.prototype.includeFile = function (file, config) {
       decodeEntities: true,
     });
 
-    let actualFilePath = file;
-    if (!utils.fileExists(file)) {
```
rachx: Probably should be added back so base/addressable pages can be properly resolved

damithc:
> Do you intend to use boilerplate files as addressable pages (those in site.json)?
> I realized I might not have handled that case

Possibly. CS2103 website has many index.md files that are exactly the same content and all of them are addressable.

rachx:
Okay, will need to think about/discuss how to handle that.
Currently, I will just take the base name for those cases (so it is still not possible to have multiple index.md)

After front matter is implemented/ if there is no "site.json", it might be harder to identify if it is a boilerplate file and handle accordingly (as compared to the case when a boilerplate file is included or dynamically included)

damithc:
> After front matter is implemented/ if there is no "site.json", it might be harder to identify if it is a boilerplate file and handle accordingly (as compared to the case when a boilerplate file is included or dynamically included)

Looks like it will be difficult to use boilerplates as addressable pages. In that case we don't need to support that. It will result in a bunch of duplicate files but they will be almost empty as the content duplication can be minimized using variables.

On a related note, I'm not sure if we want to remove site.json entirely as it contains some other info as well. May be we should consider info specified in site.json in conjunction with the info we generate automatically? @nicholaschuayunzhi

lib/markbind/lib/parser.js
```diff
@@ -49,9 +49,10 @@ function calculateNewBaseUrl(filePath, root, lookUp) {
   return calculate(filePath, []);
 }
 
-function calculateBoilerplateFilePath(filePath, config) {
-  const base = calculateNewBaseUrl(filePath, config.rootPath, config.baseUrlMap).relative;
-  return path.resolve(base, BOILERPLATE_DIRECTORY_NAME, path.basename(filePath));
+function calculateBoilerplateFilepath(filePath, pathWithinBoilerplate, config) {
```
acjh: Why the change from `*FilePath` to `*Filepath`? The latter is inconsistent with the rest of this file.

lib/markbind/lib/parser.js
```diff
@@ -91,6 +92,7 @@ Parser.prototype._preprocess = function (node, context, config) {
   element.attribs = element.attribs || {};
   element.attribs[ATTRIB_CWF] = path.resolve(context.cwf);
   if (element.name === 'include') {
+    const isBoilerplate = _.hasIn(element.attribs, 'boilerplate');
```
acjh: The check `_.hasIn(element.attribs, 'boilerplate');` appears 3 times.

There are 5 switch-cases based on `element` type:
- `if (element.name === 'include')`
- `else if (element.name === 'dynamic-panel')`
- `else if ((element.name === 'morph' || element.name === 'panel') && _.hasIn(element.attribs, 'src'))`
- `else if (element.children && element.children.length > 0)`
- default

The check is done in 3 of 5 cases, and doesn't hurt to do for the other 2.

I propose assigning `isBoilerplate` and `element.attribs[ATTRIB_BOILERPLATE]` before the first `if`.
Then we avoid `pathWithinBoilerplate || path.basename(filePath)` in `calculateBoilerplateFilePath`.

lib/markbind/lib/parser.js
```diff
@@ -359,14 +377,11 @@ Parser.prototype.includeFile = function (file, config) {
     });
 
     let actualFilePath = file;
-    if (!utils.fileExists(file)) {
-      const boilerplateFilePath = calculateBoilerplateFilePath(file, config);
+    if (!utils.fileExists(file)) { // if a boilerplate file is an addressable page in site.json
+      // search for file with the same name in the boilerplate folder
```
acjh: These comments look out-of-place.

lib/markbind/lib/parser.js
acjh: In fact, this is not the only repetition. See `isUrl`, `includeSrc`, `includeSrcPath` etc.

I've thought about the other repetitions before, but opted for duplication "as needed" previously because I didn't have time to explore the implications. However, with my current understanding, this seems like a trade-off between DRY and unnecessary/premature computation.

Alternatively, those could be abstracted into helper functions.

@Gisonrg Your thoughts?

rachx: @acjh 
Something like this?
https://github.com/MarkBind/markbind/pull/167/commits/6173f3c424c336ac7a9452004e4a4e8d354aa78a

lib/markbind/lib/parser.js
```diff
@@ -193,36 +206,18 @@ Parser.prototype._preprocess = function (node, context, config) {
...
-      }
-      filePath = path.resolve(path.dirname(context.cwf), includeSrcPath); // updated path (no hash)
+    if (!isUrl && includeSrc.hash) {
+      element.attribs.fragment = includeSrc.hash.substring(1); // save hash to fragment attribute
     }
     element.attribs.src = filePath;
-    this.dynamicIncludeSrc.push({ from: context.cwf, to: filePath });
+    this.dynamicIncludeSrc.push({ from: context.cwf, to: actualFilePath, asIfTo: filePath });
     return element;
   } else if ((element.name === 'morph' || element.name === 'panel') && _.hasIn(element.attribs, 'src')) {
```
acjh: `&& _.hasIn(element.attribs, 'src')` is no longer necessary?

lib/markbind/lib/parser.js
acjh: @rachx Nice cleanup :)

> I propose assigning `isBoilerplate` and `element.attribs[ATTRIB_BOILERPLATE]` before the first `if`.
Then we avoid `pathWithinBoilerplate || path.basename(filePath)` in `calculateBoilerplateFilePath`.

This was not addressed? Maybe even
`calculateBoilerplateFilePath(element.attribs.boilerplate || path.basename(filePath), config);` might be clearer.

lib/markbind/lib/parser.js
```diff
@@ -90,13 +90,29 @@ Parser.prototype._preprocess = function (node, context, config) {
   const self = this;
   element.attribs = element.attribs || {};
   element.attribs[ATTRIB_CWF] = path.resolve(context.cwf);
+
+  const hasSrc = _.hasIn(element.attribs, 'src');
+  if (!hasSrc) {
+    if (element.children && element.children.length > 0) {
+      element.children = element.children.map(e => self._preprocess(e, context, config));
```
acjh: Hmm, it seems unavoidable to do this below as the last `else if`.
I don't feel like "unnecessary computation" is good enough reason to duplicate this here.

In fact, if this check is done here, it makes #174 useless.

rachx: I duplicated this here so that nodes without a `src` attribute can exit the function early. Some of the subsequent computation e.g. `filePath` will throw an error if there is no `src` e.g. a normal `div`.

I was thinking of having a list of elements that require source (`dynamic-panel`, `include`, not sure about `panel`) before this. If those `src` are empty/null/undefined, return the error node.

Also after the file path and boilerplate is checked, return an error node if the file cannot be found. This should stop missing dynamically included files from crashing build/live reload.

acjh: > I duplicated this here so that nodes without a `src` attribute can exit the function early. Some of the subsequent computation e.g. `filePath` will throw an error if there is no `src` e.g. a normal `div`.

Perhaps `filePath` can be conditionally computed.

> I was thinking of having a list of elements that require source (`dynamic-panel`, `include`, not sure about `panel`) before this. If those `src` are empty/null/undefined, return the error node.
>
> Also after the file path and boilerplate is checked, return an error node if the file cannot be found. This should stop missing dynamically included files from crashing build/live reload.

Sounds alright. [`panel`](https://markbind.github.io/vue-strap/#accordion) doesn't require `src`.

lib/markbind/lib/parser.js
```diff
@@ -51,9 +51,9 @@ function calculateNewBaseUrl(filePath, root, lookUp) {
   return calculate(filePath, []);
 }
 
-function calculateBoilerplateFilePath(filePath, config) {
-  const base = calculateNewBaseUrl(filePath, config.rootPath, config.baseUrlMap).relative;
-  return path.resolve(base, BOILERPLATE_DIRECTORY_NAME, path.basename(filePath));
+function calculateBoilerplateFilePath(filePath, pathWithinBoilerplate, config) {
```
acjh: How about:

```js
function calculateBoilerplateFilePath(pathInBoilerplate, asIfAt, config) {
```

lib/markbind/lib/parser.js
acjh: Duplication still here? This is preferable:

> Perhaps `filePath` can be conditionally computed.

lib/markbind/lib/parser.js
```diff
@@ -97,18 +97,36 @@ Parser.prototype._preprocess = function (node, context, config) {
...
-      return createErrorNode(element, error);
+
+  const requiresSrc = ['include', 'dynamic-panel'].includes(element.name);
+  if (requiresSrc && _.isEmpty(element.attribs.src)) {
+    const error = new Error(`Empty src attribute in ${element.name} in: ${element.attribs[ATTRIB_CWF]}`);
+    this._onError(error);
+    return createErrorNode(element, error);
+  }
+
+  const hasSrc = _.hasIn(element.attribs, 'src');
```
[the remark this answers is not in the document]
acjh: Cancel this.

lib/markbind/lib/parser.js
```diff
@@ -293,34 +281,39 @@ Parser.prototype._parse = function (node, context, config) {
     element.attribs['no-close'] = 'true';
     element.attribs['no-switch'] = 'true';
     element.name = 'panel';
+    delete element.attribs.boilerplate;
     break;
   }
   case 'morph': {
```
Gisonrg: I think eventually we might deprecate this element. This block of logic duplicates with `panel` one.

test/test_site/index.md
```diff
@@ -12,7 +12,9 @@
 <include src="requirements/SpecifyingRequirements.md" name="Dynamic Include" dynamic />
 
 # Boilerplate include
-<include src="requirements/boilerTest.md" name="Boilerplate Referencing" />
+<include src="requirements/boilerTest.md" name="Boilerplate Referencing" boilerplate />
+
+<include src="requirements/notInside.md" name="Referencing specified path in boilerplate" boilerplate="folder/inside.md" dynamic/>
```
Gisonrg: a very noob question here:

So if I already know I want to use `folder/inside.md` here and not what's in `src`, do I have to write both `src` and `boilerplate` together? Can I just use something like:

`<include boilerplate="folder/inside.md" name="Boilerplate Referencing" />`

Remind me if this is already discussed when we design the `boilerplate` feature

acjh: [An error node is returned](https://github.com/rachx/markbind/commit/ebeb9c21e840224fb0883beeacbb0dc734591fc4#diff-ba2dd3dbaf0965b1ae992e6f64d7109fR106) if `src` is empty.

`src` is used as [`asIfTo`](https://github.com/MarkBind/markbind/pull/167/commits/5aeb0eb387c6f9fb3442ca4ab4df6d403e7d27f7#diff-ba2dd3dbaf0965b1ae992e6f64d7109fR130) and to determine [`filePath`](https://github.com/MarkBind/markbind/pull/167/commits/b57b9628db8ddd1178445c9e993472bc8c86b113#diff-ba2dd3dbaf0965b1ae992e6f64d7109fR117)/[`asIfAt`](https://github.com/MarkBind/markbind/pull/167/commits/388458453b4653d5c76e97fa77b0f09fa70a3c6b#diff-ba2dd3dbaf0965b1ae992e6f64d7109fR54) for `calculateBoilerplateFilePath`.
I think we should avoid inferring, for your example, between:
- `<include src="inside.md" boilerplate="folder/inside.md" name="Boilerplate Referencing" />`
- `<include src="folder/inside.md" boilerplate="folder/inside.md" name="Boilerplate Referencing" />`

rachx: `src` has to be written:

`src` is where it pretends to be. It is needed to set the context/current working file so that if the boilerplate file contains `<include src ="child.md">`, it will know which folder to look for `child.md` in.

Also, for dynamically include, the _include_.html files are created where they are supposed to be.

lib/markbind/lib/parser.js
rachx: Do I keep the code and only do it here (and not in preprocess)?
Also, should we poss the `logger` in the options or use `console.warn`?

acjh: Yes, keep the code because we should still make it correct.
Only doing it here is fine, we don't need repeated warnings.
Oh, `logger` is in cli. `console.warn` will suffice.

acjh:
Fix failing check?

rachx:
Fixed.

acjh:
@Gisonrg Good to go?

Gisonrg:
Nice extraction of common src handling logic :+1:

acjh:
@rachx You can merge this :)

Issue #135, opened 2018-02-04
Use MarkBind for the MarkBind website


damithc:
At some point we should use MarkBind for the MarkBind website :-)

> I don't think end users need to know there are two types of syntax. It would be good to document all syntax in one place, at least a summary.

> It would be cool if MarkBind itself can solve this problem of integrating documentation of markdown, vue, and markbind. :-)

Sub-issues:
- [x] #82: Document the `<include>` tag
- [x] #84: Document question body syntax, e.g. syntax for check boxes, radio buttons

nicholaschuayunzhi:
Would it be possible to incorporate the test site mentioned in #121 int our main website.

Components and functionality can be documented like in https://markbind.github.io/vue-strap/

acjh:
Updated issue description.

damithc:
Some relevant points:
1. It is preferable that user docs are updated only when a feature is released, not when it is merged. Otherwise users will see features that are not available to them. We can do that with Markbind i.e., we run the deploy command only when a new release is made.
2. By right MarkBind is able to live-integrate (pull content from) another website. We may be able to live-integrate content from the vuestrap website into the main site so that users only need to go to one website.

PR #169, opened 2018-03-17
Merge cli repo for better dependency management #139

### LINKED ISSUE #139: Consider merge two repos for better dependency management
Currently the `markbind` and `markbind-cli` are two separate repos. The initial thought is we could reuse the core library (renderer) as dependency for other projects, such as used in plugins for the editor (https://github.com/MarkBind/atom-markbind-helper, outdated).
However, this starts to bring extra project dependency issues. For example, for a change that needs to update both core library and CLI module, we have to update the core library, publish it, update CLI dependency, and make the corresponding change. Eventually, they should be in the same repo, and we could still separately publish them (I believe there are ways to do so). This could make our dependency easier to manage (no need to manually update version) and lead to improved workflow.
I think we should think about merging them into single repo.😕  Thoughts?

acjh:
Resolves #139

I did `git merge cli --allow-unrelated-histories` to preserve histories from both repos.
As such, there are 200+ commits. Here are the newest commits that are specific to this PR:
- df3a62b Move core library into lib/markbind
- 7402e2a Update eslintignore for lib/markdown-it/*
- 59830af Merge branch 'cli'
- 83b373c Restore eslint-plugin-lodash
- ~[9962681](https://github.com/MarkBind/markbind/commit/9962681f678b95de81ebb782c94c5983ce8bf306) Use local markbind~ (overwritten by force-push)
- c36f595 Use local markbind

---
1. To test this, `cd` to the `markbind` folder, pull this branch, and then run (`sudo`) `npm link`.
2. We can publish **markbind-cli** from the root, and publish **markbind** core from **./lib/markbind**.
3. Caveat: When running `npm version`, tagging is now committed for **cli** but not the **core** library.
   - This means that if you push that commit, the **markbind-cli** tag will be added to this repo.
   - Maybe create a version-specific branch that is only pushed to [MarkBind/markbind-cli](https://github.com/MarkBind/markbind-cli) repo.
   - Committing and tagging for the **core** library has to be done manually (smallest issue here).
   - ⚠ Care must be taken that maintainers with write access do not "Push all tags" in SourceTree:
     screenshot 2018-03-17 23 32 39


acjh:
It should be easy to merge changes from [MarkBind/markbind-cli](https://github.com/MarkBind/markbind-cli) when existing PRs there are merged.

acjh:
If this PR is merged, I can release [markbind-cli@1.6.1](https://github.com/MarkBind/markbind-cli/milestone/4) from this merged repo tomorrow (or Monday).

acjh:
> 3. Caveat: When running `npm version`, tagging is now committed for **cli** but not the **core** library.
>    - ...

Or we could sync the release versions, and then we can use the same Milestone and tags for both!

nicholaschuayunzhi:
To clarify

> To test this, cd to the markbind folder, pull this branch, and then run (sudo) npm link

If we're using local markbind in https://github.com/MarkBind/markbind/pull/169/commits/c36f595e64b7b423e8fb26490559c2af61bd7cc5 we don't have to run the `$ npm link` command right?

acjh:
`npm link` binds the `markbind` command and the global **node_modules/markbind-cli** to that folder.

You should have run it before, but that was the unmerged **markbind-cli** folder:
https://github.com/MarkBind/markbind-cli/wiki/Developer-Guide#development

> 3. To make sure you are using the cloned CLI program in your own terminal/console, in the cloned CLI repo, run
>
>    ```
>    $ npm link
>    ```

Gisonrg:
Thanks @acjh for the efforts!!! The merge is not easy 👏 

`npm version` has a [command flag](https://docs.npmjs.com/cli/version#git-tag-version) which disallow auto commit tag, we could enforce that by automation (could use npm scripts) and well written release guide.

Also we need to migrate the Wiki page, which works similar to change a git remote address (Wikis are just a repo).

package.json
```diff
@@ -1,43 +1,55 @@
...
+    "live-server": "^1.2.0",
+    "markbind": "file:lib/markbind",
     "nunjucks": "^3.0.0",
-    "path-is-inside": "^1.0.2"
+    "path-is-inside": "^1.0.2",
+    "walk-sync": "^0.3.1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MarkBind/markbind-cli.git"
```
Gisonrg: Should change this to `MarkBind/markbind`

acjh:
Published **markbind@1.6.0** and tested like this:

```shell
$ cd ~/Desktop
$ mkdir tmp
$ cd tmp
$ npm init
$ # Press enter 10 times.
$ npm install markbind@1.6.0
$ echo "const MarkBind = require('markbind');" > index.js
$ echo "console.log(new MarkBind());" >> index.js
$ node index.js
```

Output:

```js
Parser {
  _options: {},
  _onError: [Function: bound consoleCall],
  _fileCache: {},
  dynamicIncludeSrc: [],
  staticIncludeSrc: [],
  boilerplateIncludeSrc: [],
  missingIncludeSrc: [] }
```

acjh:
Merge worked fine.

acjh:
> Also we need to migrate the Wiki page, which works similar to change a git remote address (Wikis are just a repo).

I was thinking of going straight to #135. Easier to track future changes and PRs.

> Also might need to mark the `markbind-cli` repo as deprecated (remove the code and update the README).

I updated the [README](https://github.com/MarkBind/markbind-cli#readme). I'll remove the code when the remaining PRs are closed.

acjh:
@Gisonrg Can you activate Travis on this repo? 🙈

Issue #122, opened 2018-01-20
Defaults for addressable pages


acjh:
Currently, it's necessary to specify every addressable page in **site.json**:
https://github.com/nus-cs2103/website-base/blob/master/site.json

- Defaults, e.g. for index.md
- Possibly specify folders?
- **site.json** is also where an author specifies `title` of the page, so that has to be handled too

PR #175, opened 2018-03-22
Add front matter feature

danielbrzn:
This PR aims to add front matter support to Markbind. This feature will allow content authors to specify the parameters of a page (e.g. search keywords, page title) using a front matter tag (<frontmatter>).

Key Points: 

- A migration script has been made available at [markbind-migrate](https://github.com/danielbrzn/markbind-migrate) to transfer existing page metadata from `site.json` to the respective page

- A `siteData.json` file will be created upon each call of `serve` or `build` which will contain the metadata of each addressable page.

- A pages array is used in `site.json` to specify pages that are addressable. It now supports the usage of **globs** to allow the author more flexibility in defining addressable pages. This works with existing Markbind sites. An example is as follows.

```
"pages": [
    {
    "glob" : "**/index.md"
    },
    {
    "glob" : "**/projectList.md"
    },
    {
    "glob" : "**/print.md"
    },
    {
    "glob" : "**/definitions.md"
    },
  ]
```

- A `titlePrefix` parameter in `site.json` is allowed for the user to specify prefixes that should be applied to the titles of all addressable pages. (e.g.: CS2103/T Website)


lib/Site.js
```diff
@@ -356,22 +357,55 @@ Site.prototype.generatePages = function () {
...
+    });
+  },
+  );
+};
+
+/**
+ * Builds the siteData file that specifies addressable pages and their associated keywords
+ */
+Site.prototype.buildSiteData = function () {
+  const filePaths = walkSync(this.rootPath, { directories: false })
```
rachx: Instead of filtering, I think you can use something like `{ directories: false, ignore: [ '_site' ] }` and the `globs` property to only match `.md` files

lib/Site.js
```diff
@@ -356,22 +357,55 @@ Site.prototype.generatePages = function () {
...
+    .filter(currentPath => currentPath.endsWith('.md') && !currentPath.startsWith('_site'));
+  const frontMatterData = [];
+  const promises = [];
+  filePaths.forEach((currentFile) => {
+    promises.push(
+      new Promise((resolve, reject) => {
```
rachx: If we are using `fs-extra-promise`, we can use `fs.readFileAync(path).then...` ?

damithc:
hmm... in CS2103 website (se-book) many of the addressable pages are actually boilerplates.  Eventually we may have to support something like this as well:
```
---
<include src="frontmatter.md" />
---
```
Related to the above, I wonder if we should use `<frontmatter>` tags instead of `---` to be consistent with other markbind syntax.

damithc:
To clarify further, we often present the same content in slightly different ways using different addressable pages that include the same primary content (e.g., full.md, print.md, index.md embed.md etc.), which means some front matter content such as title, keywords etc may need to be reused as well.

lib/Site.js
danielbrzn: I tried using globs to ignore but I didn't get too far with it. I'll give your suggestion a shot.

lib/Site.js
danielbrzn: Yeah that's true. I'll switch the methods.

lib/Site.js
rachx: Actually, maybe it is not needed. `_site` folder should have been emptied

danielbrzn:
Hm, to clarify further. Let's take the IntelliJ section as an example.

```
    {
      "src": "book/intellij/index.md",
      "title": "CS2103/T Textbook - Intellij IDEA"
    },
    {
      "src": "book/intellij/print.md",
      "title": "CS2103/T Textbook - Intellij IDEA [Printable]"
    },
```

Let's say we defined front matter (fm) in the index.md file, we should be able to reuse fm from that file in the print.md file?

`---` was chosen to denote fm as it's how YAML defines it, and many parsers are also written to parse fm in this manner. Currently we're using `fastmatter` for this purpose. Perhaps we could do some sort processing on the backend to add a segment id to front matter? Then we can simply just reference it in a manner as follows:

```
<include src="requirements/EstablishingRequirements.md#frontmatter" />
```

I'm also open to using `<frontmatter>` tags as well, we should be able to parse it.

lib/Site.js
danielbrzn: Yes you're right, I just gave it another try. The folder is emptied at this point so the filtering for `_site` files is extraneous. I'll remove it.

damithc:
> Let's say we defined front matter (fm) in the index.md file, we should be able to reuse fm from that file in the print.md file?

Although I haven't done so yet, the goal was to move all the print.md, full.md, etc to boilerplates because they all contain the same content. Once frontmatter is introduced, we cannot do that any more because now each of them will contain different content as frontmatter for each topic is different, unless  this kind of syntax is supported.
```
---
<include src="frontmatter.md" />
---
```
That way, we can write the frontmatter in a separate file (one for each topic) but use boilerplates for print.md, full.md etc.

danielbrzn:
I understand what you mean. I'll see how best to implement this 👍

damithc:
> Would you be using the following code inside a boilerplate file?

Yes, that's what I was thinking. That way, the boilerplate file can be used in many places and each time it will pick the frontmatter in the current folder. 

BTW, authors might want to use variables inside the front matter too e.g., if the page title were to contain the module code
```
---
title: {{module}} Text Book: Requirements
---
```

That's why I thought it may be easier and more future proof if we used the same syntax for frontmatter as we use elsewhere.

lib/Page.js
```diff
@@ -71,6 +80,38 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+        frontMatterData = $(includedFrontMatter).get(FIRST_ELEMENT_INDEX).children[FIRST_ELEMENT_INDEX].data;
+      } else {
+        frontMatterData = frontMatter.get(FIRST_ELEMENT_INDEX).children[FIRST_ELEMENT_INDEX].data;
+      }
+      const formattedMatter = `${FRONT_MATTER_FENCE + NEW_LINE}${frontMatterData}${FRONT_MATTER_FENCE}`;
+      // Parse front matter data
+      const parsedData = fm(formattedMatter);
+      parsedData.attributes.src = this.pageSrc;
+      parsedData.attributes.title = this.pageTitlePrefix + parsedData.attributes.title;
+      this.frontMatter = parsedData.attributes;
```
nicholaschuayunzhi: Do we need to add `src` and `title` to `parsedData.attributes` here? I think we can use a new object instead.

lib/Page.js
```diff
@@ -71,6 +80,38 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+      this.title = this.frontMatter.title;
+      resolve(includedPage);
+    } else {
+      // Page is addressable but does not have a specified title
+      this.frontMatter = { src: this.pageSrc };
+      resolve(includedPage);
```
nicholaschuayunzhi: We can combine line 110 and 106 into a line after the if else bllock

lib/Page.js
```diff
@@ -118,8 +160,12 @@ Page.prototype.generate = function (builtFiles) {
         this.collectIncludedFiles(markbinder.getBoilerplateIncludeSrc());
         this.collectIncludedFiles(markbinder.getMissingIncludeSrc());
       })
-      .then(resolve)
-      .catch(reject);
+      .then(() => {
+        resolve(this.frontMatter);
+      })
+      .catch((err) => {
+        reject(err);
```
nicholaschuayunzhi: we can still keep `.catch(reject)`

lib/Site.js
```diff
@@ -44,7 +42,7 @@ const SITE_CONFIG_DEFAULT = {
   },
 };
 
-const INDEX_MARKDOWN_DEFAULT = '# Hello world\nWelcome to your page generated with MarkBind.\n';
+const INDEX_MARKDOWN_DEFAULT = '<frontmatter>\ntitle: "Hello World"\n</frontmatter>\n\n# Hello world\nWelcome to your page generated with MarkBind.\n';
```
nicholaschuayunzhi: Error from linting:
```
  45:1   error  Line 45 exceeds the maximum line length of 110  max-len
```

lib/Site.js
```diff
@@ -25,14 +25,12 @@ const SITE_ASSET_FOLDER_NAME = 'asset';
...
 const SITE_CONFIG_DEFAULT = {
   baseUrl: '',
-  pages: [
-    {
-      src: 'index.md',
-      title: 'Hello World',
-    },
+  addressable: [
+    'index.md',
   ],
```
nicholaschuayunzhi: error from linting:
```
  32:16  error  There should be no linebreak after '['          array-bracket-newline
  34:3   error  There should be no linebreak before ']'         array-bracket-newline
```

lib/Site.js
```diff
@@ -356,24 +363,39 @@ Site.prototype.generatePages = function () {
...
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
-      .then(resolve)
-      .catch(reject);
+      .then((frontMatterData) => {
+        this.frontMatterData = frontMatterData;
+        resolve();
+      })
+      .catch((err) => {
+        reject(err);
```
nicholaschuayunzhi: can do a 
```js
 .then(resolve)
 .catch(reject);
```
instead of resolving in line 381

If i'm not wrong, line 379 can also be changed to 
```js
.then(frontMatterData => this.frontMatterData = frontMatterData)
```

lib/Site.js
```diff
@@ -356,24 +363,39 @@ Site.prototype.generatePages = function () {
...
+      });
   });
 };
 
+/**
+ * Builds the index of files to be traversed as addressable pages
+ */
+Site.prototype.buildIndex = function () {
+  const addressableFiles = this.siteConfig.addressable;
+  const filePathIndex = walkSync(this.rootPath, { directories: false, ignore: ['_boilerplates'] })
```
nicholaschuayunzhi: use `BOILERPLATE_FOLDER_NAME` here

rachx:
Can you add/share the script to convert existing site to front matter?

lib/Site.js
```diff
@@ -356,24 +363,39 @@ Site.prototype.generatePages = function () {
...
+    .filter(currentPath => addressableFiles.some(fileName => currentPath.includes(fileName)));
```
rachx: Should we check for the exact match in the file base name? (`index.md` should match but not `other-index.md`)
Also are we going to support/use glob pattern in the future e.g. (`**/index.md`)

lib/Page.js
danielbrzn: Hm, we could. So a new object that has the fields from `parsedData.attributes` and `title` and `src`? It does seem like more work though, since we're essentially cloning `parsedData.attributes`

lib/Page.js
danielbrzn: Will fix 👍

lib/Site.js
danielbrzn: Good catch.

It's possible that we implement that in the future, perhaps using walkSync's glob support.

lib/Site.js
danielbrzn: The change can't be made to line 379 as it'll be an error according to ESLint. [no-return-assign](https://eslint.org/docs/rules/no-return-assign)

lib/Page.js
```diff
@@ -71,6 +80,37 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+ * @param includedPage a page with its dependencies included
+ */
+Page.prototype.collectFrontMatter = function (includedPage) {
+  return new Promise((resolve) => {
+    const $ = cheerio.load(includedPage);
+    const frontMatter = $('frontmatter');
+    const includedFrontMatter = $('frontmatter div');
+    if (frontMatter.get(FIRST_ELEMENT_INDEX) !== undefined) {
+      let frontMatterData = '';
+      if (includedFrontMatter.get(FIRST_ELEMENT_INDEX) !== undefined) {
```
rachx: I think you can use `.first()` instead of `get(FIRST_ELEMENT_INDEX)`.
Also, `children().first()`

Also, is there only one/do we only care about one `includedFrontMatter`?

danielbrzn: For both cases of frontmatter, only the attributes from the first occurrence of `<frontmatter>` tag is taken to represent the data of that particular page. 

`frontmatter` is for front matter that is defined in the page itself, whereas `includedFrontMatter` is for front matter that has been included from another file. It is possible that there may be multiple occurrences of the `<frontmatter>` tag in a single page.

test/test_site/expected/index.html
```diff
@@ -11,7 +11,10 @@
 </head>
 <body>
 <div id="app" class="container-fluid">
-    <div>
+    <frontmatter>
+  title: Hello World
+</frontmatter>
```
acjh: Can we not render `frontmatter` elements?

acjh:
This is not compatible with the existing Search feature right? 😕

test/test_site/expected/index.html
acjh: Yes, since it's unused.

danielbrzn:
Made a new PR to update the Typeahead component.

https://github.com/MarkBind/vue-strap/pull/17

lib/Site.js
```diff
@@ -260,13 +263,16 @@ Site.prototype.generate = function (baseUrl) {
 };
 
 /**
- * Build all pages of the site
+ * Build all pages and the index of the site
```
acjh: But I see this in `Site.prototype.generate`:

```diff
+ .then(() => this.buildIndex())
  .then(() => this.buildSourceFiles())
```

How about:

```js
.then(() => this.collectAddressablePaths()) // The current buildIndex
.then(() => this.buildSourceFiles())
.then(() => this.writeSiteData())
```

Let's stick to one term, rather than mixing with "Site index".
"siteData" is consistent with "pageData" in `Site.prototype.createPageData` 👍

lib/Site.js
```diff
@@ -357,24 +363,40 @@ Site.prototype.generatePages = function () {
...
   this.pageModels.forEach((page) => {
     processingFiles.push(page.generate(builtFiles));
   });
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
-      .then(resolve)
+      .then((frontMatterData) => {
+        this.frontMatterData = frontMatterData;
+        resolve();
+      })
```
acjh:
```diff
+ .then((frontMatterData) => { this.frontMatterData = frontMatterData; })
  .then(resolve)
```

lib/Site.js
```diff
@@ -357,24 +363,40 @@ Site.prototype.generatePages = function () {
...
       .catch(reject);
   });
 };
 
+/**
+ * Builds the index of files to be traversed as addressable pages
+ */
+Site.prototype.buildIndex = function () {
```
acjh: Rename this to `Site.prototype.collectAddressablePaths`.
Rename `this.filePaths` to `this.addressablePaths`.

lib/Page.js
```diff
@@ -118,7 +165,9 @@ Page.prototype.generate = function (builtFiles) {
         this.collectIncludedFiles(markbinder.getBoilerplateIncludeSrc());
         this.collectIncludedFiles(markbinder.getMissingIncludeSrc());
       })
-      .then(resolve)
+      .then(() => {
+        resolve(this.frontMatter);
+      })
```
acjh: This change is unnecessary. We have reference to the `Page`s in `this.pageModels` in **Site.js**.

lib/Site.js
```diff
@@ -393,7 +415,15 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
 
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
-      .then(() => logger.info('Pages rebuilt'))
+      .then((updatedFrontMatter) => {
+        logger.info('Pages rebuilt');
+        updatedFrontMatter.forEach((currentData) => {
```
acjh: As mentioned in `Page.prototype.generate`, `page.generate` shouldn't return `updatedFrontMatter`.
You should be able to get it from each `page` in `this.pageModels`.

```js
this.pageModels.forEach((page) => {
  const currentData = page.frontMatter;
  // ...
});
```

Then revert:

```diff
- .then(() => logger.info('Pages rebuilt'))
```

lib/Site.js
```diff
@@ -393,7 +415,15 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
...
+          const toUpdate = this.frontMatterData.findIndex(frontMatter => frontMatter.src === currentData.src);
+          this.frontMatterData[toUpdate] = currentData;
+        });
+      }).then(() => fs.writeJsonAsync(path.join(this.rootPath, 'siteData.json'),
```
acjh:
```js
})
.then(() => this.writeSiteData())
```

lib/markbind/lib/parser.js
```diff
@@ -438,7 +439,7 @@ Parser.prototype.renderFile = function (file, config) {
       }
       const fileExt = utils.getExtName(file);
       if (fileExt === 'md') {
-        const inputData = md.render(data.toString());
+        const inputData = md.render(fm(data.toString()).body);
```
acjh: What does this do?

lib/Page.js
```diff
@@ -8,10 +8,21 @@ const FsUtil = require('./util/fsUtil');
 const pathIsInside = require('path-is-inside');
 
 const MarkBind = require('./markbind/lib/parser');
+const cheerio = require('cheerio');
+const fm = require('fastmatter');
+
+const FIRST_ELEMENT_INDEX = '0';
+const FRONT_MATTER_FENCE = '---';
+const NEW_LINE = '\n';
```
acjh: Let's not have `FIRST_ELEMENT_INDEX` and `NEW_LINE` as constants.

asset/js/setup.js
```diff
@@ -10,30 +10,26 @@ function setup() {
...
+      frontmattersearch,
     },
     data() {
       return {
-        searchData: titleArray,
+        searchData: siteData.pages,
       };
     },
     methods: {
       searchCallback(match) {
```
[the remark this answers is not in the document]
danielbrzn: `searchData` is used [here](https://github.com/MarkBind/vue-strap/blob/411dcba73bebb57d2bb1b2ead7cfd6708631f72f/src/Typeahead.vue#L103).

'searchCallback` is used [here](https://github.com/MarkBind/vue-strap/blob/411dcba73bebb57d2bb1b2ead7cfd6708631f72f/src/Typeahead.vue#L149).

The `Typeahead` component as it is now won't support our front matter search because it is only doing matches on a primitive array. The PR https://github.com/MarkBind/vue-strap/pull/17 adds a method that allow matching on an array of objects.

asset/js/setup.js
```diff
@@ -10,30 +10,26 @@ function setup() {
...
-        window.location.pathname = route.replace('.md', '.html');
+        window.location.pathname = match.src.replace('.md', '.html');
       },
     },
   });
   VueStrap.installEvents(vm);
 }
 
-jQuery.getJSON('../../site.json')
+jQuery.getJSON(`${window.location.origin}/siteData.json`)
```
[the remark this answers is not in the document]
danielbrzn: Hmm yeah. Should I do a check for the existence of the siteData.json file? 

I was thinking: Check `${window.location.origin}/siteData.json` if not, then `${window.location.origin}/website/siteData.json`

lib/Page.js
```diff
@@ -71,6 +83,40 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
   });
 };
 
+/**
+ * Records the front matter data of the current page into this.frontMatter
+ * @param includedPage a page with its dependencies included
+ */
+Page.prototype.collectFrontMatter = function (includedPage) {
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
danielbrzn: Hm, I guess it's not necessary. Nothing that is async happening inside the block.

asset/js/setup.js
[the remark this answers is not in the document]
acjh: The 2 lines that you linked are:
- `return this.data.filter(value => {`
- `this.onHit(this.items[this.current], this)`

Am I missing something?

> The `Typeahead` component as it is now won't support our front matter search because it is only doing matches on a primitive array. The PR MarkBind/vue-strap#17 extends the `Typeahead` component by adding a method that allow matching on an array of objects.

The "array of objects" is from what we write to **siteData.json** right? Can't we preprocess the `title` and `keywords`? It's more maintainable than a custom "subclass"-ed component.

asset/js/setup.js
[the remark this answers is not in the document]
acjh: We cannot hardcode `website` because it depends on `baseUrl` in **site.json**.

lib/Site.js
```diff
@@ -25,26 +25,23 @@ const SITE_ASSET_FOLDER_NAME = 'asset';
...
 
 const SITE_CONFIG_DEFAULT = {
   baseUrl: '',
-  pages: [
-    {
-      src: 'index.md',
-      title: 'Hello World',
-    },
-  ],
+  addressable: ['**/index.md'],
```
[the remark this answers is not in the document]
danielbrzn: 1. Yes. Instead, the author can specify pages that are meant to be addressable, without having to list the full path of every page that's meant to be addressable.

2. Yes it does. Why is that bad?

lib/Site.js
```diff
@@ -167,7 +167,9 @@ Site.prototype.createPageData = function (config) {
   const resultPath = path.join(this.outputPath, setExtension(config.pageSrc, '.html'));
   return new Page({
     content: '',
-    title: config.title || '',
+    title: '',
```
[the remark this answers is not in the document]
danielbrzn: The title may or may not be prefixed with a defined `titlePrefix` later on in `Page.js`

asset/js/setup.js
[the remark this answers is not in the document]
danielbrzn: Okay I think it's clearer here. It's used in the HTML to display the searchbar: https://github.com/MarkBind/vue-strap/pull/17

That's correct. Hmm.. How could we preprocess that in a way that is understandable by `Typeahead.vue`? Currently it does matches from simple arrays like follows:
`['Oregon', 'Arizona', 'New York']`

An example of how `Typeahead.vue` can be used now is shown [here](http://yuche.github.io/vue-strap/#typeahead)

What we need is a way for it to match against titles and keywords from the pages that we have.

asset/js/setup.js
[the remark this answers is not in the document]
danielbrzn: Ah alright.. Hm.. how about just getting the `window.location` then? Downside is that it will only work on the front page of the website.

lib/Site.js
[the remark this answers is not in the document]
acjh: 1. You didn't clearly state this breaking change. 😕 (cc: @Gisonrg)

2. Not sure what complications there might be.

asset/js/setup.js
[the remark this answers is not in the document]
acjh: > Okay I think it's clearer here. It's used in the HTML to display the searchbar: MarkBind/vue-strap#17

Ah, I was searching in the code. Thanks for pointing that out, much clearer now.

> How could we preprocess that in a way that is understandable by `Typeahead.vue`?

We could concatenate the `title` and `keywords`.
Can you post a screenshot of how the proposed `FrontMatterSearch` component looks?

asset/js/setup.js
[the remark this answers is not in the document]
acjh: What's the improvement over the current relative path (`../..`)?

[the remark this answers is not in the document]
danielbrzn: I guess it's clearer. But we can switch back to the relative path if it's preferable?

lib/markbind/lib/parser.js
danielbrzn: This was meant to only render the body of a html file (minus the front matter), but it is not necessary. Let me remove it.

asset/js/setup.js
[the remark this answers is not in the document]
acjh: Hmm, why was it `../..` before? I think we only tested it locally.

Is it run from the current page (which meant it didn't need `../..`)?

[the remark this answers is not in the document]
danielbrzn: Nope! It's run from the `/markbind/js` folder which means we needed `../..` to access the JSON file.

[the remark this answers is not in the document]
acjh: Then it's preferable.

asset/js/setup.js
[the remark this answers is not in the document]
acjh: Ah, got objects to work!

```html
<typeahead
  :data="searchData"
  :template="titleTemplate"
  template-name="title"
  placeholder="Search"
  match-case
></typeahead>
```

```js
data () {
  return {
    searchData: [
      {
        title: 'Title 1',
        keywords: ['Keyword', 'drowyeK'],
        value: function () { return [this.title].concat(this.keywords).join(' '); },
        indexOf: function (query) { return this.value().indexOf(query) },
        toLowerCase: function () { return this.value().toLowerCase(); }
      },
      {
        title: 'Title 2',
        keywords: ['Keyword', 'drowyeK'],
        value: function () { return [this.title].concat(this.keywords).join(' '); },
        indexOf: function (query) { return this.value().indexOf(query) },
        toLowerCase: function () { return this.value().toLowerCase(); }
      }
    ],
    'titleTemplate': '{{ item.title }}<br><sub>{{ item.keywords }}</sub>',
  }
},
```

screenshot 2018-04-15 01 12 04

This works by reverse-engineering **Typeahead.vue**:

```js
primitiveData () {
  if (this.data) {
    return this.data.filter(value => {
      value = this.matchCase ? value : value.toLowerCase()
      var query = this.matchCase ? this.value : this.value.toLowerCase()
      return this.matchStart ? value.indexOf(query) === 0 : value.indexOf(query) !== -1
    }).slice(0, this.limit)
  }
}
```

lib/Site.js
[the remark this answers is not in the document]
acjh: So this will break #178, #188 and #190...

Why wasn't this designed to be non-breaking? 😢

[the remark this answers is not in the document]
danielbrzn: #188 and #190 can be migrated to the new format using https://github.com/danielbrzn/markbind-migrate.

As for #178, I have taken a look at it and found out the areas that have to be changed should this PR be merged.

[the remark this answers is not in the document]
acjh: That's irreversible, right?

I'm not convinced if that's an improvement. It's unnecessarily restrictive, especially since we're reverse-generating **siteData.json** in the existing `pages` format.

I'm inclined to this:

> 3. I think this is a better design: https://github.com/angular/angular-cli/wiki/stories-asset-configuration
We can change this later. To me, `pages` is still more intuitive and consistent with MarkBind's design.

```js
pages: [
  {
    src: 'index.md',
    title: 'Hello World'
  },
  {
    glob: '**/index.md'
  }
]
```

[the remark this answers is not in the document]
acjh: > The main idea of having front matter was to ease author burden because it's a pain to have to specify every addressable page, together with its title inside `site.json`.

I believe that's solved by glob rather than Front Matter, right? Front Matter simply lets you put the `title` in the file itself, instead of in one place. That's not necessarily better, definitely not worth a breaking change in usability.  (*PR description still doesn't mention breaking change.*)

> As to the glob suggestion, would we be using the existing site.json, and then specifying a glob to collect pages that are not specified by the author?

Yes, as proposed above and expanded on here with `titlePrefix`:

```js
pages: [
  {
    src: 'index.md',
    title: 'Hello World'
  },
  {
    glob: '**/index.md',
    titlePrefix: 'CS2103/T Website'
  },
  {
    glob: '**/projectList.md',
    titlePrefix: 'CS2103/T Website - Project'
  }
]
```

As it is, I'm opposed to merging this.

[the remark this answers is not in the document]
danielbrzn: >I believe that's solved by glob rather than Front Matter, right? Front Matter simply lets you put the title in the file itself, instead of in one place. That's not necessarily better, definitely not worth a breaking change in usability. (PR description still doesn't mention breaking change.)

Yes.

To confirm, we'll generate `siteData.json` from `site.json` which is able to specify globs and src. Are titles still to be specified within `site.json`?

Also, `titlePrefix` was meant to be a global attribute, should we be supporting it on a per-glob basis?

asset/js/setup.js
```diff
@@ -10,30 +10,32 @@ function setup() {
...
 function setupWithSearch(siteData) {
-  const routeArray = jQuery.map(siteData.pages, object => object.src);
-  const titleArray = jQuery.map(siteData.pages, object => object.title);
-  const { typeahead } = VueStrap.components;
+  const { frontmattersearch } = VueStrap.components;
   const vm = new Vue({
     el: '#app',
     components: {
-      typeahead,
+      frontmattersearch,
```
acjh: Revert to `typeahead`?

lib/Page.js
```diff
@@ -6,12 +6,20 @@ const Promise = require('bluebird');
 const logger = require('./util/logger');
 const FsUtil = require('./util/fsUtil');
 const pathIsInside = require('path-is-inside');
-
+const cheerio = require('cheerio');
+const fm = require('fastmatter');
```
acjh: Restore newline after this line.

lib/Page.js
```diff
@@ -71,6 +80,39 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
   });
 };
 
+/**
+ * Records the front matter data of the current page into this.frontMatter
+ * Removes the front matter data of the current page before it is rendered
```
acjh: "front matter ~`data`~"

lib/Page.js
```diff
@@ -84,7 +126,7 @@ Page.prototype.generate = function (builtFiles) {
       userDefinedVariablesMap: this.userDefinedVariablesMap,
       rootPath: this.rootPath,
     })
-      .then(result => markbinder.resolveBaseUrl(result, {
+      .then(result => markbinder.resolveBaseUrl(this.collectFrontMatter(result), {
```
acjh:
```diff
+ .then(result => this.collectFrontMatter(result))
  .then(result => markbinder.resolveBaseUrl(result, {
```

lib/Site.js
```diff
@@ -25,26 +25,23 @@ const SITE_ASSET_FOLDER_NAME = 'asset';
 const TEMPLATE_ROOT_FOLDER_NAME = 'template';
 const TEMPLATE_SITE_ASSET_FOLDER_NAME = 'markbind';
 const USER_VARIABLES_PATH = '_markbind/variables.md';
+const TITLE_PREFIX_SEPARATOR = ' - ';
```
acjh: Add a newline before this, since this has nothing to do with paths.

lib/Site.js
```diff
@@ -25,26 +25,23 @@ const SITE_ASSET_FOLDER_NAME = 'asset';
...
     '*.md',
   ],
   deploy: {
     message: 'Site Update.',
   },
 };
 
-const INDEX_MARKDOWN_DEFAULT = '# Hello world\nWelcome to your page generated with MarkBind.\n';
+const INDEX_MARKDOWN_DEFAULT = '<frontmatter>\ntitle: "Hello World"\n</frontmatter>\n\n# Hello world\n'
+  + 'Welcome to your page generated with MarkBind.\n';
```
acjh:
```js
const INDEX_MARKDOWN_DEFAULT = '<frontmatter>\n'
  + 'title: "Hello World"\n'
  + '</frontmatter>\n\n'
  + '# Hello world\n'
  + 'Welcome to your page generated with MarkBind.\n';
```

lib/Site.js
```diff
@@ -64,6 +61,10 @@ function Site(rootPath, outputPath) {
   this.pageTemplatePath = path.join(__dirname, TEMPLATE_ROOT_FOLDER_NAME, PAGE_TEMPLATE_NAME);
   this.pageTemplate = ejs.compile(fs.readFileSync(this.pageTemplatePath, 'utf8'));
   this.pageModels = [];
+
+  // Addressable paths and front matter data array
+  this.addressablePaths = [];
+  this.frontMatterData = [];
```
acjh: Is maintaining this state necessary? Why not assign locally in `writeSiteData`?

lib/Site.js
```diff
@@ -357,24 +378,38 @@ Site.prototype.generatePages = function () {
...
+      .then(() => { this.pageModels.forEach(page => this.frontMatterData.push(page.frontMatter)); })
       .then(resolve)
       .catch(reject);
   });
 };
 
+/**
+ * Builds the index of paths to be traversed as addressable pages
+ */
+Site.prototype.collectAddressablePaths = function () {
```
acjh: Move to before `Site.prototype.collectBaseUrl` (group similar methods).

lib/Site.js
```diff
@@ -275,6 +280,22 @@ Site.prototype.buildSourceFiles = function () {
   });
 };
 
+/**
+ * Build data index of the site
+ */
+Site.prototype.writeSiteData = function () {
```
acjh: Move to after `Site.prototype.copyMarkBindAsset`.

test/test_site/site.json
```diff
@@ -1,14 +1,7 @@
...
-    {
-      "src": "index.md",
-      "title": "Hello World"
-    },
-    {
-      "src": "bugs/index.md",
-      "title": "Open Bugs"
-    }
+  "addressable": [
+    "**/index.md"
```
acjh: Show example of `titlePrefix` and `keywords` if you're adding support in this PR.

lib/Site.js
danielbrzn: Yes. I'm accessing it from `Site.prototype.regenerateAffectedPages` and `Site.prototype.generatePages`.

acjh: Why not assign locally in `writeSiteData`?
It's used immediately and ever only used once each time it is generated.

danielbrzn:
Made the necessary changes. It'll now support globs specified by the user in `site.json`, together with page entries that already exist @acjh

Gisonrg:
I've merged the FrontMatterSearch vue component, are you planning to use it in this PR? 😛

lib/Page.js
```diff
@@ -71,6 +81,44 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+      frontMatterData = $(includedFrontMatter)[0].children[0].data;
+    } else {
+      frontMatterData = frontMatter[0].children[0].data;
+    }
+    const formattedMatter = `${FRONT_MATTER_FENCE}\n${frontMatterData}${FRONT_MATTER_FENCE}`;
+    // Parse front matter data
+    const parsedData = fm(formattedMatter);
+    parsedData.attributes.src = this.pageSrc;
+    parsedData.attributes.title = this.pageTitlePrefix + parsedData.attributes.title;
+    this.frontMatter = parsedData.attributes;
```
acjh:
```js
if (frontMatter.length) {
  const frontMatterData = frontMatter.find('div').length
    ? frontMatter.find('div')[0].children[0].data
    : frontMatter.children[0].data;
  const frontMatterWrapped = `${FRONT_MATTER_FENCE}\n${frontMatterData}${FRONT_MATTER_FENCE}`;
  const parsedData = fastmatter(frontMatterWrapped);
  this.frontMatter = parsedData.attributes;
  this.frontMatter.src = this.pageSrc;
  this.frontMatter.title = this.pageTitlePrefix + this.frontMatter.title;
```

lib/Site.js
acjh: Move to after `Site.prototype.copyMarkBindAsset`.

lib/Site.js
```diff
@@ -407,6 +444,22 @@ Site.prototype.copyMarkBindAsset = function () {
...
+  return new Promise((resolve, reject) => {
+    fs.writeJsonAsync(path.join(this.rootPath, 'siteData.json'),
+                      { pages: this.pageModels.map(page => (page.frontMatter)) })
```
acjh:
```js
const siteDataPath = path.join(this.rootPath, SITE_DATA_NAME);
const siteData = { pages: this.pageModels.map(page => (page.frontMatter)) };
fs.writeJsonAsync(siteDataPath, siteData)
```

lib/Site.js
```diff
@@ -407,6 +444,22 @@ Site.prototype.copyMarkBindAsset = function () {
...
+      .then(() => logger.info('Site data built'))
+      .then(resolve)
+      .catch((error) => {
+        // if error, remove the site and temp folders
```
acjh: No need to copy this comment.

lib/Site.js
```diff
@@ -250,7 +278,9 @@ Site.prototype.generate = function (baseUrl) {
       .then(() => this.collectBaseUrl())
       .then(() => this.collectUserDefinedVariablesMap())
       .then(() => this.buildAssets())
+      .then(() => this.collectAddressablePaths())
```
acjh: Rename this to `collectAddressablePages` and move to:

```diff
  this.readSiteConfig(baseUrl)
+   .then(() => this.collectAddressablePages())
    .then(() => this.collectBaseUrl())
```

Let's preserve the "shape" of pages so we don't do this in `Site.prototype.generatePages`:

```js
const index = this.siteConfig.pages.findIndex(currentPage => currentPage.src === src);
```

lib/Page.js
```diff
@@ -71,6 +81,44 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+    // Title specified in site.json will override title specified in front matter
+    if (this.title) {
+      this.title = this.pageTitlePrefix + this.title;
+    } else {
+      this.title = this.frontMatter.title;
+    }
```
acjh: This doesn't seem to "override title specified in front matter" (it does "take priority").
Isn't `this.frontMatter` used by `Site` though?

lib/Site.js
```diff
@@ -26,25 +26,30 @@ const TEMPLATE_ROOT_FOLDER_NAME = 'template';
...
   ignore: [
     '_site/*',
-    '*.json',
+    'site.json',
     '*.md',
   ],
+  pages: [
+    {
+      glob: '**/index.md',
+    },
```
acjh: Can you add this above (rather than replace here) for a clean diff and to show that both work together?

```diff
  pages: [
    {
      src: 'index.md',
      title: 'Hello World',
    },
+   {
+     glob: '**/index.md',
+   },
  ],
```

test/test_site/site.json
```diff
@@ -19,7 +9,13 @@
     "testUtil/*",
     "expected/*"
   ],
+  "pages": [
+    {
+    "glob" : "**/index.md"
```
acjh: Move this back up:

```diff
  "pages": [
    {
      "src": "index.md",
      "title": "Hello World"
    },
    {
-     "src": "bugs/index.md",
-     "title": "Open Bugs"
+     "glob": "**/index.md",
    },
  ],
```

test/test_site/site.json
```diff
@@ -19,7 +9,13 @@
...
+    }
+  ],
   "deploy": {
     "message": "Site Update."
-  }
+  },
+  "titlePrefix": "Markbind Test Site"
```
acjh: MarkBind with a capital B :)

lib/Site.js
danielbrzn: This has been changed

lib/Site.js
acjh: @danielbrzn Did you miss this?

acjh:
@nicholaschuayunzhi Are you satisfied with the requested changes?

lib/Site.js
danielbrzn: Yup let me fix it

lib/Site.js
```diff
@@ -357,14 +398,17 @@ Site.prototype.generatePages = function () {
...
   const { baseUrl } = this.siteConfig;
   const builtFiles = {};
-  const pages = this.siteConfig.pages || [];
   const processingFiles = [];
+  const pages = this.addressablePages || [];
   this.pageModels = pages.map(page => this.createPageData({
     baseUrl,
-    pageSrc: page.src,
     title: page.title,
+    pageSrc: page.src,
```
nicholaschuayunzhi: we can revert this reordering

lib/Site.js
```diff
@@ -193,6 +208,30 @@ Site.prototype.createPageData = function (config) {
...
+                                 {
+                                   directories: false,
+                                   globs: userGlobs,
+                                   ignore: [BOILERPLATE_FOLDER_NAME],
+                                 });
+  // Add pages collected by walkSync without duplication
+  filePathIndex.forEach((currentPath) => {
+    const index = pages.findIndex(currentPage => currentPage.src === currentPath);
+    if (index === -1) {
+      userPages.push({ src: currentPath });
```
acjh:
```js
globPaths.forEach((globPath) => {
  const index = addressablePages.findIndex(page => page.src === globPath);
  if (index === -1) {
    addressablePages.push({ src: globPath });

File: lib/Site.js [hunk 46]
```diff
@@ -193,6 +208,30 @@ Site.prototype.createPageData = function (config) {
...
 
+
+/**
+ * Builds the index of paths to be traversed as addressable pages
+ */
+Site.prototype.collectAddressablePages = function () {
+  const { pages } = this.siteConfig;
+  const userGlobs = pages.filter(currentPage => currentPage.glob).map(currentGlob => currentGlob.glob);
+  const userPages = pages.filter(currentPage => currentPage.src);
+  const filePathIndex = walkSync(this.rootPath,
```
acjh:
```js
const addressableGlobs = pages.filter(page => page.glob).map(page => page.glob);
const addressablePages = pages.filter(page => page.src);
const globPaths = walkSync(this.rootPath,
```

lib/Site.js
```diff
@@ -193,6 +208,30 @@ Site.prototype.createPageData = function (config) {
   });
 };
 
+
+/**
+ * Builds the index of paths to be traversed as addressable pages
```
acjh: "~`Builds`~`Collects` the ~`index of`~ paths to be traversed as addressable pages"

lib/Site.js
```diff
@@ -247,10 +286,12 @@ Site.prototype.generate = function (baseUrl) {
   fs.emptydirSync(this.outputPath);
   return new Promise((resolve, reject) => {
     this.readSiteConfig(baseUrl)
+      .then(() => this.collectAddressablePages())
       .then(() => this.collectBaseUrl())
       .then(() => this.collectUserDefinedVariablesMap())
       .then(() => this.buildAssets())
       .then(() => this.buildSourceFiles())
+      .then(() => this.writeSiteData())
```
acjh: Can be after `copyMarkBindAsset`.

lib/Site.js
```diff
@@ -407,6 +452,22 @@ Site.prototype.copyMarkBindAsset = function () {
   });
 };
 
+/**
+ * Build data index of the site
```
acjh: "~`Build`~`Writes the site` data ~`index of the site`~`to a file`"

asset/js/setup.js
```diff
@@ -10,30 +10,32 @@ function setup() {
...
-  const routeArray = jQuery.map(siteData.pages, object => object.src);
-  const titleArray = jQuery.map(siteData.pages, object => object.title);
   const { typeahead } = VueStrap.components;
   const vm = new Vue({
     el: '#app',
     components: {
       typeahead,
     },
     data() {
+      const utilMethods = {
```
acjh: Rename to `helpers`.

lib/Page.js
```diff
@@ -71,6 +81,42 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
   });
 };
 
+/**
+ * Records the front matter of the current page into this.frontMatter
+ * Removes the front matter of the current page before it is rendered
```
acjh: ~`of the current page`~
~`of the current page`~

lib/Page.js
```diff
@@ -84,6 +130,7 @@ Page.prototype.generate = function (builtFiles) {
       userDefinedVariablesMap: this.userDefinedVariablesMap,
       rootPath: this.rootPath,
     })
+      .then(result => this.collectFrontMatter(result))
```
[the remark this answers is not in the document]
danielbrzn: Nope, it should not. The front matter is removed during the page generation process.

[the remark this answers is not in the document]
acjh: Does a static include result in a page having front matter?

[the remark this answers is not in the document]
danielbrzn: Yes. It'll inherit the frontmatter from the first static include if the page itself doesn't have its own front matter.

test/test_site/expected/siteData.json
```diff
@@ -0,0 +1,16 @@
+{
+  "pages": [
+    {
+      "title": "L - Hello World",
```
acjh: Where's `L` from?

lib/Page.js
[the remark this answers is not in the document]
acjh: @damithc Is that expected behaviour?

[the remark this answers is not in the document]
damithc: > Yes. It'll inherit the frontmatter from the first static include if the page itself doesn't have its own front matter.

Can give an example where this is useful? My immediate thought is not to treat the _first_ include differently from any other includes.

test/test_site/expected/siteData.json
danielbrzn: Was testing prefixes and forgot to revert.

acjh: Looks like we should add a check for **siteData.json** in [test.js](https://github.com/MarkBind/markbind/blob/8bf51b9b9b687f2c29b0d774ea2fe5410240a130/test/test_site/testUtil/test.js#L22).

lib/Site.js
```diff
@@ -193,6 +206,29 @@ Site.prototype.createPageData = function (config) {
...
+  const { pages } = this.siteConfig;
+  const addressableGlobs = pages.filter(page => page.glob).map(page => page.glob);
+  this.addressablePages = pages.filter(currentPage => currentPage.src);
+  const globPaths = walkSync(this.rootPath,
+                             {
+                               directories: false,
+                               globs: addressableGlobs,
+                               ignore: [BOILERPLATE_FOLDER_NAME],
+                             });
+  // Add pages collected by walkSync without duplication
```
rachx: Use lodash `union` method?

rachx:
Does this fix https://github.com/MarkBind/markbind/issues/122 ?

test/test_site/expected/siteData.json
danielbrzn: Will add this

acjh:
@danielbrzn There are merge conflicts. Can you rebase for a clean diff?

lib/Page.js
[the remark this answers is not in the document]
danielbrzn: One useful scenario is when you want to include front matter from another file. I've updated the user guide with how it can be used here: [here[(https://deploy-preview-175--markbind.netlify.com/userguide/contentauthoring)

The changes to `site.json` are also mentioned [here](https://deploy-preview-175--markbind.netlify.com/userguide/siteconfiguration)

danielbrzn:
@rachx I guess it does.

1) Defaults are handled by the glob pattern
2) Able to specify folders with globs
3) Titles can either be in `site.json` or as front matter.

@acjh Ready!

lib/Page.js
```diff
@@ -71,6 +82,44 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+    const frontMatterWrapped = `${FRONT_MATTER_FENCE}\n${frontMatterData}${FRONT_MATTER_FENCE}`;
+    // Parse front matter data
+    const parsedData = fm(frontMatterWrapped);
+    this.frontMatter = parsedData.attributes;
+    this.frontMatter.src = this.src;
+    // Title specified in site.json will override title specified in front matter
+    this.frontMatter.title = (this.title || this.frontMatter.title || '');
+    if (this.titlePrefix) {
+      this.frontMatter.title = this.frontMatter.title
+        ? this.titlePrefix + TITLE_PREFIX_SEPARATOR + this.frontMatter.title : this.titlePrefix;
```
acjh: This is clearer:

```js
? this.titlePrefix + TITLE_PREFIX_SEPARATOR + this.frontMatter.title
: this.titlePrefix;
```

lib/Page.js
```diff
@@ -71,6 +82,44 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+    }
+  } else {
+    // Page is addressable but no front matter specified
+    const formattedTitle = this.titlePrefix
+      ? this.titlePrefix + (this.title ? TITLE_PREFIX_SEPARATOR + this.title : '') : this.title;
```
acjh: This is clearer:

```js
? this.titlePrefix + (this.title ? TITLE_PREFIX_SEPARATOR + this.title : '')
: this.title;
```

lib/Site.js
```diff
@@ -193,6 +208,25 @@ Site.prototype.createPageData = function (config) {
...
+Site.prototype.collectAddressablePages = function () {
+  const { pages } = this.siteConfig;
+  const addressableGlobs = pages.filter(page => page.glob).map(page => page.glob);
+  this.addressablePages = pages.filter(page => page.src);
+  const globPaths = walkSync(this.rootPath,
+                             {
+                               directories: false,
+                               globs: addressableGlobs,
+                               ignore: [BOILERPLATE_FOLDER_NAME],
+                             }).map(globPath => ({ src: globPath }));
```
acjh:
```js
const globPaths = walkSync(this.rootPath, {
  directories: false,
  globs: addressableGlobs,
  ignore: [BOILERPLATE_FOLDER_NAME],
}).map(globPath => ({ src: globPath }));
```

test/test_site/testUtil/test.js
```diff
@@ -18,16 +21,22 @@ for (let i = 0; i < expectedPaths.length; i += 1) {
...
+    if (actualFilePath.includes('siteData.json')) {
+      if (!_.isEqual(JSON.parse(expected), JSON.parse(actual))) {
+        throw new Error('Site data does not match with the expected file.');
+      }
+    } else {
+      try {
+        diffHtml(expected, actual);
+      } catch (err) {
+        throw new Error(`${err.message} in ${expectedFilePath}`);
+      }
```
acjh: Hmm, since it's quite distinct, this may be clearer:

```js
const parsed = path.parse(actualFilePath);
if (parsed.ext === '.html') {
  // compare html files
  // ...
} else if (parsed.base === 'siteData.json') {
  // compare site data
  // ...
}
```

lib/Site.js
```diff
@@ -13,11 +13,13 @@ const logger = require('./util/logger');
 
 const _ = {};
 _.uniq = require('lodash/uniq');
+_.unionWith = require('lodash/unionWith');
```
acjh: Alphabetical order.

danielbrzn:
Thank you for painstakingly looking through the code 🙏

lib/Page.js
```diff
@@ -71,6 +82,46 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
...
+/**
+ * Records the front matter page into this.frontMatter
+ * Removes the front matter page before it is rendered
+ * @param includedPage a page with its dependencies included
+ */
+Page.prototype.collectFrontMatter = function (includedPage) {
+  const $ = cheerio.load(includedPage);
+  const frontMatter = $('frontmatter');
+  if (frontMatter.length) {
+    const frontMatterData = frontMatter.find('div').length
```
Gisonrg: This block is not very readable. Can you add some comments here?

docs/userGuide/siteConfiguration.md
```diff
@@ -33,7 +37,8 @@ Let's examine a typical `site.json` file:
...
 | **baseUrl** | The url base when you deploy the website. For example, if you are using Github Pages to host your deployed website, and the resulted page is `https://markbind.github.io/site-demo-cs2103`, then your `baseUrl` would be `/site-demo-cs2103` (relative to your domain, default is ` ` (empty)). Later you could use this variable to help you create top-level navigation link in the document. **You may need to change the `baseUrl` when deploy to different repo.** |
-| **pages** | An array of the pages config tells MarkBind all the source files that needs to be rendered. The `src` is the page to be rendered; `title` is the page title for the generated web page. |
+| **titlePrefix** | The prefix for all page titles. This prefix will be prepended to all rendered pages. |
+| **pages** | An array of the pages config tells MarkBind all the source files that needs to be rendered. The `src` is the page to be rendered; `title` is the page title for the generated web page. In addition, globs can be used to define pages that will be rendered. |
```
Gisonrg: Can you explain what happen if a page got collected by globs by don't have the title/frontmatter there?

acjh:
@danielbrzn Any response to @Gisonrg's review?

lib/Page.js
```diff
@@ -135,7 +145,11 @@ Page.prototype.generate = function (builtFiles) {
       userDefinedVariablesMap: this.userDefinedVariablesMap,
       rootPath: this.rootPath,
     })
-      .then(result => this.collectFrontMatter(result))
+      .then((result) => {
+        this.collectFrontMatter(result);
+        return this.removeFrontMatter(result);
+      },
+      )
```
acjh: })

docs/userGuide/siteConfiguration.md
```diff
@@ -33,7 +37,8 @@ Let's examine a typical `site.json` file:
...
-| **pages** | An array of the pages config tells MarkBind all the source files that needs to be rendered. The `src` is the page to be rendered; `title` is the page title for the generated web page. |
+| **titlePrefix** | The prefix for all page titles. This prefix will be prepended to all rendered pages. |
+| **pages** | An array of the pages config tells MarkBind all the source files that needs to be rendered. The `src` is the page to be rendered; `title` is the page title for the generated web page. Titles specified in this array will take priority over titles specified within the front matter of individual pages. In addition, globs can be used to define pages that will be rendered. Front matter for the pages will be automatically indexed. However, if front matter/title is not specified, the page will have a default title that consists only of the titlePrefix. |
```
acjh: - "front matter <code>or \`title`</code>"
- <code>\`titlePrefix`</code>

danielbrzn:
After rebasing, I noticed that `package-lock.json` was not updated, although new dependencies were added like `vows` and `winston` were added. @rachx

acjh:
> After rebasing, I noticed that `package-lock.json` was not updated, although new dependencies like `vows` and `winston` were added.

Thanks for highlighting; missed it in my review. Is `vows` a typo?

The build failure seems to be a different *matter* though.

danielbrzn:
Nope, `vows` isn't a typo.

image

nicholaschuayunzhi:
Seems like `yarn` is not resolving the dependencies for `htmlparser` correctly and we are reverting to an old version of `htmlparser`.

Currently when travis detects `yarn.lock` in the project and installs the modules using yarn which results in the wrong html parser being installed. I have tried regenerating the yarn.lock but it seems like it doesnt work. 

The solution i have is either
- Delete `yarn.lock`
  - i'd prefer this as we do not use yarn as our package manager
- Force `npm install` on travis

These have been tested on: https://github.com/nicholaschuayunzhi/markbind/pull/5
https://travis-ci.org/nicholaschuayunzhi/markbind

danielbrzn:
Looking at the `yarn.lock` that exists in our repo, it does seem that it's pulling the wrong version of `htmlparser2` as shown below. v3.9.2 is the last published version of the module.

image

I think that using `npm` on travis would be ideal since it more closely resembles the environment that we actually use.

nicholaschuayunzhi:
@Gisonrg do we need the yarn.lock?

acjh:
**yarn.lock** hasn't been updated in 8 months. Any idea why this wasn't a problem in previous builds?

acjh:
For reference:

```diff
-  <strong><code>As a</code></strong> <code>&lt;use type/role&gt;</code> <strong><code>I can</code></strong> <code>&lt;function&gt;</code> <strong><code>so that</code></strong> <code>&lt;benefit&gt;</code></p>
+  <strong><code>As a</code></strong> <code>&lt;use type=&quot;&quot; role=&quot;&quot;&gt;</code> <strong><code>I can</code></strong> <code>&lt;function&gt;</code> <strong><code>so that</code></strong> <code>&lt;benefit&gt;</code></p>
```

In particular:

```diff
- <code>&lt;use type/role&gt;</code>
+ <code>&lt;use type=&quot;&quot; role=&quot;&quot;&gt;</code>
```

acjh:
> The solution i have is either
> - Delete `yarn.lock`
>   - i'd prefer this as we do not use yarn as our package manager
> - Force `npm install` on travis

Let's proceed with the former.

danielbrzn:
Getting a pretty odd test failure, @nicholaschuayunzhi any idea why?

nicholaschuayunzhi:
I didnt mock `writeJsonAsync`. I suggest you use `outputJsonAsync` instead as i have mocked it and it also creates the directory if it does not exist.

acjh:
Hmm, there should not be changes to **vue-strap.min.js**.

danielbrzn:
Updated.

acjh:
@danielbrzn You can merge this :)

PR #178, opened 2018-03-23
Add jest for unit testing

nicholaschuayunzhi:
Continues from the PR https://github.com/MarkBind/markbind-cli/pull/28

Currently working to add tests for `index.js` and `Page.js`.

I need advice with creating tests for boilerplates, user defined variables and different cli commands. Also, how should i test `parser.js`?


Gisonrg:
For `parser.js` we are actually interested in `_preprocess` and `_parse`, but they seem to has large and complicated logic though. If it's too difficult, since it is already covered in the testing with test site, it might be fine for now (you can start the review with what you have for now so it's easier and faster to review) :)

test/unit/Page.test.js
```diff
@@ -0,0 +1,15 @@
+const Page = require('../../lib/Page');
+
+test('Page#collectIncludedFiles collects included files from 1 dependency object', () => {
+  const page = new Page({});
+  page.collectIncludedFiles([{ to: 'somewhere' }]);
+
+  expect(page.includedFiles).toEqual({ somewhere: true });
+});
+
+test('Page#collectIncludedFiles ignores other keys in dependency', () => {
```
danielbrzn: Perhaps in addition to these tests, a test for `Page#collectIncludedFiles` to not include anything from a dependency object could be made as well?

test/unit/Site.test.js
```diff
@@ -0,0 +1,323 @@
...
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/css/markbind.css'))).toEqual(true);
+
+  // js
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/js'))).toEqual(true);
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/js/setup.js'))).toEqual(true);
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/js/vue.min.js'))).toEqual(true);
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/js/vue-strap.min.js'))).toEqual(true);
+
+  // fonts
+  expect(fs.existsSync(path.resolve('inner/_site/markbind/fonts/glyphicons-halflings-regular.eot')))
```
danielbrzn: For consistency with the above tests, you should add a check for the directory `inner/_site/markbind/fonts`.

Actually, could it be removed? The check for the existence of the file implicitly checks for the existence of the directory.

test/unit/Site.test.js
```diff
@@ -0,0 +1,323 @@
...
+
+  // _boilerplates
+  expect(fs.existsSync(path.resolve('_markbind/boilerplates'))).toEqual(true);
+
+  // user defined variables
+  expect(fs.existsSync(path.resolve('_markbind/variables.md'))).toEqual(true);
+  expect(fs.readFileSync(path.resolve('_markbind/variables.md'), 'utf8')).toEqual(USER_VARIABLES_DEFAULT);
+
+  // site.json
+  expect(fs.existsSync(path.resolve('site.json'))).toEqual(true);
```
danielbrzn: `siteData.json` will also be generated in the future once https://github.com/MarkBind/markbind/pull/175 is merged.

Should I commit on top of your branch to add tests for this?

test/unit/Site.test.js
```diff
@@ -0,0 +1,323 @@
...
+  };
+  fs.vol.fromJSON(json, '');
+
+  const site = new Site('./', '_site');
+  await site.removeAsset('toRemove.jpg');
+  expect(fs.existsSync(path.resolve('_site/toRemove.jpg'))).toEqual(false);
+  expect(fs.existsSync(path.resolve('_site/dontRemove.png'))).toEqual(true);
+});
+
+test('Site baseurls are correct for sub nested subsites', async () => {
```
danielbrzn: This test is the same as the above one.

test/unit/Site.test.js
```diff
@@ -0,0 +1,323 @@
...
+
+  const site = new Site('./', '_site');
+  await site.readSiteConfig();
+  expect(site.siteConfig).toEqual(JSON.parse(SITE_JSON_DEFAULT));
+});
+
+test('Site read site config for custom site config', async () => {
+  const customSiteJson = {
+    baseUrl: '',
+    pages: [
```
danielbrzn: This `site.json` will have to be changed to have the new attribute `addressable` once https://github.com/MarkBind/markbind/pull/175 is merged.

acjh:
@nicholaschuayunzhi Any response to @danielbrzn's review?

nicholaschuayunzhi:
@danielbrzn @acjh, sorry for the late reply. I have updated the pr according to your comments. I agree that   we can just for the content of the file. If the file does not exist, the test will still fail but with an error which could be more beneficial for debugging instead of a:
```
    expect(received).toEqual(expected)
    Expected value to equal:
      true
    Received:
      false
```
Considering that you have allowed for `site.json` to use the key `pages` as well as globs in `pages`, i think we can keep the test as it is. Then we can write separate unit tests for the new changes.

Regarding `siteData.json`, we do need to update the test to account for it.

acjh:
> Regarding `siteData.json`, we do need to update the test to account for it.

Can be a separate PR.

__mocks__/fs-extra-promise.js
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Utils
+ */
+
+/**
+ * Remove directory recursively
+ * @param {string} dirPath
+ * @see https://stackoverflow.com/a/42505874/3027390
+ */
+function rimraf(dirPath) {
```
Gisonrg: Couldn't tell the function's usage from the name...should we rename it?

test/unit/Site.test.js
```diff
@@ -0,0 +1,294 @@
...
+  fs.vol.fromJSON(json, '');
+  const site = new Site('inner/', 'inner/_site');
+  await site.generate();
+  const paths = Object.keys(fs.vol.toJSON());
+  const originalNumFiles = 13;
+  const expectedNumBuilt = 11;
+  expect(paths.length).toEqual(originalNumFiles + expectedNumBuilt);
+
+  // site
+  expect(fs.existsSync(path.resolve('inner/_site'))).toEqual(true);
```
Gisonrg: I think `.toEqual(true)` is same as `.toBeTruthy()`

ref: https://facebook.github.io/jest/docs/en/using-matchers.html#truthiness

test/unit/Site.test.js
Gisonrg: Actually I think it's ok not checking the font files... they won't block the site being rendering in the browser anyway :P

test/unit/Site.test.js
acjh: `.toEqual(true)` is a subset of `.toBeTruthy()`. I think the former is more correct.

nicholaschuayunzhi: `toBeTruthy` tests if a value is true in a boolean context
`toBe` checks for exact equality using Object.is
`toEqual` is the same as toBe for Boolean values but for object it matches properties instead of identity 

```js
expect(1).toBeTruthy(); // passes
expect(1).toBe(true); // fails
expect(1).toEqual(true); // fails

const obj1 = { a: 1 };
const obj2 =  { a: 1 };

expect(obj1).toBe(obj2); // fails
expect(obj1).toEqual(obj2); // passes
```

Personally i would either use `.toBe` or `.toEqual`. Is it safe to assume that fs.existsSync will return a boolean? If that is the case then `.toBeTruthy` wlll be safe to use.

nicholaschuayunzhi:
@acjh, I have rebased on top of master and **mocked logging**. Am I able to merge this?

acjh:

PR #186, opened 2018-03-30
Add option to specify baseurl in build command

nicholaschuayunzhi:
Gives users command line argument to specify base url. This is useful for build for preview sites, eg: on Netlify.

Usage:

| command | what it does |
|---|---|
|`markbind build --baseurl override-base-url`| build site with baseUrl: "override-base-url"|
|`markbind build --baseurl`| build site with baseUrl: ""|

Things to note:

Option is `--baseurl` and **is** case-sensitive, was wondering if this would be better as the user can type this more quickly? Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,19 +16,21 @@ npm i -g markbind-cli
   Usage: markbind  <command>
 
 
-    Commands:
+  Options:
 
-      include [options] <file>  process all the fragment include in the given file
-      render [options] <file>   render the given file
-      init [root]               init a markbind website project
-      serve [options] [root]    build then serve a website from a directory
-      build [root] [output]     build a website
-      deploy                    deploy to Github Pages
+    -V, --version  output the version number
+    -h, --help     output usage information
 
-    Options:
 
-      -h, --help     output usage information
-      -V, --version  output the version number
+  Commands:
+
+    include [options] <file>         process all the fragment include in the given file
+    render [options] <file>          render the given file
+    init [root]                      init a markbind website project
+    serve [options] [root]           build then serve a website from a directory
+    deploy                           deploy the site to the repo's Github pages.
+    build [options] [root] [output]  build a website
+
 ```
 
 ## Documentation
--- a/index.js
+++ b/index.js
@@ -11,6 +11,9 @@ const htmlBeautify = require('js-beautify').html;
 const liveServer = require('live-server');
 const chokidar = require('chokidar');
 
+const _ = {};
+_.isBoolean = require('lodash/isBoolean');
+
 const logger = require('./lib/util/logger');
 const fsUtil = require('./lib/util/fsUtil');
 const Site = require('./lib/Site');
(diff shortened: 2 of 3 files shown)


damithc:
> Option is --baseurl and is case-sensitive, was wondering if this would be better as the user can type this more quickly? Any thoughts?

May be use `baseurl` for both as CLI option and in code i.e., `{{baseurl}}`?

lib/Site.js
```diff
@@ -133,13 +133,14 @@ Site.initSite = function (rootPath) {
...
 
-Site.prototype.readSiteConfig = function () {
+Site.prototype.readSiteConfig = function (baseUrl) {
   return new Promise((resolve, reject) => {
     const siteConfigPath = path.join(this.rootPath, SITE_CONFIG_NAME);
     fs.readJsonAsync(siteConfigPath)
       .then((config) => {
         this.siteConfig = config;
-        resolve(config);
+        this.siteConfig.baseUrl = baseUrl || this.siteConfig.baseUrl;
```
acjh: I believe this won't work when `baseUrl` is an empty string.

acjh:
> May be use `baseurl` for both as CLI option and in code i.e., `{{baseurl}}`?

[`baseUrl`](https://www.google.com.sg/search?q="baseUrl"+node+-typescript) seems more common in Node.js, though Jekyll uses [`baseurl`](https://github.com/jekyll/jekyll/search?q=baseurl).

@Gisonrg Your thoughts?

lib/Site.js
nicholaschuayunzhi: Ah yes my bad. see: https://stackoverflow.com/a/3088161

nicholaschuayunzhi:
Perhaps I should keep it as `baseUrl` first, to stay consistent with the current semantics?

Gisonrg:
Any one would do in my opinion :)

index.js
```diff
@@ -208,14 +211,17 @@ program
 
 program
   .command('build [root] [output]')
+  .option('--baseUrl [baseUrl]', 'overrides baseUrl in site.json, leave argument empty for empty baseUrl')
```
Gisonrg: `leave argument empty for empty baseUrl`

Maybe just say default value is empty base url

index.js
```diff
@@ -208,14 +211,17 @@ program
...
   .description('build a website')
-  .action((root, output) => {
+  .action((root, output, cmd) => {
```
Gisonrg: Follow the variable naming of other command: s/cmd/option/

nicholaschuayunzhi:
Thanks for the review @Gisonrg, regarding https://github.com/MarkBind/markbind/pull/186#discussion_r178843691,

I've changed it to 
```
optional flag which overrides baseUrl in site.json, leave argument empty for empty baseUrl
```

To make it clearer. I thinking using the wore `overrides` implies to the user that the original behaviour would be using `baseUrl in site.json`

README.md
```diff
@@ -13,22 +13,24 @@ npm i -g markbind-cli
 
 ## Usage
 ```
-  Usage: markbind  <command>
+  Usage: index  <command>
```
Gisonrg: markbind

Gisonrg:
>Did you mean to change cmd to options? 

Yea since the name `options` is more readable and is used in `serve` command :)

acjh:

damithc:
🤔 Shouldn't we mention this new option in the README.md?

damithc:
@nicholaschuayunzhi this feature doesn't seem to be documented in the user guide

PR #190, opened 2018-04-12
Deploy changes to user guide on Netlify

nicholaschuayunzhi:
This pr introduces a netlify configuration which forces Netlify to install the `package.json` dependencies in `docs/` which contains the latest published version of `markbind-cli`, currently `v1.6.3`.

We cannot specify this dependency in `package.json` in the root directory as npm does not allow a dependency to be the same as the package name.

To set up Netlify:

https://deploy-preview-2--markbind-nic.netlify.com/userguide/netlifypreview

You can ignore the project setup (first 4 steps) and continue with the Netlify setup.

In step 5 , you can leave the build and publish directory to be blank.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,0 +1,2520 @@
+{
+  "name": "markbind-docs",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "a-sync-waterfall": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/a-sync-waterfall/-/a-sync-waterfall-1.0.0.tgz",
+      "integrity": "sha1-OOgxnXk3niRiiEW1O5ZyKyng5Hw="
+    },
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+    },
+    "accepts": {
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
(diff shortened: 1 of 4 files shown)


nicholaschuayunzhi:
> Users need to update package.json and package.lock.json every time they upgrade markbind?

If you are referring to markbind site repositories, the site owner has to do it and the version for markbind should reflect the version he wants the site to be built in so that Netlify can install the correct version of markbind in their environment and build.

On a related note:

Currently what we advise is to install markbind globally `$ npm install markbind-cli -g` which allows us to to use markbind as a command. 

I am thinking eventually, site owners should use a local version of markbind (specified by package.json) as some sites may depend on older version of markbind. This can be specified by doing `$ npm install markbind-cli --save`.

nicholaschuayunzhi:
I have updated the [user guide](https://deploy-preview-2--markbind-nic.netlify.com/userguide/netlifypreview) to show a simple way to set up Netlify. The catch is that Netlify will always use the latest version of MarkBind when building. However, it does not require any `package.json` and `package.lock.json`. 

To use a specific version of MarkBind, we would still need to follow the old workflow which i have kept in the site.

damithc:
Got it. 👍

damithc:
Does this setup work for markbind/markbind? 
Note that the site is inside the /docs folder, not the root folder

I tried the setup and got this error
image

damithc:
When showing how to fill up netlify settings, show the screenshot  of the empty form and give the values as regular text. That way, it is easy to update in future and people can copy paste the value from our doc to netlify.

damithc:
On a separate but related note, I wonder if this will work for repos that has submodules, such as CS2103 website.

damithc:
@nicholaschuayunzhi I invited you as an admin to the netlify for this site

nicholaschuayunzhi:
I've deployed a fork of 2103 website on netlify. 
https://heuristic-jones-050f39.netlify.com/

Seems like the textbook is built https://heuristic-jones-050f39.netlify.com/book/index.html
But i have not fully checked it.

nicholaschuayunzhi:
@damithc , seems like i have solution for our repository that lets us preview our documentation without this .toml file while still installing only the latest published version of `markbind-cli`

> When showing how to fill up netlify settings, show the screenshot of the empty form and give the values as regular text. That way, it is easy to update in future and people can copy paste the value from our doc to netlify.

PR #192, opened 2018-04-17
Replace nunjucks with mustache.js

rachx:
Currently, we only use nunjucks to insert the values of baseUrl and user-defined variables.

nunjucks does way more than that e.g. support math operations. This makes variable names like `{{ gly-search }}` not possible as - is taken as the math minus symbol.

We can use a simpler library like mustache.js instead


acjh:
Hmm, I'm not sure if we should go for this. 😕

PR #200, opened 2018-04-23
Fix included front matter parsing

danielbrzn:
This PR aims to fix the parsing of front matter that is included from other files.

```
<seg id="bst">
  <frontmatter>
    title: Binary Search Tree
    keywords: bst, avl, red-black
  </frontmatter>
</seg>
```

Using the above snippet as an example (taken from the [User Guide](https://markbind.github.io/markbind/userGuide/contentAuthoring.html)), the whitespace from the indentation before the closing `</frontmatter>` tag affects the parsing of the data and causes it not to be captured properly. By adding a newline to properly wrap the front matter, the parsing is done correctly.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -96,7 +96,7 @@ Page.prototype.collectFrontMatter = function (includedPage) {
     const frontMatterData = frontMatter.find('div').length
       ? frontMatter.find('div')[0].children[0].data
       : frontMatter[0].children[0].data;
-    const frontMatterWrapped = `${FRONT_MATTER_FENCE}\n${frontMatterData}${FRONT_MATTER_FENCE}`;
+    const frontMatterWrapped = `${FRONT_MATTER_FENCE}\n${frontMatterData}\n${FRONT_MATTER_FENCE}`;
     // Parse front matter data
     const parsedData = fm(frontMatterWrapped);
     this.frontMatter = parsedData.attributes;


danielbrzn:
`setup.js` needs more fine tuning, will ping when ready for review. @acjh

danielbrzn:
Ready! @acjh

asset/js/setup.js
```diff
@@ -9,6 +9,13 @@ function setup() {
   VueStrap.installEvents(vm);
 }
 
+let baseUrl = window.location.origin;
+
+// environment check
+if (window.location.href.includes('github.io')) {
+  baseUrl = `${baseUrl}/website`;
```
acjh: Hmm, that's hardcoding the repository name...

danielbrzn: What about splitting the `window.location.pathname` and taking the first result? 

e.g.
```javascript
if (window.location.href.includes('github.io')) {
baseUrl = `${baseUrl}/${window.location.pathname.split('/')[1]}`; // http://username.github.io/reponame
}
```

acjh: Better. Probably a stricter check:

```js
if (window.location.origin.endsWith('github.io')) {
  const repository = window.location.pathname.split('/')[1];
  baseUrl += '/' + repository;
}
```

But this still wouldn't work for sites hosted elsewhere. Could we use nunjucks render when copying this asset?

@Gisonrg What do you think?

danielbrzn: That's true. I suppose it would be possible, so all the instances of baseUrl will be replaced by the author's specified URL? 

After which, we can use `window.location.origin` and concatenate with the baseUrl.

acjh: Yes.

Actually, `window.location.origin` prefix is unnecessary:

```js
let baseUrl = '';

// Support gh-pages
if (window.location.origin.endsWith('github.io')) {
  const repository = window.location.pathname.split('/')[1];
  baseUrl = `/${repository}`;
}
```

PR #206, opened 2018-04-25
Change logging setting to be based on maximum files

rachx:


acjh:
@rachx You can merge this :)

PR #207, opened 2018-04-25
Assign build-time baseUrl in page template

nicholaschuayunzhi:
`baseUrl` is recorded based on build time baseUrl.

This ensures that `baseUrl` is the one author specifies.
- using baseUrl in site.json 
- using the `--baseUrl` flag in `markbind build`

Thanks @rachx for suggesting solution.


nicholaschuayunzhi:
I have tested here: https://nicholaschuayunzhi.github.io/website-1/
Please note that i had to change my baseUrl to website-1 because i have to repositories named website 😅

lib/template/page.ejs
```diff
@@ -16,5 +16,6 @@
 </body>
 <script src="<%- asset.vue %>"></script>
 <script src="<%- asset.vueStrap %>"></script>
+<script>const baseUrl = '<%= baseUrl %>'</script>
```
acjh: Use block to differentiate this:

```html
<script>
    const baseUrl = '<%- baseUrl %>';
</script>
```

If we don't need to escape it (`<%=`), let's keep it consistent (`<%-`).

acjh: Trailing semicolon!

acjh:

PR #214, opened 2018-05-16
Add warning: body tag may cause formatting errors

### LINKED ISSUE #148: <body> tag in included HTML causes formatting errors
Wrapping the content of a HTML document within a `<body>` tag disrupts the formatting of the entire site. 

Actual:
image

HTML:
```
<body><p>This is a HTML document.</p></body>
```

Expected:
image

HTML:
```
<p>This is a HTML document.</p>
```

acjh:

• [x] Other, please explain: Add warning for a potential user (author) error

Closes #148 

> when you create a html file in webstorm, they auto generate the formatting for you with the body tag.

> a user may be wondering why the formatting is messed up when he/she tries to include the HTML file.

Explanation: https://stackoverflow.com/questions/2035462/multiple-htmlbody-html-body-in-same-file

- In `Parser.prototype._preprocess` if `element.name === 'body'` then `console.warn`.

Documentation seems unnecessary as this is not MarkBind-specific.

```html
<body><p>This is a HTML document.</p></body>
```

No change in behaviour except a warning in the console.

\-

**Testing instructions:**

- Add the example code above in **test/test_site/index.md** and run `$ markbind serve`.

There should be a warning, but MarkBind is able to build. The formatting is messed up.

### DIFF
--- a/lib/markbind/lib/parser.js
+++ b/lib/markbind/lib/parser.js
@@ -239,8 +239,14 @@ Parser.prototype._preprocess = function (node, context, config) {
     element.attribs.src = filePath;
     this.dynamicIncludeSrc.push({ from: context.cwf, to: actualFilePath, asIfTo: filePath });
     return element;
-  } else if (element.children && element.children.length > 0) {
-    element.children = element.children.map(e => self._preprocess(e, context, config));
+  } else {
+    if (element.name === 'body') {
+      // eslint-disable-next-line no-console
+      console.warn(`<body> tag found in ${element.attribs[ATTRIB_CWF]}. This may cause formatting errors.`);
+    }
+    if (element.children && element.children.length > 0) {
+      element.children = element.children.map(e => self._preprocess(e, context, config));
+    }
   }
 
   return element;


damithc:
If a site has submodules which are also markbind sites (e.g., cs2103 website), building a submodule creates a `_site` folder containing a lot of html files. The body tag in those files can trigger this warning when building the outer site.

acjh:
I couldn't reproduce that by running:

```
$ markbind serve book
$ markbind serve
```

What's the warning message?

damithc:
What I did was:
```
cd book
markbind serve
cd ..
markbind serve
```

Got about thousand warnings ... about the body tag in each html file.

yamgent:
Can't reproduce it on my side either, even with the `cd book`, `markbind serve`, `cd ..` and `markbind serve`.

> building a submodule creates a _site folder containing a lot of html files. The body tag in those files can trigger this warning when building the outer site.

That seems kind of strange though, since markbind shouldn't traverse through ignored directories. Is `_site` still listed as an ignored directory in the `site.json` configuration?

```json
  "ignore": [
    "_site/*",
```

damithc:
Try with https://github.com/nus-cs2103/website-base
I was able to reproduce it just now with a fresh set up. No changes to site.json in repo.

acjh:
Repro-ed.

> No changes to site.json in repo.

Actually, there's a major change from using `src` to `glob`. The `glob` includes pages in **book**.
The fix is to use `"glob": "{,!(book)/**/}index.md"` etc. instead of `"glob": "**/index.md"`.

yamgent:
Now able to reproduce it.

> The fix is to use "glob": "{,!(book)/**/}index.md" etc. instead of "glob": "**/index.md".

Is there a reason why we are not making `ignore` a higher priority to `glob`? It doesn't seem very obvious to the user that the ignore for `_site` is being overridden.

acjh:
It's actually unrelated.
`ignore` just decides which files to copy to `_site`. Note that there's `site.json`, `*.md` etc.
`glob` explicitly asks MarkBind to "process files that match this pattern and generate pages".

yamgent:
> `ignore` just decides which files to copy to `_site`. Note that there's s`ite.json`, `*.md etc`.
> `glob` explicitly asks MarkBind to "process files that match this pattern and generate pages".

Hmm OK, perhaps we can implement a similar `ignore` feature for `pages` then. Then one can just exclude an entire directory instead of having to use regex to do the same.

PR #210, opened 2018-05-15
Support favicon

yamgent:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] Documentation update
• [X] New feature

Fixes #96.

**What is the rationale for this request?**
User would like to add a favicon to their website. This is only possible if the HTML `<head>` tag contains a link to the image of the favicon.

**What changes did you make? (Give an overview)**
* Modified `page.ejs` to insert the favicon path if it is enabled.
* Added the `favIconPath` parameter in `site.json` and passed it to the page template generator.
* Modified the user guide to mention about the new parameter.

**Provide some example code that this change will affect:**

After:
```html
<html>
  <head>
    ...
    <link rel="icon" href="favicon.png"> 
    ...
  </head>
   ...
</html>
```

**Is there anything you'd like reviewers to focus on?**
The documentation & tests.

**Testing instructions:**
1. Insert an additional line to a markbind website's `site.json`:
```json
"favIconPath": "favicon.png",
```
1. Add a `favicon.png` to the root directory of the markbind website.
1. Execute `markbind serve`.
1. The new webpage should now show the favicon on the browser's title bar.

PR #221, opened 2018-05-18
Favicon: fix missing starting slash if baseUrl is empty

yamgent:

• [X] Bug fix

Follows https://github.com/MarkBind/markbind/pull/210#issuecomment-389815226.

favicon is usually root-relative (e.g. `/favicon.png`), but if `baseUrl` is empty, then the code generates an incorrect path for favicon (i.e. no longer root-relative, e.g. `favicon.png`).

When generating favicon, the first parameter to `url.join()` calls will start with '/', so that a starting slash is always added regardless of the `baseUrl` value.

Before:

```html
<link rel="icon" href="favicon.png">
```

After:

```html
<link rel="icon" href="/favicon.png">
```

NIL

**Testing instructions:**

1. Use the following settings for `site.json`:
```json
"baseUrl": "",
"faviconPath": "favicon.png",
```
1. Do `markbind serve`.
1. The generated html file should have the favicon URL specified with a starting `/`:
```html
<link rel="icon" href="/favicon.png">
```

### DIFF
--- a/lib/Site.js
+++ b/lib/Site.js
@@ -428,12 +428,12 @@ Site.prototype.generatePages = function () {
 
   let faviconUrl;
   if (faviconPath) {
-    faviconUrl = url.join(baseUrl, faviconPath);
+    faviconUrl = url.join('/', baseUrl, faviconPath);
     if (!fs.existsSync(path.join(this.rootPath, faviconPath))) {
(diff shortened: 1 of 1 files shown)


yamgent:
I also verified that `url.join('/', '/website');` will return `/website`, **not** `//website`.

PR #231, opened 2018-05-22
Allow variables to reference earlier variables

### LINKED ISSUE #230: Allow meta-variables
I was trying to do the following in my `variables.md` so that I can use a more meaningful name for my icon. 
`<div id="icon_prereq">{{glyphicon_education}}</div>`

That is, I can write `{{icon_prereq}}` instead of `{{glyphicon_education}}` in my code.

Looks like it (i.e., including a variable in the value of a variable) cannot be done at the moment (v1.7.4)

yamgent:

• [X] New feature

This also indirectly fixes #230. However, a deliberate design consideration was made: we restrict the referencing to only allow earlier declared variables (yes, glyphicons counts as "earlier declared variables").

```html
<!-- Allowed: -->
<span id="some_icon">{{glyphicon_education}}</span>

<!-- Allowed: -->
<span id="first">This is the first variable.</span>
<span id="second">{{first}}</span> <!-- The value will be rendered correctly -->

<!-- Forbidden: -->
<span id="third">{{fourth}}</span> <!-- Nothing will be rendered -->
<span id="fourth">This variable is declared AFTER third.</span>
```

The restriction is enforced to avoid circular referencing:
```html
<!-- Without the restriction, circular referencing is possible -->
<span id="first">{{second}}</span>
<span id="second">{{first}}</span>
```

User wants to be able to refer to glyphicons inside the content of a variable. This can be partially achieved by allowing variables to refer to earlier declared variables (glyphicons are declared internally in MarkBind before `variables.md` is parsed).

Made nunjucks parse the contents of variables declared in `variables.md` one by one, earliest first.

```html
<div id="icon_prereq">{{glyphicon_education}}</div>
```

_Before:_ Nothing is rendered.
_After:_ The icon is rendered correctly.

-----

```html
<div id="first">First variable</div>
<div id="second">{{first}}</div>
```

When using `{{second}}`:
_Before:_ Nothing is rendered.
_After:_ The value "First variable" is shown.

Documentation

~~The documentation sample code. Is there a way to escape the `{{}}` inside a multi-line code?~~

~~Things I tried:~~
* ~~Using `<span></span>` trick: Doesn't work, the span tag is rendered.~~
* ~~Using [`{% raw %}`](https://mozilla.github.io/nunjucks/templating.html#raw): Unfortunately this always gets overridden because we call nunjucks multiple times (last time I debug it was called twice).~~
* ~~Using `{% raw %}` twice: The temporary workaround works... until vue comes along and removes it, because it is not wrapped with [`<span v-pre>`](https://vuejs.org/v2/api/#v-pre).~~
* ~~Using `<span v-pre>`, `<markdown>`, `{% raw %}`: Doesn't work inside code snippets or just doesn't render properly.~~

**Testing instructions:**
1. Declare new variables inside `variables.md`.
```html
<span id="icon_prereq">{{glyphicon_education}}</span>
<span id="forty_two">42</span>
<span id="some_value">{{forty_two}}</span>
```
2. Use the variables in a markdown file (e.g. `index.md`):
```
{{icon_prereq}}

{{some_value}}
```
3. The icon and the value should be rendered properly:
iconand42

### DIFF
--- a/docs/userGuide/contentAuthoring.md
+++ b/docs/userGuide/contentAuthoring.md
@@ -161,6 +161,29 @@ Each variable must have an id and the value can be any MarkBind-compliant code f
 
 In any file in your site, you can include the variable by surrounding the variable's id with double curly braces, e.g. <code>{<span></span>{options}}</code>
 
+Variables can refer to other variables that are declared earlier, including helpful variables provided by MarkBind.
+
+<tip-box>
+<div>
+This is allowed, MarkBind helpful variables can be used.<br>
+
+<code>\<span id="right_hand">{<span></span>{glyphicon_hand_right}}\</span></code>
+</div>
+<div>
+This is also allowed, the second variable will be assigned the contents of the first variable.
+
+<code>\<span id="first">This is the first variable.\</span></code><br>
+<code>\<span id="second">{<span></span>{first}}\</span></code><br>
+<code>\<span id="third">{<span></span>{second}}\</span></code>
+</div>
+<div>
+This is forbidden, as the fourth variable is not declared yet.
+
+<code>\<span id="third">{<span></span>{fourth}}\</span></code><br>
+<code>\<span id="fourth">This is the fourth variable.\</span></code>
+</div>
+</tip-box>
+
 ### Use Searchbar
 
 To use the searchbar, add the following markup to your file.
--- a/docs/userGuide/contentAuthoring.md
+++ b/docs/userGuide/contentAuthoring.md
@@ -177,7 +177,7 @@ This is also allowed, the second variable will be assigned the contents of the f
 <code>\<span id="third">{<span></span>{second}}\</span></code>
 </div>
 <div>
-This is forbidden, as the fourth variable is not declared yet.
+This is forbidden, as the fourth variable is not declared yet, and will not be rendered correctly.
 
 <code>\<span id="third">{<span></span>{fourth}}\</span></code><br>
 <code>\<span id="fourth">This is the fourth variable.\</span></code>
--- a/docs/userGuide/contentAuthoring.md
+++ b/docs/userGuide/contentAuthoring.md
@@ -184,6 +184,41 @@ This is forbidden, as the fourth variable is not declared yet, and will not be r
 </div>
 </tip-box>
 
+Note: If the variable being referenced contains HTML tags, MarkBind may escape the tags and render it literally. For example:
+
+<tip-box>
+<div>
+If we declare the variables as follow:<br>
+
+<code>\<span id="right_hand">{<span></span>{glyphicon_hand_right}}\</span></code><br>
+<code>\<span id="right_hand_2">{<span></span>{right_hand}}\</span></code>
+</div>
+<div>
+The following will be rendered:<br>
+
+{{glyphicon_hand_right}}<br>
+\<span class='glyphicon glyphicon-hand-right' aria-hidden='true'></span>
+</div>
+</tip-box>
+
+You must use the `safe` filter when using such variables:
+
+<tip-box>
+<div>
+If we use the safe filter for the second variable:<br>
+
+<code>\<span id="right_hand">{<span></span>{glyphicon_hand_right}}\</span></code><br>
+<code>\<span id="right_hand_2">{<span></span>{right_hand | safe}}\</span></code>
+</div>
+<div>
+The following will be rendered:<br>
+
+{{glyphicon_hand_right}}<br>
+{{glyphicon_hand_right}}
+</div>
+</tip-box>
+
+
 ### Use Searchbar
 
 To use the searchbar, add the following markup to your file.
--- a/lib/Site.js
+++ b/lib/Site.js
@@ -6,6 +6,7 @@ const findUp = require('find-up');
 const fs = require('fs-extra-promise');
 const ghpages = require('gh-pages');
 const ignore = require('ignore');
+const nunjucks = require('nunjucks');
 const path = require('path');
 const Promise = require('bluebird');
 const walkSync = require('walk-sync');
@@ -295,7 +296,8 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
     const $ = cheerio.load(content);
     $.root().children().each(function () {
       const id = $(this).attr('id');
-      const html = $(this).html();
+      // Process the content of the variable with nunjucks, in case it refers to other variables.
+      const html = nunjucks.renderString($(this).html(), userDefinedVariables);
       userDefinedVariables[id] = html;
     });
 
--- a/test/test_site/index.md
+++ b/test/test_site/index.md
@@ -11,6 +11,8 @@ title: Hello World
 
 {{finalized_value}}
 
+{{reference_level_4}}
+
 # Normal include
 <include src="requirements/EstablishingRequirements.md" />
 
--- a/test/test_site/index.md
+++ b/test/test_site/index.md
@@ -6,6 +6,11 @@ title: Hello World
 
 <div class="website-content">
 
+# Variables that reference another variable
+{{education_icon}}
+
+{{finalized_value}}
+
 # Normal include
 <include src="requirements/EstablishingRequirements.md" />
 
--- a/test/unit/Site.test.js
+++ b/test/unit/Site.test.js
@@ -189,6 +189,32 @@ test('Site read site config for custom site config', async () => {
   expect(site.siteConfig).toEqual(customSiteJson);
 });
 
+test('Site resolves variables referencing other variables', async () => {
+  const json = {
+    'lib/asset/glyphicons.csv': '',
+    'lib/template/page.ejs': PAGE_EJS,
+    'site.json': SITE_JSON_DEFAULT,
+    '_markbind/variables.md':
+    '<span id="level1">variable</span>'
+    + '<span id="level2">{{level1}}</span>'
+    + '<span id="level3">{{level2}}</span>'
+    + '<span id="level4">{{level3}}</span>',
+  };
+  fs.vol.fromJSON(json, '');
+
+  const site = new Site('./', '_site');
+  await site.collectBaseUrl();
+  await site.collectUserDefinedVariablesMap();
+
+  const root = site.userDefinedVariablesMap[path.resolve('')];
+
+  // check all variables
+  expect(root.level1).toEqual('variable');
+  expect(root.level2).toEqual('variable');
+  expect(root.level3).toEqual('variable');
+  expect(root.level4).toEqual('variable');
+});
+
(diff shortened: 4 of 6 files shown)


docs/userGuide/contentAuthoring.md
```diff
@@ -161,6 +161,22 @@ Each variable must have an id and the value can be any MarkBind-compliant code f
...
+<!-- This is allowed, the second variable will be assigned the contents of the first variable. -->
+<span id="first">This is the first variable.</span>
+<span id="second">{ {first}}</span>
+<span id="third">{ {second}}</span>
+
+<!-- This is forbidden, the fourth variable is not declared yet. -->
+<span id="third">{ {fourth}}</span>
+<span id="fourth">This is the fourth variable.</span>
+```
+
```
Chng-Zhi-Xuan: Changed text so that additional space not required. Copy pasting the updated text into `diffChecker` has no issues as well

```
<tip-box>
<div>
This is allowed, MarkBind helpful variables can be used.<br>

`<span id="right_hand">`<code>{<span/>{glyphicon_hand_right}}</code>`</span>`
</div>
<div>
This is also allowed, the second variable will be assigned the contents of the first variable.

`<span id="first">This is the first variable.</span>`<br>
`<span id="second">`<code>{<span/>{first}}</code>`</span>`<br>
`<span id="third">`<code>{<span/>{second}}</code>`</span>`
</div>
<div>
This is forbidden, the fourth variable is not declared yet.

`<span id="third">`<code>{<span/>{fourth}}</code>`</span>`<br>
`<span id="fourth">This is the fourth variable.</span>`
</div>
</tip-box>
```
will be rendered as
variable-tipbox

test/test_site/index.md
```diff
@@ -6,6 +6,12 @@ title: Hello World
 
 <div class="website-content">
 
+{{education_icon}}
+
+{{finalized_value}}
+
+{{reference_level_4}}
+
```
Chng-Zhi-Xuan: Should add a heading to inform what is this section trying to test.

How about `# Test Variables`?

docs/userGuide/contentAuthoring.md
```diff
@@ -161,6 +161,29 @@ Each variable must have an id and the value can be any MarkBind-compliant code f
 
 In any file in your site, you can include the variable by surrounding the variable's id with double curly braces, e.g. <code>{<span></span>{options}}</code>
 
+Variables can refer to other variables that are declared earlier, including helpful variables provided by MarkBind.
+
+<tip-box>
+<div>
+This is allowed, MarkBind helpful variables can be used.<br>
+
+`<span id="right_hand">`<code>{<span/>{glyphicon_hand_right}}</code>`</span>`
```
acjh: - Let's do `<span></span>` as it's not a proper [empty element](https://developer.mozilla.org/en-US/docs/Glossary/Empty_element).
- We can keep it in a single `<code>` element (and avoid the space in browser) by escaping `<`:
  `<code>\<span id="right_hand">{<span></span>{glyphicon_hand_right}}\</span></code>`

docs/userGuide/contentAuthoring.md
```diff
@@ -161,6 +161,29 @@ Each variable must have an id and the value can be any MarkBind-compliant code f
...
+</div>
+<div>
+This is also allowed, the second variable will be assigned the contents of the first variable.
+
+`<span id="first">This is the first variable.</span>`<br>
+`<span id="second">`<code>{<span/>{first}}</code>`</span>`<br>
+`<span id="third">`<code>{<span/>{second}}</code>`</span>`
+</div>
+<div>
+This is forbidden, the fourth variable is not declared yet.
```
acjh: "This is forbidden, `as` the fourth variable is not declared yet."

lib/Site.js
```diff
@@ -295,7 +296,7 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
     const $ = cheerio.load(content);
     $.root().children().each(function () {
       const id = $(this).attr('id');
-      const html = $(this).html();
+      const html = nunjucks.renderString($(this).html(), userDefinedVariables);
```
acjh: Shall we have a comment before this line? It won't be obvious why this was done.

test/test_site/index.md
```diff
@@ -6,6 +6,13 @@ title: Hello World
 
 <div class="website-content">
 
+# Variable that references other variables
```
acjh: More like "Variables that reference another variable"

yamgent:
Changes made.

nicholaschuayunzhi:
> The documentation sample code. Is there a way to escape the {{}} inside a multi-line code?

Im sorry but im having trouble finding this issue in the preview documentation.

yamgent:
> Im sorry but im having trouble finding this issue in the preview documentation.

Both [Zhi Xuan](https://github.com/MarkBind/markbind/pull/231#discussion_r190489617) and [Aaron](https://github.com/MarkBind/markbind/pull/231#discussion_r190503178) have already provided workarounds for this issue. I will update the original post to strike that out.

lib/Site.js
```diff
@@ -295,7 +296,9 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
     const $ = cheerio.load(content);
     $.root().children().each(function () {
       const id = $(this).attr('id');
-      const html = $(this).html();
+      // process the contents of the variable with nunjucks
+      // in case it refers to other variables
```
acjh: Can this be in one line? Also, it should be sentence case (start with a capital letter).

yamgent: Fixed it

nicholaschuayunzhi:
Sorry for the late comment. Would it be possible to add a test similar to https://github.com/MarkBind/markbind/blob/be8e7acc14b2ce503acaebbb5e30f42ca3107c07/test/unit/Site.test.js#L192-L233

for the new behaviour?

yamgent:
> Sorry for the late comment. Would it be possible to add a test similar to

Good idea, will do, thanks.

yamgent:
Unit test added.

docs/userGuide/contentAuthoring.md
```diff
@@ -161,6 +161,29 @@ Each variable must have an id and the value can be any MarkBind-compliant code f
...
+</div>
+<div>
+This is also allowed, the second variable will be assigned the contents of the first variable.
+
+<code>\<span id="first">This is the first variable.\</span></code><br>
+<code>\<span id="second">{<span></span>{first}}\</span></code><br>
+<code>\<span id="third">{<span></span>{second}}\</span></code>
+</div>
+<div>
+This is forbidden, as the fourth variable is not declared yet.
```
nicholaschuayunzhi: Forbidden suggests that the site would not build. Is this case?

test/unit/Site.test.js
```diff
@@ -189,6 +189,32 @@ test('Site read site config for custom site config', async () => {
...
+    'lib/asset/glyphicons.csv': '',
+    'lib/template/page.ejs': PAGE_EJS,
+    'site.json': SITE_JSON_DEFAULT,
+    '_markbind/variables.md':
+    '<span id="level1">variable</span>'
+    + '<span id="level2">{{level1}}</span>'
+    + '<span id="level3">{{level2}}</span>'
+    + '<span id="level4">{{level3}}</span>',
+  };
+  fs.vol.fromJSON(json, '');
```
nicholaschuayunzhi: Could we account for different variables and glyphicons (if possible)? Also could we add sub sites as well?

docs/userGuide/contentAuthoring.md
yamgent: No, the site will continue to build, but the variable will not render properly.

nicholaschuayunzhi: So technically the variable will be set to `undefined`? Maybe we could add on and say that your variables would not render correctly on build?

test/unit/Site.test.js
yamgent: > Could we account for different variables and glyphicons (if possible)?

What do you mean by different variables? Do you mean level4 referring to both level3 and level2?

Will modify the test to include glyphicons.

> Also could we add sub sites as well?

The current implementation of MarkBind does not allow cross-referencing variables from other sub sites, does it? Anyway that seems to be out of scope of the current PR.

nicholaschuayunzhi: > Could we account for different variables and glyphicons (if possible)?

Oh i meant to say adding another word instead of 'variable'. Perhaps changing level3 to a glyphicon?

> The current implementation of MarkBind does not allow cross-referencing variables from other sub sites, does it? Anyway that seems to be out of scope of the current PR.

It doesnt. I was just concerned if there was any unwanted new behaviour was introduced where a subsite variable can reference an outersite. As for adding this new behaviour, i agree this is out of scope.

test/unit/Site.test.js
```diff
@@ -191,14 +191,14 @@ test('Site read site config for custom site config', async () => {
...
+    'lib/asset/glyphicons.csv': 'glyphicon-plus',
     'lib/template/page.ejs': PAGE_EJS,
     'site.json': SITE_JSON_DEFAULT,
     '_markbind/variables.md':
     '<span id="level1">variable</span>'
     + '<span id="level2">{{level1}}</span>'
-    + '<span id="level3">{{level2}}</span>'
-    + '<span id="level4">{{level3}}</span>',
+    + '<span id="level3">{{glyphicon_plus}}</span>'
+    + '<span id="level4">{{level3 | safe}}</span>',
```
yamgent: I don't really like this workaround, but without doing so, nunjucks will escape the angular brackets and the glyphicon code will be render literally:

---

`variables.md`:
```html
<span id="icon_prereq">{{glyphicon_education}}</span>
<span id="icon_ref_prereq_nosafe">{{icon_prereq}}</span>
<span id="icon_ref_prereq_safe">{{icon_prereq | safe}}</span>
```

`index.md`
```
{{icon_prereq}}

This variable didn't use 'safe' when referencing: {{icon_ref_prereq_nosafe}}.

This variable did: {{icon_ref_prereq_safe}}.
```

Results:

230-safe

yamgent: Disabling HTML escape globally (which is the same effect of applying "safe" to all variables) doesn't seem to be a good alternative either.

nicholaschuayunzhi: Perhaps we should document this. 

I was also thinking some people may want to reference glyphicons in other variables and maybe concatenate them like

`<span id="icon_alert_prereq">{{ icon_alert | safe}}{{icon_prereq | safe}}</span>`

This will be fine right?

yamgent: Noted, will document about the flaw.

> ```
> <span id="icon_alert_prereq">{{icon_alert | safe}}{{icon_prereq | safe}}</span>
> ```
>
> This will be fine right?

Yep, that works.

PR #235, opened 2018-05-23
Fix page title prefix get repeated at every live reload

### LINKED ISSUE #233: page title prefix get repeated at every live reload
Steps:
1. specify a title prefix in site.json e.g., `"titlePrefix": "Sandbox"`
2. markbind serve
3. edit some code and save, to trigger a live reload.

Observation: every live reload causes a repetion of the page title prefix. 

image
image

yamgent:

• [X] Bug fix

Fixes #233.

While the website is running live through `markbind serve`, modifying a page causes the page title to re-prepend the title prefix to the title. Therefore, reloading the page three times will cause three title prefixes to appear on the page title for example.

Moved the prefix prepending logic to the `prepareTemplateData()` function so that the title prefix only gets added in at the render stage, rather than at the page (re-)processing stage.

Before | After
---- | ----
title - before | title - after

NIL

**Testing instructions:**
1. Run `markbind serve` on a website with a `titlePrefix` set (e.g. `"titlePrefix": "Prefix"`).
2. Modify the page multiple times (change, save, change, save, ...).
3. The new page should still have one title prefix (e.g. "Prefix - XXXXXX").

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -67,12 +67,16 @@ function unique(array) {
 }
 
 Page.prototype.prepareTemplateData = function () {
+  const prefixedTitle = this.titlePrefix
+    ? this.titlePrefix + (this.title ? TITLE_PREFIX_SEPARATOR + this.title : '')
+    : this.title;
+
   return {
     asset: this.asset,
     baseUrl: this.baseUrl,
     content: this.content,
     faviconUrl: this.faviconUrl,
-    title: this.title,
+    title: prefixedTitle,
   };
 };
 
@@ -114,9 +118,7 @@ Page.prototype.collectFrontMatter = function (includedPage) {
       title: this.title || '',
     };
   }
-  this.title = this.titlePrefix
-    ? this.titlePrefix + (this.frontMatter.title ? TITLE_PREFIX_SEPARATOR + this.frontMatter.title : '')
-    : this.frontMatter.title;
+  this.title = this.frontMatter.title;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Page.js
```diff
@@ -107,22 +111,13 @@ Page.prototype.collectFrontMatter = function (includedPage) {
...
-    }
   } else {
     // Page is addressable but no front matter specified
-    const formattedTitle = this.titlePrefix
-      ? this.titlePrefix + (this.title ? TITLE_PREFIX_SEPARATOR + this.title : '')
-      : this.title;
     this.frontMatter = {
       src: this.src,
-      title: formattedTitle,
+      title: (this.title || ''),
```
acjh: Remove parentheses.

lib/Page.js
```diff
@@ -107,22 +111,13 @@ Page.prototype.collectFrontMatter = function (includedPage) {
...
     };
   }
-  this.title = this.frontMatter.title;
```
acjh: We should sync `this.title`.

lib/Page.js
```diff
@@ -67,12 +67,16 @@ function unique(array) {
 }
 
 Page.prototype.prepareTemplateData = function () {
+  const titleWithPrefix = this.titlePrefix
+    ? this.titlePrefix + (this.frontMatter.title ? TITLE_PREFIX_SEPARATOR + this.frontMatter.title : '')
+    : this.frontMatter.title;
```
acjh: - How about: `const prefixedTitle`
- Use `this.title`.

yamgent:
* Made changes w.r.t. review comments
* Cleaned up the second commit message's language

PR #237, opened 2018-05-23
Docs: Improve description of site.json options

acjh:

• [x] Documentation update

Clarity, conciseness and consistency.

- Rephrased descriptions
- Added link to **contentAuthoring.html#specifying-path-reference** for `baseUrl` option

\-

\-

**Testing instructions:**

\-


docs/userGuide/siteConfiguration.md
```diff
@@ -37,12 +37,12 @@ Let's examine a typical `site.json` file:
...
+| **faviconPath** | The location of the favicon. Default: `favicon.ico`. If the favicon was recently changed, you may need to force-refresh to see the new image. |
+| **titlePrefix** | The prefix for all page titles. |
+| **pages** | An array of pages to be rendered. The `src` is the file; `title` is the page title for the generated web page. Titles specified here take priority over titles specified in the [front matter](contentAuthoring.html#front-matter) of individual pages. Alternatively, `glob` can be used to define a file pattern. If `title` is not specified in the front matter, the page will have `titlePrefix` as its title. |
+| **ignore** | An array of file patterns to be ignored. By default, MarkBind will copy all the files as assets into the output folder. The ignore pattern follows the glob pattern used in [`.gitignore`](https://git-scm.com/docs/gitignore#_pattern_format). The entry `*.md` ignores all markdown source files; you may want to ignore the Git directory `.git/*`. |
```
yamgent: * The link for `.gitignore` wasn't obvious when I read the rendered version (it appears the same as a non-clickable one). Maybe get rid of the code?
* Also maybe we can say "**For example,** the entry `*.md` ..."
* Is there a reason why the user might want to ignore the `.git` directory? (is it for performance reasons?)

docs/userGuide/siteConfiguration.md
```diff
@@ -37,12 +37,12 @@ Let's examine a typical `site.json` file:
...
+| **deploy** | The settings for [auto deployment to Github pages](ghpagesDeployment.html). |
```
yamgent: Could probably mention about the `message` attribute:

"The `message` is the commit message used when deploying to Github pages."

docs/userGuide/siteConfiguration.md
```diff
@@ -37,12 +37,12 @@ Let's examine a typical `site.json` file:
...
-| **deploy** | Settings for the auto Github page deployment. Please refer this [doc](ghpagesDeployment.html) for more details. |
+| **baseUrl** | The base url relative to your domain. Default: <code></code>&nbsp;(empty). For example, if you are using Github Pages to host your deployed website and it is published at `https://markbind.github.io/site-demo-cs2103`, then your `baseUrl` would be `/site-demo-cs2103`. You can use this variable for [specifying path reference](contentAuthoring.html#specifying-path-reference) of images and links. **You may need to change the `baseUrl` when deploying to a different repo.** |
+| **faviconPath** | The location of the favicon. Default: `favicon.ico`. If the favicon was recently changed, you may need to force-refresh to see the new image. |
+| **titlePrefix** | The prefix for all page titles. |
```
yamgent: Should we show an example of how the title prefix will be prepended to the title?

docs/userGuide/siteConfiguration.md
acjh: > - Is there a reason why the user might want to ignore the .git directory? (is it for performance reasons?)

No significant effect on performance, just unnecessary copying.

docs/userGuide/siteConfiguration.md
acjh: "prefix" should be self-descriptive. Added a mention of the separator.

docs/userGuide/siteConfiguration.md
```diff
@@ -37,12 +37,12 @@ Let's examine a typical `site.json` file:
...
+| **baseUrl** | The base url relative to your domain. Default: <code></code>&nbsp;(empty). For example, if you are using Github Pages to host your deployed website and it is published at `https://markbind.github.io/site-demo-cs2103`, then your `baseUrl` would be `/site-demo-cs2103`. You can use this variable for [specifying path reference](contentAuthoring.html#specifying-path-reference) of images and links. **You may need to change the `baseUrl` when deploying to a different repo.** |
+| **faviconPath** | The location of the favicon. Default: `favicon.ico`. If the favicon was recently changed, you may need to force-refresh to see the new image. |
+| **titlePrefix** | The prefix for all page titles. The separator <code>-</code> will be inserted by MarkBind. |
+| **pages** | An array of pages to be rendered. The `src` is the file; `title` is the page title for the generated web page. Titles specified here take priority over titles specified in the [front matter](contentAuthoring.html#front-matter) of individual pages. Alternatively, `glob` can be used to define a file pattern. If `title` is not specified in the front matter, the page will have `titlePrefix` as its title. |
```
yamgent: I feel that this could use a bit of rearrangement (from "Titles specified here" onwards: it talks about titles, then jumps to `glob`, then suddenly jumps back to title).

Something like this:
```
An array of pages to be rendered. The `src` is the file; `title` is the page title for the generated web page. Alternatively, `glob` can be used to define a file pattern. Titles specified here take priority over titles specified in the [front matter](contentAuthoring.html#front-matter) of individual pages. If `title` is not specified in the front matter, the page will have `titlePrefix` as its title. 
```

acjh: I considered that, but `title` is not used with `glob`. And the last line may imply: "If `title` is not specified in the front matter (even if it's in **site.json**), the page will have `titlePrefix` as its title."

Issue #204, opened 2018-04-25
Highlight keyword matches in search results


damithc:
Current: The first match is highlighted. The entire entry is highlighted.
![image](https://user-images.githubusercontent.com/1673303/39242661-66b2bd98-48bd-11e8-91c0-760a6977c843.png)

Suggested: highlight all matches. Highlight only the word matched.

acjh:
Link: https://nus-cs3281.github.io/website/

PR #240, opened 2018-05-23
Support searching for headings

### LINKED ISSUE #162: Support searching for headings
Now that users are able to search for page titles, may be we can also index level 1-3 headings?

nusjzx:

• [x] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Resolves #162 and #204 

highlighting and extracting the search result helps users to identify the results they want easily,
searching for headings and jump to the headings on click helpers users to go to website sections easily.

Follows MarkBind/vue-strap#30 — implementation of highlighting and extracting
while the template is changed in setup.js

jumping to headings: siteData now includes headings also, `collectHeadings` collect h1, h2, h3 headings, and put them in an object where the key is the id , value is the text content. Jumping is done by adding a anchor after the url.

**Testing instructions:**
1. in vue-strap `npm run build`
2. copy paste vue-strap.min.js to markbind
3. test on 3281 website


acjh:
Can you add some screenshots?

yamgent:
Missing step for testing instruction: After cloning the CS3281 website, go to `common/header.md`, and replace the use of `<typeahead>` with `<searchbar>`:

```html
        <searchbar :data="searchData" placeholder="Search" :template="titleTemplate" template-name="title" :on-hit="searchCallback"></searchbar>
```

nusjzx:
Screenshot:
image

image

acjh:
Is that generating multiple entries for a single page? Doesn't seem right.
Also, are long headings truncated?

nusjzx:
long headings are not truncated, as we have each heading per entry

asset/js/setup.js
```diff
@@ -9,29 +9,26 @@ function setup() {
...
-                      + '<span v-if="item.headings" v-html="item.headings | highlight value"></span>',
+            + '<span v-if="item.keywords" v-html="item.keywords | highlightAndExtract value<br />"></span>'
+            + '<span v-if="item.heading" v-html="item.heading.value | highlight value"></span>',
       };
     },
     methods: {
       searchCallback(match) {
-        window.location = `${baseUrl}/${match.src.replace('.md', '.html')}`;
+        const page = `${baseUrl}/${match.src.replace('.md', '.html')}`;
+        window.location = Object.keys(match.heading).length === 0 ? page : `${page}#${match.heading.id}`;
```
yamgent: How about:

```js
const page = `${baseUrl}/${match.src.replace('.md', '.html')}`;
const anchor = (Object.keys(match.heading).length === 0) ? '' : `#${match.heading.id}`;
window.location = `${page}${anchor}`;
```

nusjzx:
prof just checked the effects, this is what he suggested

asset/js/setup.js
```diff
@@ -9,27 +9,26 @@ function setup() {
   VueStrap.installEvents(vm);
 }
 
+
```
yamgent: Extra blank line

lib/Page.js
```diff
@@ -136,6 +137,14 @@ Page.prototype.removeFrontMatter = function (includedPage) {
   return $.html();
 };
 
+Page.prototype.collectHeadings = function (includedPage) {
+  const $ = cheerio.load(includedPage);
+  $('h1,h2,h3').map((i, heading) => {
```
yamgent: This should use `forEach` rather than `map` so that we can avoid the `return 0;`

yamgent: Sorry, the correct name should be `each()` for this case.

damithc:
Noticed that in your staged version the drop-down of the nav bar is too long.
image

nusjzx:
@damithc  noted

yamgent:
This PR needs to be rebased to the latest master, there's a conflicting file.

danielbrzn:
Looked through the PR and the live demo available at https://nusjzx.github.io/website-base/.

It's a good start to making the search more functional and user-friendly, especially given that important headings might be hidden behind panels that the user has to click in order to see.

Some comments:

- Perhaps bolding the matches between the user's query and results might be a better experience for the user instead of highlighting the matches. It might seem a bit jarring to users.

- There's a noticeable delay between the user typing a query, and having the query + results show up, which doesn't make for a very good user experience. You can observe how the results are instantly served [here](https://nus-cs3281.github.io/website). The delay could be due to the highlighting of the match, or because the data source is large (948 pages in `siteData.json`).

yamgent:
> There's a noticeable delay between the user typing a query, and having the query + results show up, which doesn't make for a very good user experience. You can observe how the results are instantly served here. The delay could be due to the highlighting of the match, or because the data source is large (948 pages in siteData.json).

I highly suspect it is mainly due to the large data source. If the highlighting of the match is a problem then it also means we are also rendering too many items in the list.

Two things we could possibly do to make this less of an issue:
* Load the data in the background, and show a loading spinner in the UI while it is loading. While it doesn't reduce the delay, it gives a visible feedback to the user of what is going on.
* When processing the list items one by one, if an item is matched already, then it can be shown on the screen first while we process the rest of the items (rather than the current implementation where we have to process *everything* before we show everything in one go).

damithc:
Performance improvements can be done as separate PRs if appropriate.

acjh:
> implementation of highlighting and extracting is in MarkBind/vue-strap#28

Please change this to:

> Follows MarkBind/vue-strap#30 — implementation of highlighting and extracting

And update this PR where required.

lib/Page.js
```diff
@@ -132,6 +133,13 @@ Page.prototype.removeFrontMatter = function (includedPage) {
   return $.html();
 };
 
+Page.prototype.collectHeadings = function (includedPage) {
+  const $ = cheerio.load(includedPage);
+  $('h1,h2,h3').each((i, heading) => {
+    this.headings[$(heading).attr('id')] = $(heading).text();
+  });
+};
```
acjh: - Move this above `collectIncludedFiles`.
- Add jsdoc.

acjh:
@yamgent The changes in **setup.js** make this PR a breaking change, to be noted in the next release.

test/test_site/expected/siteData.json
```diff
@@ -1,14 +1,33 @@
...
+        "normal-include": "Normal include",
+        "establishing-requirements": "Establishing Requirements",
+        "include-segment": "Include segment",
+        "dynamic-include": "Dynamic include",
+        "boilerplate-include": "Boilerplate include",
+        "nested-include": "Nested include",
+        "html-include": "HTML include",
+        "include-from-another-markbind-site": "Include from another Markbind site",
+        "feature-list": "Feature list"
+      },
```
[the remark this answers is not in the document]
yamgent: > * This isn't used? I'm surprised that this test isn't failing:

What do you mean by "This isn't used"?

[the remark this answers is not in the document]
acjh: This isn't used by test_site (and appeared out of nowhere).

`$ npm run test` fails locally.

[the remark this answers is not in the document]
yamgent: > This isn't used by test_site (and appeared out of nowhere).

While test_site doesn't use `searchbar`, this data is still generated regardless of whether `searchbar` is used or not.

So the behavior is intended, and the headings do match the contents of the test_site.

> `$ npm run test` fails locally.

I cannot reproduce the test failing locally on Windows.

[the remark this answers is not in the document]
acjh: My bad.

In that case, let's add a searchbar to **docs/common/header.md** instead.

lib/Page.js
```diff
@@ -81,6 +81,18 @@ Page.prototype.prepareTemplateData = function () {
   };
 };
 
+/**
+ * Records all h1-h3 headings id-text pair into this.headings
```
yamgent: Records all h1 to h3 headings ~~id-text pair~~ into this.headings

Mentioning id-text pair would reveal too much details imo.

test/test_site/expected/siteData.json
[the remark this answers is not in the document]
yamgent: @nusjzx Reminder to add a searchbar for our MarkBind documentation website too.

lib/Page.js
```diff
@@ -80,6 +81,18 @@ Page.prototype.prepareTemplateData = function () {
   };
 };
 
+/**
+ * Records all h1-h3 headings into this.headings
+ * @param includedPage a page with its dependencies included
+ */
+Page.prototype.collectHeadings = function (includedPage) {
```
acjh: Should be `renderedPage`.

```js
/**
 * Records h1,h2,h3 headings into this.headings
 * @param renderedPage a page with its headings rendered
 */
Page.prototype.collectHeadings = function (renderedPage) {
```

test/test_site/expected/siteData.json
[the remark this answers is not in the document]
acjh: @nusjzx Let's add it to **docs/** instead of **test_site/**.

yamgent:
@danielbrzn Mind taking a look at this again, specifically the code? Thanks! :)

lib/Page.js
```diff
@@ -80,6 +81,18 @@ Page.prototype.prepareTemplateData = function () {
   };
 };
 
+/**
+ * Records all h1-h3 headings into this.headings
```
acjh: @nusjzx Note the change in the first line, from https://github.com/MarkBind/markbind/pull/240#discussion_r192317692:

> ```js
> /**
>  * Records h1,h2,h3 headings into this.headings
> ```

nusjzx: noted

damithc:
> A handful of headings (e.g. Glyphicons) in MarkBind docs use h4, which can't be searched.
> Should we "upgrade" those? The current headings look pleasant.

Alternatively, we can include h4 in search index.

acjh:
h4 seems noisy:

screenshot 2018-06-01 17 29 52

yamgent:
> A handful of headings (e.g. Glyphicons) in MarkBind docs use h4, which can't be searched.
> Should we "upgrade" those? The current headings look pleasant.

I think it is good to "upgrade" them.

If the new headings doesn't look pleasant, then I think the root of the issue is that the style for different headings are not distinct enough from each other, and we should fix that rather than skipping headings.

yamgent:
> @yamgent The changes in setup.js make this PR a breaking change, to be noted in the next release.

@acjh if it is a breaking change, should that make the version become 2.0.0? According to the [semantic versioning scheme](https://semver.org/), since our new **setup.js** in this PR breaks existing websites if the authors do not make any changes to their HTML code on their navbar, it qualifies as not having a backwards-compatible feature.

acjh:
A minor version will do. 1.x is still not production-ready for now.
We can follow those guidelines when we are ready for 2.0.

PR #327, opened 2018-07-11
Update markbind and site-nav CSS to work well in Bootstrap 4

### LINKED ISSUE #325: Migration: Fix "component references" page covering up site navigation

* **MarkBind Version:** None (currently a problem in `bootstrap-migration` branch).

Serve the MarkBind's documentation, and go to the "Component References" page.

The layout width of the page content seems to be too big when the window size is smaller, which allows the user to scroll further to the right and cover up the site nagivation:

doesnotscale

Chng-Zhi-Xuan:

• [X] Bug fix

Fixes #325 

Rendering of certain components inside documentation was buggy due to new CSS in Bootstrap 4.

- Override `<code>` tag's CSS to not break character from words
- Improve looks and rendering of `site-nav` elements
- Modified elements in `test_site` to look correct and have more meaningful filler text

Before |
--- |
doesnotscale |

After |
--- |
markbind-css-after |

- Width of `#page-content` before responsive media query kicks in

**Testing instructions:**
- Ensure you have `vue-strap.min.js` built from [vue-strap migration branch](https://github.com/MarkBind/vue-strap/tree/bootstrap-migration).
- Run `markbind serve docs`, check if elements render properly in both desktop / responsive mode.
- Run `markbind serve` in `test_site` and check there as well.

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -31,7 +31,8 @@ code {
     background-color: #fff;
 }
 
-@media (max-width: 767px) {
+/* Bootstrap small(sm) responsive breakpoint */
+@media (max-width: 767.98px) {
     .dropdown-menu > li > a {
         white-space: normal;
     }
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -2,6 +2,10 @@ blockquote {
     font-size: 14px;
 }
 
+code {
+    word-break: normal;
+}
+
 .btn:active,
 .btn:focus {
     box-shadow: none !important;
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -6,10 +6,6 @@ code {
     word-break: normal;
 }
 
-pre {
-    margin-bottom: 0;
-}
-
 .btn:active,
 .btn:focus {
     box-shadow: none !important;
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -6,6 +6,10 @@ code {
     word-break: normal;
 }
 
+pre {
+    margin-bottom: 0;
+}
+
 .btn:active,
 .btn:focus {
     box-shadow: none !important;
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -8,8 +8,9 @@
 #page-content {
     flex: 1;
     margin-left: 285px;
+    overflow-x: hidden;
     padding-left: 20px;
-    padding-right: 20px;
+    padding-right: 0;
     transition: 0.4s;
     -webkit-transition: 0.4s;
 }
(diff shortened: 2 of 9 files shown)


asset/css/markbind.css
```diff
@@ -6,6 +6,10 @@ code {
     word-break: normal;
 }
 
+pre {
+    margin-bottom: 0;
+}
```
yamgent: Is there a reason why we are doing this? It is removing line spacings across many different areas of the documentation, and the missing line spacing is making the documentation rather hard to read. An example:

asset/css/site-nav.css
```diff
@@ -140,7 +143,7 @@
 
 /* Responsive site navigation */
 
-@media screen and (max-width: 767px) {
+@media screen and (max-width: 1006px) {
```
yamgent: I am fine with `1006px`, although personally I think it doesn't look like a nice number compared to like `999px`. :P

asset/css/site-nav.css
```diff
@@ -8,8 +8,9 @@
 #page-content {
     flex: 1;
     margin-left: 285px;
+    overflow-x: hidden;
```
yamgent: One of the disadvantage is that if your screen width is really small (e.g. 360px), no scroll bar will be shown, the content will be cut off and the user cannot see them. For example, this table's "Description" column cannot be seen at all: 



In our current MarkBind website, the scroll bar will be shown and user can scroll right:

asset/css/site-nav.css
```diff
@@ -8,8 +8,9 @@
...
     padding-left: 20px;
```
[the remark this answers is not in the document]
Chng-Zhi-Xuan: Then the page contents will touch the divider line between itself and the site navigation. 

It is not obvious in our documentation since `website-content` class has additional margins to space things out. 

I removed `padding-right` as there is a built in padding from the app container already which helps us avoid being too close to the scroll bar.

asset/css/site-nav.css
Chng-Zhi-Xuan: The old media query had the `#page-content` div squished to 416 pixels width when the window width was 786 pixels. This was too narrow to read and triggered a much more dramatic transition effect.

So I took the maximum width of Panel in responsive mode (window at 767 pixels or lower), which was 655 pixels and subtracted the squished width of 400+ pixels. The result was 239 pixels.

239 + 767 = 1006 😃. So now the `#page-content` div will have a minimum of 655 pixels of width  when the window width is above 767 pixels.

acjh: Hmm, that should be in a code comment. In general, stick to Bootstrap's [responsive breakpoints](https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints).

Issue #335, opened 2018-07-13
Improve MarkBind's responsive / mobile usability


Chng-Zhi-Xuan:
* **MarkBind Version:** 1.9.0

Table below will be updated as more issues are found.
(Tested in Chrome's responsive mode for iPhone 6/7/8)

Areas of improvement in user guide |
--- |
![responsive-table](https://user-images.githubusercontent.com/31084833/42675822-4837a79c-86a8-11e8-86f5-37e7bafbde27.PNG)<br><h2>1. Tables are not horizontally scroll-able.</h2> |
![responsive-panel-button](https://user-images.githubusercontent.com/31084833/42675848-6086e330-86a8-11e8-865f-289f0c0c7c5d.PNG)<br><h2>2. Panel buttons are misaligned. The button group also takes up too much width, limiting length of the header.</h2> |
![responsive-font](https://user-images.githubusercontent.com/31084833/42675885-7198a5a0-86a8-11e8-9b81-fe8e643ac38b.PNG)<br><h2>3. Font size for headers need better scaling<br>(H4 currently too large).</h2> |
![responsive-image](https://user-images.githubusercontent.com/31084833/42675924-96592860-86a8-11e8-922a-06abbda26f84.PNG)<br><h2>4. Large images are cut off, should have scaled down in size<br> or have it scroll-able.</h2> |
![responsive-popover](https://user-images.githubusercontent.com/31084833/42676562-3c02745e-86ab-11e8-80f0-b4af7948e095.PNG)<br><h2>5. Popover with image is cut off</h2> |
![responsive-footer](https://user-images.githubusercontent.com/31084833/42675962-bb5f7178-86a8-11e8-9fe9-da9cb386812c.PNG)<br><h2>6. Format our footer so it looks good on mobile.</h2> |
![responsive-minimised-btn](https://user-images.githubusercontent.com/31084833/42803881-e95b5ce8-89d9-11e8-8f80-4d3b0d3cfd02.PNG)<br><h2>7. Minimised button text does not wrap when overflowed.</h2> |
<h2>8. Site navigation bar on small devices (576px width or smaller) takes up too much space, we should hide it. User can still access it by "Request for desktop site" on their mobile browsers.</h2>

**Fix status**
- [x] Area 1 - #349
- [x] Area 2 - MarkBind/vue-strap#75
- [x] Area 3 - #350
- [x] Area 4 - MarkBind/vue-strap#76 and #354 
- [x] Area 5 - #357
[...]

PR #350, opened 2018-07-16
Refactor long <h4> in user guide to bold text

Chng-Zhi-Xuan:

• [x] Documentation update

Part of #335 

Heading font sizes on mobile site was too large.

- Change long instances of `<h4>` into bold markdown text using `**`

Before | After 
--- | ---
335-font-size-before | 335-font-size-afterv2

- Readability of new format on mobile.

**Testing instructions:**
- Use `Netlify Preview` and reduce your window size.
- Go through the document and check the readability of the content.

### DIFF
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -2,9 +2,7 @@
 
 ## Rich formatted text in headings and titles
 
-<p style="font-size:18px;">
-  Using the normal syntax, you are only able to use markdown formatting on headings. If you would like more styling options, you can define an element within the component that acts as your heading. This is done by adding a <md>`slot`</md> attribute with the correct name to that element.
-</p>
+Using the normal syntax, you are only able to use markdown formatting on headings. If you would like more styling options, you can define an element within the component that acts as your heading. This is done by adding a <md>`slot`</md> attribute with the correct name to that element.
 
 <tip-box border-left-color="#00B0F0">
   <i style="font-style: normal; font-weight: bold; color: dimgray">Example</i><br>
--- a/docs/userGuide/components/modal.md
+++ b/docs/userGuide/components/modal.md
@@ -1,6 +1,6 @@
 # Modal
 
-#### Modals are to be used together with the [Trigger](#trigger) component for activation.
+**Modals are to be used together with the [Trigger](#trigger) component for activation.**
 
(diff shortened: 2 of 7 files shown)


yamgent:
Ok. However, I took a look at this with regards to `website-base`, and didn't think that much space was saved with the reduction in heading. And for h4 and below (see second row and third row below), we can no longer differentiate between the headings and the normal text because the sizes are almost the same.

To me, I think both "Before" and "After" are acceptable, but this is probably down to personal taste. @acjh what do you think about it?

Before | After
-- | --
 | 
 | 
 | 

Url of the pages shown above:
https://nus-cs2103.github.io/website-base/admin/programming-languages.html
https://nus-cs2103.github.io/website-base/se-book-adapted/chapters/ides.html

asset/css/markbind.css
```diff
@@ -33,6 +33,32 @@ code {
 
 /* Bootstrap small(sm) responsive breakpoint */
 @media (max-width: 767.98px) {
+
+    /* 70% of Bootstrap 4 default font-size for headings */
+    h1 {
```
yamgent: Bootstrap also overrides classes of `h1` (i.e. it does `h1, .h1 {`). Should we consider doing the same?

docs/userGuide/components/modal.md
```diff
@@ -1,6 +1,6 @@
 # Modal
 
-#### Modals are to be used together with the [Trigger](#trigger) component for activation.
+<p><strong>Modals are to be used together with the <md>[Trigger](#trigger)</md> component for activation.</strong></p>
```
yamgent: Can just use standard Markdown syntax, it renders the same way:

```
**Modals are to be used together with the [Trigger](#trigger) component for activation.**
```

Chng-Zhi-Xuan:
Bold text was intended to substitute `<h4>`. Long blocks of text should remain as is for easier readability (especially on mobile). 

Will just have that paragraph use the default responsive font size from Bootstrap.

docs/userGuide/components/advanced.md
```diff
@@ -2,7 +2,7 @@
 
 ## Rich formatted text in headings and titles
 
-<p style="font-size:18px;">
+<p>
```
yamgent: The paragraph tag can just be removed then.

Issue #336, opened 2018-07-14
Level 4 headings not showing up in search results


damithc:
v1.9.0
context: https://nus-cs2103.github.io/website-base/
I've set search indexing to level 4 https://github.com/nus-cs2103/website-base/blob/master/site.json

In the case below, the [level 4 heading `Objects as Abstractions`](https://nus-cs2103.github.io/website-base/se-book-adapted/chapters/oop.html#objects-as-abstractions) did not show up in results.
![image](https://user-images.githubusercontent.com/1673303/42724150-8f8d7e1c-879f-11e8-81d4-bed640a1d32e.png)

Furthermore, some odd entries with `%` show up in results.

yamgent:
> In the case below, the level 4 heading Objects as Abstractions did not show up in results.

One possible reason could be that because the content is contained inside a panel, and the panel itself is dynamic, so the content wasn't loaded in when the headings were scanned, and we didn't pick it up as a result.

damithc:
> One possible reason could be that because the content is contained inside a panel, and the panel itself is dynamic, so the content wasn't loaded in when the headings were scanned, and we didn't pick it up as a result.

hmm... I thought index building is not dependent on page loading. Isn't the index built even before a single page is loaded?

yamgent:
> Isn't the index built even before a single page is loaded?

Yes, that is correct. What I meant is that, the index is built with reference to the html pages generated by MarkBind during a `markbind build`, and if a panel is dynamic, then the headings won't be visible on the html source.

damithc:
I see. I guess this affects nested panels of 2nd level and below (as the heading is specified in the panel text).
But these are expanded panels, which means they are effectively non-dynamic?

Chng-Zhi-Xuan:
Expanded Panels are still dynamic. 

The HTML will be removed when you close them. The difference is that since `expanded = true`, the HTML contents will be there when loaded and the algorithm in `setup.js` will index the headings. 

The reason it only affects 2nd level and below is because the first level's panel header will always be rendered. Since the 2nd level and below is contained within the 1st level's body, it is dynamically rendered when opened.

nusjzx:
cannot reproduce the %% problem. https://nusjzx.github.io/website-base/

yamgent:
Regarding the `%%` problem, can't reproduce this problem on a blank-slate website either, since MarkBind resolves the formatting before we start building the search index, so the raw `%%` syntax should be long gone by then.

`index.md`:
[...]

PR #351, opened 2018-07-17
Search panel headings

nusjzx:

• [ ] Documentation update
• [X] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Relates #336 

1. Headings in dynamic panels are not included in search result.
2. panels not expanded should not be searchable.

collect headings from panels which are expanded.

**Testing instructions:**
1.pull https://github.com/MarkBind/vue-strap/pull/80
2. run `npm run build` in vue-strap repo
3. copy vue-strap.min.js to markbind repo
4. test on 2103 website.
eg: "pros": the pros and cons entry in the software engineer section are not indexed before(It's a panel header)
      "abstraction" : `object as abstraction` entry is there same as above.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -165,21 +165,53 @@ Page.prototype.prepareTemplateData = function () {
 };
 
 /**
- * Records h1,h2,h3 headings into this.headings
- * @param renderedPage a page with its headings rendered
+ * Records headings inside rendered page into this.headings
  */
-Page.prototype.collectHeadings = function (renderedPage) {
-  const $ = cheerio.load(renderedPage);
+Page.prototype.collectHeadings = function () {
+  this.collectHeadingsInContent(fs.readFileSync(this.resultPath));
+};
+
+/**
+ * Records headings inside content into this.headings
+ * @param content that contains the headings
+ */
+Page.prototype.collectHeadingsInContent = function (content) {
+  let $ = cheerio.load(content);
+  $('panel').each((index, panel) => {
+    if (panel.attribs.header) {
+      this.collectHeadingsInContent(md.render(panel.attribs.header));
+    }
+  }).filter((index, panel) => panel.attribs.expanded !== undefined)
+    .each((index, panel) => {
+      if (panel.attribs.src) {
+        const src = panel.attribs.src.split('#')[0];
+        const includePath = path.resolve(this.resultPath,
+                                         path.relative(this.sourcePath,
+                                                       this.baseUrl
+                                                         ? path.relative(this.baseUrl, src)
+                                                         : src.substring(1)));
+        const includeContent = fs.readFileSync(includePath);
+        if (panel.attribs.fragment) {
+          $ = cheerio.load(includeContent);
+          this.collectHeadingsInContent($(`#${panel.attribs.fragment}`).html());
+        } else {
+          this.collectHeadingsInContent(includeContent);
+        }
+      } else {
+        this.collectHeadingsInContent($(panel).html());
+      }
+    });
+  $ = cheerio.load(content);
   if (this.headingIndexingLevel > 0) {
     let headingsSelector = 'h1';
     for (let i = 2; i <= this.headingIndexingLevel; i += 1) {
       headingsSelector += `, h${i}`;
     }
+    $('panel').remove();
     $(headingsSelector).each((i, heading) => {
       this.headings[$(heading).attr('id')] = $(heading).text();
     });
   }
-  return renderedPage;
 };
 
 /**
@@ -317,7 +349,6 @@ Page.prototype.generate = function (builtFiles) {
       .then(result => markbinder.resolveBaseUrl(result, fileConfig))
       .then(result => fs.outputFileAsync(this.tempPath, result))
       .then(() => markbinder.renderFile(this.tempPath, fileConfig))
-      .then(result => this.collectHeadings(result))
       .then((result) => {
         this.content = htmlBeautify(result, { indent_size: 2 });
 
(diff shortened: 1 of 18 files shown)


yamgent:
Currently there's a mismatch in behavior between this PR and MarkBind/vue-strap#80, will wait until MarkBind/vue-strap#80 is fully finished before reviewing resumes.

yamgent:
I can't search for the heading "Bye" with the markdown below.

Also, it seems that `# Hi` is only searchable when `expanded` is specified, but since the heading is always visible, it should always be searchable regardless of whether it is expanded or not?

`index.md`:

```markdown

<searchbar :data="searchData" placeholder="Search" :on-hit="searchCallback"></searchbar>

<panel header="# Hi" expanded>
  <markdown>
  ## Bye
  Bye
  </markdown>
</panel>
```

lib/Site.js
```diff
@@ -522,6 +522,11 @@ Site.prototype.generatePages = function () {
   });
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
+      .then(() => {
+        this.pages.forEach((page) => {
+          page.collectHeadings(fs.readFileSync(page.resultPath));
```
yamgent: The test fails because `page` is a mock object rather than the an actual `Page` object, so `page.resultPath` is undefined (class attributes do not get mock).

Perhaps redesign the method such that it is unnecessary for Site.js to call `fs.readFileSync(page.resultPath)`?

lib/Site.js
```diff
@@ -522,6 +522,12 @@ Site.prototype.generatePages = function () {
   });
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
+      .then((res) => {
```
yamgent: The `res` doesn't seem to be necessary (the result is an undefined array, and is not used by subsequent code), so it can be removed (just `.then(() => {`)

lib/Page.js
```diff
@@ -165,21 +165,50 @@ Page.prototype.prepareTemplateData = function () {
...
+  $('panel').each((index, panel) => {
+    if (panel.attribs.header) {
+      this.collectHeadingsInContent(md.render(panel.attribs.header));
+    }
+  }).filter((index, panel) => panel.attribs.expanded !== undefined)
+    .each((index, panel) => {
+      if (panel.attribs.src) {
+        const includePath = path.resolve(this.resultPath,
+                                         path.relative(this.sourcePath,
+                                                       path.relative(this.baseUrl, panel.attribs.src)));
```
yamgent: There seems to be some implementation error computing the include path. I get this error:

Console:
```
warning: Error: ENOENT: no such file or directory, open 'C:\Users\other._include_.html'
warning: Error: ENOENT: no such file or directory, open 'C:\Users\other._include_.html'
error: ENOENT: no such file or directory, open 'C:\Users\other._include_.html'
```

`index.md`:
```html
<panel header="Included panel" src="other.md" expanded>
</panel>
```

`other.md`:
```markdown
# Other Heading

Other heading
```

lib/Page.js
```diff
@@ -165,21 +165,52 @@ Page.prototype.prepareTemplateData = function () {
...
  */
-Page.prototype.collectHeadings = function (renderedPage) {
-  const $ = cheerio.load(renderedPage);
+Page.prototype.collectHeadings = function () {
+  this.collectHeadingsInContent(fs.readFileSync(this.resultPath));
+};
+
+/**
+ * Records headings inside content into this.headings
+ * @param content that needs to collect headings
```
yamgent: ` * @param content that contains the headings`

lib/Page.js
```diff
@@ -165,21 +165,52 @@ Page.prototype.prepareTemplateData = function () {
 };
 
 /**
- * Records h1,h2,h3 headings into this.headings
- * @param renderedPage a page with its headings rendered
+ * Records headings in rendered page into this.headings
```
yamgent: To be consistent with the below method's comment:

` * Records headings inside rendered page into this.headings`

lib/markbind/lib/parser.js
```diff
@@ -284,10 +284,10 @@ Parser.prototype._parse = function (node, context, config) {
                     || utils.fileExists(calculateBoilerplateFilePath(element.attribs.boilerplate,
                                                                      element.attribs.src, config));
     if (fileExists) {
-      const { src, fragment } = element.attribs;
+      const { src } = element.attribs;
       const resultDir = path.dirname(path.join('{{hostBaseUrl}}', path.relative(config.rootPath, src)));
       const resultPath = path.join(resultDir, utils.setExtension(path.basename(src), '._include_.html'));
-      element.attribs.src = fragment ? `${resultPath}#${fragment}` : resultPath;
+      element.attribs.src = resultPath;
```
yamgent: There is a bug in the rendering of panels with fragments after this line of code is changed. This suggests that the fragment is necessary after all:

yamgent: @nusjzx Btw this bug still exist, I presume you are not done with the PR, right? :P

test/test_site/index.md
```diff
@@ -38,4 +38,33 @@ siteNav: site-nav.md
 # Include from another Markbind site
 <include src="sub_site/index.md" />
 
+# Panel without src
+<panel header="## Panel without src" expanded>
+<markdown>
+### panel without src content heading
+<markdown>
```
yamgent: Is this supposed to be the closing tag? `</markdown>`

test/test_site/index.md
```diff
@@ -38,4 +38,33 @@ siteNav: site-nav.md
 # Include from another Markbind site
 <include src="sub_site/index.md" />
 
+# Panel without src
+<panel header="## Panel without src" expanded>
```
yamgent: The heading inside the panel is the same as the heading outside the panel. Maybe you can omit the heading outside the panel if it is the same as the one inside the panel?

This may be applicable for the ones below too.

PR #356, opened 2018-07-18
Hide site navigation bar when on mobile devices

Chng-Zhi-Xuan:

Part of #335 

Site navigation toggle button on mobile device takes up too much screen space.

- Add CSS to hide site navigation at Bootstrap XS responsive breakpoint.
- Update site navigation section in the user guide.

Before | After
--- | ---
335-site-nav-before | 335-site-nav-after

- Transition between medium (992px) and extra small (576px) breakpoints

**Testing instructions:**
- Use `Netlify Preview` and use responsive mode in chrome to test.

### DIFF
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -185,3 +185,36 @@
         margin-left: 300px;
     }
 }
+
+/* Bootstrap extra small(xs) responsive breakpoint */
+@media screen and (max-width: 575.98px) {
+
+    #site-nav {
+        padding-right: 0;
+    }
+
+    #site-nav-btn-wrap {
+        padding: 0;
+        width: 0;
+    }
+
+    #page-content {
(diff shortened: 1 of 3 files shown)


asset/css/site-nav.css
```diff
@@ -185,3 +185,36 @@
...
+        padding: 0;
+    }
+
+    footer {
+        margin-left: -15px;
+    }
+
+    /* Site navigation open button toggle */
+
+    #site-nav.open {
```
yamgent: Do we need these? The site navigation is not supposed to be open-able anyway right?

Chng-Zhi-Xuan: Edge case where a user can open the site nav and reduce the window size 😰

yamgent: How to trigger this edge case?

Issue #377, opened 2018-07-26
Specify location of content inserted into <head> of a page


Chng-Zhi-Xuan:
This is from a discussion continuation in #355 

**Existing behaviour**

Content could only be inserted at the bottom of a page's `<head>` tag

**Changes requested**

User would like more control on which contents are inserted at the top or bottom of a page's `<head>` tag.

PR #378, opened 2018-07-26
Allow user to control position of content to be inserted in <head>

Chng-Zhi-Xuan:

#377 

User would like to control the position of content placed in `<head>`

- Introduce `<head-top>` and `<head-bottom>` tags as position specifiers.
- Changed `collectHeadFiles` method to account for position logic  and be more streamlined. 
- Update relevant section in user guide
- Updated insertion test in `test_site`

Page source of test_site |
--- |
377-after |

- Wording in user guide

**Testing instructions:**
- Checkout this PR and navigate to the `test_site` folder.
- Run `markbind serve` and check console for success messages.
- Check page source to see correct insertion.

### DIFF
--- a/docs/userGuide/contentAuthoring.md
+++ b/docs/userGuide/contentAuthoring.md
@@ -385,6 +385,10 @@ While authoring your website, you may want to have your own CSS or Javascript fi
 - Author your `<style>` elements for CSS files and `<link>` elements for Javascript files using HTML as shown below.
     - Ensure that your URLs start from the root directory, by using <code>{<span></span>{baseUrl}}/</code> when you are referencing your files.
 
+- All content is inserted near the bottom of the `<head>` tag by default.
+    - Use the `<head-top>` tag to specify content that you wish to place at the top of the `<head>` tag.
+    - The `<head-top>` tag is optional, you may omit it from your head file if you do not use it. 
+
 ```css
 /* In yourCSSFolder/subfolder/myCustomStyle.css */
 p {
@@ -399,8 +403,11 @@ alert("Welcome to my website!");
 
 ```html
 <!-- In _markbind/head/compiledRef.md -->
+<head-top>  <!-- HTML within this tag will be placed first, directly after the starting <head> tag -->
+  <script src="{{baseUrl}}/yourScriptFolder/myCustomScript.js"></script>
+</head-top>
+<!-- Any content outside the <head-top> tag will default to the bottom position -->
 <link rel="stylesheet" href="{{baseUrl}}/yourCSSFolder/subfolder/myCustomStyle.css">
-<script src="{{baseUrl}}/yourScriptFolder/myCustomScript.js"></script>
 ```
 
 - Specify the head file in pages that uses it, by specifying the [front matter](#front-matter) `head` attribute.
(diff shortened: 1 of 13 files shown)


docs/userGuide/contentAuthoring.md
```diff
@@ -385,6 +385,10 @@ While authoring your website, you may want to have your own CSS or Javascript fi
 - Author your `<style>` elements for CSS files and `<link>` elements for Javascript files using HTML as shown below.
     - Ensure that your URLs start from the root directory, by using <code>{<span></span>{baseUrl}}/</code> when you are referencing your files.
 
+- Use `<head-top>` and `<head-bottom>` tags to control where your content will be placed.
+  - You must enclose your content in at least one tag for it to be placed in the `<head>`.
+  - If the other tag is not needed, you may omit it in your head file.
```
yamgent: `  - It is not mandatory to use both tags, you may omit either one of them in your head file if it is unused.`

docs/userGuide/contentAuthoring.md
```diff
@@ -385,6 +385,10 @@ While authoring your website, you may want to have your own CSS or Javascript fi
 - Author your `<style>` elements for CSS files and `<link>` elements for Javascript files using HTML as shown below.
     - Ensure that your URLs start from the root directory, by using <code>{<span></span>{baseUrl}}/</code> when you are referencing your files.
 
+- Use `<head-top>` and `<head-bottom>` tags to control where your content will be placed.
```
yamgent: "to control where your content will be placed" may be quite vague, so we can be more explicit here.

`... tags to insert the content at the top and bottom of <head> respectively.`

docs/userGuide/contentAuthoring.md
damithc: Can we avoid this extra work to the author? i.e., normally added to the bottom but if these tags are present, they will be honored.

yamgent: Might get a bit confusing though if the user only uses `<head-bottom>` but not `<head-top>`.

```html
<script src="appearsToBeInsertedAtTop-butIsInsertedAtBottom.js" />

<head-bottom>
<script src="alwaysAtBottom.js" />
</head-bottom>
```

damithc: May be we don't need the bottom tag at all?

Chng-Zhi-Xuan: **Messy example**
```html
<script src="blahA"></script>
<!-- random comment -->
<script src="blahB"></script>
<head-top>
  <script src="kingOfScripts"></script>
</head-top>
<!-- random comment 2 -->
<script src="blahC"></script> 
```

Implementing just the `<head-top>` tag seems unbalanced, as the user can insert it anywhere they want :/

damithc: **Common things should be _easy_ to do. Uncommon things should be _possible_ to do.** The common thing is adding to the bottom, which should be possible with the least effort.

Our instructions can be _Normally added to the bottom. If you want to add to the top, enclose them in a `<top>` tag_.

In the example we give, we can show the neat way to do it so that users do the same. But it's ok even if the top tag is in the middle. We just pick it up from the middle and add the rest to the bottom.

acjh:
> I am OK with `<head-top>` and `<head-bottom>`, but I am not sure whether having hyphens in a HTML tag name is allowed

Apparently, it's **required**.
https://stackoverflow.com/questions/22545621/do-custom-elements-require-a-dash-in-their-name

Chng-Zhi-Xuan:
**Update**
- Deprecate `<head-bottom>` tag, default content to be inserted at the bottom.
- Add regex removal of excessive line breaks
- Update test file

Chng-Zhi-Xuan:
**Update**
- Squashed commits

PR #394, opened 2018-07-30
Remove container fluid class from #app and add page content wrapper for styling

### LINKED ISSUE #390: Move footer outside "#app" div

Due to how `container-fluid` works, everything that is placed inside it will have a padding added on its left and right side. If we did not intervene with this css style, the footer will look like this (notice the white space at the left and right of footer):

We resolved that issue by specifying `margin-left: -15px` and `margin-right: -15px` as shown below:

https://github.com/MarkBind/markbind/blob/2f9461b4ea63ada3e4a2d6b2bdd062f996498e3d/asset/css/markbind.css#L68-L74

However, this workaround is counter-intuitive, as users wanting to override the css of `container-fluid` must now also modify the css of `footer`, as seen in https://github.com/se-edu/learningresources/commit/d8f206f04b5565dba134066dbf379109bc4683da.

As such, a better workaround on our site would be to change `page.ejs`:

https://github.com/MarkBind/markbind/blob/2f9461b4ea63ada3e4a2d6b2bdd062f996498e3d/lib/template/page.ejs#L20-L22

such that the footer is outside the `container-fluid` (if we do not want the padding effect of `container-fluid`, we should not include it inside in the first place):

```ejs
<div id="app">
    <div class="container-fluid">
        <%- content %>
    </div>
    <%- footer %>
</div>
```

Chng-Zhi-Xuan:

• [x] Bug fix

Fixes #390

User would like to standardise styling the site without having to deal with negative margins on the footer element.

-  Localised `container-fluid` class to only components that specify it.
- Adjusted values in `site-nav.css` to compensate for removed padding around border.
- Introduce a unique `#content-wrapper` div for styling page contents. 

- Is there a need to document overriding the `container-fluid` class with the user's custom CSS?

**Testing instructions:**
- Use `Netlify Preview` and remove the `website-content` class from inside the `page-content` div.
- Everything should look the same, in both desktop and responsive mode.


yamgent:
Bootstrap overrides the default margins of `<body>` and sets it to 0px (the default by the browser such as Chrome is 8px). Perhaps add a style to `<body>` to override this? But the user should still be able to override our defaults if necessary:

```
body {
  margin: 8px;
}
```

Chng-Zhi-Xuan:
Then the issue of the footer not being flush to the edge will resurface again.

The current work around is having the users create their own wrapper to add padding / margins while authoring in index.md.

The alternative is your original solution to shift it outside and insert it via the template. Then more testing is required to see if that implementation breaks the "sticky" feature of the footer.

yamgent:
> The alternative is your original solution to shift it outside and insert it via the template. Then more testing is required to see if that implementation breaks the "sticky" feature of the footer.

Let's try that then.

Chng-Zhi-Xuan:
**Update**
- Changes made after discussion with Prof.
- Add `container-fluid` class back but localise it to the authored contents.
- Revert `#app` selector changes.
- Override `container-fluid` class to increase spacing of website content from side edges.

**Disucssion**
- Is there a need to document overriding the `container-fluid` class?

asset/css/markbind.css
```diff
@@ -23,6 +23,12 @@ pre > code.hljs {
     outline: none !important;
 }
 
+/* Bootstrap responsive container class */
+.container-fluid {
+    max-width: 1200px;
```
yamgent: Why are we restricting the maximum width to 1200px?

lib/Page.js
```diff
@@ -72,6 +73,17 @@ function Page(pageConfig) {
  * Util Methods
  */
 
+function addContentWrapper(pageData) {
+  const $ = cheerio.load(pageData);
+  if ($.root().hasClass(CONTENT_WRAPPER_CLASS)) {
+    return pageData;
+  }
+  $(CONTENT_WRAPPER_CLASS).remove();
```
yamgent: What is the purpose of this line of code? Users should be able to add multiple `container-fluid` if he wants inside the website. They cannot be nested, but they can be used multiple times.

yamgent:
> Is there a need to document overriding the container-fluid class?

Overriding as in overriding the CSS? Don't think it's necessary, I think it is expected that the authors will know that Bootstrap classes can be overridden if they want, seems weird to be targeting specifically `container-fluid` class in that sense.

asset/css/markbind.css
Chng-Zhi-Xuan: Prof mentioned the content is difficult to read when it is too stretched out

Chng-Zhi-Xuan:
WIll add Padding / Margins once the screen size is small enough.

lib/Page.js
Chng-Zhi-Xuan: @yamgent 

If container-fluid class can be added multiple times. We shouldn't be using `container-fluid` as our page wrapper? It should be unique as using media queries for `container-fluid` affects the user using it within.

We can have our unique page wrapper inherit the responsive values from `container-fluid`, which is `overflow-x: hidden` and `width: 100%`.

yamgent: https://stackoverflow.com/questions/23062439/am-i-only-supposed-to-have-one-bootstrap-3-container

While we are using Bootstrap 4, I don't think the philosophy behind `container-fluid` has changed. So yes, one can use `container-fluid` multiple times. It just cannot be nested, that's all.

yamgent:
Hmm, looks like NavBar is getting affected by the 1000px limit change though:

Chng-Zhi-Xuan:
**Update**
- Changed page content wrapper from `.container-fluid` class to `#content-wrapper` id.
- Have `#content-wrapper` inherit responsive values from `.container-fluid`.

lib/Page.js
```diff
@@ -72,6 +73,17 @@ function Page(pageConfig) {
  * Util Methods
  */
 
+function addContentWrapper(pageData) {
+  const $ = cheerio.load(pageData);
+  if ($.root().has(`#${CONTENT_WRAPPER_ID}`).length) {
+    return pageData;
+  }
+  $(`#${CONTENT_WRAPPER_ID}`).remove();
```
yamgent: Actually, this line doesn't really do anything right? Consider the situation:

* The document already has an element of id `#content-wrapper`: Line 79 returns it straight away, so this line doesn't get executed.
* The document does not have an element of id `#content-wrapper`: The flow of execution will come to here, but since no such element exist, this statement has no effect.

Chng-Zhi-Xuan: My mistake.

`Line 78` is supposed to check if the root has a id of `content-wrapper`. It wasn't intended to check its contents if there exists an id of `content-wrapper`.

Chng-Zhi-Xuan:
**Update**
- Squashed changes requested

Issue #133, opened 2018-02-01
Logging for Site operations


nicholaschuayunzhi:
Found this as a todo in `Site.js` in `markbind-cli`

https://github.com/MarkBind/markbind-cli/blob/25b595df8e91ac97f626004f1d9b770dca4adf2a/lib/Site.js#L80-L86

Just collating all the required operations to be logged for user feedback. A loading bar of sorts would be helpful as well. 

#### Build
- build start
- build info
  1. collect baseUrl
  2. build assets
  3. build source files
     - generate pages
  4. copy markbind asset
- build end + timing (speed)
#### Live Reload
- live reload start
- reload info
  - number of pages rebuilt
- live reload end + timing + ready to inspect changes

acjh:
- Verbosity should be one of the [options](https://github.com/MarkBind/markbind-cli#usage).
- Related: #117 (Provide timestamp in logging)

acjh:
https://github.com/MarkBind/markbind/issues/183#issuecomment-377523330:

> At some point we should adopt a proper logging framework so that we can increase logging verbosity when we want to capture more info.

ang-zeyu:
Closing this issue as the lower level details are unapplicable at this point, moved to #1715

Issue #117, opened 2018-01-19
[Request for feature] Provide timestamp in logging


okkhoy:
Currently, when live preview is running and file changes, the log on console only reads `info: Reload for file changed: <path>`

E.g:

```
info: Press CTRL+C to stop ...
info: Reload for file change: <path>\cs2103\website-jan18\schedule\week1\lecture.md
info: Reload for file change: <path>\cs2103\website-jan18\schedule\week1\lecture.md
```

As in the above example, we cannot say which line corresponds to when the file was  changed. It would help to have time stamp info here, to help identify the latest (or which changes went into the serving).

damithc:
For large websites, live reloading is not very useful due to the performance problem we have now. For smaller websites, live reloading is almost instantaneous, so the timestamp does not add much value (the loaded version is guaranteed to be the latest version)

acjh:
Resolved by #398
```
info: [14:35:25] Reload for file change: <path>\cs2103\website-jan18\schedule\week1\lecture.md
```

PR #398, opened 2018-07-31
Add progress bar and additional logging for building process

### LINKED ISSUE #388: Add a console progress bar when building website

When doing a `markbind build` and `markbind serve`, the console does not print anything to indicate progress. User will stare at a blank screen until the website finish building.

Progress reporting would be nice, especially if the websites are big and might take longer than 5 seconds to build.

nusjzx:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Resolves #388 

When doing a markbind build and markbind serve, the console does not print anything to indicate progress. User will stare at a blank screen until the website finish building.

Logging building process and add timestamp for live reloading

NA

**Testing instructions:**
1. test on 2103 website to see the log info

### DIFF
--- a/lib/Site.js
+++ b/lib/Site.js
@@ -9,6 +9,7 @@ const ignore = require('ignore');
 const nunjucks = require('nunjucks');
 const path = require('path');
 const Promise = require('bluebird');
+const ProgressBar = require('progress');
 const walkSync = require('walk-sync');
 
 const _ = {};
@@ -381,11 +382,13 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
 };
 
 Site.prototype.generate = function (baseUrl) {
+  const startTime = new Date();
   // Create the .tmp folder for storing intermediate results.
   logger.profile(GENERATE_SITE_LOGGING_KEY);
   fs.emptydirSync(this.tempPath);
   // Clean the output folder; create it if not exist.
   fs.emptydirSync(this.outputPath);
+  logger.info(`Website generation started at ${startTime.toLocaleTimeString()}`);
   return new Promise((resolve, reject) => {
     this.readSiteConfig(baseUrl)
       .then(() => this.collectAddressablePages())
@@ -395,6 +398,11 @@ Site.prototype.generate = function (baseUrl) {
       .then(() => this.buildSourceFiles())
       .then(() => this.copyMarkBindAsset())
       .then(() => this.writeSiteData())
+      .then(() => {
+        const endTime = new Date();
+        const totalBuildTime = (endTime - startTime) / 1000;
+        logger.info(`Website generation complete! Total build time: ${totalBuildTime}s`);
+      })
       .then(resolve)
       .catch((error) => {
         rejectHandler(reject, error, [this.tempPath, this.outputPath]);
@@ -408,6 +416,7 @@ Site.prototype.generate = function (baseUrl) {
  */
 Site.prototype.buildSourceFiles = function () {
   return new Promise((resolve, reject) => {
+    logger.info('Generating pages...');
     this.generatePages()
       .then(() => fs.removeAsync(this.tempPath))
       .then(() => logger.info('Pages built'))
@@ -476,6 +485,7 @@ Site.prototype.removeAsset
  = delay(Site.prototype._removeMultipleAssets, 1000);
 
 Site.prototype.buildAssets = function () {
+  logger.info('Building assets...');
   return new Promise((resolve, reject) => {
     const ignoreConfig = this.siteConfig.ignore || [];
     const outputFolder = path.relative(this.rootPath, this.outputPath);
@@ -487,6 +497,7 @@ Site.prototype.buildAssets = function () {
         assets.map(asset => fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, asset))),
       )
       .then(copyAssets => Promise.all(copyAssets))
+      .then(() => logger.info('Assets built'))
       .then(resolve)
       .catch((error) => {
         rejectHandler(reject, error, []); // assets won't affect deletion
@@ -522,8 +533,12 @@ Site.prototype.generatePages = function () {
     title: page.title,
     searchable: page.searchable !== 'no',
   }));
+  const progressBar = new ProgressBar(`[:bar] :current / ${this.pages.length} pages built`,
+                                      { total: this.pages.length });
+  progressBar.render();
   this.pages.forEach((page) => {
     processingFiles.push(page.generate(builtFiles)
+      .then(() => progressBar.tick())
       .catch((err) => {
         logger.error(err);
         return Promise.reject(new Error(`Error while generating ${page.sourcePath}`));
--- a/index.js
+++ b/index.js
@@ -135,7 +135,7 @@ program
     const site = new Site(rootFolder, outputFolder);
 
     const addHandler = (filePath) => {
-      logger.info(`Reload for file add: ${filePath}`);
+      logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       Promise.resolve('').then(() => {
         if (fsUtil.isSourceFile(filePath)) {
           return site.rebuildAffectedSourceFiles(filePath);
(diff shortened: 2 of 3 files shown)


yamgent:
* #388 mentions having a progress bar. You need to incorporate a progress bar to the build process. [This progress bar package](https://www.npmjs.com/package/progress) seems to be used by a lot of famous packages and tools (e.g. eslint).
* Seems like you are attempting #133 and #117 as well? [As mentioned by Aaron](https://github.com/MarkBind/markbind/issues/133#issuecomment-362311504), it is best to have the capability to turn this off as this may become very verbose.

yamgent:
As discussed online, the progress bar will be to determine the number of pages that is being built:

```
info: Building pages...
[####################........................................] 34% (13 out of 38 pages built)
```

lib/Site.js
```diff
@@ -239,6 +240,7 @@ Site.prototype.readSiteConfig = function (baseUrl) {
         this.siteConfig = config;
         this.siteConfig.baseUrl = (baseUrl === undefined) ? this.siteConfig.baseUrl : baseUrl;
         resolve(this.siteConfig);
+        logger.verbose('Site configuration read.');
```
yamgent: `Site configuration loaded`

yamgent: Also, move this above `resolve(this.siteConfig);`

lib/Site.js
```diff
@@ -325,6 +327,7 @@ Site.prototype.collectAddressablePages = function () {
   // Add pages collected by walkSync without duplication
   this.addressablePages = _.unionWith(this.addressablePages, globPaths,
                                       ((pageA, pageB) => pageA.src === pageB.src));
+  logger.verbose('Addressable pages collected.');
```
yamgent: `Addressable pages collected` (remove full stop)

lib/Site.js
```diff
@@ -338,7 +341,7 @@ Site.prototype.collectBaseUrl = function () {
     pre[path.dirname(now)] = true;
     return pre;
   }, {});
-
+  logger.verbose('Base url collected.');
```
yamgent: `baseUrl collected` (calling it "baseUrl", which is used by the user, instead of "Base url", will give user some familiarity).

lib/Site.js
```diff
@@ -378,14 +381,17 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
       userDefinedVariables[id] = html;
     });
   });
+  logger.verbose('User-defined variables collected.');
```
yamgent: `User-defined variables collected` (remove full stop)

lib/Site.js
```diff
@@ -378,14 +381,17 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
...
 };
 
 Site.prototype.generate = function (baseUrl) {
   // Create the .tmp folder for storing intermediate results.
+  const startTime = new Date();
```
yamgent: Move this above the comment (`startTime` is not part of the .tmp folder logic)

lib/Site.js
```diff
@@ -378,14 +381,17 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
...
   logger.profile(GENERATE_SITE_LOGGING_KEY);
   fs.emptydirSync(this.tempPath);
   // Clean the output folder; create it if not exist.
   fs.emptydirSync(this.outputPath);
+  logger.info('Start generating website.');
```
yamgent: Use a timestamp for this, otherwise this message doesn't seem to be necessary because the user knows that he is building the website, that's why he ran `markbind build` or `markbind serve`. :/

```javascript
logger.info(`Website generation started at ${startTime.toLocaleTimeString()}`);
```

lib/Site.js
```diff
@@ -395,6 +401,7 @@ Site.prototype.generate = function (baseUrl) {
       .then(() => this.buildSourceFiles())
       .then(() => this.copyMarkBindAsset())
       .then(() => this.writeSiteData())
+      .then(() => logger.info(`Build end. building time: ${(new Date() - startTime) / 1000}s`))
```
yamgent: Would be better to move the calculation out of the string interpolation to make things clearer.

I improved the message also:

```javascript
.then(() => {
    const endTime = new Date();
    const totalBuildTime = (endTime - startTime) / 1000;
    logger.info(`Website generation complete! Total build time: ${totalBuildTime}s`));
}
```

lib/Site.js
```diff
@@ -408,6 +415,7 @@ Site.prototype.generate = function (baseUrl) {
  */
 Site.prototype.buildSourceFiles = function () {
   return new Promise((resolve, reject) => {
+    logger.info('Start building pages.');
```
yamgent: `Generating pages...`

lib/Site.js
```diff
@@ -476,6 +484,7 @@ Site.prototype.removeAsset
  = delay(Site.prototype._removeMultipleAssets, 1000);
 
 Site.prototype.buildAssets = function () {
+  logger.info('Start building assets.');
```
yamgent: `Building assets...`

lib/Site.js
```diff
@@ -487,6 +496,7 @@ Site.prototype.buildAssets = function () {
         assets.map(asset => fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, asset))),
       )
       .then(copyAssets => Promise.all(copyAssets))
+      .then(() => logger.info('Assets built.'))
```
yamgent: `Assets built` (remove full stop)

lib/Site.js
```diff
@@ -514,16 +524,19 @@ Site.prototype.generatePages = function () {
...
   }
-
   this._setTimestampVariable();
   this.pages = addressablePages.map(page => this.createPage({
     faviconUrl,
     pageSrc: page.src,
     title: page.title,
     searchable: page.searchable !== 'no',
   }));
+  const progressBar = new ProgressBar(`[:bar] :current / ${this.pages.length} pages are built`,
```
yamgent: drop the `are`, so that it just says `X/Y pages built`.

acjh:
How do I set verbosity?

yamgent:
The verbose log messages are currently printed in the log, but not in the console. Perhaps we can add a `--verbose` option to `index.js`?

nusjzx:
@acjh the new commit contains the verbose option

acjh:
1. Can we set the option in `logger` from **index.js** instead?
1. Let's support `-v` shorthand.
1. Let's support repeated options (pretty standard) and use a `Number` instead of a `Boolean` flag.
   See: https://github.com/tj/commander.js/#coercion

yamgent:
As discussed, there's a lot of issues with the implementations of the verbose logging:
1. The current implementation, putting the option in `index.js`, won't work because the tests will fail (see Travis).
1. The implementation of putting the option inside `logger` (as suggested by @acjh above) will require the logger to be passed manually to all the `Page`s and `Site`s, because the logger instance created in `index.js` is not shared by everyone, so the option does not propagate inwards. Therefore, there will be quite a significant revamping effort of the logging system.

Since issue #388 is mainly about the progress bar (and not anything else), maybe we should just focus on the progress bar only in this PR? :P Strictly speaking, the verbose logging (and its options) should be implemented in a separate PR.

lib/Site.js
```diff
@@ -338,7 +339,6 @@ Site.prototype.collectBaseUrl = function () {
     pre[path.dirname(now)] = true;
     return pre;
   }, {});
-
```
yamgent: Unnecessary line change here.

damithc:
Tried the new version. I don't see any progress-bar in the console 🤔 Can give an example screenshot?

nusjzx:
image

damithc:
Mine:
image

v1.12.0

nusjzx:
found out this issue https://github.com/npm/npm/issues/12098

yamgent:
The package's issue tracker also has this issue: https://github.com/visionmedia/node-progress/issues/170.

Think all mintty-based terminals (which gitbash uses) has this issue. Can reproduce the missing progress bar issue when I run MarkBind on cygwin's mintty terminal.

PR #399, opened 2018-07-31
Update live preview on variable file change

### LINKED ISSUE #387: Live preview: Update webpages when `variables.md` is modified

1. Create a new website with `markbind init`.
1. Use an existing variable on `index.md` (e.g. `{{example}}`)
1. Do `markbind serve`.
1. Modify the contents of the variable that you used in `index.md` by modifying `variables.md`. Save `variables.md`.

After saving `variables.md`, the live preview server should rebuild `index.md` and the new variable content should show up.

`index.md` still shows the old variable content (i.e. you must restart the live preview server to see the changes)

Chng-Zhi-Xuan:

Resolves #387 

User would like to not reset live preview server to view changes in `variables.md` file.

- Re-generate every single page if `variables.md` file is changed. (Brute force solution)

- Code style

**Testing instructions:**
1. Checkout this PR
2. Download [387-test-v2.zip](https://github.com/MarkBind/markbind/files/2244430/387-test-v2.zip) and extract
3. Run `markbind serve`
4. Change `_markbind/variables.md` 
5. Both pages should be updated.

### DIFF
--- a/lib/Site.js
+++ b/lib/Site.js
@@ -380,6 +380,20 @@ Site.prototype.collectUserDefinedVariablesMap = function () {
   });
 };
 
+/**
+ * Collects the user defined variables map in the site/subsites
+ * if there is a change in the variables file
+ * @param filePaths array of paths corresponding to files that have changed
+ */
+Site.prototype.collectUserDefinedVariablesMapIfNeeded = function (filePaths) {
+  const variablesPath = path.resolve(this.rootPath, USER_VARIABLES_PATH);
+  if (filePaths.includes(variablesPath)) {
+    this.collectUserDefinedVariablesMap();
+    return true;
+  }
+  return false;
+};
+
 Site.prototype.generate = function (baseUrl) {
   // Create the .tmp folder for storing intermediate results.
   logger.profile(GENERATE_SITE_LOGGING_KEY);
@@ -549,9 +563,13 @@ Site.prototype.generatePages = function () {
 Site.prototype.regenerateAffectedPages = function (filePaths) {
   const builtFiles = {};
   const processingFiles = [];
+  const shouldRebuildAllPages = this.collectUserDefinedVariablesMapIfNeeded(filePaths);
+  if (shouldRebuildAllPages) {
+    logger.warn('Rebuilding all pages as variables file was changed');
+  }
   this._setTimestampVariable();
   this.pages.forEach((page) => {
-    if (filePaths.some(filePath => page.includedFiles[filePath])) {
+    if (shouldRebuildAllPages || filePaths.some(filePath => page.includedFiles[filePath])) {
       // eslint-disable-next-line no-param-reassign
       page.userDefinedVariablesMap = this.userDefinedVariablesMap;
       processingFiles.push(page.generate(builtFiles)


yamgent:
Actually, not really sure whether this feature would be useful in the end.

While it is nice that MarkBind automatically helps us rebuild the pages, the brute force method is equivalent to resetting the live preview server.

If the user is doing changes in `variables.md` multiple times then MarkBind is going to rebuild multiple times, and the server will not be responsive when the pages are rebuild. Therefore, the user might want some control in the rebuild process in that case, so I think it would be better for the user to rebuild the site themselves as and when they want to? (which is what we do currently) @damithc

acjh:
Agreed. How about `this._logIfVariableFileChanged(filePaths)` that informs the user instead?

damithc:
I discussed this with @Chng-Zhi-Xuan y'day. We wanted to try the simple brute-force approach first because
1. It's still faster than restarting
2. changes to variable files are rare. The live preview usually doesn't trigger until the IDE saves the file.
3. It is easy to stop the live preview while editing a variable. (you just have to stop the server).

As this fix is easier, we can dogfood that first. We can always revert it it worsens the situation or improve further at a later time.

BTW, I suggested giving a warning in the console that a full rebuild has been triggered and it will take longer than usual.

lib/Site.js
```diff
@@ -549,9 +549,15 @@ Site.prototype.generatePages = function () {
 Site.prototype.regenerateAffectedPages = function (filePaths) {
   const builtFiles = {};
   const processingFiles = [];
+  const variablePath = path.resolve(this.rootPath, USER_VARIABLES_PATH);
+  const variableFileChanged = filePaths.includes(variablePath);
+  if (variableFileChanged) {
+    this.collectUserDefinedVariablesMap();
+    logger.warn('Rebuilding all pages as variable file was changed');
+  }
```
acjh: 1. Maintain plural `variables`, e.g. `variables file` and `variablesPath`.
1. Can we extract this to:

```js
const shouldRebuildAllPages = this.collectUserDefinedVariablesMapIfNeeded(filePaths);
if (shouldRebuildAllPages) {
  logger.warn('Rebuilding all pages as variables file was changed');
}
```
where:
```js
Site.prototype.collectUserDefinedVariablesMapIfNeeded = function (filePaths) {
  const variablesPath = path.resolve(this.rootPath, USER_VARIABLES_PATH);
  if (filePaths.includes(variablesPath)) {
    this.collectUserDefinedVariablesMap();
    return true;
  }
  return false;
}
```

3. Should short-circuit:
```js
if (shouldRebuildAllPages || filePaths.some(filePath => page.includedFiles[filePath])) {
```

lib/Site.js
```diff
@@ -541,6 +541,20 @@ Site.prototype.generatePages = function () {
   });
 };
 
+/**
+ * Checks if variables file has been changed and rebuild variables map when necessary
```
acjh: - Phrase it similar to the function that it wraps:
  "Recollects the user defined variables map in the site/subsites if needed"

- Place it just under the function that it wraps.

lib/Site.js
```diff
@@ -573,6 +577,20 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
   });
 };
 
+/**
+ * Recollects the user defined variables map in the site/subsites
+ * if there is a change in the variables file
+ * @param filePaths array of paths corresponding to files that have changed
+ */
+Site.prototype.recollectUserDefinedVariablesMapIfNeeded = function (filePaths) {
```
acjh: - Let's go with `collect` rather than `recollect` (similar for the jsdoc).
  Just as it's not **re**[scrollIntoViewIfNeeded](
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded) or **re**[layoutIfNeeded](
https://developer.apple.com/documentation/uikit/uiview/1622507-layoutifneeded).

- I should have been more specific when I said "Place it just under the function that it wraps".
  I meant `collectUserDefinedVariablesMap`.

Chng-Zhi-Xuan:
**Update**
- Refactored and squashed

lib/Site.js
```diff
@@ -549,9 +563,13 @@ Site.prototype.generatePages = function () {
...
   const builtFiles = {};
   const processingFiles = [];
+  const shouldRebuildAllPages = this.collectUserDefinedVariablesMapIfNeeded(filePaths);
+  if (shouldRebuildAllPages) {
+    logger.warn('Rebuilding all pages as variables file was changed');
+  }
   this._setTimestampVariable();
   this.pages.forEach((page) => {
-    if (filePaths.some(filePath => page.includedFiles[filePath])) {
+    if (shouldRebuildAllPages || (filePaths.some(filePath => page.includedFiles[filePath]))) {
```
acjh: Unnecessary parentheses around `(filePaths.some(...))`

Chng-Zhi-Xuan:
**Update**
- Squashed changes requested

PR #405, opened 2018-08-02
Fix scroll to heading if anchor is in the same page

### LINKED ISSUE #402: Scroll to target goes off target if already at the target
v1.11.0

Steps.
* Go to https://nus-cs2103.github.io/website-base
* search for `role`
* hit Enter --> page scrolls to the correct location
* search for `role` again
* hit Enter

Expected: page remains at the same scroll position
Actual: page scrolls a bit, hiding the correct target

nusjzx:

• [ ] Documentation update
• [X] Bug fix
• [ ] New feature
• [ ] Other, please explain:

fIxes #402

when jumping to the target in the same page, result covered by the navbar

add `scrollToUrlAnchorHeading` to search callback, so that the scrolling is done every jumping

**Testing instructions:**
1. test on 2103 website

### DIFF
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -61,6 +61,7 @@ function setupWithSearch(siteData) {
         const page = `${baseUrl}/${match.src.replace('.md', '.html')}`;
         const anchor = match.heading ? `#${match.heading.id}` : '';
         window.location = `${page}${anchor}`;
+        scrollToUrlAnchorHeading();
       },
     },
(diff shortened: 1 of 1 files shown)


nusjzx:
This issue is because we are scrolling by the navbar height if there is hash change. If we scrolling by the navbar height in the searchbar callback and remove the hashChange callback, link on the website will be off. Also user will rarely search for the same thing twice.

PR #420, opened 2018-08-18
Remove clear line

### LINKED ISSUE #384: Do not clear console when MarkBind is launched

When MarkBind is launched, the console is always cleared. This behavior is counter-intuitive for console users.

We can get rid of this behavior by removing line 28: https://github.com/MarkBind/markbind/blob/2f9461b4ea63ada3e4a2d6b2bdd062f996498e3d/index.js#L26-L30

jamos-tay:

• [X] Bug fix

Fixes #384 

Prevent console clear when markbind is started.

Remove `clear()` call and `require` reference.

**Testing instructions:**
* Run `markbind`
* Console does not clear

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@
 
 // Entry file for Markbind project
 const chokidar = require('chokidar');
-const clear = require('clear');
 const fs = require('fs-extra-promise');
 const htmlBeautify = require('js-beautify').html;
 const liveServer = require('live-server');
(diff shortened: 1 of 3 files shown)


jamos-tay:
Sure 👍

acjh:
The **package-lock.json** file should be updated.

jamos-tay:
My bad, updated

yamgent:
Looks generally OK, however, I notice that your branch branched out from a very old version `1.7.2` (we are now at `1.12.0` :P).

old

1. Try to rebase on the latest `master` commit so that we can be confidently sure that the fix works.
1. Also might be a good time to squash the commits together into one.

PR #423, opened 2018-08-19
Fix help command message

### LINKED ISSUE #422: Incorrect 'help' command given when an invalid command is entered
**MarkBind Version: 1.12.0**

When the user specifies an invalid command, they are told to run `markbind help`, which seems to also be invalid.

capture

The correct command should be `markbind --help`. Alternatively, `markbind help` could be added.

jamos-tay:

• [X] Bug fix

Fixes #422 

Fixes the invalid command message

Fixed the invalid command message

**Testing instructions:**
Enter `markbind invalid-command`
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -259,7 +259,7 @@ program.parse(process.argv);
 
 if (!program.args.length || !ACCEPTED_COMMANDS.includes(process.argv[2])) {
   if (program.args.length) {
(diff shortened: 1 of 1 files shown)


acjh:
I don't think this is a good idea. The message should be fixed instead.

jamos-tay:
Sure, I'll change it over then

PR #424, opened 2018-08-27
Add a --force-reload flag to markbind serve

### LINKED ISSUE #403: Add `--full-reload` option for `markbind serve`
Users might want a way to force all files to be rebuilt when making changes to a single document.

For example, we do not rebuild files when nunjucks files are changed, because we cannot detect the dependencies for those cases. Therefore, the user might want a full reset, and a `--full-reload` option means that the user need not have to do this manually on their own.

jamos-tay:

• [X] New feature

Fixes #403 

Changing certain files will not trigger a rebuild. User might want a way to force a rebuild of all files.

Add a `--force-reload` flag to `markbind serve` to allow user to rebuild all files whenever a file is changed.

When set, any change detected in the directory will immediately trigger a rebuild of all files.

The `delay` logic has been modified slightly:

* A few functions using `delay` was relying on it to wrap a variable in an array (e.g. `site.buildAssets(filePath)` actually gets converted to `Site.buildMultipleAssets([filePath])` by `delay` which seems a little odd). Wrapper logic was moved to the function itself.

* Allow `delay` to support multiple arguments


acjh:
Complications regarding rest parameters:
- https://github.com/MarkBind/markbind-cli/pull/31#discussion_r173388278
- https://github.com/MarkBind/markbind-cli/pull/31#discussion_r173615973

jamos-tay:
Hi @acjh, I've looked through the two comment threads and sorry, but could you explain a little more about what changes I should make?

My rationale for changing the delay function was to add an extra parameter `buildAsset(filepaths, forceReload)`, which was giving me problems as delay only allows one array parameter.

What I gathered from the first thread was that the old definition of delay adds a `[]` so it converts single arguments arg to an array:

```
  return function (arg) {
    
    // arg is a single value "test.html"
    
    context = this;
    if (Array.isArray(arg)) {
      pendingArgs = pendingArgs.concat(arg);
    } else {
      pendingArgs.push(arg);
    }
    
    // pendingArgs contains ["test.html"]
    
    ...

    const funcPromise = func.apply(context, [pendingArgs]); 
    // => func.apply(context, [["test.html"]])
    // => func(["test.html"]) is called instead of func("test.html")
```

Which is necessary for cases like `buildAsset("filepath") -> buildAsset(["filepath"])`.

But this seems strange to me because I expected delay to be a wrapper function that can be used with any `func`, it shouldn't be dependent on `func`'s definition. Is there any reason that we should enforce delay to only work with functions that take in one array?

Regarding the use of `...args`, the purpose was just to allow variable number of parameters for `func`. It should still work even if `func` doesn't use rest parameters.

For example:
```
function buildAsset(filepaths, forceReload) { // Not using rest
   ...
}

delayedbuildAsset = delay(buildAsset, 1000);

/*
    delayedbuildAsset is set to function(...args){
        ...
        buildAsset.apply(args);
    }
*/

delayedbuildAsset(["some filepath"], true); // Expect to call buildAsset(["some filepath"], true)

/*
    function(...args){
        // args is set to [["some filepath"], true]
        buildAsset.apply(args);
        // => buildAsset.apply([["some filepath"], true])
        // => calls buildAsset(["some filepath"], true) correctly
    }
*/
```

acjh:
If I recall correctly, your change would cause this:

```js
pendingArgs = [];
// ...
pendingArgs = [filePaths, forceReload];
// ...
pendingArgs = [filePaths, forceReload, filePaths, forceReload];
```

Did you test?

jamos-tay:
Ah I see... I didn't know delay aggregated the params and executed them all at once, I thought it was just a simple delay function, my bad.

I moved the forceReload param to the Site constructor, I think that might be simpler for now.

docs/userGuide/userQuickStart.md
```diff
@@ -104,6 +104,7 @@ Live reload is enabled to regenerate the site for changes, so you could see the
 |---|---|
 | `-p`, `--port <port>` | The port used for serving your website. |
 | --no-open | Don't open browser automatically. |
+| --force-reload | Forces a full reload of all site files when a file is changed. |
```
acjh: How about a `-f` short flag? If yes, move above `-p`, else move above `--no-open`.

index.js
```diff
@@ -129,12 +129,13 @@ program
   .description('build then serve a website from a directory')
   .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
   .option('--no-open', 'do not automatically open the site in browser')
+  .option('--force-reload', 'forces a full reload of all site files when a file is changed')
```
acjh: "~`forces`~`force` a full reload ..." — let's use imperative mood.

lib/Site.js
```diff
@@ -119,6 +119,7 @@ function Site(rootPath, outputPath) {
   this.baseUrlMap = {};
   this.siteConfig = {};
   this.userDefinedVariablesMap = {};
+  this.forceReload = forceReload;
```
acjh: Alphabetical order.

lib/Site.js
```diff
@@ -443,7 +444,8 @@ Site.prototype.buildSourceFiles = function () {
 };
 
 Site.prototype._rebuildAffectedSourceFiles = function (filePaths) {
-  const uniquePaths = _.uniq(filePaths);
+  const wrappedFilepaths = Array.isArray(filePaths) ? filePaths : [filePaths];
```
acjh: - `wrapped` seems like an odd way to describe an "ensuring an array", but I don't have a better name at the moment. 
- We should not lose the capital `P` for consistency.

acjh:
> Ah I see... I didn't know delay aggregated the params and executed them all at once, I thought it was just a simple delay function, my bad.

It was named after [Promise.delay](http://bluebirdjs.com/docs/api/promise.delay.html), but we can rename if there is a better name used by other libraries.

> - Allow `delay` to support multiple arguments

Remember to update PR description.

lib/Site.js
jamos-tay: How about just filePathArray?

jamos-tay:
I see, thanks for the information

index.js
```diff
@@ -127,14 +127,15 @@ program
...
+  .option('-f, --force-reload', 'force a full reload of all site files when a file is changed')
   .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
   .option('--no-open', 'do not automatically open the site in browser')
   .action((root, options) => {
     const rootFolder = path.resolve(root || process.cwd());
     const logsFolder = path.join(rootFolder, '_markbind/logs');
     const outputFolder = path.join(rootFolder, '_site');
 
-    const site = new Site(rootFolder, outputFolder);
+    const site = new Site(rootFolder, outputFolder, options.forceReload);
```
yamgent: Seems weird for the `Site`'s constructor to accept this parameter (because this is a build configuration parameter, rather than a site configuration parameter).

Maybe we should pass `options.forceReload` when calling `site.rebuildAffectedSourceFiles()` instead?

jamos-tay: Yes, I tried that previously, but I ran into some issues because of the way delay() was implemented, it only accepts a function that takes single array of paths so I can't add a second parameter to it (see PR comments for more info).

Maybe I can set the variable on Site separately?

yamgent: > Maybe I can set the variable on Site separately?

I guess it is ok for now, might be worth a revisit in case there's more options in the future though.

damithc:
@jamos-tay this feature doesn't seem to be reloading nunjucks code though :-(

PR #425, opened 2018-08-30
Support live preview for .mbd and .mbdf files

### LINKED ISSUE #331: Support live preview for .fr files
Context: I've been trying to use a different extension (`.fr`) for files containing page fragments, as opposed to addressable pages. The reasons:
* I can use `**/*.md` in `site.json` to indicate all pages that need to be rendered.  
* for my own convenience to distinguish full pages from page fragments. 

However, it seems `.fr` files are not captured by the live preview feature. When I edit a .fr file, there seems to be some activity in the `markbind serve` console but the changes don't appear in the rendered pages.

Suggestion: Capture changes to any file that is included irrespective of extension, not just .md files.

jamos-tay:

Fixes #331 

Certain markbind specific file extensions should be eligible for live preview

Recognize`mbd` and `mbdf` file extensions as source files.

### DIFF
--- a/docs/userGuide/developingASite.md
+++ b/docs/userGuide/developingASite.md
@@ -40,6 +40,13 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 | `-p`, `--port <port>` | The port used for serving your website. |
 | --no-open | Don't open browser automatically. |
 
+Note: Live reload is only supported for the following file types:
+
+- `.html`
+- `.md`
+- `.mbd` (MarkBind file)
+- `.mbdf` (MarkBind fragment)
+
 ## Build the static site
 
 ```
--- a/lib/markbind/lib/parser.js
+++ b/lib/markbind/lib/parser.js
@@ -179,7 +179,7 @@ Parser.prototype._preprocess = function (node, context, config) {
       return createErrorNode(element, e);
     }
 
-    const isIncludeSrcMd = utils.getExtName(filePath) === 'md';
+    const isIncludeSrcMd = utils.isMarkdownFileExt(utils.getExt(filePath));
 
     if (isIncludeSrcMd && context.source === 'html') {
       // HTML include markdown, use special tag to indicate markdown code.
@@ -375,8 +375,8 @@ Parser.prototype.includeFile = function (file, config) {
       const { parent, relative } = calculateNewBaseUrls(file, config.rootPath, config.baseUrlMap);
       const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
       const fileContent = nunjucks.renderString(data, userDefinedVariables);
-      const fileExt = utils.getExtName(file);
-      if (fileExt === 'md') {
+      const fileExt = utils.getExt(file);
+      if (utils.isMarkdownFileExt(fileExt)) {
         context.source = 'md';
         parser.parseComplete(fileContent.toString());
       } else if (fileExt === 'html') {
@@ -430,8 +430,8 @@ Parser.prototype.renderFile = function (file, config) {
         reject(err);
         return;
       }
-      const fileExt = utils.getExtName(file);
-      if (fileExt === 'md') {
+      const fileExt = utils.getExt(file);
+      if (utils.isMarkdownFileExt(fileExt)) {
         const inputData = md.render(data.toString());
         context.source = 'md';
         parser.parseComplete(inputData);
--- a/lib/markbind/lib/utils.js
+++ b/lib/markbind/lib/utils.js
@@ -1,6 +1,8 @@
 const fs = require('fs');
 const path = require('path');
 
+const markdownFileExts = ['md', 'mbd', 'mbdf'];
+
 module.exports = {
   getCurrentDirectoryBase() {
     return path.basename(process.cwd());
@@ -22,7 +24,7 @@ module.exports = {
     }
   },
 
-  getExtName(file) {
+  getExt(file) {
     const ext = file.split('.').pop();
     if (!ext || ext === file) {
       return '';
@@ -42,6 +44,10 @@ module.exports = {
       path.dirname(filename),
       path.basename(filename, path.extname(filename)) + ext,
     );
+  },
+
+  isMarkdownFileExt(ext) {
+    return markdownFileExts.includes(ext);
   },
 
   isUrl(filePath) {
--- a/lib/util/fsUtil.js
+++ b/lib/util/fsUtil.js
@@ -1,10 +1,10 @@
 const path = require('path');
 
+const sourceFileExtNames = ['.html', '.md', '.mbd', '.mbdf'];
+
 module.exports = {
-  isHtml: filePath => path.extname(filePath) === '.html',
-  isMarkdown: filePath => path.extname(filePath) === '.md',
   isSourceFile(filePath) {
-    return this.isMarkdown(filePath) || this.isHtml(filePath);
+    return sourceFileExtNames.includes(path.extname(filePath));
   },
 
   isInRoot: (root, fileName) => {
--- a/test/test_site/testIncludeMbd.mbd
+++ b/test/test_site/testIncludeMbd.mbd
@@ -1,0 +1,1 @@
+*MarkBind supports .mbd files.*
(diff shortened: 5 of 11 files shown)


nicholaschuayunzhi:
I think this is an existing bug on live. 

Issue is that live reload checks for pages specified in `site.json` and recursively checks files that have been included.

For the above set up, `test.html` is not `site.json` and has not been included in any page, thus the live reload does not rebuild it.

yamgent:
> I think this is an existing bug on live.
> 
> Issue is that live reload checks for pages specified in site.json and recursively checks files that have been included.
> 
> For the above set up, test.html is not site.json and has not been included in any page, thus the live reload does not rebuild it.

Thanks for the substantial explanation :+1:. @jamos-tay this bug needs to be fixed in this PR as well, otherwise adding these extensions alone won't have any useful effect for live preview.

acjh:
That should be a separate bug:
"Live reload does not copy assets that are not in site.json nor included"

yamgent:
> That should be a separate bug:
> "Live reload does not copy assets that are not in site.json nor included"

Ah ok, I think I see the confusion surrounding this.

There are two ways of using `.html` files:
1. Including the `.html` file inside Markdown files. For example:
   ```markdown
   <include src="test.html" />
   ```
   If the content of `test.html` is updated, then the rendered Markdown file should be updated as well (otherwise it is no longer a live preview).
2. Using `.html` files as standalone files (they can still be accessed manually in the server, but may not necessary be included by any Markdown file).
   
The first way already works on `master`. So we are not fixing that problem.

So I am under the presumption that we are fixing this for the second way, which is why I was requesting that bug to be fixed.

---

On another unrelated note, the `.mbd` and `.mbdf` needs a bit more work, because `parser.js` does not recognize them as source files correctly:

```
PS C:\Users\Tan Wang Leng\Desktop\New folder> markbind serve
  __  __                  _      ____    _               _
 |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
 | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
 | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
 |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|

 v1.12.0
info: Website generation started at 08:45:06
info: Building assets...
info: Assets built
info: Generating pages...
[-] 0 / 1 pages built
error: Error: Unsupported File Extension: 'mbd'
warning: Error: Error while generating C:\Users\Tan Wang Leng\Desktop\New folder\index.mbd
warning: Error: Error while generating C:\Users\Tan Wang Leng\Desktop\New folder\index.mbd
error: Error while generating C:\Users\Tan Wang Leng\Desktop\New folder\index.mbd
```

acjh:
> So I am under the presumption that we are fixing this for the second way, which is why I was requesting that bug to be fixed.

Correct. But it does not need to be fixed in this PR.

yamgent:
> Correct. But it does not need to be fixed in this PR.

Ok, I see. 👌

lib/util/fsUtil.js
```diff
@@ -1,10 +1,12 @@
 const path = require('path');
 
+const sourceFileExtensions = ['.html', '.md', '.mbd', '.mbdf'];
+
 module.exports = {
   isHtml: filePath => path.extname(filePath) === '.html',
   isMarkdown: filePath => path.extname(filePath) === '.md',
```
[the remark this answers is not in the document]
jamos-tay: Ah, okay.

Bug with the parser should be fixed, I added a test just in case.

yamgent:
`.mbd` files are still proper Markdown files (they are named differently because we want them to represent Markdown+extended MarkBind).

So when we use `.mbd` files as addressable pages in our `site.json`, they should be rendered if it contains Markdown syntax. But it doesn't render when I try it. :P

Replicating what Prof said in https://github.com/MarkBind/markbind/issues/331#issuecomment-404757610 (highlights by me):

> As discussed, we can support four extensions. They are recommendations only; user need not follow. **we treat all four types the same way.**
> 
> * `.md` : pure markdown files
> * `.html` : pure html files
>
> The above two have existing definitions. **It's not nice to redefine them to fit out purpose.**
> 
> * `.mb` : markbind files
> * `.mbf` : markbind fragments that do not represent addressable pages
>
> I don't like that md and mb look almost the same though :-p mbd and mbdf?

acjh:
Let's not treat **.html** files the same way (as in rendering Markdown) though.
I remember there was a related discussion, but I can't find it at the moment.

yamgent:
The file will be generated, but the Markdown syntax in the file does not render properly.

jamos-tay:
Odd, it seems to be fine on my side?

image

Could you try pulling from the branch again? Sorry for the inconvenience

test/test_site/testIncludeMbd.mbd
```diff
@@ -0,0 +1 @@
+<p>Markbind supports .mbd files.</p>
```
yamgent: Might want to use some Markdown syntax in the test.

Also, please leave a newline at the end of the file so that we can avoid the `No newline at end of file` message (same for the other files).

jamos-tay: Sure 👍

docs/userGuide/developingASite.md
```diff
@@ -40,6 +40,13 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 | `-p`, `--port <port>` | The port used for serving your website. |
 | --no-open | Don't open browser automatically. |
 
+Note: Live reload is only supported for the following file types:
+
+- `.html`
+- `.md`
+- `.mbd` (Markbind file)
+- `.mbdf` (Markbind fragment)
```
acjh: MarkBind with a capital B.

test/test_site/testIncludeMbd.mbd
```diff
@@ -0,0 +1 @@
+*Markbind supports .mbd files.*
```
acjh: MarkBind with a capital B.

lib/markbind/lib/utils.js
```diff
@@ -1,6 +1,8 @@
 const fs = require('fs');
 const path = require('path');
 
+const markdownFileExtensions = ['md', 'mbd', 'mbdf'];
```
acjh: This should be renamed since `getFileExt` returns `.md` etc.

lib/markbind/lib/utils.js
```diff
@@ -49,6 +51,10 @@ module.exports = {
     return r.test(filePath);
   },
 
+  isMarkdownFile(ext) {
```
acjh: `isMarkdownFileExt`

jamos-tay:
Updated

lib/util/fsUtil.js
```diff
@@ -1,10 +1,10 @@
 const path = require('path');
 
+const sourceFileExtensions = ['.html', '.md', '.mbd', '.mbdf'];
```
acjh: `*FileExts` seems like an abbrievation of `*FileExtensions`.
Also, our `utils.getExtName` returns without `.` while `path.extname` returns with.

Let's rename `utils.getExtName` to `utils.getExt`.
And then rename this to `sourceFileExtNames`.

`isSourceFile` and `isMarkdownFileExt` are fine.

lib/markbind/lib/parser.js
```diff
@@ -376,8 +376,8 @@ Parser.prototype.includeFile = function (file, config) {
       const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
       const fileContent = nunjucks.renderString(data, userDefinedVariables);
       const fileExt = utils.getExtName(file);
-      if (fileExt === 'md') {
-        context.source = 'md';
+      if (utils.isMarkdownFileExt(fileExt)) {
+        context.source = fileExt;
```
acjh: Do we need `context.source` to be the actual `fileExt`? Seems cleaner if we can treat custom Markdown file extensions as `'md'`.

acjh:
Please squash the commits.
Try to use meaningful commit messages instead of "PR Fixes" so that we can consider merging them as-is.

acjh:
*Remember to comment when you update a PR with a force-push because GitHub doesn't send notification emails for that.*

jamos-tay:
Okay, I'll keep that in mind next time

PR #428, opened 2018-09-05
Support marking keywords for search

### LINKED ISSUE #404: Support marking keywords for search
It is nice if we can mark keywords in the text as targets for search i.e., to be included in search index
e.g.,
```html
The quick brown <span class="keyword">fox</span> jumps over
the lazy <span class="keyword">dog</span>
```

These keywords can be accumulated under the index of the heading that is just above it.

jamos-tay:

• [X] New feature

Fixes #404 

Users might want certain keywords to be tagged to headers so they can be searched with the search bar.

Allow keywords to be tagged to headers.

User adds `<keyword>` tags after the header, e.g.

```
<searchbar :data="searchData" placeholder="Search" :on-hit="searchCallback"></searchbar>
# C#
<keyword>Visual Studio</keyword>
<keyword>.NET</keyword>
text text text text text text text text text text text 
# Java
<keyword>Eclipse</keyword>
<keyword>NetBeans</keyword>
<keyword>IntelliJ</keyword>
text text text text text text text text text text text 
```

Keywords will be added after the header name, allowing them to be searched.

image

Instead of using `<span class="keyword"></span>` as suggested by #404, I decided to use a keyword tag instead:
* Keyword doesn't have to be part of text
* Keywords can be placed directly under header for clarity
* More succinct


lib/Page.js
```diff
@@ -243,7 +243,15 @@ Page.prototype.collectHeadingsInContent = function (content) {
     $('modal').remove();
     $('panel').remove();
     $(headingsSelector).each((i, heading) => {
-      this.headings[$(heading).attr('id')] = $(heading).text();
+      const headingId = $(heading).attr('id');
+      const keywords = [];
+      $(heading).next().children('keyword').each((index, element) => {
```
acjh: This isn't general enough for the use case in the target issue. Keywords don't always come directly after a heading in a single `<p>` element.

asset/css/markbind.css
```diff
@@ -92,3 +92,7 @@ footer {
 #flex-div {
     flex: 1;
 }
+
+keyword {
+    display: none;
+}
```
acjh: CSS files are usually ordered by increasing specificity (and then alphabetical order).

acjh: This doesn't fulfil the requirements in the target issue at all. Remove this style.

jamos-tay:
I was thinking instead of inline keywords, we enforce keywords to be placed directly after the header, so it's completely separate from the text. I.e. we don't allow inline keywords at all, if the user wants to add a keyword he should explicitly place it below the header.

My worry with inline is that the keywords seem arbitrarily tagged to certain headers, so it may be a bit confusing, plus if headers get deleted or text gets moved then the tagging behavior might be a little unpredictable, especially if `include`s are involved. I can see cases in which we might want this behavior though.

Should I implement both, or just stick to what the original issue suggested?

Something like this:
```html
# Header
<span class="header-keyword">testing</span> <!-- need to hide this or it gets rendered as text -->

This is good for catching <span class="keyword">regressions</span>
```

damithc:
I prefer inline keywords because when a text segment moves, the keywords should move with it i.e., the keyword exist for the text segment, not for the heading. Otherwise we could end up with orphan keywords. However, our search results are based on headings, which means we have to figure out how to connect the inline keywords to the correct heading.

jamos-tay:
Okay, no problem

lib/Page.js
```diff
@@ -193,29 +194,58 @@ Page.prototype.prepareTemplateData = function () {
   };
 };
 
+/**
+ * Gets the closest header to an element.
+ * @param $ html content
+ * @param headingsSelector jQuery selector for selecting headings
+ * @param element to find closest heading
+ */
+function getClosestHeader($, headingsSelector, element) {
```
acjh: `getClosestHeading` and fix all wrongly used `headers`.

Don't confuse Headers and Headings: http://www.altformat.org/index.asp?pid=389

lib/Page.js
```diff
@@ -193,29 +194,58 @@ Page.prototype.prepareTemplateData = function () {
...
+  const prevHeader = $(element).prevAll(headingsSelector);
```
acjh: Plural `prevHeaders`.

lib/Page.js
```diff
@@ -193,29 +194,58 @@ Page.prototype.prepareTemplateData = function () {
...
+  if ($(prevHeader).length === 0) {
+    if ($(element).parent().length === 0) {
+      return null;
+    }
+    return getClosestHeader($, headingsSelector, $(element).parent());
+  }
+  return $(prevHeader.first());
```
acjh: [`.first()`](
https://api.jquery.com/first/) already returns a jQuery element.

test/test_site/testKeyword.md
```diff
@@ -0,0 +1 @@
+<span class="keyword">included keyword</span>
```
acjh: Newline at end of file.

lib/Page.js
jamos-tay: TIL, thanks!

lib/Page.js
```diff
@@ -226,28 +256,49 @@ Page.prototype.collectHeadingsInContent = function (content) {
...
   }
 };
 
+/**
+ * Concatenates keywords in this.keywords to heading text in this.headings.
+ */
+Page.prototype.concatenateHeadingsAndKeywords = function () {
+  Object.keys(this.keywords).forEach((headingId) => {
+    const keywordString = this.keywords[headingId].join(', ');
+    this.headings[headingId] += ` : ${keywordString}`;
```
acjh: Remove space before colon, or use a different symbol. I'd suggest a pipe `|`.

test/test_site/index.md
```diff
@@ -16,6 +16,26 @@ head: myCustomHead.md
 
 {{reference_level_4}}
 
+# Heading with keyword
+<span class="keyword">keyword</span>
```
acjh: Add another keyword here to test multiple keywords.

lib/Page.js
```diff
@@ -226,28 +256,49 @@ Page.prototype.collectHeadingsInContent = function (content) {
...
+    $('.keyword').each((i, keyword) => {
+      let closestHeading = getClosestHeading($, headingsSelector, keyword);
+      if (!closestHeading) {
+        if (!lastHeading) {
+          logger.warn(`Missing heading for keyword: ${$(keyword).text()}`);
+          return;
+        }
+        closestHeading = lastHeading;
+      }
+      const headingId = $(closestHeading).attr('id');
```
yamgent: Was wondering whether line 283 - 287 can be refactored to a different method, took me a while to understand what was going on (maybe `linkKeywordToHeading(keyword, heading)`)?

docs/userGuide/contentAuthoring.md
```diff
@@ -274,6 +276,17 @@ To use the searchbar within a navbar, add the following markup to your file. The
 </li>
 ```
 
+### Keywords
+
+You can also specify custom keywords to include under a heading by including keyword class:
```
yamgent: ... to ~~include~~ be included under a heading, by ~~including keyword class~~ tagging the words with the `keyword` class.

docs/userGuide/contentAuthoring.md
```diff
@@ -274,6 +276,17 @@ To use the searchbar within a navbar, add the following markup to your file. The
...
+### Keywords
+
+You can also specify custom keywords to include under a heading by including keyword class:
+
+```html
+# Heading
+This is good for catching <span class="keyword">regressions</span>
+```
+
+These keywords will be tagged to the immediate heading before it. Users will be able to search for these keywords using the searchbar.
```
yamgent: These keywords will be ~~tagged~~ linked to the immediate...

lib/Page.js
```diff
@@ -194,28 +195,57 @@ Page.prototype.prepareTemplateData = function () {
 };
 
 /**
- * Records headings inside rendered page into this.headings
+ * Gets the closest heading to an element.
```
yamgent: For consistency, remove the full stop at the end.

lib/Page.js
```diff
@@ -226,26 +256,57 @@ Page.prototype.collectHeadingsInContent = function (content) {
...
+Page.prototype.linkKeywordToHeading = function ($, keyword, heading) {
+  const headingId = $(heading).attr('id');
+  if (!(headingId in this.keywords)) {
+    this.keywords[headingId] = [];
   }
+  this.keywords[headingId].push($(keyword).text());
+};
+
+/**
+ * Concatenates keywords in this.keywords to heading text in this.headings.
```
yamgent: For consistency, remove the full stop at the end.

PR #440, opened 2018-09-28
Allow custom site.json path

### LINKED ISSUE #434: Allow user to specify site.json path
Follow up on discussion with @damithc on 19 Sep

Users may want to deploy single markbind project to multiple websites with different configurations, for example two information sites can share a common template, except a few extra pages are deployed for each one.

We can allow the user to specify the `site.json` file path in `markbind serve` so the same project can have multiple deploy options without having to swap the file out.

jamos-tay:

Fixes #434 

Users may want to specify a custom path for the site.json

Added a `--site-config filepath` option to `markbind serve` that allows user to specify filepath.

I changed some behavior in the Site.js constructor (line 100~) since folder paths (like `config/site.json` were failing. Would this work? Alternatively we can ban `/` in the site.json path.

### DIFF
--- a/docs/userGuide/developingASite.md
+++ b/docs/userGuide/developingASite.md
@@ -42,8 +42,6 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 | `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
 | --no-open | Don't open browser automatically. |
 
-Note: The site config file must be in the root folder.
-
 Note: Live reload is only supported for the following file types:
 
 - `.html`
--- a/docs/userGuide/developingASite.md
+++ b/docs/userGuide/developingASite.md
@@ -37,7 +37,9 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 
 | Options | Description |
 |----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
+| `-s`, `--site-config <filepath>` | Specify a custom filepath for site.json. |
 | --no-open | Don't open browser automatically. |
 
 Note: Live reload is only supported for the following file types:
--- a/docs/userGuide/developingASite.md
+++ b/docs/userGuide/developingASite.md
@@ -39,8 +39,10 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 |----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
-| `-s`, `--site-config <filepath>` | Specify a custom filepath for site.json. |
+| `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
 | --no-open | Don't open browser automatically. |
+
+Note: The site config file must be in the root folder.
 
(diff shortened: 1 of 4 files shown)


lib/Site.js
```diff
@@ -99,9 +98,8 @@ const GENERATE_SITE_LOGGING_KEY = 'Generate Site';
 const MARKBIND_WEBSITE_URL = 'https://markbind.github.io/markbind/';
 const MARKBIND_LINK_HTML = `<a href='${MARKBIND_WEBSITE_URL}'>MarkBind ${CLI_VERSION}</a>`;
 
-function Site(rootPath, outputPath, forceReload = false) {
-  const configPath = findUp.sync(SITE_CONFIG_NAME, { cwd: rootPath });
```
yamgent: Removing this line of code means that the user can no longer serve inside the subdirectories, but have to always go back to the root directory and serve. Not sure whether it is a good idea since it can be a convenience feature for the user.

jamos-tay: Hmm alright, then it might be better to enforce that all alternate site.json files be in the root folder. I don't really want to mess with the config path logic in case something messes up

lib/Site.js
```diff
@@ -99,7 +99,7 @@ const GENERATE_SITE_LOGGING_KEY = 'Generate Site';
 const MARKBIND_WEBSITE_URL = 'https://markbind.github.io/markbind/';
 const MARKBIND_LINK_HTML = `<a href='${MARKBIND_WEBSITE_URL}'>MarkBind ${CLI_VERSION}</a>`;
 
-function Site(rootPath, outputPath, forceReload = false) {
+function Site(rootPath, outputPath, forceReload = false, siteConfigPath = SITE_CONFIG_NAME) {
   const configPath = findUp.sync(SITE_CONFIG_NAME, { cwd: rootPath });
```
yamgent: Change `SITE_CONFIG_NAME` to `siteConfigPath`, so that user do not have to necessary serve it at the root.

docs/userGuide/developingASite.md
```diff
@@ -37,9 +37,13 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 
 | Options | Description |
 |----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
+| `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
 | --no-open | Don't open browser automatically. |
 
+Note: The site config file must be in the root folder.
```
yamgent: With the change proposed, this is no longer enforced.

lib/Site.js
jamos-tay: Hi, you mean allow siteConfigPath to be a filepath (like `configs/mysite.json`)? I think if we add that functionality then `markbind serve` must be run in the root folder (not subdirectory), since markbind currently uses the site.json folder to identify the root folder.

Alternatively we can have a different method of locating the root folder.

yamgent: No, I mean that it be changed so that it becomes from:

```js
const configPath = findUp.sync(SITE_CONFIG_NAME, { cwd: rootPath });
```

to 

```js
const configPath = findUp.sync(siteConfigPath, { cwd: rootPath });
```

Making this change still allows the user to serve the site inside subdirectories, unless the directory of `mysite.json` is different from `site.json` (then again in such a case the directory where `mysite.json` should take precedent as the root directory). `siteConfigPath` need not be a filepath for it to work (it can be just `mysite.json`).

jamos-tay: Ah, gotcha

PR #446, opened 2018-10-02
Fix keyword with panel headings

### LINKED ISSUE #444: Keywords inside panel are not searchable

* **MarkBind Version:** 1.13.0

```html
<panel header="# Some heading">
Some <span class="keyword">content</span>.
</panel>
```

When searching for `content`, it is associated with `# Some heading`.

The keyword cannot be searched at all.

jamos-tay:

• [X] Bug fix

Fixes #444 

The issue was that the panel was unexpanded, and those don't get indexed by search so keywords weren't indexed either.

Keywords should tag to the closest header that is indexed by the search data.

Following behaviour applies for keywords within a panel:

If the panel is expanded:

- Keyword tags to the closest header within the panel if there is one
- Otherwise, keyword tags to the panel header if there is one (must have `#`)
- Otherwise, keyword tags to the closest header

If the panel is unexpanded:

- Keyword tags to the panel header if there is one, ignoring headers within the panel
- Otherwise, keyword tags to the closest header

Added tests for cases.

### DIFF
--- a/test/test_site/index.md
+++ b/test/test_site/index.md
@@ -26,6 +26,25 @@ layout: default
   <span class="keyword">panel keyword</span>
 </panel>
 
+# Panel with heading with keyword
+<panel header="# Panel with heading" expanded>
+  <span class="keyword">panel keyword</span>
+</panel>
+
+# Expanded panel without heading with keyword
+<panel header="# Panel without heading with keyword" expanded>
+
+  # Keyword should be tagged to this heading, not the panel heading
+  <span class="keyword">panel keyword</span>
+</panel>
+
+# Unexpanded panel with heading with keyword
+<panel header="# Panel with heading with keyword">
+
+  # Keyword should be tagged to the panel heading, not this heading
+  <span class="keyword">panel keyword</span>
+</panel>
+
 # Heading with included keyword
 <include src="testKeyword.md" />
 
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -231,7 +231,7 @@ function getClosestHeading($, headingsSelector, element) {
  */
 Page.prototype.collectHeadingsAndKeywords = function () {
   const $ = cheerio.load(fs.readFileSync(this.resultPath));
-  this.collectHeadingsAndKeywordsInContent($(`#${CONTENT_WRAPPER_ID}`).html(), null);
+  this.collectHeadingsAndKeywordsInContent($(`#${CONTENT_WRAPPER_ID}`).html(), null, false);
 };
 
 /**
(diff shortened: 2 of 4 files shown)


jamos-tay:
I can't seem to figure out the problem with deploy - I don't think it's a bug caused by my changes, the only change is that expanded panels are scanned now for keywords (which shouldn't be a problem), and the scanning is causing an error when building the `docs`,

I'm encountering this really strange bug which happens even on `master`. Place the following files in the `Markdown` folder:

```
Markbind
  -- All the markbind source files --
  test-site/
    index.md: <include src="a/panel.md" />
    a/
      panel.md: <panel header="Some panel" src="b/test.html#a" expanded></panel>
      b/
        test.html: <div id="a">hello world</div>
```

When I run `markbind build` within the test-site folder, it runs perfectly. However, when I run `node index.js build test-site` within the main folder, it fails with the following error

image

It seems to boil down to this line of code:

image

`path.relative` is returning different values for the two commands (the error one has an extra `../`. But it shouldn't depend on the working directory?

jamos-tay:
Never mind, turns out its an issue with `path`:

test/test_site/index.md
```diff
@@ -25,6 +25,25 @@ head: myCustomHead.md, myCustomHead2.md
...
 
+# Panel with heading with keyword
+<panel header="# Panel with heading" expanded>
+  <span class="keyword">panel keyword</span>
+</panel>
+
+# Expanded panel without heading with keyword
+<panel header="# Panel without heading with keyword" expanded>
+
+  # Keyword should be tagged to the this heading, not the panel heading
```
yamgent: `to the this heading` -> `to this heading` (remove the `the`)

yamgent:
> When I run `markbind build` within the test-site folder, it runs perfectly. However, when I run `node index.js build test-site` within the main folder, it fails with the following error

All build files should be either inside the `.temp` folder or the `_site` folder. Likely inside the `.temp` folder at this stage of the build. (Note that of course `.temp` folder only appears temporary during the build process; it is deleted when the build completes/fails).

jamos-tay:
Hi - I _think_ I've resolved the issue but I'm not sure I did it correctly.

This is regarding this line change in `Page.js`, see the diff:

image

The include path being returned is correct when running `markbind build`, but not `node index.js build docs --baseUrl`.

The reason it's only failing on the `node` command is the `baseUrl` : for the first case, the baseUrl is `markbind/`, but for the second case its empty. So the second one used the other side of the ternary, and that's why it was causing problems.

I suspect there's something wrong with the usage of `path.relative`. I gathered from [here](https://stackoverflow.com/questions/31023972/node-path-relative-returns-incorrect-path) that `relative` treats file paths with a file as a folder. I.e. for "a/b.txt", b.txt is interpreted as a directory rather than a file name. So `path.relative("a/b.txt", "a/c/d.txt")` actually returns `../c/d.txt` when you would expect  `c/d.txt`. `this.sourcePath` was the path for the specific file, so when I retrieved the directory name with `dirname` and it worked fine. But that makes the other logic kind of weird, it might be just a coincidence that all the "../"s line up nicely. Maybe I can investigate this in a future PR.

Nevertheless tests are passing and the keyword behaviour is working properly:

capture
capture2

acjh:
Nice find. Then `path.dirname(this.sourcePath)` should apply in both cases?

jamos-tay:
It should... I tried it but it didn't work though

I think there's another issue with the path. The src is a relative path (like `userGuide/component/file.html` but the sourcePath is an absolute path (like `C:\Users\Desktop\...`) I think [path.relative](https://nodejs.org/api/path.html#path_path_relative_from_to) requires either both relative or both absolute because I'm getting weird results. I.e.

image

image

I'm expecting C to be `../userGuide ...` but I'm getting `../../userGuide ...` which is causing problems. Will try and find a workaround.

jamos-tay:
I'm not really sure how to properly fix this for now.

The way I understand it is this:

Find relative path from base to src attribute
Find relative path from from sourcePath to the above
Resolve from the resultPath

Which is fine and all if the baseUrl is provided, but it keeps giving me an extra `../` is the baseUrl is empty for the reasons above. I manually removed the extra `../` for now, I'm not sure what other paths I can explore yet, but at least it's working.

yamgent:
@jamos-tay Thanks for attempting a fix, understand that it is difficult to know what is going on, I myself have very little clue on why the original code was written that way...

...which is why I have taken the liberty to debug and rewrite the entire resolution logic for you, according to the best of my understanding.

I hope that this is correct. I believe it should be so, since I have tested on both MarkBind's `docs` and your example website given in https://github.com/MarkBind/markbind/pull/446#issuecomment-426351004, and the rewritten logic seems to work for both cases so far :).

Add a new commit in your PR with the following diff:

```diff
diff --git a/lib/Page.js b/lib/Page.js
index 17f47a1..d470b27 100644
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -276,10 +276,14 @@ Page.prototype.collectHeadingsAndKeywordsInContent = function (content, lastHead
       if (panel.attribs.src) {
         const src = panel.attribs.src.split('#')[0];
 
-        const srcPath = path.relative(path.dirname(this.sourcePath),
-                                      path.relative(this.baseUrl.substring(1), src.substring(1)));
-        const includePath = path.resolve(path.dirname(this.resultPath),
-                                         this.baseUrl ? srcPath : srcPath.substring(3));
+        const buildInnerDir = path.dirname(this.sourcePath);
+        const resultInnerDir = path.dirname(this.resultPath);
+
+        const includeRelativeToBuildRootDirPath = this.baseUrl ? path.relative(this.baseUrl, src) : src.substring(1);
+        const includeAbsoluteToBuildRootDirPath = path.resolve(this.rootPath, includeRelativeToBuildRootDirPath);
+        const includeRelativeToInnerDirPath = path.relative(buildInnerDir, includeAbsoluteToBuildRootDirPath);
+        const includePath = path.resolve(resultInnerDir, includeRelativeToInnerDirPath);
+
         const includeContent = fs.readFileSync(includePath);
         if (panel.attribs.fragment) {
           $ = cheerio.load(includeContent);
```

The variable names are long, but they are deliberately written that way so that I can tell what is going on. Feel free to shorten it, but please add comments as a form of substitution, so that future developers will be less confused by what is going on.

jamos-tay:
@yamgent Thanks a lot! This was tripping me up for the longest time 😆

I've checked both cases on my side and they're working too.

PR #454, opened 2018-10-15
Fix anchors for panel headers

### LINKED ISSUE #442: Show anchor icons for heading inside panels

* **MarkBind Version:** 1.13.0

Followup from https://github.com/MarkBind/markbind/issues/426#issuecomment-425449396.

```html
<panel header="# Some heading">
    ... Some content here ...
</panel>
```

An anchor icon appears right beside the heading.

No anchor icon show up beside the heading.

jamos-tay:

• [X] Bug fix

Fixes #442 

Anchors were not being generated for panel headers, and in a previous version do not redirect to the page when clicked.

Anchors should now be generated for panel headers correctly.

The panel headers are stored in the panel's `header` attribute and rendered by vue, so headings weren't being picked up by `Page`. I worked around this by checking all panels for the `header` attribute and adding the anchor html there.

As for the links not working previously (see https://github.com/MarkBind/markbind/issues/442#issuecomment-425638038), I added some code to `setup.js` to manually redirect the page when an anchor is clicked, it seems to be working as expected now.

### DIFF
--- a/test/test_site/expected/markbind/js/setup.js
+++ b/test/test_site/expected/markbind/js/setup.js
@@ -15,7 +15,7 @@ function flattenModals() {
   });
 }
 
-function setupAnchorVisibility() {
+function setupAnchors() {
   jQuery('h1, h2, h3, h4, h5, h6').each((index, heading) => {
     jQuery(heading).on('mouseenter', function () {
       jQuery(this).children('.fa.fa-anchor').show();
@@ -24,12 +24,17 @@ function setupAnchorVisibility() {
       jQuery(this).children('.fa.fa-anchor').hide();
     });
   });
+  jQuery('.fa-anchor').each((index, anchor) => {
+    jQuery(anchor).on('click', function () {
+      window.location.href = jQuery(this).attr('href');
+    });
+  });
 }
 
 function executeAfterMountedRoutines() {
   flattenModals();
   scrollToUrlAnchorHeading();
-  setupAnchorVisibility();
+  setupAnchors();
 }
 
 function setupSiteNav() {
--- a/test/test_site/expected/markbind/js/setup.js
+++ b/test/test_site/expected/markbind/js/setup.js
@@ -16,13 +16,9 @@ function flattenModals() {
 }
 
 function setupAnchors() {
-  jQuery('h1, h2, h3, h4, h5, h6').each((index, heading) => {
-    jQuery(heading).on('mouseenter', function () {
-      jQuery(this).children('.fa.fa-anchor').show();
-    });
-    jQuery(heading).on('mouseleave', function () {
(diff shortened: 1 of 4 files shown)


jamos-tay:
The tests are passing locally on my machine (windows) but for some reason Travis seems to be failing.

Somehow when rendered on Travis the block of code I added causes the `src` attribute of panels to change, all the `\` are converted to `/`:

image

test/test_site/testUtil/diffHtml.js
```diff
@@ -89,7 +89,7 @@ const diffHtml = (expected, actual) => {
       insidePath = true;
     }
 
-    if (isDiff(part) && !insidePath) {
+    if (isDiff(part) && !insidePath && !isPathSeparatorDiff(part.value)) {
```
nicholaschuayunzhi: The change here allows for this situation to pass:
`expected.html`
```html
<div> \ </div>
```

`actual.html`
```html
<div> / <div>
```

nicholaschuayunzhi:
> Somehow when rendered on Travis the block of code I added causes the src attribute of panels to change, all the \ are converted to /.

The reason is because the Travis is running a unix env and the path separator generated will be `/` whereas on Windows, it will be `\`.

As for the diff that is being shown, I would think my diffing code will account for that and not trigger an error. Sorry about that 🙏. I will help to look into the matter.

Also see my comment about your fix!

nicholaschuayunzhi:
@jamos-tay The fix has been merged to master, you can go ahead and rebase on master and also drop commits `d6e6851` and `eb2f83a`

jamos-tay:
Sure, updated 👍

yamgent:
Seems OK, but user-usability wise, I actually expected the anchor to have appeared when the mouse is inside the panel's header box, but as of now you must hover the mouse inside the header text in order to reveal it.

anchor

jamos-tay:
Yeah... the dimensions of the header seems to be different for headers in panels.

I'll try and find a fix

jamos-tay:
@yamgent Hey, think it should be fixed now.

There's still a minor nit where the anchor will disappear if the user hovers over the header text first, then moves the mouse right until the empty space (basically `mouseleave` gets triggered for the header), but fixing it would be quite troublesome and I think it's small enough to ignore.

test/test_site/expected/markbind/js/setup.js
```diff
@@ -24,12 +24,17 @@ function setupAnchorVisibility() {
       jQuery(this).children('.fa.fa-anchor').hide();
     });
   });
+  jQuery('.fa-anchor').each((index, anchor) => {
```
yamgent: Remember to update the content of `test/test_site/expected/markbind/js/setup.js` with the new code that you have.

PR #465, opened 2018-12-12
Support Page Navigation Sidebar

### LINKED ISSUE #219: Support page navigation sidebar
v1.7.3

suggestion: provide a way to include a sidebar navigation, possibly auto-generated from the headings in the page.
e.g., right hand side of this page https://nus-cs2030.github.io/1718-s2/lec01/index.html

Chng-Zhi-Xuan:

• [x] Documentation update
• [x] New feature

Resolves #219 

User would like to be able to use a page navigation component that is automatically generated from headings within the page.

Prototype 2 Demo |
--- |
page-nav-prototype-2 |

Prototype 3, Responsive Design Demo |
--- |
page-nav-responsive

To Do | Done?
--- | ---
Responsive design | ✔️  
Change specifying `pageNav` within `front-matter` from `Boolean` to `{1 ... 6} or default` | ✔️ 
Add specifying `pageNavHeader` to `front-matter` as the header text for page navigation | ✔️ 
Adaptive CSS if `site-nav` or `footer` is not specified in the page | ✔️  
Page navigation automatically scrolls to active heading as user scroll the page | ✔️ 
Indentation of smaller headings | ✔️  
Include headings in `Panel headers` and `Panel contents` if expanded  | ✔️ 
Live preview via `markbind serve` re-inserts page navigation | ✔️ 
Update test site to use page navigation | ✔️ 
Change page navigation and scroll bar styling from default Bootstrap theme | ✔️ 
Update documentation for new feature and using it | ✔️ 
Support auto expanding and collapsing of panels upon clicking an anchor link | ⚠️<br> On hold 

 
References |
--- |
[Bootstrap Scrollspy](https://getbootstrap.com/docs/4.0/components/scrollspy/) |
[Custom Scrollbars](https://css-tricks.com/custom-scrollbars-in-webkit/) |
[Scrollspy Event Handling](https://stackoverflow.com/questions/48693913/bootstrap-4-activate-bs-scrollspy-event-is-not-firing) |

**Testing instructions:**
1. Use Netlify Preview
1. Go to the "Docs" tab then "Using Components" sub section and the page navigation bar should appear.


Chng-Zhi-Xuan:
**Discussion:**
1.  Currently page navigation is enabled via specifying a boolean value in `frontmatter`. Via 'pageNav: true'. Then it uses the `headingIndexingLevel` to select which headings are added to the navigation bar. <br><br> I was thinking if we should decouple it from `headingIndexingLevel` and specify a set of numbers {1, ... 6} instead?

2. Colour scheme, font size and styling suggestions are also welcome.

damithc:
This page has a well-done site-nave and page-nav if you need some inspiration :-) https://angular.io/guide/architecture-components

yamgent:
> 1. Specify a second variable in `frontmatter`, like `pageNavHeader: "My page nav header"`.

Oh I see. Second variable seems OK? If the author does not specify, it can be the name of the page itself.

> This was intentional as I felt having 2 scrollbars (page scroll) adjacent to each other was an eyesore. Hiding it would result in users perceiving that it is not scroll-able.

Hmm, looking at the Angular documentation for inspiration. :) The page nav scrollbar for the Angular documentation is on the right side, but it is of a different color compared to the actual page's scrollbar. It is also thinner than the default OS scrollbar. That style might be worth considering as well?

Chng-Zhi-Xuan:
**Update:**
- Prototype 2 is up, updated `To do` and `Demo` in PR comment 
    - Used `.small` class on `#pageNav` to fit more headings
    - Indentation of smaller headings is `+5% margin` of its immediate parent
- Removed headings from `Panel headers` in component documentation as it cluttered the page navigation severely with buggy behaviour (Bootstrap Scrollspy skips when "spied" elements are very close together)

asset/css/markbind.css
```diff
@@ -105,3 +108,14 @@ footer {
...
+
+#page-nav {
+    border-left: solid 1px lightgrey;
+    display: inline-block;
+    direction: rtl;
+    overflow: auto;
+    max-height: 80vh;
+    position: fixed;
+    width: 300px;
+    right: 0;
```
yamgent: Remember to sort the attributes in alphabetical order.

damithc:
@Chng-Zhi-Xuan What's the planned behavior for preloaded but collapsed sections? For dynamically-loaded sections?

Chng-Zhi-Xuan:
> @Chng-Zhi-Xuan What's the planned behavior for preloaded but collapsed sections? For dynamically-loaded sections?

@damithc , currently it is hard coded to not index content within any collapsed panels. 

It is easy to index it as the content is already there in the raw html file, however the `href` upon clicking would not work since it is hidden by `vuestrap` upon loading the page. The `href` will work once any of the content is visible on the page (expanded panel). 

I wanted to avoid having the users click on a `href` and wondering why it doesn't jump to that page. 

What I may explore is maybe have the page navigation dynamically change depending on which headers are visible at any given point in time. However it is most likely difficult to code as well as having performance issues on deeply nested panels having its contents revealed in 1 click.

damithc:
> I wanted to avoid having the users click on a `href` and wondering why it doesn't jump to that section.

What if the panel is expanded by default but collapsed by the user?

> What I may explore is maybe have the page navigation dynamically change depending on which headers are visible at any given point in time. However it is most likely difficult to code as well as having performance issues on deeply nested panels having its contents revealed in 1 click.

Sure, we can omit that from the first version.

Chng-Zhi-Xuan:
> What if the panel is expanded by default but collapsed by the user?

If the panel is expanded, it will get indexed. Upon collapsing by the user, the header link stays there but would not be able to jump to that header. 

The current planned behaviour is indexing any visible content upon loading the page, and not changing it if certain contents become visible / invisible.

acjh:
We could expand panels based on `window.location.hash`, so it works even when sharing links.

Related:
- [`scrollToUrlAnchorHeading`](https://github.com/MarkBind/markbind/blob/b7d191c9f810b8687f808ca6b978abe8b06b1dc5/asset/js/setup.js#L5-L11) in **setup.js**
- [`window.onhashchange`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onhashchange)

Chng-Zhi-Xuan:
> We could expand panels based on `window.location.hash`, so it works even when sharing links.

To clarify, you are suggesting that hidden headings within any panels will be expanded using a javascript function, upon clicking a `href` whose heading is not yet visible?

Might be tricky given that panels can be deeply nested. Also need to test behaviour with Bootstrap's ScrollSpy.

acjh:
> To clarify, you are suggesting that hidden headings within any panels will be expanded using a javascript function, upon clicking a `href` whose heading is not yet visible?

Correct.

Chng-Zhi-Xuan:
**Update:** 

1. Updated PR description To-Dos and added responsive design demo.

lib/Page.js
```diff
@@ -227,24 +242,65 @@ function getClosestHeading($, headingsSelector, element) {
...
-    headingsSelector += `, h${i}`;
+Page.prototype.collectNavigableHeadings = function (content) {
+  if (this.validPageNavigationSpecifier()) {
+    const { pageNav } = this.frontMatter;
+    let elementSelector;
+    // Use specified navigation level or default in this.headingIndexingLevel
+    if (pageNav === 'default') {
+      elementSelector = generateHeadingSelector(this.headingIndexingLevel);
+    } else if (Number.isInteger(pageNav)) {
+      elementSelector = generateHeadingSelector(parseInt(pageNav, 10));
```
yamgent: I believe the radix parameter is optional, so it can be `parseInt(pageNav)` right?

lib/Page.js
```diff
@@ -485,6 +541,87 @@ Page.prototype.insertSiteNav = function (pageData) {
...
+          while (headingStack.length > 1 && topOfHeadingStack > currentHeadingLevel) {
+            headingHTML += '</nav>\n';
+            headingStack.pop();
+            topOfHeadingStack = headingStack[headingStack.length - 1];
+          }
+          if (topOfHeadingStack < currentHeadingLevel) {
+            // Increase nesting level by 1
+            headingHTML += nestedHeadingHTML;
+          } else {
+            headingHTML += currentHeadingHTML;
```
yamgent: There seems to be a mixture of determining how to partition the headings, and actual generation of the HTML code. Is it possible to separate the two? Otherwise the function gets really hard to read.

lib/Page.js
Chng-Zhi-Xuan: > If the input string begins with "0", radix is eight (octal) or 10 (decimal).  Exactly which radix is chosen is implementation-dependent.  ECMAScript 5 specifies that 10 (decimal) is used, but not all browsers support this yet.  **For this reason always specify a radix when using parseInt.** 

Taken from [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)

lib/Page.js
Chng-Zhi-Xuan: The actual HTML generation needs to be concurrent with partitioning. It builds it from top to bottom and at each stage. I guess I can apply SLAP and bring the HTML generation to another function, similar to `formatSiteNav`?

yamgent: > I guess I can apply SLAP and bring the HTML generation to another function, similar to `formatSiteNav`?

I think that's a good place to start. We can see if there's any further improvements to be made after following that.

Chng-Zhi-Xuan:
**Update:**

- Rebase to latest master, squashed commits
- Update test site to use page navigation
- PR description To-Dos updated
- Documentation still in progress
- Encountering some issues in implementing `Support auto expanding and collapsing of panels upon clicking an anchor link`, this functionality might be pushed to another PR.

**Discussion:**

Regarding the issue of whether page navigation exists in a layout. Currently to activate the page navigation feature, the user needs to input a number or `default` within the `frontmatter` of a page.

However, the layouts are file specific. **So is there a good case to have a stand-alone file to have page navigation supported in layouts?** Since the file will only have 2 lines (heading level and page navigation header text).

damithc:
> However, the layouts are file specific. **So is there a good case to have a stand-alone file to have page navigation supported in layouts?** Since the file will only have 2 lines (heading level and page navigation header text).

Introduce a new file e.g., `settings.md` into layouts that can contain this kind of settings?

Chng-Zhi-Xuan:
> Introduce a new file e.g., settings.md into layouts that can contain this kind of settings?

To extend your idea, currently the layout files for footer and other items are separated. It would be neater if all the different files (footer, siteNav, css) are all stored in their respective folders and have a simple `settings.md` file to "store" the file names required. Basically abstracting layouts to just a front matter file.

E.G, content within `settings.md`:
```
css: colorTheme1.css, blackTheme2.css
footer: simpleFooter.md
pageNav: default
pageNavHeader: "This is a list of stuff"
siteNav: mySiteNav.md
script: myScript1.js, myScript2.js
```

Then all the different file types are just stored in their respective folders within `_/markbind`

damithc:
> Then all the different file types are just stored in their respective folders within `_/markbind`

That way the same files can be used in different layouts right? That would be highly desirable.

Chng-Zhi-Xuan:
**Update:**

- Ready for review
- Updated PR description To Dos
- Added new `slim-scroll` class to `markbind.css`
- Added documentation for page navigation within `Tweaking The Page`.

**Bug:**
- The scollspy functionality skips headings if there are HTML tags within the headings. 
  - Within `Tweaking The Page` section, the heading with `<head>` enclosed within a `<code>` tag will not show up as active.

**Discussion**
- For the bug above, it probably can't be fixed and would require the author to avoid having tags within headings. Would that be OK?
- Would like reviewers to focus on the styling of the new scrollbar (which can be adopted within any scroll-able HTML element as well as the page navigation `.active` heading style.

damithc:
Nice progress!

For the bug, check if a heading with a font icon is affected. If it is, it's a significant problem because users can use font icons in headings often.  Also, how about things like these? `# My heading %%(contd)%%`

Use font color (rather than vertical bars) to indicate selected heading? As we already have two scroll bars on either side, using vertical bars to show current heading could become a case of 'too many vertical bars'? Alternatively, we can use dot or something.

Should we move the scrollbar to be on the other side? It is true that having two scrollbars near to each other is a problem. Having one in the middle of the page doesn't look nice too, especially because the current one does attract a significant amount of attention. Alternativley, we can try to make it even thinner/lighter but not sure if that will affect the usability.

Overall though, we  can release the feature as soon as the functionality is correct and good enough. UI tweaks can be done in future releases. Again, good work!

Chng-Zhi-Xuan:
> For the bug, check if a heading with a font icon is affected. If it is, it's a significant problem because users can use font icons in headings often. Also, how about things like these? # My heading %%(contd)%%

I did some extensive testing. So far `Glyphicons` and `Font-Awesome` icons have no issues. Inline Markdown syntax (`%%`, `==`, `**`) also causes no issues. The issues come in when using certain parenthesis symbols: `{` ,`}`, `(`, `)`, `{`, `}`, `[`, `]` anywhere in the heading, regardless of whether it is in a span or not. 

[Other users facing same issue](https://stackoverflow.com/questions/21295090/bootstrap-scrollspy-not-working-when-links-contains-parentheses)

> Use font color (rather than vertical bars) to indicate selected heading? 

I could make the original text slightly lighter, and have the `.active` class darken + bold + enlarge the words?

> Alternatively, we can use dot or something.

I would prefer a dot as well but I could not find a method of having a dot rendered purely from CSS `.active` class change.

> Should we move the scrollbar to be on the other side?

Preview |
--- |
right-scroll-bar

I replaced the scrollbar for `<body>` with a grey slim-scrollbar and then have an accent colour scrollbar for `pageNav`. I feel the accent colour (can be changed to something more subtle) is ok since it is very unlikely that the page is scroll-able but the `pageNav` is not.

yamgent:
> The issues come in when using certain parenthesis symbols: `{` ,`}`, `(`, `)`, `{`, `}`, `[`, `]` anywhere in the heading, regardless of whether it is in a span or not.
> 
> [Other users facing same issue](https://stackoverflow.com/questions/21295090/bootstrap-scrollspy-not-working-when-links-contains-parentheses)

That's... kinda annoying, especially since it is not uncommon to have URLs with these symbols these days. For example: https://en.wikipedia.org/wiki/JavaScript#Prototype-based_(object-oriented)

The issue's answer contains a further link to another issue, which states that [HTML5 allows any characters in its id except for space](https://stackoverflow.com/a/79022). The reason is probably to make URLs more user friendly (rather than having to replace `(` with `%28` for example, which makes the URL more ugly).

As such, I find it hard to justify modifying the anchor URLs on our side to replace the symbols in order to circumvent the problem with scrollspy. We may just have to live with the bug until the Bootstrap team finds a good way of fixing this problem?

asset/css/page-nav.css
```diff
@@ -0,0 +1,51 @@
...
+#page-nav-content-wrapper {
+    margin-right: 300px;
+    transition: 0.4s;
+    -webkit-transition: 0.4s;
+}
+
+/* Responsive site navigation */
+
+/* Hides page navigation on screen widths smaller than 1600px */
+@media screen and (max-width: 1599.98px) {
```
yamgent: I switched to my laptop for the first time in reviewing this PR, and the page navigation didn't appear, even though my resolution is 1920x1080.

I can't seem to figure out the exact problem, will let you know once I have further time to dive into this PR.

damithc:
> It is clickable and functions just as normal, but the Scrollspy will not toggle the text to be `.active`, thus seem like "skipping" it. As long as there are parenthesis in the text, the Scrollspy does not toggle `.active` class on that specific heading.

> As such, I find it hard to justify modifying the anchor URLs on our side to replace the symbols in order to circumvent the problem with scrollspy. We may just have to live with the bug until the Bootstrap team finds a good way of fixing this problem?

Indeed it is annoying but not necessarily show-stopping. I guess we can live with it for the time being.

> I will change the main `slim-scroll` class to be a lighter grey and slimmer (5px). As for the `accented-slim-scroll`, maybe a pastel blue of `#aec6cf` ?

Sure, let's give that a try. Try to use bootstrap colors though, to make the feature theaming-compatible.

> Darken only seems more "modern". The current behaviour of `pageNav` is that it can activate on multiple headings as once so underline might seem excessive.

I'm OK with darken-only 👍

asset/css/page-nav.css
```diff
@@ -0,0 +1,36 @@
+/* Page navigation */
+
+#page-nav {
+    border-left: solid 1px lightgrey;
+    display: inline-block;
+    direction: rtl;
```
yamgent: `display` should come after `direction`.

src/Page.js
```diff
@@ -227,24 +242,76 @@ function getClosestHeading($, headingsSelector, element) {
...
 
 /**
- * Records headings and keywords inside rendered page into this.headings and this.keywords respectively
+ * Checks if page.frontMatter has a valid page navigation specifier
+ * @returns boolean
  */
-Page.prototype.collectHeadingsAndKeywords = function () {
-  const $ = cheerio.load(fs.readFileSync(this.resultPath));
-  this.collectHeadingsAndKeywordsInContent($(`#${CONTENT_WRAPPER_ID}`).html(), null, false);
+Page.prototype.validPageNavigationSpecifier = function () {
```
yamgent: Rename the method name to `isPageNavigationSpecifierValid` so that it is understood that it returns a boolean (then perhaps the comment `@returns boolean` is no longer necessary?)

src/Page.js
```diff
@@ -227,24 +242,76 @@ function getClosestHeading($, headingsSelector, element) {
...
+    let elementSelector;
+    // Use specified navigation level or default in this.headingIndexingLevel
+    if (pageNav === 'default') {
+      elementSelector = generateHeadingSelector(this.headingIndexingLevel);
+    } else if (Number.isInteger(pageNav)) {
+      elementSelector = generateHeadingSelector(parseInt(pageNav, 10));
+    } else {
+      // Not a valid specifier
+      return;
+    }
```
yamgent: How about like this, we refactor out this logic:

```js
Page.prototype.generateElementSelectorForPageNav = function (pageNav) {
  if (pageNav === 'default') {
    return generateHeadingSelector(this.headingIndexingLevel);
  } else if (Number.isInteger(pageNav)) {
    return = generateHeadingSelector(parseInt(pageNav, 10));
  } else {
    // Not a valid specifier
    return undefined;
  }
}
```

Then:

* `collectNavigableHeadings()` can directly call this method. `isPageNavigationSpecifierValid()` will just return 
* `generateElementSelectorForPageNav(pageNav) !== undefined`.

Justification: To avoid duplicating the logic of handling pageNav as much as possible (lest we forget to change both places if there's a change in the behaviour of pageNav in the future).

asset/css/page-nav.css
Chng-Zhi-Xuan: If you zoom in within the browser, `pageNav` will disappear as well.

Also, within Windows, if your UI scaling is more than 100%, browsers will treat your resolution as being smaller than actual.

src/Site.js
```diff
@@ -655,13 +662,23 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
...
   logger.info(`Rebuilding ${processingFiles.length} pages`);
 
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
+      .then(() => {
+        this.pages.forEach((page) => {
+          if (this.modifiedPageSrc.includes(page.src)) {
+            page.collectHeadingsAndKeywords();
+            page.insertPageNav();
+            page.concatenateHeadingsAndKeywords();
```
yamgent: This part (and some affected parts such as `collectHeadingsAndKeywords`), is concentrated on regenerating the headings and keywords whenever it is modified during the live preview right? It is already done in #558 and #568, so changes will have to be made accordingly, so just a heads up.

docs/userGuide/glossary.md
```diff
@@ -34,7 +34,16 @@ MarkBind is able to take Markdown files (`.md`) and HTML files (`.html`) as sour
...
 
-_Site Navigation Menu_ (_siteNav_ for short) is a fixed menu on the left side of your page contents, that shows a road map of the main pages of your site.
+_Site Navigation Menu_ (_siteNav_ for short) is a fixed menu on the **left** side of your page contents, that shows a road map of the main pages of your site.
 </span>
 
-</div>
\ No newline at end of file
+</div>
+
+#### Page Navigation Menu
```
yamgent: The indentation is off with the rest of the entries.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -239,6 +240,38 @@ A siteNav has a fixed width of 300 pixels for its contents. A siteNavs is [_resp
 There is no limit to the number of nesting levels or the number of items in the menu, but note that any content exceeding a height of 1000 pixels will be cut off.
 
 
+<hr><!-- ======================================================================================================= -->
+
+## Page Navigation Menus
+
+**You can also add a <trigger trigger="click" for="modal:pageStructure-pageNavidationMenu">_page navigation menu_</trigger> to a page.**
+
+<modal title="Site Naviation Menu" id="modal:pageStructure-pageNavidationMenu">
```
yamgent: title should be "Page Navigation Menu"?

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -239,6 +240,38 @@ A siteNav has a fixed width of 300 pixels for its contents. A siteNavs is [_resp
...
+  <include src="glossary.md#page-navigation-menu" />
+</modal>
+
+Steps to add a page navigation menu ==(_pageNav_ for short)==:
+1. Specify your pageNav within the `<frontmatter>` of a page with <tooltip content="The value `default` will use `headingIndexingLevel` within `site.json`.">`"default"`</tooltip> or a <tooltip content="HTML defines six levels of headings, numbered from <br>`1 to 6`.">`heading level`</tooltip>.
+2. `[Optional]` You may also specify page navigation header text within `<frontmatter>` that will be placed at the top of the page navigation menu.
```
yamgent: * We currently use "(Optional)" instead of \`[Optional]\` in our documentation.
* Would it be better to say "page navigation title" instead of "header text"?

asset/css/page-nav.css
yamgent: > Also, within Windows, if your UI scaling is more than 100%, browsers will treat your resolution as being smaller than actual.

I see, that could be the problem. I wonder if there's any way around it, because it feels like a pity to miss out this feature when there's space for it (I forced it to always render and it seems to have sufficient space):

pagenav

Chng-Zhi-Xuan: My original intention was to give the contents a minimum of 1000px width for the `<body>` content. (`siteNav` and `pageNav` both are `300px`)

But upon checking the `siteNav.CSS` which uses `(max-width: 991.98px)`, it might be more appropriate to use `(max-width: 1199.98px)[Bootstrap Large responsive breakpoint]` + `100px` to accommodate the `300px` siteNav. Thus leaving the `<body>` content with a minimum of `692px` width.

src/Page.js
Chng-Zhi-Xuan: This implementation will result in unnecessary processing since `generateElementSelectorForPageNav(pageNav)` will invoke `generateHeadingSelector` if the specifier is valid.

We can still do the `generateElementSelectorForPageNav(pageNav)` abstraction to avoid duplicate logic within `collectNavigableHeadings`. 

```Javascript
  const { pageNav } = this.frontMatter;
  const elementSelector = this.generateElementSelectorForPageNav(pageNav);
  if (elementSelector === undefined) {
    return;
  }
  // Rest of function
```
To make the specifier checking more efficient, maybe we can store the result of `isPageNavigationSpecifierValid()` in a variable (`this.hasValidPageNavigationSpecifier`)?

yamgent: > This implementation will result in unnecessary processing since `generateElementSelectorForPageNav(pageNav)` will invoke `generateHeadingSelector` if the specifier is valid.

You are right, it is indeed inefficient.

> To make the specifier checking more efficient, maybe we can store the result of `isPageNavigationSpecifierValid()` in a variable (`this.hasValidPageNavigationSpecifier`)?

Not sure whether it is worth going this far actually.

I guess the original code is OK then, sometimes it necessary to have the "duplicates" in order to get the trade offs.

asset/css/page-nav.css
yamgent: > I don't see a way around it since its something we can't control. We can only make it slightly better by implementing the solution above (dropping `1599.98px` to `1299.98px`).

I see. Let's try the solution then. I am not very sure about the common resolutions and scalings that students have, but if multiple students are facing the same problem and their screen space is reasonably big, then we may need to think of some other solutions as well.

Chng-Zhi-Xuan:
**Update:**
- Rebase to latest master
- Applied changes requested
- Changed style of `.slim-scroll` and `#page-nav`

**Discussion:**
- It seems like Bootstrap nor Bootswatch has themes for scrollbars. Furthermore, the only colour utilities within Bootstrap are for text and background, so it is not easy to apply it to scrollbars (which consists of several sub-components).<br><br> So tentatively I am using MarkBind's blue (`#00b0ef`) for `.slim-scroll-blue` instead of Bootstrap's colour. 

- The current darken effect for `.active` headings within `pageNav` is quite subtle. Having it darken + bold seems to be more visually appealing. Thoughts?

Current style | Bold style
--- | ---
current-style | bold-style

yamgent:
> * The current darken effect for `.active` headings within `pageNav` is quite subtle. Having it darken + bold seems to be more visually appealing. Thoughts?

On my side the difference seems to be obvious enough, so I don't think it needs change. I am open to changing it if more people think otherwise.

Will get a review on this PR tomorrow, kinda exhausted now. :sweat_smile:

damithc:
Looks good. This is how it appears on my side:
image

Noticeable enough. Will it be too much if we also use blue text color to indicate the selected item?

Shall we keep the normal page scrollbar untouched as users might see it as part of the browser rather than the page? i.e., they will not expect it to change from page to page within the same website

The PageNav scroll bar can be thin but grey as it doesn't need to attract attention either. 

The below is a bit odd from the reader POV (i.e., the PageNav hilights the item recently hidden rather than the item currently visible), but can't be helped I guess?

image

Chng-Zhi-Xuan:
> Noticeable enough. Will it be too much if we also use blue text color to indicate the selected item?

Original intention was to style it similar to `siteNav` for consistency. Using blue text would make it look like a generic link.

> Shall we keep the normal page scrollbar untouched as users might see it as part of the browser rather than the page?

We should do this since the scrollbar default style depends on the browser. Other well designed websites keep the default style as well (except for Angular JS documentation).

> The PageNav scroll bar can be thin but grey as it doesn't need to attract attention either.

Will change to default scrollbar for `<body>` and slim grey for `pageNav`

> The below is a bit odd from the reader POV 

This is controlled by the [`data-offset`](https://getbootstrap.com/docs/4.0/components/scrollspy/#options) parameter, which I've set to `100px`. 

In your example, the `Trigger` heading has not reached within `100px` from the top of the browser, thus the Scrollspy will still highlight the previous heading.

damithc:
> Original intention was to style it similar to `siteNav` for consistency. Using blue text would make it look like a generic link.

Yes, that sounds reasonable.

Chng-Zhi-Xuan:
**Update:**
- Rebase to latest master + squashed commits
- Applied changes requested
- Added page navigation to `Reusing Contents` subsection as it was rather long

asset/css/page-nav.css
```diff
@@ -0,0 +1,49 @@
...
+    max-height: 80vh;
+    overflow: auto;
+    position: fixed;
+    right: 0;
+    transition: 0.4s;
+    width: 300px;
+    -webkit-transition: 0.4s;
+}
+
+#page-nav a:link, #page-nav a:visited {
```
yamgent: Put it in two lines, like so:

```css
#page-nav a:link,
#page-nav a:visited {
```

src/Page.js
```diff
@@ -227,25 +242,80 @@ function getClosestHeading($, headingsSelector, element) {
...
+  const { pageNav } = this.frontMatter;
+  const elementSelector = this.generateElementSelectorForPageNav(pageNav);
+  if (elementSelector === undefined) {
+    return;
+  }
+  const $ = cheerio.load(content);
+  $('modal').remove();
+  $(elementSelector).each((i, elem) => {
+    // Check if heading or panel is already inside an unexpanded panel
+    let insideUnexpandedPanel = false;
```
yamgent: `insideUnexpandedPanel` -> `isInsideUnexpandedPanel` (to make it clearer that the variable is a boolean).

Chng-Zhi-Xuan:
**Update:**
- Applied changes requested.
- Squashed Commits.

Chng-Zhi-Xuan:
**Update:**

- Re-ordered logic to insert page navigation within `Page.generate`. This is to avoid a double file write in previous implementation (called `insertPageNav()` within `Site.updateSiteData()`).

Chng-Zhi-Xuan:
**Update:** 

- Fixed indexing headers in html tags outside `content-wrapper`

yamgent:
As spoken, have to rebase this PR now with the new anchor generated ID, we should have no problems with the pesky `<`, `>` and other weird symbols now.

Chng-Zhi-Xuan:
**Update:**

- Rebased to latest master
- Anchor IDs now generate properly and does not introduce any "skipping" behaviour within Scrollspy.

damithc:
> * Anchor IDs now generate properly and does not introduce any "skipping" behaviour within Scrollspy.

That's great news! :100:

PR #467, opened 2018-12-12
Add element tagging functionality

### LINKED ISSUE #435: Allow authors to show/hide certain elements on deploy
Follow up on discussion with @damithc on 19 Sep

The user may want to be able to show/hide certain elements from a page depending on the deploy configuration.

Cited use case:

When generating multiple sites with only slightly differing content, it might be easier to store all the information in one page, and allow the author to show/hide them. For example, if I have a Java tutorial page and a C# tutorial page, the text content can be exactly the same, only the code will differ:

```
# Print a line

<!-- Example syntax -->
<div class="require-flag --java">
    System.out.println("Hello world");
</div>
<div class="require-flag --Csharp">
    Console.WriteLine("Hello world");
</div>
```

The user can then deploy a Java or a C# version of the site without having two projects.

jamos-tay:

• [X] New feature

Fixes #435 

The user may want to be able to show/hide certain elements from a page depending on the deploy configuration.

Cited use case:

When generating multiple sites with only slightly differing content, it might be easier to store all the information in one page, and allow the author to show/hide them. For example, if I have a Java tutorial page and a C# tutorial page, the text content can be exactly the same, only the code will differ:

```
# Print a line

<!-- Example syntax -->
<div tag="lang--java">
    System.out.println("Hello world");
</div>
<div tag="lang--Csharp">
    Console.WriteLine("Hello world");
</div>
```

The user can then deploy a Java or a C# version of the site without having two projects.

Functionality implemented based on what was discussed here: https://github.com/MarkBind/markbind/issues/435#issuecomment-423080705

Tags (`{tagType}--{tagName}`) are added to any HTML element.
User can specify which tags to include in site.json or frontmatter. Frontmatter takes precedence.
Only elements with tags in site.json are displayed, rest are hidden.

Simplified from `include: { type: 'lang', include: ['java'] }` to just `include: { 'lang': ['java'] }` though.

Behavior works according to spec, but I'm wondering if it's a little unintuitive - do we want to mandate categorizing tags into different types (`lang` etc.) rather than just having general use tags, without the `{type}--`? I.e. `include : ['tag1', 'tag2' ... ]` I don't think we lose any functionality by having the latter.


damithc:
> Behavior works according to spec, but I'm wondering if it's a little unintuitive - do we want to mandate categorizing tags into different types (`lang` etc.) rather than just having general use tags, without the `{type}--`? I.e. `include : ['tag1', 'tag2' ... ]` I don't think we lose any functionality by having the latter.

Yes, that's fine too.

Shouldn't we allow multiple tags i.e., `<div tags="java, basic, optional">`

Tagged elements are shown by default or hidden by default?

jamos-tay:
They're always hidden by default unless a matching tag is specified.

I'm not sure about matching tags... Does that mean if a single tag matches, we display the element?

acjh:
1. **General-use tags**
   >> without the `{type}--`? I.e. `include : ['tag1', 'tag2' ... ]`
   >
   > Yes, that's fine too.

   Agreed. `tag1` should be able to contain dashes though.

1. **Multiple tags**
   > Shouldn't we allow multiple tags i.e., `<div tags="java, basic, optional">`

   Yes. Shall we follow HTML `class` attribute syntax? i.e. `<div tag="java basic optional">`

   > Does that mean if a single tag matches, we display the element?

   Yes, I think so.

1. **Tagged elements are shown by default or hidden by default**.
   > They're always hidden by default unless a matching tag is specified.

   Agreed, with an important exception: if the config is `undefined` or `null`, then show all.

1. **`include:` and `"includedTags":` keywords**

   `"renderTags":` is more appropriate as it's in MarkBind's post-rendering phase.

1. **Specifying in front matter**

   Is there a compelling reason to specify in front matter? It seems sufficient to be in **site.json**.

1. **Tagging functionality as a plugin**

   The clean implementation makes me wonder if we can consider a plugin concept, e.g. in Eslint.
   For a start, we can have a `postRender` entry point. This means something like this in **site.json**:
   ```js
   {
     "plugins": [
       "renderTags"
     ],
     "context": { // Can consider a different name
       "renderTags": ["shown", "shown-2"]
     }
   }
   ```
   And in **src/plugins/render-tags.js**:
   ```js
   module.exports = {
     postRender: function (content, siteContext, pageConfig) {
       const renderTags = siteContext.renderTags;
       // ...
       return $.html();
     },
   };
   ```
   And in **Page.js**:
   ```diff
   - .then(result => this.filterTags(this.frontMatter.includedTags, result))
   + .then(result => this.postRender(result, siteContext, pageConfig))
   ```
   ```js
   Page.prototype.postRender = function (content, siteContext, pageConfig) {
     let postRenderedContent = content;
     plugins.forEach((plugin) => { // Might be able to do `reduce` instead of `forEach`
       if (plugin.postRender) {
         postRenderedContent = plugin.postRender(postRenderedContent, siteContext, pageConfig);
       }
     });
     return postRenderedContent;
   };
   ```

1. **src/** directory (can be discussed in a separate issue)

   As suggested with **src/plugins/render-tags.js**, let's start using **src/** instead of **lib/**.
   - **src/** is pretty common.
   - **libs/** implies libraries' code, rather than *this* library's code. It's more nested but clearer.

   Current directory structure:
   ```
   lib/
     asset/
     markbind/
       lib/
         markdown-it/
         parser.js
         utils.js
       package-lock.json
       package.json
     template/
     util/
     Page.js
     Site.js
   package-lock.json
   package.json
   ```
   Proposed directory structure:
   ```
   src/
     asset/
     lib/
       markbind/
         src/
           lib/
             markdown-it/
           parser.js
           utils.js
         package-lock.json
         package.json
     template/
     util/
     Page.js
     Site.js
   package-lock.json
   package.json
   ```

damithc:
> 2\. es. Shall we follow HTML `class` attribute syntax? i.e. `<div tag="java basic optional">`

Sounds good.

> 2\. Does that mean if a single tag matches, we display the element?

For this round, yes, we can do that. But eventually we are are going to need a `exclude : ['tag1', 'tag2' ... ]` and some sort of precedence (e.g., left-to-right) so as to provide more control over what gets included/excluded

> 4\. `"renderTags":` is more appropriate as it's in MarkBind's post-rendering phase.

I prefer to not use `include` too as it is already a keyword in MarkBind. But should we use keywords that are more intuititve to the author e.g., `keep`/`omit`? `render` sounds like a dev term. 

**site.json vs frontmatter:** note that tag filtering need not be site-wide. Ideally, there should be a way to specify it for a group of files.

yamgent:
> 6\. **Tagging functionality as a plugin**
>     The clean implementation makes me wonder if we can consider a plugin concept, e.g. in Eslint.
>     For a start, we can have a `postRender` entry point. This means something like this in **site.json**:
>     ```js
>     {
>     ```

I think that it is a good idea going in the direction of allowing MarkBind to be extensible in the future. Separating the core logic with the "additional" features would create a good foundation for such a plugin system.

jamos-tay:
Changes:

Simplified tags:
`tags : ['tag-1', 'tag-2']`

I think we can just use the `tags` keyword everywhere, might be more succinct.
Removed the whole `--` system, the user now just specifies a list of tags to include.
Allow multiple tags, space delimited. Element displays if at least one tag matches
All elements will be displayed if tags are not specified, or empty


I left in the front matter functionality - do we want it in? Either way it's only a couple of lines of code changed so it's not a big thing

Is having both include and exclude necessary? I think it's sufficient to either include all and the user specifies `exclude`, or exclude all and the user specifies `include`. Is there a use case where we have to have both? 

Regarding the plugin thing, I think that's a great idea. It should be in its own PR though, I'll create another issue. We can migrate other functionalities (like anchors) to plugins too

damithc:
> Is having both include and exclude necessary? I think it's sufficient to either include all and the user specifies `exclude`, or exclude all and the user specifies `include`. Is there a use case where we have to have both?

While one can be used in place of the other, depending on the reuse context, one is usually more convenient than the other. Sometimes you want to say `show everything except x` (e.g., show everything except optional topics) in which case more tags can be added to the page and they will appear in pages without needing additional configurations in frontmatter/site.json.  At other times you want to say `show only x` (e.g., show only java examples) in which case one can add more tags and they will not appear in the pages unless specifically requested.

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+```html
+# Welcome to my site
+
+<p tags="language--english">Hello</p>
+<div tags="language--french">Bonjour</div>
+<span tags="language--spanish">Hola</span>
+```
+
+Tagged elements to include are specified in the `site.json` file, under the `tags` option:
```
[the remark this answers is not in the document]
yamgent: Also `siteConfiguration.md` will probably need to reflect this new option?

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+```
+{
+  ...
+  "tags": ["language--english"]
+}
+```
+
+When compiled, only elements that match tags specified in the `site.json` files will be displayed. All other tagged elements will be hidden. In this case, only the element with the `language--english` tag will be displayed. This is helpful when creating multiple versions of a page without creating multiple copies.
```
yamgent: ...of a page, without having to maintain separate copies. ~~creating multiple copies.~~

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+Elements can also have multiple tags, separated by a space. An element will be displayed if any tag matches the one in `site.json`.
```
yamgent: ~~Elements can also have~~ **You can also use multiple tags in a single HTML element. Specify each tag in the `tag` attribute ** separated by a space. An element will be displayed if any **of the tags** matches the one in `site.json`.

lib/Page.js
```diff
@@ -333,6 +334,28 @@ Page.prototype.concatenateHeadingsAndKeywords = function () {
   });
 };
 
+/**
+ * Filters out elements on the page based on config tags
+ * @param content of the page
+ * @param tags to filter
+ */
+Page.prototype.filterTags = function (tags, content) {
+  if (!tags || tags.length === 0) {
```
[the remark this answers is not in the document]
jamos-tay: Hmm how about this behaviour:

If `tags` property is not present in `site.json`, show all tags
If `tags` property is present but set to `[]`, show no tags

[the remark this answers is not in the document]
yamgent: Sure, that will work too.

docs/userGuide/siteConfiguration.md
```diff
@@ -50,6 +51,7 @@ Let's examine a typical `site.json` file:
...
 | **ignore** | An array of file patterns to be ignored. By default, MarkBind will copy all the files as assets into the output folder. The ignore pattern follows the [glob pattern used in .gitignore](https://git-scm.com/docs/gitignore#_pattern_format). For example, `*.md` ignores all markdown source files. You may want to ignore the Git directory `.git/*`. |
 | **deploy** | The settings for [auto deployment to Github pages](ghpagesDeployment.html). The `message` is the commit message used for the deployment commit. |
+| **tags** | A list of tags to filter page elements. See [Using Tags](usingComponents.html#using-tags) for more information. |
```
yamgent: Add a `Do not specify this option if you want to show everything.` will make it clearer without being too elaborate.

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+```
+{
+  ...
+  "tags": ["language--english"]
+}
+```
+
+When compiled, only elements that match tags specified in the `site.json` files will be displayed. All other tagged elements will be hidden. In this case, only the element with the `language--english` tag will be displayed. This is helpful when creating multiple versions of a page without having to maintain separate copies. If this option is not specified, all tagged elements will be displayed.
+
+**You can also use multiple tags in a single HTML element. Specify each tag in the `tag` attribute** separated by a space. An element will be displayed if **any of the tags** matches the one in site.json.
```
yamgent: Sorry, just spotted a typo. It should be `tags`, not `tag`? Same as with the code snippet below.

damithc: To add, can we avoid the term `compiled` as it can be misinterpreted? Also, is it misleading to say show/hide as it sounds like a client-side CSS thing? I assume this is more like include/omit thing that happens at the rendering stage, right?

jamos-tay: Yup.

Sure, I changed it to `You can use tags to selectively filter certain elements on web pages.`

jamos-tay:
Updated

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+You need to specify the tags to include in `site.json`, under the `tags` option:
+
+```
+{
+  ...
+  "tags": ["language--english"]
+}
+```
+
+When compiled, only elements that match tags specified in the `site.json` files will be displayed. All other tagged elements will be hidden. In this case, only the element with the `language--english` tag will be displayed. This is helpful when creating multiple versions of a page without having to maintain separate copies. If this option is not specified, all tagged elements will be displayed.
```
yamgent: **During rendering**, only elements that match tags specified in the `site.json` files will be **rendered**. All other tagged elements will be **filtered out**. In this case, only the element with the `language--english` tag will be **rendered**.... If this option is not specified, all tagged elements will be **rendered**.

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+**You can also use multiple tags in a single HTML element. Specify each tag in the `tags` attribute** separated by a space. An element will be displayed if **any of the tags** matches the one in site.json.
```
yamgent: ... element will be **rendered** if ...

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
 This is good for catching <span class="keyword">regressions</span>
 ```
 
+# Using Tags
+
+You can use tags to selectively filter certain elements on web pages.
```
acjh: "You can use tags to selectively filter ~`certain`~`HTML` elements ~`on web pages`~`when building a site`."

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+```
+{
+  ...
+  "tags": ["language--english"]
+}
+```
+
+During rendering, only elements that match tags specified in the `site.json` files will be rendered. All other tagged elements will be filtered out. In this case, only the element with the `language--english` tag will be rendered. This is helpful when creating multiple versions of a page without having to maintain separate copies. If this option is not specified, all tagged elements will be rendered.
+
+**You can also use multiple tags in a single HTML element. Specify each tag in the `tags` attribute** separated by a space. An element will be rendered if **any of the tags** matches the one in site.json.
```
acjh: <code>\`site.json\`</code>

Actually, my preference at work now is using **bold** for filenames (e.g. **site.json**) to differentiate from code.
The answers and comments on [this meta.stackoverflow.com question](https://meta.stackoverflow.com/questions/303145/how-should-files-names-be-formatted) make a fairly good case for *italics*.
However, MarkBind's existing convention is <code>\`backticks\`</code> so let's stick to that.

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+```html
+<p tags="language--english language--spanish">No!</p>
+<p tags="language--french">Non!</p>
+```
+
+Alternatively you can also specify tags for a page individually in the page's `frontmatter`:
```
acjh: It's not clear whether the tags are used by the site to filter the page, or used by the page to filter elements.

How about: "Alternatively, you can also specify tags to render for a page in the front matter."

> "Alternatively`,` you can also specify tags `to render` for a page ~`individually`~ in the ~`page's`~ ~<code>\`</code>~ front` `matter ~<code>\`</code>~."

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
...
+
+```
+<frontmatter>
+  title: "Hello World"
+  tags: ["language--english"]
+</frontmatter>
+```
+
+Tags in the `frontmatter` will take precedence over the ones in `site.json`.
```
acjh: "~<code>\`</code>~front` `matter~<code>\`</code>~"

src/Page.js
```diff
@@ -52,6 +52,7 @@ function Page(pageConfig) {
   this.baseUrlMap = pageConfig.baseUrlMap;
   this.content = pageConfig.content || '';
   this.faviconUrl = pageConfig.faviconUrl;
+  this.tags = pageConfig.tags;
```
acjh: Alphabetical order.

src/Page.js
```diff
@@ -344,6 +345,28 @@ Page.prototype.concatenateHeadingsAndKeywords = function () {
   });
 };
 
+/**
+ * Filters out elements on the page based on config tags
+ * @param content of the page
+ * @param tags to filter
```
acjh: Use order of parameters.

test/test_site/testTagDivs.md
```diff
@@ -0,0 +1,24 @@
...
+<div tags="tag--shown-2">
+Div with shown-2 tag
+</div>
+
+# Div with multiple tags
+<div tags="tag--shown tag--other">
+Div with multiple tags
+</div>
+
+# Div with shown on overwrite tag
```
acjh: Override is the correct term, as used in **Page.js** [above](https://github.com/MarkBind/markbind/pull/467/files#diff-6e6e830d57ea503d762ce14927385fc6R426):

> ```js
> // Included tags specified in front matter will override included tags specified in site.json
> ```

See: https://stackoverflow.com/questions/8651562/overwrite-or-override

acjh:
Agreed, `site.json` should take precedence. I guess this followed Layouts, which also needs this change.

docs/userGuide/usingComponents.md
jamos-tay: Sorry what's this one again? It's rendering like this for me:
image

jamos-tay:
Sure - can change the order.

docs/userGuide/usingComponents.md
acjh: Correct, edited.

test/test_site/testTagDivs.md
```diff
@@ -0,0 +1,24 @@
...
+<div tags="tag--shown-2">
+Div with shown-2 tag
+</div>
+
+# Div with multiple tags
+<div tags="tag--shown tag--other">
+Div with multiple tags
+</div>
+
+# Div with should be overriden tag
```
acjh: overrid`d`en

acjh: should-be-overridden

jamos-tay: Updated

test/test_site/testEmptyTags.md
```diff
@@ -0,0 +1,8 @@
+<frontmatter>
+title: Test Layouts
+tags : []
```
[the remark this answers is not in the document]
jamos-tay: Yeah probably left over from when we changed the overwrite behavior, removed

docs/userGuide/usingComponents.md
```diff
@@ -205,4 +205,47 @@ If you want to include a keyword without rendering it to the page, add `d-none`
 This is good for catching <span class="keyword">regressions</span>
 ```
 
+# Using Tags
```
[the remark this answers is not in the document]
acjh: Keywords and Tags aren't Components though?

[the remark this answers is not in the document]
acjh: `### Keywords` could be changed to `#### Keywords` under `### Use Searchbar`.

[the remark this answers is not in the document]
yamgent: > Keywords and Tags aren't Components though?

That is true, an alternative would be to move it to under `contentAuthoring.md`.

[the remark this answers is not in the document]
jamos-tay: > `### Keywords` could be changed to `#### Keywords` under `### Use Searchbar`.

I think this is the best solution, it's a little strange to put it in content authoring as that page contains more general info.

jamos-tay:
Updated

test/test_site/expected/siteData.json
```diff
@@ -72,6 +77,22 @@
...
+        "front-matter-should-be-overridden-by-site.json": "Front matter should be overridden by site.json",
+        "div-with-shown-tag": "Div with shown tag",
+        "div-with-shown-2-tag": "Div with shown-2 tag",
+        "div-with-multiple-tags": "Div with multiple tags",
+        "div-with-should-be-overridden-tag": "Div with should be overridden tag",
+        "div-with-hidden-tag": "Div with hidden tag"
+      },
+      "title": "Hello World",
+      "tags": [
+        "tag--should-be-overridden"
```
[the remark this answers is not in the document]
jamos-tay: Ah I see the problem - I overwrote the tags but the front matter gets written to `siteData.json` directly, so while the page renders correctly the site data is wrong. Should be fixed with update.

yamgent:
@jamos-tay Due to recent restructuring, the documentation needs to be updated.

jamos-tay:
Updated docs, moved tags doc to 'Tweaking the page structure'

damithc:
Some suggestions regarding docs:
* Maybe rename the heading to `Toggling alternative contents in a page`? It's better to state the intention rather than the mechanism as readers wouldn't know that tags can be used for until they read this section.
* The section can be cross-listed under `Creating contents` -> `Creating content variations` i.e., give a brief explanation and give link.
* I wonder if using languages (English, French, ...) as an example would cause people to think of this as a mechanism for internationalization? There are different (beter) mechanisms for that, right?
* Use the `Example` label as done in the sections above, for consistency?

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,61 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+You need to specify the tags to include in `site.json`, under the `tags` option:
+
+```
+{
+  ...
+  "tags": ["language--java"]
+}
+```
+
+During rendering, only elements that match tags specified in the `site.json` files will be rendered. All other tagged elements will be filtered out. In this case, only the element with the `language--java` tag will be rendered. This is helpful when creating multiple versions of a page without having to maintain separate copies. If this option is not specified, all tagged elements will be rendered.
```
yamgent: `If this option is not specified...` should be a paragraph on its own:

```markdown
If the `tags` option is not specified in `site.json`, all tagged elements will be rendered.
```

docs/userGuide/reusingContents.md
```diff
@@ -483,4 +483,8 @@ Then, in a page-specific CSS file,
 }
 ```
 
+#### Creating similar variations of a site
+
+Tags are a good way to create multiple variations of a page within the same markdown file, for example filtering content to create different views. See [Toggling alternative contents in a page]({{ baseUrl }}/userGuide/tweakingThePageStructure.html#toggling-alternative-contents-in-a-page) for more information.
```
yamgent: 1. Rephrased the sentence such that the terms used are consistent throughout the documentation.

```markdown
... within the same source file, such as to filter content for creating multiple different versions of the same page. See [_User Guide: Tweaking the Page Structure → Toggling alternative contents in a page]({{ baseUrl }}/userGuide/tweakingThePageStructure.html#toggling-alternative-contents-in-a-page_) section for more information.
```

docs/userGuide/siteConfiguration.md
```diff
@@ -114,6 +115,10 @@ The ignore pattern follows the [glob pattern used in .gitignore](https://git-scm
   The branch that will be deployed to in the remote repo.
 </div>
 
+#### **`tags`**
+
+**A list of tags to filter page elements.** Do not specify this option if you want to show everything. See [Using Tags](tweakingThePageStructure.html#using-tags) for more information.
```
yamgent: 1. Expanded a bit more details to the user on how the tags work.

```markdown
**A list of tags to filter page elements.**  Page elements with the specified tags are retained, while elements tagged with other tags are removed. Do not specify this option if you want to show everything. See [_User Guide: Tweaking the Page Structure → Using Tags_](tweakingThePageStructure.html#using-tags) section for more information.
```

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,61 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+{{ icon_example }} Attaching multiple tags to an element:
+```html
+# For loops
+
+<p tags="language--java language--C#">for (int i = 0; i < 5; i++) { ... }</p>
+<span tags="language--python">for i in range(0,5): ...</span>
+```
+</div>
+
+Alternatively, you can also specify tags to render for a page in the front matter.
```
yamgent: 1. "also" is not needed when "Alternatively" is already used.

```markdown
Alternatively, you can specify tags to render for a page in the front matter.
```

src/Page.js
```diff
@@ -344,6 +345,28 @@ Page.prototype.concatenateHeadingsAndKeywords = function () {
...
 
+/**
+ * Filters out elements on the page based on config tags
+ * @param tags to filter
+ * @param content of the page
+ */
+Page.prototype.filterTags = function (tags, content) {
+  if (tags === undefined) {
+    return content;
+  }
```
yamgent: I wonder if we should warn the user if the tags field that they used is not a list (or help the user convert the single value to a list):

```
<frontmatter>
  tags: "language--C#"
</frontmatter>
```

```
error: TypeError: tags.forEach is not a function
warning: Error: Error while generating C:\Users\Tan Wang Leng\Desktop\New folder\index.md
```

jamos-tay: Sure, I can wrap it in an array if it isn't one

jamos-tay: I've tested this manually and it works - not sure how to add an automated test for it since the tags get overwritten.

src/Page.js
```diff
@@ -344,6 +348,31 @@ Page.prototype.concatenateHeadingsAndKeywords = function () {
...
+  if (!_.isArray(tags)) {
+    return this.filterTags([tags], content);
+  }
```
acjh: - use `Array.isArray`.
- use a variable instead of recursion? similar to `Site.prototype._rebuildAffectedSourceFiles`.
  ```js
  const tagsArray = Array.isArray(tags) ? tags : [tags];
  ```

test/test_site/expected/testTags.html
```diff
@@ -0,0 +1,59 @@
...
+    <div tags="tag--shown-2">
+      Div with shown-2 tag
+    </div>
+    <h1 id="div-with-multiple-tags">Div with multiple tags<a class="fa fa-anchor" href="#div-with-multiple-tags"></a></h1>
+    <div tags="tag--shown tag--other">
+      Div with multiple tags
+    </div>
+    <h1 id="div-with-should-be-overridden-tag">Div with should be overridden tag<a class="fa fa-anchor" href="#div-with-should-be-overridden-tag"></a></h1>
+
+    <h1 id="div-with-hidden-tag">Div with hidden tag<a class="fa fa-anchor" href="#div-with-hidden-tag"></a></h1>
```
acjh: - Why are `div-with-should-be-overridden-tag` and `div-with-hidden-tag` rendered?
- What's overridden?

src/Page.js
yamgent: > I've tested this manually and it works - not sure how to add an automated test for it since the tags get overwritten.

I guess we are hitting the limit of having only one `test_site`. Having multiple `test_site` would allow us to test different combinations.

I may just open a new issue for it, no need to worry about it for now.

jamos-tay:
Updated

test/test_site/expected/testTags.html
acjh: Ah, thanks for the clarification!

docs/userGuide/reusingContents.md
```diff
@@ -483,4 +483,7 @@ Then, in a page-specific CSS file,
 }
 ```
 
+#### Creating similar variations of a site
```
yamgent: `#### Creating multiple variations of a page`.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+
+<div class="indented">
+
+{{ icon_example }} Attaching tags to elements:
+```html
+# Print 'Hello world'
+
+<p tags="language--java">System.out.println("Hello world");</p>
+<div tags="language--C#">Console.WriteLine("Hello world");</div>
+<span tags="language--python">print("Hello world")</span>
```
yamgent: Is there a reason why we use different HTML tags for different languages: `<p>`, `<div>` and `<span>`? It doesn't come across naturally in terms of what an author would do, since there's no reason to put code snippets of different languages in different types of HTML tags.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+```
+</div>
+
+You need to specify the tags to include in `site.json`, under the `tags` option:
```
yamgent: The indentation of this should be aligned with the example (put it inside `<div class="indented">`)

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+
+```
+{
+  ...
+  "tags": ["language--java"]
+}
+```
+
+During rendering, only elements that match tags specified in the `site.json` files will be rendered. All other tagged elements will be filtered out. In this case, only the element with the `language--java` tag will be rendered. This is helpful when creating multiple versions of a page without having to maintain separate copies.
```
yamgent: The first statement should probably come right after the statement `Tags are specified by the `tags` attribute...`, since at this portion of the text, it should be explaining the example, not giving additional comments about the feature.

The second statement should also be indented together with the example.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+**You can also use multiple tags in a single HTML element. Specify each tag in the `tags` attribute** separated by a space. An element will be rendered if **any of the tags** matches the one in `site.json`.
+
+<div class="indented">
+
+{{ icon_example }} Attaching multiple tags to an element:
+```html
+# For loops
+
+<p tags="language--java language--C#">for (int i = 0; i < 5; i++) { ... }</p>
+<span tags="language--python">for i in range(0,5): ...</span>
```
yamgent: 1. Explain that as long as either `language--java` or `language--C#` is used, the code snippet will render.
1. I don't find the python example very useful here. Remove it?

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+```
+</div>
+
+Alternatively, you can specify tags to render for a page in the front matter.
+
+<div class="indented">
+
+{{ icon_example }} Specifying tags in front matter:
+```
```
yamgent: Format it as html.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -306,4 +306,63 @@ In this case `myHead.md` will override the `chapterLayout/head.md`.
...
+
+<p tags="language--java">System.out.println("Hello world");</p>
+<div tags="language--C#">Console.WriteLine("Hello world");</div>
+<span tags="language--python">print("Hello world")</span>
+```
+</div>
+
+You need to specify the tags to include in `site.json`, under the `tags` option:
+
+```
```
yamgent: Format it as json.

docs/userGuide/tweakingThePageStructure.md
jamos-tay: It's supposed to demonstrate that the `tags` attribute can be on different types of elements, but I get how that can be a little confusing, will change

jamos-tay:
Updated

yamgent:
@jamos-tay can you rebase this? Thanks!

damithc:
@jamos-tay is it possible to override frontmatter tags of a specific page/glob from site.json? Or site.json tags can only apply to the entire site?

yamgent:
> @jamos-tay is it possible to override frontmatter tags of a specific page/glob from site.json? Or site.json tags can only apply to the entire site?

Currently it applies to the entire site only, no way to do it for a particular page. 

The current overriding mechanism is also rather aggressive atm: I realize that it is not true that we don't lose any functionality for not categorizing tags into different groups. Consider this scenario:

---

`helloWorld-fragment.mbdf`:

```html
<div tags="level--beginner">

This is how you do Hello World:

<code tags="lang-cs">Console.WriteLine("Hello World");</code>
<code tags="lang-java">System.out.println("Hello World");</code>

</div>

<div tags="level--advanced">

Let's do a refresher. Recall that to print anything, you would use <code tags="lang-cs">Console.WriteLine()</code><code tags="lang-cs">println()</code>.

<question>
How would you print the string "Hello World"?
  ...
</question>

</div>
```

`index-cs.mbd`:

```html
<frontmatter>
    tags: lang-cs
</frontmatter>

<include src="helloWorld-fragment.mbdf" />
```

`index-java.mbd`:

```html
<frontmatter>
    tags: lang-java
</frontmatter>

<include src="helloWorld-fragment.mbdf" />
```

I want two different websites: for beginner, and for advanced. Both of these websites will render `index-cs.mbd` and `index-java.mbd`. I have no way of specifying either `level--beginner` or `level--advanced` without accidentally overriding the `lang--*` tags.

damithc:
@jamos-tay perhaps we can do another iteration for this feature to iron out these rough edges?

jamos-tay:
Hmm... I think what we can do is a tag merge, i.e. instead of overwriting the frontmatter tags completely we merge them instead. We probably should have done it this way earlier.

damithc:
> @jamos-tay is it possible to override frontmatter tags of a specific page/glob from site.json? Or site.json tags can only apply to the entire site?

@jamos-tay don't forget this one too.

In addition, should we also have a way to specify tags at the point of inclusion, the same way we allow overriding variables at the point of inclusion?

Issue #449, opened 2018-10-06
Support a way to include only if the target exists


damithc:
Current: if the target for an `include` is not found, it is shown as an error.

Suggested: give a way to include only if the target exists i.e., ignore if the target does not exist e.g., via an `if-exists` attribute

Justification: It will get rid of the need to have empty files/divs when the include is not applicable. For example, in the se-book, not all topics have an exercises.md or a prerequisites div. This is particularly useful for boilerplates as the same boilerplate can be used for many places.

The downside is that it can make bugs go unnoticed e.g., if the file actually exists but the include URL has a typo

PR #468, opened 2018-12-12
Support optional include

Xenonym:

Implements the `optional` attribute for `<include>` as suggested in #449 (originally suggested as `if-exists`).

The `optional` attribute will allow authors to include documents only if they exist, simplifying the use of boilerplate that have sections not used by every document using the boilerplate.

The `_preprocess` method in the parser was modified to only emit an error if the document does not exist and the document was included *without* the `optional` attribute. If a document included with `optional` does not exist, an empty `<div>` is emitted in its place with no errors.

```html
<include src="exercises.md" optional/> 
```

**Testing instructions:**
1. If documents are included with the `optional` attribute, there will not be an error when the it does not exist and the page should render as normal.

### DIFF
--- a/docs/userGuide/includingContents.md
+++ b/docs/userGuide/includingContents.md
@@ -25,11 +25,14 @@
 
     - `inline` (optional): make the included result an inline element. (wrapped in `<span>` tag)
 
+    - `optional` (optional): include the document only if it exists.
+
     Examples:
     ```html
     <include src="EstablishingRequirements.md#preview" inline/>
     <include src="../common/RequirementsVsSystemSpecification.md"/>
     <include src="../index.html" />
+    <include src="UserStories.md" optional/>
     ```
 
     <br/>
--- a/lib/markbind/lib/parser.js
+++ b/lib/markbind/lib/parser.js
@@ -68,6 +68,11 @@ function createErrorNode(element, error) {
   return Object.assign(element, _.pick(errorElement, ['name', 'attribs', 'children']));
 }
 
+function createEmptyNode() {
+  const emptyElement = cheerio.parseHTML('<div></div>', true)[0];
+  return emptyElement;
+}
+
 function isText(element) {
   return element.type === 'text' || element.type === 'comment';
 }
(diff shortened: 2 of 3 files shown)


lib/markbind/lib/parser.js
```diff
@@ -131,7 +136,8 @@ Parser.prototype._preprocess = function (node, context, config) {
       actualFilePath = calculateBoilerplateFilePath(element.attribs.boilerplate, filePath, config);
       this.boilerplateIncludeSrc.push({ from: context.cwf, to: actualFilePath });
     }
-    if (!utils.fileExists(actualFilePath)) {
+    const isOptional = element.name === 'include' && _.hasIn(element.attribs, 'if-exists');
+    if (!utils.fileExists(actualFilePath) && !isOptional) {
```
yamgent: Since this portion of the entire code is dedicated to handle missing files, the logic for creating an empty node for an optional include can be handled directly here, instead of being handled separately in line 153, 157-160:

```javascript
if (!utils.fileExists(actualFilePath) {
  if (isOptional) {
    return createEmptyNode();
  }
  
  this.missingIncludeSrc.push({ from: context.cwf, to: actualFilePath });
  ...
}
```

docs/userGuide/includingContents.md
```diff
@@ -25,8 +25,11 @@
 
     - `inline` (optional): make the included result an inline element. (wrapped in `<span>` tag)
 
+    - `if-exists` (optional): include the document only if it exists.
+
     Examples:
     ```html
+    <include src="exercises.md" if-exists/>
```
yamgent: Put it below all the other existing examples (after `<include src="../index.html" />`); it is not a typical use case so it shouldn't be at the top.

test/unit/parser.test.js
```diff
@@ -55,6 +55,39 @@ test('includeFile replaces <include> with <div>', async () => {
   expect(result).toEqual(expected);
 });
 
+test('includeFile replaces <include if-exists> with empty <div> if file does not exist', async () => {
```
yamgent: `'includeFile replaces <include src="doesNotExist.md" if-exists> with empty <div>'` makes the intention clearer.

yamgent:
On a side note, seems like the keyword `optional` would be much more familiar term than `if-exists` for this use case. @damithc

damithc:
> On a side note, seems like the keyword `optional` would be much more familiar term than `if-exists` for this use case.

That sounds good too. 👍

Xenonym:
Thanks for the comments @yamgent! Have made the necessary changes.

docs/userGuide/includingContents.md
```diff
@@ -25,11 +25,14 @@
...
     - `inline` (optional): make the included result an inline element. (wrapped in `<span>` tag)
 
+    - `optional` (optional): include the document only if it exists.
+
     Examples:
     ```html
     <include src="EstablishingRequirements.md#preview" inline/>
     <include src="../common/RequirementsVsSystemSpecification.md"/>
     <include src="../index.html" />
+    <include src="exercises.md" optional/>
```
yamgent: 2. For the sake of consistency with the other examples, let's use a filename that suits the overall theme (e.g. `UserStories.md`).

yamgent:
Also you can consider squashing your commits. I think one single commit will do.

Xenonym:
@yamgent thanks for the suggestion! Test added, docs updated and commits squashed.

PR #475, opened 2018-12-18
Add serve --one-page option to live preview a single page

### LINKED ISSUE #441: Allow live previewing a single page in force-rebuild mode
Is it possible to allow launching `markbind serve` for a specific page but in force-rebuild mode?

Justification: faster live preview when the author is interested in only one page but wants all changes (including nunjucks) to be reflected in that page

Xenonym:

Resolves #441.

Authors may want to preview only a single page in their website, that also includes changes from other parts of the website (eg. changes from an included page). We can allow them the option of just rendering that one page itself without having to render the entire website on each change, speeding up preview times.

I added a `--one-page <file>` option to the CLI that allows an author to specify a single page to preview. `Site()` and `Site.js#generatePages()` was then modified to only render that page when a page has been specified via `--one-page`. Finally, `live-server` was configured to open the specified page directly.

**Testing instructions:**
1. Choose a specific page in a Markbind website to serve with `--one-page` eg. `markbind serve --one-page userGuide/components.md`.
2. Note that only `_site/userGuide/components.html` and its required files are rendered.
3. Note that the browser opens to that specific page.
4. Note that changes to the site are detected and rendered, including those not in `components.md`.

### DIFF
--- a/docs/userGuide/developingASite.md
+++ b/docs/userGuide/developingASite.md
@@ -40,6 +40,7 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 | `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
 | `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
+| `--one-page <file>` | Render and serve only a single page from your website. |
 | --no-open | Don't open browser automatically. |
 
 Note: Live reload is only supported for the following file types:
--- a/index.js
+++ b/index.js
@@ -130,13 +130,20 @@ program
   .option('-f, --force-reload', 'force a full reload of all site files when a file is changed')
   .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
   .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
+  .option('--one-page <file>', 'render and serve only a single page in the site')
   .option('--no-open', 'do not automatically open the site in browser')
   .action((root, options) => {
     const rootFolder = path.resolve(root || process.cwd());
     const logsFolder = path.join(rootFolder, '_markbind/logs');
     const outputFolder = path.join(rootFolder, '_site');
 
-    const site = new Site(rootFolder, outputFolder, options.forceReload, options.siteConfig);
+    if (options.onePage) {
+      // replace slashes for paths on Windows
+      // eslint-disable-next-line no-param-reassign
+      options.onePage = fsUtil.ensurePosix(options.onePage);
+    }
+
+    const site = new Site(rootFolder, outputFolder, options.onePage, options.forceReload, options.siteConfig);
 
     const addHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
@@ -176,7 +183,7 @@ program
 
     // server config
     const serverConfig = {
-      open: options.open,
+      open: options.open && options.onePage ? `/${options.onePage.replace(/\.(md|mbd)$/, '.html')}` : false,
       logLevel: 0,
       root: outputFolder,
       port: options.port || 8080,
--- a/docs/userGuide/userQuickStart.md
+++ b/docs/userGuide/userQuickStart.md
@@ -105,6 +105,7 @@ Live reload is enabled to regenerate the site for changes, so you could see the 
 | `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
 | `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
+| `--one-page <file>` | Render and serve only a single page from your website. |
 | --no-open | Don't open browser automatically. |
 
 ### Build the static site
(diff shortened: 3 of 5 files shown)


index.js
```diff
@@ -130,13 +130,21 @@ program
...
     const rootFolder = path.resolve(root || process.cwd());
     const logsFolder = path.join(rootFolder, '_markbind/logs');
     const outputFolder = path.join(rootFolder, '_site');
+    const forceReload = options.onePage || options.forceReload;
 
-    const site = new Site(rootFolder, outputFolder, options.forceReload, options.siteConfig);
+    if (options.onePage) {
+      // replace slashes for paths on Windows
+      // eslint-disable-next-line no-param-reassign
+      options.onePage = options.onePage.replace('\\', '/');
```
yamgent: On Windows nodejs supports both `/` and `\`. Is there any particular reason why this was done?

index.js
```diff
@@ -130,13 +130,21 @@ program
   .option('-f, --force-reload', 'force a full reload of all site files when a file is changed')
   .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
   .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
+  .option('--one-page <file>', 'render and serve only a single page in the site')
   .option('--no-open', 'do not automatically open the site in browser')
   .action((root, options) => {
     const rootFolder = path.resolve(root || process.cwd());
     const logsFolder = path.join(rootFolder, '_markbind/logs');
     const outputFolder = path.join(rootFolder, '_site');
+    const forceReload = options.onePage || options.forceReload;
```
yamgent: Why is the `forceReload` needed in a one page scenario?

docs/userGuide/developingASite.md
```diff
@@ -40,6 +40,7 @@ MarkBind will generate the site in a folder named `_site` in the current directo
 | `-f`, `--force-reload` | Force a full reload of all site files when a file is changed. |
 | `-p`, `--port <port>` | The port used for serving your website. |
 | `-s`, `--site-config <file>` | Specify the site config file (default: site.json) |
+| `--one-page <file>` | Render and serve only a single page from your website. |
```
yamgent: `userQuickStart.md` also has a table that needs updating too.

lib/Site.js
```diff
@@ -126,6 +126,7 @@ function Site(rootPath, outputPath, forceReload = false, siteConfigPath = SITE_C
   this.siteConfig = {};
   this.siteConfigPath = siteConfigPath;
   this.userDefinedVariablesMap = {};
+  this.onePage = onePage;
```
yamgent: * should be in alphabetical order
* `onePagePath` rather than just `onePage`?

index.js
Xenonym: On PowerShell and `cmd.exe`, paths are normally written with `\`. However, we internally normalise all paths to use `/` as the separator (including when storing the value of `pages.src` in `siteData.json`). Since the value passed to `--one-path` is directly used to look up which page to serve, I believe it would be better to allow users to enter `\` as the path separator when specifying the file path to `--one-page`, which would be less confusing than if an error was emitted when `\` is used.

damithc:
Good job tackling this feature request @Xenonym 
Is the rendering faster when this feature is used on big sites such as https://github.com/se-edu/se-book ?

Xenonym:
No problem @damithc! 

Initial rendering times of [se-edu/se-book](https://github.com/se-edu/se-book) averaged over 5 runs:
`markbind serve`: 59.872s
`markbind serve --one-page gitAndGithub/index.md`: 14.115s

On average, there should be ~4x speedup during the initial render. 
Changes made during the live preview when using `serve --one-page` will pretty much be instantaneous compared to just using `serve`.

damithc:
> On average, there should be ~4x speedup during the initial render.
> Changes made during the live preview when using `serve --one-page` will pretty much be instantaneous compared to just using `serve`.

Nice. Looking forward to using this feature. 👍

index.js
Xenonym: Replacing `options.onePage.replace('\\', '/')` with `path.normalize()` actually breaks the lookup. This is because the default `site.json` uses paths with `/`. While we use `path.` functions for filesystem operations, `page.src` is still written with `/` into `siteData.json`. 

`path.normalize()` on a Windows system converts all path seperators to `\`, which breaks the lookup since pages are located by `page.src`, which needs `/` to match.

acjh: How about `path.posix.normalize`?

Xenonym: @acjh thank you for the suggestion! Unfortunately, `path.posix.normalize` on Windows-style paths does nothing:
```js
> path.posix.normalize('userGuide\\userQuickStart.md')
'userGuide\\userQuickStart.md'
```
This is because, according to the [Node.js docs for `path.normalize`](https://nodejs.org/docs/latest-v8.x/api/path.html#path_path_normalize_path), only `/` is considered a path separator under POSIX, and thus `\` will not be replaced. 

(The phrasing is clearer in the original PR for this section nodejs/node#12700, but was subsequently changed.)

acjh: Understood. Let's add a function in utils similar to ensure-posix-path's [`ensurePosix`](https://github.com/stefanpenner/ensure-posix-path/blob/master/index.js).

Xenonym: @acjh Sure, that will work!

Xenonym:
Glad to hear it @damithc! :D

PR #489, opened 2018-12-22
Swap layout override priority

jamos-tay:

• [X] Other, please explain: Changing behavior of existing feature

Follow up from #467 

Swap the priority of loading layouts, `site.json` takes precedence over front matter instead of the other way around.

Added test to match.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -399,8 +399,8 @@ Page.prototype.collectFrontMatter = function (includedPage) {
     this.frontMatter.src = this.src;
     // Title specified in site.json will override title specified in front matter
     this.frontMatter.title = (this.title || this.frontMatter.title || '');
-    // Layout specified in front matter will override layout specified in site.json
-    this.frontMatter.layout = (this.frontMatter.layout || this.layout || '');
+    // Layout specified in site.json will override layout specified in the front matter
+    this.frontMatter.layout = (this.layout || this.frontMatter.layout || LAYOUT_DEFAULT_NAME);
   } else {
     // Page is addressable but no front matter specified
     this.frontMatter = {
--- a/test/test_site/site.json
+++ b/test/test_site/site.json
@@ -5,8 +5,7 @@
   "pages": [
     {
       "src": "index.md",
-      "title": "Hello World",
-      "layout": "default"
+      "title": "Hello World"
     },
     {
       "src": "testEmptyFrontmatter.md",
--- a/test/test_site/site.json
+++ b/test/test_site/site.json
@@ -6,7 +6,7 @@
     {
       "src": "index.md",
       "title": "Hello World",
-      "layout": "testLayout"
+      "layout": "default"
     },
     {
       "src": "testEmptyFrontmatter.md",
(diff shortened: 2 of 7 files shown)


acjh:
I believe that was done to customize the layout for a page. The change still has to support that.

jamos-tay:
How should it be specified though? I'm worried while this change might be more 'correct' but we end up lose some functionality.

If the user wants to customize the page then he should probably not use globs in the site.json

src/Page.js
```diff
@@ -398,8 +398,8 @@ Page.prototype.collectFrontMatter = function (includedPage) {
     this.frontMatter.src = this.src;
     // Title specified in site.json will override title specified in front matter
     this.frontMatter.title = (this.title || this.frontMatter.title || '');
-    // Layout specified in front matter will override layout specified in site.json
-    this.frontMatter.layout = (this.frontMatter.layout || this.layout || '');
+    // Layout specified in site.json will override layout specified in the front matter
+    this.layout = (this.layout || this.frontMatter.layout || LAYOUT_DEFAULT_NAME);
```
yamgent: It can still be stored in `this.frontMatter.layout` (see `this.frontMatter.title` two lines above).

jamos-tay:
Updated

test/test_site/testLayouts.md
```diff
@@ -1,6 +1,7 @@
 <frontmatter>
 title: Test Layouts
 head: overwriteLayoutHead.md
+layout: testLayout
 </frontmatter>
 
-# Uses a layout
\ No newline at end of file
+# Uses a front matter layout
```
yamgent: In `site.json`, no layout is specified.
In the frontmatter, `testLayout` is specified.
The final output still uses `default` instead of `testLayout`.

That doesn't seem right. 😕

jamos-tay: My bad, fixed it, there was some issue with the glob layouts

jamos-tay:
Updated and Rebased

yamgent:
@jamos-tay The documentation needs to be updated.

jamos-tay:
Updated and rebased on master

jamos-tay:
Squashed and rebased

test/test_site/site.json
```diff
@@ -6,7 +6,7 @@
     {
       "src": "index.md",
       "title": "Hello World",
-      "layout": "testLayout"
+      "layout": "default"
```
acjh: should we specify this? I think `"default"` should be the *default*.

nicholaschuayunzhi:
I have a diff when doing file compare for `siteData.json`. Red is expected, and green is what markbind generated.
```diff
{
      "headings": {
- "uses-a-site.json-layout%2C-overriding-front-matter": "Uses a site.json layout, overriding front matter"
+  "uses-a-site-json-layout-overriding-front-matter": "Uses a site.json layout, overriding front matter"
      },
      "title": "Hello World",
      "head": "overwriteLayoutHead.md",
      "layout": "testLayout",
      "src": "testLayoutsOverride.md"
    },
...

Want to clarify if the line in the expected file is correct.

acjh:
I believe expected needs to be updated for #578.

yamgent:
@nicholaschuayunzhi thanks for spotting the error, my apologies for the error, I didn't re-run the tests when merging this PR.

The fix is now available on the master branch, in commit f02c338.

PR #490, opened 2018-12-23
Update error message for --one-page for cases where specified page is not in site.json

yamgent:

• [X] Other, please explain: Clarifying the error message

The page may exist on the file-system, but if it is not specified in `site.json`, it does not exist from MarkBind's point of view.

Updated the warning message to include that possibility.

NIL

NIL

**Testing instructions:**
`markbind serve docs --one-page userGuide/components/dropdown.md`


yamgent:
Updated PR.

PR #495, opened 2018-12-24
Plugins: Support shorthand syntax processing

### LINKED ISSUE #492: Provide a shorter syntax for specifying complicated panel headings
Current: a complicated panel/modal heading (i.e., when the heading cannot be specified using inline markdown as `heading="#Notes to remember")` can be specified using the following `slot` syntax:
```html
<panel>
    <span slot="header" class="card-title">
        Complicated content to be used as heading ...
    </span>
</panel>
```

Suggested: provide a shorter syntax e.g,
```html
<panel>
    <heading>
        Complicated content to be used as heading ...
    </heading>
</panel>
```

jamos-tay:

• [X] New feature

Fixes #492 

We want to be able to support shorthand syntax for certain elements.

Have a feature that transpiles shorthand syntax into proper Markbind syntax before processing.

The current implementation is just a proof of concept. Better to migrate this function to external plugin once #474  is merged.

We can generalize this plugin so other shorthands can be added easily as well. I'm thinking we can handle these all together since they are too small to warrant their own functions/plugins.

Something like this perhaps?
```
{
  'panel>span[heading]' : (ele) => // modify ... ,
}
```

### DIFF
--- a/docs/userGuide/usingPlugins.md
+++ b/docs/userGuide/usingPlugins.md
@@ -152,6 +152,7 @@ MarkBind has a set of default plugins that it uses to carry out some of its feat
 Default Plugin | Functionality
 --- | ---
 `anchors` | Attaches anchor links to the side of headings.
+`shorthandSyntax` | Allows for certain syntax shorthands.
 
 Although not advised, you can disable these by passing `"off": true` in the `pluginsContext`.
 
--- a/src/plugins/default/markbind-plugin-shorthandSyntax.js
+++ b/src/plugins/default/markbind-plugin-shorthandSyntax.js
@@ -1,0 +1,22 @@
+const cheerio = module.parent.require('cheerio');
+
+// Convert panel headings: <span heading>
+function convertPanelHeadings($) {
+  $('panel>span[heading]').each((i, element) => {
+    $(element).attr('slot', 'header');
+    $(element).addClass('card-title');
+    $(element).removeAttr('heading');
+  });
+}
+
+/**
+ * Converts shorthand syntax to proper Markbind syntax
+ * @param content of the page
+ */
+module.exports = {
+  postRender: (content) => {
+    const $ = cheerio.load(content, { xmlMode: false });
+    convertPanelHeadings($);
+    return $.html();
+  },
+};
--- a/test/functional/test_site/expected/index.html
+++ b/test/functional/test_site/expected/index.html
@@ -402,6 +402,11 @@ specification that specifies how the product will address the requirements. </sp
           <div>
             <p>Should be blank:</p>
           </div>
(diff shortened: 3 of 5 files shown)


jamos-tay:
Oh yeah, js plugin definitely. I'm just suggesting we have some kind of framework that can be generalized to different syntax transformations.

yamgent:
> I'm just suggesting we have some kind of framework that can be generalized to different syntax transformations.

Understand the point, my question originated because you mentioned the following:

> We can generalize this plugin so other shorthands can be added easily as well. I'm thinking we can handle these all together since they are too small to warrant their own functions/plugins.

Will look forward to your actual implementation.

jamos-tay:
Reviving this PR now that #474 is merged.

Moved to `shorthandSyntax` plugin.

---

This plugin is kind of empty right now. I'm not sure if it's worth exposing to public (documenting) at the moment. How about:

Add a mechanism for 'always-run' plugins i.e. without having to include it in `plugins`. The way I see it plugins are good for two things:
- Allowing users to enable/disable features
- Keeping the code base clean

This was made as a plugin more for the sake of keeping the code base clean, users are unlikely to need to disable it. Having it always on saves the user from having to add it in every time.

This can also be applied to other features that should be always enabled, such as
- Anchors
- Reading page variables from outer (#682)
- Popovers (#551)

jamos-tay:
Waiting on #702 so this can be converted to always-on

jamos-tay:
Sorry for late update. Plugin has been converted to default (always on) and rebased.

yamgent:
Propose a merge commit message.

jamos-tay:
How's this?

```
Plugins: Shorthand syntax processing

Certain MarkBind elements such as panels require a lot of boilerplate 
HTML to be written in the .md file, which is tedious. We would like to 
have syntax shorthands for such element to make it easier for the 
site author.

Let's create a plugin that can convert shorthand syntaxes into proper 
MarkBind syntax simply, which serves as a framework for more of these
to be added.
```

yamgent:
Revised your commit message to use `<span heading>` as an example:

```
Plugins: Support shorthand syntax processing (#495)

Certain MarkBind elements such as panels require a lot of boilerplate 
HTML to be written in the .md file, which is tedious. We would like to 
have syntax shorthands for such element to make it easier for the 
site author, so that instead of writing:

    <panel>
        <span slot="header" class="card-title">
            Heading
        </span>
    </panel>

Authors can write this instead to achieve the same effect:

    <panel>
        <span heading>
            Heading
        </span>
    </panel>

Let's create a plugin that can convert shorthand syntaxes into proper 
MarkBind syntax simply, which serves as a framework for more of these
to be added.
```

PR #517, opened 2019-01-01
Support optional includes for page fragments

### LINKED ISSUE #497: Extend 'optional include' functionality to page fragments
Current: #468 allows `<include src="foo.mbd" optional />` to fail sliently if `foo.mbd` doesn't exist.

Suggestion: Allow `<include src="foo.mbd#bar" optional />` to fail sliently if `bar` doesn't exist in `foo.mbd`

Justification: This remove the need to have empty divs/spans in files when the div doesn't apply.

Xenonym:

Resolves #497.

Currently, we allow optional inclusion of whole documents with the `optional` tag eg. 
`<include src="doc.md" optional />`. This PR extends this functionality to page fragments as well eg.
`<include src="doc.md#segment" optional />` will be included only if `#segment` exists.

When a page fragment that is included does not exist, we check first if the `optional` tag is used. If so, we include an empty string in place, leading to an empty `<div>`. If not, an error is emitted that the page fragment included does not exist. 

```html
<include src="doc.md#exists" optional />
<include src="doc.md#doesNotExist" optional />
<include src="doc.md#doesNotExist" />
```

In previous versions, inclusion of nonexistent page fragments lead to a `<div>` with the string `null` being included instead. Since this behavior is likely to be confusing with the addition of `optional`, I have chosen to change this so that an error is emitted when an included page fragment does not exist and `optional` was not used. I have also added tests to verify this new behavior alongside those that verify the behavior of `optional` with page fragments.

**Testing instructions:**
1. Include a page fragment with `optional` eg. `<include src="doc.md#exists" optional />`.  The content of the page fragment should be included.
2. Include a nonexistent page fragment with `optional` eg. 
`<include src="doc.md#doesNotExist" optional />`.  There should not be an error and the page should render as normal.
3. Include a nonexistent page fragment normally eg. `<include src="doc.md#doesNotExist" />`.  An error should be emitted.

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -158,8 +158,8 @@ Parser.prototype._preprocess = function (node, context, config) {
     element.name = isInline ? 'span' : 'div';
     element.attribs[ATTRIB_INCLUDE_PATH] = filePath;
 
-    if (isOptional) {
-      // optional files have been handled earlier, so just delete the attribute here
+    if (isOptional && !includeSrc.hash) {
+      // optional includes of whole files have been handled, but segments still need to be processed
       delete element.attribs.optional;
     }
 
@@ -219,6 +219,26 @@ Parser.prototype._preprocess = function (node, context, config) {
       const $ = cheerio.load(segmentSrc);
       const htmlContent = $(includeSrc.hash).html();
       let actualContent = htmlContent;
+
+      if (actualContent === null) {
+        if (isOptional) {
+          // set empty content for optional segment include that does not exist
+          actualContent = '';
+        } else {
+          const error = new Error(
+            `No such segment '${includeSrc.hash.substring(1)}' in file: ${actualFilePath}`
+              + `\nMissing reference in ${element.attribs[ATTRIB_CWF]}`,
+          );
+          this._onError(error);
+          return createErrorNode(element, error);
+        }
+      }
+
+      if (isOptional) {
+        // optional includes of segments have now been handled, so delete the attribute
+        delete element.attribs.optional;
+      }
+
       if (isIncludeSrcMd) {
         if (context.mode === 'include') {
           actualContent = isInline ? actualContent : utils.wrapContent(actualContent, '\n\n', '\n');
--- a/test/unit/parser.test.js
+++ b/test/unit/parser.test.js
@@ -128,6 +128,87 @@ test('includeFile replaces <include src="doesNotExist.md" optional> with empty <
   expect(result).toEqual(expected);
 });
 
+test('includeFile replaces <include src="include.md#exists"> with <div>', async () => {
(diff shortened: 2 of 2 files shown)


src/lib/markbind/src/parser.js
```diff
@@ -219,6 +219,25 @@ Parser.prototype._preprocess = function (node, context, config) {
...
       const htmlContent = $(includeSrc.hash).html();
       let actualContent = htmlContent;
+
+      if (actualContent === null) {
+        if (isOptional) {
+          // set empty content for optional segment include that does not exist
+          actualContent = '';
+        } else {
+          const error = new Error(
+            `No such segment ${includeSrc.hash.substring(1)}`,
```
yamgent: 1. The included file should be mentioned to make it clearer.
2. The file referencing the included file should also be mentioned, because the error is printed on the console as well.

So the final error message should be something like this:
```javascript
          const error = new Error(
            `No such segment '${includeSrc.hash.substring(1)}' in file: ${actualFilePath}`
              + `\nMissing reference in ${element.attribs[ATTRIB_CWF]}`,
          );
```

Xenonym: Good point, I have changed the error message to be more detailed!

yamgent:
> This probably needs a slight update to docs?

Hmm, it didn't cross my mind for this feature, as we do not explicitly mention that other attributes such as `inline` and `trim` works for fragments as well in the current documentation, because it seems to be implied.

Although I agree that a minor mention might make it clearer and more obvious for the author. Maybe such a mention will work?

```diff
diff --git a/docs/userGuide/reusingContents.md b/docs/userGuide/reusingContents.md
index e9e957a..15dea5b 100644
--- a/docs/userGuide/reusingContents.md
+++ b/docs/userGuide/reusingContents.md
@@ -186,14 +186,14 @@ Attributes:
   <include src="frament.md" trim />
   \```
 
-**You can `<include>` a fragment of a file** by specifying the `#fragment-id` at the end of the `src` attribute value, provided the fragment is wrapped in a `<div>`/`<span>`/`<seg>` tag with the matching `id`.
+**You can `<include>` a fragment of a file** by specifying the `#fragment-id` at the end of the `src` attribute value, provided the fragment is wrapped in a `<div>`/`<span>`/`<seg>` tag with the matching `id`. You can use the attributes for documents (such as `inline` and `optional`) with fragments as well.
 
 <div class="indented">
 
 {{ icon_example }} Including a fragment from a file:
 \```html
 Some text
-<include src="docs/tips.md#tip-1" />
+<include src="docs/tips.md#tip-1" trim />
 Some other text
 \```
 ```

damithc:
Yes, it is awkward to mention it in the current doc. The problem is, fragment inclusion is mention _after_ the attributes. Let's leave it be.

PR #521, opened 2019-01-02
Allow empty <frontmatter> tags

### LINKED ISSUE #515: Allow empty <frontmatter> tags
Current: an empty `<frontmatter>` tag causes a build error.

Suggested: allow empty `<frontmatter>` tags

Xenonym:

Resolves #515.

Currently, having an empty `<frontmatter>` tag in pages causes a build error. We should just ignore empty `<frontmatter>` tags instead of failing to build the site.

The check for `<frontmatter>` tags was changed such that only non-empty `<frontmatter>` tags are processed. Empty `<frontmatter>` tags are ignored.

```html
<frontmatter></frontmatter>
```
**Testing instructions:**
1. Create a page with an empty `<frontmatter>` tag. The page should still be able to build.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -384,7 +384,7 @@ Page.prototype.collectIncludedFiles = function (dependencies) {
 Page.prototype.collectFrontMatter = function (includedPage) {
   const $ = cheerio.load(includedPage);
   const frontMatter = $('frontmatter');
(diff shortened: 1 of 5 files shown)


Xenonym:
@yamgent noted, test file created!

PR #525, opened 2019-01-03
Improve the default `site.json` settings generated during init

### LINKED ISSUE #507: Improve the default `site.json` settings generated during init
Justification: It lowers the initial barrier to entry as any folder with .md files can be served as a MarkBind site. The 'quick start' can omit mentioning `markbind init` and `site.json` altogether. 

If the site.json is missing, we can use defaults for all configuration options and consider all *.md and *.mbd files as pages.

### LINKED ISSUE #563: Prevent glob from building .md files in the _markbind folder
Using general glob patterns like  `**/*.md` can cause `.md` files in the `_markbind` folder such as `footer.md`, `head.md` to be selected as pages for building.

```
  "pages": [
    {
      "glob": "**/*.md"
    },
  ]
```

Should we add an exclusion for the `_markbind` folder?

amad-person:

Fixes #507.
Fixes #563.

As per #507, the default site.json generated during `markbind init` can be improved by serving all .md and .mbd files by default (and ignoring .mbd files and the .git directory). Hence users need to tweak `site.json` only when they require some additional setup, e.g. overriding a page title.

Updated the default `site.json` that is generated by the `markbind init` command as follows:
1. Added  `*.md` and `*.mbd` to the pages array so that all .md and .mbd files will be served by default.
2. Added `*.mbd` and `.git/*` to the ignore array so that these files will be ignored while copying over to the generated site's directory. 

Default site.json generated by `markbind init` after this change:

```
{
  "baseUrl": "",
  "titlePrefix": "",
  "ignore": [
    "_markbind/layouts/*",
    "_markbind/logs/*",
    "_site/*",
    "site.json",
    "*.md",
    "*.mbd",
    ".git/*"
  ],
  "pages": [
    {
      "src": "index.md",
      "title": "Hello World"
    },
    {
      "glob": "**/index.md"
    },
    {
      "glob": "**/*.+(md|mbd)"
    }
  ],
  "deploy": {
    "message": "Site Update."
  }
}
```

N.A.

**Testing instructions:**

Run the test suite / build markbind locally and test.


Xenonym:
1. Would it make sense to ignore `.gitignore` as well since we are already ignoring `.git/*`?
2. The glob patterns `*.md` and `*.mbd` would only match `.md` and `.mbd` files in the root directory of the site eg. `inside/doc.md` would not be matched and rendered. Is this the intended behavior?

amad-person:
Yeah...I think we can remove `.gitignore` since we're ignoring everything in `.git` already.
As for the second point - would updating the patterns to `**/*.md` and `**/*.mbd` work?

Xenonym:
@amad-person they *should*, but it doesn't seem to work in testing, see #526.

acjh:
Why are `.md` and `.mbd` in both `ignore` and `glob`?

amad-person:
@acjh from what I understand, the requirement is that all .md and .mbd files should be rendered as HTML files i.e. they should be in `glob`. But the original markdown files shouldn't be copied over to the generated site directory, so .md and .mbd are in `ignore` also.

Xenonym:
@amad-person `**/*.md` and `**/.*mbd` would work! In fact, you could combine them into one pattern: `**/*.+(md|mbd)`.

yamgent:
> However, this would also render all files in the `_markbind` directory eg. `_markbind/variables.md`, which I am not sure is useful. `{,!(_markbind)/**/}*.+(md|mbd)` _would_ exclude those files, but it causes the unit test to fail, so I suspect that's intended behaviour?

There's currently no way to exclude pages, and the current codebase doesn't explicitly exclude the `_markbind/` directory as well. So the behaviour is expected, but I don't think it is a desirable behaviour.

It should be possible to correct the behaviour in `Site.prototype.collectAddressablePages`, no need to resort to complicated regex for now. :P

acjh:
No, [boilerplate files are a feature in MarkBind](https://markbind.github.io/markbind/userGuide/includingContents.html#handling-repetitive-boilerplate-files) and that constant is used in other places.

src/Site.js
```diff
@@ -364,7 +365,7 @@ Site.prototype.collectAddressablePages = function () {
     globPages.concat(walkSync(this.rootPath, {
       directories: false,
       globs: [addressableGlob.glob],
-      ignore: [BOILERPLATE_FOLDER_NAME],
+      ignore: [BOILERPLATE_FOLDER_NAME, CONFIG_FOLDER_NAME],
```
acjh: Can remove `BOILERPLATE_FOLDER_NAME` since it's a subdirectory of `CONFIG_FOLDER_NAME`?

PR #550, opened 2019-01-05
Markbind help: update output to match userguide

### LINKED ISSUE #535: markbind help: update output to match userguide
As a result of the recent user guide update, the output of the `markbind help` command is slightly out of sync with the user guide  https://markbind-master.netlify.com/userguide/clicommands

Suggestions: update it to match the user guide.

JYL123:

• [X ] Enhancement to an existing feature

Fixes #535 

Update `index.js` command descriptions to match the updates in [user guide](https://markbind-master.netlify.com/userguide/clicommands). The modified command overview is as follows from terminal:

commands

**Testing instructions:**
Run `markbind -h` from terminal

### DIFF
--- a/docs/userGuide/cliCommands.md
+++ b/docs/userGuide/cliCommands.md
@@ -37,14 +37,14 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
 Generates the site to the directory named `_site` in the current directory.
 
 **`OPTIONS`:**
-* `<output>`<br>
+* `[output]`<br>
   Put the generated files in the specified directory<br>
   {{ icon_example }} `../myOutDir`
-* `<root> <output>`<br>
+* `[root] [output]`<br>
   Read source files from the `<root>` directory and put the generated files in the specified `<output>` directory<br>
   {{ icon_example }} `./myWebsite ../myOutDir`
 * `--baseUrl <base>`<br>
-  Override the `baseUrl` property (read from the `site.json`) with the give `<base>` value.<br>
+  Override the `baseUrl` property (read from the `site.json`) with the given `<base>` value.<br>
   {{ icon_example }} `--baseUrl staging`
 
 **{{ icon_examples }}**
@@ -57,7 +57,7 @@ Generates the site to the directory named `_site` in the current directory.
 
 **Format**: `markbind deploy`
 
-**Description**: Deploy the site to the repo's Github pages.
+**Description**: Deploy the website to the repo's GitHub pages.
 
 Deploys the site to the repo's Github pages by pushing everything in the generated site (default dir: `_site`) to the `gh-pages` branch of the current git working directory's remote repo.
 
@@ -69,12 +69,12 @@ Deploys the site to the repo's Github pages by pushing everything in the generat
 
 **Format:** `markbind init [<OPTIONS>]`
 
-**Description:** Init a markbind website project.
+**Description:** Init a MarkBind website project.
 
 Initializes a directory into a MarkBind site by creating a skeleton structure for the website which includes a `index.md` and a `site.json`.
 
 **`OPTIONS`:**
-* `<root>`<br>
+* `[root]`<br>
   Root directory. Default is the current directory.<br>
   {{ icon_example }} `./myWebsite`
 
@@ -88,7 +88,7 @@ Initializes a directory into a MarkBind site by creating a skeleton structure fo
 
 **Format:** `markbind serve [<OPTIONS>]`
 
-**Description:** Build then serve a website from a directory.
+**Description:** Serve a website from a directory.
 
 Does the following steps:
 1. Builds the site and puts the generated files in a directory named `_site`.
@@ -100,7 +100,7 @@ Does the following steps:
 </modal>
 
 **`OPTIONS`:**
-* `<root>`<br>
+* `[root]`<br>
   Root directory. Default is the current directory.<br>
   {{ icon_example }} `./myWebsite`
 * `-f`, `--force-reload`<br>
--- a/docs/userGuide/cliCommands.md
+++ b/docs/userGuide/cliCommands.md
@@ -30,18 +30,20 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
 
 ### `build` Command
 
-**Format:**: `markbind build [OPTIONS] `
+**Format**: `markbind build [<OPTIONS>] `
 
-**Description**: Generates the site to the directory named `_site` in the current directory.
+**Description**: Build a website.
 
-**`[OPTIONS]`:**
-* `[output]`<br>
+Generates the site to the directory named `_site` in the current directory.
+
+**`OPTIONS`:**
+* `<output>`<br>
   Put the generated files in the specified directory<br>
   {{ icon_example }} `../myOutDir`
-* `[root] [output]`<br>
+* `<root> <output>`<br>
   Read source files from the `<root>` directory and put the generated files in the specified `<output>` directory<br>
   {{ icon_example }} `./myWebsite ../myOutDir`
-* `--baseUrl [base]`<br>
+* `--baseUrl <base>`<br>
   Override the `baseUrl` property (read from the `site.json`) with the give `<base>` value.<br>
   {{ icon_example }} `--baseUrl staging`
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -52,10 +52,10 @@ program
 
 program
   .command('serve [root]')
-  .description('build then serve a website from a directory')
+  .description('build then serve a website from a directory, and open a live preview of the website')
   .option('-f, --force-reload', 'force a full reload of all site files when a file is changed')
-  .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
-  .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
+  .option('-p <port>, --port <port>', 'port for server to listen on (Default is 8080)')
+  .option('-s <file>, --site-config <file>', 'specify the site config file (default: site.json)')
```
Xenonym: This change breaks the options: `-p` and `-s` do not have an effect, and the long options `--port` and `--site-config` are not recognised.

The reason is because the library we are using for the CLI, [commander.js](https://github.com/tj/commander.js), parses the first argument to `.option` [as a list of flags](http://tj.github.io/commander.js/#Command.prototype.option):

> The `flags` string should contain both the short and long flags, separated by comma, a pipe or space.

I don't think we can make this specific change without breaking the CLI.

index.js
```diff
@@ -172,6 +172,7 @@ program
 
 program
   .command('build [root] [output]')
+  .option('<output>', 'put the generated files in the specified directory')
```
Xenonym: As with earlier, commander.js expects a list of flags as the first argument to `.option()`. Passing in anything else doesn't seem to have meaningful results.

index.js
JYL123: Hi @Xenonym, thanks for the explanation. I should have tested it myself. I got it now. Then I will remove my modification.

index.js
JYL123: Thanks for pointing out again. I will remove this line as well. For this issue, is there any other parts that i have missed which is supposed to be done?

Xenonym:
@JYL123 comparing between the [user guide](https://markbind-master.netlify.com/userguide/clicommands) and the output of `markbind <command> -h`, these are the differences I notice:
1. The commands are listed in the user guide in alphabetical order, but not so in `markbind`. We could alphabetise the commands in `index.js` to match the user guide.
2. Optional arguments are denoted with `[]` in `markbind` eg. `--baseUrl [baseUrl]` but marked with `<>` in the user guide. Since we cannot change the flag description without changing the behaviour of the CLI, we should probably change the user guide to match instead.

yamgent:
@Xenonym excellent review, good work. :+1:

@damithc just checking, how much of the matching were you looking for when you created #535? Were you expecting a close match like this:

```
$ markbind --help

  ...

  Commands:

  build [options] [root] [output]  Generates the site to the directory named _site
                                   in the current directory.
  deploy                           Deploys the site to the repo's Github pages by 
                                   pushing everything in the generated site (default 
                                   dir: _site) to the gh-pages branch of the current 
                                   git working directory's remote repo. 

  ...
```

Xenonym:
> It's better to be as close as possible, right? For example, use the same notation for options `markbind build [<OPTIONS>]`.

The description in the CLI is actually [parsed by commander.js](http://tj.github.io/commander.js/#Command.prototype.command) to determine what arguments the command takes. I don't think we can change it without breaking the commands.

damithc:
> The description in the CLI is actually [parsed by commander.js](http://tj.github.io/commander.js/#Command.prototype.command) to determine what arguments the command takes. I don't think we can change it without breaking the commands.

I see. I don't mind if the user guide is changed to match the format used by CLI, if that format is good enough. In the user guide, I was trying to follow the format used by Git documentation.

JYL123:
Hi, thanks for clarifying this issue. is it okay that I make the following changes:
1.  for user guide, refer to this [Git documentation](https://git-scm.com/docs/git), can I remove `<>` for `OPTIONS`, `root`, `output` and leave the rest as it is? because from [the documentation](https://git-scm.com/docs/git), I see `<>` is used for extra values added within `[]`.  So for example, my imagination for `build` is : 
```
build Command
Format: markbind build [OPTIONS]

OPTIONS:
* output
...
* root output
...
* --baseUrl <base>
...
```
 
2. for `index.js`, I will add `Description` from user guide for each command description, and add option descriptions from user guide, such as `<output>
Put the generated files in the specified directory`, in flag descriptions. 

 Please correct me if anything is wrong or not appropriate. Thanks in advance.

damithc:
> 1. because from [the documentation](https://git-scm.com/docs/git), I see `<>` is used for extra values added within `[]`.

It looks like `<>`  is used to indicate a placeholder where a value is expected. e.g., `<root>` means it it expects a value such as `c:\` whereas options without `<>` e.g., `--base-url` are keywords that should be repeated exactly as they are. What do you think?

damithc:
> Thanks prof. I see. It makes sense. Then I think the user guide is good. I will undo my change to the user guide, namely keep the user guide as follows:
> 
> ```
> Format:  markbind build [<OPTIONS>]
> ...
> OPTIONS:
> * <output>
> ...
> * <root> <output>
> ...
> ```
> And change the descriptions in `index.js` as the comment above.

Got it. I'll leave the senior devs to make the final decision w.r.t. the exact format to follow.

yamgent:
> 2. for `index.js`, I will add `Description` from user guide for each command description, and add option descriptions from user guide, such as `<output> Put the generated files in the specified directory`, in flag descriptions.

Is that actually possible with commander.js? I don't see a way of specifying the description for the compulsory arguments...

Also, I am in favour of such a format:

> Example: In the user guide, the first paragraph of the description is 1-2 sentence overview of the command/parameter that will be reused in the CLI help. Further details can be in subsequent paragraphs.

Basically, the help description in the CLI should be as short as possible (which is what we have currently). The user guide should also have the short description, but then it will be followed by the details for the command.

For example, the help description for `markbind build` should be a single sentence summary:

```
$ markbind build --help

  Usage: build [options] [root] [output]

  Builds the website.


  Options:
```

But the user guide should contain the single sentence summary + additional description of what it does:

---

## `build` Command
**Format:** `markbind build [<OPTIONS>]`

**Description:** Builds the website. Generates the site to the directory named _site in the current directory.

...
---

yamgent:
There are two other issues that I have spotted.

1. `<arg>`: compulsory argument. `[arg]` = optional argument. The user guide is not consistent on this front.
1. Options are different from arguments. Options have a flag (e.g. `-n` or `--no-open`). Arguments do not have a flag. The user guide is not making the distinction clear.

JYL123:
@yamgent Thanks for the review!

docs/userGuide/cliCommands.md
```diff
@@ -69,7 +69,7 @@ Deploys the site to the repo's Github pages by pushing everything in the generat
 
 **Format:** `markbind init [<OPTIONS>]`
 
-**Description:** Init a markbind website project.
+**Description:** Init a Markbind website project.
```
yamgent: The B needs to be capitalized as well, that is how we brand our project. (i.e. `MarkBind`, not `Markbind`)

docs/userGuide/cliCommands.md
```diff
@@ -105,16 +105,16 @@ Does the following steps:
   {{ icon_example }} `./myWebsite`
 * `-f`, `--force-reload`<br>
    Force live reload to process all files in the site, instead of just the relevant files. This option is useful when you are modifying a file that is not a file type monitored by the <trigger trigger="click" for="modal:cliCommands-livePreview">live preview</trigger> feature.
-* `-p <port>`, `--port <port>`<br>
+* `-n, --no-open`<br>
+   Don't open a live preview in the browser automatically.
+* `-o, --one-page [file]`<br>
+   Render and serve only a single page from your website.<br>
+  {{ icon_example }} `--one-page guide/index.md`
+* `-p [port]`, `--port [port]`<br>
```
yamgent: You got it the wrong way round. :P

Things like `[root]` are the optional arguments (users can omit `root` when using `markbind serve`). They should be wrapped with `[]`.

Things like `<port>` are the compulsory arguments (users must specify a port number when using `-p`). They should be wrapped with `<>`.

Most of the commands and options have this issue as well, please do rectify them, thanks!

PR #553, opened 2019-01-05
Allow specifying variables in include

### LINKED ISSUE #545: Allow customizing at the point of inclusion
Suggestied feature: the following code will include `foo.md` as per normal but will replace all spans/divs with id `title` in `foo.md` with `New Title` and `level` with `2`:
```html
<include src=”foo.md”>
  <span id=”title”>New Title</span>
  <span id=”level”>2</span>
</include>
```
`foo.md`:
```html
## <span id="title">Dummy Title</span>

This is a level <span id="level"/> topic.

The level of this topic is <span id="level"/>.
```
Rationale: this will greatly increase code reuse as it allows authors to write templates  and reuse them while customizing them at the point of reuse (somewhat similar to Jekyll tempates https://jekyllrb.com/docs/layouts/).

Points to consider: What about variations that cannot be captured in spans e.g., html attribute values? For example, if we want to use the following code fragment as a template but we want the background color to be confirguragle at the point of reuse

```html
   <box background-color="white" border-color="#cccccc"><md><span id="content"/></md></box>
```

jamos-tay:

Fixes #545 

Users may want to create `include` file templates with variables that can be filled in, allowing reuse. However, they should only apply to each file. For example,

```markdown
## {{ title }}

This is a level {{ level }} topic.
```

Allow users to specify variables in an `include` tag

Allow users to specify Markbind variables in an include tag:

```
<include src=”foo.md”>
  <span id="title"><span>New Title</span></span> // same syntax as variables.md
  <span id="level">2</span>
</include>
```

These add on to (and overwrite) variables in `_markbind/variables.md`, and only apply to the context of the included file, and any files that the file includes.

We use nunjuck's `or (default)` syntax to allow for default values in case the variable is missing: `{{ title or "Untitled" }}`

Users cannot reference other variables in the include, like this:

```
<include src=”foo.md”>
  <span id="inner">Inner</span>
  <span id="outer">Outer and {{ inner }}</span>
</include>
```

This is because `{{ inner }}` gets cleared when the original file is rendered. We can have a separate syntax for these, but it's a little unwieldly.

Referencing variables in `_markbind/variables.md` is okay.


damithc:
Does this work too? It's not related to includes but possibly related to how the feature is implemented. 
```html
<span id="color">red</span>

<font color="{{ color }}">my text</font>
```

damithc:
This PR affects the following statement, right?
image

docs/userGuide/reusingContents.md
```diff
@@ -101,6 +101,13 @@ As mentioned in _User Guide: MarkBind Syntax_, you can use built-in variables to
 <p/>
 
 
+##### Variables: Defaults
+
+You can specify a default value for a variable, which is displayed when the variable is not specified in `variables.md`. This is done by adding `or (defaultValue)` within the curly braces.
```
yamgent: `or defaultValue`?

docs/userGuide/reusingContents.md
```diff
@@ -101,6 +101,13 @@ As mentioned in _User Guide: MarkBind Syntax_, you can use built-in variables to
...
+
+<code>My name is {<span></span>{ name or "Anonymous" }}.</span></code>
```
yamgent:
```
<div class="indented">

{{ icon_example }} If `name` is not declared in `variables.md`:<br>
<code>My name is {<span></span>{ name or "Anonymous" }}.</span></code> {{ icon_arrow_right }} My name is Anonymous.
</div>
```

docs/userGuide/reusingContents.md
```diff
@@ -257,6 +264,27 @@ In other words, **`<include>` interprets the reused code relative to the origina
 
 <hr><!-- ======================================================================================================= -->
 
+#### Specifying Variables in the Include Tag
+
+**It is possible to include variables in the include tag.** For example,
```
yamgent: Use `{{ icon_example }}` and `<div class="indented">` whenever an example is listed.

src/lib/markbind/src/parser.js
```diff
@@ -205,7 +205,23 @@ Parser.prototype._preprocess = function (node, context, config) {
     let fileContent = self._fileCache[actualFilePath]; // cache the file contents to save some I/O
     const { parent, relative } = calculateNewBaseUrls(filePath, config.rootPath, config.baseUrlMap);
     const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
-    fileContent = nunjucks.renderString(fileContent, userDefinedVariables);
+    const includedVariables = context.includedVariables || {};
+
+    if (element.children) {
```
yamgent: Add a comment above to say what this portion of the logic is about. E.g.:

```js
// process variables declared within the include
```

src/lib/markbind/src/parser.js
```diff
@@ -205,7 +205,23 @@ Parser.prototype._preprocess = function (node, context, config) {
...
+      element.children.forEach((child) => {
+        if (child.name !== 'span') {
+          return;
+        }
+        if (!child.attribs.id) {
+          // eslint-disable-next-line no-console
+          console.warn(`Missing 'id' in variable for ${element.attribs.src} include.`);
```
yamgent: Can also include the name of the current file? `\nMissing reference in ${element.attribs[ATTRIB_CWF]}`.

This is so that the author has an easier time finding the offending variable.

test/test_site/index.md
```diff
@@ -94,6 +94,22 @@ layout: default
...
+  <span id="includedVariableAsAttribute">color: blue</span>
+  <span id="includedVariableAsHtmlElement"><span>Included variable within html element</span></span>
+  <span id="includedVariableOverriding">Included variable overriding variables.md</span>
+  <span id="includedVariableInIncludedFile">Included variable in included file</span>
+  <span id="includedVariableWithGlobalVariable">Included variable with {{ includedGlobalVariable }}</span>
+</include>
+
+# Included variables should not leak into other files
+
+<include src="testIncludeVariableLeak.md" />>
```
yamgent: Remove the extra `>`.

test/test_site/testIncludeVariableLeak.md
```diff
@@ -0,0 +1,2 @@
+
```
yamgent: Remove the blank line at the start of the file, don't think it is needed.

test/test_site/testIncludeVariables.md
```diff
@@ -0,0 +1,24 @@
+
```
yamgent: Remove the blank line at the start of the file, don't think it is needed.

docs/userGuide/reusingContents.md
acjh: Why is there `</span>` right before `</code>`?

yamgent: > Why is there `</span>` right before `</code>`?

My bad, didn't spot the error. The `</span>` is not necessary. Edited the original post.

docs/userGuide/reusingContents.md
```diff
@@ -257,6 +267,33 @@ In other words, **`<include>` interprets the reused code relative to the origina
 
 <hr><!-- ======================================================================================================= -->
 
+#### Specifying Variables in the Include Tag
```
yamgent: 1. Let's follow the same heading level as "\<include> Inside an Included File".
1. Let's use "\<include>" rather than "Include Tag" to follow the style of the rest of the documentation.

```markdown
##### Specifying Variables in an `<include>`
```

docs/userGuide/reusingContents.md
```diff
@@ -257,6 +267,33 @@ In other words, **`<include>` interprets the reused code relative to the origina
...
+
+**It is possible to include variables in the include tag.**
```
yamgent:
```markdown
**It is possible to declare variables in an `<include>`.**
```

docs/userGuide/reusingContents.md
```diff
@@ -257,6 +267,33 @@ In other words, **`<include>` interprets the reused code relative to the origina
...
+
+<div class="indented">
+
+{{ icon_example }} Specifying `title` and `author` variables in an `include` tag:
```
yamgent:
```markdown
... in an `<include>` tag:
```

docs/userGuide/reusingContents.md
```diff
@@ -257,6 +267,33 @@ In other words, **`<include>` interprets the reused code relative to the origina
...
+
+<code>
+# {<span></span>{ title }}
+
+Author: {<span></span>{ author }}
+</code>
+</div>
+
+
+These variables work the same way as variables in `_markbind/variables.md`, except they only apply to the included file. This can be helpful for creating a template for multiple files.
```
yamgent: 1. Fixing the grammar.
2. Rephrased the explanation of it being a template mechanism.

```markdown
..., except that they only apply to the included file. They allow the included file to be reused as a template, for different source files using different variable values.
```

damithc: Side question: do they also work as nunjucks variables? e.g., `{{ (size | int) - 1 }}` would evaluate to 3 if `<span id="size">4</span>`

damithc:
> Does this work too? It's not related to includes but possibly related to how the feature is implemented.
> 
> ```
> <span id="color">red</span>
> 
> <font color="{{ color }}">my text</font>
> ```

@jamos-tay make sure this is reflected somewhere in the docs too. i.e., together with the `d-none` attribute, this provides a convenient way to declare local variables and use them in a page.

docs/userGuide/reusingContents.md
jamos-tay: Tested and it works - but do we want to document this? It seems like a small enough case, and it might confuse the user since it introduces another api

jamos-tay:
> @jamos-tay make sure this is reflected somewhere in the docs too. i.e., together with the `d-none` attribute, this provides a convenient way to declare local variables and use them in a page.

@damithc Sorry I think I misunderstood, were you referring to 
```
<span id="color">red</span>

<font color="{{ color }}">my text</font>
```
within the same `.md` file? Because that doesn't work... I thought you were asking if you could put variables as element attributes.

I don't think there's a way to create a local variable for a page just yet, but it should be put it in a different PR. I'm pretty sure it requires separate logic (such as placing in the frontmatter) because of the way nunjucks is called.

docs/userGuide/reusingContents.md
damithc: We already claim that MarkBind is generally compatible with nunjucks syntax. Here, we are using nunjucks syntax for a MarkBind feature. It's better to clarify that these variables are in fact equivalent to nunjucks variables declared within the scope of the included file. However, it can be mentioned in https://markbind.github.io/markbind/userGuide/markBindSyntax.html#support-for-nunjucks rather than under this feature. What do you think?

jamos-tay:
Updated

damithc:
> within the same `.md` file? Because that doesn't work... I thought you were asking if you could put variables as element attributes.
> 
> I don't think there's a way to create a local variable for a page just yet, but it should be put it in a different PR. I'm pretty sure it requires separate logic (such as placing in the frontmatter) because of the way nunjucks is called.

I see. I thought we got that feature for free. Never mind then. It can be a feature to be added in the future.

docs/userGuide/reusingContents.md
```diff
@@ -145,6 +155,17 @@ You must use the `safe` filter when using such variables:
 <code>{<span></span>{ right_hand_2 }}</code> {{ icon_arrow_right }} {{glyphicon_hand_right}}
 </div>
 
+**Variables can also be used in element attributes.**
+
+<div class="indented">
+
+{{ icon_example }} Specifying a variable as an attribute:
+
+<code>\<span id="color">red</span>\</code><br>
```
yamgent: The wrong HTML tag is escaped, it didn't render correctly.

docs/userGuide/reusingContents.md
```diff
@@ -145,6 +155,17 @@ You must use the `safe` filter when using such variables:
...
+
+<code>\<font color="{<span></span>{ color }}">my text</font>\</code>
```
yamgent: Same problem (wrong HTML tag escaped).

docs/userGuide/reusingContents.md
yamgent: Also it is not clear that the two different code exist in two different places, especially when you look at this portion in the live preview. Maybe indicate one is in `variables.md` and one is in `index.md`?

jamos-tay:
Updated, I decided to delete that section since there was a misunderstanding.

src/lib/markbind/src/parser.js
```diff
@@ -205,7 +205,25 @@ Parser.prototype._preprocess = function (node, context, config) {
...
+    const includedVariables = context.includedVariables || {};
+    if (element.children) {
+      element.children.forEach((child) => {
+        if (child.name !== 'span') {
+          return;
+        }
+        if (!child.attribs.id) {
+          // eslint-disable-next-line no-console
+          console.warn(`Missing reference in ${element.attribs[ATTRIB_CWF]}\n`
+                     + `Missing 'id' in variable for ${element.attribs.src} include.`);
```
yamgent: 1. The "Missing 'id'" part should actually be in front, in order to match the style of other error messages that we show as well.
2. I realise that the "Missing reference..." should be rephrased (because it is not a missing reference, but rather, a missing variable name), apologies about that.

So it should be something like this:

```diff
diff --git a/src/lib/markbind/src/parser.js b/src/lib/markbind/src/parser.js
index 69c40e5..b71b50f 100644
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -215,8 +215,8 @@ Parser.prototype._preprocess = function (node, context, config) {
         }
         if (!child.attribs.id) {
           // eslint-disable-next-line no-console
-          console.warn(`Missing reference in ${element.attribs[ATTRIB_CWF]}\n`
-                     + `Missing 'id' in variable for ${element.attribs.src} include.`);
+          console.warn(`Missing 'id' in variable for referenced file: ${element.attribs.src}.\n`
+                      + `Missing variable name in ${element.attribs[ATTRIB_CWF]}\n`);
           return;
         }
         includedVariables[child.attribs.id] = cheerio.html(child.children);
```

damithc:
@jamos-tay do look into how variables propagate via nested includes. Ideally, the variable value should apply to nested includes too and outermost value should override values specified in inner includes.

yamgent:
> Ideally, the variable value should apply to nested includes too and outermost value should override values specified in inner includes.

Do you mean that for the following scenario, the final output should be "Original variable"? The behaviour would look very odd to authors, since the inner page that is "closer" to the final included page should be allowed to have a say over the final variable value, rather than the outer page.

---

`index.md`
```html
<include src="a.md">
    <span id="someVariable">Original variable</span>
</include>
```

`a.md`
```html
<include src="b.md">
    <span id="someVariable">Nested include variable</span>
</include>
```

`b.md`
```html
{{someVariable}}
```

---

damithc:
> Do you mean that for the following scenario, the final output should be "Original variable"? The behaviour would look very odd to authors, since the inner page that is "closer" to the final included page should be allowed to have a say over the final variable value, rather than the outer page.

Yes, that feels odd but from a reuse perspective that should be the behavior (opposite of how inheritance works). Otherwise, someone who wants to resue this bit of code has no way to customize the code without modifying it directly, right?

yamgent:
> Otherwise, someone who wants to resue this bit of code has no way to customize the code without modifying it directly, right?

That is indeed true, I think that is the desirable behaviour.

In order to make it easier to explain this behaviour, maybe we should consider adopting a different term rather than calling them all "variables"? Maybe "page variables" vs "global variables"? So we can say that "If a page uses a `{{variable}}`, then:

1. Page variables override global variables.
1. Outer page variables override inner page variables.

damithc:
> In order to make it easier to explain this behaviour, maybe we should consider adopting a different term rather than calling them all "variables"?

That is a good idea. One issue is that we use the exact syntax used for nunjucks variables. Are we using nunjucks behind the scene to implement variables? This could become a problem because nunjucks variables are likely to behave differently from ours and users don't have a way to tell the two apart (note that a MarkBind file can contain some nunjucks code too). Is it possible to use a slightly different syntax e.g., `{< foo >}`?

> 1. Page variables override global variables.

This should be the other way around too?

> Maybe "page variables" vs "global variables"?

Or stay away from _variables_ altogether. _customization_?

yamgent:
> That is a good idea. One issue is that we use the exact syntax used for nunjucks variables. Are we using nunjucks behind the scene to implement variables? This could become a problem because nunjucks variables are likely to behave differently from ours and users don't have a way to tell the two apart (note that a MarkBind file can contain some nunjucks code too). Is it possible to use a slightly different syntax e.g., `{< foo >}`?

Yes, we are using nunjucks behind the scene to implement variables.

And yes there is a chance the authors may mix them up. We can explore using a different syntax in another PR (that would be a breaking change)?

> This should be the other way around too?

Got it, global ones should have the highest priority. @jamos-tay you will have to rework this. :P

damithc:
> And yes there is a chance the authors may mix them up. We can explore using a different syntax in another PR (that would be a breaking change)?

It would break the syntax introduced in this PR, right? :-) If the new syntax is introduced in this PR, it will not break existing features.
Anyway, it is a big change, and I'm not sure how hard it is to implement, especially if we cannot piggyback on nunjucks anymore. We can discuss on Saturday.

jamos-tay:
Hmm... I think it might be possible to reuse the same syntax. I just have to check that if a variable already exists (i.e. declared by an outer file), don't overwrite it.

I'll need a way to detect whether a variable is a global or page variable though, but I think it can be done without using a hacky solution.

damithc:
> Hmm... I think it might be possible to reuse the same syntax. I just have to check that if a variable already exists (i.e. declared by an outer file), don't overwrite it.

Reusing the same syntax is appropriate only if a normal nunjucks variable behaves exactly the same way as the ones used for includes. Are we going to apply the same outermost-overrides behavior to normal nunjucks variables as well? i.e., those declared using `{% set x = 'foo' %}`

jamos-tay:
Hmm... Nunjucks overwrites, so it'll use the inner most by default. We might be able to write an [extension](https://mozilla.github.io/nunjucks/api.html#custom-tags) that will overwrite `set`'s behavior. Alternatively we could strip out {% set %} tags and parse them manually.

If not, are we considering moving the entire variable system off nunjucks? i.e. {< variable >} everywhere? That might be a lot of work just for this feature... considering the global `variables.md` would have to be migrated over, we'd have to write our own parser logic, and also all older sites would break.

damithc:
> Hmm... Nunjucks overwrites, so it'll use the inner most by default. We might be able to write an [extension](https://mozilla.github.io/nunjucks/api.html#custom-tags) that will overwrite `set`'s behavior. Alternatively we could strip out {% set %} tags and parse them manually.
> 
> If not, are we considering moving the entire variable system off nunjucks? i.e. {< variable >} everywhere? That might be a lot of work just for this feature... considering the global `variables.md` would have to be migrated over, we'd have to write our own parser logic, and also all older sites would break.

Yup, this is a big decision. Better to make it now rather than later but we have to think it through. Although the issue came up in this PR, we are actually dealing with a much wider issue, namely, the need to support the _outer-overrides-inner_ behavior that is central to reuse but opposite of the normal template behavior. The problem seem to be that we cannot support two opposing behaviors using the same internal library and/or syntax.

src/lib/markbind/src/parser.js
```diff
@@ -205,7 +205,25 @@ Parser.prototype._preprocess = function (node, context, config) {
     let fileContent = self._fileCache[actualFilePath]; // cache the file contents to save some I/O
     const { parent, relative } = calculateNewBaseUrls(filePath, config.rootPath, config.baseUrlMap);
     const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
-    fileContent = nunjucks.renderString(fileContent, userDefinedVariables);
+
+    // process variables declared within the include
+    const includedVariables = context.includedVariables || {};
+    if (element.children) {
+      element.children.forEach((child) => {
```
acjh: Let's abstract this into a function.
```js
const includedVariables = extractVariables(element, context.includedVariables);
```

```js
function extractVariables(node, variables) {
  // Clone variables, modify the clone and return it
  // ...
}
```

jamos-tay:
Updated

Note: We currently do not support include variables referencing variables within the same include:

index.md:
```html
<include src="a.md">
  <span id="a">a</span>
  <span id="b">b & {{ a }}</span>
</include>
```

This is because {{ a }} gets rendered to an empty string when nunjucks processes `index.md`. Making this work would require some rework the page generation logic, which should be in another PR. This rework is also necessary to handle certain cases with `set`.

jamos-tay:
Rebased on master

test/test_site/index.md
```diff
@@ -95,6 +95,24 @@ head: myCustomHead.md, myCustomHead2.md
...
 ## <include src="testTrimInclude.md" trim inline />
 
+# Include with custom variables
+
+<include src="testIncludeVariables.md">
+  <span id="includedVariable">Included variable</span>
+  <span id="includedVariableWithMarkdown">__**Included variable with markdown**__</span>
+  <span id="includedVariableAsAttribute">color: blue</span>
+  <span id="includedVariableAsHtmlElement"><span>Included variable within html element</span></span>
+  <span id="includedVariableGlobalOverriden">Included variable overriden by global variable</span>
```
yamgent: General typo: "overriden" is spelt with two d (i.e. "overridden").

test/test_site/testIncludeVariables.md
```diff
@@ -0,0 +1,39 @@
...
+</include>
+
+# Inner included variables should not leak into other files
+<include src="testIncludeVariableLeakInner.md" />
+
+# Test included variable with global variable
+{{ includedVariableWithGlobalVariable }}
+
+# Test included variable with local variable
+{{ includedVariableWithLocalVariable }}
```
yamgent: Is this supposed to be rendered blank?

docs/userGuide/reusingContents.md
```diff
@@ -259,6 +269,32 @@ In other words, **`<include>` interprets the reused code relative to the origina
...
+
+<code>
+# {<span></span>{ title }}
+
+Author: {<span></span>{ author }}
+</code>
+</div>
+
+These variables work the same way as variables in `_markbind/variables.md`, except that they only apply to the included file. They allow the included file to be reused as a template, for different source files using different variable values.
+
```
yamgent: The documentation can explain the precedence of variables (global > outer include > inner include).

test/test_site/testIncludeVariables.md
jamos-tay: My bad, was a remainder from when the functionality was removed since it wasn't compatible with nunjucks

jamos-tay:
Updated

test/test_site/_markbind/variables.md
```diff
@@ -7,3 +7,6 @@
 <span id="reference_level_2">{{reference_level_1}}</span>
 <span id="reference_level_3">{{reference_level_2}}</span>
 <span id="reference_level_4">{{reference_level_3}}</span>
+
+<span id="includedVariableGlobalOverridden">Included variable overriding page variables</span>
```
yamgent: `Included variable overriding page variables` -> `Global variable overriding page variables`?

test/test_site/index.md
```diff
@@ -95,6 +95,24 @@ head: myCustomHead.md, myCustomHead2.md
...
+# Include with custom variables
+
+<include src="testIncludeVariables.md">
+  <span id="includedVariable">Included variable</span>
+  <span id="includedVariableWithMarkdown">__**Included variable with markdown**__</span>
+  <span id="includedVariableAsAttribute">color: blue</span>
+  <span id="includedVariableAsHtmlElement"><span>Included variable within html element</span></span>
+  <span id="includedVariableGlobalOverridden">Included variable overridden by global variable</span>
+  <span id="includedVariableInnerOverridden">Included variable overriding inner variable</span>
+  <span id="includedVariableInIncludedFile">Included variable in included file</span>
```
yamgent: `includedVariableInIncludedFile` -> `includedVariableInOuterIncludedFile` (which include is the value coming from is not very obvious with the names).

test/test_site/_markbind/variables.md
acjh: Also, use `kebab-case` for `id` (and `class`).

jamos-tay:
Updated

PR #554, opened 2019-01-05
Use light grey for siteNav text

Xenonym:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] Enhancement to an existing feature

Resolves #542.

**What is the rationale for this request?**
We can have lighter coloured text for the site navigation sidebar to give more relative importance to the main text.

**What changes did you make? (Give an overview)**
Adjust the `color`for `siteNav` text to be `#444444`.

**Testing instructions:**
1. Observe that in MarkBind documents with a navigation sidebar, the colour of the text is `#444444`.

PR #560, opened 2019-01-07
 Indicate currently loaded page in siteNav

### LINKED ISSUE #544: siteNav: indicate selected item
v1.15.2

Suggestion: indicate the currently loaded item in the siteNav. An example (from Hexo):

image

Xenonym:

Resolves #544.

Most site navigation menus in other websites indicate the current page the user is on. We can also indicate which page the user is on in MarkBind's siteNav to allow for easier navigation.

I added a check to mark the link referring to the current page being processed in `formatSiteNav` as `class='current'` and also added a `#site-nav a.current` CSS rule to highlight pages marked as `current`, thus highlighting the current loaded page in the siteNav. If the current page link is in a dropdown that is collapsed by default, that dropdown will automatically be expanded.

Note that the current colour being used to highlight the current page `#0072EC` is rather close to the colour for a dropdown `#0076FF`. This should not be a problem once #554 is merged.

**Testing instructions:**
1. Render and open a MarkBind page with a siteNav. If the page is listed in the siteNav, it should be highlighted.
2. If the page is inside a dropdown that is collapsed by default, it should be expanded automatically.

### DIFF
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -30,7 +30,9 @@
     -webkit-transition: 0.4s;
 }
 
-#site-nav a:link, #site-nav a:visited, #site-nav a:hover, #site-nav a:active {
+#site-nav a:active,
+#site-nav a:link,
+#site-nav a:visited {
     color: #565656;
     text-decoration: none;
 }
@@ -114,7 +116,9 @@
 
 /* Hamburger button */
 
-.menu-top-bar, .menu-middle-bar, .menu-bottom-bar {
+.menu-top-bar,
+.menu-middle-bar,
+.menu-bottom-bar {
     background-color: darkslategrey;
     height: 4px;
     margin: 5px 10%;
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -41,6 +41,10 @@
     color: #0076FF;
 }
 
+#site-nav a.current {
+    color: #0072EC;
+}
+
 #site-nav-btn {
     background: transparent;
     cursor: pointer;
--- a/src/Page.js
+++ b/src/Page.js
@@ -129,7 +129,7 @@ function formatFooter(pageData) {
   return $.html();
 }
 
-function formatSiteNav(renderedSiteNav) {
+function formatSiteNav(renderedSiteNav, src) {
   const $ = cheerio.load(renderedSiteNav);
   const listItems = $.root().find('ul').first().children();
   if (listItems.length === 0) {
@@ -137,6 +137,10 @@ function formatSiteNav(renderedSiteNav) {
   }
   // Tidy up the style of the unordered list <ul>
   listItems.parent().attr('style', 'list-style-type: none; margin-left:-1em');
+
+  // Highlight current page
+  const currentPageHtmlPath = src.replace(/\.(md|mbd)$/, '.html');
+  listItems.find(`a[href='{{baseUrl}}/${currentPageHtmlPath}']`).addClass('current');
 
   listItems.each(function () {
     // Tidy up the style of each list item
@@ -148,7 +152,7 @@ function formatSiteNav(renderedSiteNav) {
         // Double wrap to counter replaceWith removing <li>
         nestedList.parent().wrap('<li style="margin-top:10px"></li>');
         // Recursively format nested lists without dropdown wrapper
-        nestedList.parent().replaceWith(formatSiteNav(nestedList.parent().html()));
+        nestedList.parent().replaceWith(formatSiteNav(nestedList.parent().html(), src));
       }
     // Found nested list, render dropdown menu
     } else if ($(this).children('ul').length) {
@@ -156,8 +160,10 @@ function formatSiteNav(renderedSiteNav) {
       const dropdownTitle = $(this).contents().not('ul');
       // Replace the title with the dropdown wrapper
       dropdownTitle.remove();
-      // Check if dropdown is expanded by default
-      if (dropdownTitle.toString().includes(DROPDOWN_EXPAND_KEYWORD)) {
+      // Check if dropdown is expanded by default or if the current page is in a dropdown
+      const shouldExpandDropdown = dropdownTitle.toString().includes(DROPDOWN_EXPAND_KEYWORD)
+        || Boolean(nestedList.find(`a[href='{{baseUrl}}/${currentPageHtmlPath}']`).text());
+      if (shouldExpandDropdown) {
         const expandKeywordRegex = new RegExp(DROPDOWN_EXPAND_KEYWORD, 'g');
         const dropdownTitleWithoutKeyword = dropdownTitle.toString().replace(expandKeywordRegex, '');
         const rotatedIcon = cheerio.load(DROPDOWN_BUTTON_ICON_HTML, { xmlMode: false })('i')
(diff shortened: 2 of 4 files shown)


asset/css/site-nav.css
```diff
@@ -30,7 +30,7 @@
     -webkit-transition: 0.4s;
 }
 
-#site-nav a:link, #site-nav a:visited, #site-nav a:hover, #site-nav a:active {
+#site-nav a:link, #site-nav a:visited, #site-nav a:active {
```
Chng-Zhi-Xuan: Good catch on the redundant `#site-nav a:hover`

damithc:
@Xenonym what happens if the user clicked on a lower-level menu item that is collapsed by default? It's fine if that case is not handled in this PR but keep in mind that we should be able to extend the current solution to cover that case.

damithc:
On a separate note, blue is also used to indicate links. Should we use some other means e.g., use a light background to mark the clicked item?

src/Page.js
```diff
@@ -482,7 +488,7 @@ Page.prototype.insertSiteNav = function (pageData) {
     throw new Error(`More than one <navigation> tag found in ${siteNavPath}`);
   } else if (siteNavDataSelector('navigation').length === 1) {
     const siteNavHtml = md.render(siteNavDataSelector('navigation').html().trim().replace(/\n\s*\n/g, '\n'));
-    const formattedSiteNav = formatSiteNav(siteNavHtml);
+    const formattedSiteNav = formatSiteNav(siteNavHtml, this.src);
```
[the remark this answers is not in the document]
Xenonym: The first time we render the siteNav for a page, we pass in `this.src` so that `formatSiteNav` can locate the link to set `class=current` to at `Page.js#143`. When we recursively format the same siteNav for nested lists in `Page.js#157` and `Page.js#184`, we do not pass in the `src` parameter since we do not need to highlight the same link repeatedly for the same page.

Xenonym:
> @Xenonym what happens if the user clicked on a lower-level menu item that is collapsed by default? It's fine if that case is not handled in this PR but keep in mind that we should be able to extend the current solution to cover that case.

If the link is in a dropdown that is collapsed by default, the link will still be highlighted but the dropdown will not be expanded automatically. I assume it will be better if the dropdown is also expanded in this scenario?

damithc:
> I assume it will be better if the dropdown is also expanded in this scenario?

I think so too.

> Not in the siteNav right? #554 would mean that the only items that are blue in the siteNav are these highlighted links. Or are you referring to potential confusion between links in the body and the siteNav?

Yes, I'm referring to the confusion arising from the fact that users unconsciously associate blue text with a clickable link. But it's not a big deal.

src/Page.js
```diff
@@ -129,7 +129,7 @@ function formatFooter(pageData) {
   return $.html();
 }
 
-function formatSiteNav(renderedSiteNav) {
+function formatSiteNav(renderedSiteNav, src) {
```
yamgent: > The first time we render the siteNav for a page, we pass in `this.src` so that `formatSiteNav` can locate the link to set `class=current` to at `Page.js#143`. When we recursively format the same siteNav for nested lists in `Page.js#157` and `Page.js#184`, we do not pass in the `src` parameter since we do not need to highlight the same link repeatedly for the same page.

I see. However, we don't seem to use the content of `src` at all in the method, I wonder if we can substitute it with a boolean value instead?

Xenonym: No, because we actually do use the value of `src` at `Page.js#143` to find the correct link to highlight:

```js
listItems.find(`a[href='{{baseUrl}}/${src.replace(/\.(md|mbd)$/, '.html')}']`)
      .attr('class', 'current');
```

Xenonym:
@damithc a lower-level menu item that is collapsed by default should now expand when the user clicks on it.

damithc:
> @damithc a lower-level menu item that is collapsed by default should now expand when the user clicks on it.

Nice!

yamgent:
Only one minor issue: This doesn't work for `http://127.0.0.1:8080/markbind/userGuide/index.html`. However, this is because `userGuide/index.html` just includes `userGuide/gettingStarted.md`, rather than being the actual "Getting Started" page itself.

Not sure how we can legitimately solve that issue in this PR, because `userGuide/index.md` and `userGuide/gettingStarted.md` are ultimately different pages, even if they display the same content.

asset/css/site-nav.css
acjh: Let's do:
```css
#site-nav a:active,
#site-nav a:link,
#site-nav a:visited {
```

so it's distinct from e.g.:
```css
#site-nav .active .link a:visited {
```

https://stylelint.io/user-guide/rules/selector-list-comma-newline-after/

test/test_site/expected/markbind/css/site-nav.css
```diff
@@ -30,7 +30,7 @@
     -webkit-transition: 0.4s;
 }
 
-#site-nav a:link, #site-nav a:visited, #site-nav a:hover, #site-nav a:active {
+#site-nav a:link, #site-nav a:visited, #site-nav a:active {
```
acjh: @Xenonym Same here.

src/Page.js
```diff
@@ -138,6 +138,9 @@ function formatSiteNav(renderedSiteNav) {
   // Tidy up the style of the unordered list <ul>
   listItems.parent().attr('style', 'list-style-type: none; margin-left:-1em');
 
+  // highlight current page
+  listItems.find(`a[href='{{baseUrl}}/${src.replace(/\.(md|mbd)$/, '.html')}']`).attr('class', 'current');
```
acjh: - Begin comments with a capital letter: `// highlight` → `// Highlight`
- Instead of `.attr('class', 'current')`, which replaces the `class`, use `.addClass('current')`.
- Where do we remove the `class`?
- Let's assign `${src.replace(/\.(md|mbd)$/, '.html')}` to a constant since it's clearer and also used below.

src/Page.js
```diff
@@ -148,16 +151,18 @@ function formatSiteNav(renderedSiteNav) {
...
     } else if ($(this).children('ul').length) {
       const nestedList = $(this).children('ul').first();
       const dropdownTitle = $(this).contents().not('ul');
       // Replace the title with the dropdown wrapper
       dropdownTitle.remove();
-      // Check if dropdown is expanded by default
-      if (dropdownTitle.toString().includes(DROPDOWN_EXPAND_KEYWORD)) {
+      // Check if dropdown is expanded by default or if the current page is in a dropdown
+      const expandDropdown = dropdownTitle.toString().includes(DROPDOWN_EXPAND_KEYWORD)
+        || nestedList.find(`a[href='{{baseUrl}}/${src.replace(/\.(md|mbd)$/, '.html')}']`).text();
```
acjh: - Wrap in `Boolean(nestedList...text())` to be clear that we're using its truthiness?
- Rename constant to `shouldExpandDropdown` to reflect boolean type.

src/Page.js
acjh: >> Where do we remove the class?
>
> We don't, the class is needed for `#site-nav a.current` in `site-nav.css` to be applied

Ah right, I was thinking of page nav. For site nav, a page will always be `current`.

src/Page.js
```diff
@@ -138,6 +138,10 @@ function formatSiteNav(renderedSiteNav) {
   // Tidy up the style of the unordered list <ul>
   listItems.parent().attr('style', 'list-style-type: none; margin-left:-1em');
 
+  // Highlight current page
+  const curPage = src.replace(/\.(md|mbd)$/, '.html');
```
yamgent: `curPage` -> `currentPageHtmlPath`

PR #568, opened 2019-01-09
Optimize site data update during live preview

nicholaschuayunzhi:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

For large sites, live reload re-collects and generate all heading data, causing long reload times for small changes.

Only re-collect heading info for update pages. If `force-reload` flag is called, all site data is reloaded, ie a reload results in a full build.

Previously for `cs2103/website-base` live reload incurred additional 23s for reload (on my machine). Currently it has been reduced to on average 3-5s for changes. The speed depends on the number of affected files during a reload.

Order and naming of default params created for `generateSiteData`.

**Testing instructions:**
1. Update header
1. Update front matter
1. Observe search is updated

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -454,7 +454,7 @@ Site.prototype.generate = function (baseUrl) {
       .then(() => this.buildSourceFiles())
       .then(() => this.copyMarkBindAsset())
       .then(() => this.copyLayouts())
-      .then(() => this.generateSiteData())
+      .then(() => this.updateSiteData())
       .then(() => {
         const endTime = new Date();
         const totalBuildTime = (endTime - startTime) / 1000;
@@ -656,7 +656,7 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
 
   return new Promise((resolve, reject) => {
     Promise.all(processingFiles)
-      .then(() => this.generateSiteData())
+      .then(() => this.updateSiteData(shouldRebuildAllPages ? undefined : filePaths))
       .then(() => logger.info('Pages rebuilt'))
       .then(resolve)
       .catch(reject);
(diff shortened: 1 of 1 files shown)


src/Site.js
```diff
@@ -667,11 +667,15 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
 /**
  * Uses heading data in built pages to generate heading and keyword information for siteData
  * Subsequently writes to siteData.json
+ * @param filePaths array of paths corresponding to files that have changed
+ * @param generateAllPages if all pages heading data should be generated
  */
-Site.prototype.generateSiteData = function () {
+Site.prototype.generateSiteData = function (filePaths = [], generateAllPages = true) {
```
yamgent: 1. `generateAllPages` -> `generateForAllPages`? Since in this case we are generate site data **for** pages, not generating the pages themselves.
1. Should we swap the order of the parameters? `generateForAllPages` seems to be a much more powerful option than `filePaths`, so `generateForAllPages` can be first?

src/Site.js
```diff
@@ -667,11 +667,15 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
 /**
  * Uses heading data in built pages to generate heading and keyword information for siteData
  * Subsequently writes to siteData.json
+ * @param filePaths array of paths corresponding to files that have changed
```
yamgent: Using the word "changed" here sound a bit strange (I know it is following the comments in other method, but those methods are usually doing purely regeneration, rather than initial generation + regeneration).

Perhaps the name of the method can be `updateSiteData()` rather than `generateSiteData()`? To make it clearer that the changes are accumulative.

nicholaschuayunzhi:
@yamgent thanks for the review :smile: ! I've updated the pr based on your comments.

src/Site.js
```diff
@@ -667,11 +667,15 @@ Site.prototype.regenerateAffectedPages = function (filePaths) {
 /**
  * Uses heading data in built pages to generate heading and keyword information for siteData
  * Subsequently writes to siteData.json
+ * @param generateForAllPages if all pages heading data should be generated
+ * @param filePaths optional array of updated file paths during live preview
  */
-Site.prototype.generateSiteData = function () {
+Site.prototype.updateSiteData = function (generateForAllPages = true, filePaths = []) {
```
acjh: They are mutually exclusive when used right? How about accepting only `filePaths` with no default value, and then assigning `const updateForAllPages = filePaths === undefined;` internally?

Then calling it would be `updateSiteData(shouldRebuildAllPages ? undefined : filePaths)`.

Or pass and check `true` instead of `undefined`.

nicholaschuayunzhi: I thought about this but I was afraid that `updateSiteData(shouldRebuildAllPages ? undefined : filePaths)` would be confusing for other devs. I've no objections to it though. I think my main goal was to keep the function call argumentless (ie unchanged) in build process.

yamgent:
Do squash the commits. Thanks!

PR #573, opened 2019-01-10
Suppress error message for parser test

jamos-tay:

• [X] Bug fix

This error has been showing up in the [testing console](https://travis-ci.org/MarkBind/markbind/builds/477772901?utm_source=github_status&utm_medium=notification) for a while now:

image

I traced it back to the `includeFile replaces <include src="include.md#exists" optional> with <div>` test which was added by #517.

An error is expected so behavior is correct. Suppress the error message by supplying a no-op error handler.

### DIFF
--- a/test/unit/parser.test.js
+++ b/test/unit/parser.test.js
@@ -192,7 +192,9 @@ test('includeFile replaces <include src="include.md#doesNotExist"> with error <d
   const baseUrlMap = {};
   baseUrlMap[ROOT_PATH] = true;
 
-  const markbinder = new MarkBind();
+  const markbinder = new MarkBind({
+    errorHandler: () => {},
+  });
   const result = await markbinder.includeFile(indexPath, {
     baseUrlMap,
     rootPath: ROOT_PATH,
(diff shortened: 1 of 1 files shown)


test/unit/parser.test.js
```diff
@@ -192,7 +192,9 @@ test('includeFile replaces <include src="include.md#doesNotExist"> with error <d
   const baseUrlMap = {};
   baseUrlMap[ROOT_PATH] = true;
 
-  const markbinder = new MarkBind();
+  const markbinder = new MarkBind({
+    errorHandler: () => {},
```
acjh: Since we're constructing and testing the error message in `expected`, how about extracting that and `expect(e.message).toEqual` that here?

```diff
+ const expectedErrorMessage = `No such segment 'doesNotExist' in file: ${includePath}`
+   + `\nMissing reference in ${indexPath}`;

const markbinder = new MarkBind({
+ errorHandler: (e) => { expect(e.message).toEqual(expectedErrorMessage); },
});

// ...

const expected = [
  '# Index',
- `<div style="color: red">No such segment 'doesNotExist' in file: ${includePath}`,
- `Missing reference in ${indexPath}</div>`,
+ `<div style="color: red">${expectedErrorMessage}</div>`,
  '',
].join('\n');
```

PR #578, opened 2019-01-14
 Restore old anchor generation behaviour from before v1.15.0

### LINKED ISSUE #576: Intra-page links are not compatible with GFMD
v1.16.0

Current: The auto-generated link to heading `#Foo: Bar` will be `#foo-%3A-bar` instead of `#foo-bar` which is what GFMD generates.

Are we stuck with the curent or is it possible to go back to the previous GFMD-compatible state?

Xenonym:

• [X] Bug fix

Resolves #576.

When we upgraded `markdown-it-anchor` to `5.0.0` in #469, we also changed the anchor tag generation behaviour: instead of ignoring special characters, the new version now URL-encodes them instead. For example, "Guideline: Avoid Unsafe Shortcuts" becomes `guideline%3A-avoid-unsafe-shortcuts`. This breaks compatibility with headings generated by GFMD, which expects `guideline-avoid-unsafe-shortcuts`.

I added [`@sindresorhus/slugify`](https://github.com/sindresorhus/slugify) which provides slugify behaviour similar to what we had previously, and configured `markdown-it-anchor` to use this new library.

```markdown
## Guideline: Avoid Unsafe Shortcuts
```

This change needs to be coupled with a change to MarkBind/vue-strap to revert similar changes made in markbind/vue-strap#92.

**Testing instructions:**
1. Create a markdown document with headers that have special characters eg. `:`. The anchor link should not contain special characters.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,6 +61,15 @@
         }
       }
     },
+    "@sindresorhus/slugify": {
+      "version": "0.8.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/slugify/-/slugify-0.8.0.tgz",
+      "integrity": "sha512-Y+C3aG0JHmi4nCfixHgq0iAtqWCjMCliWghf6fXbemRKSGzpcrHdYxGZGDt8MeFg+gH7ounfMbz6WogqKCWvDg==",
+      "requires": {
+        "escape-string-regexp": "^1.0.5",
+        "lodash.deburr": "^4.1.0"
+      }
+    },
     "a-sync-waterfall": {
       "version": "1.0.0",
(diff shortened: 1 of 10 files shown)


Xenonym:
@yamgent if you are referring to jprichardson/string.js#212, note that the regex in question are not the ones used in the logic of `slugify`. I am not sure if the problem of unsafe user input applies in MarkBind's context: `slugify` is used to generate anchors, and thus the inputs must necessarily come from MarkBind source files written by the author, which probably will not contain malicious code unless the author seeks to crash his own computer.

Nevertheless, I do agree that if there is a external package available that is maintained with this functionality, we should probably go with that instead. Will look into [slugify](https://github.com/simov/slugify). Thanks for the heads up!

PR #584, opened 2019-01-16
Add check to ensure root folder when creating new Site object

### LINKED ISSUE #583: `markbind serve` should not serve a non-existent root directory

1. `markbind init` on an empty directory.
1. Do `markbind serve nonExistentRoot` (`nonExistentRoot/` does  **not** exist on the new website).

The command should fail, since the root directory given does not exist.

MarkBind creates a new folder `nonExistentRoot` and generates `_site/` inside the new folder. It seems to just serve the folder that was just init.

marvinchin:

• [X] Bug fix

Fixes #583 

Constructing a new Site object with a non-existent root folder should not be valid.

Add a check when constructing a new Site object to ensure that the provided root folder exists. If it does not exist, an error is thrown.

```js
// Should throw error
const site = new Site("non_existent_path", "_site");
```

Whether the Site object constructor is the best place to perform the check, or if there is a more appropriate place to do it.

**Testing instructions:**

Executing `./index.js serve non_existent_root_folder` should print an error message and terminate early.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,7 +69,13 @@ program
       options.onePage = fsUtil.ensurePosix(options.onePage);
     }
 
-    const site = new Site(rootFolder, outputFolder, options.onePage, options.forceReload, options.siteConfig);
+    let site;
+    try {
+      site = new Site(rootFolder, outputFolder, options.onePage, options.forceReload, options.siteConfig);
+    } catch (err) {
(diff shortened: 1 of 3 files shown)


marvinchin:
Fix underlying bug causing this issue in #595 instead.

Issue #574, opened 2019-01-11
System tests: allow more than one `test_site`


yamgent:
**Tell us about your environment**

* **MarkBind Version:** 1.16.0

Currently, we only have a `test/test_site`, which is a single MarkBind website, that we use to do all our system testing.

However, there may be cases where we need more than one MarkBind websites for testing.

For example, in the upcoming tags feature, we want to (1) test that tags in `site.json` overrides tags in pages, and (2) test that tags in individual pages actually work as expected when no overriding occurs. We currently cannot do that as we are [forced to choose one over another](https://github.com/MarkBind/markbind/pull/467#discussion_r246788142), because you cannot test both of them at the same time in one single website.

PR #585, opened 2019-01-16
Allow references to external scripts

### LINKED ISSUE #505: Layout: Allow references to external scripts

* **MarkBind Version:** 1.15.2

I want to refer to an external script at the bottom of the page. For example:

```html
   ...
   <script src="..\markbind\js\setup.js"></script>
   <script src="https://example.com/some-script.js"></script> 
   <script src="..\markbind\layouts\default\scripts.js"></script>
</html>
```

`https://example.com/some-script.js` is the external script that I am referring to.

There's no way of doing that with our current layout system.

Xenonym:

Resolves #505.

Users may want to refer to external scripts hosted online to add functionality such as mathematical formatting eg. [MathJax](https://www.mathjax.org) or charts eg. [Plotly.js](https://github.com/plotly/plotly.js).

I modified `page.ejs` to include any scripts in `assets.externalScripts` after the default MarkBind scripts, but before `_markbind\layouts\default\scripts.js` so that the user can override the behaviour of external scripts if necessary. Next, I modified `Site.js` to process `externalScripts` specified in either `pages` or in the config root, such that the user can choose to either reference external scripts only in specific pages or for the entire site.

```json
  "pages": [
    {
      "src": "index.md",
      "externalScripts": [
        "https://cdn.plot.ly/plotly-latest.min.js"
      ]
    }
  ],
  "externalScripts": [
    "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML"
  ],
```
I added a test page `testExternalScripts.md` for the per-page case, but have not added a corresponding test for the config root case since it will affect all pages in the test site. It would be cleaner if we do it in a separate test site after #574.

**Testing instructions:**
1. Reference some external scripts by specifying an array of external scripts as `externalScripts` at either `pages` or at the config root. 
	- If referenced at `pages`, they should only be referenced on those specific pages. 
	- If referenced at the config root, they should be referenced on all pages in the site.
2. The scripts should be referenced after the default MarkBind scripts, but before `_markbind\layouts\default\scripts.js`.

### DIFF
--- a/docs/userGuide/siteConfiguration.md
+++ b/docs/userGuide/siteConfiguration.md
@@ -28,11 +28,17 @@ Here is a typical `site.json` file:
       "src": "index.md",
       "title": "Hello World",
       "layout": "normal",
-      "searchable": "no"
+      "searchable": "no",
+      "externalScripts": [
+        "https://cdn.plot.ly/plotly-latest.min.js"
+      ]
     },
     {
       "glob": "**/index.md"
     }
+  ],
+  "externalScripts": [
+    "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML"
   ],
   "deploy": {
     "message": "Site Update.",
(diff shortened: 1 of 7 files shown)


docs/userGuide/siteConfiguration.md
```diff
@@ -82,6 +88,7 @@ Here is a typical `site.json` file:
 * **`title`**: The page `<title>` for the generated web page. Titles specified here take priority over titles specified in the [front matter](addingPages.html#front-matter) of individual pages.
 * **`layout`**: The [layout]({{ baseUrl }}/userGuide/advancedTopics.html#page-layout) to be used by the page. Default: `default`.
 * **`searchable`**: Specifies that the page(s) should be excluded from searching. Default: `yes`.
+* **`externalScripts`**: An array of external scripts to be refrenced on the page.
```
yamgent: refrenced -> referenced

yamgent: As discussed, also maybe describe where the scripts are inserted on the page, right now it is not very obvious.

yamgent:
Implementation wise looks good. :+1:

docs/userGuide/siteConfiguration.md
Xenonym: I added a brief description! Let me know if that's okay.

PR #588, opened 2019-01-16
Add multiple test site support

### LINKED ISSUE #574: System tests: allow more than one `test_site`

jamos-tay:

• [X] New feature

Fixes #574 

Some features, such as tagging, must be tested on different sites since the test cases are mutually exclusive.

Allow support for any number of test sites.

Revamped test directory structure:

- Old:
```
test/
  unit/
    (unit tests)
  test_site/
    (site files)
    test.sh
    test.bat
    testUtil/
      (test utils)
```

- New:
```
test/
  unit/
    (unit tests)
  functional/
    test.sh
    test.bat
    testUtil/
      (test utils)
    test_site/
      (test site files)
    test_site_2/
      (test site 2 files)
    test_site etc... /
      (...)
```

Creating a new test site:

- Create a folder in `test/functional`
- Run `markbind init` in the folder, write test site as per normal
- Create an `expected` directory within the folder and populate with expected files
- Add the folder name to these two lines:

test.sh
```
set sites=test_site test_site_2   // Add the folder name here
```

test.bat
```
declare -a sites=("test_site" "test_site_2")   // Add the folder name here
```

Note: Ensure `expected` directory is ignored (Add `expected/*` to the site.json ignore), and that no `src` or `glob` matches any of the expected files (the default glob `**/*.+(md|mbd)` causes some problems)

The current `test_site_2` folder is just there to show the feature works, it stores a fresh init site. Will remove before merge. No edits made to test_site, except for removing 3 lines from the site.json

How do we want to organize the tests? For example, say we have:
- Feature A: Test case A1 and Test case A2 are mutually exclusive
- Feature B: Test case B1 and Test case B2 are mutually exclusive

We could do something like this:
- Site 1: Other tests
- Site 2: A1, B1
- Site 3: A2, B2


jamos-tay:
Rebased, updated

I placed all the `test_site` related changes in a separate commit, so you should just have to look at the `Add multiple test site functionality` commit

test/functional/test.bat
```diff
@@ -0,0 +1,21 @@
+@ECHO off
+
+set sites=test_site test_site_2
+
+for %%a in (%sites%) do ( 
+	
+	echo(
```
yamgent: Should be 4 spaces, not `\t`.

test/functional/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+declare -a sites=("test_site" "test_site_2")
+
+for site in "${sites[@]}"
+do
+	# print site name
+	echo
```
yamgent: Same here with the tabs, should be 4 spaces.

test/functional/testExternalScripts.md
```diff
@@ -0,0 +1,6 @@
+The external script [MathJax 2.75](https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-MML-AM_CHTML) should be included, and the following MathJax content should render:
```
yamgent: These files shouldn't be outside? They are supposed to be inside `test_site`.

EDIT: `test_site` already has these files, so delete this one.

test/functional/testLayoutsOverride.md
```diff
@@ -0,0 +1,7 @@
+<frontmatter>
```
yamgent: Same as above, should be inside `test_site`.

test/functional/testTagDivs.md
```diff
@@ -0,0 +1,24 @@
+# Div with shown tag
```
yamgent: Same as above, should be inside `test_site`.

test/functional/testTags.md
```diff
@@ -0,0 +1,8 @@
+<frontmatter>
```
yamgent: Same as above, should be inside `test_site`.

test/functional/testExternalScripts.md
jamos-tay: My bad I think something went wrong during the rebase process

jamos-tay:
Rebased, updated

jamos-tay:
Yup, removed it

jamos-tay:
Rebased to fix conflicts

PR #590, opened 2019-01-16
Add spinner to be shown before content is mounted

### LINKED ISSUE #484: Show a spinner while loading page
Current: When a page loads, there is an initial period when a partially-rendered page (ugly!) appears before the page settles into its final form.

Suggested: Show a spinner while the page is loading 

On a related note, there is a slide-in animation (i.e., some content moves from left-to-right) that happens during the page loading (e.g., https://markbind.github.io/markbind/) that need not be there. Not sure where it is coming from. Noticeable only when the page loading is slow.

marvinchin:

• [x] New feature

Resolves #484 

Time taken for the page to render causes flash of unstyled content. To avoid having the user see the partially rendered page, we instead show a spinner until content is ready.

- Added an overlay with a spinner to the generated page markup
- Added method `fadeOutSpinner` to `setup.js` to remove the overlay when the page is done rendering
- Updated the expected site generated to include the spinner markup and related assets

spinner 1
Example of spinner being shown (slow!)

- This spinner is shown briefly on every page load - will this be annoying to the user?
- CSS for the spinner is taken from https://www.w3schools.com/howto/howto_css_loader.asp - do we need to attribute this in the source code?

**Testing instructions:**

Load any page. The spinner should appear while the page is rendering, and fade out when the page is done rendering.


asset/css/markbind.css
```diff
@@ -121,3 +121,32 @@ footer {
...
+    position: fixed;
+    z-index: 100;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    background: white;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
yamgent: Attributes should be listed in the alphabetical order.

asset/css/markbind.css
```diff
@@ -121,3 +121,32 @@ footer {
...
+    border-radius: 50%;
+    width: 120px;
+    height: 120px;
+    animation: spin 2s linear infinite;
+}
+
+@keyframes spin {
+    0% { transform: rotate(0deg); }
+    100% { transform: rotate(360deg); }
+}
```
yamgent: As stated earlier, we can use the Bootstrap's spinner, so these should no longer be necessary.

asset/js/setup.js
```diff
@@ -33,10 +33,15 @@ function setupAnchors() {
   });
 }
 
+function fadeOutSpinner() {
+  jQuery('#spinner-overlay').fadeOut(100);
```
yamgent: As stated earlier, this should be `.remove()` rather than fade out, to make it more pleasant for the user to navigate.

src/Page.js
```diff
@@ -449,6 +450,11 @@ Page.prototype.insertFooter = function (pageData) {
   return `${pageData}\n${nunjucks.renderString(footerContent, userDefinedVariables)}`;
 };
 
+
+Page.prototype.insertSpinner = function (pageData) {
+  return `${pageData}\n${SPINNER_HTML}`;
+};
```
yamgent: The addition of the spinner need not be done through code, because the HTML content is always going to be the same.

Please add the spinner directly inside `page.ejs` instead.

damithc:
Thanks for tackling this issue @marvinchin 

As @yamgent the current implementation somehow gives the impression that the website is 'slow'. The main factor is there is a blank page (and a spinner) shown for a while; other websites don't usually do that.

Although this solution was suggested by me, now I wonder if the current status quo (i.e., showing partial content) is actually better than the solution (w.r.t. user perception)? The main problem with the current status is that the partially rendered menus look ugly/broken. Is there a way to fix that instead so that the partial rendering doesn't look so bad?

marvinchin:
The main cause of the unstyled content appears to be waiting for the vue and vue-strap files to load. This delay is more prominent in dev environment as we aren't doing compression or caching (if this is where the original observation came from, this may be the reason). The delay doesn't seem too visible to me in production. 

Users on a slow network might still face this problem. If this is a concern, we could take the approach suggested by @yamgent and show the spinner only if the assets have not been loaded after 1 second. 

Happy to further discuss how else we can tackle this problem, or if we should handle it at all!

damithc:
> The delay doesn't seem too visible to me in production.

It depends on the site I think. e.g., this always show a broken page briefly when loading for the first time https://nus-cs2103-ay1819s2.github.io/cs2103-website/

marvinchin:
That's true, first loads might always face this issue as the scripts aren't cached locally yet. 

Do you have any thoughts about showing the spinner only after loading for some time then? We should also hide the content until it is fully rendered. This should avoid the user seeing the unstyled content without showing the spinner unless loading is taking too long.

damithc:
> That's true, first loads might always face this issue as the scripts aren't cached locally yet.

Try viewing a different page using the `Schedule` dropdown of https://nus-cs2103-ay1819s2.github.io/cs2103-website/. The problem happens in every page.

yamgent:
> The main problem with the current status is that the partially rendered menus look ugly/broken. Is there a way to fix that instead so that the partial rendering doesn't look so bad?

We can add temporary classes to the components, before vue-strap finishes loading and starts to replace the components with actual implementations.

In the case of a `<navbar>`, we can try to add Bootstrap navbar classes to it so that at least they are positioned correctly while vue-strap is loading:

```javascript
$('navbar').addClass('navbar navbar-expand-md fixed-top');
```

marvinchin:
Do you think we should hide the content entirely until the required assets are done loading? Or is there some other way we should style them? Transitioning from temporary styles to the fully rendered styles may look strange otherwise.

On the note of reducing the time taken to fully render the page, I've taken a closer look at the CS2103 website, and it seems like rendering is blocked until `siteData.json` is loaded. The problem seems to be that we are loading images etc. before `siteData.json`. As the `siteData.json` is used only for search and is not user facing, we might be able to load it asynchronously to avoid it blocking the render.

yamgent:
> Do you think we should hide the content entirely until the required assets are done loading? Or is there some other way we should style them? Transitioning from temporary styles to the fully rendered styles may look strange otherwise.

We can add on top of the styles I have mentioned with `visibility: hidden;` so that they stay hidden but the position of the content do not jump around too much. 

I think it is fine to not hide the words, since the user will still be allowed to get a basic sense of the website's content, regardless of whether the style has finished rendering or not. When I visit websites, things that appear "hidden", only to show up later, are usually always ads. :P

marvinchin:
Noted, will make a separate PR for these changes! Thanks for the inputs! 🙂

marvinchin:
Hi @yamgent @damithc I've made the changes requested. To summarise, we 1. hide content until it is ready to be rendered fully, and 2. show a spinner if rendering is not completed after 1 second.

Please do let me know if this behaviour is acceptable or if I should make any other changes!

src/template/page.ejs
```diff
@@ -6,22 +6,27 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title><%= title %></title>
-    <link rel="stylesheet" href="<%- asset.bootstrap %>">
-    <link rel="stylesheet" href="<%- asset.bootstrapVue %>">
-    <link rel="stylesheet" href="<%- asset.fontAwesome %>" >
-    <link rel="stylesheet" href="<%- asset.glyphicons %>" >
-    <link rel="stylesheet" href="<%- asset.highlight %>">
```
yamgent: These should not be moved to the bottom:
1. The convention is to put stylesheets at the head.
2. Authors may need to override the stylesheets, and they currently do that by adding additional stylesheets in head. Moving it to the bottom prevents them from overriding the stylesheets.

src/template/page.ejs
```diff
@@ -6,22 +6,27 @@
...
     <link rel="stylesheet" href="<%- asset.markbind %>">
-    <link rel="stylesheet" href="<%- asset.layoutStyle %>">
-    <% if (siteNav) { %><link rel="stylesheet" href="<%- asset.siteNavCss %>"><% } %>
```
yamgent: Same comment as previous

src/template/page.ejs
marvinchin: The reason I moved them to the bottom is to reduce the time taken to start showing the spinner. Otherwise, we have to wait for the other much larger assets to load before the spinner is shown. On slower connections this delay can be several seconds. 

If this issue is not a concern, I will remove these changes!

yamgent: The main bottleneck of the loading process is vue-strap and bulky/embedded page content and assets (e.g. youtube videos). Stylesheets themselves are fine and do not usually take up most of the loading time.

marvinchin: Will remove these changes!

marvinchin:
Thanks for looking at it!

Also, just to update - @damithc wants to compare the difference in perceived load times between showing the spinner vs just showing the semi-rendered page. If showing the spinner makes the page load look slower, we may choose not include this feature. I'll keep you posted once he has decided whether or not to go ahead with this!

marvinchin:
@damithc has given the go ahead to include this.

Thanks everyone for looking at this!

marvinchin:
@yamgent I believe we spoke about this and the decision was that we can assume our users use JavaScript. Do you think this use case is something we should account for?

If it is, we can modify the CSS and use some inline JavaScript to show this spinner only to users with JavaScript enabled.

pyokagan:
afaik usually server side rendering is used to deal with the FOUC problem.

<sup>(Notice how https://se-edu.github.io/learningresources/ doesn't have any FOUC :stuck_out_tongue: )</sup>

yamgent:
Vue, a major framework used by MarkBind, is using JavaScript. Website visitors can be assumed to be using JavaScript.

If the concern is that users with JavaScript disabled do not know why the web page is blank, we can add a warning message on the overlay to ask the user to enable JavaScript.

pyokagan:
No, the concern is that this unnecessarily handicaps MarkBind and prevents it from being used in websites that do not need any of the flashy Javascript-required features.

https://vuepress.vuejs.org/ doesn't need the reader to have Javascript, why should MarkBind?

yamgent:
> No, the concern is that this unnecessarily handicaps MarkBind and prevents it from being used in websites that do not need any of the flashy Javascript-required features.

Yes, you are right in that we can be more gentle to website that generally do not rely on dynamic content or dynamic features that MarkBind provides.

> https://vuepress.vuejs.org/ doesn't need the reader to have Javascript, why should MarkBind?

Thanks for mentioning the vuepress project. I checked their project and I think their idea of using Vue as a server-side rendering, rather than client-side rendering, is pretty nice since it allows them to reduce their reliance on JavaScript to purely just dynamic features. This is certainty an idea we can consider exploring in the future, so that Vue and vue-strap no longer need to be bundled together with MarkBind websites.

---

@marvinchin

> In the meantime, we could make the overlay and spinner visible only on browsers with javascript enabled to avoid having the page be entirely unusable for these users using the approach I suggested above.

 Pardon me for forgetting if we discuss this before, but if we do it in this order, the un-rendered content will flash for a second before the overlay is enabled right? I don't think that will look very nice to the user.

> Alternatively, if we don't foresee this enhancement bringing much value to users we could just shelve it for now.

Thanks for offering to shelve the enhancement, if we indeed end up doing so, do check with Prof to see if you can claim credit for the work you have done for this PR.

For the issue of the partial rendering temporary "breaking" the website, you can still consider the method that I have suggested earlier to make the website less ugly:

> We can add temporary classes to the components, before vue-strap finishes loading and starts to replace the components with actual implementations.
> 
> In the case of a `<navbar>`, we can try to add Bootstrap navbar classes to it so that at least they are positioned correctly while vue-strap is loading:
> 
> ```js
> $('navbar').addClass('navbar navbar-expand-md fixed-top');
> ```

yamgent:
> If we don't apply it across MarkBind immediately, maybe we can consider using it for the TopNav, as that seems to me the main cause of FOUC at the moment?

Sure, that seems like a reasonable start. @marvinchin interested?

yamgent:
@marvinchin there is a merge conflict with this PR, would need you to resolve it, thanks!

PR #596, opened 2019-01-19
CLI: Add command aliases

### LINKED ISSUE #566: CLI: Add command aliases
Inspired from npm's CLI. Instead of typing `npm install` and `npm test`, npm users can type `npm i` and `npm t`.

Perhaps we can support the same for MarkBind? Like so:

```bash
markbind b      # alias for 'markbind build'
markbind s -n   # alias for 'markbind serve --no-open'
markbind d      # alias for 'markbind deploy'
```

luyangkenneth:

Resolves #566

To enhance the user experience of the CLI

- Added aliases to all the CLI commands
- Updated the user guide

**Testing instructions:**

1) `markbind s` should be equivalent to `markbind serve` (and so on for the other commands)

2) `markbind --help` should automatically indicate that command aliases are available:

```
$ markbind --help

  Usage: markbind  <command>


  Options:

    -V, --version  output the version number
    -h, --help     output usage information


  Commands:

    init|i [root]                      init a markbind website project
    serve|s [options] [root]           build then serve a website from a directory
    deploy|d                           deploy the site to the repo's Github pages.
    build|b [options] [root] [output]  build a website
```

### DIFF
--- a/docs/userGuide/cliCommands.md
+++ b/docs/userGuide/cliCommands.md
@@ -30,9 +30,9 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
 
 ### `build` Command
 
-**Format:**: `markbind build [<OPTIONS>] `
+**Format:** `markbind [build | b] [<OPTIONS>] `
 
-**Description**: Generates the site to the directory named `_site` in the current directory.
+**Description:** Generates the site to the directory named `_site` in the current directory.
 
 **`OPTIONS`:**
 * `<output>`<br>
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -16,7 +16,8 @@ const fsUtil = require('./src/util/fsUtil');
 const logger = require('./src/util/logger');
 const Site = require('./src/Site');
 
-const ACCEPTED_COMMANDS = ['version', 'init', 'build', 'serve', 'deploy'];
+const ACCEPTED_COMMANDS = ['init', 'build', 'serve', 'deploy'];
```
luyangkenneth: Removed `version` from this list because it's technically not a command / `markbind version` is a no-op

index.js
```diff
@@ -205,7 +210,9 @@ program
 
 program.parse(process.argv);
 
-if (!program.args.length || !ACCEPTED_COMMANDS.includes(process.argv[2])) {
+if (!program.args.length
+  || !ACCEPTED_COMMANDS.includes(process.argv[2])
+  || !ACCEPTED_COMMANDS_ALIAS.includes(process.argv[2])) {
```
yamgent: The warning now shows for all commands, even the non-alias versions. The logic should be to check that it does not exist in both the non-alias **and** the alias list.

docs/userGuide/cliCommands.md
```diff
@@ -30,9 +30,9 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
 
 ### `build` Command
 
-**Format:**: `markbind build [<OPTIONS>] `
+**Format:** `markbind [build | b] [<OPTIONS>] `
```
yamgent: Currently we use square brackets to mean optional. The commands themselves aren't optional, so maybe use `<build | b>` instead? Or drop the brackets altogether: `build|b` (which is how `markbind --help` renders).

Or an alternative would be to specify it on a separate line? Something similar to [how npm docs does it](https://docs.npmjs.com/cli/install):

```markdown
**Format:** `markbind build [OPTIONS]`
**Alias:** `markbind b`
```

luyangkenneth: Ah, thanks for clarifying. I like how it looks on a separate line - just pushed the changes for that 🙂

PR #609, opened 2019-01-20
Fix siteNav example in documentation

### LINKED ISSUE #559: siteNav example in documentation is styled incorrectly

* **MarkBind Version:** 1.16.0

In our [documentation for siteNav](https://markbind.github.io/markbind/userGuide/tweakingThePageStructure.html#site-navigation-menus), the main level links are incorrectly highlighted:

Incorrectly highlighted siteNav example

The correct example should not have any links highlighted:

Corrected siteNav example

The reason is because we style links in the siteNav to be `black` using the `id` (`#site-nav a:link`) in `site-nav.css`, but the example in the documentation is not in an element with `id='site-nav'`. Thus, it takes on the default colour for links (`a`) specified in `bootstrap.min.css` which is `#007bff`.

Xenonym:

• [X] Documentation update

Resolves #559.

The documentation for the siteNav component and example code are styled incorrectly as they appear different from the actual siteNav. This may lead to user confusion.

I added a `.site-nav` class that only has styles related to links and had both the example siteNav in the docs and the real siteNav use that class style, so that the styles of both will be consistent across both. I also changed the styles of the example code given so that they reflect the correct indentation to use.

**Testing instructions:**
1. Render a page with the siteNav. It should have the same styles as before.
2. Check the example siteNav render and code to ensure that they appear identical to the actual siteNav render and code.

### DIFF
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -30,14 +30,16 @@
     -webkit-transition: 0.4s;
 }
 
-#site-nav a:active,
-#site-nav a:link,
-#site-nav a:visited {
+/*  Standard style order for links, see CSS2§5.11.3:
+    https://www.w3.org/TR/CSS2/selector.html#dynamic-pseudo-classes */
+.site-nav__a:link,
+.site-nav__a:visited,
+.site-nav__a:active {
     color: #565656;
     text-decoration: none;
 }
 
-#site-nav a:hover {
+.site-nav__a:hover {
     color: #0076FF;
 }
 
--- a/docs/_markbind/variables.md
+++ b/docs/_markbind/variables.md
@@ -1,6 +1,8 @@
-<span id="showBaseUrl">
+<span id="showBaseUrlCode">
 <code>{<span></span>{ baseUrl }}</code>
 </span>
+
+<span id="showBaseUrlText">{<span></span>{ baseUrl }}</span>
 
 <span id="markbind_blue">#00B0F0</span>
 
--- a/src/Page.js
+++ b/src/Page.js
@@ -141,6 +141,9 @@ function formatSiteNav(renderedSiteNav, src) {
   // Tidy up the style of the unordered list <ul>
   listItems.parent().attr('style', 'list-style-type: none; margin-left:-1em');
 
+  // Set class of <a> to ${SITE_NAV_ID}__a to style links
+  listItems.find('a[href]').addClass(`${SITE_NAV_ID}__a`);
+
   // Highlight current page
   const currentPageHtmlPath = src.replace(/\.(md|mbd)$/, '.html');
   listItems.find(`a[href='{{baseUrl}}/${currentPageHtmlPath}']`).addClass('current');
--- a/docs/userGuide/markBindSyntax.md
+++ b/docs/userGuide/markBindSyntax.md
@@ -322,14 +322,14 @@ More media blocks, embedding services and additional options can be found in [Ma
 
 <div id="intraSiteLinks">
 
-Links to files of the generated site (e.g., an HTML page or an image file) should be specified as absolute paths and should start with {{ showBaseUrl }} (which represents the root directory of the project).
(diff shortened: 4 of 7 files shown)


src/Page.js
```diff
@@ -487,7 +487,8 @@ Page.prototype.insertSiteNav = function (pageData) {
     siteNavDataSelector('navigation').replaceWith(formattedSiteNav);
   }
   // Wrap sections
-  const wrappedSiteNav = `<div id="${SITE_NAV_ID}">\n${siteNavDataSelector.html()}\n</div>`;
+  const wrappedSiteNav = `<div class="${SITE_NAV_ID}" id="${SITE_NAV_ID}">`
```
Chng-Zhi-Xuan: Although having the same name for IDs and classes works perfectly fine. We should avoid it as it is prone to confusion.

Since the class styles the `<a>` tag only, we can rename the class to `site-nav__link` ? 

[Reference to BEM naming convention](https://medium.freecodecamp.org/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849)

docs/_markbind/variables.md
```diff
@@ -1,7 +1,9 @@
-<span id="showBaseUrl">
+<span id="showBaseUrlCodeStyle">
 <code>{<span></span>{ baseUrl }}</code>
 </span>
 
+<span id="showBaseUrl">{<span></span>{ baseUrl }}</span>
```
Chng-Zhi-Xuan: How about `<code>{<span></span>{ baseUrl }}</code>` -> `{{ showBaseUrlCode }}`

and `{<span></span>{ baseUrl }}` -> `{{ showBaseUrlText }}`

asset/css/site-nav.css
```diff
@@ -114,7 +116,9 @@
 
 /* Hamburger button */
 
-.menu-top-bar, .menu-middle-bar, .menu-bottom-bar {
+.menu-top-bar, 
+.menu-middle-bar, 
+.menu-bottom-bar {
```
Chng-Zhi-Xuan: Good that you integrated the style from a review comment in #560

src/Page.js
acjh: `site-nav__link` would be directly applied to a `link` though.

https://seesparkbox.com/foundry/bem_by_example

Xenonym: `site-nav__a`?

acjh: I mean `site-nav__*` should go directly on the element.

The selector shouldn't be `.site-nav__a a { }`.

Xenonym: I could change all generated links to follow `.site-nav__a`, but this probably means we should follow BEM in all our CSS files for consistency - is that what we want?

Alternatively, we could simply go with `.site-nav-link`/`.site-nav-a`.

acjh: I'm open to either. @yamgent?

yamgent: > I could change all generated links to follow `.site-nav__a`,

Let's go with this (i.e. the BEM naming convention).

asset/css/site-nav.css
```diff
@@ -30,14 +30,14 @@
     -webkit-transition: 0.4s;
 }
 
-#site-nav a:active,
-#site-nav a:link,
-#site-nav a:visited {
+.site-nav__a:link,
+.site-nav__a:visited,
+.site-nav__a:active {
```
yamgent: Should be alphabetical order actually? :P

Chng-Zhi-Xuan: > a:hover MUST come after a:link and a:visited
> a:active MUST come after a:hover
> [Source](https://www.w3schools.com/css/css_link.asp)

So active being the last is correct.

acjh: That's for the case where the link states have different styles.
I'm okay with that order here, but let's add that reference as a comment.

Xenonym: I have added a link to the relevant section in the CSS standard for a more authoritative reference!

Issue #589, opened 2019-01-16
List unused files


acjh:
**Tell us about your environment**

* **MarkBind Version:** 1.16.1

**What did you do?**

1. `markbind init` in an empty directory.
2. create an `unused.md` file.
3. `markbind build`

**What did you expect to happen?**

MarkBind should be able to inform me of unused files, so that I can keep my files up to date.
- This may be specified by a flag on the `build` command, or a separate command altogether.
  - Probably for performance reasons.
- I should be able to ~ignore~ *whitelist* specific files that I want to keep, so that these don't appear.

**What actually happened? Please include the actual, raw output.**

```shell
  __  __                  _      ____    _               _ 
 |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
 | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
 | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
 |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|
                                                           
 v1.16.1
info: Website generation started at 00:00:00
info: Building assets...
info: Assets built
info: Generating pages...
[=] 1 / 1 pages built
info: Pages built
info: Website generation complete! Total build time: 0.125s
info: Build success!
info: Site data built
```

acjh:
This may be implemented by abstracting a `FileManager` class that tracks the read and copied files.

Related: #138 proposes having a static instance of file cache, which we can do under the hood of a `FileManager` class.

The implementation should allow us to enable/disable the tracking and caching, so that we can evaluate the performance costs.

PR #614, opened 2019-01-22
Add informative error for cyclic dependency

### LINKED ISSUE #594: include: cyclic references cause stack overflow
* NodeJS v9.3.0
* **MarkBind Version:** v1.16.1

**a.md**
```html
<include src="b.md" />
```
**b.md**
```html
<include src="a.md" />
```

The includes should be resolved automatically. At least, a user-friendly error should given, to allow the user to understand what happens.

MarkBind errored with a RangeError.
```
error: RangeError: Maximum call stack size exceeded
Error while preprocessing '/path/to/file/a.md'
```
However, the RangeError and "Maximum call stack size" error isn't helpful in determining what went wrong.

sijie123:

• [x] Other, please explain: Not exactly a bug fix / new feature, but an enhancement to current error message. Should I tag it under Enhancement?

Resolves #594.

Currently, authors might unintentionally create cyclic references in their `<include />` tag. For example, two files 
a.md
```
<include src="b.md" />
```
and b.md
```
<include src="a.md" />
```
would cause MarkBind to error with an uninformative RangeError: Maximum call stack size exceeded.

I created a custom error object (that extends Javascript's Error), which keeps track of the last 5 files processed before an error occurred. This helps the user identify the cyclic dependency loop.
When a cyclic dependency occurs, users would see
```
error: Cyclic reference detected.
Error while preprocessing '/path/to/file/a.md'
Last 5 files processed: 
        /path/to/file/a.md
        /path/to/file/b.md
        /path/to/file/a.md
        /path/to/file/b.md
        /path/to/file/a.md
```
I'm not sure if this is the best way to implement cyclic dependency checks. Please let me know if you can think of a better way to keep track and print visited files, or if creating the custom error class is not appropriate.
**[NEW]** I've implemented our custom call stack to keep track of the files. This enables me to throw an error when its above 200 items, and additionally keep track of the last 5 files. As a future extension, perhaps we can extend this into a full-blown file manager class.
**[NEW]** thanks @nicholaschuayunzhi : I realise there's no issue when including dynamic cyclic reference during the preprocess stage. Dynamic references cause the error only when Page.js tries to `collectHeadingsAndKeywordsInContent`. Thanks for helping me spot the gap; I've also added try/catch on the error there, although the error message is not as informative because there's no longer a recursion tree.

**Testing instructions:**
Ensure that you have fetched my PR, then create two files: `a.md` and `b.md` with contents as above.


nicholaschuayunzhi:
Thanks for the PR @sijie123 !

I tried a special case where

`a.md`
```md
<include src="b.md" />
```
`b.md`
```md
<include src="a.md" dynamic />
```

Error:

```
Unhandled rejection TypeError: Cannot read property 'join' of undefined
    at RangeError.CyclicReferenceError.toString (path\to\project\markbind\src\lib\markbind\src\parser.js:47:79)
    at Object.warn (path\to\project\markbind\src\util\logger.js:30:42)
    at rejectHandler (path\to\project\markbind\src\Site.js:142:10)
    at readSiteConfig.then.then.then.then.then.then.then.then.then.then.catch (path\to\project\markbind\src\Site.js:472:9)
    at tryCatcher (path\to\project\markbind\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (path\to\project\markbind\node_modules\bluebird\js\release\promise.js:513:31)
    at Promise._settlePromise (path\to\project\markbind\node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (path\to\project\markbind\node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (path\to\project\markbind\node_modules\bluebird\js\release\promise.js:689:18)
    at Async._drainQueue (path\to\project\markbind\node_modules\bluebird\js\release\async.js:133:16)
    at Async._drainQueues (path\to\project\markbind\node_modules\bluebird\js\release\async.js:143:10)
    at Immediate.Async.drainQueues [as _onImmediate] (path\to\project\markbind\node_modules\bluebird\js\release\async.js:17:14)
    at runCallback (timers.js:756:18)
    at tryOnImmediate (timers.js:717:5)
    at processImmediate [as _immediateCallback] (timers.js:697:5)
```

Could you help to investigate? Thank you!

Original behaviour prints `maximum call stack exceeded`

src/lib/markbind/src/parser.js
```diff
@@ -266,7 +285,19 @@ Parser.prototype._preprocess = function (node, context, config) {
...
+      } catch (e) {
+        if (e.message === 'Maximum call stack size exceeded') {
+          const err = new CyclicReferenceError();
+          err.addFileToStack(childContext.cwf);
+          throw err;
+        } else if (e instanceof CyclicReferenceError) {
+          e.addFileToStack(childContext.cwf);
+          throw e;
+        }
+        throw e;
```
acjh: Interesting implementation. Let's move `CyclicReferenceError` into its own file and implement a `CyclicReferenceError.handle` class method, so that we can do this:

```js
try {
  element.children = element.children.map(e => self._preprocess(e, childContext, config));
} catch (e) {
  CyclicReferenceError.handle(e);
}
```

However, this implementation still waits for a call stack error to occur and depends on a hardcoded `e.message` check.
I wonder if a `FileManager` (suggested in #589) would help. We can proactively track and ensure acyclic file references.

yamgent: > I wonder if a `FileManager` (suggested in #589) would help.

I agree with letting FileManager handling this problem. @sijie123 might want to consider #589 first?

sijie123: Hmm, yea, I agree that it's kinda odd to wait for the error to occur and then checking, where we wcould have done something proactively first.

However, in that case FileManager would have to implement a subroutine similar to preprocess in order to detect any cyclic references, or even unused files for that matter.
I think it might be easier for me to implement a solution to the cyclic reference problem, then apply it to FileManager later.

src/lib/markbind/src/handlers/cyclicReferenceError.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  toString() {
+    return `${this.message} \nLast 5 files processed: ${'\n\t'}${this.fileStack.join('\n\t')}`;
+  }
+}
+
+function handle(err, filePath) {
+  if (err.message === 'Maximum call stack size exceeded') {
```
yamgent: Hmm, this is still too reliant on Nodejs handling of stack overflow errors. Their behaviour does not have to be consistent across different Nodejs versions, and Nodejs developers can change this any time they want, which means our feature may break.

I think it would be better if we are able to keep a counter ourselves, and to stop parsing if we hit a certain depth (e.g. if the depth goes beyond 200, then stop and throw error). That way, we do not rely on Nodejs to crash for detecting cyclic references.

test/unit/parser.test.js
```diff
@@ -298,6 +298,97 @@ test('includeFile replaces <include src="include.md#doesNotExist" optional> with
   expect(result).toEqual(expected);
 });
 
+test('includeFile detects cyclic references for static includes', async () => {
```
yamgent: for static includes -> for static **cyclic** includes

test/unit/parser.test.js
```diff
@@ -298,6 +298,97 @@ test('includeFile replaces <include src="include.md#doesNotExist" optional> with
...
+  const expected = [
+    '# Index',
+    `<div style="color: red">${expectedErrorMessage}</div>`,
+    '',
+  ].join('\n');
+
+  expect(result).toEqual(expected);
+});
+
+test('includeFile processes successfully for dynamic includes', async () => {
```
yamgent: for dynamic includes -> for dynamic cyclic includes

src/lib/markbind/src/handlers/cyclicReferenceError.js
sijie123: Thanks! I agree that this err.message sounds really hacky.
I've now implemented our own call stack in a newer commit.

sijie123:
Updated, and ready for review.

src/Page.js
```diff
@@ -332,22 +333,23 @@ Page.prototype.collectHeadingsAndKeywords = function () {
...
+  this.collectHeadingsAndKeywordsInContent($(`#${CONTENT_WRAPPER_ID}`).html(), null, false, []);
 };
 
 /**
  * Records headings and keywords inside content into this.headings and this.keywords respectively
  * @param content that contains the headings and keywords
  */
-Page.prototype.collectHeadingsAndKeywordsInContent = function (content, lastHeading, excludeHeadings) {
+Page.prototype.collectHeadingsAndKeywordsInContent = function (content, lastHeading, excludeHeadings,
+                                                               callStack) {
```
yamgent: `callStack` sounds a bit weird (because it is not directly relevant to the pages), name it `sourceTraversalStack`?

sijie123: Just tested on CS2103T website:
```
info: Building assets...
info: Assets built
info: Generating pages...
[=========================================================================================================================================================] 714 / 714 pages buil
info: Pages built
info: Website generation complete! Total build time: 170.116s
http://0.0.0.0:8080 is already in use. Trying another port.
info: Site data built
info: Serving "/workspace/cs2103-website/_site" at http://127.0.0.1:43455
```
All's good :D

sijie123: callStack -> sourceTraversalStack: Changed. Thanks!

src/lib/markbind/src/handlers/cyclicReferenceError.js
```diff
@@ -0,0 +1,20 @@
+const MAX_RECURSIVE_DEPTH = 200;
+const ERROR_MESSAGE = 'Cyclic reference detected.';
+
+class CyclicReferenceError extends Error {
+  static get MAX_RECURSIVE_DEPTH() {
+    return MAX_RECURSIVE_DEPTH;
```
nicholaschuayunzhi: Personally i feel that writing a static getter is the same as writing a constant. We could remove the constant above and update the getter as:

```suggestion
    return 200;
```

src/lib/markbind/src/handlers/cyclicReferenceError.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  static get ERROR_MESSAGE() {
+    return ERROR_MESSAGE;
```
nicholaschuayunzhi: This static class variable is not used. We can remove it. Alternatively we could treat it as static class variable declaration like: 
```suggestion
    return 'Cyclic reference detected.;
```

And use `CyclicReferenceError.ERROR_MESSAGE` in the constructor.

sijie123: Hmm, I was of the impression that we should declare the ERROR_MESSAGE like a constant, i.e. `const ERROR_MESSAGE = ...`
But I guess, since we're only going to use it once, it doesn't make too much sense to declare it this way.
Removed. Thanks!

yamgent:
@sijie123 will need you to resolve conflict.

sijie123:
Conflict resolved.

Issue #620, opened 2019-01-23
usingComponents.md: Warn user not to set enableSearch to false if using searchbar


marvinchin:
**Tell us about your environment**

* **MarkBind Version:** v1.17.0

**What did you do? Please include the actual source code causing the issue.**
Set `enableSearch` to `false` in `site.json`

```json
{
  ....
  "enableSearch": false
}
```

**What did you expect to happen?**
Site should load without search

**What actually happened? Please include the actual, raw output.**
Site does not load
The following error is shown in the Chrome console:
![screenshot 2019-01-23 at 2 05 34 pm](https://user-images.githubusercontent.com/17447681/51586380-ff554d00-1f17-11e9-9cfb-f18b46162572.png)

marvinchin:
Site is able to load after the `searchbar` component is removed from `header.md`. Perhaps we should note in the documentation that if the `searchbar` component is used, then `enableSearch` must be set to true in `site.json`.

yamgent:
> Perhaps we should note in the documentation that if the `searchbar` component is used, then `enableSearch` must be set to true in `site.json`.

We can put a warning on the searchbar component in the "Using Components" reference page to warn the user not to set `enableSearch` to false.

Xenonym:
@marvinchin do you mean that the site fails to load when there is _both_ a `<searchbar>` component that uses `:data="searchData"` *and* when `"enableSearch": false`? I don't believe there's a restriction that `<searchbar>` can only be used in the header or that it can only use `:data="searchData"`.

Second error is reported in #619.

marvinchin:
@Xenonym yes, that seems to be the source of the error. Thanks for clarifying!

Xenonym:
In that case, it seems the issue is that `<searchbar>` does not react well to being passed an invalid or nonexistent `:data` array. 

You can use `<searchbar>` with `"enableSearch": false`, so long as you do not refer to a nonexistent `searchData`. Is that a valuable use case? I can imagine a user might build a custom `searchData` for a searchbar that is only on a single page (maybe a searchbar for a very long glossary page, for example).

marvinchin:
Yes, I believe that the intention of allowing the user to specify the `:data` attribute would be to allow the user to specify the source of the search data.

I think the problem here might be that the connection between a  `<searchbar>` component with attribute `:data="searchData"`, and setting `"enableSearch": true` in `site.json` was not explicitly mentioned in the user documentation. I'll make the necessary updates to include this!

PR #621, opened 2019-01-23
Load site data asynchronously

### LINKED ISSUE #608: Render blocked on loading siteData.json

Open any MarkBind generated site with search enabled.

Page should render before fetching `siteData.json`.

`siteData.json` is fetched before the page is fully rendered, and the render blocks until `siteData.json` is fetched.

marvinchin:

Resolves #608.

To avoid having render being blocked on waiting for `siteData.json` to load, we first render the page then load it asynchronously.

- Embed the `enableSearch` variable in the HTML document
- Update `setup.js` to fetch `siteData.json` asynchronously

I've embedded the `enableSearch` variable in the HTML document to allow us to decide whether to use `setup()` or `setupWithSearch()` without first having to load `siteData.json` to preserve the original behavior. However, I was wondering whether or not it would be possible for us to simply use `setupWithSearch()` for both cases, since sites with search disabled cannot use the `searchbar` component anyway (see #620). This would allow us to avoid having to embed `enableSearch` in the HTML document.

**Testing instructions:**
Observe the chrome network tab. `siteData.json` should be loaded after the page is rendered completely.

### DIFF
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -36,6 +36,15 @@ function setupAnchors() {
 function removeLoadingOverlay() {
   jQuery('#loading-overlay').remove();
 }
+
+function getSearchData(vm) {
+  jQuery.getJSON(`${baseUrl}/siteData.json`)
+    .then((siteData) => {
+      // eslint-disable-next-line no-param-reassign
+      vm.searchData = siteData.pages;
+    });
+}
+
 function executeAfterMountedRoutines() {
   flattenModals();
   scrollToUrlAnchorHeading();
@@ -84,11 +93,7 @@ function setup() {
   setupPageNav();
 }
 
-function setupWithSearch(siteData) {
-  if (!siteData.enableSearch) {
-    setup();
-    return;
-  }
+function setupWithSearch() {
   const { searchbar } = VueStrap.components;
   // eslint-disable-next-line no-unused-vars
   const vm = new Vue({
@@ -98,7 +103,7 @@ function setupWithSearch(siteData) {
     },
     data() {
       return {
-        searchData: siteData.pages,
+        searchData: [],
       };
     },
     methods: {
@@ -110,12 +115,15 @@ function setupWithSearch(siteData) {
     },
     mounted() {
       executeAfterMountedRoutines();
+      getSearchData(this);
     },
   });
   setupSiteNav();
   setupPageNav();
 }
 
-jQuery.getJSON(`${baseUrl}/siteData.json`)
-  .then(siteData => setupWithSearch(siteData))
-  .catch(() => setup());
+if (enableSearch) {
+  setupWithSearch();
+} else {
+  setup();
+}
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -37,7 +37,7 @@ function removeLoadingOverlay() {
   jQuery('#loading-overlay').remove();
 }
 
-function getSearchData(vm) {
+function updateSearchData(vm) {
   jQuery.getJSON(`${baseUrl}/siteData.json`)
     .then((siteData) => {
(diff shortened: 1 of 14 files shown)


test/test_site/expected/bugs/index.html
```diff
@@ -6,13 +6,13 @@
...
-    <link rel="stylesheet" href="..\markbind\css\github.min.css">
-    <link rel="stylesheet" href="..\markbind\css\markbind.css">
-    <link rel="stylesheet" href="..\markbind\layouts\default\styles.css">
+    <link rel="stylesheet" href="../markbind/css/bootstrap.min.css">
+    <link rel="stylesheet" href="../markbind/css/bootstrap-vue.min.css">
+    <link rel="stylesheet" href="../markbind/css/font-awesome.min.css" >
+    <link rel="stylesheet" href="../markbind/css/bootstrap-glyphicons.min.css" >
+    <link rel="stylesheet" href="../markbind/css/github.min.css">
+    <link rel="stylesheet" href="../markbind/css/markbind.css">
+    <link rel="stylesheet" href="../markbind/layouts/default/styles.css">
```
yamgent: Just curious, which OS are you on? We should be consistent with the slash direction, so let's stick to `\`.

yamgent: We should be consistent with the slash direction, so let's stick to `\`.

marvinchin:
@yamgent I've made the changes!

test/test_site/expected/bugs/index.html
acjh: We should use `/`, especially for URLs. Is `\` generated by Windows here?

yamgent: Yes `\` is generated by Windows. We currently update test files by copying over content from `_site` to `expected`, so the slashes are going to change depending on the OS that the person is using when he/she updates the tests, which is not really ideal...

Maybe code-wise we should standardize how MarkBind output slashes for file paths (can be a separate PR for that).

acjh: Agreed.

marvinchin: @acjh @yamgent could I check if I should change the slash direction in my PR, or should we leave it to be fixed in another PR?

Thanks for looking into it!

yamgent: > could I check if I should change the slash direction in my PR, or should we leave it to be fixed in another PR?

Those will be dealt with by the person that is resolving https://github.com/MarkBind/markbind/issues/628.

marvinchin:
@yamgent if we do not use `enableSearch`, we can simply make the call to fetch `siteData.json` regardless of whether or not search is enabled. If `enableSearch` is disabled, then the call will be unnecessary but should not affect the rest of the site.

I suppose this is a tradeoff between making an unnecessary call, and injecting the `enableSearch` into the document. I was just thinking that injecting the variable into the document felt a little dirty, so I was thinking if there was any way to avoid that 😛

marvinchin:
If that's the case, perhaps not unifying `setup()` and `setupWithSearch()` might still make sense, since that way we do not need to add the `searchbar` component or the required attributes/methods to the vue instance we are creating if `enableSearch` is false.

asset/js/setup.js
```diff
@@ -33,6 +33,14 @@ function setupAnchors() {
   });
 }
 
+function getSearchData(vm) {
```
yamgent: Rename `getSearchData()` to `updateSearchData()` (methods with names that start with `get` are usually accessor methods).

marvinchin:
@yamgent fixed!

marvinchin:
@yamgent could I check if we are intending to merge this? Just wanted to let you know that I've resolved the merge conflicts! 🙂

PR #623, opened 2019-01-23
Fix unexpected indentation in the first line of fenced text

### LINKED ISSUE #618: Unexpected indentation in the first line of fenced text
v1.17.0

Current:
image
Code:
image

Problem: there seems to be an additional space at the start of the first line

amad-person:

• [X] Bug fix

Fixes #618.

Fenced code blocks seem to have an unexpected indentation in the first line. 
As discussed in #618, this is because the `<code>` tag does not have the  `display: block` style attribute. Since the `hljs` class has this style tag by [default](https://github.com/highlightjs/highlight.js/blob/master/src/styles/default.css), this bug can be fixed by adding `class="hljs"` to the `<code>` tag.

In `/markdown-it/index.js#15`,

`return '<pre class="hljs"><code>' + markdownIt.utils.escapeHtml(str) + '</code></pre>';` 

was changed to:

`return '<pre><code class="hljs">' + markdownIt.utils.escapeHtml(str) + '</code></pre>';`

Note that this bug only arises when no language is specified after the backticks. If a language is specified, then the current code assigns the `hljs` class to `<code>`.

The syntax for adding code blocks remains the same, only the way they are being displayed will change.

Make sure the indentation is correct for fenced code blocks, and that inline code blocks are not being affected by this change.

**Testing instructions:**

Add a fenced code block (with no language specified) and build the website. Check that there is no unexpected indentation in the first line of the block.

### DIFF
--- a/src/lib/markbind/src/lib/markdown-it/index.js
+++ b/src/lib/markbind/src/lib/markdown-it/index.js
@@ -12,7 +12,7 @@ const markdownIt = require('markdown-it')({
       }
     }
 
-    return '<pre class="hljs"><code>' + markdownIt.utils.escapeHtml(str) + '</code></pre>';
+    return '<pre><code class="hljs">' + markdownIt.utils.escapeHtml(str) + '</code></pre>';
   }
 });
 const slugify = require('@sindresorhus/slugify');


damithc:
In addition to unwanted indentation, there is also an unnecessary scrollbar. I hope both problems are related and fixed by this PR?
image

@yamgent can we release this soon? I have a bunch of exercises where the 'expected output' is affected by the extra indentation problem, potentially causing confusion among students.

damithc:
> In addition to unwanted indentation, there is also an unnecessary scrollbar. I hope both problems are related and fixed by this PR?

Actually, the scrollbar doesn't appear in _every_ case of extra indentation. Not sure what's causing it. 
Examples: https://nus-te3201.github.io/2019/programming/toc/lists.html

yamgent:
> Actually, the scrollbar doesn't appear in _every_ case of extra indentation. Not sure what's causing it.

The latest release which contains this PR should fix this problem. The scrollbar was caused by some styling issues related to paddings.

damithc:
> The latest release which contains this PR should fix this problem. The scrollbar was caused by some styling issues related to paddings.

Yes, seems to fixed. Good work @amad-person and reviewers.

PR #627, opened 2019-01-23
Implement page variable functionality

### LINKED ISSUE #575: Allow declaration of variable values in the same source file
From https://github.com/MarkBind/markbind/pull/553#issuecomment-451707842.

~~`index.md`:~~

~~\<span id="color">red\</span>~~
~~\<font color="{{ color }}">my text\</font>~~

~~However, a better syntax is needed, because outside of `variables.md`, `<span>`s are not always variables.~~

---

EDIT:

Per further discussions below, support such a syntax:

`index.md`:
```html
<variable id="color">red</variable>

<font color="{{ color }}">my text</font>
```

jamos-tay:

Fixes #575 

Extension of included variable functionality, allow users to specify local variables within the same page.

Users can specify local variables with the `variable` key word:

```html
<variable name="color">red</variable>
```

They can use it within the page like normal variables.

Should variables cascade into includes? I'm thinking no, it can be confusing if the user is trying to trace where a variable came from and can't find it. If the user wishes for them to be used they can include it specifically:

```html
<variable name="x">5</variable>
<include ... >
  <span id="x">{{ x }}</span>
</include>
```

---

All page variables are processed before any rendering is done, so users cannot change the value of variables mid page:

```
<variable name="x">5</variable>
{{ x }} // 6
<variable name="x">6</variable>
{{ x }} // 6
```

I don't think supporting this is a good idea since we won't be able to use nunjucks and would have to write our own parser. Users can use `{% set %}` as an alternative. Currently a warning is shown if the user tries to reassign a variable.


damithc:
> All page variables are processed before any rendering is done, so users cannot change the value of variables mid page:

should we call them constants instead? 

`<const name="MAX_SIZE">5</const>`

jamos-tay:
Aren't the global variables kind of like constants too, in that case?

damithc:
> We could, but aren't the global variables kind of like constants too, in that case?

At the moment, they are too.

yamgent:
> should we call them constants instead?
> 
> `<const name="MAX_SIZE">5</const>`

But if the users can do a `{% set %}` on them, then it is confusing to call them constants? :P

acjh:
We may also consider implementing overriding the sub-sites' variables in future.

damithc:
> But if the users can do a `{% set %}` on them, then it is confusing to call them constants? :P

That's true.

We have several types of variables each behaving in a different way; we should come up with a good terminology to differentiate them.

jamos-tay:
Sure - maybe we should just leave it as variable for now then since that's how `variables.md` is named?

jamos-tay:
Implemented page variables + included variables, should be ready

test/test_site/_markbind/variables.md
```diff
@@ -10,3 +10,5 @@
 
 <span id="global_variable_overriding_included_variable">Global variable overriding included variable</span>
 <span id="included_global_variable">Global variable</span>
+<span id="page_global_variable">Global Variable</span>
```
yamgent: In context of the test that uses the variable, we can just use `included_global_variable` rather than declaring a separate `page_global_variable`?

docs/userGuide/reusingContents.md
```diff
@@ -157,6 +157,20 @@ You must use the `safe` filter when using such variables:
 <code>{<span></span>{ right_hand_2 }}</code> {{ icon_arrow_right }} {{glyphicon_hand_right}}
 </div>
 
+### Page Variables
+
+**You can also declare variables for use within a single page.** These variables work exactly the same as regular variables, except they only apply to the page they are declared in. This can be done by using the `variable` tag.
```
yamgent: ...except **that their values** only apply...

`variable` -> `<variable>`

docs/userGuide/reusingContents.md
```diff
@@ -157,6 +157,20 @@ You must use the `safe` filter when using such variables:
...
+
+<div class="indented">
+
+{{ icon_example }} Declaring page variables<br>
```
yamgent: End with a colon.

test/test_site/testPageVariables.md
```diff
@@ -0,0 +1,7 @@
+<variable name="included_variable_overriding_page_variable">Page Variable Overridden by Included Variable</variable>
+
+# Page Variables Should Not Leak Into Inner Files
+Should be blank: {{ page_variable }}
```
yamgent: The current behavior seems to have no difference with `{% set ...=... %}` (the only additional benefit we get is that you can declare it in any part of the page, but that's it :/).

Is it not possible to bring page variable into included files?

yamgent: Also the name of this file is somewhat misleading; it tests the page variables w.r.t. it being used inside an `<include>`, not as a _general_ page for testing page variables.

jamos-tay: Ahh yes bringing them in is not a problem, but I was wondering if we should actually in the original post:

>Should variables cascade into includes? I'm thinking no, it can be confusing if the user is trying to trace where a variable came from and can't find it. If the user wishes for them to be used they can include it specifically:
>
> ```html
> <variable name="x">5</variable>
> <include ... >
>   <span id="x">{{ x }}</span>
> </include>
> ```

jamos-tay:
Updated, page variables can now cascade into included files.

Page variables are now treated exactly like included variables, global > outer > inner priority etc.. The only difference is how they're defined (span vs variable)

test/test_site/testPageVariablesInInclude.md
```diff
@@ -0,0 +1,15 @@
+# Outer Page Variable Overriding Inner Page Variable
+<variable name="outer_page_variable_overriding_inner_page_variable">Inner Page Variable Overridden by Outer Page Variable</variable>
```
yamgent: Name should be `"page_variable"`?

docs/userGuide/reusingContents.md
```diff
@@ -157,6 +157,20 @@ You must use the `safe` filter when using such variables:
...
+
+<div class="indented">
+
+{{ icon_example }} Declaring page variables:<br>
+
+`<variable name="full_name">John Doe</variable>`<br>
+<code>My name is {<span></span>{ full_name }}. This is {<span></span>{ full_name }}'s site.</code>
+</div>
+
+Note: These variables will cascade into [`include` files]({{ baseUrl }}/userGuide/reusingContents.html#the-include-tag). If the same variable is defined in a chain of pages (e.g. `a.md` includes `b.md` includes `c.md`...), variables defined in the top-most page will take precedence. Global variables (`_markbind/variables.md`) will take precedence over any page variables.
```
yamgent: `include` -> `<include>`

Also not sure whether the word "cascade" is a good word to use, but I can't think of a better word choice. 🤔

jamos-tay: These variables will also be applied to [`<include>` files]

jamos-tay:
Updated

damithc:
Good work on this feature @jamos-tay (and @yamgent, @acjh). I've started using it and worked great so far. What do you guys think of allowing the reuse context to access page variables using the `#` operator? 

e.g., `<include src="foo.md#title"/>` (where `title` is a page variable in `foo.md`)

acjh:
Isn't that the syntax for fragments?

damithc:
> Isn't that the syntax for fragments?

Yes. a variable is very similar like a hidden fragment, right? Of course name clashes is a concern but in that case we can follow the same rules as multiple segments with same ID?

jamos-tay:
Hmm... I think that implementation might have a few unwanted implications.

For example, if the reasoning is `<include src="foo.md#title"/>` renders `<variable name="title">` fragment in the parent page and therefore `title` should apply to the parent page, then wouldn't  `<include src="foo.md"/>` render all the variables in `foo.md` in the parent page and thus all inner variables should be applied? Which seems to be a case of an inner page modifying an outer page's content, which might go against what we want.

damithc:
> For example, if the reasoning is `<include src="foo.md#title"/>` renders `<variable name="title">` fragment in the parent page and therefore `title` should apply to the parent page, then wouldn't `<include src="foo.md"/>` render all the variables in `foo.md` in the parent page and thus all inner variables should be applied? Which seems to be a case of an inner page modifying an outer page's content, which might go against what we want.

Accessing from a variable _value_ from the parent doesn't mean the variable is raised to the level of the parent. It's simply reading the _value_ only.

I was hoping to use variables as _properties_ of the page. e.g., title, summary, level, difficulty etc. Similar to a public varaible of a Java object, it makes sense to be able to access properties of the object from outside.

Of course, I can use `<span id="title" class="d-none">abc</span>`  instead. Using variables is slightly more convenient.

jamos-tay:
Hmm okay, that seems reasonable

It would have to be treated separately from a regular include, though

damithc:
Moved the discussion to https://github.com/MarkBind/markbind/issues/682

PR #632, opened 2019-01-25
Add diff printing for easier debugging

nicholaschuayunzhi:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

When html file compare fails on travis, there was no useful infomation. Also the diff does not finish for all files if there is a diff in one file.

Print diffs to console. Red = removed from expected, green = added by new MarkBind, grey = common
Log can be viewed here:
https://travis-ci.org/nicholaschuayunzhi/markbind/builds/484199960?utm_source=github_status&utm_medium=notification

If code and log is understandable. I use [jsdiff](https://github.com/kpdecker/jsdiff#examples) but i do not use the `colors` library.

Consideration: we may move to jest snapshot testing in near future?

**Testing instructions:**
Add some diffs to html file and run test.

### DIFF
--- a/test/functional/testUtil/diffPrinter.js
+++ b/test/functional/testUtil/diffPrinter.js
@@ -1,11 +1,9 @@
-class DiffPrinter {
-  /* eslint-disable lodash/prefer-constant */
-  static get ANSI_RED() { return '\u001B[31m'; }
-  static get ANSI_GREEN() { return '\u001B[32m'; }
-  static get ANSI_GREY() { return '\u001B[90m'; }
-  static get ANSI_RESET() { return '\u001B[0m'; }
-  /* eslint-enable lodash/prefer-constant */
+const ANSI_RED = '\u001B[31m';
+const ANSI_GREEN = '\u001B[32m';
+const ANSI_GREY = '\u001B[90m';
+const ANSI_RESET = '\u001B[0m';
 
+class DiffPrinter {
   /**
    * Prints line of text in colour provided
    * @param {string} text text to print, default: no text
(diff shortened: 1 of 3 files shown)


test/test_site/testUtil/diffHtml.js
```diff
@@ -84,7 +86,9 @@ const diffHtml = (expected, actual) => {
     insidePath
     && startsWithClosedPath(fragment);
 
-  diff.forEach((part) => {
+  let error = false;
+  const diffPrinter = new DiffPrinter(filePathName);
+  diff.forEach((part, index) => {
```
yamgent: I did some extensive testing.

The new algorithm that uses `diffPrinter.printDiff()` seems to rely on the fact that the indices are sorted, but is that always truly the case? I was at one point able to find a case whereby this assumption seems to be violated, as it printed out two different parts of the file without the "...". I can't reproduce this bug again though, I don't know how to properly trigger it. :/

nicholaschuayunzhi:
@yamgent i have updated the code to check for path separators. The logic is also re-written to be more understandable. Currently the log is printed https://travis-ci.org/nicholaschuayunzhi/markbind/builds/489411595?utm_source=github_status&utm_medium=notification

Im not too sure why but there are some white text in the diff printing. Its not reproduce-able locally and the areas of white text changes when the test is re run on travis.

test/functional/testUtil/diffPrinter.js
```diff
@@ -0,0 +1,111 @@
+class DiffPrinter {
+  /* eslint-disable lodash/prefer-constant */
+  static get ANSI_RED() { return '\u001B[31m'; }
+  static get ANSI_GREEN() { return '\u001B[32m'; }
+  static get ANSI_GREY() { return '\u001B[90m'; }
+  static get ANSI_RESET() { return '\u001B[0m'; }
```
yamgent: > ```
> /* eslint-disable lodash/prefer-constant */
> ```

These can actually just be constants then? :P No good reason to put them as methods for now.

test/functional/testUtil/diffPrinter.js
```diff
@@ -0,0 +1,111 @@
...
+  /* eslint-enable lodash/prefer-constant */
+
+  /**
+   * Prints line of text in colour provided
+   * @param {string} text text to print, default: no text
+   * @param {string} colour colour of text, default: no colour
+   */
+  static printLine(text = '', colour = 'none') {
+    let ansiEscCode = '';
+    switch (colour) {
```
yamgent: Callers could probably just pass in `ANSI_RED`, `ANSI_GREEN`, etc directly if they are constants.

nicholaschuayunzhi: I'm thinking that the callers should not know that we're using ANSI escape code to render colour.

yamgent: You have a point.

PR #639, opened 2019-01-27
Add scrollspy only when pageNav is enabled

### LINKED ISSUE #619: JS error on default MarkBind site: Cannot read property 'scrollIntoView' of null

* **MarkBind Version:** 1.17.0

1. Create a new MarkBind site with `markbind init`.
2. Preview the site with `markbind serve`.

The site should preview normally with no errors.

An uncaught `TypeError` is logged by the JS console in Chrome:
```
setup.js:67 Uncaught TypeError: Cannot read property 'scrollIntoView' of null
    at jQuery.on (setup.js:67)
    at dispatch (vue-strap.min.js:39)
    at m.handle (vue-strap.min.js:39)
    at Object.trigger (vue-strap.min.js:39)
    at vue-strap.min.js:39
    at Function.each (vue-strap.min.js:28)
    at k.fn.init.each (vue-strap.min.js:28)
    at k.fn.init.trigger (vue-strap.min.js:39)
    at n.t._activate (bootstrap-utility.min.js:6)
    at n.t._process (bootstrap-utility.min.js:6)
```
`setup.js:67` is part of `setupPageNav`:
```js
65	function setupPageNav() {
66	jQuery(window).on('activate.bs.scrollspy', (event, obj) => {
67	    document.querySelectorAll(`a[href='${obj.relatedTarget}']`).item(0).scrollIntoView(false);
68	  }); 
69	}
```

I guess this has something to do with the page nav implementation.

Xenonym:

• [X] Bug fix

Resolves #619.

Previously, we added the scrollspy plugin for page navigation to all pages, regardless of whether or not page nav was enabled. This caused JS errors in the browser for pages without a valid page nav.

I added a check to `page.ejs` to only add scrollspy when the page nav is enabled.

**Testing instructions:**
1. `markbind serve` a page that does not have a page nav specified. The page should render normally without JS errors.
2. `markbind serve` a page that has a page nav specified. The page should render normally without JS errors, and the page nav should work normally.

### DIFF
--- a/src/template/page.ejs
+++ b/src/template/page.ejs
@@ -18,7 +18,7 @@
     <%- headFileBottomContent %>
     <% if (faviconUrl) { %><link rel="icon" href="<%- faviconUrl %>"><% } %>
 </head>
-<body data-spy="scroll" data-target="#page-nav" data-offset="100">
+<body <%_ if (pageNav) { %> data-spy="scroll" data-target="#page-nav" data-offset="100" <%_ } %>>
 <div id="loading-overlay">
     <div id="spinner-layout">
         <div id="spinner"></div>
--- a/src/Page.js
+++ b/src/Page.js
@@ -231,7 +231,7 @@ Page.prototype.prepareTemplateData = function () {
     faviconUrl: this.faviconUrl,
     headFileBottomContent: this.headFileBottomContent,
     headFileTopContent: this.headFileTopContent,
-    pageNav: this.frontMatter.pageNav,
+    pageNav: this.isPageNavigationSpecifierValid(),
     siteNav: this.frontMatter.siteNav,
     title: prefixedTitle,
   };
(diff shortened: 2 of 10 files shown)


src/template/page.ejs
```diff
@@ -18,7 +18,7 @@
     <%- headFileBottomContent %>
     <% if (faviconUrl) { %><link rel="icon" href="<%- faviconUrl %>"><% } %>
 </head>
-<body data-spy="scroll" data-target="#page-nav" data-offset="100">
+<body <%_ if (pageNav) { %> data-spy="scroll" data-target="#page-nav" data-offset="100" <%_ } %>>
```
Chng-Zhi-Xuan: Within `Page.js`

```Javascript
    headFileTopContent: this.headFileTopContent,
    pageNav: this.frontMatter.pageNav, // Should change this to be this.isPageNavigationSpecifierValid
    siteNav: this.frontMatter.siteNav,
```

Because the current logic to insert a `page-nav` is through that function. Page navigation is not inserted if the specifier has wrong syntax.

PR #643, opened 2019-01-28
Use only forward slashes for paths during build

### LINKED ISSUE #628: Always use forward slash for file paths when generating the website

* **MarkBind Version:** 1.17.0

Repeat the steps on 2 different platforms (Windows vs Mac/Linux):
1. `markbind init`
2. `markbind build`
3. Check the file paths inside the generated `index.html`.

The file paths generated should use a consistent slash (i.e. `/`), regardless of the OS that the author is running.

* On Windows, the slash generated is `\`.

```html
<script src="markbind\js\setup.js"></script>
<script src="markbind\layouts\default\scripts.js"></script>
</html>
```

* On Mac & Linux, the slash generated is `/`.

```html
<script src="markbind/js/setup.js"></script>
<script src="markbind/layouts/default/scripts.js"></script>
</html>
```

Xenonym:

Resolves #628.

When MarkBind generates paths during site build, it uses the default path separator of the OS it is running on (`\` on Windows, `/` on Mac/Linux). For consistency, MarkBind should use the same path seperator regardless of OS.

I created a temporary `assets` object which has all the paths run through `ensurePosix` to build the template with all forward slashes its paths. I also modified the parser to generate paths with forward slashes for dynamic `panel` includes.

**Testing instructions:**
1. `markbind build` a site. The generated HTML should use forward slashes regardless of OS.


src/Page.js
```diff
@@ -214,8 +214,18 @@ Page.prototype.prepareTemplateData = function () {
...
     : this.title;
 
+  // construct temporary asset object with only POSIX-style paths
+  const asset = {};
+  Object.entries(this.asset).forEach(([key, value]) => {
+    if (key === 'externalScripts') {
+      asset[key] = value;
+    } else {
+      asset[key] = FsUtil.ensurePosix(value);
+    }
```
nicholaschuayunzhi: We could shorten this to:
```js
asset[key] = key === 'externalScripts' ? value : FsUtil.ensurePosix(value);
```

src/Page.js
```diff
@@ -214,8 +214,18 @@ Page.prototype.prepareTemplateData = function () {
     ? this.titlePrefix + (this.title ? TITLE_PREFIX_SEPARATOR + this.title : '')
     : this.title;
 
+  // construct temporary asset object with only POSIX-style paths
+  const asset = {};
+  Object.entries(this.asset).forEach(([key, value]) => {
+    if (key === 'externalScripts') {
```
nicholaschuayunzhi: Since `ensurePosix` requires a `string`, perhaps we could use `typeof`

Xenonym:
> Have you considered writing a wrapper over `path`. While i understand it may hurt build time but it could potentially result in less developer error.

While it would be possible to have our own `path` that always uses forward slashes, I am not sure if it will create interoperability issues with other libaries, since they will most likely be relying on the default `path`with the platform specific `path.sep`. Passing paths between other libraires and our own code in this case would probably lead to unexpected results.

src/Page.js
Xenonym: As recommended by [`eslint-plugin-lodash/prefer-lodash-typecheck`](https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/prefer-lodash-typecheck.md), I have opted for `_.isString(value)`instead.

nicholaschuayunzhi:
@Xenonym Thanks for the changes, please rebase on `master`, thanks!

yamgent:
As discussed on Slack the `enablePosix()` thing is duplicated across two different util scripts. Perhaps the one in `src/lib/markbind/src/utils.js` can stay, while methods that are calling the `src/util/fsUtil.js` version can switch to `src/lib/markbind/src/utils.js`?

Xenonym:
@yamgent consolidated `ensurePosix` uses to the one from `src/lib/markbind/src/utils.js`!

PR #652, opened 2019-01-31
Update site.json and frontmatter tag behavior

jamos-tay:

Follow up from #467 

The current tagging overwrite behavior is too aggressive. We want to support the following use case:
 
> `helloWorld-fragment.mbdf`:
> 
> ```
> <div tags="level--beginner">
> 
> This is how you do Hello World:
> 
> <code tags="lang-cs">Console.WriteLine("Hello World");</code>
> <code tags="lang-java">System.out.println("Hello World");</code>
> 
> </div>
> 
> <div tags="level--advanced">
> 
> Let's do a refresher. Recall that to print anything, you would use <code tags="lang-cs">Console.WriteLine()</code><code tags="lang-cs">println()</code>.
> 
> <question>
> How would you print the string "Hello World"?
>   ...
> </question>
> 
> </div>
> ```
> 
> `index-cs.mbd`:
> 
> ```
> <frontmatter>
>     tags: lang-cs
> </frontmatter>
> 
> <include src="helloWorld-fragment.mbdf" />
> ```
> 
> `index-java.mbd`:
> 
> ```
> <frontmatter>
>     tags: lang-java
> </frontmatter>

Introduced group tags. Group tags are represented as `group--tagName`. No changes to current tag syntax, e.g. `tags: ['lang--java', 'lang--c++', 'level--advanced']`. (I think this is more convenient than using an object `{ lang : java }`, also it is consistent with usage in html files)

Instead of overriding, site and frontmatter tags are merged, but only one tag of a single group will be applied. For example:

`site.json` : `['a', 'b', 'lang--java']`
`frontmatter` : `['c', 'd', 'lang--cs', 'level--beginner']`
Merged: `['a', 'b', 'c', 'd', 'lang--java', 'level--beginner']`

Only the first `--` matters, tags like `['group--Name--name']` are possible
Tags without `--` are treated as a group tag with no name, they can be used on their own. We don't have to worry about overriding behavior for these.

Users can turn off tags completely by specifying a non-existent tag name:

`site.json` : `['lang--off', 'x--off']`
`frontmatter` : `['lang--java', 'x']`
`lang--java` and `x` tags will no longer show.

Does the above behavior work? I'm open to revising it. 

I'll update the documentation if it's okay.


yamgent:
> `site.json` : `['lang--off', 'x--off']`
> `frontmatter` : `['lang--java', 'x']`
> `lang--java` and `x` tags will no longer show.

Seems reasonable for now, although that means that if authors really use `lang--off` as a tag name it would be mildly confusing.

Also, I see `x` as a tag with no group, rather than a group with no actual tag name. Another potential source of confusion?

damithc:
> Instead of overriding, site and frontmatter tags are merged, but only one tag of a single group will be applied.

Will there be a need to allow more than one tag? e.g., `level--basic` and `level--intermediate` tags need to be allowed to create a document for intermediate level readers (they need to see basic level stuff too).

---

Another approach is to have a mechanism like the following (assuming we allow tags to be specified at the point of inclusion too): `*` means all, `+` means show, `-` means hide; tags applied left to right, cumulatively.
* file has content tagged with `level.1`, `level.2`, `level.3`
* frontmatter has the tags: `-level.*, +level.1` → only level 1 stuff are shown by default
* at the point of inclusion we specify `+level.2` → when applied on top of frontmatter tags, levels 1-2  stuff are shown
* site.json: `-level.*, +level.3` → only level 3 stuff are shown now

yamgent:
> Another approach is to have a mechanism like the following (assuming we allow tags to be specified at the point of inclusion too): `*` means all, `+` means show, `-` means hide; tags applied left to right, cumulatively.
> 
> * file has content tagged with `level.1`, `level.2`, `level.3`
> * frontmatter has the tags: `-level.*, +level.1` → only level 1 stuff are shown by default
> * at the point of inclusion we specify `+level.2` → when applied on top of frontmatter tags, levels 1-2  stuff are shown
> * site.json: `-level.*, +level.3` → only level 3 stuff are shown now

I think this approach is the best, it allows the author to have ultimate control over how the overriding behaviour works. Let's do that.

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -403,6 +403,65 @@ Alternatively, you can specify tags to render for a page in the front matter.
 ```
 </div>
 
-Tags in `site.json` will take precedence over the ones in the front matter.
+Tags in `site.json` will be merged with the ones in the front matter.
```
yamgent: Due to how they are merged, and with the additional wildcard feature, site.json will still "take precedence". This aspect needs to be better explained?

test/functional/test_site/index.md
```diff
@@ -64,7 +65,52 @@ head: myCustomHead.md, myCustomHead2.md
 # Heading with hidden keyword
 <span class="keyword d-none">invisible keyword</span>
 
-<include src="testTagDivs.md" />
```
yamgent: The tags in the test looks ok. But is there any reason we moved the tags to here? The `index.md` is getting really crowded. 😅 Putting it separate would make it neater.

src/Page.js
```diff
@@ -442,14 +443,26 @@ Page.prototype.filterTags = function (tags, content) {
   if (tags === undefined) {
     return content;
   }
-  const tagsArray = Array.isArray(tags) ? tags : [tags];
   const $ = cheerio.load(content, { xmlMode: false });
+  const tagOperations = this.frontMatter.tags.map(tag => ({
+    // Trim leading + or -, replace * with .*, add ^ and $
+    tagExp: `^${_.escapeRegExp(tag.replace(/^(\+|-)/g, '')).replace(/\\\*/, '.*')}$`,
```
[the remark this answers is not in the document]
jamos-tay: The original `*` gets replaced by `\*` when it is escaped, so we have swap it back out, e.g.
`nus.edu*` -> `nus\.edu\*` -> `nus\.edu.*`

test/functional/test_site/index.md
jamos-tay: Sure, moved it to separate file. Also marked the divs that should be hidden so it's easier to verify

jamos-tay:
Updated, but having the same issue as #474 ... will ask on slack

jamos-tay:
Rebased

yamgent:
Need you to resolve conflict again. :P I will merge this first next time.

jamos-tay:
Rebased

PR #664, opened 2019-02-03
Add temporary styles to prevent FOUC

### LINKED ISSUE #657: Minimize FOUC problem
- [ ] No FOUC in showcase websites
- [ ] No FOUC in TopNav

marvinchin:

Resolves #657.

`navbar` component rendering only after Vue has loaded causes FOUC. To avoid the page looking broken while Vue loads, we add temporary styles.

The `dropdown` component, similarly, is always expanded until Vue is loaded. This breaks the layout of the page.

Add temporary styles to `navbar` and `dropdown` during loading.

tempstyles

- I've removed the spinner and overlay after implementing the temporary styles as they both seem to address the same issue, but the temporary styles seems to address it more simply.
- Not sure if there are any other "critical" VueStrap components that require temporary styles in order to fully consider #657 resolved.

**Testing instructions:**
Open the Netlify preview page in Chrome with network speed set to `Fast 3G` or `Slow 3G`. The page should not look broken while it loads.
Another sample (with `dropdown` component) is available at https://marvinchin.github.io/website-base-loader/

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -212,3 +212,7 @@ footer {
 .temp-navbar li {
     display: block;
 }
+
+.temp-dropdown {
+    display: none;
+}
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -202,3 +202,13 @@ footer {
     0% { transform: rotate(0deg); }
     100% { transform: rotate(360deg); }
 }
+
+/* Temporary Styles */
+
+.temp-navbar {
+    display: flex;
+}
+
+.temp-navbar li {
+    display: block;
+}
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -159,50 +159,6 @@ footer {
     border-radius: 20px;
 }
 
-/* Spinner */
-
-#loading-overlay {
-    background: white;
-    bottom: 0;
-    display: none;
-    left: 0;
-    position: fixed;
-    right: 0;
-    top: 0;
-    z-index: 100;
-}
-
-#spinner-layout {
-    align-items: center;
-    animation: show 1s;
-    animation-fill-mode: forwards;
-    animation-timing-function: step-end;
-    display: flex;
-    height: 100%;
-    justify-content: center;
-    visibility: hidden;
-    width: 100%;
-}
-
-@keyframes show {
-    0% { visibility: hidden; }
-    100% { visibility: visible; }
-}
-
-#spinner {
-    animation: spin 1s linear infinite;
-    border: 16px solid lightgray;
-    border-radius: 50%;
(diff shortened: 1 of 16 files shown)


damithc:
What if we do our own TopNav, based on BootStrap? I believe we are doing our own PageNav and SiteNav already?

damithc:
> What if we do our own TopNav, based on BootStrap? I believe we are doing our own PageNav and SiteNav already?

I meant as a more permanent solution. This PR can serve as a stop-gap solution for the time being.

marvinchin:
That might help avoid having the TopNav render be blocked by Vue, but I'm concerned that:
1. Users will not have as much freedom to customise the TopNav (e.g. using Dropdowns, which is a VueStrap component)
2. Even if we do this for the TopNav, other VueStrap components used in the user content could still cause the initial render to look distorted

yamgent:
> What if we do our own TopNav, based on BootStrap? I believe we are doing our own PageNav and SiteNav already?

That is certainty possible, we do not write PageNav and SiteNav as Vue components, but rather as something that gets rendered automatically during `markbind build`, which fits the definition of static site generation.

Going in this direction means less reliance on vue-strap, which means FOUC can be avoided.

However, from the code maintainability point of view, this is bad. The code written in Vue is definitely cleaner and easier to read, compared to code written using pure JavaScript.

damithc:
> That is certainty possible, we do not write PageNav and SiteNav as Vue components, but rather as something that gets rendered automatically during `markbind build`, which fits the definition of static site generation.

I'm OK with this approach. We can also do it in steps, i.e., using a dirty solution while we implement a more permanent solution.

> However, from the code maintainability point of view, this is bad. The code written in Vue is definitely cleaner and easier to read, compared to code written using pure JavaScript.

I guess we have to use it sparingly and only when necessary.

asset/css/markbind.css
```diff
@@ -202,3 +202,13 @@ footer {
     0% { transform: rotate(0deg); }
     100% { transform: rotate(360deg); }
 }
+
+/* Temporary Styles */
```
yamgent: We need to explain the rationale for having and using temporary styles (which is to avoid the FOUC problem).
I think this should be the best place to put the entire explanation in.

yamgent: Then perhaps we can also name all references as "Temporary **Anti-FOUC** Styles" rather than just "Temporary Styles"? This may make it even more clearer to the developers how these temporary styles are used.

marvinchin: Fixed!

damithc:
This seems to have improved the FOUC situation a bit e.g., https://nus-te3201.github.io/2019/

Issue #448, opened 2018-10-05
Add a Windows-specific CI such as AppVeyor


yamgent:
We currently utilize Travis as our sole CI, which does testing on Linux for our PRs.

We should consider adding AppVeyor as a second CI so that we can run the tests on Windows too, to be assured that our code runs on Windows without any issues as well (since it is possible that the same code might behave differently on different OS).

nicholaschuayunzhi:
> (since it is possible that the same code might behave differently on different OS).

One of the differences would be the forward and backward slashes in generated in paths. The testing should still be fine on that part though because we ignore such differences during our tests.

yash-chowdhary:
Is it possible for a non-member of the MarkBind organization to enable AppVeyor on the repository? AppVeyor needs Organization Access (which only members can grant) to be able to build on each commit/PR.

damithc:
> Is it possible for a non-member of the MarkBind organization to enable AppVeyor on the repository? AppVeyor needs Organization Access (which only members can grant) to be able to build on each commit/PR.

You can create a dummy org to try the set up. Once the set up is working, PR here to add the required files and also specify the steps one of the org admins can follow to set it up for this repo.

PR #677, opened 2019-02-07
Convert relative src references to absolute paths

### LINKED ISSUE #598: Src tag: unintuitive as resources are relative to included folder

* **MarkBind Version:** v1.16.1

Folder structure:
```
root/
- index.md
- a/
- - img.png
- - subsite.md
```
_index.md_
```html
...
<include src="a/subsite.md" />
```
_subsite.md_
```html
...
<img src="img.png" />
```

The subsite.md/html works, because the references are correct. 
I would also expect any site reusing the subsite to work too. In this case, index.md should work and display img.png properly.

However, the image link is broken when viewing index.md/html. Upon inspection, the img src in index.html points to `img.png`, suggesting that the content of subsite.md is just copied into index.md.
To get the image to work on both sites, the code has to be changed to ``. However, this is tedious and unintuitive for authors to write.

**Possible solution**
In addition to the pre-process step of include, additionally pre-process the src tags too, replacing all relative URLs with their absolute path equivalent. For example, we could use `{{baseUrl}}/a/img.png` in the example above.
- However, we need to be able to detect relative URLs vs absolute one, as well as handle additional edge cases like "http(s)://..." or "#".

**Extension**
This issue is also applicable to other media elements such as videos, downloadable files, css and hrefs.

### LINKED ISSUE #661: Ensure syntax is reasonably author-friendly
- [ ] relative URLs should not need {{ baseUrl }} #9

sijie123:

• [x] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Fixes #598 
Fixes #661

Currently, authors can use dynamic references in include tags.
However, if authors use dynamic resource references in src tags for pic/img etc, these references may cause broken links when re-used. This request converts all dynamic resource references to static references, so it will work consistently across all sites and sub-sites. At the same time, this request should be backward compatible, so that authors using older workarounds should expect their site to continue working.

During the preprocess stage, in addition to preprocessing includes, also preprocess media resources (e.g. pic) to convert all dynamic references to static ones with respect to the root path (where site.json is located)
Additionally, removed unnecessary use of {{baseUrl}} from the user guide, changing all such links to static references. However, I did not touch mbdf files within `userGuide/syntax` and `userGuide/plugins`, as relative references might make it confusing to write these files in the future.

Refer to #598 
Folder structure:
```
root/
- index.md
- a/
- - img.png
- - subsite.md
```
_index.md_
```html
...
<include src="a/subsite.md" />
```
_subsite.md_
```html
...
<img src="img.png" />
```
**New syntax changes**
Authors can now also use relative src references in `img`, `pic` tags, as well as relative href attributes in `a` and `link` tags.
To use
* relative src attributes: All references without a leading '/' or {{...}} are treated as relative references.
* static src attributes: All references with a leading '/' or starts with `{{variables}}` are treated as absolute paths, and will not be modified.

**Breaking changes**
There should be no breaking changes. I tested against the old unit tests (I didn't modify anything) and code is passing. I have also added unit tests to ensure that future changes will retain this behaviour.
My PR might break existing documents if authors used incorrect URI statements like `/{{baseUrl}}/path/to/file`. Notice that this string, in the original version, essentially resolves to `//rootFolder/path/to/file` with 2 forward slashes. While previous versions might be lenient, my conversion requires stricter syntax and might result in broken links.

Are there scenarios where this change would cause something to break? I've tested on unit tests and CS2103-website.

**Testing instructions:**
Replicate the folder structure as above. Be sure to include dynamic references to media resources. Run markbind serve, then view the generated webpage.

### DIFF
--- a/docs/userGuide/syntax/links.mbdf
+++ b/docs/userGuide/syntax/links.mbdf
@@ -45,15 +45,47 @@ MarkBind home is at [here][1]
 
 <div id="intraSiteLinks">
 
-Links to files of the generated site (e.g., an HTML page or an image file) should be specified as absolute paths and should start with {{ showBaseUrlCode }} (which represents the root directory of the project).
+Links to files of the generated site (e.g., an HTML page or an image file) can be specified either as relative paths or absolute paths.
 
+Absolute paths:
 <div class="indented">
+Links should start with {{ showBaseUrlCode }} (which represents the root directory of the project). 
 
 {{ icon_example }} Here's how to specify a link to (1) a page, and (2) an image, using the {{ showBaseUrlCode }}:
 
-1. <code>Click [here]({<span></span>{ baseUrl }}/userGuide/reusingContents.html).</code>
+1. <code>Click [here]({{ showBaseUrlCode }}/userGuide/reusingContents.html).</code>
 2. `![](`{{ showBaseUrlCode }}`/images/preview.png)`
 
+</div>
+
+Relative paths:
+
+<div class="indented">
+Links to files can also be specified relative to the file that includes it.
+
+{{ icon_example }} Assuming that we have the following folder structure:
+```
+c:/course/
+  ├── textbook/
+  |      ├── subsite.md
+  |      ├── image.png
+  |      └── site.json
+  ├── index.md
+  └── site.json
+```
+
+Within `textbook/subsite.md`, we can refer to the image using:
+```html
+<img src="image.png" />
+```
+Within `index.md`, we can also display the image using
+```html
+<img src="textbook/image.png" />
+```
+or by including `subsite.md`:
+```
+<include src="textbook/subsite.md" />
+```
 </div>
 </div>
 
--- a/test/functional/test_site/expected/sub_site/testReuse._include_.html
+++ b/test/functional/test_site/expected/sub_site/testReuse._include_.html
@@ -1,0 +1,18 @@
+<p>This is a page from another Markbind site. The purpose of this page is to ensure that reuse works as expected. All the following images should display correctly.</p>
+<p>Some variables:</p>
+<div></div>
+<p>IMG tags:
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="https://dummyimage.com/600x400/000/fff"></p>
+<p>PIC tags:
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+  <pic src="https://dummyimage.com/600x400/000/fff"></pic>
+</p>
+<p>Anchor link:
+  <a href="https://dummyimage.com/600x400/000/fff">External Image</a></p>
--- a/test/functional/test_site/sub_site/testReuse.md
+++ b/test/functional/test_site/sub_site/testReuse.md
@@ -1,0 +1,23 @@
+This is a page from another Markbind site.
+The purpose of this page is to ensure that reuse works as expected.
+All the following images should display correctly.
+
+Some variables:
+<variable name="imgFolder">images</variable>
+
+IMG tags:
+<img src="images/I'm not allowed to use my favorite tool.png">
+<img src="{{imgFolder}}/I'm not allowed to use my favorite tool.png">
+<img src="{{baseUrl}}/{{imgFolder}}/I'm not allowed to use my favorite tool.png">
+<img src="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png">
+<img src="https://dummyimage.com/600x400/000/fff">
+
+PIC tags:
+<pic src="images/I'm not allowed to use my favorite tool.png"></pic>
+<pic src="{{imgFolder}}/I'm not allowed to use my favorite tool.png"></pic>
+<pic src="{{baseUrl}}/{{imgFolder}}/I'm not allowed to use my favorite tool.png"></pic>
+<pic src="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png"></pic>
+<pic src="https://dummyimage.com/600x400/000/fff"></pic>
+
+Anchor link:
+<a href="https://dummyimage.com/600x400/000/fff">External Image</a>
--- a/docs/_markbind/variables.md
+++ b/docs/_markbind/variables.md
@@ -16,7 +16,7 @@
 <variable name="icon_info">:fas-info-circle:</variable>
 <variable name="icon_ticked">:far-check-square:</variable>
 
-<variable name="link_live_preview">[live preview]({{ baseUrl }}/userGuide/glossary.html#live-preview)</variable>
+<variable name="link_live_preview">[live preview](userGuide/glossary.html#live-preview)</variable>
 
 <variable name="tooltip_root_directory"><tooltip content="The directory that contains all the project files. It is also the directory in which the `site.json` configuration file is located.">root directory</tooltip></variable>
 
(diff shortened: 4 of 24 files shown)


sijie123:
> I found an edge case you didn't cover. If you put a `<pic>` component within `<panel>` it will not change the `src`.
> 
> In `index.md`	In `images/subsite.md`
> evidence-1`<include src="images/subsite.md"></include>`	evidence-2 `<panel header="### test-pic">`
>   `<pic src="om.jpg" alt="om-pokemon"></pic>`
> `</panel>`
> **EDIT:**
> Checked out `PR 667` instead of `PR 677` for testing.
> 
> Tested on `PR 677` and the above example works.

Thanks for testing though - that was a case I didn't think of, but somehow it works :P
I'll be sure to add it to my unit tests.

damithc:
How's this going?
I suppose the old syntax will continue to work even after this is merged?
Don't forget to update the docs too.

sijie123:
I've been working on SE-EDU over the past couple of weeks, so I haven't done much here.
I'll be fixing this issue for other media tags over the next week or so, and I'll write some tests for it.

The old syntax will continue to work, and yes, thanks for the reminder. Will update the docs too.

sijie123:
Just pushed an update to this PR.
Now fixes relative URLs in src attributes in `img` and `pic`. Additionally, also supports `href` attrib in `a` tag.

Also works for URLs involving variables: {{baseUrl}}/path/to/file. However, this involved a hack: checking whether the string starts with "{{". I'm not sure if this is a good idea.

I have updated the PR issue with some things I need help with, from reviewers.

sijie123:
@Chng-Zhi-Xuan something's broken right now, please don't review it yet! :)

sijie123:
I've updated the code and added unit tests.
Tested with:
* Unit and Functional tests
* CS2103T Website

sijie123:
> How's this going?
> I suppose the old syntax will continue to work even after this is merged?
> Don't forget to update the docs too.

Prof @damithc I didn't update the [docs](https://markbind.org/userGuide/reusingContents.html#reusing-contents-across-sites) in the end.
If anything, the current docs used to be wrong and the correct behaviour is "fixed" by this PR.
The current docs actually doesn't say anything about references breaking if we reuse images and other media content through including sub-sites.
This PR "fixes the bug" and allows the user to include images in sub-sites in any way they prefer. It is more intuitive: static and dynamic references both work in a way that truly "allows you to include the pages you want ... without having to change anything in the source files"

damithc:
Perhaps also remove all unnecessary `{{ baseUrl }}`s in the docs site?

sijie123:
Okay :))
Done with removing unnecessary {{baseUrl}}s from the docs too.
Edited the PR message to reflect this.

docs/userGuide/syntax/links.mbdf
```diff
@@ -45,15 +45,44 @@ MarkBind home is at [here][1]
...
+Links to files can also be specified relative to the file that includes it.
+
+{{ icon_example }} Assuming that we have the following folder structure:
+```
+/
+- index.md
+- sub/
+- - img.png
+- - subsite.md
+```
```
damithc: Make it look like this?
image

sijie123: That's a nice way to represent the file structure. Will do that. Thanks!

test/functional/test_site/sub_site/testReuse.md
```diff
@@ -0,0 +1,16 @@
+This is a page from another Markbind site.
+The purpose of this page is to ensure that reuse works as expected.
+All the following images should display correctly.
+
+IMG tags:
+<img src="images/I'm not allowed to use my favorite tool.png">
+<img src="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png">
+<img src="https://dummyimage.com/600x400/000/fff">
+
```
yamgent: Do you think we should also test for using "non-baseUrl" variables as well? E.g.

```
<variable name="somePath">images</variable>
<img src="{{somePath}}/images/I'm not allowed to use my favorite tool.png">
```

What do you think?

yamgent:
Btw good job in attempting to organize your commit. I will retain your commits when merging your PR, so don't worry about your efforts going to waste. :+1:

In fact, perhaps this project could start moving towards such commit organizations for future PRs.

damithc:
> Btw good job in attempting to organize your commit. I will retain your commits when merging your PR, so don't worry about your efforts going to waste.

We can enable CanIHasReview in case someone wants to do the PR that way. Just need to give write permission to CIHR.

yamgent:
@sijie123 need you to resolve conflict for this PR; all PRs were affected by the recent master branch rebase.

sijie123:
Done. Thanks for the heads-up!

test/functional/test_site/expected/sub_site/testReuse._include_.html
```diff
@@ -0,0 +1,18 @@
...
+<div></div>
+<p>IMG tags:
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="https://dummyimage.com/600x400/000/fff"></p>
+<p>PIC tags:
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
```
Chng-Zhi-Xuan: When running `markbind serve` on the `test_site` folder and replacing the `_site` contents into `expected` folder

I encountered this:
path diff

It works fine when viewing in the browser though.

yamgent: Nice catch! The paths need to be further processed with `ensurePosix()`, so that on Windows, the path separator will consistently be `/`.

Could have caught this via CI if #448 is implemented. :P

yamgent:
Updated original post to include "fixes #661".

test/functional/test_site/expected/sub_site/testReuse._include_.html
sijie123: Nice catch! Thanks @Chng-Zhi-Xuan .

damithc:
@sijie123 this doesn't seem to work when the file containing the relative link is included somewhere else. Is it supposed to?

damithc:
> @sijie123 this doesn't seem to work when the file containing the relative link is included somewhere else. Is it supposed to?

In fact, it seems to work for `` but not for ``

sijie123:
> @sijie123 this doesn't seem to work when the file containing the relative link is included somewhere else. Is it supposed to?

If I understand your question correctly, are you trying to include files across different sites? (ie site_one trying to include site_two/file.md, where site_two is not a sub-site or sub-folder of site-one)
Right now, we don't have cross-site support, so the included file must be a sub-site or within a sub-folder of the originating site for relative URLs to work. Otherwise, the author will still have to use absolute paths.


> > @sijie123 this doesn't seem to work when the file containing the relative link is included somewhere else. Is it supposed to?
> 
> In fact, it seems to work for `` but not for ``

This however, seems to be an issue. I implemented the code with the assumption that markdown code will first be converted to html before we do any parsing. It seems that this is not the case? (@yamgent / @Xenonym: do you guys know how our parser layers work?)

damithc:
> If I understand your question correctly, are you trying to include files across different sites? (ie site_one trying to include site_two/file.md, where site_two is not a sub-site or sub-folder of site-one)
> Right now, we don't have cross-site support, so the included file must be a sub-site or within a sub-folder of the originating site for relative URLs to work. Otherwise, the author will still have to use absolute paths.

It's within the same site, not even a subsite, but the problem is limited to only the `` syntax.

yamgent:
> (@yamgent / @Xenonym: do you guys know how our parser layers work?)

`Parser._preprocess()` is called before any markdown-to-html conversion takes place. It is directly called by `Parser.includeFile()`, which is in turn called here:

https://github.com/MarkBind/markbind/blob/40f3e7a31cd73e5b4ba4978b3160d8cb0ed66e56/src/Page.js#L776-L796

`markbinder.renderFile()` is the one that is converting markdown to HTML.

> In fact, it seems to work for `` but not for ``

I completely forgotten about the markdown syntax equivalents, we are doing the relative src referencing in the wrong pass.

sijie123:
😱 I think I'll have to discuss this issue with you on Thursday...
Right now I think maybe we might need to add an extra layer after markdown converts everything?
I was hoping to be able to not introduce an additional layer, but I think we don't have much choice ._.

yamgent:
Yes looks like we will have to introduce an additional `.then(result => ...)` (I presume that is what you meant).

Then again, if a developer fears adding a new layer, that is a sign that that part of the code is too intertwined, and needs to be refactored into more logical units. 😛

sijie123:
For future reference, the follow-up issue is resolved by #826.

PR #680, opened 2019-02-08
Use a simpler :prefix-name: syntax for icon fonts

### LINKED ISSUE #612: Use a simpler syntax for glyphicons/fontawesome
Current syntax: `{{ fas_home }}`
Suggested: `:fas-home:`

Inspired by https://www.npmjs.com/package/markdown-it-fontawesome
Not sure if it is possible though.

Xenonym:

Resolves #612.

Currently, we use the `{{ prefix_name }}` syntax for inserting icon fonts from Font Awesome and Glyphicons.  It would be ideal to support a simpler `:prefix-name:` syntax, similar to how we support `:emoji:`.

- I added [`markdown-it-regexp`](https://github.com/rlidwka/markdown-it-regexp) to match `:prefix-name:` and replace them with the appropriate icon font HTML.
- I added `aria-hidden="true"` to the FA icons to be consistent with Glyphicons and for accessibility. 
([FA's accessibility documentation](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility#web-fonts-decorative), [Bulletproof Accessible Icon Fonts](https://www.filamentgroup.com/lab/bulletproof_icon_fonts.html#accessibility))
- I updated the default `site-nav.md`, the `test_site`'s `site-nav.md`, and the user guide to use the new `:prefix-name:` syntax.
- I removed the use of icon font variables from the default site, unit tests and `test_site`.
- I created a corresponding PR MarkBind/vue-strap#97 in the `vue-strap` repo to parse `:prefix-name:` in `<panel>` headers.

```markdown
- FA university icon: :fas-university:
- Glyphicon icon: :glyphicon-home:
```

**Testing instructions:**
1. Add the above sample code to a MarkBind page and `markbind serve`. The icons should render:
Example render of :prefix-name: icon fonts


docs/index.md
```diff
@@ -94,7 +94,7 @@ In the example below, there is an expandable panel that can reveal more content.
 
 A binary tree is a tree data structure in which each node has at most two children, which are referred to as the left child and the right child. Primitive data types on the other hand ...
 
-<panel header="%%{{ icon_info }} Some example primitive data types%%">
+<panel header="%%:fas-info-circle: Some example primitive data types%%">
```
damithc: Change the definition of the variable in variables.md instead?

damithc: Applies to other `icon_*` variables too.

Xenonym: Currently, the `icon_*` variables have icon fonts of the form `{{ icon_font | safe }}`. I assumed it was a workaround for including them in other variables, in which case the new syntax does not need it anymore.

Did you actually mean for them to be semantic aliases instead? If so, I will add them back in.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -26,6 +26,19 @@ markdownIt.use(require('markdown-it-mark'))
...
+      let iconFontType = match[1];
+      let iconFontName = match[2];
+
+      if (iconFontType === 'glyphicon') {
+        return `<span aria-hidden="true" class="glyphicon glyphicon-${iconFontName}"></span>`;
+      } else {
+        return `<span aria-hidden="true" class="${iconFontType} fa-${iconFontName}"></span>`;
+      }
+    }
+  ))
```
acjh: Let's abstract this into `./markdown-it-icons`.

docs/index.md
damithc: > Did you actually mean for them to be semantic aliases instead? If so, I will add them back in.

Yes, that was my intention. Easier to remember.

Xenonym: @damithc I have added the variables back!

src/lib/markbind/src/lib/markdown-it/index.js
Xenonym: @acjh done!

src/lib/markbind/src/lib/markdown-it/markdown-it-icons.js
```diff
@@ -0,0 +1,13 @@
+module.exports = require('markdown-it-regexp')(
+    /:(fa[brs]|glyphicon)-([a-z-]+):/m,
+    (match, utils) => {
+        let iconFontType = match[1];
+        let iconFontName = match[2];
+
+        if (iconFontType === 'glyphicon') {
+            return `<span aria-hidden="true" class="glyphicon glyphicon-${iconFontName}"></span>`;
+        } else {
```
yamgent: Add a comment that this is for "Font Awesome" icons? Or perhaps check that the `iconFontType` is "fas"/"far"/"fab" before rendering it as a Font Awesome icon?

Chng-Zhi-Xuan:
Seems like there are some icons that are not rendered.

In `reusingcontents.md` |
--- |
unrendered-icons |

src/lib/markbind/src/lib/markdown-it/markdown-it-icons.js
Chng-Zhi-Xuan: To add on, maybe use a switch statement? Since checking `fas`/`far`/`fab` within `if else` is quite long.

Xenonym: The regex `:(fa[brs]|glyphicon)-([a-z-]+):` should only catch `:fab-`/`:far-`/`:fas-`/`:glyphicon-` anyway, so a check would be redundant. Will add in a comment instead!

Xenonym:
> Seems like there are some icons that are not rendered.

@Chng-Zhi-Xuan good catch! Turns out a preceding line break from the last `<div>` tag is needed, since we need to interpret the text as Markdown for `markdown-it` to parse the icon fonts.

Xenonym:
> A search for `{{glyphicon` still returns files that are using the old syntax.

Yes, since these are still used to test variables. Since we are only deprecating the syntax for the next release, I think we still have to keep them to ensure that existing uses don't break until when we remove the syntax altogether. Also, there are some issues related to testing since we will not have static built-in variables after we remove the `{{ prefix_name }}` syntax. From the PR:

>Some `{{ prefix_name }}` usages that were not changed because they are implicitly used to test variables:
>* In `Site.js#FOOTER_DEFAULT`:
  `'This is a dynamic height footer that supports variables {{glyphicon_tags}}'`
>* In `Site.test.js#test('Site resolves variables referencing other variables')`:
  `'<span id="level3">{{glyphicon_plus}}</span>'`

> When we remove icon font variables, the only remaining built-in variable with HTML will be `{{ MarkBind }}`. Since `{{ MarkBind }}` changes with every new version, I am unsure if it is a good candidate for automated testing.

docs/index.md
```diff
@@ -103,7 +103,7 @@ In the example below, there are expandable panels that are nested within each ot
 
 <box>
 
-<panel header="{{ far_list_alt }} Exercises" no-close >
+<panel header=":far-list-alt: Exercises" no-close >
```
Chng-Zhi-Xuan: Another un-rendered icon

unrendered-icon

Chng-Zhi-Xuan:
> When we remove icon font variables, the only remaining built-in variable with HTML will be `{{ MarkBind }}`. Since `{{ MarkBind }}` changes with every new version, I am unsure if it is a good candidate for automated testing.

I think it is fine to swap over to the new syntax completely, We can just add unit tests in `test_site` to test any old syntax required.

acjh:
@Chng-Zhi-Xuan By completely, do you mean including `{{ MarkBind }}`?

Note that this PR does not (should not) introduce a new syntax for variables, just "constant" icons.

Chng-Zhi-Xuan:
@acjh I am referring to the Glyphicons using the old syntax. Since this PR is meant to improve icon syntax only.

docs/index.md
Xenonym: @Chng-Zhi-Xuan This should be rendered once you include the `vue-strap.min.js` built from MarkBind/vue-strap#97 in `assets`.

Chng-Zhi-Xuan:
@Xenonym, I looked through again and seems ok.

However, we should change our documentation and tests to follow the new icon syntax in this PR. "Implicit testing" of old syntax by intentionally leaving some *icons with old syntax* in templates or tests is not clean.

Please do change the `FOOTER_DEFAULT` template in `Site.js` which uses icons, to the new syntax. I found other instances in `Site.test.js` and `test_site` using the old icon syntax as well. 

Once you're done, I can give the green light 👍

Xenonym:
> Please do change the `FOOTER_DEFAULT` template in `Site.js` which uses icons, to the new syntax. I found other instances in `Site.test.js` and `test_site` using the old icon syntax as well.

@Chng-Zhi-Xuan made the necessary changes! Here's a summary:
- Removed `{{glyphicons_tag}}` from `FOOTER_DEFAULT`. The use of `{{MarkBind}}` and `{{timestamp}}` already serve as a means to demonstrate variables.
- Removed `{{education_icon}}` from `test_site`.
- Modified `test('Site resolves variables referencing other variables')` to use a custom variable instead.
- Used a `<span>` instead of `<font>` tag for the code example in `reusingContents.md` since the [`<font>` tag is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font#Summary).

damithc:
Gave it a try. Works as intended. Nice work @Xenonym and reviewers.

One small nit. This extra space can be misleading (not sure if it was introduced in this PR or not).

image

Chng-Zhi-Xuan:
That extra space was already present before the icon syntax change PR. The blank space is caused by the `<code>` tag having padding on the left and right side. There are 2 inline code blocks (`fab-` and `file-code` etc) so the result looks like an additional space in-between.

Snippet from `icons.mbdf`
```HTML
   * _Solid_ (prefix: `fas-`) e.g., :fas-file-code: (actual name `file-code`, MarkBind name **`fas-`**`file-code`)
   * _Regular_ (prefix: `far-`) e.g., :fas-file-code: (actual name `file-code`, MarkBind name **`far-`**`file-code`)
   * _Brands_ (prefix: `fab-`): e.g., :fab-github-alt: (actual name `github-alt`, MarkBind name **`fab-`**`github-alt`)
```

damithc:
> That extra space was already present before the icon syntax change PR. The blank space is caused by the `<code>` tag having padding on the left and right side. There are 2 inline code blocks (`fab-` and `file-code` etc) so the result looks like an additional space in-between.

I guess it is a legacy problem. It's better to fix it though. We can sacrifice the bolding. On a related note, I hope we can add a proper way to highlight/bold parts of `code` some day.

PR #681, opened 2019-02-08
Support specifying include variables inline

### LINKED ISSUE #642: Variables in includes: provide a simpler syntax
Current: syntax for specifying _include-variables_:
```html
<include src="foo.md">
  <span id="x">bat</span>
</include>
```

Suggested: also allow this syntax:
```html
<include src="foo.md" x="bat" />
```

Not sure how hard it is though.

jamos-tay:

Fixes #642 

Provide a simpler syntax for specifying include variables

Authors can include variables inline by adding `var-...` attributes:

```
<include src="foo.md" var-x="bat" />
==
<include src="foo.md">
  <span id="x">bat</span>
</include>
```

Used `var` instead of `data` as it's more variable related

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -94,6 +94,15 @@ function Parser(options) {
  */
 function extractIncludeVariables(includeElement, contextVariables) {
   const includedVariables = { ...contextVariables };
+  Object.keys(includeElement.attribs).forEach((attribute) => {
+    if (!attribute.startsWith('var-')) {
+      return;
+    }
+    const variableName = attribute.replace(/^var-/, '');
+    if (!includedVariables[variableName]) {
+      includedVariables[variableName] = includeElement.attribs[attribute];
+    }
+  });
   if (includeElement.children) {
     includeElement.children.forEach((child) => {
       if (child.name !== 'span') {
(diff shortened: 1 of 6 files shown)


src/lib/markbind/src/parser.js
```diff
@@ -94,6 +94,15 @@ function Parser(options) {
  */
 function extractIncludeVariables(includeElement, contextVariables) {
   const includedVariables = { ...contextVariables };
+  Object.keys(includeElement.attribs).forEach((attribute) => {
+    if (!attribute.startsWith('var-')) {
+      return;
+    }
+    const variableName = attribute.replace(/^var-/, '');
+    if (!includedVariables[variableName]) {
+      includedVariables[variableName] = includeElement.attribs[attribute];
```
yamgent: Is it possible to run through the variable through nunjucks?

jamos-tay: Hmm I think at this point the variable would have already been ran through nunjucks based on the preprocessing

jamos-tay: I tested it with `<include src="foo.md" var-x="bat {{ color }}" />` where color is a global var and it works correctly

jamos-tay:
Rebased, see comment

src/lib/markbind/src/parser.js
yamgent: You are right, I stand corrected.

damithc:
@jamos-tay there was a suggestion from @acjh to use `data-` as the prefix (apparently, that's what HTML standard recommends). Was it changed to `var-` later?

acjh:
Since the `<include>` tag is not in the rendered HTML, `var-` attributes are acceptable.

jamos-tay:
Yeah, I figured `var-` might be better since it's more related to variables. `data` might collide with some other libraries.

PR #691, opened 2019-02-13
[WIP] Integrate Agolia Search

### LINKED ISSUE #656: Minimal support 3rd party search services
- [x] Fix problems preventing Algolia search integration
- [x] Set up Algolia search for an example site

marvinchin:

• [X] New feature

Resolves #656.

Allow users to integrate Agolia search with their site

- Streamline search related configs
- Add support for Agolia search 

**Testing instructions:**
For the `se-book` repository:

Update `site.json`:
```js
{
    // ... some config
    "search": {
        "provider": "agolia",
        "config": {
            "apiKey": API_KEY_HERE,
            "indexName": INDEX_NAME
        }
    }
}
```

Update `navigation.md`:
```md
...
      <searchbar placeholder="Search" agolia menu-align-right></searchbar>
...
```

Agolia search results should be displayed when user enters query in search bar.

### DIFF
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -108,7 +108,7 @@ function setup() {
   setupPageNav();
 }
 
-function setupWithSearch() {
+function setupWithMarkbindSearch() {
   const { searchbar } = VueStrap.components;
   // eslint-disable-next-line no-unused-vars
   const vm = new Vue({
@@ -137,8 +137,8 @@ function setupWithSearch() {
   setupPageNav();
 }
 
-if (enableSearch) {
-  setupWithSearch();
+if (enableMarkbindSearch) {
+  setupWithMarkbindSearch();
 } else {
   setup();
 }
--- a/src/Page.js
+++ b/src/Page.js
@@ -65,9 +65,9 @@ function Page(pageConfig) {
   this.layout = pageConfig.layout;
   this.layoutsAssetPath = pageConfig.layoutsAssetPath;
   this.rootPath = pageConfig.rootPath;
-  this.enableSearch = pageConfig.enableSearch;
   this.plugins = pageConfig.plugins;
   this.pluginsContext = pageConfig.pluginsContext;
+  this.search = pageConfig.search;
   this.searchable = pageConfig.searchable;
   this.src = pageConfig.src;
   this.template = pageConfig.pageTemplate;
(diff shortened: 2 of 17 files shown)


acjh:
Is it possible to improve plugin support such that this can be implemented as a plugin?

It feels odd to integrate an optional third-party service to such an extent in MarkBind.

If necessary, then it can be done in a separate PR.

marvinchin:
That might be possible. I'm thinking of extending the plugin support to allow adding scripts and stylesheets in the `preRender` stage (through some helper methods e.g. `addExternalScript`, `addExternalStylesheet`, `addRawScript`).

Alternatively, we could just modify the DOM directly as part of the `postRender`, but I feel that this might be a little less clean.

What are your thoughts on this?

acjh:
> I'm thinking of extending the plugin support to allow adding scripts and stylesheets in the `preRender` stage (through some helper methods e.g. `addExternalScript`, `addExternalStylesheet`, `addRawScript`).

Those are independent of the content right? They can be functions at the same level as `preRender`.
We should try to handle this using page.ejs.

We can accept raw HTML to be more flexible, then we just need `addScript` and `addStylesheet`.

> Alternatively, we could just modify the DOM directly as part of the `postRender` (no additional support needed), but I feel that this might be a little less clean.

Agreed, don't do this.

marvinchin:
I'm concerned that having to write
```
addStylesheet("<link rel="stylesheet" href="STYLESHEET_LINK">");
```
would be clunky compared to something like
```
addExternalStylesheet("STYLESHEET_LINK");
```

There might be more methods available (which might be a little more confusing), but I wonder if it is worth the tradeoff for a more pleasant syntax.

acjh:
There is more to a `<link>` and a `<script>` than that:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

It is better to be flexible for plugin authors.
Examples of templating scripts in other programming languages (didn't find a good JS example):
- Python: http://dev.splunk.com/view/webframework-djangobindings/SP-CAAAESW
- C#: https://stackoverflow.com/questions/37727972/what-is-section-scripts-and-what-it-is-used-for

Plugin can also load HTML file of styles and/or scripts and pass it as string.

marvinchin:
Alright, will do that instead then! 🙂

PR #693, opened 2019-02-14
Make page variables no longer cascade into inner files

### LINKED ISSUE #674: Page variables should only be overridden by request
v1.17.2

Context:
`parent.md`:
```html
<variable name="title">Parent page</variable>

Refer the <include src="child.md#summary" />
```
`child.md`:
```html
<variable name="title">Child page</variable>

<span id="summary">
summary of page: {{ title }}
</span>
```

Current: the parent.md renders `Refer the summary of page: Parent page`
Expected: `Refer the summary of page: Child page`

Suggested: override page variables only when specifically requested, as follows:
```html
<variable name="title">Parent page</variable>

<include src="child.md#summary">
  <span id="title">{{ title }}</span>
</include>
```

Rationale: to avoid accidental overriding. We can expect variable names such as `title` to occur multiple times within an include chain. This problem tripped me up several times when doing #668

jamos-tay:

• [X] Bug fix

Fixes #674 

Page variables should not propagate so as to differentiate them from include variables.

Changes page variables so they no longer cascade into includes.

To cascade a variable, use the include variable syntax:

```
<variable name="x">Variable</variable>

<include src="foo.md>
  <span id="x">{{ x }}</span>
</include>
```


damithc:
Gave it a try. Looks like it works as intended. Nice work @jamos-tay

PR #694, opened 2019-02-14
Mark failed tests for tags and variables

jamos-tay:

• [X] Other, please explain: Testing

Some of the variable and tag tests, especially those testing overriding behavior are hard to verify correctness since it is difficult to tell at first glance if the output is incorrect

Added a `**TEST FAIL**` to incorrect outputs so they show up clearer in index.html.

Awaiting #693, #652 to finish

### DIFF
--- a/test/functional/test_site/index.md
+++ b/test/functional/test_site/index.md
@@ -42,7 +42,7 @@ tags: ["tag-frontmatter-shown", "tag-included-file", "+tag-exp*", "-tag-exp-hidd
 {{ page_variable_referencing_page_variable }}
 
 # Global Variable overriding Page Variable
-<variable name="page_global_variable_overriding_page_variable">Page Variable overridden by Global Variable</variable>
+<variable name="page_global_variable_overriding_page_variable">**TEST FAIL**: Page Variable overridden by Global Variable</variable>
 {{ page_global_variable_overriding_page_variable }}
 
 # Test Page Variable and Included Variable Integrations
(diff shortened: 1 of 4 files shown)


acjh:
I don't understand this. If outputs are incorrect, why is Travis passing?

jamos-tay:
I think it's helpful when verifying tests locally before pushing to travis, I usually check the built index.html because it's easier than sieving the console output. If say a test breaks, it displays this:

Incorrect:
image

Expected:
image

Which is hard to notice just by looking at the index.html. Basically we want to prefix values that are meant to be hidden/overridden with TEST FAIL so if the test breaks this shows up:

image

test/functional/test_site/index.md
```diff
@@ -43,7 +43,7 @@ head: myCustomHead.md, myCustomHead2.md
 {{ page_variable_referencing_page_variable }}
 
 # Global Variable overriding Page Variable
-<variable name="page_global_variable_overriding_page_variable">Page Variable overridden by Global Variable</variable>
+<variable name="page_global_variable_overriding_page_variable">**TEST FAIL**: Page Variable overridden by Global Variable</variable>
```
yamgent: Instead of `**TEST FAIL**`, maybe use `**Incorrect input**` or `**Should not appear**`?

Also, I was wondering whether there will be any value in converting some of these tests to unit tests, which might allow a cleaner implementation? Agree that generally it is difficult to read the .html output, although we can also consider breaking up `index.html` further because it is getting rather huge and messy. :P

yamgent: > ```markdown
> # Global Variable overriding Page Variable
> ```

Also since headings usually generate quite a long html code (because it also generates the anchor ID and the anchor icon itself), perhaps we can minimize the use of headings (only tests that are testing the features of headings should use it).

jamos-tay: Hmm sure I'll keep that in mind, I agree it might be helpful to do something about index before it gets too large

How about we convert them to `h4`s so they don't get indexed?

I'll use **Should not appear** as it is more descriptive

jamos-tay:
PRs are merged, so updated

test/functional/test_site/index.md
yamgent: `h4` still generates an unnecessary ID though.

```html
<h4 id="global-variable-overriding-page-variable">Global Variable overriding Page Variable</h4>
```

How about just using bold?

```markdown
**Global Variable overriding Page Variable**
```

jamos-tay: Sure - I can handle that in a separate PR.

In the meantime is this one okay?

PR #696, opened 2019-02-16
Replace <font> in user docs with <span style>

Xenonym:

• [X] Documentation update

As suggested in https://github.com/MarkBind/markbind/pull/680#pullrequestreview-204102924.

The `<font>` HTML tag is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font#Summary) and should not be used in our documentation.

I replaced uses of `<font color="[colour]">` in our documentation with `<span style="color: [colour]">`.

Note that the SVG files for FA icons use a `<font>` tag as well. While the SVG `<font>` tag is also [deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/font), we probably shouldn't arbitrarily modify upstream assets unless necessary.

### DIFF
--- a/docs/_markbind/variables.md
+++ b/docs/_markbind/variables.md
@@ -8,8 +8,8 @@
 
 <variable name="icon_arrow_down">:fas-arrow-down:</variable>
 <variable name="icon_arrow_right">:fas-arrow-right:</variable>
-<variable name="icon_check_blue"><font color="{{ markbind_blue }}">:fas-check-circle:</font></variable>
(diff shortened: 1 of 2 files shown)


Xenonym:
Proposed commit message:

```
Replace <font> in user docs with <span style>

There are multiple instances in our user docs where we use <font>
to inline style text. However, <font> has been deprecated [1] and
should not be used in our user docs.

Let's replace uses of <font> with <span style> instead to inline style
text in our user docs.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font#Summary
```

yamgent:
Nice :+1:.

Issue #660, opened 2019-02-03
Reasonable workflow support for GitHub-based projects


damithc:
- [x] A way to convert GitHub project to MarkBind
- [x] A way to auto-deploy to GitHub pages #457

Issue #597, opened 2019-01-19
Reduce replication of constants


marvinchin:
**Tell us about your environment**

* **MarkBind Version:**
MarkBind v1.16.1

**What did you expect to happen?**
Constants can be defined in a single file and reused where needed

**What actually happened? Please include the actual, raw output.**
Some constants are reused across `Site.js`, `Page.js`, and `cliUtil.js` (e.g. `LAYOUT_FOLDER_PATH`)

nicholaschuayunzhi:
For different sub logic it has been suggested in https://github.com/MarkBind/markbind/pull/698#discussion_r270254377 to have file naming structure of `{{main module}}.{{sub module}}.js`

Additionally it is suggested that constants should be placed in `constants.js` https://github.com/MarkBind/markbind/pull/698#discussion_r270965603

PR #698, opened 2019-02-17
Automatically convert GitHub-based projects into MarkBind websites

### LINKED ISSUE #675: Automatically convert GitHub Wikis to MarkBind sites

* **MarkBind Version:** 1.17.2

**Rationale:** [GitHub Wikis](https://help.github.com/articles/about-wikis/) are used commonly for project documentation (eg. [Angular.JS](https://github.com/angular/angular.js/wiki), [d3](https://github.com/d3/d3/wiki)). It would be viable to support automatic conversions for wikis as they are all flat directories of Markdown files in a Git repository, which should be relatively simple to implement.

In addition, we can also convert a wiki's [custom footers and sidebars](https://help.github.com/articles/creating-a-footer-or-sidebar-for-your-wiki/) (`_Footer.md`, `_Sidebar.md`) to our corresponding `footer.md` and `site-nav.md`. Additionally, many wikis use a sidebar that is generated by [`github-wiki-sidebar`](https://github.com/adriantanasa/github-wiki-sidebar), whose output can cleanly translate to our own `site-nav.md`:
```markdown
* [Installation](./Installation)
* [Roadmap](./Roadmap)
* [Usage](./Usage)
  * [Command line modifiers](./Usage%3A-Command-line-modifiers)
  * [Default mode (enquire)](./Usage%3A-Default-mode-(enquire))
  * [Tutorial mode](./Usage%3A-Tutorial-mode)
* [Examples](./Examples)
```
Alternatively, we can parse the file listing of the GitHub wiki to generate the site nav.

amad-person:

• [X] New feature

Fixes #675. Part of #660. 

Provides a way to convert an existing Github wiki or a docs folder into a MarkBind website. It will be easier to integrate MarkBind into the workflow for Github-based project.

1. Added a new option for the init command: `--convert`/`-c`. Users can run `markbind init -c` in the desired directory to convert it into a MarkBind project.
2. Converting the folder into a MarkBind project includes the following:
    - Running a usual `markbind init`.
    - Then, copying over contents of `README.md` (or `Home.md`) to `index.md`
    - Adding an `about.md` page if not present.
    - Adding a site navigation menu to all pages by updating `navigation.md` in the default layout. As mentioned in #675, if `_Sidebar.md` is present in the root directory, this will be used for the site navigation menu. Otherwise a menu will be generated.
    - Adding a footer to the website by updating `footer.md` in the default layout. As mentioned in #675, if `_Footer.md` is present in the root directory, this will be used for the site navigation menu. Otherwise the default footer generated by `markbind init` will be used.
    - Adding a top navigation bar to all pages. **This is done by adding a nav bar into `header.md` in the default layout.**
    - Updating site config so all pages have the default layout applied to them.

- Correctness of `test.bat` file.

**Testing instructions:**

Run `markbind init -c` in a folder of your choice and build the site. Run `npm run test(win)` to run the UTs and FT.

**Proposed Commit Message:**
```
Automatically convert GitHub-based projects into MarkBind websites

GitHub projects commonly have wikis or docs folders for project
documentation. As MarkBind is especially optimized as a project
documentation tool, we would want to make it easier for projects to
adopt it in their workflow.

Let's add the ability to automatically convert a GitHub wiki or docs
folder into a MarkBind website.
```


amad-person:
1. I'm not sure where to document this functionality. Should I add information in [MarkBind in the Project Workflow](https://markbind.org/userGuide/markBindInTheProjectWorkflow.html) and include a link to the section in [CLI Commands](https://markbind.org/userGuide/cliCommands.html)?

2. For testing this, should I add a dummy folder and provide an expected output folder (similar to our functional testing)?

damithc:
> 1. I'm not sure where to document this functionality. Should I add information in [MarkBind in the Project Workflow](https://markbind.org/userGuide/markBindInTheProjectWorkflow.html) and include a link to the section in [CLI Commands](https://markbind.org/userGuide/cliCommands.html)?

That sounds about right.

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+  const fileNames = ['README.md', 'Home.md'];
+  for (let i = 0; i < fileNames.length; i += 1) {
+    const filePath = path.join(rootPath, fileNames[i]);
+    if (fs.existsSync(filePath)) {
+      fs.copyAsync(filePath, indexPagePath)
+        .catch(() => Promise.reject(new Error(`Failed to copy over ${fileNames[i]}`)));
+      break;
+    }
+  }
+  return Promise.resolve();
```
nicholaschuayunzhi: This function resolves before `copyAsync` resolves. Maybe you should declare a variable above and reassign it to the new promise, after which you can return it at the end of the function?

```js

let promise = Promise.resolve();
...
// logic for copying file
promise = fs.copyAsync(filePath, indexPagePath)
        .catch(() => Promise.reject(new Error(`Failed to copy over ${fileNames[i]}`)));
...
return promise;
```

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+ * @param rootPath
+ */
+Site.prototype.addIndexPage = function (rootPath) {
+  const indexPagePath = path.join(rootPath, INDEX_MARKDOWN_FILE);
+  const fileNames = ['README.md', 'Home.md'];
+  for (let i = 0; i < fileNames.length; i += 1) {
+    const filePath = path.join(rootPath, fileNames[i]);
+    if (fs.existsSync(filePath)) {
+      fs.copyAsync(filePath, indexPagePath)
+        .catch(() => Promise.reject(new Error(`Failed to copy over ${fileNames[i]}`)));
```
nicholaschuayunzhi: This rejected promise never gets passed down the promise chain in `Site.convert`

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+      .then(() => this.addSiteNavToDefaultLayout(rootPath))
+      .then(() => this.addDefaultLayoutToSiteConfig(rootPath))
+      .then(() => this.addTopNavToAddressablePages(rootPath))
+      .then(resolve)
+      .catch(reject);
+  });
+};
+
+/**
+ * Adds an index page to site if not present.
```
nicholaschuayunzhi: From the logic is seems like `index.md` is generated only if `README.md` or `Home.md` are present which seems to contradict the comment.

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
   });
 };
 
+/**
+ * Converts an existing Github wiki or docs folder to a MarkBind website.
+ */
+Site.prototype.convert = function (rootPath) {
```
nicholaschuayunzhi: `rootPath` is already a member variable of site, We could use that instead.

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+  const siteLayoutPath = path.join(rootPath, LAYOUT_FOLDER_PATH);
+  const siteLayoutFooterDefaultPath = path.join(siteLayoutPath, LAYOUT_DEFAULT_NAME, 'footer.md');
+  const wikiFooterPath = path.join(rootPath, WIKI_FOOTER_PATH);
+
+  return new Promise((resolve, reject) => {
+    fs.accessAsync(wikiFooterPath)
+      .then(() => {
+        const footerContent = fs.readFileSync(wikiFooterPath, 'utf8');
+        const wrappedFooterContent = `<footer>\n\t${footerContent}\n</footer>`;
+        return fs.outputFileSync(siteLayoutFooterDefaultPath, wrappedFooterContent);
```
nicholaschuayunzhi: Could we use `outputFileAsync` here?

src/Site.js
acjh: @nicholaschuayunzhi Nice catch! It's clearer to `return fs.copyAsync(...).catch(...);` directly.

src/Site.js
```diff
@@ -104,6 +113,24 @@ const SITE_NAV_DEFAULT = '<navigation>\n'
   + '* [Home :glyphicon-home:]({{baseUrl}}/index.html)\n'
   + '</navigation>\n';
 
+const TOP_NAV_DEFAULT = '<navbar placement="top" type="inverse">\n'
+  + '  <a slot="brand" href="{{baseUrl}}/index.html" title="Home" class="navbar-brand">'
+  + '<i class="far fa-file-image"></i></a>\n'
+  + '  <li><a href="{{baseUrl}}/index.html" class="nav-link">HOME</a></li>\n'
+  + '  <li><a href="{{baseUrl}}/about.html" class="nav-link">ABOUT</a></li>\n'
+  + '  <li>\n'
+  + '    <a href="https://github.com/MarkBind/markbind" target="_blank" class="nav-link">\n'
```
nicholaschuayunzhi: The github link seem unrelated. Is there a way to infer the wiki's code repository?

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+ * Converts an existing Github wiki or docs folder to a MarkBind website.
+ */
+Site.prototype.convert = function (rootPath) {
+  return new Promise((resolve, reject) => {
+    this.addIndexPage(rootPath)
+      .then(() => this.addAboutPage(rootPath))
+      .then(() => this.addFooterToDefaultLayout(rootPath))
+      .then(() => this.addSiteNavToDefaultLayout(rootPath))
+      .then(() => this.addDefaultLayoutToSiteConfig(rootPath))
+      .then(() => this.addTopNavToAddressablePages(rootPath))
```
nicholaschuayunzhi: Could we remind user to update their `baseUrl` in `site.json` here?

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+Site.prototype.addSiteNavToDefaultLayout = function (rootPath) {
+  const siteLayoutPath = path.join(rootPath, LAYOUT_FOLDER_PATH);
+  const siteLayoutSiteNavDefaultPath = path.join(siteLayoutPath, LAYOUT_DEFAULT_NAME, 'navigation.md');
+  const wikiSiteNavPath = path.join(rootPath, WIKI_SITE_NAV_PATH);
+
+  return new Promise((resolve, reject) => {
+    fs.accessAsync(wikiSiteNavPath)
+      .then(() => {
+        const siteNavContent = fs.readFileSync(wikiSiteNavPath, 'utf8');
+        const wrappedSiteNavContent = `<navigation>\n${siteNavContent}\n</navigation>`;
```
nicholaschuayunzhi: For wiki's with side bars the urls are not resolved correctly:

https://github.com/SFML/SFML/wiki

image

nicholaschuayunzhi: Also the styling seems to be off to the side for both types of generation of site navs

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+      .then(() => this.readSiteConfig())
+      .then(() => this.collectAddressablePages())
+      .then(() => {
+        this.addressablePages
+          .filter(addressablePage => !addressablePage.src.startsWith('_'))
+          .forEach((page) => {
+            const addressablePagePath = path.join(rootPath, page.src);
+            const topNavRelativePath = path.relative(rootPath, topNavDefaultPath);
+            if (topNavRelativePath.length > 0) {
+              const includeTopNavCode = `<include src="${topNavRelativePath}" />\n\n`;
```
nicholaschuayunzhi: This seems quite expensive, i'm not too sure but is there a way to to specify this in layouts/front matter?

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
   });
 };
 
+/**
+ * Converts an existing Github wiki or docs folder to a MarkBind website.
+ */
+Site.prototype.convert = function (rootPath) {
+  return new Promise((resolve, reject) => {
+    this.addIndexPage(rootPath)
```
acjh: Do `return this.addIndexPage(rootPath)...` directly.

See: [What is the explicit promise construction antipattern and how do I avoid it?](https://stackoverflow.com/q/23803743/8601760)

src/Site.js
acjh: Also, I'm pretty sure `i` will always be `fileNames.length` in this arrow function.

src/Site.js
amad-person: > For wiki's with sidebars the urls are not resolved correctly:
You're right, I think I might need to preprocess `_Sidebar.md` instead of just copying it over.

> Also the styling seems to be off to the side for both types of generation of site navs
If I'm not wrong, it's because of a bug that's mentioned in #739.

src/Site.js
amad-person: Sorry, I'm not clear on what you mean...the `baseUrl` is set during `markbind init`, right?

src/Site.js
amad-person: > The github link seem unrelated. Is there a way to infer the wiki's code repository?

I'm not sure how to go about implementing this actually, should I let the user specify the Github URL as a parameter for the `--convert` command?

src/Site.js
amad-person: Noted, I'll update that!

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+          .forEach((page) => {
+            const addressablePagePath = path.join(rootPath, page.src);
+            const topNavRelativePath = path.relative(rootPath, topNavDefaultPath);
+            if (topNavRelativePath.length > 0) {
+              const includeTopNavCode = `<include src="${topNavRelativePath}" />\n\n`;
+              const addressablePageContent = fs.readFileSync(addressablePagePath, 'utf8');
+              const updatedAddressablePageContent = includeTopNavCode + addressablePageContent;
+              return fs.outputFileAsync(addressablePagePath, updatedAddressablePageContent);
+            }
+            return Promise.resolve();
```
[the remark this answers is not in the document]
amad-person: Noted, I'll update this.

src/Site.js
amad-person: As of now, there is no support for a `header.md` like we have with footers. 
I think I would probably need to use the functionality in #733 once it's added...

src/Site.js
amad-person: Thanks for the link! I'll make the changes.

src/Site.js
nicholaschuayunzhi: After `markbind init`, the `baseUrl` in `site.json` is the default `''`. This is unlikely to be the case for most github repositories. I was wondering if it would be good to remind the authors to update it appropriately, or even better, do it for them (maybe we could guess it based on the repository, similar to the navigation)

amad-person: Ah, I see! Perhaps we could set it to be the root directory name?

src/Site.js
nicholaschuayunzhi: Usually the wiki has a `.wiki` attached to the end. I'm thinking we could get the git remote and parse out the actual repository's url. This may require an external library though.

> should I let the user specify the Github URL as a parameter for the --convert command

I think its a bit weird to open an option just for that. Also what happens if they do not specify it? Alternatively i think we could write a comment and have the link point to MarkBind or we could just remove the link.

@Chng-Zhi-Xuan any thoughts?

Xenonym:
If you try `--convert` with [zaproxy/zaproxy/wiki](https://github.com/zaproxy/zaproxy/wiki), it will fail out of the box as the page [FAQformauth](https://github.com/zaproxy/zaproxy/wiki/FAQformauth) contains the text `username={%username%}&password={%password%}`, which Nunjucks tries and fails to parse as variables:
```
error: uncaughtException: (C:\Users\user\Desktop\zaproxy.wiki\FAQformauth.md) [Line 31, Column 114]
  unknown block tag: username date=Mon Mar 04 2019 16:21:20 GMT+0800 (Malay Peninsula Standard Time), pid=27616, uid=null, gid=null, cwd=C:\Users\user\Desktop\zaproxy.wiki, execPath=C:\Program Files\nodejs\node.exe, version=v8.15.0, argv=[C:\Program Files\nodejs\node.exe, C:\Program Files\nodejs\node_modules\markbind-cli\index.js, serve], rss=90669056, heapTotal=78016512, heapUsed=46236248, external=927768, loadavg=[0, 0, 0], uptime=142101
Template render error: (C:\Users\user\Desktop\zaproxy.wiki\FAQformauth.md) [Line 31, Column 114]
  unknown block tag: username
    at Object.exports.prettifyError (C:\Users\user\Documents\GitHub\CS3281-2\markbind\node_modules\nunjucks\src\lib.js:34:15)
    at new_cls.render (C:\Users\user\Documents\GitHub\CS3281-2\markbind\node_modules\nunjucks\src\environment.js:472:27)
    at new_cls.renderString (C:\Users\user\Documents\GitHub\CS3281-2\markbind\node_modules\nunjucks\src\environment.js:328:21)
    at Object.module.exports.renderString (C:\Users\user\Documents\GitHub\CS3281-2\markbind\node_modules\nunjucks\index.js:80:14)
    at fs.readFile (C:\Users\user\Documents\GitHub\CS3281-2\markbind\src\lib\markbind\src\parser.js:508:36)
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:511:3)
```

We should either parse the wiki pages to "escape" problematic syntax, or at least document that `--convert` may not work if there is MarkBind/Nunjucks syntax in wiki pages.

yamgent:
> We should either parse the wiki pages to "escape" problematic syntax, or at least document that `--convert` may not work if there is MarkBind/Nunjucks syntax in wiki pages.

I think in general, when we fail a conversion, we should just retain the original `.md` file content, but print a warning to say that we have failed for this `.md`, and continue on with our business.

For edge cases in general, we can try to apply fixes wherever possible, a good rule of thumb is that if the fix is not too complicated in code, and the fix is not "controversial" in nature, then we can consider handling the edge case. Otherwise we shouldn't worry about them too much.

For this edge case though, while I think escaping could be a good try, the problem is that there is no known consistent way of escaping Nunjucks stuff (this is a known bug btw, we have always been abusing `<span>` in the past to get around the problem of Nunjucks interpreting stuff that is not supposed to be interpreted as Nunjucks code).

src/Site.js
```diff
@@ -385,6 +412,193 @@ Site.prototype.createPage = function (config) {
...
+  return new Promise((resolve, reject) => {
+    fs.accessAsync(topNavDefaultPath)
+      .catch(() => {
+        if (fs.existsSync(topNavDefaultPath)) {
+          return Promise.resolve();
+        }
+        return fs.outputFileAsync(topNavDefaultPath, TOP_NAV_DEFAULT);
+      })
+      .then(() => this.readSiteConfig())
+      .then(() => this.collectAddressablePages())
```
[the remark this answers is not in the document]
amad-person: Removed the duplicate calls.

src/Site.js
amad-person: Currently, the logic is to copy over content from `README.md` or `Home.md` if they exist, to the default `index.md` that is generated during `markbind init`.
Updated the comment to reflect the current behaviour.

amad-person: @yamgent @acjh should we have some other default content in `index.md` for converted wikis/docs folders? Or is the current logic okay? 
Also, using `README.md` / `Home.md` seems a bit hacky, but I'm not sure how else to do this as of now...

amad-person:
Currently, the conversion process doesn't parse the `.md` files, so errors (if any) will arise only on `markbind build`. 

Any ideas for getting these build errors? Maybe I can parse the files before starting the conversion and log the errors - so users can fix them and start again?

nicholaschuayunzhi:
Personally I feel that if we handle nunjucks error well, the author would be able to identify the problem and adjust accordingly? Perhaps this isn't in the scope of this feature? @yamgent Do you agree?

yamgent:
> Personally I feel that if we handle nunjucks error well, the author would be able to identify the problem and adjust accordingly? Perhaps this isn't in the scope of this feature? @yamgent Do you agree?

I agree that this PR shouldn't be going inside the nunjucks code and modify stuff for this current edge case, the fixing of the errors should be done manually by the author. Printing a user-friendly message will make that experience better, so I agree that this isn't the scope of the current PR.

src/Site.js
yamgent: Apologies for the late reply.

We can use `README.md` as `index.md` if there isn't already an original `index.md`. This behaviour is not dissimilar to how [GitHub Pages](https://github.blog/2016-12-09-publishing-with-github-pages-now-as-easy-as-1-2-3/) do it, so if we want to mimic this behaviour, I think it is reasonable.

If neither `README.md`/`Home.md` nor `index.md` is present, which means we cannot and won't have an `index.html`, doing a `markbind serve` will cause it to present to the user a list of files and directories on the root path. To me, I think that is OK, since that way authors can easily check the content of different .html files, rather than having to look at a default index.html which he is going to change anyway. In short: Nope, probably don't need to supply a default `index.md` for that situation.

amad-person:
So, would adding the following in this PR be okay?

1. Printing a message to the console 
    > `--convert` may not work if there is MarkBind/Nunjucks syntax 

    (for all cases, whether there might be build errors or not)

2. Adding this warning to the documentation.

yamgent:
> 1. Printing a message to the console
>    > `--convert` may not work if there is MarkBind/Nunjucks syntax
>    
>    
>    (for all cases, whether there might be build errors or not)

I don't think this message would be very useful, such messages are generally not paid attention to, because it works, the error message is redundant, and if it doesn't, users will see the error churned out by nunjucks and may not make the connections between the error and this message.

The documentation is fine and we should definitely include it.

src/Site.js
amad-person: Thanks!

amad-person:
**Updates** after discussion -
Print a message to let users know they can update `{{ baseUrl }}`.
Two formats of site navigation files need to be supported:
1. `[[Name|URL]]`
2. `[Name](URL)`

src/Site.js
```diff
@@ -387,6 +414,158 @@ Site.prototype.createPage = function (config) {
...
+    .then(() => this.addTopNavToAddressablePages());
+};
+
+/**
+ * Copies over README.md or Home.md to default index.md if present.
+ */
+Site.prototype.addIndexPage = function () {
+  const indexPagePath = path.join(this.rootPath, INDEX_MARKDOWN_FILE);
+  const fileNames = ['README.md', 'Home.md'];
+  for (let i = 0; i < fileNames.length; i += 1) {
```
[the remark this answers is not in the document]
amad-person: What is the convention to break out of `forEach()`?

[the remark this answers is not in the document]
acjh: We should not break out of `forEach`. [`fileNames.find(...)`](
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) is more appropriate.

src/util/fsUtil.js
```diff
@@ -28,4 +28,12 @@ module.exports = {
     path.dirname(normalizedFilename),
     path.basename(normalizedFilename, path.extname(normalizedFilename)) + ext,
   ),
+
+  removeExtension: (filePathWithExt) => {
```
marvinchin: Just wondering - would a file path with `.` in the path but not the file itself pose a problem? (e.g. `~/.folderWithDot/fileWithNoExt`)

Would doing the check on the last item in the path, rather than the entire path itself, resolve this problem?

Xenonym: Could `removeExtension` be written as:

```js
removeExtension: (filePathWithExt) => path.join(
	path.dirname(filePathWithExt),
	path.basename(filePathWithExt, path.extname(filePathWithExt)),
),
```
Should achieve the same effect, but using standard library components. `setExtension` can then be refactored to use `removeExtension`.

src/Site.js
```diff
@@ -387,6 +414,166 @@ Site.prototype.createPage = function (config) {
...
+            const updatedAddressablePageContent = includeTopNavCode + addressablePageContent;
+            return fs.outputFileAsync(addressablePagePath, updatedAddressablePageContent);
+          }
+          return Promise.all(outputFilesPromises);
+        });
+    });
+};
+
+Site.prototype.printBaseUrlMessage = function () {
+  logger.info('The default base URL of your site is set to \'\'.\n'
```
yamgent: The default is actually `/`, not `\`.

src/Site.js
```diff
@@ -46,12 +49,15 @@ const PROJECT_PLUGIN_FOLDER_NAME = '_markbind/plugins';
 const SITE_CONFIG_NAME = 'site.json';
 const SITE_DATA_NAME = 'siteData.json';
 const SITE_NAV_PATH = '_markbind/navigation/site-nav.md';
+const TOP_NAV_PATH = '_markbind/common/topNav.md';
```
yamgent: `topNav` -> `top-nav`? (to be more consistent with the file naming convention in general).

src/Site.js
```diff
@@ -84,6 +90,9 @@ const SITE_CONFIG_DEFAULT = {
   },
 };
 
+const ABOUT_MARKDOWN_DEFAULT = '# About\n'
+  + 'Welcome to your **About Us** page.\n';
```
yamgent: Just curious, is an "About" page a default page provided by GitHub wikis?

src/Site.js
yamgent: If there's no easy way of deducing the actual repository's URL, it is OK to not include this.

src/Site.js
amad-person: Nope, it isn't provided by default. I had actually discussed this with prof @damithc and we decided to include an about.md page as this conversion feature works for both wikis and `docs` folders.

yamgent: I see, thanks for the clarification

src/util/fsUtil.js
yamgent: > `setExtension` can then be refactored to use `removeExtension`.

@amad-person Do refactor the code as suggested, thanks!

src/Site.js
```diff
@@ -387,6 +409,191 @@ Site.prototype.createPage = function (config) {
...
+    .catch(() => this.buildSiteNav(siteLayoutSiteNavDefaultPath));
+};
+
+/**
+ * Parses _Sidebar.md to create a MarkBind friendly site navigation menu.
+ * Formats supported: [[Name|URL]] and [Name](URL).
+ */
+Site.prototype.parseSidebarFileContent = function (rawSiteNavContent) {
+  let siteNavContent = '';
+  const doubleSquareBracketsFormatRegex = /\[\[.+\|{0,1}.?\]\]/g;
```
yamgent: `{0,1}` can be a `?` instead?

src/Site.js
```diff
@@ -387,6 +409,191 @@ Site.prototype.createPage = function (config) {
...
+  const markdownLinkFormatRegex = /(?:__|[*#])|\[(.*?)\]\(.*?\)/g;
+  const baseUrlSubst = '{{baseUrl}}';
+  if (rawSiteNavContent.match(doubleSquareBracketsFormatRegex)) {
+    rawSiteNavContent.split('\n').forEach((navLine) => {
+      const nameStartIdx = navLine.indexOf('[[');
```
yamgent: There seems to be two different methods employed to search for double square brackets.  I believe regular expression has a way for you to extract out the URL text and URL link itself, so that we don't have to manually use `.indexOf()` and etc. to manually extract them.

src/Site.js
```diff
@@ -412,6 +434,188 @@ Site.prototype.createPage = function (config) {
...
+        .forEach((page) => {
+          const addressablePagePath = path.join(this.rootPath, page.src);
+          const topNavRelativePath = path.relative(this.rootPath, topNavDefaultPath);
+          if (topNavRelativePath.length > 0) {
+            const includeTopNavCode = `<include src="${topNavRelativePath}" />\n\n`;
+            const addressablePageContent = fs.readFileSync(addressablePagePath, 'utf8');
+            const updatedAddressablePageContent = includeTopNavCode + addressablePageContent;
+            return fs.outputFileAsync(addressablePagePath, updatedAddressablePageContent);
+          }
+          return Promise.all(outputFilesPromises);
```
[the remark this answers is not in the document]
acjh: In addition, some statements can be moved out of `map`.
```js
const topNavRelativePath = path.relative(this.rootPath, topNavDefaultPath);
if (topNavRelativePath.length == 0) {
    return Promise.resolve();
}
const includeTopNavCode = `<include src="${topNavRelativePath}" />\n\n`;
const outputFilesPromises = this.addressablePages
    .filter(addressablePage => !addressablePage.src.startsWith('_'))
    .map((page) => {       
        const addressablePagePath = path.join(this.rootPath, page.src);
        const addressablePageContent = fs.readFileSync(addressablePagePath, 'utf8');
        const updatedAddressablePageContent = includeTopNavCode + addressablePageContent;
        return fs.outputFileAsync(addressablePagePath, updatedAddressablePageContent);
    });
return Promise.all(outputFilesPromises);
```

This function modifies the source files?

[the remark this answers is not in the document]
nicholaschuayunzhi: This function inserts a top nav to every addressable page. I dont think there's a clean way to insert a top nav without modifying the original wiki site in some way.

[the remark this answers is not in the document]
acjh: Oh, this is specific to conversion. If not entirely possible to abstract this out of `Site`, then shall we at least put this in a different file (e.g. Site.convert.js)? We can expect the caller to `require` that file after Site.js.

src/Site.js
```diff
@@ -412,6 +434,188 @@ Site.prototype.createPage = function (config) {
...
+/**
+ * Parses _Sidebar.md to create a MarkBind friendly site navigation menu.
+ * Formats supported: [[Name|URL]] and [Name](URL).
+ */
+Site.prototype.parseSidebarFileContent = function (rawSiteNavContent) {
+  let siteNavContent = '';
+  const doubleSquareBracketsFormatRegex = /\[\[([\w-]*)(\|)?([\w-]*)?\]\]/g;
+  const markdownLinkFormatRegex = /(?:__|[*#])|\[(.*?)\]\(.*?\)/g;
+  const baseUrlSubst = '{{baseUrl}}';
+  if (rawSiteNavContent.match(doubleSquareBracketsFormatRegex)) {
```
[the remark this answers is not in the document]
amad-person: Noted, I'll update the logic as discussed in the meeting.

src/Site.js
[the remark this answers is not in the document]
nicholaschuayunzhi: @amad-person could you update the feature based on https://github.com/MarkBind/markbind/pull/698#discussion_r270254377 ?

test/unit/Site.test.js
```diff
@@ -351,6 +353,69 @@ test('Site read correct user defined variables', async () => {
...
+    'node_modules/@fortawesome/fontawesome-free/css/all.min.css': '',
+    'node_modules/@fortawesome/fontawesome-free/webfonts/font1.svg': '',
+    'node_modules/@fortawesome/fontawesome-free/webfonts/font2.ttf': '',
+
+    'inner/_markbind/layouts/default/footer.md': '',
+    'inner/_markbind/layouts/default/head.md': '',
+    'inner/_markbind/layouts/default/header.md': '',
+    'inner/_markbind/layouts/default/navigation.md': '',
+    'inner/_markbind/layouts/default/scripts.js': '',
+    'inner/_markbind/layouts/default/styles.css': '',
```
nicholaschuayunzhi: I think you should test that the original site was just a wiki and then you call `init` and subsequently `convert`. No need for `generate`.

src/Site.js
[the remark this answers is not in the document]
nicholaschuayunzhi: @acjh @yamgent  how should we handle the constants in `Site.js` used in the conversion feature. It seems like defining constants in that manner is not good for modularity.

Should we:
 - Make the used constants static variables of `Site`
 - Copy the constants in `Site.convert.js` and refactor in the future
 - Keep the feature in `Site` and refactor everything in one shot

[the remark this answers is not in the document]
yamgent: > * Copy the constants in `Site.convert.js` and refactor in the future

This seems like a pretty good option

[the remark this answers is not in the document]
acjh: Projects usually have file(s) for constants, e.g. https://github.com/vuejs/vue/blob/dev/src/shared/constants.js

Let's start with `src/constants.js`.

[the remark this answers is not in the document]
nicholaschuayunzhi: Thanks for the input @acjh @yamgent. From discussion today, we already have an issue tracking this https://github.com/MarkBind/markbind/issues/597. Perhaps the change is out of scope of the pull request? Perhaps we shall do all the refactoring in one shot and when there is less on going development at the moment. 

So to be clear what i mean is that the conversion logic will be be in `Site.js`.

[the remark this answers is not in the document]
yamgent: I am cool with it.

amad-person:
#### Updates after discussion on 04/04/19

We decided to do the following for generating site navigation:
1. If `_Sidebar.md` exists, copy over the file without parsing it. Display a message to the user saying they might need to modify the generated site navigation if needed. 
2. Otherwise, generate site navigation based on folder structure (this is being done already).

Reasons: There are a couple of edge cases that we didn't think about that makes parsing `_Sidebar.md` difficult. 😢 
For example,

```
[![SFML logo](https://www.sfml-dev.org/images/logo.png)](https://www.sfml-dev.org)
```
1. Parsing this correctly would require testing `` pattern does not exist after extracting the text in `[]`.
2. The link is an external URL, so a check would need to be done before converting it into a MarkBind-friendly link (i.e. adding `{{baseUrl}}` and `.html`).

Additionally, we don't have a hold on the number of formats supported by a Github wiki's `_Sidebar.md` as this isn't documented officially. So far we came across two formats. 

It was decided that this functionality can be added in a future PR if needed 😅

test/unit/Site.test.js
yamgent: Has this already been addressed? Resolve this conversation if it has.

amad-person:
Noted, I will rebase and group the necessary commits together!

amad-person:
Updates: 

- Changed the top navigation logic to use the new `header.md` feature. The conversion feature still works as expected.

- Added unit and functional tests. 
I modified both `test.sh` and `test.bat` to execute the `init -c` command before building and comparing the files. For the functional test, I had to update `test.js` to take in an argument for the expected output directory for `test_site_convert`. This is because unlike other FTs, this feature needs to convert the `test_site_convert` folder first. If the expected output is in this folder, this will also get 'converted'. Hence, I put the expected output in a separate folder and modified `test.js` accordingly, hope this is okay! 

Please check if the `test.bat` file has been updated correctly!

test/functional/test.bat
```diff
@@ -17,5 +17,28 @@ for %%a in (%sites%) do (
...
+
+    node ../../index.js build %%a
+
+    node testUtil/test.js %%a "test_site_convert_expected"
+
+    rmdir /s /q %%a/_markbind
+    rmdir /s /q %%a/_site
+    del %%a/about.md %%a/index.md %%a/site.json
+
+    if errorlevel 1 (
```
yamgent: This is the wrong `errorlevel`; this `errorlevel` belongs to the `del` command rather than the `node testUtil/test.js` command. We want to test whether `node testUtil/test.js` failed, rather than whether `del` failed.

test/functional/test.sh
```diff
@@ -21,6 +21,34 @@ do
...
+    node ../../index.js build "$site"
+
+    # run our test script to compare html files
+    node testUtil/test.js "$site" "test_site_convert_expected"
+
+    # delete generated files
+    rm -rf "$site"/_markbind "$site"/_site
+    rm "$site"/about.md "$site"/index.md "$site"/site.json
+
+    if [ $? -ne 0 ]
```
yamgent: Same problem as the Windows' version: the return code `$?` belong to the `rm` command, rather than the `node testUtil/test.js` command.

test/functional/test.sh
```diff
@@ -21,6 +21,34 @@ do
...
+    echo "Running $site tests"
+
+    # convert site
+    node ../../index.js init "$site" -c
+
+    # build site
+    node ../../index.js build "$site"
+
+    # run our test script to compare html files
+    node testUtil/test.js "$site" "test_site_convert_expected"
```
yamgent: This shouldn't be hard code, it defeats the purpose of having the `sites_convert` array in the first place.

test/functional/testUtil/test.js
```diff
@@ -11,7 +11,8 @@ function readFileSync(...paths) {
 }
 
 const sitePath = process.argv[2];
-const expectedDirectory = path.join(sitePath, 'expected');
+const expectedPath = process.argv[3];
+const expectedDirectory = expectedPath === undefined ? path.join(sitePath, 'expected') : expectedPath;
```
yamgent: This seems rather unnecessary. `test.js` has no business in knowing that `test_site_convert` exist. Instead, `sitePath` should be `test_site_convert_expected`, so that this addition is not needed.

Please also rename the directories accordingly:

* `test_site_convert/` -> `test_site_convert/non_markbind_site/` 
* `test_site_convert_expected/` -> `test_site_convert/`

test/functional/test_site_convert_expected/_Sidebar.html
```diff
@@ -0,0 +1,71 @@
+<!DOCTYPE html>
```
yamgent: @nicholaschuayunzhi I find it quite strange that `_Sidebar.md` (and `_Footer.md`) is getting rendered here, but I understand that we don't really have a way to stop MarkBind from rendering certain files. Your thoughts?

docs/userGuide/markBindInTheProjectWorkflow.md
```diff
@@ -36,5 +36,29 @@ You can keep the user docs in a separate directory (say `user-docs`) and set up
 Similarly, you can keep the dev docs in a separate directory (sey `dev-docs`) and set up Netlify to deploy the site when there is an update to the `master` branch; that way, developers can see the latest version of dev-docs via the Netlify site.
 </div>
 
+#### Converting an existing Github project
```
nicholaschuayunzhi: How about:
```suggestion
#### Converting existing project documentation/wiki
```
We would want index `wiki` in the search.

docs/userGuide/markBindInTheProjectWorkflow.md
```diff
@@ -36,5 +36,29 @@ You can keep the user docs in a separate directory (say `user-docs`) and set up
...
+
+MarkBind supports the automatic conversion of an existing Github wiki or `docs` folder containing Markdown files.
```
nicholaschuayunzhi:
```suggestion
MarkBind supports the automatic conversion of an existing GitHub wiki or `docs` folder containing Markdown files.
```

docs/userGuide/markBindInTheProjectWorkflow.md
```diff
@@ -36,5 +36,29 @@ You can keep the user docs in a separate directory (say `user-docs`) and set up
...
+
+A MarkBind conversion involves the following:
+1. Adding a Home page: If your project already has a `README.md` or `Home.md`, the content will be copied over to `index.md`. Otherwise, a default home page will be added.
```
nicholaschuayunzhi: For numbered lists, instead of `1. 2. 3. 4.`, you can repeat the `1.s` to make editing easier

test/unit/Site.test.js
```diff
@@ -351,6 +353,86 @@ test('Site read correct user defined variables', async () => {
   expect(subsub.number).toEqual('9999');
 });
 
+test('Site convert generates correct assets', async () => {
+  const json = {
+    'src/template/page.ejs': PAGE_EJS,
+
+    'inner/index.md': INDEX_MD_DEFAULT,
+
```
nicholaschuayunzhi: Do remove these lines 😅

test/unit/Site.test.js
```diff
@@ -351,6 +353,86 @@ test('Site read correct user defined variables', async () => {
...
+  const paths = Object.keys(fs.vol.toJSON());
+  const originalNumFiles = Object.keys(json).length;
+  const expectedNumBuilt = 15;
+  expect(paths.length).toEqual(originalNumFiles + expectedNumBuilt);
+
+  // footer
+  expect(fs.existsSync(path.resolve('inner/_markbind/layouts/default/footer.md'))).toEqual(true);
+
+  // site navigation
+  expect(fs.existsSync(path.resolve('inner/_markbind/layouts/default/navigation.md'))).toEqual(true);
```
nicholaschuayunzhi: If possible, are you able to check the contents that are auto-generated here?

test/unit/Site.test.js
```diff
@@ -351,6 +353,86 @@ test('Site read correct user defined variables', async () => {
...
+    'src/template/page.ejs': PAGE_EJS,
+
+    'inner/index.md': INDEX_MD_DEFAULT,
+
+    'inner/_Footer.md': '',
+    'inner/_Sidebar.md': '',
+    'inner/Home.md': '',
+  };
+  fs.vol.fromJSON(json, '');
+  await Site.initSite('inner/');
```
nicholaschuayunzhi: Since we don't have a functional test for site conversion, could we check for `site.json` and `_markbind` directory here?

src/Site.js
```diff
@@ -435,6 +456,140 @@ Site.prototype.createPage = function (config) {
...
+  const siteLayoutSiteNavDefaultPath = path.join(siteLayoutPath, LAYOUT_DEFAULT_NAME, 'navigation.md');
+  const wikiSiteNavPath = path.join(this.rootPath, WIKI_SITE_NAV_PATH);
+
+  return fs.accessAsync(wikiSiteNavPath)
+    .then(() => {
+      const siteNavContent = fs.readFileSync(wikiSiteNavPath, 'utf8');
+      const wrappedSiteNavContent = `<navigation>\n${siteNavContent}\n</navigation>`;
+      logger.info(`Copied over the existing _Sidebar.md file to ${path.relative(this.rootPath,
+                                                                                siteLayoutSiteNavDefaultPath)
+      }\n\tCheck https://markbind.org/userGuide/tweakingThePageStructure.html#site-navigation-menus`
```
nicholaschuayunzhi: The `\t` here mis-aligns the message. While you could add 6 spaces, I think its not very clean. I'm fine with logging another info line directly after.

docs/userGuide/markBindInTheProjectWorkflow.md
```diff
@@ -36,5 +36,29 @@ You can keep the user docs in a separate directory (say `user-docs`) and set up
...
+2. Adding an About Us page: If your project already an `About` page, the content will be copied over to `about.md`. Otherwise, a default about page will be added.
+3. Adding a top navigation bar.
+4. Adding a site navigation menu: If your project has a valid `_Sidebar.md` file, it will be used as the site navigation menu. Otherwise, the menu will be built from your project's directory structure and contain links to all addressable pages.   
+5. Adding a custom footer: If your project has a valid `_Footer.md` file, it will be used as the website footer. Otherwise, a default footer will be added.
+
+<box type="warning">
+    Conversion might not work if your project files have existing MarkBind/Nunjucks syntax.
```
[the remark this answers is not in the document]
amad-person: This was mentioned in a [comment](https://github.com/MarkBind/markbind/pull/698#issuecomment-469166028) in this PR

src/Site.js
amad-person: I'll log the info line directly after.

test/functional/testUtil/test.js
amad-person: After renaming the directories as suggested, I'm not sure I understand...if the `sitePath` is `test_site_convert`, won't the conversion be run on this folder instead of the `non_markbind_site`?

test/unit/Site.test.js
amad-person: I added in a functional test for site conversion, could you help me take a look 😅

test/functional/testUtil/test.js
yamgent: Sorry, I forgot that the `init` was in place. Is it not possible to modify `test.sh/bat` to run init on `test_site_convert/non_markbind_site/`, then copy the output files to `test_site_convert/`?

My goal here is to make the test as self-contained as possible, so that (1) there would not be two different separate folders for the same test case, and that (2) we can reuse the original logic of comparing with the `expected/` folder. This is to make the maintenance easier so that people don't get confused about a different test folder structure.

docs/userGuide/markBindInTheProjectWorkflow.md
yamgent: Would the numbering make sense here, as these are not really steps for the authors? Should we use bullet points here instead?

test/unit/Site.test.js
nicholaschuayunzhi: The functional test checks for the built converted site but not the converted files. Sorry if i wasn't clear.

docs/userGuide/markBindInTheProjectWorkflow.md
amad-person: As they aren't steps that readers need to follow, using bullet points makes sense...I've updated the documentation to reflect this.

docs/userGuide/markBindInTheProjectWorkflow.md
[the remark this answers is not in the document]
amad-person: I have removed `MarkBind` from `MarkBind/Nunjucks syntax`

test/functional/testUtil/test.js
amad-person: Updated to follow logic you've mentioned, please take a look!

test/functional/test_site_convert_expected/_Sidebar.html
nicholaschuayunzhi: Thanks for pointing that out @yamgent . If users would want to prevent rendering of `_Sidebar.md`, they would have to specify their own addressable pages. 

That being said, I can understand why it would it be confusing and annoying to users to see additional html files that do not make sense (eg Sidebar with header and footer). I'm thinking we may need a page ignore mechanism to prevent its rendering as an addressable page which should be out of scope for this PR. Maybe the files prefixed with `_` should not be addressable?

Alternatives would be to remove the said files as they are no longer used by the site but I'm afraid of tampering with the author's original files as he may want to reuse them/refer to them locally. Potentially in the future, we could store all the original unused files into some location maybe `_original/` and ignore the entire folder for finding addressable pages.

docs/userGuide/markBindInTheProjectWorkflow.md
```diff
@@ -36,5 +36,29 @@ You can keep the user docs in a separate directory (say `user-docs`) and set up
 Similarly, you can keep the dev docs in a separate directory (sey `dev-docs`) and set up Netlify to deploy the site when there is an update to the `master` branch; that way, developers can see the latest version of dev-docs via the Netlify site.
 </div>
 
+#### Converting existing project documentation/wiki
+
+MarkBind supports the automatic conversion of an existing GitHub wiki or `docs` folder containing Markdown files.
+
+A MarkBind conversion involves the following:
+- Adding a Home page: If your project already has a `README.md` or `Home.md`, the content will be copied over to `index.md`. Otherwise, a default home page will be added.
+- Adding an About Us page: If your project already an `About` page, the content will be copied over to `about.md`. Otherwise, a default about page will be added.
```
yamgent: > - Adding an About Us page: If your project already an `About` page, ...

Do our code actually check for this?

src/Site.js
```diff
@@ -435,6 +456,140 @@ Site.prototype.createPage = function (config) {
...
+};
+
+/**
+ * Adds top navigation menu to default layout of site.
+ */
+Site.prototype.addTopNavToDefaultLayout = function () {
+  const siteLayoutPath = path.join(this.rootPath, LAYOUT_FOLDER_PATH);
+  const siteLayoutHeaderDefaultPath = path.join(siteLayoutPath, LAYOUT_DEFAULT_NAME, 'header.md');
+
+  return fs.outputFileAsync(siteLayoutHeaderDefaultPath, TOP_NAV_DEFAULT);
```
yamgent: Does this mean that, if an author calls `markbind init -c` on an existing site (could be by accident), `_markbind/layout/default/header.md` will always be overridden even if it already exist?

test/functional/test_site_convert_expected/_Sidebar.html
yamgent: Thanks for your response. In view of the limited time, let's consider such issues in a future PR.

docs/userGuide/markBindInTheProjectWorkflow.md
amad-person: Yes, the conversion process checks for `about.md` and uses it if present.

amad-person: I should word it differently: `If your project already has an about.md, it will be used as the About page. ...`

src/Site.js
amad-person: Yep, you're right. I'm not sure how to prevent this from happening...since the use case doesn't consider running a conversion more than once / running it on an existing MarkBind site

amad-person:
> The conversion crashes if I do the following:
> 
> 1. Create an empty folder with a `test.md` (content of `test.md` does not matter).
> 2. Run `markbind init -c`.
> 3. MarkBind crashes with the following error message, and the conversion is not complete.
> 
> ```
> > markbind init -c
>   __  __                  _      ____    _               _ 
>  |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
>  | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
>  | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
>  |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|
>                                                            
>  v2.1.1
> info: Initialization success.
> info: Converting to MarkBind website.
> error: Path must be a string. Received undefined
> ```

I have pushed the fix for this! It was crashing in cases where `Home.md` or `README.md` were not present in the directory.

src/Site.js
acjh: Check for `site.json`?

nicholaschuayunzhi: How about if `init` is ran with `-c` flag, we check for `site.json` first?

yamgent:
> I have pushed the fix for this! It was crashing in cases where `Home.md` or `README.md` were not present in the directory.

Thanks for the fix, it works now.

On a side note, with this use case (only `test.md`), the default `index.md` will look out of place (because it uses the wrong topNav and siteNav). However, I will argue that maybe it is just that our default `index.md` weren't that useful to begin with, so I am OK with leaving it as it is for now, and tackle that problem in future PRs.

src/Site.js
yamgent: Um how about just checking for the existence of `_markbind/layout/default/header.md` and not doing anything if it exist? This would be in line with how we deal with `about.md`, and our general policy of not touching existing files with our defaults for other kinds of files as well (remember that `markbind init` itself can also be run more than once).

amad-person: But `markbind init` definitely creates a `header.md` file, so this check would mean the top nav content will never be written to `header.md`.

amad-person:
Pushed a fix for aborting conversion on existing MarkBind sites using a `site.json` check.

test/functional/test.sh
```diff
@@ -21,6 +21,35 @@ do
...
+if [ $? -ne 0 ]
+then
+    echo "Test result: $site FAILED"
+    exit 1
+fi
+
+# delete generated files
+rm -rf "$site_convert"/_site
+rm -rf "$site_convert"/non_markbind_site/_markbind "$site_convert"/non_markbind_site/_site
+rm "$site_convert"/non_markbind_site/about.md "$site_convert"/non_markbind_site/index.md "$site_convert"/non_markbind_site/site.json
```
nicholaschuayunzhi: It test fails, the generated files will be left in the test site.

test/functional/test.bat
```diff
@@ -17,5 +17,31 @@ for %%a in (%sites%) do (
...
+    xcopy %%a\non_markbind_site\_site %%a /i
+
+    node testUtil/test.js %%a
+
+    if errorlevel 1 (
+        echo Test %%a Failed
+        exit /b %errorlevel%
+    )
+
+    rmdir /s /q %%a\_site
```
nicholaschuayunzhi: If test fails, generated files will left in the test site.

test/unit/Site.test.js
nicholaschuayunzhi: Was this addressed else where?

If no I was thinking of doing a check on the content of the navigation file here. The file is generated by the conversion and the results should be predictable.

docs/userGuide/cliCommands.md
```diff
@@ -76,10 +76,13 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
 * `<root>`<br>
   Root directory. Default is the current directory.<br>
   {{ icon_example }} `./myWebsite`
+* `-c`, `--convert`<br>
+  Convert an existing Github wiki or `docs` folder into a MarkBind website. See [Converting an existing Github project]({{ baseUrl }}/userGuide/markBindInTheProjectWorkflow.html#converting-an-existing-github-project) for more information.
```
nicholaschuayunzhi:
```suggestion
  Convert an existing GitHub wiki or `docs` folder into a MarkBind website. See [Converting an existing Github project]({{ baseUrl }}/userGuide/markBindInTheProjectWorkflow.html#converting-an-existing-github-project) for more information.
```

docs/userGuide/cliCommands.md
```diff
@@ -76,10 +76,13 @@ MarkBind Command Line Interface (CLI) can be run in the following ways:
...
 
 {{ icon_examples }}
 * `markbind init` : Initializes the site in the current working directory.
 * `markbind init ./myWebsite` : Initializes the site in `./myWebsite` directory.
+* `markbind init --convert` : Converts the current working directory into a MarkBind website.
```
nicholaschuayunzhi:
```suggestion
* `markbind init --convert` : Converts the Github wiki or `docs` folder in the current working directory into a MarkBind website.
```

test/unit/Site.test.js
amad-person: > Was this addressed else where?

Sorry, I forgot to mention that I put the content check in the second unit test (in `Site.test.js#419`). Is that what you mean?

test/functional/test.sh
amad-person: Pushed a fix for this using `trap` in `test.sh`.

test/functional/test.bat
amad-person: I used `trap` in `test.sh`, but I couldn't find a batch equivalent. Is there a better fix than copying over the delete commands to the `if` block?

nicholaschuayunzhi: I think you can set a variable and check after removing the directories. 

https://stackoverflow.com/questions/19300813/how-to-store-errorlevel-in-a-variable

nicholaschuayunzhi:
@amad-person thanks for the PR. Could you write a a Proposed Commit message for the PR (to be updated in the PR description)?

amad-person:
@nicholaschuayunzhi @yamgent updated PR description with the proposed commit message.

Issue #662, opened 2019-02-03
Give a decent ‘first contact’ impression


damithc:
- [x] Use a custom domain
- [x] No redirect in landing page

amad-person:
The custom domain we're going to use is `https://markbind.github.io` (mentioned in #626)?.

damithc:
> The custom domain we're going to use is `https://markbind.github.io` (mentioned in #626)?.

No, https://markbind.org

amad-person:
So I have to get this custom domain registered first?

damithc:
> So I have to get this custom domain registered first? Or do we already own this?

I'll take care of the first item after #626 is done.

amad-person:
For #626, I just have to configure the repo and branch settings in `site.json` and do a `markbind deploy`, right? Because I don't have access to the [markbind.github.io](https://github.com/MarkBind/markbind.github.io) repo.

damithc:
> For #626, I just have to configure the repo and branch settings in `site.json` and do a `markbind deploy`, right? Because I don't have access to the [markbind.github.io](https://github.com/MarkBind/markbind.github.io) repo.

I think so. @yamgent what do you think?

yamgent:
> For #626, I just have to configure the repo and branch settings in `site.json` and do a `markbind deploy`, right? Because I don't have access to the [markbind.github.io](https://github.com/MarkBind/markbind.github.io) repo.

Yes that should work.

amad-person:
I'm getting the following error as I don't have access rights to the `markbind.github.io` repository.

```
error: remote: Permission to MarkBind/markbind.github.io.git denied to amad-person.
fatal: unable to access 'https://github.com/MarkBind/markbind.github.io.git/': The requested URL returned error: 403
```

Changes made to `site.json`:
```
"baseUrl": "/markbind.github.io"
```
```
  "deploy": {
    "message": "Site Update.",
    "repo": "https://github.com/MarkBind/markbind.github.io.git",
    "branch": "master"
  }
```

damithc:
> I'm getting the following error as I don't have access rights to the `markbind.github.io` repository.

You don't need to actually deploy. @yamgent will deploy when along with the next release.

amad-person:
Okay!

damithc:
> `"baseUrl": "/markbind.github.io"`

I suspect this should be just `/`

damithc:
Closing this as both items are done now.

PR #699, opened 2019-02-17
Migrate MarkBind's website to markbind.github.io

### LINKED ISSUE #626: Deploy MarkBind website to markbind.github.io
Current: https://markbind.github.io redirects to https://markbind.github.io/markbind/
Suggested: Deploy the released version of the MarkBind website to markbind.github.io instead as it is not nice when the landing page is a redirect.

Note: markbind.github.io only supports serving from the master branch (does not support serving from the gh-pages)

amad-person:

• [X] Other, please explain:

Deploy the MarkBind website to `https://markbind.github.io/` instead of a redirect.

Fixes #626. 
Part of #662.

Currently, `https://markbind.github.io` redirects to `https://markbind.github.io/markbind/`.
These changes will deploy the released version of the MarkBind website directly to `https://markbind.github.io`  instead.

Changes made to `site.json`:
```
"baseUrl": "/"
```
```
  "deploy": {
    "message": "Site Update.",
    "repo": "https://github.com/MarkBind/markbind.github.io.git",
    "branch": "master"
  }
```

### DIFF
--- a/docs/site.json
+++ b/docs/site.json
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "/markbind.github.io",
+  "baseUrl": "/",
   "titlePrefix": "MarkBind",
   "pages": [
     {
--- a/docs/site.json
+++ b/docs/site.json
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "/",
+  "baseUrl": "",
   "titlePrefix": "MarkBind",
   "pages": [
     {
--- a/docs/site.json
+++ b/docs/site.json
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "/markbind",
+  "baseUrl": "/markbind.github.io",
   "titlePrefix": "MarkBind",
   "pages": [
     {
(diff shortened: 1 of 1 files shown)


damithc:
> Temporary because there's a URL conflict right now

Oops. What can we do about this? I don't think renaming markbind repo will help as GitHub remembers the old name as well. Rename the generated `markbind` folder to something else?

yamgent:
> Oops. What can we do about this? I don't think renaming markbind repo will help as GitHub remembers the old name as well. Rename the generated `markbind` folder to something else?

If I am not wrong, we should be able to just disable this repo's website when the time comes and it should work.

damithc:
> If I am not wrong, we should be able to just disable this repo's website when the time for actual deployment comes and it should work. **EDIT:** Right now the problem is having two active sites at the same time, which is why I reverted it after testing this PR out.

I see. That means we'll have to host dev docs at another repo (when we convert them to MarkBind) as the `markbind` site cannot have an active site.

PR #703, opened 2019-02-18
docs/*: rename variable main_title, main_file, etc.

damithc:

• [x] Documentation update

```
Some files use the variable `main_title` (instead of `title`)
to avoid being overwritten by a similarly named variable in a
reuse context.

As of v1.19.0, variables are no longer cascaded into inner files
by default. Hence, there is no need for such a precaution.

Let's rename those variables back to `title`.
```


damithc:
Hang on, found another variant of this problem. Will update soon.

damithc:
> Hang on, found another variant of this problem. Will update soon.

Updated. Also renamed `this_filename`/`main_filename` to `filename`

PR #707, opened 2019-02-18
Prune reference pages from search index

damithc:

• [x] Documentation update

```
Prune reference pages from the search index

The following reference pages (which contain collations of
information from other pages) appear in search results.
* Reader-Facing Features
* Full Syntax Reference
* Syntax Cheat Sheet

Let's remove these pages from the search index so that the search
can send the reader to the 'original' version of the content.
```

### DIFF
--- a/docs/site.json
+++ b/docs/site.json
@@ -14,6 +14,18 @@
     },
     {
       "glob": "userGuide/*.md"
+    },
+    {
+      "src": "userGuide/fullSyntaxReference.md",
+      "searchable": "no"
+    },
+    {
+      "src": "userGuide/syntaxCheatSheet.md",
+      "searchable": "no"
+    },
+    {
+      "src": "userGuide/readerFacingFeatures.md",
+      "searchable": "no"
     }
   ],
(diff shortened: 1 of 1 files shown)


docs/site.json
```diff
@@ -14,6 +14,18 @@
     },
     {
       "glob": "userGuide/*.md"
+    },
+    {
+      "src": "userGuide/fullSyntaxReference.md",
+      "searchable": "no"
```
yamgent: It prunes everything off though. So now this is not searchable at all, whereas previously it will just appear as two separate entries (which is not desirable but at least still searchable?)

damithc: It is not a big loss because all content appear elsewhere, which I consider to be their 'proper' location. It doesn't make sense to have two results for the exact same content. This page is not a good destination for search results due to all the dynamic panels (the user ends up at the wrong scroll position in the page due to how the page is loaded in stages). Unfortunately, this page appears at the top of the search results, above the more preferred twin result. Let's get rid of it for now. We can bring it back if we 'miss it' after a period of dog fooding.

PR #709, opened 2019-02-18
Remove website-content wrapper from pages

damithc:
• [x] Documentation update

```
Remove website-content wrapper from pages

All docs pages have a span named website-content wrapped around the main
body of content.

Nesting all content within a tag is an unnecessary complication.
The wrapper does not seem to add a significant value.

Let's remove the offending wrapper (and the corresponding css classes)
from everywhere in the site.
```

I'm not sure if this wrapper serves some useful purpose. Removing it didn't seem to break anything. We can keep it if it serves a useful purpose.


yamgent:
@damithc need you to resolve the conflict.

PR #713, opened 2019-02-19
Upgrade Font Awesome from 5.1.0 to 5.7.2

### LINKED ISSUE #510: {{ fab_markdown }} is not working
`{{ fab_markdown }}` doesn't seem to be working although other brand icons e.g., `{{ fab_github }}` are working.

Xenonym:

Fixes #510.

With #680, Font Awesome icon names no longer need to be manually hardcoded to be used when referenced with the new colon syntax. As such, we can straightforwardly upgrade our version of Font Awesome by just replacing the CSS and font files. We can also use the [`@fortawesome/fontawesome-free`](https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers#installing-free) NPM package to keep Font Awesome updated in the future.

- I installed `@fortawesome/fontawesome-free@5.7.2`.
- I moved the Glyphicon asset files into `asset/glyphicons/`, to avoid having `fonts/` (from Glyphicons) and `webfonts/` (from Font Awesome) in the same folder.
- I added `Site.js#copyFontAwesomeAsset` to copy Font Awesome files from the NPM package to `asset/fontawesome/`.
- I fixed the behavior of the `fs.copyAsync()` mock to match its actual behaviour (details in 8e438cb9bf932f8cb428d3925c54f9ac7428be3c)

```markdown
- FA new icons: :fab-markdown: :fas-biohazard:
```

**Testing instructions:**
1. Add the above example code to a MarkBind site. It should render the new FA icons:
New FA icons


acjh:
Why are glyphicons affected?

Xenonym:
> Why are glyphicons affected?

@acjh I shifted `fonts/` to `webfonts/` and edited the Glyphicon CSS files to match, since the new FA files use `webfonts/`. Since we will update FA but not Glyphicons in the future, changing the directory structure to match FA's should be more efficient.

acjh:
What's more efficient? Glyphicons and Font Awesome use separate files right?

Xenonym:
> What's more efficient? Glyphicons and Font Awesome use separate files right?

@acjh Right, efficient may not be the right choice of word. What I meant was, the new FA files use `webfonts/`, while Glyphicons uses `fonts/`. Since it would be confusing to have both `webfonts/` and `fonts/`, I decided to consolidate both font folders. I chose to change the Glyphicons CSS files to refer to `webfonts/` instead of changing the FA CSS files to use `fonts/` since we will be likely to do FA upgrades in the future, and changing the Glyphicons CSS files once would be more "efficient" for upgrading than to change the FA CSS files to refer to `fonts/` every time.

On the flip side, since Glyphicons are no longer in Bootstrap, it is unlikely that we would have to change out those files in the future as there is no longer a publicly available version of Glyphicons beyond what used to be in Bootstrap.

acjh:
Let's not change the original paths.
https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css

Also, shall we use npm to manage Font Awesome?
https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers

Xenonym:
> Let's not change the original paths.
👍 Might I suggest doing `assets/fontawesome/` and `assets/glyphicons`? I feel that having both `webfonts/` and `fonts/` in the same folder will lead to confusion in the future.

> Also, shall we use npm to manage Font Awesome?
This seems like a good idea, but we would need to change our asset logic to copy from both `assets/` and the NPM package.

yamgent:
> This seems like a good idea, but we would need to change our asset logic to copy from both `assets/` and the NPM package. Also, currently we bundle ~2MB of FA asset files with the package, but to use the NPM package would involve ~13MB, which is a not-insignificant size bump.

The size bump is only on the MarkBind application side right? When we build the website, we may not need the entire package, we can just copy their `webfonts/` folder?

Using an npm package may also make upgrading the icons more convenient in the future.

Xenonym:
> The size bump is only on the MarkBind application side right? When we build the website, we may not need the entire package, we can just copy their `webfonts/` folder?

Yes, that's right. The 13 MB is just an increase in size when a user installs MarkBind. However, built sites will still be the same size. 

> Using an npm package may also make upgrading the icons more convenient in the future.

Agreed. I was just wondering if the benefit of this outweighs the increase in install size, given that we don't use most of the Font Awesome package apart from the pre-built files.

yamgent:
> Agreed. However, I was wondering if the benefit of this outweighs the increase in install size? (given that we don't use most of the Font Awesome package apart from the pre-built files)

Hmm, I agree that it makes our app a bit more bloated. @Chng-Zhi-Xuan what do you think?

Chng-Zhi-Xuan:
I think going forward to update Font Awesome automatically is necessary, especially if it can reduce some human resource requirements to keep it updated.

MarkBind as a product is also operating under some assumptions of our users:
- The author is using it on a computer
- Has reasonable internet access

So the 13MB increase in install size is quite trivial given that a 1 minute 720 YouTube video is approximately 40MB already.

Xenonym:
I shall try to see how to upgrade Font Awesome with the NPM package then! 
Should I create another PR for that?

Chng-Zhi-Xuan:
> Should I create another PR for that?

Putting it in this PR is ok.

Xenonym:
This PR has been updated to use Font Awesome via the NPM package!

yamgent:
I will take a closer look when I get home at night, but generally I haven't spotted a big issue yet.

Meanwhile,

1. "Move glyphicon files to asset/glyphicons" commit should be wrapped at 72 characters. Also maybe clean the commit message a bit, by writing a standard paragraph rather than using a bullet point?

src/Site.js
```diff
@@ -813,6 +815,19 @@ Site.prototype.updateSiteData = function (filePaths) {
   this.writeSiteData();
 };
 
+/**
+ * Copies Font Awesome assets to the assets folder
+ */
+Site.prototype.copyFontAwesomeAsset = function () {
+  const faRootSrcPath = path.join(__dirname, '..', 'node_modules', '@fortawesome', 'fontawesome-free');
+  const faCssSrcPath = path.join(faRootSrcPath, 'css', 'all.min.css');
```
yamgent: The package itself also provides `font-awesome.min.css`, so is there a reason why we use `all.min.css` instead of `font-awesome.min.css`?

__mocks__/fs-extra-promise.js
```diff
@@ -75,6 +75,28 @@ function copyDirSync(src, dest) {
   }
 }
 
+/**
+ * Utility function to copy a directory to a destination recursively
+ * copyDirAsync needs to use fs.copyAsync as behaviour of copyAsync and copyFileSync differ
```
yamgent: This comment is unnecessary as outside of the commit, the context for this comment is missing.

Xenonym: Makes sense, will remove the comment then!

src/Site.js
Xenonym: `fontawesome.min.css` is meant to be used in combination with the other CSS files in `css/` (`brands.css`, `solid.css`, `regular.css`, `light.css`) [to enable only specific icon styles](https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#using-certain-styles). 

Since we are enabling all icon styles, we should use `all.min.css`, which [enables all icon styles in a single file](https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#using-web-fonts).

Xenonym:
@yamgent commit messages updated!

src/Site.js
yamgent: Hmm, is it misleading to call the final destination css to be `fontawesome.min.css` then? 🤔 Should we also change it to `all.min.css` for the final destination css?

yamgent:
> @yamgent commit messages updated!

Nice work. :+1:

Only one minor nit, for commit message in "Copy Font Awesome assets from NPM package", you can drop the word "Currently", it is implied.

For commit message in "Return promise directly from copyMarkbindAsset", we could have avoided using the uncommonly heard jargon, by directly explaining the problem with the anti-pattern (the original code is an exercise in futility), but the diff speaks for itself so I am open with leaving the message as it is now.

src/Site.js
Xenonym: I think the original reason for naming it `font-awesome.min.css` was to not have a `all.min.css` in the `css/` folder, since it would be confusing.

However, now that we have moved FA to `fontawesome/`, `fontawesome/css/all.min.css` would be less confusing. I will retain the original `all.min.css` naming then.

PR #714, opened 2019-02-20
Support always-on plugins

### LINKED ISSUE #702: Allow for 'always-on' plugins
Follow up from [474](https://github.com/MarkBind/markbind/pull/474#issuecomment-462358229), [495](https://github.com/MarkBind/markbind/pull/495#pullrequestreview-204547897)

Add a mechanism for 'always-run' plugins i.e. without having to include it in `plugins`, this will apply to plugins such as anchors, syntax shortening that the user should not need to turn off.

Allow disabling by accepting `"off"` in `"pluginsContext"`, similar to ESLint
https://github.com/MarkBind/markbind/blob/master/.eslintrc.js

jamos-tay:

Fixes #702 

Certain plugins are unlikely to be excluded from every project
Support for plugins that are always on by default for every project

Added support for default plugins:
- Reside in the `src/plugins/default` folder
- Folder is scanned and all plugins are included
- Overriding allowed: If there's a plugin with the same name in the project plugin folder, Markbind will use that one. A warning will be displayed if the default plugin is being overridden so they don't accidentally override
- Can be turned off by supplying `off` to pluginsContext:
```
{
  ...
  plugins : {
     // Do not need to specify plugin name in plugins
  },
  pluginsContext : {
    "anchors" : {
      "off": true
    }
  }
}
```

Also:
- Convert anchors to default plugins to test it

### DIFF
--- a/docs/userGuide/usingPlugins.md
+++ b/docs/userGuide/usingPlugins.md
@@ -146,6 +146,29 @@ This will add the following link and script elements to the page:
 - `<script src="SCRIPT_LINK"></script>`
 - `<script>alert("hello")</script>`
 
+### Advanced: Default plugins
+
+MarkBind has a set of default plugins that it uses to carry out some of its features. These are enabled by default for every project and should be left alone.
+
+Default Plugin | Functionality
+--- | ---
+`anchors` | Attaches anchor links to the side of headings.
+
+Although not advised, you can disable these by passing `"off": true` in the `pluginsContext`.
+
+Disabling the `anchors` plugin:
+
+```js
+{
+  ...
+  "pluginsContext": {
+    "anchors": {
+      "off": true
+    }
+  }
+}
+```
+
 ### Built-in plugins
 
 MarkBind has a set of built-in plugins that can be used immediately without installation.
--- a/src/Page.js
+++ b/src/Page.js
@@ -805,6 +805,19 @@ Page.prototype.generate = function (builtFiles) {
 };
 
 /**
+ * Retrieves page config for plugins
+ */
+Page.prototype.getPluginConfig = function () {
+  return {
+    headingIndexingLevel: this.headingIndexingLevel,
+    enableSearch: this.enableSearch,
+    searchable: this.searchable,
+    rootPath: this.rootPath,
+    sourcePath: this.sourcePath,
+  };
+};
+
+/**
  * Entry point for plugin pre-render
  */
 Page.prototype.preRender = function (content) {
@@ -812,7 +825,8 @@ Page.prototype.preRender = function (content) {
   Object.entries(this.plugins).forEach(([pluginName, plugin]) => {
     if (plugin.preRender) {
       preRenderedContent
-        = plugin.preRender(preRenderedContent, this.pluginsContext[pluginName] || {}, this.frontMatter);
+        = plugin.preRender(preRenderedContent, this.pluginsContext[pluginName] || {},
+                           this.frontMatter, this.getPluginConfig());
     }
   });
   return preRenderedContent;
@@ -826,7 +840,8 @@ Page.prototype.postRender = function (content) {
   Object.entries(this.plugins).forEach(([pluginName, plugin]) => {
     if (plugin.postRender) {
       postRenderedContent
-        = plugin.postRender(postRenderedContent, this.pluginsContext[pluginName] || {}, this.frontMatter);
+        = plugin.postRender(postRenderedContent, this.pluginsContext[pluginName] || {},
+                            this.frontMatter, this.getPluginConfig());
     }
   });
   return postRenderedContent;
--- a/src/Page.js
+++ b/src/Page.js
@@ -38,7 +38,6 @@ const PAGE_NAV_CONENT_WRAPPER_ID = 'page-nav-content-wrapper';
 const SITE_NAV_ID = 'site-nav';
 const TITLE_PREFIX_SEPARATOR = ' - ';
 
-const ANCHOR_HTML = '<a class="fa fa-anchor" href="#"></a>';
 const DROPDOWN_BUTTON_ICON_HTML = '<i class="dropdown-btn-icon">\n'
   + '<span class="glyphicon glyphicon-menu-down" aria-hidden="true"></span>\n'
   + '</i>';
@@ -458,29 +457,6 @@ Page.prototype.concatenateHeadingsAndKeywords = function () {
 };
 
 /**
- * Adds anchor links to headings in the page
- * @param content of the page
- */
-Page.prototype.addAnchors = function (content) {
-  const $ = cheerio.load(content, { xmlMode: false });
-  if (this.headingIndexingLevel > 0) {
-    const headingsSelector = generateHeadingSelector(this.headingIndexingLevel);
-    $(headingsSelector).each((i, heading) => {
-      $(heading).append(ANCHOR_HTML.replace('#', `#${$(heading).attr('id')}`));
-    });
-    $('panel[header]').each((i, panel) => {
-      const panelHeading = cheerio.load(md.render(panel.attribs.header), { xmlMode: false });
-      if (panelHeading(headingsSelector).length >= 1) {
-        const headingId = $(panelHeading(headingsSelector)[0]).attr('id');
-        const anchorIcon = ANCHOR_HTML.replace(/"/g, "'").replace('#', `#${headingId}`);
-        $(panel).attr('header', `${$(panel).attr('header')}${anchorIcon}`);
-      }
-    });
-  }
-  return $.html();
-};
-
-/**
  * Records the dynamic or static included files into this.includedFiles
  * @param dependencies array of maps of the external dependency and where it is included
  */
@@ -793,7 +769,6 @@ Page.prototype.generate = function (builtFiles) {
       .then(result => markbinder.resolveBaseUrl(result, fileConfig))
       .then(result => fs.outputFileAsync(this.tempPath, result))
       .then(() => markbinder.renderFile(this.tempPath, fileConfig))
-      .then(result => this.addAnchors(result))
       .then(result => this.postRender(result))
       .then(result => this.collectPluginsAssets(result))
       .then((result) => {
--- a/src/Site.js
+++ b/src/Site.js
@@ -11,7 +11,9 @@ const walkSync = require('walk-sync');
 
 const _ = {};
 _.difference = require('lodash/difference');
+_.get = require('lodash/get');
 _.has = require('lodash/has');
+_.includes = require('lodash/includes');
 _.isBoolean = require('lodash/isBoolean');
 _.isUndefined = require('lodash/isUndefined');
 _.noop = require('lodash/noop');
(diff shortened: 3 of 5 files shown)


src/plugins/default/anchors.js
```diff
@@ -0,0 +1,40 @@
...
+    headingsSelector += `, h${i}`;
+  }
+  return headingsSelector;
+}
+
+/**
+ * Adds anchor links to headers
+ */
+module.exports = {
+  postRender: (content, pluginContext, frontMatter, page) => {
```
jamos-tay: I needed access to `headingIndexingLevel,` so I added a 4th parameter, page which is the `Page` object itself... Other plugins might need access to page config so this could be helpful

I'm thinking we can use this for internal plugins, but don't document it, since we don't want to expose authors to the implementation

src/Site.js
```diff
@@ -666,32 +669,79 @@ function getPluginPath(rootPath, plugin) {
...
-      logger.warn(`Unable to load plugin ${plugin}, skipping`);
-    }
-  });
+  const defaultPlugins = findDefaultPlugins();
+
+  this.siteConfig.plugins
+    .filter(plugin => !_.includes(defaultPlugins, plugin))
+    .forEach(plugin => this.loadPlugin(plugin, false));
+  defaultPlugins
+    .filter(plugin => !this.siteConfig.pluginsContext
```
marvinchin: Just wondering - would using lodash's `get` method make this more readable? 🙂 

Possibly something like this:
```js
_.get(this.siteConfig, ["pluginsContext", plugin, "off"], false)
```

src/plugins/default/anchors.js
marvinchin: I'm just thinking it might be dangerous to pass the entire `Page` object down to plugins, since that would allow the plugin to modify or invoke the variables and methods of `Page` which could be dangerous. 😅 

What do you think about passing down a copy of the `pageConfig` object instead? Or maybe having a helper method `getDefaultPluginsContext` which constructs the context for default plugins and appends it to the other `pluginsContext`?

jamos-tay: Sure, I think I'll do the page config method, I think `getDefaultPluginsContext` might end up growing pretty large if we have a lot of default plugins =P

src/Site.js
jamos-tay: TIL, thanks

jamos-tay:
Rebased and updated

docs/userGuide/usingPlugins.md
```diff
@@ -115,4 +115,13 @@ MarkBind has a set of built-in plugins that can be used immediately without inst
...
 
+### Advanced: Default plugins
+
+Markbind has a set of default plugins that it uses to carry out some of its features. These are enabled by default for every project and should be left alone.
+
+Default Plugin | Functionality
+--- | ---
+`anchors` | Attaches anchor links to the side of headings.
+
+Although not advised, you can disable these by passing `"off": true` in the `pluginsContext`.
```
yamgent: Give a brief code snippet example, so that the user can have a better idea on how this is done (btw the code snippet given in your OP doesn't work because of missing/mismatch quotation marks).

src/plugins/default/anchors.js
yamgent: This part doesn't seem to be addressed?

src/Site.js
```diff
@@ -666,32 +670,77 @@ function getPluginPath(rootPath, plugin) {
...
   }
 
   return '';
 }
 
+/**
+ * Finds plugins in the site's default plugin folder
+ */
+function findDefaultPlugins() {
+  const globPath = path.join(__dirname, BUILT_IN_PLUGIN_DEFAULT_FOLDER_NAME);
```
acjh: This assumes all built-in plugins are default plugins (i.e. turned **on** by default), which is not true.
For example, Algolia search plugin loads unnecessary files in the site if not used, so it should be turned **off** by default.

Also, this is known at MarkBind dev time. Shall we use a constant as a whitelist?

jamos-tay: Currently there's a separate folder for plugins that are always on (which I'm calling 'default plugins')

```
src/
  plugins/
    default/
      always-on.js
      ... other default plugins
    must-be-turned-on.js
    ... other normal plugins
```

Plugins that should be off can be placed outside the default folder, where they are treated like normal plugins

I think this might be better than using a whitelist as that has to be hardcoded...

jamos-tay:
Updated and rebased

src/Site.js
acjh: Ah. `BUILT_IN_PLUGIN_DEFAULT_FOLDER_NAME` → `BUILT_IN_DEFAULT_PLUGIN_FOLDER_NAME`?

yamgent: > Ah. `BUILT_IN_PLUGIN_DEFAULT_FOLDER_NAME` → `BUILT_IN_DEFAULT_PLUGIN_FOLDER_NAME`?

Sounds better :+1:.

jamos-tay: Sure no problem

jamos-tay:
Updated, rebased

docs/userGuide/usingPlugins.md
```diff
@@ -152,5 +152,28 @@ MarkBind has a set of built-in plugins that can be used immediately without inst
 
 <include src="plugins/filterTags.mbdf" />
 
+### Advanced: Default plugins
+
+Markbind has a set of default plugins that it uses to carry out some of its features. These are enabled by default for every project and should be left alone.
```
yamgent: Markbind -> MarkBind

src/Site.js
```diff
@@ -666,32 +670,77 @@ function getPluginPath(rootPath, plugin) {
...
+ * Finds plugins in the site's default plugin folder
+ */
+function findDefaultPlugins() {
+  const globPath = path.join(__dirname, BUILT_IN_DEFAULT_PLUGIN_FOLDER_NAME);
+  if (!fs.existsSync(globPath)) {
+    return [];
+  }
+  return walkSync(globPath, {
+    directories: false,
+    globs: ['*.js'],
```
yamgent: A possible concern with this could be that a `.js` files happens to have a `preRender()` and/or `postRender()` method, but it is not an actual plugin, but just some library.

Should we have some way to allow .js files to "designate" themselves as a valid MarkBind plugin?

jamos-tay: Maybe we can use a different file extension, like .mbjs or something? Just for default built-in plugins, the other plugins can be .js since those are only loaded when specified

yamgent: Hmm, I am not sure whether it is a good idea to give a different file extension. They are still essentially normal JavaScript files, and that is not going to play well with editor configurations...

Perhaps introduce a suffix instead to mark such scripts? E.g. `<name>-mbplugin.js`.

acjh: I suggest using a prefix, similar to ESLint.

acjh: i.e. `markbind-plugin-<name>.js`

jamos-tay:
Updated

jamos-tay:
Sorry for the delay, rebased and updated

src/Site.js
```diff
@@ -667,32 +672,81 @@ function getPluginPath(rootPath, plugin) {
...
+function findDefaultPlugins() {
+  const globPath = path.join(__dirname, BUILT_IN_DEFAULT_PLUGIN_FOLDER_NAME);
+  if (!fs.existsSync(globPath)) {
+    return [];
+  }
+  return walkSync(globPath, {
+    directories: false,
+    globs: [`${MARKBIND_PLUGIN_PREFIX}*.js`],
+  })
+    .map(file => path.parse(file).name);
```
yamgent: Should be one line with the previous statement? Looks weird being a separate statement on its own.

src/Page.js
```diff
@@ -830,7 +806,8 @@ Page.prototype.preRender = function (content) {
   Object.entries(this.plugins).forEach(([pluginName, plugin]) => {
     if (plugin.preRender) {
       preRenderedContent
-        = plugin.preRender(preRenderedContent, this.pluginsContext[pluginName] || {}, this.frontMatter);
+        = plugin.preRender(preRenderedContent, this.pluginsContext[pluginName] || {},
+                           this.frontMatter, this.pageConfig);
     }
```
nicholaschuayunzhi: Passing `this.pageConfig` will allow plugins to modify the pageConfig object owned by the `Page`. While i understand `this.pageConfig` is only used in plugins, I'm thinking it would be better to construct a copy in the pre and post render functions explicitly.

Perhaps before the foreach:

```js
const config = {
  rootPath: this.rootPath,
  enableSearch: this.enableSearch,
};
```

This allows us to explicitly control what plugins have access to and prevents them from changing the object owned by `Page`. However, it does not prevent plugins from changing the config being passed to other plugins.

src/plugins/default/markbind-plugin-anchors.js
```diff
@@ -0,0 +1,40 @@
...
+    headingsSelector += `, h${i}`;
+  }
+  return headingsSelector;
+}
+
+/**
+ * Adds anchor links to headers
+ */
+module.exports = {
+  postRender: (content, pluginContext, frontMatter, pageConfig) => {
```
nicholaschuayunzhi: Would it be possible to write unit tests for this?

src/Page.js
yamgent: And then this line can be removed:
https://github.com/MarkBind/markbind/blob/7c8dfbb945b2f6c3f4a82c3caac7df6c4a6f3c9f/src/Page.js#L68

src/plugins/default/markbind-plugin-anchors.js
yamgent: To clarify: This is the part where we think could be done in a separate PR. Btw, this is different from the functional test that we already have in `test_site`.

src/Page.js
jamos-tay: What if we made a copy before passing it in?

src/plugins/default/markbind-plugin-anchors.js
jamos-tay: Sure, I'll make an issue

jamos-tay:
Sorry for lateness, updated

src/Page.js
```diff
@@ -827,10 +802,13 @@ Page.prototype.generate = function (builtFiles) {
  */
 Page.prototype.preRender = function (content) {
   let preRenderedContent = content;
+  const pageConfig = {
+  };
```
yamgent: The selection of config attributes to be given to plugins seems to be a bit too aribitrary... 

To be consistent with `postRender` below, maybe also include the `headingIndexingLevel`? Additionally, you can also include other search parameters such as `enableSearch` and `searchable`.

Finally, plugins may be interested in knowing the paths of the original source. `rootPath` and `sourcePath` can also be included.

These inclusions should also be made in `postRender`.

jamos-tay: Ahh I was thnking we could leave it blank first, then include whatever properties we may need when they show up

But sure, I can see how this is a little odd

jamos-tay:
Updated

jamos-tay:
Updated

yamgent:
Travis is currently borked, don't really know why, the builds are all getting stuck. I will return to this PR and merge it when I have the next free time.

Meanwhile the proposed commit message:

```
Support always-on plugins (#714)

When authors want to use plugins for his website, he or she must
modify site.json to enable them manually. There are no plugins that
are enabled by default.

In the future, there may be built-in MarkBind plugins that should be
enabled by default. For example, we want to move the anchor
functionality into a plugin. However, as there are no default plugins,
MarkBind will have them disabled by default, even though anchors are
a common feature in websites and it would be troublesome for authors
to enable this manually.

Let's add always-on plugins, plugins that are always enabled by default
unless the author specify in site.json to disable them. As a
"proof-of-concept", let's also refactor the anchor logic to an
always-on plugin to demonstrate how always-on plugins can be utilized.
```

Issue #610, opened 2019-01-20
Uncaught exceptions causes a silent exit


Xenonym:
**Tell us about your environment**

* **MarkBind Version:** 1.16.1

**What did you do? Please include the actual source code causing the issue.**
If an exception is thrown that is not handled, MarkBind logs the exception but does not return anything to standard output. For example, suppose we have the following MarkBind source:
```html
<!-- the following will cause an exception in Nunjucks -->
{% set error = " %}
```

**What did you expect to happen?**
MarkBind should print the exception to standard output / error, or at least indicate to the user that they should investigate the log file for more information.

**What actually happened? Please include the actual, raw output.**
`markbind build` aborts halfway with no feedback:
```
PS C:\Users\pzy5a\Desktop\testbind> markbind build
  __  __                  _      ____    _               _
 |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
 | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
 | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
 |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|

 v1.16.1
info: Website generation started at 18:13:12
info: Building assets...
info: Assets built
info: Generating pages...
[-] 0 / 1 pages built
PS C:\Users\pzy5a\Desktop\testbind>
```
However, the exception is logged:
```
2019-01-20T10:13:12.451Z - error: uncaughtException: (unknown path)
  expected block end in set statement date=Sun Jan 20 2019 18:13:12 GMT+0800 (Malay Peninsula Standard Time), pid=18160, uid=null, gid=null, cwd=C:\Users\pzy5a\Desktop\testbind, execPath=C:\Program Files\nodejs\node.exe, version=v8.15.0, argv=[C:\Program Files\nodejs\node.exe, C:\Program Files\nodejs\node_modules\markbind-cli\index.js, build], rss=63217664, heapTotal=60190720, heapUsed=26615736, external=836159, loadavg=[0, 0, 0], uptime=165999
Template render error: (unknown path)
  expected block end in set statement
    at Object.exports.prettifyError (C:\Users\pzy5a\Documents\GitHub\CS3281-2\markbind\node_modules\nunjucks\src\lib.js:34:15)
[...]

PR #721, opened 2019-02-21
Fix: Uncaught exceptions cause a silent exit

### LINKED ISSUE #265: Show error message when aborting due to nunjucks error
Context: I have a need to generate several variations of the same site. As markbind doesn't have syntax to support conditional variations yet, I've resorted to using nunjucks code (on top of markbind) for the time being. [example](https://github.com/nus-cs2103/website-base/edit/master/se-book-adapted/chapters/architecture.md)

Current: If there is an error in nunjucks code (e.g., importing a file that doesn't exist), the build fails silently. Same applies to live preview.

Suggested: instead of failing silently, show some error information from nunjucks

amad-person:

• [X] Bug fix

Fixes #265, #610.

Errors in building files with nunjucks (as well as uncaught exceptions) were causing a silent exit. This information should be shown to the user on the console. It was previously shown in the log file only. 

Added a new [Console transport](https://github.com/winstonjs/winston/blob/master/docs/transports.md#console-transport) to winston for logging errors to the console.

Also passed the path parameter to `nunjucks.renderString()` as errors thrown by Nunjucks were not showing the file paths.

Add the following statement to any file: 

```
<!-- the following will cause an exception in Nunjucks -->
{% set error = " %}
```

Console output on `markbind build`:

```
error: uncaughtException: (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/docs/about.md) [Line 14, Column 19]
  expected block end in set statement date=Thu Feb 21 2019 10:22:03 GMT+0800 (Singapore Standard Time), pid=32358, uid=501, gid=20, cwd=/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/docs, execPath=/usr/local/Cellar/node/11.10.0/bin/node, version=v11.10.0, argv=[/usr/local/Cellar/node/11.10.0/bin/node, /usr/local/bin/markbind, build], rss=85471232, heapTotal=64253952, heapUsed=37254224, external=2661243, loadavg=[2.04248046875, 2.189453125, 2.33935546875], uptime=310221
Template render error: (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/docs/about.md) [Line 14, Column 19]
  expected block end in set statement
    at Object.exports.prettifyError (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/src/lib/markbind/node_modules/nunjucks/src/lib.js:34:15)
    at new_cls.render (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/src/lib/markbind/node_modules/nunjucks/src/environment.js:472:27)
    at new_cls.renderString (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/src/lib/markbind/node_modules/nunjucks/src/environment.js:328:21)
    at Object.module.exports.renderString (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/src/lib/markbind/node_modules/nunjucks/index.js:80:14)
    at fs.readFile (/Users/aadyaamaddi/Desktop/BComp/Semester 6 1819/CS3281:2/markbind/src/lib/markbind/src/parser.js:508:36)
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:54:3)
```

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -276,7 +276,8 @@ Parser.prototype._preprocess = function (node, context, config) {
 
     // Render inner file content
     fileContent = nunjucks.renderString(fileContent,
-                                        { ...pageVariables, ...includeVariables, ...userDefinedVariables });
+                                        { ...pageVariables, ...includeVariables, ...userDefinedVariables },
+                                        { path: actualFilePath });
 
     // Delete variable attributes in include
     Object.keys(element.attribs).forEach((attribute) => {
@@ -504,7 +505,8 @@ Parser.prototype.includeFile = function (file, config) {
       const { parent, relative } = calculateNewBaseUrls(file, config.rootPath, config.baseUrlMap);
       const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
       const pageVariables = extractPageVariables(path.basename(file), data, userDefinedVariables, {});
-      const fileContent = nunjucks.renderString(data, { ...pageVariables, ...userDefinedVariables });
+      const fileContent = nunjucks.renderString(data, { ...pageVariables, ...userDefinedVariables },
+                                                { path: actualFilePath });
       const fileExt = utils.getExt(file);
       if (utils.isMarkdownFileExt(fileExt)) {
         context.source = 'md';
@@ -658,7 +660,7 @@ Parser.prototype._rebaseReference = function (node, foundBase) {
               // and let the hostBaseUrl value be injected later.
               hostBaseUrl: '{{hostBaseUrl}}',
               baseUrl: newBaseUrl,
-            });
+            }, { path: filePath });
             element.children = cheerio.parseHTML(rendered, true);
             cheerio.prototype.options.xmlMode = true;
           }
@@ -694,7 +696,7 @@ Parser.prototype._rebaseReferenceForStaticIncludes = function (pageData, element
 
   const newBase = fileBase.relative;
   const newBaseUrl = `{{hostBaseUrl}}/${newBase}`;
-  return nunjucks.renderString(pageData, { baseUrl: newBaseUrl });
+  return nunjucks.renderString(pageData, { baseUrl: newBaseUrl }, { path: filePath });
 };
 
 module.exports = Parser;
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -505,7 +505,8 @@ Parser.prototype.includeFile = function (file, config) {
       const { parent, relative } = calculateNewBaseUrls(file, config.rootPath, config.baseUrlMap);
       const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
       const pageVariables = extractPageVariables(path.basename(file), data, userDefinedVariables, {});
-      const fileContent = nunjucks.renderString(data, { ...pageVariables, ...userDefinedVariables },
+      const fileContent = nunjucks.renderString(data,
+                                                { ...pageVariables, ...userDefinedVariables },
                                                 { path: actualFilePath });
       const fileExt = utils.getExt(file);
       if (utils.isMarkdownFileExt(fileExt)) {
--- a/src/util/logger.js
+++ b/src/util/logger.js
@@ -9,9 +9,10 @@ winston.configure({
   exitOnError: false,
   transports: [
     new (winston.transports.Console)({
-      level: 'error',
+      colorize: true,
       handleExceptions: true,
       humanReadableUnhandledException: true,
+      level: 'error',
       showLevel: true,
     }),
     new DailyRotateFile({
--- a/src/util/logger.js
+++ b/src/util/logger.js
@@ -8,6 +8,12 @@ const DailyRotateFile = require('winston-daily-rotate-file');
 winston.configure({
   exitOnError: false,
   transports: [
+    new (winston.transports.Console)({
+      level: 'error',
+      handleExceptions: true,
+      humanReadableUnhandledException: true,
+      showLevel: true,
+    }),
     new DailyRotateFile({
       datePattern: 'YYYY-MM-DD',
       dirname: '_markbind/logs',
--- a/src/util/logger.js
+++ b/src/util/logger.js
@@ -12,7 +12,7 @@ winston.configure({
       colorize: true,
       handleExceptions: true,
       humanReadableUnhandledException: true,
-      level: 'error',
+      level: 'info',
       showLevel: true,
     }),
     new DailyRotateFile({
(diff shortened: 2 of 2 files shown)


src/lib/markbind/src/parser.js
```diff
@@ -504,7 +505,8 @@ Parser.prototype.includeFile = function (file, config) {
       const { parent, relative } = calculateNewBaseUrls(file, config.rootPath, config.baseUrlMap);
       const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
       const pageVariables = extractPageVariables(path.basename(file), data, userDefinedVariables, {});
-      const fileContent = nunjucks.renderString(data, { ...pageVariables, ...userDefinedVariables });
+      const fileContent = nunjucks.renderString(data, { ...pageVariables, ...userDefinedVariables },
+                                                { path: actualFilePath });
```
acjh: Move second argument to a line of its own, similar to above.

src/util/logger.js
```diff
@@ -8,6 +8,13 @@ const DailyRotateFile = require('winston-daily-rotate-file');
 winston.configure({
   exitOnError: false,
   transports: [
+    new (winston.transports.Console)({
+      colorize: true,
+      handleExceptions: true,
+      humanReadableUnhandledException: true,
+      level: 'error',
+      showLevel: true,
+    }),
```
acjh: Is this different from `console.error`? Should we remove that?

amad-person: Do you mean from this line:

https://github.com/MarkBind/markbind/blob/1686253401a4ad6ef75f67dfab0e7adc5c93543e/src/lib/markbind/src/parser.js#L82

acjh: No, this line:

https://github.com/MarkBind/markbind/blob/669325bcdefdb2b2d6ddb37b2ef6500e64555eac/src/util/logger.js#L33

amad-person: The current mechanism of logging is as follows: 

We are using `logger.js` functions to print messages to the console with `console.log()` statements and a [winston-daily-rotate-file transport](https://github.com/winstonjs/winston-daily-rotate-file) set to level `debug` that writes messages to a log file. 

Nunjucks errors were not being logged to the console because we are not calling `logger.error()` explicitly in `nunjucks.renderString()` (as we don't have error handlers for these calls). Other errors do get printed to the console because we call `logger.error()` when we catch them.

However, nunjucks errors were being printed to the log file because of the winston-daily-rotate-file transport. 

If we want to remove `console.error` (and other `console.log` statements in `logger.js`), then we should a winston console transport with the level set to `debug`. There is an option to use custom formats (color, etc.) too.

yamgent: Correct me if I am wrong, `nunjucks` is actually using `winston-daily-rotate-file` internally?

amad-person: No, `nunjucks` itself doesn't use `winston-daily-rotate-file`. `winston-daily-rotate-file` is a storage device for logs (i.e. a transport) where all our logging output goes. Currently MarkBind only uses this transport. To display logs on the console **with winston** we should have added a Console transport to our winston configuration.

The reason any log messages are being shown on the console right now is because `logger.js` explicitly outputs them using `console.log` statements.

amad-person: Hope this makes sense!

acjh: 1. Where do we configure winston for nunjucks?

2. > If we want to remove `console.error` (and other `console.log` statements in `logger.js`), then we should use the winston console transport with the level set to `debug`.

   Why is level set to `error` insufficient to remove `console.error`?

amad-person: > 1. Where do we configure winston for nunjucks?

If I'm not wrong, we are configuring winston only once in `logger.js` and using the logger instance in the rest of the application.

> 2. Why is level set to `error` insufficient to remove `console.error`?

For removing `console.error`, setting the level to `error` is sufficient. If we want to take it one step further and remove statements like `console.log(chalk.cyan('info: ') + text)` (i.e. for other levels too), `debug` would work.

acjh: 1. So how does nunjucks use winston?

2. As asked in my first comment: *Should we remove that?*

amad-person: > 1. So how does nunjucks use winston?

Our application as a whole uses winston, so all logging messages are handled by it. We don't need to configure nunjucks to use it, any message / uncaught exception from nunjucks will be logged automatically to _the log file_ (and only the log file, because that's the transport we have). 

> 2. As asked in my first comment: _Should we remove that?_

No we should not remove it, because we are calling `logger.error()` from other parts of the code. For example, our errorHandler uses it:

https://github.com/MarkBind/markbind/blob/6bbe1fa7d8a5eb02ff160050bc1fc088b9808f2e/src/Page.js#L766

acjh: 1. How does our application use winston if not through logger?

2. Logger calls winston.

amad-person: Our application does use winston through the logger, but we have also configured winston to handle logging uncaught exceptions. 
https://github.com/MarkBind/markbind/blob/6bbe1fa7d8a5eb02ff160050bc1fc088b9808f2e/src/util/logger.js#L15
That's how nunjucks errors were being logged to the log file.

acjh: 1. Thanks for checking this. It answers the question originally raised by @yamgent.

2. How about this?

amad-person: > 1. Let's set `winston.transports.Console` level to `'info'` and then remove `console.log` in `logger.error`, `logger.warn` and `logger.info`.

May I do these changes in this PR itself?

acjh: Yes please.

yamgent:
Console seems to be more noisy now though since we now print everything. :/

```
> markbind build
error: No config file found in parent directories of D:\GitHub\yamgent\markbind
error: uncaughtException: Path must be a string. Received undefined
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at Object.join (path.js:489:7)
    at Command.program.command.alias.option.description.action (D:\GitHub\yamgent\markbind\index.js:199:36)
    at Command.listener (D:\GitHub\yamgent\markbind\node_modules\commander\index.js:300:8)
    at emitTwo (events.js:126:13)
    at Command.emit (events.js:214:7)
    at Command.parseArgs (D:\GitHub\yamgent\markbind\node_modules\commander\index.js:635:12)
    at Command.parse (D:\GitHub\yamgent\markbind\node_modules\commander\index.js:457:21)
    at Object.<anonymous> (D:\GitHub\yamgent\markbind\index.js:210:9)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:191:16)
```

acjh:
Is that an actual error?

amad-person:
> Is that an actual error?

Yep, it happens when you run `markbind build` in a directory with no `site.json`.

yamgent:
> Is that an actual error?

Looks like I have accidentally discovered something worth fixing. :P Never mind my original comment then; I thought it was something that was previously printed on our own on the log.

PR #733, opened 2019-02-27
Support page top header

### LINKED ISSUE #549: Support jumbotron page header
Example: from the markdownguide

image

I wonder if we can provide support this kind of jumbotron to appear at the top of the page. The benefit is the site can be made to appear distinct simply by varying the background color. For project websites, the color can match the branding color of the project.

### LINKED ISSUE #636: Support a 'page top' container
Should we support a generic container for placing some content at the top of the page in a way that SiteNav and PageNav play along? i.e., SiteNav and PageNav should appear below this container. 

e.g., 
```html
<pagetop sticky flush>
... content
</pagetop>
```

Example user case: if the user wants to use Bootstrap navigation menus instead of our own `navbar`

### LINKED ISSUE #755: Inline styles within side navigation bars prevents overriding
* **MarkBind Version: v1.19.2**

**Issue**

Currently `site-nav` and `page-nav` uses [inline styles](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style) to format its list items. 

This prevents the author from overriding the style.

**Desired outcome**

Shift the inline styles to a class within `site-nav.css` and `page-nav.css`, which allows overriding.

Chng-Zhi-Xuan:

• [x] New feature

Resolves #636 
Resolves #549
Fixes #755 

- Changed MarkBind template to adopt the Holy Grail Layout, implemented using Flex Box (See No.1 in References).
- `site-nav` and `page-nav` are now in the same `flex-body-div` as the content, and are using `position: sticky` to have it follow the user as they are scrolling through content.
- Abstracted content wrapper handling from both `site-nav.css` and `page-nav.css` to the `#content-wrapper` ID within `markbind.css`.
- All authored sections (`site-nav`, `header`, `footer`) is appended to the page content so that we only call MarkBind parser once.
- The Page then collects the appended sections from the content, stored in a Page variable before mapping to the template variables.

**Removals (Justification in Discussion)**

- Removed `site-nav` button
- Removed support for inline footers
- Removed `scrollIntoView` 

**Implementation Discussion**

- In the new layout, expanding `site-nav` using the button (on smaller screens) will cause the content to shift, causing stuttering of the page. 
  - Solution 1, change to `position: fixed` instead of flex box when screen is small.<br> Maintain existing behaviour. However, when transitioning from `position: sticky` to `position: fixed` has stuttering and animation issues.
  - Solution 2, duplicate site navigation content to a dropdown list in the top nav<br>This solution becomes complex if we want to automate this, by inserting into an existing top nav or to render a top nav responsively if there isn't any.

- Removed `scrollIntoView` within `setup.js#pageNavSetup()`
  - Reason being that it breaks with `position: sticky` elements 

- Remove any inline `<header>` or `<footer>` elements
  - The main `index.md` file should contain content for the `<main>` part of the page, and should not overlap with `<header>` or `<footer>`
  - Encourage re-usability of headers and footers
  - Compatibility with `content-wrapper` for searching and indexing headings  

- Code style
- Edge cases that will break current layout
- Naming of variables / functions 

**To Do List** | Done ?
--- | --- |
Add initial template for header file | ✔️   |
Update `Site.test.js` | ✔️   |
Extended PR to fix #755 | ✔️ |
Allow scrolling of sticky navigation bars | ✔️ | 
Update documentation for `Footers` and `Headers` | ✔️ |

**References** |
--- |
[1. Holy Grail Layout with Flex Box](https://philipwalton.github.io/solved-by-flexbox/demos/holy-grail/)
[2. Holy Grail Layout](https://en.wikipedia.org/wiki/Holy_grail_(web_design)) |
[3. Complete guide to Flex Box](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) |
[4. Making Elements Stick in CSS](https://alligator.io/css/position-sticky/) |
[5. `overflow: hidden` breaks `position: sticky`](https://stackoverflow.com/questions/43909940/why-does-overflowhidden-prevent-positionsticky-from-working) |

**Testing instructions:**
- Download [holy-grail-demo-v2.zip](https://github.com/MarkBind/markbind/files/2960865/holy-grail-demo-v2.zip)
- Extract and run `markbind serve` in the directory
- Test new layout, header file support etc.


Chng-Zhi-Xuan:
CS3282 Team members are welcome to provide reviews 😄

acjh:
> **What changes did you make? (Give an overview)**
>
> * ...
> * ...
> * All authored sections (`site-nav`, `header`, `footer`) is appended to the page content so that we only call MarkBind parser once.

- Does this affect `[[toc]]` in page if there are headings in footer? (Should not.)
  I suppose this doesn't affect search indexing.

- Does this affect variables in footer that are undefined in variables.md? (Should not.)

- How about pre-parsing header and footer, so that they are only ever parsed once?

Chng-Zhi-Xuan:
> * Does this affect `[[toc]]` in page if there are headings in footer? (Should not.)
  I suppose this doesn't affect search indexing.

Headings are only collected within the content wrapper, which excludes page-nav, site-nav, header and footer sections.

> * Does this affect variables in footer that are undefined in variables.md? (Should not.)

I did not change anything other than removing inline footers, so existing behaviour (not affecting) should be preserved for file footers

> * How about pre-parsing header and footer, so that they are only ever parsed once?

This is a great idea, but would require us to pre-parse all `site-nav`, `footers` and `headings` as well as those contained within layout folders before generating the pages. 

As the changes required to `Site.js` seems extensive, should we push this to a separate issue?

src/Page.js
```diff
@@ -530,11 +517,38 @@ Page.prototype.removeFrontMatter = function (includedPage) {
...
+  return `${nunjucks.renderString(headerContent, userDefinedVariables)}\n${pageData}`;
+};
+
+
 /**
  * Inserts the footer specified in front matter to the end of the page
  * @param pageData a page with its front matter collected
  */
-Page.prototype.insertFooter = function (pageData) {
+Page.prototype.insertFooterFile = function (pageData) {
```
nicholaschuayunzhi: Was thinking if this can be refactored into a plugin? Similar to #714 ? Just a side thought though.

pyokagan:
@Chng-Zhi-Xuan Thanks so much for taking a look at this :+1: 

As an aside, seems like the site nav links in https://deploy-preview-733--markbind-master.netlify.com/userguide/ are broken.

damithc:
> @Chng-Zhi-Xuan Thanks so much for taking a look at this 👍

yup, thanks for tackling this major feature @Chng-Zhi-Xuan

src/Page.js
```diff
@@ -530,11 +517,38 @@ Page.prototype.removeFrontMatter = function (includedPage) {
...
 };
 
+/**
+  * Inserts the page layout's header to the start of the page
+  * @param pageData a page with its front matter collected
+  */
+Page.prototype.insertHeaderFile = function (pageData) {
+  const { header } = this.frontMatter;
+  let headerFile;
+  if (header) {
```
marvinchin: Just wondering - could `header` be an empty string`''` (this would resolve to `false`)? If so, should do a check for `header === undefined` here instead to be more explicit?

src/Page.js
```diff
@@ -758,6 +765,39 @@ Page.prototype.insertTemporaryStyles = function (pageData) {
...
+  const header = $('header');
+  if (header.length === 0) {
+    return;
+  }
+  this.headerHtml = htmlBeautify($.html('header'), { indent_size: 2 }).trim();
+  header.remove();
+  this.content = htmlBeautify($.html(), { indent_size: 2 });
+};
+
+Page.prototype.collectSiteNavigation = function () {
```
marvinchin: This method seems rather similar to `collectHeader`. Would it make sense to abstract this behaviour into another method?

marvinchin: Same applies for the `collectFooter` method.

test/functional/test_site/expected/sub_site/index.html
```diff
@@ -20,8 +20,11 @@
     <link rel="icon" href="/test_site/favicon.png">
 </head>
 <body>
-<div id="app">
-    <div id="content-wrapper">
+    <div id="app">
+
```
marvinchin: Just wondering if the empty lines here are intended? 😅

src/Page.js
Chng-Zhi-Xuan: Yes, it could be an empty string but the user would have to **explicitly** specify a `header` field in `<frontmatter>` and leave it empty. 

Also, empty string is a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), so it will still go to the `else` statement. 

Also the first statement is to check for any value specified within `<frontmatter>`, the logic of `header === undefined` is not correct (since it will be false when a valid string is given).

test/functional/test_site/expected/sub_site/index.html
Chng-Zhi-Xuan: Most likely due to the line break between template variables within `Page.ejs`, might look into it (since it is kind of ugly :P)

src/Page.js
marvinchin: Sorry, I think it should have been `header !== undefined` instead, but if the way we handle an empty header is the same as the way we handle the case where no header is defined, it might not be necessary.

Thanks for the clarification! 🙂

src/Page.js
Chng-Zhi-Xuan: Now that you mentioned it, we could abstract it to 

```Javascript
Page.prototype.collectPageSection = function (section) {
  const $ = cheerio.load(this.content, { xmlMode: false });
  const selectedSection = $(section)
  ....
  this.pageSections[section] = $.html(section);
}

Page.collectAllPageSection = function () {
  this.collectPageSection('header');
  this.collectPageSection(`#${SITE_NAV_ID}`);
  this.collectPageSection('footer');
}
```

Chng-Zhi-Xuan:
> @Chng-Zhi-Xuan Thanks so much for taking a look at this 👍
> 
> As an aside, seems like the site nav links in https://deploy-preview-733--markbind-master.netlify.com/userguide/ are broken.

Thanks for helping me find a bug!

Due to the positioning of `collectSiteNavigation`, the site navigation contents did not get parsed through  the critical line here:

```Javascript
this.content = nunjucks.renderString(this.content, { baseUrl, hostBaseUrl });
```

Likewise for the other collected sections.

src/Page.js
Chng-Zhi-Xuan: Will take a look if it is feasible. 

Although I am thinking that specifying plugins is less user intuitive compared to the current implementation of specifying via `<frontmatter>` or using layouts.

asset/js/setup.js
```diff
@@ -90,12 +90,6 @@ function setupSiteNav() {
   );
 }
 
-function setupPageNav() {
-  jQuery(window).on('activate.bs.scrollspy', (event, obj) => {
-    document.querySelectorAll(`a[href='${obj.relatedTarget}']`).item(0).scrollIntoView(false);
-  });
-}
```
yamgent: Are you planning any workarounds to restore this feature?

Chng-Zhi-Xuan: With the new layout, the sticky `page-nav` now has the height of the `<content>` instead of the viewport. So `scrollIntoView` would not work.

A different work around would need to be implemented, which is an independently scrollable sticky element. I found a good example here [Scrollable Sticky Element](http://leafo.net/sticky-kit/).

Chng-Zhi-Xuan: With the new layout, the sticky `page-nav` now has the height of the `<content>` instead of the viewport. So `scrollIntoView` would not work.

A different work around would need to be implemented, which is an independently scrollable sticky element.

**References**
[Scrollable Sticky Element](http://leafo.net/sticky-kit/)
[Sticky Sidebar](https://abouolia.github.io/sticky-sidebar/)

Chng-Zhi-Xuan:
**Update**

- Rebase to latest master
- Added meaningful commit messages
- Extended PR to fix #755 
- Fixed broken links within side navigation bar
- Added initialisation template for `header.md`
- Update side navigation bars to be scroll-able 

**Ready for review**

Would like reviewers to focus on the following:
- Code style / variable names
- Commit messages

Chng-Zhi-Xuan:
**Requesting for CS3282 Team Members Review**
@Xenonym @marvinchin @amad-person @sijie123 @luyangkenneth 

This is a large feature PR, so it is good exposure to look through and comment/review it :)

asset/css/page-nav.css
```diff
@@ -27,13 +25,22 @@
...
+.nested {
+    margin-left: 5%;
+}
+
+.no-flex-wrap {
+    flex-wrap: nowrap;
 }
 
-/* Responsive site navigation */
+
```
marvinchin: Is the number of blank lines here intended?

asset/css/markbind.css
```diff
@@ -78,12 +78,53 @@ code.hljs.inline {
     background-color: #fff;
 }
 
+/** Holy Grail Layout **/
```
marvinchin: Just wondering if the`/**` is intended instead of `/*`? I can't seem to find any other comments that use this convention in the codebase 😅

test/functional/test_site/expected/bugs/index.html
```diff
@@ -20,8 +20,11 @@
     <link rel="icon" href="/test_site/favicon.png">
 </head>
 <body>
-<div id="app">
-    <div id="content-wrapper">
+    <div id="app">
+
```
marvinchin: There seems to be a number of extraneous blank lines in the output html. Is this intended?

test/functional/test_site/expected/bugs/index.html
```diff
@@ -75,13 +81,14 @@ <h2 id="remove-extra-space-in-links">Remove extra space in links<a class="fa fa-
...
-<footer>
+            
+        </div>
+        <footer>
   <div class="text-center">
     Default footer
   </div>
 </footer>
-</div>
+    </div>
```
marvinchin: The indentation here seems off too

src/Page.js
```diff
@@ -150,7 +131,7 @@ function formatSiteNav(renderedSiteNav, src) {
     return renderedSiteNav;
   }
   // Tidy up the style of the unordered list <ul>
-  listItems.parent().attr('style', 'list-style-type: none; margin-left:-1em');
+  listItems.parent().addClass('site-nav-list');
```
marvinchin: Should we move `site-nav-list` to a constant? (similar to `page-nav`)

src/Page.js
```diff
@@ -590,31 +602,21 @@ Page.prototype.insertSiteNav = function (pageData) {
...
-  }
-  return pageData;
+  const wrappedSiteNav = `<nav id="${SITE_NAV_ID}" class="navbar navbar-light bg-transparent">\n`
+    + '<div class="sticky-top spacer-top viewport-height-90 scrollable slim-scroll">'
+    + `${siteNavDataSelector.html()}\n`
+    + '</div>\n'
+    + '</nav>\n';
+
+  return `${wrappedSiteNav}`
+    + `${pageData}`;
```
marvinchin: Would ``return `${wrappedSiteNav}${pageData}`;`` work instead? 🙂

asset/css/page-nav.css
Chng-Zhi-Xuan: Not intended, looks like `eslint` did not complain about this :P

asset/css/markbind.css
Chng-Zhi-Xuan: Because it was intended for `/** Holy Grail Layout **/` to be the main section and for `/* Header */`, `/* Footer */` etc to be sub-sections under the main section. 

I agree that it is subtle and not intuitive. Might change the commenting style to something more widely recognised.

test/functional/test_site/expected/bugs/index.html
Chng-Zhi-Xuan: I have not looked into this in depth but I might suspect it is due to `prepareTemplateData` defaulting to empty string values. 

I should test using the `<%_ if %>` syntax to avoid inserting empty strings and see if there are improvements.

test/functional/test_site/expected/bugs/index.html
Chng-Zhi-Xuan: Might be due to the indentation within `page.ejs`. Seems like a legacy issue since we have been using 4 space indentation within `page.ejs` but 2 space indentation for `htmlBeautify`.

src/Page.js
Chng-Zhi-Xuan: Yes, it would work. However, within `wrappedSiteNav` there is a line break at the end of the string. So I wrote it in this manner to visually represent that wrappedSiteNav and `pageData` is separated by one line break.

src/Page.js
Chng-Zhi-Xuan: Actually this brings up an interesting point.

Relating to Webstorm IDE (maybe other IDEs as well), where if you refactor CSS names, it will automatically change the name of that class within html strings or function parameters that take in CSS names. However, constant strings with the same name will not be refactored. 

Also, we shouldn't use a constant for a string if it is only used once throughout the whole file. 

So we should be finding single use constants and changing them back to be used inline? (`site-nav-list` only appears once)

asset/css/markbind.css
Chng-Zhi-Xuan: After digging around, this seems much better :)

```CSS
/**
  * Holy Grail Layout 
  *
  * This section covers the common styles used within Header, Footer and Side Navigation bars
  */

File: src/Page.js (hunk 10 above)
Comment (yamgent) [reply src/Page.js]: Hmm, I would say that there's still value in refactoring it even if it is a single use now, because in case in the future we need to use this somewhere else in the code, we need not do the refactoring again (which is also provided that we even realise that the same class is used here. If we forgot about it, then we are using "magic values" everywhere).

File: asset/css/page-nav.css (hunk 6 above)
Comment (acjh) [reply asset/css/page-nav.css]: ESLint doesn't check CSS files. :see_no_evil:

File: src/Page.js (hunk 1 above)
Comment (acjh) [reply src/Page.js]: We already pass `frontMatter` to plugins right?

File: asset/css/page-nav.css (hunk 6 above)
Comment (marvinchin) [reply asset/css/page-nav.css]: Is there any interest in adding a linter for css? (e.g. https://stylelint.io) 🙂

File: src/Page.js (hunk 11 above)
Comment (acjh) [reply src/Page.js]: We can move the `\n` out of `wrappedSiteNav` to here, so that the visual representation is consistent.

File: asset/css/page-nav.css (hunk 6 above)
Comment (yamgent) [reply asset/css/page-nav.css]: > Is there any interest in adding a linter for css? (e.g. https://stylelint.io) 🙂

@marvinchin PR is welcome for a css linter.

File: src/Page.js [hunk 12]
```diff
@@ -530,6 +532,33 @@ Page.prototype.removeFrontMatter = function (includedPage) {
...
+  const { header } = this.frontMatter;
+  let headerFile;
+  if (header) {
+    headerFile = path.join(HEADERS_FOLDER_PATH, header);
+  } else {
+    headerFile = path.join(LAYOUT_FOLDER_PATH, this.frontMatter.layout, LAYOUT_HEADER);
+  }
+  const headerPath = path.join(this.rootPath, headerFile);
+  if (!fs.existsSync(headerPath)) {
+    return pageData;
```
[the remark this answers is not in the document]
Chng-Zhi-Xuan: Since headers and footers are optional, an explicit error should not be logged. 

To add on, other optional items such as `site-nav` and `page-nav` are also not logged if they are not added. Furthermore, if we do log it then the console will be filled with messages if there are a significant amount of pages (with missing header files).

Although I can see explicit logging useful for debugging purposes. E.G.

```console
header file skipped: header file not found
page navigation added: {Title: Test Title} {headingIndexingLevel: 2}
footer file skipped: footer not specified, layout not specified
```
But alas, not in the scope of this PR :)

Chng-Zhi-Xuan:
@sijie123 , Thanks for looking through. To reply the CSS part: 

```CSS
.spacer-top {
    /*
      spacer-top class is a work around for MarkBind <navbar> component using "position: fixed".
      This class can be removed once the component's position type is changed.
    */
    padding-top: 70px;
}
```

Both `site-nav` and `page-nav` are using `.spacer-top` class for the reason stated in the comment 😄

test/functional/test_site/expected/bugs/index.html
Chng-Zhi-Xuan: **Update**

Tested with `<% if (...) %>` syntax but does not get rid of the blank lines.

However, adding either `-%>` or `_%>` to the end of the template variable insertion works. 

**Interesting finding:** 
The indentation before the template variables is only applied to the first line of the inserted html instead of all the lines within the inserted html. I would assume `page.ejs` is not smart enough to tell apart simple strings vs proper html :P

Chng-Zhi-Xuan:
**Update**

Review Amendments:
- Removed additional blank lines in `page-nav.css`
- Modified comments in `markbind.css`
- Added `_%>` to template variable insertions within `page.ejs` to fix blank lines in output html
- Refactored `site-nav-list` and `page-nav-nav` title to constants
- Moved last `\n` from `wrappedSiteNav` to return statement
 
Modifications:
- Rebased to latest master
- Changed `footers.mbdf` to reflect new behaviour
- Added `headers.mbdf` for new feature

Chng-Zhi-Xuan:
**Update**

- Rebase to latest master

Chng-Zhi-Xuan:
**Update**

- Added html formatting after output html is generated, before file write. Fixes indentation issues caused by new layout within `Page.ejs`.

src/Site.js
```diff
@@ -94,9 +95,19 @@ const FOOTER_DEFAULT = '<footer>\n'
   + '  </div>\n'
   + '</footer>\n';
 
+const HEADER_DEFAULT = '<header>\n'
+  + '  <div class="bg-primary display-4 text-center text-white">\n'
```
yamgent: This is subjective, but the default header generated seems to be too dominating compared to the rest of the parts of the webpage. :P

This is how it looks like now with a `markbind init` website:

test/functional/test_site/expected/sub_site/index.html
```diff
@@ -1,58 +1,64 @@
 <!DOCTYPE html>
 <html>
+
```
yamgent: The test site doesn't seem to have a page that uses our new headers.

Chng-Zhi-Xuan: I avoided placing a Jumbotron on the test site as the current top nav bar will block it. 

After a PR is done to modify `<navbar>` in the vue-strap repo, we require another PR here to update our usage of `<navbar>`. Within that PR, we can add testing header elements.

src/Site.js
Chng-Zhi-Xuan: Intended to be a large banner / Jumbotron. Similar to the one in [Markdown Guide](https://www.markdownguide.org/cheat-sheet/)

Chng-Zhi-Xuan:
**Update**

- Minor amendments for `footers.mbdf` and `headers.mbdf`

yamgent:
> * Create pull request in `vue-strap` repository to change any components using `position: fixed`. Specifically, `<navbar>` is the priority.

We need this to be merged as well as soon as possible, otherwise the empty gap rendering would look quite ugly.

Chng-Zhi-Xuan:
> > * Create pull request in `vue-strap` repository to change any components using `position: fixed`. Specifically, `<navbar>` is the priority.
> 
> We need this to be merged as well as soon as possible, otherwise the empty gap rendering would look quite ugly.

Will be working on it today 👍

damithc:
@Chng-Zhi-Xuan What if I want to have the same navbar in multiple pages but I also want a different jumbotron in each page e.g., https://www.markdownguide.org? Do I need to create many different header files and duplicate the navbar code in each (as include is not allowed in header files)?

damithc:
image

With respect to the above, why not allow the header to be cumulative, in the order they are specified?

Chng-Zhi-Xuan:
>  What if I want to have the same navbar in multiple pages but I also want a different jumbotron in each page? ...

Unfortunately yes, you have to have multiple header files if the content is different as there is no way to add content to the header file outside of authoring it.

Although supporting includes within `site-nav`, `footers` and `headers` can be done if we preprocess those files first before generating pages. 

This was mentioned by Aaron above, `"How about pre-parsing header and footer, so that they are only ever parsed once?"`.

> With respect to the above, why not allow the header to be cumulative, in the order they are specified?

I found a relevant discussion, ["Should header and footer tags be inside main tag?"](https://stackoverflow.com/questions/20470345/should-the-header-and-footer-tags-be-inside-the-main-tag).

Unless the inline `<header>` tag has content relevant to the `<main>` of the page, it should be placed outside of `<main>`. Since shifting `<header>` tags breaks the author layout, removing it ensures consistency with the pre-defined holy grail layout.

Consecutive header tags are supported in the current implementation, as long as it is within the same header file.

```html
<!-- within my-header.md -->
<header class"top-nav-stuff">
   ...
</header>
<header class="jumboBlue">
  ...
</header>
```

damithc:
What I'm trying to do is something like this:
```html
<frontmatter>
header: top-nav.md
</frontmatter>

<header class="jumboBlue">
  ...
</header>
```

That way, I can put the topnav in the header file (as it is common to all pages) and then add on a page specific-jumbotron below it in each page.

The documentation says _inline headers will be removed_. Do we even support inline headers? The docs doesn't say how to use an inline header.

Chng-Zhi-Xuan:
Sorry, missed this comment

> What I'm trying to do is something like this: ...

This is definitely possible. but can be confusing for the user. E.G

```HTML
<!-- Input by author -->
<frontmatter>
  header: shared-header.md
<frontmatter>

<header class=jumboBlue>
   Text 1
</header>

Some content 1

<header class="detailed-desc">
    Description text
</header>

Some content 2 
```

```HTML
<!-- Output by MarkBind -->
<header>
   sharedHeader
</header>
<header class="JumboBlue">
  Text 1
</header>
<header class="detailed-desc">
  Description text
</header>
<div id="content-wrapper">
  Some Content 1
  Some Content 2 
</div>
```
As we have to shift every header outside the `content-wrapper`, we might mess up what the author intended. 

Alternatives to this are less elegant than enforcing strict use of footer/header files:
- Add support for multiple inline footers and headers. However the inline headers and footer boundaries will be restricted to `<div id="content-wrapper">` and won't span the entire page if either `site-nav` or `page-nav` is specified.
- Have only top level headers (above content and below front matter) be shifted outside. A bit hacky, requires a lot of edge case checking / testing (headers in components, empty content, etc). 

> The documentation says inline headers will be removed. Do we even support inline headers? The docs doesn't say how to use an inline header.

Existing documentation specified use of Inline footers but was removed. It was supported prior to this PR. Specifically it was for inline footers. Where inline footers will override specified footers. This feature has been removed, and the new behaviour is ported over to headers as well, thus the notes being almost identical.  

I would agree that adding some context of what are inline headers and footers to the user guide would be better.

yamgent:
> I would agree that adding some context of what are inline headers and footers to the user guide would be better.

It is strange to explain a feature that no longer works though.

Chng-Zhi-Xuan:
> It is strange to explain a feature that no longer works though.

It is not really a feature, but something that can be created by the author. As inline headers and footers (within the `<body>` tag) can exist in normal HTML web pages.

damithc:
> It is not really a feature, but something that can be created by the author. As inline headers and footers (within the `<body>` tag) can exist in normal HTML web pages.

Oh? Didn't realize they are normal HTML tags. In that case shouldn't we have used some other tags e.g., `<md-header>` to avoid confusion?

Chng-Zhi-Xuan:
I would like to clarify what would be the purpose of a custom tag for headers? (For inline headers or for the tag used within `header.md`?)

Since functionally there is no special operations being performed on `<header>` tags other than insertion and collection (same as footers).

damithc:
> Since functionally there is no special operations being performed on `<header>` tags other than insertion and collection (same as `<footer>`).

If there is no difference, then it makes sense to use the same tag. 👍 
However, it feels like we ended up defining two types of headers: those that are put outside content-wrapper and those that will not be put outside the content-wrapper? 

> This is definitely possible. but can be confusing for the user. E.G
> 
> ```
> 
> <frontmatter>
>   header: shared-header.md
> <frontmatter>
> 
> <header class=jumboBlue>
>    Text 1
> </header>
> 
> Some content 1
> 
> <header class="detailed-desc">
>     Description text
> </header>
> 
> Some content 2 
> ```
> 
> ```
> 
> <header>
>    sharedHeader
> </header>
> <header class="JumboBlue">
>   Text 1
> </header>
> <header class="detailed-desc">
>   Description text
> </header>
> <div id="content-wrapper">
>   Some Content 1
>   Some Content 2 
> </div>
> ```
> 
> As we have to shift every header outside the `content-wrapper`, we might mess up what the author intended.

I can live with that, as it would solve my immediate problem. :-) 
In other words, we move all headers outside of content wrapper. 

Alternatively, we can provide some attribute (or a different tag) to differentiate between the two types of headers.

What do you think?

Chng-Zhi-Xuan:
A different tag would be appropriate (since we will be shifting the position of that tag)

Implementation should be simple, by using Cheerio to select and shift the headers with that specific tag to the top. It will be appended sequentially and keeping any classes authored on it.

PR #734, opened 2019-02-27
Add previous/next links to the user guide

damithc:
Proposed commit message (please use squash and merge):
```
Add previous/next links to the user guide

Let's add previous/next links to the bottom of the main pages of the
user guide so that readers can read the relevant pages in the correct
sequence if they want to.

These links are not added to reference pages as they are not meant
to be read in a linear sequence.

A nunjucks macro is used to minimize duplicating boilerplate code.
Using a boilerplate include is not feasible because the reused code
has conditional contents. For example, the 'next' link should not
appear in the last page of the sequence.

`<variable name='title' id='title'>` syntax is used to make the
title of a page accessible from other pages. The duplicate id attribute
can be removed after MarkBind has added support for accessing page
variables from outside of the page.
```

• [x] Documentation update

### DIFF
--- a/docs/njk/common.njk
+++ b/docs/njk/common.njk
@@ -1,0 +1,11 @@
+{% macro previous_next(previous_page, next_page) %}
+<div class="clearfix">
+{% if previous_page != ''%}
+<span class="float-left"><b-btn variant="light" href="{{ baseUrl }}/userGuide/{{ previous_page }}.html"><md>:far-arrow-alt-circle-left: <include src="{{ previous_page }}.md#title" inline /></md></b-btn></span>
+{% endif %}
+{% if next_page != ''%}
+<span class="float-right"><b-btn variant="light" href="{{ baseUrl }}/userGuide/{{ next_page }}.html"><md><include src="{{ next_page }}.md#title" inline /> :far-arrow-alt-circle-right:</md></b-btn></span>
+</div>
+{% endif %}
+<br>
+{% endmacro %}
--- a/docs/userGuide/addingPages.md
+++ b/docs/userGuide/addingPages.md
@@ -1,4 +1,4 @@
-<variable name="title">Adding Pages</variable>
+<variable name="title" id="title">Adding Pages</variable>
 <variable name="filename">addingPages</variable>
 
 <frontmatter>
@@ -46,3 +46,6 @@ Here are the steps to add a new page to your site:
 <modal large title="Live Preview" id="modal:addingPages-livePreview">
 <include src="glossary.md#live-preview" inline/>
 </modal>
+
+{% from "njk/common.njk" import previous_next %}
+{{ previous_next('authoringContents', 'markBindSyntaxOverview') }}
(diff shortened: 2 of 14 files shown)


Xenonym:
In commit message: `bottome of the main pages` -> **`bottom`**`of the main pages`

I am not sure the sequence makes sense for [Formatting Contents](https://markbind.org/userGuide/formattingContents.html) and [Using Components](https://markbind.org/userGuide/usingComponents.html) since these are essentially long lists of disparate building blocks, and most users are unlikely to use _all_ of them in their site. I don't know if anyone would realistically read these sections from start to finish at one go (at least I haven't 😅). Perhaps:
- we should put the previous/next links at the top instead, or
- move these pages to the reference sections since they do feel more like reference material. You don't need to use _all_ of them to use MarkBind proficiently.

damithc:
Thanks for the review @Xenonym 

> In commit message: `bottome of the main pages` -> **`bottom`**`of the main pages`

Updated the proposed commit message. @yamgent please use 'squash and merge' and use the proposed commit message above.

> I am not sure the sequence makes sense for [Formatting Contents](https://markbind.org/userGuide/formattingContents.html) and [Using Components](https://markbind.org/userGuide/usingComponents.html) since these are essentially long lists of disparate building blocks, and most users are unlikely to use _all_ of them in their site.

Good point. I don't think we expect people to actually _read_ all content. It is just the sequence for people to be _introduced_ to these pages, especially for a first-time reader. They could be just skimming through the content. In the case of the component page, I expect them to skim through the page to get to know which components exist but not read the details of how to use each one.

> * we should put the previous/next links at the top instead, or

We can put them on both top and bottom, although it felt a bit excessive.  Maybe just for long pages?

> * move these pages to the reference sections since they do feel more like reference material. You don't need to use _all_ of them to use MarkBind proficiently.

I'm trying to keep reference pages to contain content that are already found elsewhere in the "regular" (non-reference) pages. Basically, it is for returning readers who already know all the content and want to re-look at a specific topic in random order.

docs/njk/common.njk
```diff
@@ -0,0 +1,11 @@
...
+<div class="clearfix">
+{% if previous_page != ''%}
+<span class="float-left"><b-btn variant="light" href="{{ baseUrl }}/userGuide/{{ previous_page }}.html"><md>:far-arrow-alt-circle-left: <include src="{{ previous_page }}.md#title" inline /></md></b-btn></span>
+{% endif %}
+{% if next_page != ''%}
+<span class="float-right"><b-btn variant="light" href="{{ baseUrl }}/userGuide/{{ next_page }}.html"><md><include src="{{ next_page }}.md#title" inline /> :far-arrow-alt-circle-right:</md></b-btn></span>
+</div>
+{% endif %}
+<br>
+{% endmacro %}
```
yamgent: Insert an newline here.

damithc: Done

damithc:
Missing line break added.

PR #743, opened 2019-03-02
Standardize variable syntax

### LINKED ISSUE #710: Standardize the syntax for defining variables
Current:
Global variables use `<span id="foo">` syntax
Page variables use `<variable name="foo">` syntax

Should we allow `<variable name="foo">` syntax to be used for global variables too?

jamos-tay:

• [X] Other, please explain: Refactoring

Fixes #710 

Two different variable syntaxes are used, `<span id=...>` for include and global variables, `<variable name=...>` for page variables. For standardization we should only use one.

Change all variables to use `<variable name=...>` syntax.

**Testing instructions:**

I left compatibility for `<span id=...>` - prevents old sites from breaking. Should it be removed?

Can confirm something like this works:

```
<span id="a">1</span>
<variable name="b">2</variable>
```

### DIFF
--- a/docs/_markbind/variables.md
+++ b/docs/_markbind/variables.md
@@ -1,22 +1,22 @@
-<span id="showBaseUrlCode">
+<variable name="showBaseUrlCode">
 <code>{<span></span>{ baseUrl }}</code>
-</span>
+</variable>
 
-<span id="showBaseUrlText">{<span></span>{ baseUrl }}</span>
+<variable name="showBaseUrlText">{<span></span>{ baseUrl }}</variable>
 
-<span id="markbind_blue">#00B0F0</span>
+<variable name="markbind_blue">#00B0F0</variable>
 
-<span id="icon_arrow_down">:fas-arrow-down:</span>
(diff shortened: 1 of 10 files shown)


yamgent:
> I left compatibility for `<span id=...>` - prevents old sites from breaking. Should it be removed?

I think it should be removed eventually, don't see the value in the old syntax since it is hard to differentiate it from actual `<span>`s.

We will deprecate this feature first, then remove it by the next big release as per standard practice.

jamos-tay:
Rebased to solve merge conflict

yamgent:
Proposed commit message:

```
Standardize variable syntax (#743)

We use <span> to declare variables in include and global variables.

<span> variables are difficult to differentiate from actual <span>
tag uses. Furthermore, in places where this differentiation is
important (e.g. pages where <span>s are extremely common), we are
forced to use <variable> tags, rather than <span> tags. This results in
two different (and confusing) ways of declaring variables.

Let's standardize all variable declaration to use <variable> instead
of <span>.

The processing of <span> variables are currently left intact, but
they are deprecated and should be removed in a subsequent release.
```

Issue #582, opened 2019-01-15
Force cheerio to use our own custom htmlparser2


yamgent:
Occurs when using node v8.15.0 (which bundles npm v6.4.1).

When setting up the repo for the first time and using `npm install`, npm will install cheerio, one of our dependencies. cheerio in turn is dependent on htmlparser2.

Due to issues with the way [the original htmlparser2 parses the angular brackets](https://github.com/MarkBind/markbind/issues/101), we had to fork our [own custom version of htmlparser2](https://github.com/MarkBind/htmlparser2) in order to fix the issue.

When running `npm install` under npm v6.4.1, npm would not use our custom htmlparser2, but uses the one by the official version. This will cause an error for MarkBind v1.16.0 (provided by @Xenonym):

```
PS C:\Users\pzy5a\Documents\GitHub\markbind\docs> markbind build
  __  __                  _      ____    _               _
 |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
 | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
 | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
 |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|
 v1.16.0
info: Website generation started at 1:54:34 AM
info: Building assets...
info: Assets built
info: Generating pages...
[------------------] 0 / 18 pages builterror: TypeError: Cannot set property 'src' of null
warning: Error: Error while generating C:\Users\pzy5a\Documents\GitHub\markbind\docs\userGuide\makingTheSiteSearchable.md
error: Error: Empty src attribute in include in: C:\Users\pzy5a\Documents\GitHub\markbind\docs\userGuide\reusingContents.md
error: TypeError: Cannot set property 'src' of null
[================--] 16 / 18 pages builtwarning: Error: Error while generating C:\Users\pzy5a\Documents\GitHub\markbind\docs\userGuide\makingTheSiteSearchable.md
error: Error while generating C:\Users\pzy5a\Documents\GitHub\markbind\docs\userGuide\makingTheSiteSearchable.md
```

Solution: Use `npm shrinkwrap` to force cheerio to use our own version of htmlparser2.

Temporary workaround: If you face this issue, discard the changes made by npm for `package-lock.json`, and redo `npm install`.

----

Additional info: `package-lock-new.json` is the one whereby cheerio uses the official version of htmlparser2.

[...]

Issue #386, opened 2018-07-30
Allow custom colors for top nav bar


damithc:
v1.1.10

Current: there are only two choices for top nav bar color: black and grey

Suggestion: Allow custom colors

Justification: To create a lightweight theme effect on sites. Especially useful when instructors work with several variations of the same site and want an easy way to distinguish between them.

damithc:
> I feel that this one-off customisation should be done with custom styling in the site's own CSS file, rather than supported by MarkBind or MarkBind/vue-strap.

Fair point. The counter-point is that every website would want to brand itself. Although it is once-per-website, it might be used by most markbind sites.

If we leave it as DIY, we might still want to give guidance in documentation e.g., which class to override? Note that our target users are instructors. Until we have a WYSIWYG editor, we are forced to assume the target users have (or can learn) _some basic programming skills_ but we cannot assume _web programming expertise_. 

It also depend on the cost of the feature too. 

For the time being, for my own sites, I'm OK as long as I'm told how to override the color :-)

damithc:
Thanks @acjh Works after fixing typo `background-colour` → `background-color`
Is it possible to give different colors to main text and drop-down text?

yamgent:
Dropdown:

```css
.dropdown-item {
  color: brown !important
}
```

> main text

Main text as in the website page content itself?

damithc:
Thanks @yamgent 

> Main text as in the website page content itself?

No. I meant text in the main nav bar vs text in the dropdown items. Your solution is probably what I was looking for. Will try it out.

acjh:
If you don't want to affect other dropdowns:

```c#
.navbar .dropdown-item {
  color: brown !important
}
```

damithc:
I've tried using these on the website-base. While they work, it is quite a lot of css code to theme the whole navbar to a different color, including text and icons to match, not forgetting the hover over colors, boarders, shadows etc. I didn't manage to make it as good-looking as the default theme. I guess theming is harder than it appears at first :-p

acjh:
Indeed. But there is little value that MarkBind can add.

MarkBind should go for pleasant styling, but theming does not need further support in the codebase.
#363 (Support adding custom classes to any vue-strap component) was a step in the right direction.

Theming can be a topic to be considered when doing #371 (Give examples for adding custom classes).

damithc:
[...]

PR #745, opened 2019-03-02
Add support for light themes from bootswatch

### LINKED ISSUE #547: Support for bootswatch themes
What do you guys think of adding support to themes provided by https://bootswatch.com/ ? Does our current architecture even suitable for that kind of theming support?

### LINKED ISSUE #663: Minimal support for theming
- [ ] Provide a way to tweak a site to match a ‘brand’ color

luyangkenneth:

• [X] New feature

Fixes #547, Fixes #663

### High-level approach
The goal here is to enable themes in MarkBind! Building on the discussion in #547, the approach I'm going for is to:

1. Incorporate bootswatch into the project via its npm package, instead of the [other options](https://github.com/thomaspark/bootswatch/#usage). This is so that we can manage the dependency via npm, and be able to directly copy the css files without needing to download from a CDN.

1. During the asset-copying stage of site generation, use the custom `bootstrap.min.css` of the specified bootswatch theme instead of the default bootstrap file.

1. Allow users to change the theme of their site by specifying the name of the theme in `site.json`. At this point, this will be a global setting - no further granularity will be supported e.g. in the frontmatter or layout.

### Non-goals
- Adapt custom MarkBind components to use the overridden bootstrap css (to be done in a separate PR in the future)

### Things I would like your input on (edit: these are resolved, thanks everyone!)
**_I wanted this to get reviewed at its early draft stage_** so that we have a consensus about what I'll be working on, and how best to go about it. I would appreciate your feedback! Here are some open questions I have:

1. Is `npm install bootswatch` (no args) the right way to add it as a dependency? I feel uncomfortable seeing all the newly loosened version numbers in `package-lock.json`, and it wasn't immediately obvious to me how I can prevent that from happening.

1. Ideally we want every theme to look polished. But some of the things we're styling (e.g. code blocks, highlighting, colour of the GitHub logo) are fixed, and it might be super tedious to have to come up with custom or light/dark mode versions of everything. So I think there's some ambiguity regarding the scope of this feature, if we want to support _all_ bootswatch themes nicely. Do you have any thoughts on this?

1. Do you have any concerns about the approach / next steps / anything that I've mentioned so far?

### Example of what applying a diferent theme looks like

Screenshot 2019-03-14 at 5 07 42 PM


luyangkenneth:
Adding everyone as a reviewer to bring this to your attention. Feel free to involve yourself (or not) in the discussion 🙂

acjh:
> 1. Is `npm install bootswatch` (no args) the right way to add it as a dependency? I feel uncomfortable seeing all the newly loosened version numbers in `package-lock.json`, and it wasn't immediately obvious to me how I can prevent that from happening.

Yes. It is an issue with npm 6: https://github.com/npm/npm/issues/20434

You can prevent that from happening by (temporarily) downgrading to npm 5.
I suggest that developers use npm 5 for now, especially when modifying package-lock.json. @yamgent?

Related: https://github.com/MarkBind/markbind-cli/wiki/Developer-Guide#troubleshooting (#582)

> 2. Ideally we want every theme to look polished. But some of the things we're styling (e.g. code blocks, highlighting, colour of the GitHub logo) are fixed, and it might be super tedious to have to come up with custom or light/dark mode versions of everything. So I think there's some ambiguity regarding the scope of this feature, if we want to support _all_ bootswatch themes nicely. Do you have any thoughts on this?

If we want to claim theming support, then we should have light/dark versions of almost everything.
We don't want to (and cannot) support _all_ Bootswatch themes nicely.

From your screenshot:
- code blocks — look fine.
- highlighting — not shown.
- colour of the GitHub logo — cannot be changed: https://github.com/logos

damithc:
Shall we support only light themes at first? The need of the hour is for all MarkBind sites not to look exactly the same. That can be achieved with a range of light themes using different primary colors?

Xenonym:
> > 1. Is `npm install bootswatch` (no args) the right way to add it as a dependency? I feel uncomfortable seeing all the newly loosened version numbers in `package-lock.json`, and it wasn't immediately obvious to me how I can prevent that from happening.
> 
> Yes. It is an issue with npm 6: [npm/npm#20434](https://github.com/npm/npm/issues/20434)

@acjh @yamgent It's not so much a bug as it is [a change in how `package-lock.json` is written in `npm@6`](https://npm.community/t/issue-package-lock-json-saved-with-not-exact-package-versions/731/2). The official advice is to: 

> tl;dr: make sure your whole team is on `npm@6` or later, do a single `npm install` to upgrade your requires fields, and you should stop seeing weird diffs like these.

We could do the above in a separate PR.

acjh:
I didn't say it is a bug. It is an issue as the change is contentious.

> tl;dr: make sure your whole team is on `npm@6` or later, do a single `npm install` to upgrade your requires fields, and you should stop seeing weird diffs like these.

I disagree. But @yamgent can decide for MarkBind.

yamgent:
> I suggest that developers use npm 5 for now, especially when modifying package-lock.json. @yamgent?

Let's stick with npm 5 for now and not risk potentially breaking something.

luyangkenneth:
Thanks for your inputs everyone! We have resolved the npm issue separately in #767, and I've updated this PR description accordingly.

As discussed, we'll only support light themes for now. I'm ready to have the _code_ reviewed - please take a look 🙂 And then after everything is ok, I will proceed with writing the user docs.

docs/site.json
```diff
@@ -29,6 +29,7 @@
     }
   ],
   "headingIndexingLevel": 6,
+  "bootswatchTheme": "journal",
```
luyangkenneth: This is in a temporary commit, to illustrate what modifying the theme looks like from the user's perspective.

damithc: `"theme": "bootswatch-journal"` is more future proof, in case we add support for other themes?

src/Site.js
```diff
@@ -53,6 +54,25 @@ const LAYOUT_SCRIPTS_PATH = 'scripts.js';
...
+  'materia',
+  'minty',
+  'pulse',
+  'sandstone',
+  'simplex',
+  'sketchy',
+  'spacelab',
+  'united',
+  'yeti',
+];
```
luyangkenneth: These correspond to subfolder names within bootswatch's `node_modules` folder.

Spelling of the themes are double checked with https://bootswatch.com/api/4.json

docs/site.json
luyangkenneth: Nice, yeah that sounds better 👍

luyangkenneth: @damithc - Updated! How does this look now?

damithc: Looks good 👍

src/Site.js
```diff
@@ -830,7 +854,24 @@ Site.prototype.copyFontAwesomeAsset = function () {
  * Copies MarkBind assets to the assets folder
  */
 Site.prototype.copyMarkBindAsset = function () {
-  return fs.copyAsync(this.siteAssetsSrcPath, this.siteAssetsDestPath);
+  const maybeOverrideDefaultBootstrapTheme = () => {
+    const { theme } = this.siteConfig;
+    if (!theme || !Object.prototype.hasOwnProperty.call(SUPPORTED_THEMES_PATHS, theme)) {
```
Xenonym: Could this check be simplified to [`_.has(SUPPORTED_THEMES_PATHS, theme)`](https://lodash.com/docs/4.17.11#has)?

luyangkenneth: Good call, that definitely looks more readable. Done 👍

yamgent:
Apologies for the late comment, looks pretty good! Do update the docs in a separate commit, thanks! :+1:

luyangkenneth:
Docs are up! 🎉
- Commit: https://github.com/MarkBind/markbind/pull/745/commits/28a84fc2329e76b5a9da44af03a66918d598ab55
- Page: https://deploy-preview-745--markbind-master.netlify.com/userguide/themes

docs/userGuide/themes.md
```diff
@@ -0,0 +1,46 @@
...
+- [`bootswatch-materia`](https://bootswatch.com/materia/)
+- [`bootswatch-minty`](https://bootswatch.com/minty/)
+- [`bootswatch-pulse`](https://bootswatch.com/pulse/)
+- [`bootswatch-sandstone`](https://bootswatch.com/sandstone/)
+- [`bootswatch-simplex`](https://bootswatch.com/simplex/)
+- [`bootswatch-sketchy`](https://bootswatch.com/sketchy/)
+- [`bootswatch-spacelab`](https://bootswatch.com/spacelab/)
+- [`bootswatch-united`](https://bootswatch.com/united/)
+- [`bootswatch-yeti`](https://bootswatch.com/yeti/)
+
```
damithc: Should we use images from bootswatch here (with proper acknowledgegements)? e.g.,

docs/userGuide/themes.md
```diff
@@ -0,0 +1,46 @@
...
+{% from "njk/common.njk" import previous_next %}
+{{ previous_next('markBindInTheProjectWorkflow', '') }}
```
Xenonym: Missing a line break here:

Missing line break between previous link and footer

luyangkenneth: Nice catch - I see that this happens only at narrow viewport widths. This surprises me because, comparing with what the other docs look like, there's nothing that points to a mistake in the contents of the `.md` file.

Seems like this might be related to the link generation - I added a "next" link and the problem goes away. I'll file a new issue and investigate it separately, as it is a different concern from this PR.

docs/userGuide/themes.md
luyangkenneth: I was wondering about that too. The images have straightfoward URLs and should be easy to add to our docs. I think it might be reasonable to do that with proper acknowledgements, but I'm not fully sure about that.

damithc: Themes are very visual things; it's best if we can provide a visual comparison. Also, we should avoid sending our users to external sites as much as possible.

Alternatively, we can duplicate the files in our site (but acknowledge the source). I think they use MIT license which allows duplication.

docs/userGuide/themes.md
luyangkenneth: Documented this bug in #786

docs/userGuide/themes.md
luyangkenneth: Makes sense, I'll go ahead with that then!

luyangkenneth:
@damithc I've added the images for easier visual comparison between the themes. How does this look now?

Screenshot 2019-03-20 at 7 43 44 PM

damithc:
> @damithc I've added the images for easier visual comparison between the themes. How does this look now?

Looks good, although the pics blends too well with the page due to white background in both. Perhaps put the whole lot inside a `<box>` so that the gray background can define the edge of each pic? Just a thought. I'm OK with the current one too.

luyangkenneth:
Nice, yeah I agree it looks better:

Screenshot 2019-03-20 at 8 14 41 PM

luyangkenneth:
@yamgent How about the PR title? :)

yamgent:
> @yamgent How about the PR title? :)

The PR title alone is insufficient in providing the context of the change.

Proposed merge commit message:

```
Add built-in support for light themes from bootswatch (#745)

To use any custom Bootstrap themes, authors must manually copy over the
theme's .css files to the website's asset folders, and configure the
layout's head.md to use the .css files.

Let's add built-in support for bootswatch light themes, by providing a
"theme" option in site.json, so that authors can just directly specify
the name of the theme in order to use it. For example:

    {        
        "theme": "bootswatch-cerulean"
    }
```

damithc:
Commit message: you can also mention why only light themes are being added.

luyangkenneth:
Whoops I thought you meant just the title of the merge commit. What you proposed sounds good to me!

Re: light themes, maybe we can append this to what you wrote:

```
Dark themes from Bootswatch are not yet supported because they require
our custom MarkBind components to inherit the Bootstrap styling classes,
to be done in #782.
```

damithc:
I'm testing out this feature now. Is it possible to get the navbar to change color too? 
Like this:
image

luyangkenneth:
@damithc I've investigated this briefly and I think this involves modifying our Navbar custom component to use the `bg-primary` class, in place of the `bg-dark` class it currently uses.

For example, if we apply the Journal theme (https://bootswatch.com/journal/), the generated website will have a black navbar (the second one in the list) as opposed to the pink navbar that we see in the thumbnail.

---

If we change the bootstrap class, the default navbar color will also change, based on the default bootstrap theme.

### Using `bg-dark` (current)
Screenshot 2019-03-25 at 4 37 45 PM

### Using `bg-primary`
Screenshot 2019-03-25 at 4 37 31 PM

But we can probably work around this by modifying how we specify our `<navbar>` in the user guide's `header.md`.

Would you like us to move in this direction? / What are your thoughts?

damithc:
> Would you like us to move in this direction? / What are your thoughts?

It would be good to take advantage of the different nav bar colors as the navbar plays an important role in making the site look 'different'. It depends on how much effort it will take though. May be senior devs have some thoughts?

acjh:
Let's continue this discussion in #386.

PR #748, opened 2019-03-03
Site.js: remove site generation profiling

### LINKED ISSUE #738: Fix missing space in log output
Current:
image

A space is missing? This can be phrased in a more readable way? e.g., `1234 ms`

tawAsh1:
• [X] Bug fix

Fixes #738 

Get rid of the logger.profile()
use Date() and show generate time on debug level

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -528,11 +528,6 @@ Site.prototype.generate = function (baseUrl) {
       .then(resolve)
       .catch((error) => {
         rejectHandler(reject, error, [this.tempPath, this.outputPath]);
-      })
-      .finally(() => {
-        const endTime = new Date();
-        const totalBuildTime = endTime - startTime;
-        logger.debug(`Total generation time: ${totalBuildTime}ms`);
       });
   });
 };
--- a/src/Site.js
+++ b/src/Site.js
@@ -113,7 +113,6 @@ const USER_VARIABLES_DEFAULT = '<span id="example">\n'
   + 'More generally, surround the segment\'s id with double curly braces.\n'
   + '</span>';
 
-const GENERATE_SITE_LOGGING_KEY = 'Generate Site';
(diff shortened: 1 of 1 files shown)


src/Site.js
```diff
@@ -531,7 +529,11 @@ Site.prototype.generate = function (baseUrl) {
       .catch((error) => {
         rejectHandler(reject, error, [this.tempPath, this.outputPath]);
       })
-      .finally(() => logger.profile(GENERATE_SITE_LOGGING_KEY));
+      .finally(() => {
+        const endTime = new Date();
+        const totalBuildTime = endTime - startTime;
+        logger.debug(`Total generation time: ${totalBuildTime}ms`);
```
acjh: Let's:
1. rename this variable to `totalGenerationTime`.
2. use seconds, to be consistent with the existing `totalBuildTime`.
3. remove the `profile` export in logger.js.

yamgent: Should we downgrade the profile logging in `logger.js#profile()` instead, like so:

```js
logger.profile('test', { level: 'debug' });
```

Otherwise, if we are not planning to do any extensive profiling in the future, there's not really a reason for this portion of the code to stay.

tawAsh1: It seems necessary to upgrade the version of logger (winston) from 2 to 3 in order to execute that code.  ` logger.profile('test', { level: 'debug' }); `
https://github.com/winstonjs/winston/blob/master/UPGRADE-3.0.md

yamgent: Let's just get rid of this part of the code then.

yamgent:
Your first post and PR title will need to be updated, the PR title can be `Site.js: remove site generation profiling`?

tawAsh1:
Title updated

tawAsh1:
Thank you for your reviews. 🙇

PR #751, opened 2019-03-05
Add imported variables

### LINKED ISSUE #682: Allow accessing page variables from outside
e.g., allow `<include src="foo.md#title"/>` (where `title` is a page variable in `foo.md`)

Some dicussion about this feature can be found [here](https://github.com/MarkBind/markbind/pull/627#issuecomment-460223203)

jamos-tay:

Fixes #682 

We want to be able to import page variables from inner files.

Page variables can be imported from inner files.

```
index.md
<include src="foo.md#title"/>
{{ title }}

foo.md
<variable name="title">Title</variable>
```

This PR changes the way variables are handled, hence WIP because there might be some side effects.
Variables are now only rendered after all preprocessing is done per file, this is necessary because we need to render `include` to retrieve the variable value before the variable can be rendered on the top level.

To do: Document and test integrations with other variable types


damithc:
A quick comment: this is a bit more restrictive than what I had in mind.  From `parent.md` I should be able to access `child1.md#title` as well as `child2.md#title`. i.e., _access_ page variables, not _import_ page variables.

jamos-tay:
@damithc Ahh, I see

How would it be referenced though? I don't think `.`s are allowed in nunjucks variable names. Something like `{{ child1#title }}` and `{{ child2#title }}`?

acjh:
They are used at the point of `<include>` as usual.

damithc:
> How would it be referenced though? I don't think `.`s are allowed in nunjucks variable names. Something like `{{ child1#title }}` and `{{ child2#title }}`?

Minimally we can just use `include` syntax?

Assuming `child.md` has variables `title` and `src`
```
# <include src="foo.md#title" inline />

<variable name="child_src"><include src="foo.md#src" inline /></variable>

<panel src="{{ child_src }}"></panel>
```

damithc:
As discussed at the meeting today, in light of implementation complications of the requested feature, perhaps something like this (closer to the current implementation)?
```html
<import src="foo.md#title"/>
{{ title }}

<import src="bar.md#title" as="bar_title" />
{{ bar_title }}
```

damithc:
@jamos-tay can revive this PR?

jamos-tay:
I rebased with newer codebase. I'm getting a weird error with test that I can't seem to fix:

image

I think changing the order pages are rendered messed it up somewhat. Will try looking into it later

jamos-tay:
Okay, I'm stumped... can someone guide me as to how `{{baseUrl}}` gets converted? For some reason the `{{hostBaseUrl}}` for sub_site is being resolved to `test_site` and I'm not too sure why that's is happening

acjh:
What do you expect it to be resolved to?

jamos-tay:
If I'm not wrong, just `/`

acjh:
`test_site/site.json` specifies `"baseUrl": "/test_site"`.

sijie123:
@jamos-tay could you pull the newest version (as of 18 Apr) and test again?
Previously, the implementation of the rewrite might be a little haphazard. The implementation now should work correctly (fingers crossed).
The test errors come from the reuse section. I was working on a feature that rewrites relative urls (eg "../images/photo.png") into its absolute path ("/docs/userGuide/images/photo.png" for example). I'm not sure how this will affect your feature.

jamos-tay:
@acjh Even for the sub_site? I'm not exactly sure how sub sites work with MarkBind

@sijie123 Ahh I see... I tried rebasing off master, it fixed one of the diffs but this one still isn't working:

index.md
```
# Include segment from another Markbind site
<include src="sub_site/testReuse.md#imageTest" />
```

The one above (`<include src="sub_site/testReuse.md" />`)  seems to be okay though? Which is odd...

yamgent:
> Even for the sub_site? I'm not exactly sure how sub sites work with MarkBind

Yes, for the sub_site it should still be `/test_site`, because we are not building the sub_site directly, but instead including the files from test_site.

> Ahh I see... I tried rebasing off master, it fixed one of the diffs but this one still isn't working:
> 
> index.md
> 
> ```
> # Include segment from another Markbind site
> <include src="sub_site/testReuse.md#imageTest" />
> ```
> The one above (`<include src="sub_site/testReuse.md" />`) seems to be okay though? Which is odd...

Maybe it is not resolving the relative path properly with the segment included? Does your variable import resolve the path correctly?

acjh:
>> Even for the sub_site? I'm not exactly sure how sub sites work with MarkBind
>
> Yes, for the sub_site it should still be `/test_site`, because we are not building the sub_site directly, but instead including the files from test_site.

👍

In sub_site, `{{ baseUrl }}` → `{{ hostBaseUrl }}/sub_site` → `/test_site/sub_site`.

jamos-tay:
Okay, I think I've finally found the problem

Since the rendering order was changed, hostBaseUrl was being rendered to nothing too early so I had to update it with `hostBaseUrl: '{{hostBaseUrl}}'` to prevent it from being rendered.

damithc:
> Okay, I think I've finally found the problem

Nice 👍

test/functional/test_site/expected/index.html
```diff
@@ -183,6 +183,16 @@ <h1 id="page-variable-referencing-included-variable">Page Variable Referencing I
             <div></div>
             Page Variable Referencing Included Variable
           </div>
+          <div>
+            <h1 id="imported-variable">Imported Variable<a class="fa fa-anchor" href="#imported-variable"></a></h1>
+            <div></div>
```
yamgent: The normal text "Imported Variable" did not appear here so this test case failed?

jamos-tay: Ah my bad, I think it might've disappeared when I was moving stuff around

jamos-tay:
Updated

src/lib/markbind/src/parser.js
```diff
@@ -296,8 +327,28 @@ Parser.prototype._preprocess = function (node, context, config) {
...
+        }
+        if (!context.importedVariables) {
+          // eslint-disable-next-line no-param-reassign
+          context.importedVariables = {};
+        }
+        if (!context.importedVariables[includeSrc.pathname]) {
+          // eslint-disable-next-line no-param-reassign
+          context.importedVariables[includeSrc.pathname] = [];
+        }
+        // eslint-disable-next-line no-param-reassign
```
yamgent: No need to disable eslint rule for this one.

yamgent: Nevermind, we can let someone else clean this up as part of first-timer issue, let's merge it if there's no major issues with the code.

jamos-tay: Ahh sure

ang-zeyu:
@damithc this and page variables are proving to be rather problematic and restrictive in terms of the overall architecture

seeking some understanding about what these provide on top of nunjuck's `{% import / set %}`

**Problem:**

Our processing order is like such currently:

html (to extract variables + imported variables) -> process with nunjucks -> html (to process `<include>`) -> markdown -> html (a lot of miscellaneous operations like processing relative links)

The issue is **parsing html before markdown**, this:
1. requires html - markdown patches for htmlparser2, which is very difficult to implement correctly (e.g. #1205)
2. prevents features like #896 with wild markdown syntax from being implemented
3. repeated parsing (performance)

ideally we should be always rendering markdown before html, I'd like to get the process down to a simple 3 steps:
nunjucks -> markdown -> html

However, the 3rd and 5th steps in the current processing order are mergable, but the first isn't because we need variables to process with nunjucks, making this impossible.

To my knowledge, the current page / import variable system provides only the **following on top of nunjucks**:
- external json variables (can be reimplemented using a nunjucks tag easily)
- ❔

damithc:
@ang-zeyu I actually don't have any knowledge of the internal workings. I can only comment on the external behavior.

So, if you need my input, you'll have to 'dumb it down' to the external behavior aspect only :-p

Keep in mind that we are in a crowded market and we aim to go the extra mile to optimize for a niche market. As a result, we sometimes end up adding esoteric features that might exact a rather heavy price on the elegance of the architecture. That doesn't mean we always choose in favor of the feature over architecture elegance, but sometimes we have to.

In this case, what are our choices, from external behavior point of view?

ang-zeyu:
> So, if you need my input, you'll have to 'dumb it down' to the external behavior aspect only :-p

This is what I need actually, sorry for the extra implementation details, its only to reason for the removal of the features

From external behaviour pov, it boils down to
1. Remove the `<variable>` (not including `<variable>` within `<include>`) and `<import>` feature, which nunjucks already has
   - massive code architecture illustrated above
   - the only feature we have currently not supported is external json variables; we can adapt the code to a nunjucks tag instead (e.g. `{% set-external %}`)
2. Retain it
   - feature redundancy
   - one more syntax to remember

would be great if any of the older devs @acjh @yamgent @marvinchin have more context on why these features was implemented as well

damithc:
I'm the one who asked for this feature. At that time there was a situation where I needed this feature and there was no other way to do it. I can't remember if we actually considered 'Why not use nunjucks instead?' at the time. Anyway, back then we also had a serious limitation with nunjucks: nunjucks could not be used inside sub-sites because nunjucks imports need to be specified relative to the the project root folder.  The imports in the sub-site will not work when it is part of a bigger site. Have we solved that problem already?

ang-zeyu:
I realised live preview also currently dosen't include nunjucks sources yet. But there is a convenient api in nunjucks for that, it should be rather straightforward to implement.

---

Dug up some more old threads, if I read these right,

#674 - `<variable>`s within `<include><variable>...</variable></include>` - no change, we keep this
https://github.com/MarkBind/markbind/pull/553#issuecomment-451707842 - page variables suggestion, #627 - page variable implementation pr

the common theme amongst all these is the ability to "override inner from outer", which is supported by `<include><variable>...</variable></include>`.
page `<variable>`s were created hence to allow this, because nunjucks' `{% set %}` indiscriminately overrides everything.
similarly, `<import>`s don't override `<include><variable>...</variable></include>`, while `{% import %}` always does

there is actually a hacky workaround this https://stackoverflow.com/questions/56277938/not-able-overriding-variable-value-from-end-file-using-nunjucks, but we shouldn't have to resort to this lengthy syntax


> The imports in the sub-site will not work when it is part of a bigger site. Have we solved that problem already?

I couldn't find any mention for this in the threads, I suppose it was implicitly noted by everyone, but its fixed anyway (#931) 

---


propose:
- add deprecation warnings
- per @jamos-tay's suggestion here https://github.com/MarkBind/markbind/pull/553#issuecomment-455474517, instead move the html like `<variable>` syntax back into `{% set %}`, and `<import>` into `{% import %}`
   - we claim that we provide nunjucks support, and this does not change. what it does is only allows `<include><variable>...</variable></include>` to be used properly (which is our custom syntax)
- add live preview support for nunjucks dependencies
- then going with deprecation of this after some time; the architectural benefits are simply too big to ignore

damithc:
> I couldn't find any mention for this in the threads, I suppose it was implicitly noted by everyone, but its fixed anyway (#931)

Yes, looks like fixed. Haven't tried it yet though. Will do soon.

> the common theme amongst all these is the ability to "override inner from outer", which is supported by `<include><variable>...</variable></include>`.

This is something I'm not willing to give up. It's the holy grail of reuse and MarkBind aims to facilitate a high level of reuse. Authors should be able to reuse a piece of content, read variables of it from outside, and also override those attributes from outside to make the content fit the reuse context. As long as we don't violate that principle, I don't mind removing the custom import feature in favor of nunjucks' import syntax even if the syntax is slightly longer (as long as not super long).

ang-zeyu:
> This is something I'm not willing to give up. It's the holy grail of reuse and MarkBind aims to facilitate a high level of reuse. Authors should be able to reuse a piece of content, read variables of it from outside, and also override those variables from outside to make the content fit the reuse context. As long as we don't violate that principle, I don't mind removing the custom import feature in favor of nunjucks' import syntax even if the syntax is slightly longer (as long as not super long).

no change here, my proposal only makes `<include><variable>...</variable></include>` override `{% set / import %}`

I realised the caveat here is that existing behaviours relying on the reverse (`{% set %}` overriding our custom syntax `<include><variable>...</variable></include>`) would be broken.
is there any value in supporting this?

damithc:
> no change here, my proposal only makes `<include><variable>...</variable></include>` override `{% set / import %}`

Good to hear that.

> I realised the caveat here is that existing behaviours relying on the reverse (`{% set %}` overriding our custom syntax `<include><variable>...</variable></include>`) would be broken.
> is there any value in supporting this?

May want to compare the current CS2103 website before and after to see if there is any change. If that site is not affected, unlikely any other existing MarkBind site will be affected.

ang-zeyu:
> May want to compare the current CS2103 website before and after to see if there is any change. If that site is not affected, unlikely any other existing MarkBind site will be affected.

Sounds good 👍

PR #752, opened 2019-03-05
Ignore *.mbdf and *.njk files in user docs site and by default

Xenonym:

• [X] Documentation update
• [X] Bug fix

We now use the `.mbdf` and `.njk` extensions in the user documentation site. We should ignore these files in `site.json` so that MarkBind does not copy them over to the built site. We should also ignore these files by default as these files are source files not meant to be viewed by site visitors.

I added `*.mbdf` and `*.njk` to the `site.json` for the user documentation site and in the default `site.json`.

**Testing instructions:**
1. Build the user documentation site. It should no longer include `*.mbdf` and `*.njk` files in the built site.


yamgent:
Proposed commit message:

```
*.mbdf and *.njk files are MarkBind source files. They should not be
copied over to the final built sites.

Our user docs, and newly initialized websites, do not ignore such files,
so they are always copied over.

Let's modify the site.json settings for the user docs and the new
website defaults in order to ignore *.mbdf and *.njk files.
```

PR #756, opened 2019-03-07
Update package-lock.json to be compatible with npm 6.5.0

yamgent:

• [X] Other, please explain: package-lock.json update

* No one in the development team uses npm 5, so it does not make sense to enforce using npm 5.
* Even using an older npm version causes few changes in package-lock.json anyway, so if someone install new things, it is inevitable that someone will have to update the package-lock.json.

The package-lock.json is ~~generated~~ updated by running:

    npm i

Then, `package-lock.json` is modified so that cheerio uses our htmlparser instead of the original htmlparser.

NIL

Check that it does not break anything.

**Testing instructions:**
NIL

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       "integrity": "sha512-X3Ur/A/lIbbP8W0pmwgqtDXIxhQmxPaiwY9SKP7kF9wvZfjZRwMvbJE92ozUhF3UDK3DCKaV7oGqmI1rP/zqWA==",
       "dev": true,
       "requires": {
-        "chalk": "2.3.2",
-        "esutils": "2.0.2",
-        "js-tokens": "3.0.2"
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^3.0.0"
       },
       "dependencies": {
         "ansi-styles": {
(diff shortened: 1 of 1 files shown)


yamgent:
We now have to be careful not to overwrite this line (https://github.com/MarkBind/markbind/pull/756/commits/2bd91210c37eba6c24122c53baa03c77bc17d0b8) every time we update `package-lock.json`.

Otherwise everything seems to be fine for now.

yamgent:
As discussed on slack:

> On the issue of the npm version, I am still not convinced that updating `package-lock.json` with npm 6 is a good decision for now (the corresponding diff can be seen in this PR[1]). There is some questionable change in the version specification (meaning there could be a time when someone updates `package-lock.json` and suddenly our entire build just breaks). I would check all of the package versions if I have the time, but this could be a huge change and I am strapped for time recently to make that kind of check.
> 
> Let's stick with using an older npm 5 version. npm 5.10.0 seems to still be making some minor changes[2]. ~~Let's go lower and try npm 5.9.0.~~ (My own edit: We have since decided to use npm 5.8.0, because 5.9.0 didn't have a proper release.
>
> On a side note, does anyone from the senior dev team know what was the npm version that was used to maintain our `package-lock.json` since last year?
>
> [1] - https://github.com/MarkBind/markbind/pull/756
> [2] - https://github.com/MarkBind/markbind/pull/758#discussion_r263443847

PR #757, opened 2019-03-07
Add <meta name="generator"> tags to generated HTML

### LINKED ISSUE #754: Add meta tag for algolia search
Algolia team has asked if we can include the following meta tag in MarkBind sites that has Algoli plugin enabled.
`<meta name="generator" content="markbind">`

Xenonym:

Resolves #754.

As per #754, the Agolia team has asked us to include a `<meta name="generator">` tag to identify sites generated by MarkBind.

I added a `<meta name="generator">` tag to the page template as requested:

```html
<meta name="generator" content="MarkBind 1.19.2">
```

Since the current version number is now included in generated HTML, the `test_site` will need to be updated for each new version.

**Testing instructions:**
1. `markbind build` a MarkBind site. The new `<meta>` tag should be in the header.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -17,6 +17,8 @@ const logger = require('./util/logger');
 const MarkBind = require('./lib/markbind/src/parser');
 const md = require('./lib/markbind/src/lib/markdown-it');
 
+const CLI_VERSION = require('../package.json').version;
+
 const FOOTERS_FOLDER_PATH = '_markbind/footers';
 const HEAD_FOLDER_PATH = '_markbind/head';
 const LAYOUT_DEFAULT_NAME = 'default';
@@ -238,6 +240,7 @@ Page.prototype.prepareTemplateData = function () {
     faviconUrl: this.faviconUrl,
     headFileBottomContent: this.headFileBottomContent,
     headFileTopContent: this.headFileTopContent,
+    markBindVersion: `MarkBind ${CLI_VERSION}`,
     pageNav: this.isPageNavigationSpecifierValid(),
     siteNav: this.frontMatter.siteNav,
     title: prefixedTitle,
--- a/src/template/page.ejs
+++ b/src/template/page.ejs
@@ -4,6 +4,7 @@
     <%- headFileTopContent %>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="generator" content="<%- markBindVersion %>">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title><%= title %></title>
     <link rel="stylesheet" href="<%- asset.bootstrap %>">
--- a/test/functional/test_site/expected/bugs/index.html
+++ b/test/functional/test_site/expected/bugs/index.html
@@ -4,6 +4,7 @@
     <meta name="default-head-top">
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="generator" content="MarkBind 1.20.0">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Open Bugs</title>
     <link rel="stylesheet" href="../markbind/css/bootstrap.min.css">
--- a/test/functional/test_site/expected/index.html
+++ b/test/functional/test_site/expected/index.html
@@ -9,6 +9,7 @@
     <!-- End of top level head file content insertion -->
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="generator" content="MarkBind 1.20.0">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Hello World</title>
     <link rel="stylesheet" href="markbind/css/bootstrap.min.css">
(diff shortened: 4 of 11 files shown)


amad-person:
Could you please elaborate on why you chose to include the MarkBind version in the `<meta>` tag? Thanks!

> Since the current version number is now included in generated HTML, the test_site will need to be updated for each new version.

I'm wondering if including the version in the tag justifies this overhead?

Xenonym:
> Could you please elaborate on why you chose to include the MarkBind version in the <meta> tag? Thanks!

@amad-person since there isn't an established standard for `<meta name="generator">`, I just went by other static site generators. Most static site generators seem to include it. See: [Hexo](https://github.com/hexojs/hexo/blob/6b9acc00d64ad28f66b83118ebfe25578ae27ee6/lib/plugins/filter/meta_generator.js#L4), [Jekyll](https://github.com/jekyll/jekyll/blob/68e633a56ec712acde4b2527b6f00125b419997a/docs/_includes/top.html#L6), [Hugo](https://github.com/gohugoio/hugo/blob/7829474088f835251f04caa1121d47e35fe89f7e/common/hugo/hugo.go#L54), [Gatsby](https://github.com/gatsbyjs/gatsby/blob/db0c2ade03acbc02c5803bfb1007c56db14556c4/packages/gatsby/cache-dir/static-entry.js#L56). On the other hand, [Docasaurus](https://github.com/facebook/Docusaurus/blob/643addb7295e3d0544fb3bf6ddca060d517c7dc4/v1/lib/core/Head.js#L40) doesn't. 

hexojs/site#649 and gatsbyjs/gatsby#8221 suggest that the version number was included for analytics, so they they could track the use of different versions online. I thought it might be useful to us for the same reason as well.

amad-person:
Looks like most static site generators have this. Thanks for checking and clarifying @Xenonym!

> but not including the version because of testing seems a bit strange; we don't normally let testability dictate how features should work in other scenarios.

Good point!

LGTM.

Xenonym:
> I was thinking perhaps we could add a short commit message explaining to our future selves on why we took this decision to add a meta tag? Quote Agolia, or #754, or say that it's part of HTML5 standard perhaps?

When the commit is merged, it will be linked to this PR. Would there be merit in repeating what is here in the commit message?

yamgent:
> When the commit is merged, it will be linked to this PR. Would there be merit in repeating what is here in the commit message?

Putting a commit message would:

1. Allow people to know the context of the change without having to visit GitHub, so people can just view the message with `git log` making it self-sufficient (which also means the message last longer, since GitHub can be gone at any time).
2. Make the change more concise: PRs usually have more chatter and may include too much information.

You can suggest a proposed commit message (something like this: https://github.com/MarkBind/markbind/pull/743#issuecomment-469961993)

In your commit message, other than what @sijie123 mentioned (that it is part of HTML5 standard, and makes it easier for other external providers such as Agolia to identify MarkBind generated websites), you can also talk about why we put version numbering in value, using roughly what you have mentioned in this post: https://github.com/MarkBind/markbind/pull/757#issuecomment-470950453

Btw,

> Since the current version number is now included in generated HTML, the `test_site` will need to be updated for each new version.

This is fine (thanks for the heads up btw). We will modify our release procedure to ensure that that happens every time a new release occurs.

damithc:
Yes it is best not to rely on GitHub entirely. Some of our projects started in GoogleCode and moved to GitHub later. We may have to migrate to something else in a future date. 
It is good if the PR author can propose a full commit message for the merge commit, which can be quite useful for the person merging the PR, especially when using 'squash and merge'.

Xenonym:
Proposed commit message:

```
Add <meta name="generator"> tags to generated HTML

The Agolia team has requested that we add <meta name="generator"> 
(#754) tags to identify pages generated by MarkBind. Since this tag
is part of HTML5 [1] and can be used by other external providers to
identify MarkBind generated sites, we can include this tag in all
MarkBind pages. We can also include the version number to allow 
tracking the use of different MarkBind versions, as other projects
have done [2, 3].

Let's include the <meta name="generator"> tag to identify MarkBind
pages, as well as including the version number to track the use of
different MarkBind versions.

[1]: https://www.w3.org/TR/html5/document-metadata.html#generator
[2]: https://github.com/hexojs/site/pull/649
[3]: https://github.com/gatsbyjs/gatsby/pull/8221
```

Xenonym:
@yamgent rebased to update the version number.

damithc:
Just curious: is there a way to use this tag (or a similar tag) to get a sense of how many pages in the Internet were built using MarkBind? For example, searching for the tag via a search engine

yamgent:
> For example, searching for the tag via a search engine

That would be quite a specialized and niche search, I don't recall Google Search supporting such a filter. Anyone interested in such a result would probably need to build their own web crawler to check?

Xenonym:
> Just curious: is there a way to use this tag (or a similar tag) to get a sense of how many pages in the Internet were built using MarkBind? For example, searching for the tag via a search engine

@damithc [Wappalyser](https://www.wappalyzer.com/) can track uses of web technologies, [provided the app is "widely-used"](https://www.wappalyzer.com/docs#new-application) (emphasis mine):

> Only widely used applications should be added. When creating a pull request, **please include ten or more links to websites that use the application, a GitHub page with at least 1,000 stars or anything that will help establish the size of the user base**.

Could be something to look at once MarkBind grows big enough.

damithc:
Thanks for checking @Xenonym 
Yes, let's hope to get there some day :-)

Issue #746, opened 2019-03-03
Use Prettier to Format Code


marvinchin:
Right now, we are manually formatting code to comply with the linting rules. Having to do this can be somewhat annoying to developers, and can also result it some inconsistencies across the codebase (e.g. where/when to break to a new line).

Instead, we can use a code formatter such as [Prettier](https://prettier.io) to automatically format the code for us. This would avoid the overhead of formatting our own code and makes the codebase look more consistent.

yamgent:
I recall eslint already providing some options to automate the fixing of eslint code style violations.

marvinchin:
Yes it does, but outside of fixing linting errors, Prettier provides better support for formatting code consistently (e.g. there are many ways to break a line exceeding 80 lines, Prettier enforces a consistent way of breaking them).

Not sure if the rest of the team sees much value in this, but it was a tool I used in other projects that I found really helpful so I brought it up 🙂

Chng-Zhi-Xuan:
Seems like a good developer QOL enhancement. 

I'm pretty tired of using `./node_modules/.bin/eslint . --fix` as well :P

damithc:
It can be used individually without having to enforce on the entire project first?

marvinchin:
Formatting files individually as we work on them might cause large, unnecessary diffs (since prettier will format the entire file). I'm concerned that it would make the PR process more difficult.

damithc:
> Formatting files individually as we work on them might cause large, unnecessary diffs (since prettier will format the entire file). I'm concerned that it would make the PR process more difficult.

I see. I have no objections. Will leave the decision to senior devs :-)

yamgent:
I am open to giving this a try. Open a trial PR with Prettier and we can see how it goes.

Also a concern would be whether we can override the styles dictated by Prettier when necessary. For example, with regards to breaking a line exceeding 80 lines, sometimes it may enforce a style that may not be ideal in terms of readability. I would like to disable the enforcement for that particular line in such a situation.

marvinchin:
Sure, I'll do it tonight 🙂 
Do we intend to adhere to the current eslint rules? Because Prettier's formatting rules may differ from some of our existing rules (e.g. the `operator-linebreak` and `array-bracket-newline` rules). There are certain ways around this, but it is a bit of a workaround and may not play nice with all editor plugins.

Regardless, I will open a PR with the Prettier formatting and see what everyone thinks about it!

yamgent:
[...]

PR #758, opened 2019-03-07
[RFC] Apply Prettier Formatting

marvinchin:
#746 

Attempt to use Prettier to format working files.

As recommended in the Prettier docs, I've disabled all formatting related eslint rules (Prettier will take care of formatting). Code quality rules are still retained.

You can use Prettier plugins (available on most IDEs) to automatically format your code while coding 🙂 

Do let me know what you guys think about this, and whether or not we should go ahead with it!

EDIT:
Just to note, this PR is to show how Prettier formatting would look like, and is **not meant to be merged into master**


.eslintignore
```diff
@@ -1,4 +1,3 @@
 *.min.*
 src/lib/markbind/src/lib/markdown-it/*
-
-!.eslintrc.js
+.eslintrc.js
```
acjh: Linting this file is intentional. It is ignored by default, so this line would be redundant anyway.

.eslintrc.js
```diff
@@ -2,55 +2,46 @@
...
+  plugins: ['lodash', 'prettier'],
+  extends: ['airbnb-base', 'plugin:lodash/recommended', 'prettier'],
+  rules: {
+    'func-names': 'off',
+    'function-paren-newline': 'off',
+    'lodash/prefer-lodash-method': [0],
+    'lodash/prefer-noop': [0],
+    'max-len': ['error', { code: 110 }],
+    'no-underscore-dangle': 'off',
+    'prettier/prettier': 'error',
```
acjh: Avoid changing formatting and config in the same commit unless syncing configs.
Also, this violates the Eslint rule overrides at the top of the file.

I kept double quotes as their documentation uses that and many projects do too, so it's easier to copy rules as-is.

package-lock.json
```diff
@@ -64,7 +64,7 @@
     "@sindresorhus/slugify": {
       "version": "0.8.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/slugify/-/slugify-0.8.0.tgz",
-      "integrity": "sha512-Y+C3aG0JHmi4nCfixHgq0iAtqWCjMCliWghf6fXbemRKSGzpcrHdYxGZGDt8MeFg+gH7ounfMbz6WogqKCWvDg==",
+      "integrity": "sha1-VVC3+gZPOoqCZRRjrWNTeAVMctA=",
```
acjh: Why would you even commit and push this?

marvinchin: I did npm install using npm 5 as you suggested [here](https://github.com/MarkBind/markbind/pull/745#issuecomment-468982463), and this was the output generated.

.eslintignore
marvinchin: Without this line `npm run lint` seems to be picking up `.eslintrc.js` on my machine, not sure why it doesn't seem to be consistent.

package-lock.json
acjh: v5.10.0?

marvinchin: Yes. On further investigation the issue seems to be reported [here](https://github.com/npm/npm/issues/17749), though for a different npm version.

acjh:
To clarify, I found *the disregard of [this comment](https://github.com/MarkBind/markbind/issues/746#issuecomment-470390412)* by this PR offensive and felt strongly about it:

>> Do we intend to adhere to the current eslint rules?
>
> Our current eslint rules must be followed.

I understand now that you spoke to @yamgent about it in person. Please post a brief update there.

Thank you for your feedback and I will work on improving my tone.

acjh:
> As recommended in the Prettier docs, I've disabled all formatting related eslint rules (Prettier will take care of formatting).

I like that tooling would remove ambiguity, just not the choice of formatting rules.

The `prettier` plugin overrides many Eslint rules from Airbnb. Prettier should be used to format based on the project's rules, not be allowed to freely define the formatting rules of the project.

marvinchin:
To update on our conversation with @nicholaschuayunzhi and @yamgent today:

We were discussing how to make Prettier play nice with our existing eslint rules, and the decision was to remove the formatting rules first to let the team observe how Prettier formatting works on its own and see if those are to the team's preference (hence this PR).

I apologise if there was any confusion regarding the intent or context of this PR - I'll include updates on any related offline discussions in the future to avoid ambiguity.

marvinchin:
Also, to give a little more context as to why it was tricky to make prettier work alongside our existing eslint configs:

Prettier is an **opinionated** code formatter, so it does not follow the formatting rules defined by eslint, but instead assumes full responsibility for the formatting of the code. This naturally means any formatting related eslint rules will no longer be relevant if prettier is used (code quality rules are still useful).

As @acjh pointed out, this would mean letting prettier define the formatting rules of the project. The advantage to this is that it removes the cognitive overhead and ambiguity in thinking about how code should be formatted, but we lose the ability to define more granular rules for formatting if they differ from prettier's way of formatting code.

nicholaschuayunzhi:
I apologize for not stepping in to clear up the confusion earlier.

I too was surprised by how aggressive Prettier was. I do agree with @acjh, it makes more sense for Prettier to enforce our current eslint rules if possible - many of the configurations made by different developers were purposeful. I apologize for not thinking too deeply about that. However, that being said, we would not have been able to make a well informed judgement without the PR by @marvinchin, so much thanks for trying it out!

damithc:
Looks like this is not going to be an easy decision as there are pros and cons. In that case shall we maintain the status quo? Especially because this doesn't directly value add to our immediate goal which is V2? While it may be possible to tweak prettier to a level we are all OK with, the required effort may distract our attention away from the goal at hand. We can always revisit it again at a less-pressure time.

Thanks @marvinchin for experimenting with prettier and others for the feedback.

Other possible takeaways:
* Remember to update GitHub threads with a summary of what was discussed off-line or using other channels. Even slack discussions may not be visible to future contributors.
* 'Opinionated' tools are better adopted at the start of a project. They are harder to adopt in ongoing projects as it is unlikely that every project member's opinion will agree with the tool's.

marvinchin:
Thanks everyone for the inputs! 🙂

PR #765, opened 2019-03-12
Fix incorrect regex for matching icon syntax

### LINKED ISSUE #764: Unexpected behavior when using variables and icons
v1.20.0

These variables don't seem to behave as expected:
in `variables.md`:
```html
<variable name="year">2019</variable>
<variable name="AY">18/19</variable>
<variable name="cs3281_time">MON 1200 - 1400</variable>
<variable name="icon_dislike">:fas-thumbs-down:</variable>
```
in another file:
```html
{{ year }}<br>
{{ AY }}<br>
{{ cs3281_time }}<br>
{{ icon_dislike }}<br>
```
output:
image

Xenonym:

• [X] Bug fix

Fixes #764.

The current regex for matching icon syntax has the `m` multi-line flag turned on, which causes unexpected results in Markdown with special characters.

Remove the `m` flag.

In `variables.md`:
```html
<variable name="cs3281_time">MON 1200 - 1400</variable>
<variable name="icon_dislike">:fas-thumbs-down:</variable>
```

In a MarkBind page:
```md
{{ cs3281_time }}<br>
{{ icon_dislike }}<br>
```

**Testing instructions:**
The above sample code should render correctly:

### DIFF
--- a/test/unit/markdown-it-icons.test.js
+++ b/test/unit/markdown-it-icons.test.js
@@ -1,0 +1,35 @@
+const markdownIt = require('markdown-it')()
+  .use(require('../../src/lib/markbind/src/lib/markdown-it/markdown-it-icons'));
+
+test('markdown-it-icons renders icon syntax correctly', () => {
+  const source = ':fab-font-awesome: :glyphicon-home:';
+
+  const result = markdownIt.renderInline(source);
+  const expected = [
+    '<span aria-hidden="true" class="fab fa-font-awesome"></span>',
+    '<span aria-hidden="true" class="glyphicon glyphicon-home"></span>',
+  ].join(' ');
+
+  expect(result).toEqual(expected);
+});
+
+test('markdown-it-icons renders icon syntax in multi-line source correctly', () => {
+  const source = [
+    'text with special characters: before icons',
+    ':fab-font-awesome:',
+    'text with special characters: between icons',
+    ':glyphicon-home:',
+    'text with special characters: after icons',
+  ].join('\n');
+
+  const result = markdownIt.renderInline(source);
+  const expected = [
+    'text with special characters: before icons',
+    '<span aria-hidden="true" class="fab fa-font-awesome"></span>',
+    'text with special characters: between icons',
+    '<span aria-hidden="true" class="glyphicon glyphicon-home"></span>',
+    'text with special characters: after icons',
+  ].join('\n');
+
+  expect(result).toEqual(expected);
+});
--- a/src/lib/markbind/src/lib/markdown-it/markdown-it-icons.js
+++ b/src/lib/markbind/src/lib/markdown-it/markdown-it-icons.js
@@ -1,5 +1,5 @@
 module.exports = require('markdown-it-regexp')(
-    /:(fa[brs]|glyphicon)-([a-z-]+):/m,
+    /:(fa[brs]|glyphicon)-([a-z-]+):/,
     (match, utils) => {
         let iconFontType = match[1];
         let iconFontName = match[2];


damithc:
On a somewhat related note, it would be good if there is a way to unit-test regular expressions.

yamgent:
Am I missing something really obvious here? 🤔 I don't see how the `m` multi-line flag caused the bug in the first place (like how it someone did only a partial match).

Xenonym:
> On a somewhat related note, it would be good if there is a way to unit-test regular expressions.

@damithc since regexes are used to implement specific functions, I think that the unit test should be testing the behaviour of the functions rather than the regexes themselves, since these functions may evolve to not use regexes for their behaviour.

> Am I missing something really obvious here? 🤔 I don't see how the m multi-line flag caused the bug in the first place (like how it someone did only a partial match).

@yamgent To be fair, it's not very obvious: it's an unexpected interaction between `RegExp` and `markdown-it-regexp`. Consider the example source code (after replacing variables):
```html
MON 1200 - 1400<br>
:fas-thumbs-down:<br>
```
1. 	`markdown-it` parses this source as a `inline` token, with `MON 1200 ` being parsed as `text` as [they do not contain special characters that may be part of new syntax](https://github.com/markdown-it/markdown-it/blob/d533c27017dbb9663422019a08adcf15265cf9fb/lib/rules_inline/text.js#L7-L10). 
2. `markdown-it-regexp` parses the remaining source, which contains a match due to the `m` flag. However, it only replaces the first 17 characters (the length of `:fas-thumbs-down:`), leading to replacing `- 1400<br>\n:fas-t` in the source instead.

If we remove the `m` flag, `markdown-it-regexp` will not match tokens beyond line boundaries, which matches the behavior of other inline syntax as they do not involve multi-line syntax.

damithc:
> @damithc since regexes are used to implement specific functions, I think that the unit test should be testing the behaviour of the functions rather than the regexes themselves, since these functions may evolve to not use regexes for their behaviour.

That works, although the cost of testing the entire function can limit the number of test cases.
An alternative is to define the regex as a constant and use it in the functions as well as in separate lightweight unit tests where only the regex is tested. Not sure if this is possible in JS though.

yamgent:
> @yamgent To be fair, it's not very obvious: it's an unexpected interaction between `RegExp` and `markdown-it-regexp`. Consider the example source code (after replacing variables):
> 
> ```
> MON 1200 - 1400<br>
> :fas-thumbs-down:<br>
> ```
> 
> 1. `markdown-it` parses this source as a `inline` token, with `MON 1200 ` being parsed as `text` as [they do not contain special characters that may be part of new syntax](https://github.com/markdown-it/markdown-it/blob/d533c27017dbb9663422019a08adcf15265cf9fb/lib/rules_inline/text.js#L7-L10).
> 2. `markdown-it-regexp` parses the remaining source, which contains a match due to the `m` flag. However, it only replaces the first 17 characters (the length of `:fas-thumbs-down:`), leading to replacing `- 1400<br>\n:fas-t` in the source instead.
> 
> If we remove the `m` flag, `markdown-it-regexp` will not match tokens beyond line boundaries, and should only match icon syntax at the start of the `markdown-it` parser source position (after all tokens before the icon font have been parsed).

@Xenonym thanks for the detailed explanation, appreciate it. This part can be in the commit message so that future developers will know what happened and not get lost. :+1:

yamgent:
Also is it possible to unit test it? We could use 

```html
MON 1200 - 1400<br>
:fas-thumbs-down:<br>
```

as one of the test cases.

Xenonym:
> An alternative is to define the regex as a constant and use it in the functions as well as in separate lightweight unit tests where only the regex is tested. Not sure if this is possible in JS though.

This is possible, but would it be confusing to have very different regexes from multiple functions reside in the same file? We have multiple regexes in MarkBind (`Page.js`, `Site.js`, `filterTags.js`).

Xenonym:
@yamgent unit tests added and commit message updated!

damithc:
> This is possible, but would it be confusing to have very different regexes from multiple functions reside in the same file? We have multiple regexes in MarkBind (`Page.js`, `Site.js`, `filterTags.js`).

Doesn't need to be in the same file. Just need to define the regex in one place and use it both from the test and the production code. Anyway, this is just a thought. Perhaps there are better ways to test regex correctness? We can't be the only ones getting bitten by regex errors :-p Unlike other code, it is very hard for a reviewer to spot regex errors, which means it is good to have another way to ensure their accuracy.

test/unit/markdown-it-icons.test.js
```diff
@@ -0,0 +1,35 @@
+const markdownIt = require('markdown-it')()
+  .use(require('../../src/lib/markbind/src/lib/markdown-it/markdown-it-icons'));
+
+test('markdown-it-icons renders icon syntax correctly', async () => {
```
nicholaschuayunzhi: Correct me if i'm wrong but I don't think these functions have to be `async`

Xenonym: Yeah you are right, thanks for catching this! Have removed the `async`.

yamgent: The rest of the test cases in `parser.test.js` and `Site.test.js` are also using `async`, seems like we can clean those up in a separate PR.

damithc:
Seems to be working now 👍

yamgent:
> That works, although the cost of testing the entire function can limit the number of test cases.
> An alternative is to define the regex as a constant and use it in the functions as well as in separate lightweight unit tests where only the regex is tested. Not sure if this is possible in JS though.

I surveyed around and the general consensus is that, if the regex is complex enough, then usually the pattern matching would be a method of its own, so unit testing the regex would simply be unit testing a normal method. For example, if we take this PR, it would be refactored into something like this theoretically:

```js
const getFontTypeAndName = (markdownSource) => {
    const match = markdownSource.regex(/:(fa[brs]|glyphicon)-([a-z-]+):/);
    
    if (!match) {
        return { iconFontType: '', iconFontName: '' };
    }
    
    return { iconFontType: match[1], iconFontName: match[2] };
};

const replaceSyntaxWithSpan = (matchInfo, utils) => {
    if (matchInfo.iconFontType === 'glyphicon') {
        return `<span aria-hidden="true" class="glyphicon glyphicon-${matchInfo.iconFontName}"></span>`;
    } else { // If icon is a Font Awesome icon
        return `<span aria-hidden="true" class="${matchInfo.iconFontType} fa-${matchInfo.iconFontName}"></span>`;
    }
};

    getFontTypeAndName,   // unfortunately we can't do this
    replaceSyntaxWithSpan
);
```

Unfortunately the only thing that is stopping us is that `markdown-it-regexp` seems to only accept regex, rather than a method that does regex (I marked that line with `// unfortunately we can't do this`).

So if we want to test the regex, then the only way to do it for this scenario is to declare it as a constant and test that constant.

For the rest of the codebase that uses regex, I think the refactoring method works for them, and we can refactor them into a structure similar to `getFontTypeAndName()` for unit testing.

PR #766, opened 2019-03-13
Fix trim not working when including fragments

### LINKED ISSUE #673: include/trim doesn't trim blank lines
v1.17.2

Current:

`foo.mbd`:
```html
<span id="bar">

    abc

</span>
```

`x<include src="foo.md#bar" inline trim/>x` should result in `xabcx` but results in the below instead:
```
x
abc
x
```

marvinchin:

• [x] Bug fix

Fixes #673.

Trim was not working when including fragments.

The error was caused because we trimmed only the **file content** when the trim attribute is present. However, the actual content in the fragment is not trimmed.

This PR ensures that the trim method is called on the content that will actually be included.

**Source**:
foo.md
```md
x<include src="bar.md#fragment" trim inline />x
```

bar.md
```
<span id="fragment">

  abc

</span>
```

**Output**:
```
xabcx
```

**Testing instructions:**
The example in #673 should now work.


src/lib/markbind/src/parser.js
```diff
@@ -305,14 +305,12 @@ Parser.prototype._preprocess = function (node, context, config) {
     delete element.attribs.inline;
     delete element.attribs.trim;
 
-    fileContent = isTrim ? fileContent.trim() : fileContent;
-
     if (includeSrc.hash) {
       // directly get segment from the src
       const segmentSrc = cheerio.parseHTML(fileContent, true);
       const $ = cheerio.load(segmentSrc);
       const htmlContent = $(includeSrc.hash).html();
```
Xenonym: Could we call this variable `fragmentContent` instead? I think trying to mentally differentiate between `htmlContent` and `fileContent` is confusing.

test/functional/test_site/testTrimIncludeFragment.md
```diff
@@ -0,0 +1,5 @@
+<span id="fragment">
+
+    Fragment with leading spaces and newline
+
+</span>
```
Chng-Zhi-Xuan: Small inconsistency, fragment files have the extension of `.mbdf` instead of `.mb` (given that you explicitly named the file a fragment).

test/functional/test_site/index.md
```diff
@@ -133,6 +133,10 @@ tags: ["tag-frontmatter-shown", "tag-included-file", "+tag-exp*", "-tag-exp-hidd
 
 ## <include src="testTrimInclude.md" trim inline />
 
+# Trimmed include fragment
```
Xenonym: Could this test be better written as a `parser.js` unit test?

test/functional/test_site/index.md
```diff
@@ -133,6 +133,10 @@ tags: ["tag-frontmatter-shown", "tag-included-file", "+tag-exp*", "-tag-exp-hidd
...
+
+## Before <include src="testTrimIncludeFragment.md#fragment" trim inline /> After
```
Xenonym: Could we make the Before and After parts more visually distinct? For example:
```
Before | <include src="testTrimIncludeFragment.md#fragment" trim inline /> | After
```
or `Before` and `After` could be in a different colour?

marvinchin:
@Xenonym @Chng-Zhi-Xuan I've updated the PR based on your comments 🙂

yamgent:
Propose a merge commit message.

marvinchin:
**Commit Message:**
```
Fix trim not working when including fragments

When a fragment is included with the trim attribute specified, the content remains untrimmed.

The error was caused because only the file content is trimmed when the trim attribute is present. However, the actual content in the fragment is not trimmed.

Let's update the parser to ensure that the actual content to be included is trimmed.
```

yamgent:
Amended your commit message to wrap at 72 characters + make it more concise.

```
Fix trim not working when including fragments

The trim attribute only works when including an entire file. When a
fragment is included instead with the trim attribute specified, the
content remains untrimmed.

Let's fix the parser to ensure that included fragments are also trimmed.
```

PR #768, opened 2019-03-14
Allow inclusion/exclusion of headings for indexing

### LINKED ISSUE #737: Give a way to exclude/include a specific heading from the search index
Current: We can set which heading level in a page to use for search indexing

Suggested: Also give a way to specify that a certain heading should be omitted from the search index although it qualifies based on the heading level. Similarly, we can give a way to specify that a certain heading should be indexed even though it does not qualify based on the heading level.

marvinchin:

Fixes #737.

When using the built-in markbind search, users are able to set the heading level to be indexed. We want to users to be able to specify that certain headings should be indexed or not be indexed regardless of the heading index level.

To support this, we allow users to add the `{.index}` or `{.noindex}` attributes to headings to indicate that the heading should either be indexed or not be indexed.

Update heading selectors to include any heading with the `index` class, and exclude any heading with the `noindex` class.

```md
###### This will always be indexed {.index}

# This will never be indexed {.noindex}

```

**Testing instructions:**

See related headings in the test site.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -205,15 +205,17 @@ function formatSiteNav(renderedSiteNav, src) {
 }
 
 /**
- * Generates a heading selector based on the indexing level
+ * Generates a selector for headings with level inside the headingIndexLevel
+ * or with the index attribute, that do not also have the noindex attribute
  * @param headingIndexingLevel to generate
  */
 function generateHeadingSelector(headingIndexingLevel) {
-  let headingsSelector = 'h1';
+  let headingsSelectors = ['.always-index:header', 'h1'];
   for (let i = 2; i <= headingIndexingLevel; i += 1) {
-    headingsSelector += `, h${i}`;
+    headingsSelectors.push(`h${i}`);
   }
-  return headingsSelector;
+  headingsSelectors = headingsSelectors.map(selector => `${selector}:not(.no-index)`);
+  return headingsSelectors.join(',');
 }
 
 function unique(array) {
--- a/test/functional/test_site/expected/index.html
+++ b/test/functional/test_site/expected/index.html
@@ -147,6 +147,10 @@
     </nav>
     <a class="nav-link py-1" href="#test-plugin-in-markbind-plugins">Test plugin in markbind/plugins&#x200E;</a>
     <a class="nav-link py-1" href="#markbind-plugin-pre-render">Markbind Plugin Pre-render&#x200E;</a>
+    <a class="nav-link py-1" href="#test-search-indexing">Test search indexing&#x200E;</a>
+    <nav class="nav nav-pills flex-column my-0" style="margin-left: 5%; flex-wrap: nowrap;">
+      <a class="nav-link py-1" href="#level-6-header-outside-headingsearchindex-with-always-index-attribute-should-be-indexed">Level 6 header (outside headingSearchIndex) with always-index attribute should be indexed&#x200E;</a>
+    </nav>
 
   </nav>
 </nav>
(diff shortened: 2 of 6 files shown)


test/functional/test_site/index.md
```diff
@@ -219,3 +219,9 @@ tags: ["tag-frontmatter-shown", "tag-included-file", "+tag-exp*", "-tag-exp-hidd
 <div id="test-markbind-plugin">
   Markbind Plugin Pre-render Placeholder
 </div>
+
+# Test search indexing
+
+## Level 2 header with noindex attribute should not be indexed {.noindex}
+
+###### Level 6 header with index attribute should be indexed {.index}
```
yamgent: I think it would be clearer if you say that the heading level is inside/outside the `headingSearchIndex`, so that in case this fails in the future (failing may be due to changed `headingSearchIndex`) the context would be much clearer.

E.g.
> Level 6 header **(outside headingSearchIndex)** with index attribute...

test/functional/test_site/index.md
```diff
@@ -219,3 +219,9 @@ tags: ["tag-frontmatter-shown", "tag-included-file", "+tag-exp*", "-tag-exp-hidd
 <div id="test-markbind-plugin">
   Markbind Plugin Pre-render Placeholder
 </div>
+
+# Test search indexing
```
yamgent: Hmm, "Level 1 header without attributes specified should be indexed" to be more consistent with the rest?

src/Page.js
```diff
@@ -209,11 +209,12 @@ function formatSiteNav(renderedSiteNav, src) {
...
  */
 function generateHeadingSelector(headingIndexingLevel) {
-  let headingsSelector = 'h1';
+  let headingsSelectors = ['.index:header', 'h1'];
   for (let i = 2; i <= headingIndexingLevel; i += 1) {
-    headingsSelector += `, h${i}`;
+    headingsSelectors.push(`h${i}`);
   }
-  return headingsSelector;
+  headingsSelectors = headingsSelectors.map(selector => `${selector}:not(.noindex)`);
```
yamgent: The function comment needs to be updated since it is no longer just based on indexing level alone any more.

test/functional/test_site/index.md
marvinchin: I intended for this to be a "heading" which describes the section containing the two headings below it. This is mainly because the sidebar links have indentation based on the heading level, and without this level 1 heading, the links generated below will appear to be under the "Markbind Plugin Pre-render‎" section instead.

Screenshot 2019-03-14 at 4 40 24 PM

With this in mind, I'm not exactly sure if "Level 1 header without attributes specified should be indexed" would accurately describe this section 😅

yamgent: I see, my bad then, I thought this was part of the test cases at first glance.

test/functional/test_site/expected/index.html
```diff
@@ -557,6 +561,9 @@ <h1 id="test-plugin-in-markbind-plugins">Test plugin in markbind/plugins<a class
           <h1 id="markbind-plugin-pre-render">Markbind Plugin Pre-render<a class="fa fa-anchor" href="#markbind-plugin-pre-render"></a></h1>
           <p>Node Modules Plugin Post-render</p>
         </div>
+        <h1 id="test-search-indexing">Test search indexing<a class="fa fa-anchor" href="#test-search-indexing"></a></h1>
+        <h2 class="noindex" id="level-2-header-inside-headingsearchindex-with-noindex-attribute-should-not-be-indexed">Level 2 header (inside headingSearchIndex) with noindex attribute should not be indexed</h2>
```
Xenonym: Should we remove the `index` and `noindex` classes from the final rendered HTML?

marvinchin: Thanks for pointing it out! I think that it makes sense to have it omitted in the output HTML, but our current site building process seems to make this difficult because 1) the index is only built **after** the HTML is generated and written to the file system, and 2) the index is built by reading the generated HTML file.

I think it might be possible to overcome this by separating the generation step from the writing step, and inserting the indexing step in between the two, but I feel like this might complicate the build process.

We *could* remove these classes in `setup.js`, but I'm not sure if there is any value to it.

Would love to get inputs from the rest on the team on this! 🙂

Xenonym: > We could remove these classes in `setup.js`, but I'm not sure if it's a good idea.

There's no point removing those classes on page load is there? It would still be in the rendered HTML.

marvinchin: It would remove those classes from the affected elements after the initial render, but agreed - I don't think there's much value in doing that.

yamgent:
Update the commit message to reflect the new changes (e.g. `{always-index}`). Also perhaps they can be squashed?

marvinchin:
@yamgent done!

PR #689, opened 2019-02-11
Fix tagging documentation

jamos-tay:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] Documentation update

**What is the rationale for this request?**

The tagging documentation was moved from [tweakingThePageStructure](https://markbind.github.io/markbind/userGuide/tweakingThePageStructure.html) to [usingPlugins](https://markbind.github.io/markbind/userGuide/usingPlugins.html) when #474 was merged, but the outdated documentation wasn't removed, probably due to when the documentation was reshuffled

**What changes did you make? (Give an overview)**

Removed outdated documentation and linked it to the new, correct one
Created a folder for documentation of plugins

PR #770, opened 2019-03-14
Fix tagging documentation under plugins

jamos-tay:

• [X] Documentation update

Somehow during the documentation restructuring the filterTags documentation was changed to the older, incorrect one (#474). It should be the one from #689.

Replaced it with the newer one (#689)

### DIFF
--- a/docs/userGuide/siteConfiguration.md
+++ b/docs/userGuide/siteConfiguration.md
@@ -50,7 +50,14 @@ Here is a typical `site.json` file:
     "*.mbd",
     ".git/*"
   ],
-  "tags": ["tag1", "tag2"],
+  "plugins" : [
+    "filterTags"
+  ],
+  "pluginsContext" : {
+    "filterTags" : {
+      "tags": ["tag1", "tag2"]
+    }
+  },
   "headingIndexingLevel": 4
 }
 ```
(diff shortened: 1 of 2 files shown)


jamos-tay:
Updated

docs/userGuide/siteConfiguration.md
```diff
@@ -50,7 +50,14 @@ Here is a typical `site.json` file:
...
     ".git/*"
   ],
-  "tags": ["tag1", "tag2"],
+  "plugins" : [
+    "filterTags"
+  ],
+  "pluginsContext" : {
+    "filterTags" : {
+      "tags": ["tag1", "tag2"]
+    }
```
yamgent: The addition is good, but there's a lack of explanation of the `plugins` and the `pluginsContext` attribute.

The `tags` explanation below will also seems to be out of place now (since we will be explaining plugin settings for all kinds of plugin). Maybe we can say something like "the example above uses `tags` as an example of configuring plugin settings, refer to the Tag plugin for more details" or something similar?

jamos-tay: Sure, I reworded it

jamos-tay:
Updated, rebased

PR #774, opened 2019-03-15
test: fix permissions for test/functional/test.sh

luyangkenneth:

• [X] Bug fix

The following error shows up when doing `npm run test` on master. This points to a lack of execution permissions on the local machine (I use a mac).

Screenshot 2019-03-15 at 8 40 42 PM

Upon investigation, it seems that the permissions were modified in #588 (see screenshot below for relevant section of the diff) when moving the then `test.sh` to a different directory and creating a new `test.sh`.
Screenshot 2019-03-15 at 8 50 23 PM

**What changes did you make?**
1. `chmod +x test/functional/test.sh` (All our executable files are currently set to 755, except for this one.)
2. Removed the `before_install` config in `.travis.yml` because we shouldn't need it anymore

**Testing instructions:**
1. `npm run test` locally
2. Tests pass on travis


.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
   - '8'
-before_install:
-  - chmod +x ./test/functional/test.sh
```
yamgent: This should not be removed?

PR #777, opened 2019-03-18
Convert line endings to LF and enforce with .gitattributes config file

luyangkenneth:
This PR is to standardize the line endings for all text files in our codebase.

I noticed that our codebase has a majority of LF (Unix) encoded files, with a few CRLF (Windows) encoded files sprinkled here and there. While this hasn't caused any issues for us so far, I feel that this is an important step forward, given that MarkBind is all about file processing.

**What changes did you make?**
1. Convert CRLF to LF in unconverted text files ([done automagically with git](https://help.github.com/en/articles/dealing-with-line-endings#refreshing-a-repository-after-changing-line-endings))
2. Add the `.gitattributes` config file to enforce that for all future changes, git will detect all text files and automatically normalize them to LF

If you're interested, check out the `The New System` section of [this article](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/) for some further reading.

**Test plan:**
- CI
- `markbind serve` works as expected


marvinchin:
I agree that enforcing standardisation is definitely welcome, especially since we don't have a consistent environment for developers. This would help us to avoid the strange/confusing diffs that sometimes happen due to these invisible differences.

marvinchin:
Also, did you push to the wrong remote? 😛 (I believe MarkBind workflow uses branches from our individual remotes - I believe this isn't documented anywhere, but probably should to help new contributors understand the workflow)

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
yamgent: The [git documentation](https://git-scm.com/docs/gitattributes) says that:

```
Set to string value "auto"
When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. When the file has been committed with CRLF, no conversion is done.
```

So it is still possible for people to push files with CRLF to the repo, since the conversion is only done on pull, according to the last line? A better enforcement would be needed, because adding this only hides the problem. 😕

luyangkenneth: I think the documentation is written quite vaguely, so you might have taken away the wrong understanding from it.

My understanding (which I believe is correct because I spent some time playing around with the settings) is that:
- The `auto` setting converts (all files that are determined by Git to be text and not binary) from CRLF to LF **_upon check in_**.
- This means that I could be committing a CRLF file, but what actually goes into the repo is the converted LF version:

Screenshot 2019-03-19 at 1 00 24 AM

- The file remains as CRLF on my local repo. I haven't tested it, but my hunch is that it will get overwritten by the actual LF version when pulling from remote later on. In any case, the source of truth (our git remote) will remain as LF, and any inconsistencies will be limited to individual clones.
- As this is a git feature, it makes sense to me that it only controls the way files are versioned in the repo, rather than modifying what's on your local machine. So I think this is all the enforcement power there is to `.gitattributes`.

The gist is, after I looked through the documentation and tested different settings, I came to the conclusion that `* text=auto` does what we want in the easiest and most pain-free way.

luyangkenneth:
Thanks for the review @yamgent! I see where you're coming from re: how this will add a new entry into the blame log, for when future developers wish to look at the version history.

My thinking here is to have a bias for action - aka how do we do the right thing and focus on moving the project forward, while we're also aware of tradeoffs like polluting the blame log? In this case I'd take the view that it's not _that_ painful (in the github UI at least) to navigate past one additional commit in the file history when trying to pinpoint the origin of a specific code change.

I would also love to have a system where whitespace-only changes such as these don't show up in a file's blame log. Unfortunately I've only seen it at large companies that heavily customise their VCS, so I don't know of a cheap way for us to move to something like that right now.

What are your thoughts and do you see any further changes that you'd like me to make to this PR? 🙂

damithc:
Perhaps we can delay this until the end of semester, as the PR might mess up RepoSense output of the MarkBind repos?

.gitattributes
yamgent: Thanks for your explanation, checked it out on my side, seems to be working as you explained. :+1:

yamgent:
> Perhaps we can delay this until the end of semester, as the PR might mess up RepoSense output of the MarkBind repos?

@damithc does RepoSense use any 'ignore whitespace' options when analysing commits?

damithc:
> @damithc does RepoSense use any 'ignore whitespace' options when analysing commits?

Previously we had an option like that but we disabled it because it increases the processing time by a lot (like x10 times!)

@yamidark is that option available?

yamidark:
>Previously we had an option like that but we disabled it because it increases the processing time by a lot (>like x10 times!)
>
>@yamidark is that option available?

IIRC, the option that we removed to increase the processing time is the copy/move mechanism (tracking lines that were copied/moved back to their original/parent author).

In fact, the current `git blame` function used in RepoSense already uses the `-w` option for ignoring whitespaces changes:

https://github.com/reposense/RepoSense/blob/ba7c96d7ccb84c372d114ee02c3491855158bbd2/src/main/java/reposense/git/GitBlame.java#L29

damithc:
Thanks @yamidark 
@yamgent it looks like changing the line endings might not affect RepoSense output.

yamgent:
> @yamgent it looks like changing the line endings might not affect RepoSense output.

Noted, I will merge this PR now then so that the changes are incorporated immediately for other PRs.

PR #778, opened 2019-03-18
Use Sets instead of Dictionaries for Page.includedFiles

### LINKED ISSUE #772: Use of Javascript dictionaries for includedFiles
* **MarkBind Version:**
Markbind 1.20.0

We currently use includedFiles to check what pages we need to rebuild when they are updated.
When a file is included, we do something similar to:
https://github.com/MarkBind/markbind/blob/07814ff70cac23c4a6b78c6fa6b99fe59a25ca61/src/Page.js#L487
to indicate that it should be flagged for updating.

I notice that includedFiles is a dictionary, containing only {file: true} values.
With ES6, we can actually use the Set data structure to store this information, instead of tagging everything as true. It's a minor coding style thing, but I felt it's just odd to use a dictionary of `true`s.

If we use a set, then in Site.js we can simply do
```
if (shouldRebuildAllPages || filePaths.some(filePath => page.includedFiles.has(filePath)))`
```
which is semantically a bit clearer than `page.includedFiles[filePath]` (<-- what is the return data type?)

Would this be a good idea? It doesn't really affect our correctness now, but it's a low hanging fruit that can be resolved easily.

sijie123:

• [x] Other, please explain: Coding style "fix"

Resolves #772 

We use includedFiles to check what pages we need to rebuild when a page is updated. When a file is included, we do something similar to `this.includedFiles[dependency.to] = true;`. includedFiles is essentially a dictionary containing all {file: true} values.

Let's use the ES6 Set data structure to store this information. 

Use ES6 Set data structure instead of a dictionary.
This allows us to use `page.includedFiles.has(filePath)` when checking for existence, which is semantically clearer (it clearly returns a boolean) as compared to `page.includedFiles[filePath]`.

Hmm... Hopefully nothing breaks? 😓 

**Testing instructions:**

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -89,7 +89,7 @@ function Page(pageConfig) {
   this.headFileTopContent = '';
   this.headings = {};
   this.headingIndexingLevel = pageConfig.headingIndexingLevel;
-  this.includedFiles = {};
+  this.includedFiles = new Set();
   this.keywords = {};
   this.navigableHeadings = {};
 }
@@ -484,7 +484,7 @@ Page.prototype.addAnchors = function (content) {
  */
 Page.prototype.collectIncludedFiles = function (dependencies) {
   dependencies.forEach((dependency) => {
-    this.includedFiles[dependency.to] = true;
+    this.includedFiles.add(dependency.to);
   });
 };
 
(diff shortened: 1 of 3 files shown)


sijie123:
Superseded by #852. Thanks for picking up this PR!
Closing.

PR #780, opened 2019-03-18
Parser: Fix crash if no href given to anchor or link tag

### LINKED ISSUE #779: anchor tag without href attribute causing error
v1.21.0

Problem: This code (used to create a named anchor in a page) causes MarkBind to throw an error `<a id="example"></a>`

I suspect it is because the tag is missing the `href` attribute.

sijie123:

• [x] Bug fix
Fixes #779 

As per #779, markbind crashes if no href is given to an anchor or link tag. This is because Markbind will try to rewrite a non-existent URL.

This commit fixes the bug by introducing an extra check for the href attribute. Only if the href attribute is present, do we proceed to rewrite URL to their absolute paths.
I have also included some additional functional tests to protect against regression for this bug.

Are there any more such cases? img and pic are okay because they must contain a src tag. Otherwise markbind will raise an intended error.

**Testing instructions:**
Create a md file containing
```
<a>Some anchor</a>
```
and ensure that markbind does not crash when processing this file.

### DIFF
--- a/test/functional/test_site/sub_site/testReuse.md
+++ b/test/functional/test_site/sub_site/testReuse.md
@@ -19,5 +19,8 @@ PIC tags:
 <pic src="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png"></pic>
 <pic src="https://dummyimage.com/600x400/000/fff"></pic>
 
-Anchor link:
+Anchor:
 <a href="https://dummyimage.com/600x400/000/fff">External Image</a>
+<a href="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png">Link to picture</a>
+<a id="namedAnchor">Named Anchor</a>
+<a>Anchor with no attributes</a>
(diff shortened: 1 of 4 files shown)


damithc:
Guys, let's do a quick fix for this as some of the existing sites don't work due to this issue.

test/functional/test_site/sub_site/testReuse.md
```diff
@@ -19,5 +19,8 @@ PIC tags:
...
 <pic src="https://dummyimage.com/600x400/000/fff"></pic>
 
-Anchor link:
-<a href="https://dummyimage.com/600x400/000/fff">External Image</a>
\ No newline at end of file
+Anchor:
+<a href="https://dummyimage.com/600x400/000/fff">External Image</a>
+<a href="{{baseUrl}}/images/I'm not allowed to use my favorite tool.png">Link to picture</a>
+<a id="namedAnchor">Named Anchor</a>
+<a>Useless anchor</a>
```
Xenonym: Could we have a less emotional name for this? 😅 
`Anchor with no attributes` perhaps?

Issue #506, opened 2018-12-28
Relationship between frontmatter and site.json


damithc:
1. Is it correct to say that anything that can be specified in frontmatter can be specified in site.json?
2. If yes, any property in frontmatter can be _overridden_ by site.json?

Chng-Zhi-Xuan:
1. Yes, currently each page's variables are initialised with values provided within `site.json`

2. It is possible with some modifications to the code base. Currently the values are stored in different places.

The ones in `site.json`'s `pages` object are direct variables in a page (`page.variable`) while the ones in `frontmatter` are stored in the `frontmatter` object of a page (`page.frontmatter.variable`).

So currently there are no overlaps. Although the components I have coded before (siteNav and footer), do not check if it was specified directly in a page variable, but only in `page.frontmatter`.

damithc:
Thanks @Chng-Zhi-Xuan 

> any property in frontmatter can be overridden by site.json?

Does it make sense to go in that direction? It certainly makes things easier to explain to the user. i.e., we have a bunch of _page properties_ that can be specified in frontmatter and overridden in site.json.

What do you all think?

yamgent:
Standardizing it such that all properties can be overridden by `site.json` is a good idea, it makes the page reusable by many different projects.

damithc:
Created #647

PR #784, opened 2019-03-20
Allow any frontmatter property to be overridden by site.json

### LINKED ISSUE #647: Allow any frontmatter property to be overridden by site.json
Current: some frontmatter properties can be overridden by site.json, but not all.

Suggestion: allow any frontmatter property to be overridden by site.json

See discussion in #506

jamos-tay:

Fixes #647 

We want to be able to globally override all frontmatter properties in a page

Allow any property in the frontmatter to be overriden by `globalOverride` if specified in the site.json,

```
{
  "globalOverride": {
      "arbitraryProperty": "..."
  }
}
```

**Proposed commit message:**
```
Allow any frontmatter property to be overridden by site.json

Some front matter properties can be overridden by site.json, but not 
all.

Let's allow any front matter property to be overriden by specifying 
them at the site.json level. Additionally, let's allow the user to 
define global front matter properties that will be applied to all 
pages.
```

### DIFF
--- a/docs/userGuide/siteConfiguration.md
+++ b/docs/userGuide/siteConfiguration.md
@@ -106,7 +106,7 @@ _(Optional)_ **The theme for the generated site.** Uses the default Bootstrap th
 * **`layout`**: The [layout](tweakingThePageStructure.html#page-layouts) to be used by the page. Default: `default`.
 * **`searchable`**: Specifies that the page(s) should be excluded from searching. Default: `yes`.
 * **`externalScripts`**: An array of external scripts to be referenced on the page. Scripts referenced will be run before the layout script.
-* **`frontMatter`**: Specifies properties to add to the front matter of a page or glob of pages. Overrides any existing properties if they have the same name.
+* **`frontMatter`**: Specifies properties to add to the front matter of a page or glob of pages. Overrides any existing properties if they have the same name, and overrides any front matter properties specified in `globalOverride`.
 
 <span id="page-property-overriding">
 <box type="warning">
@@ -161,7 +161,7 @@ The following properties will apply to `index.md`:
 
 #### **`globalOverride`**
 
-**Globally overrides properties in the front matter of all pages.** Any property included in the global override will automatically be merged with the front matter of every single page, and override them if the property exists. Also overrides front matter properties specified in `pages`.
+**Globally overrides properties in the front matter of all pages.** Any property included in the global override will automatically be merged with the front matter of every single page, and override them if the property exists.
 
 #### **`ignore`**
 
--- a/docs/userGuide/siteConfiguration.md
+++ b/docs/userGuide/siteConfiguration.md
@@ -29,7 +29,10 @@ Here is a typical `site.json` file:
       "searchable": "no",
       "externalScripts": [
         "https://cdn.plot.ly/plotly-latest.min.js"
-      ]
+      ],
+      "frontmatter": {
+        "header": "header.md"
+      }
     },
     {
       "glob": "**/index.md"
@@ -42,6 +45,9 @@ Here is a typical `site.json` file:
     "message": "Site Update.",
     "repo": "https://github.com/myorg/myrepo.git",
     "branch": "gh-pages"
+  },
+  "globalOverride": {
+    "footer": "my-footer.md"
   },
   "ignore": [
     "_site/*",
@@ -100,6 +106,7 @@ _(Optional)_ **The theme for the generated site.** Uses the default Bootstrap th
 * **`layout`**: The [layout](tweakingThePageStructure.html#page-layouts) to be used by the page. Default: `default`.
 * **`searchable`**: Specifies that the page(s) should be excluded from searching. Default: `yes`.
 * **`externalScripts`**: An array of external scripts to be referenced on the page. Scripts referenced will be run before the layout script.
+* **`frontMatter`**: Specifies properties to add to the front matter of a page or glob of pages. Overrides any existing properties if they have the same name.
 
 <span id="page-property-overriding">
 <box type="warning">
@@ -151,6 +158,10 @@ The following properties will apply to `index.md`:
 #### **`externalScripts`**
 
 **An array of external scripts to be referenced on all pages.** To reference an external script only on specific pages, `externalScripts` should be specified in `pages` instead. Scripts referenced will be run before the layout script.
+
+#### **`globalOverride`**
+
+**Globally overrides properties in the front matter of all pages.** Any property included in the global override will automatically be merged with the front matter of every single page, and override them if the property exists. Also overrides front matter properties specified in `pages`.
 
 #### **`ignore`**
 
--- a/src/Page.js
+++ b/src/Page.js
@@ -56,10 +56,12 @@ function Page(pageConfig) {
   this.baseUrlMap = pageConfig.baseUrlMap;
   this.content = pageConfig.content || '';
   this.faviconUrl = pageConfig.faviconUrl;
+  this.frontmatterOverride = pageConfig.frontmatter || {};
   this.layout = pageConfig.layout;
   this.layoutsAssetPath = pageConfig.layoutsAssetPath;
   this.rootPath = pageConfig.rootPath;
   this.enableSearch = pageConfig.enableSearch;
+  this.globalOverride = pageConfig.globalOverride;
   this.plugins = pageConfig.plugins;
   this.pluginsContext = pageConfig.pluginsContext;
   this.searchable = pageConfig.searchable;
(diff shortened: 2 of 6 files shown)


jamos-tay:
My bad, I can repush with a better message

"Allow global override of frontmatter properties"?

yamgent:
My apologies, maybe I wasn't clear, but when I mean a merge commit message, it is [something like this](https://github.com/MarkBind/markbind/pull/743#issuecomment-469961993).

Basically it is a new format that we are adopting to explain the context of the change. The first paragraph is usually the current situation, the second paragraph explains why the current situation needs changes, and the third paragraph introduces our changes.

If you are strapped for time, just tell me here and I will try to come up one for you.

jamos-tay:
Sure, no problem

```
Global override for frontmatter properties #784

Currently only certain properties such as `layout` in the frontmatter can be overridden by the site.json, but not properties such as `footer`.

We would like to allow frontmatter properties on every page to be globally overridden to give the author more flexibility. This also allows us to arbitrarily add properties to every page.

We introduce a `globalOverride` property in the site.json which is merged with every page's frontmatter, and overrides any similar properties.
```

jamos-tay:
Rebased and updated

yamgent:
Thanks for the commit message! I tidy up a bit to suit our convention better:

```
Add global override for frontmatter properties (#784)

Certain properties such as `layout` in the frontmatter can be overridden
by the site.json, but not properties such as `footer`. Overriding
properties also apply to only a single `page` or `glob` entry.

We would like to allow all frontmatter properties on every page to be 
globally overridden to give the author more flexibility. By providing
it as a  global option, this also allows us to add the same set of
properties and their values to every page.

Let's introduce a `globalOverride` property in the site.json which is 
merged with every page's frontmatter, and overrides any similar 
properties.
```

Rough description on changes made:
* Message should be wrapped at 72 characters.
* Title should start with imperative verbs like "Add".
* Deleted "Currently" for the first paragraph - it is implied.
* Added "`Overriding
properties also apply to only a single `page` or `glob` entry.`" and rephrased the rest accordingly.
* Change "We" to "Let's" for the last paragraph (this is where the commit comes in and "takes action").

----

On a side note,

> Current: some frontmatter properties can be overridden by site.json, but not all.
> 
> Suggestion: allow any frontmatter property to be overridden by site.json
> 
> See discussion in #506

I don't think the original post expected it to be globally overridden, but rather that the individual `page` and `glob` entries can also override the properties. I know the subsequent posts did cover about global overrides, but unless I missed some discussion about that, I feel that the globals are a nice complement feature but not the full solution, so the spirit of the original post isn't really completely fulfilled by this PR?

damithc:
> I don't think the original post expected it to be globally overridden, but rather that the individual `page` and `glob` entries can also override the properties. I know the subsequent posts did cover about global overrides, but unless I missed some discussion about that, I feel that the globals are a nice complement feature but not the full solution, so the spirit of the original post isn't really completely fulfilled by this PR?

yes, that's correct. The original expectation is to be able to override at src/glob level, not globally.

jamos-tay:
So we want something like this?

```
    {
      "glob": "**/index.md"
      "someProperty": ...
    }
```

damithc:
> So we want something like this?
> 
> ```
>     {
>       "glob": "**/index.md"
>       "someProperty": ...
>     }
> ```

Yes 👍

jamos-tay:
Updated

User can now specify a `frontmatter` property in pages:
```
    {
      "src": "index.md",
      "title": "Hello World",
      "frontmatter": {
        "property": "val"
      }
    },
``` 

`globalOverride` will override pages, pages will override frontmatter

damithc:
> User can now specify a `frontmatter` property in pages:

Nice. Ensure that if a property is specified multiple times (e.g., via multiple globs) the latest value takes precedence, based on the order given in the site.json.

> `globalOverride` will override pages, pages will override frontmatter

An alternative is to rename this as `defaultFrontmatter` and let them be overridden by values specified at page level (but in site.json). That way, the user can specify the typical values as defaults and override only where some pages needs to differ. But in any case, site.json should override frontmatter in individual files, to comply with our top-down overriding policy. What do you think?

src/Page.js
```diff
@@ -516,9 +518,13 @@ Page.prototype.collectFrontMatter = function (includedPage) {
   } else {
     // Page is addressable but no front matter specified
     this.frontMatter = {
-      src: this.src,
-      title: this.title || '',
-      layout: LAYOUT_DEFAULT_NAME,
+      ...{
```
nicholaschuayunzhi: Is there a reason why this is in an object? 
Perhaps we could do

```js
const defaultAttributes = {
        src: this.src,
        title: this.title || '',
        layout: LAYOUT_DEFAULT_NAME,
}

this.frontMatter = {
    ...defaultAttributes 
```

Additionally, this could collapse with the logic above? the title check and layout check seems redundant there.

src/Page.js
```diff
@@ -516,9 +518,13 @@ Page.prototype.collectFrontMatter = function (includedPage) {
...
+        src: this.src,
+        title: this.title || '',
+        layout: LAYOUT_DEFAULT_NAME,
+      },
+      ...this.frontmatterOverride,
+      ...this.globalOverride,
```
nicholaschuayunzhi: Here value specified by `site.json`, namely `this.title` would be overriden by `frontmatterOverride` and `globalOverride` if they specified a title. Would this be a concern?

src/Page.js
nicholaschuayunzhi: On second thought perhaps we could do something like
```js
let front matter = { default }
if (specified front matter exists)
    front matter = specified front matter

// handle overrides
```

jamos-tay: Hmm this is a bit tricky because the default src and title actually takes priority over the specified front matter, we only use the specified one if they are undefined.

Overriding with ... seems to include undefineds:

```
a = { x: 5 }
b = { x: undefined }
{...a, ...b } // { x: undefined }
```

src/Page.js
jamos-tay: Probably not, it was discussed [here](https://github.com/MarkBind/markbind/issues/647#issuecomment-459240830) that it's easier to keep in consistent than specially handle it

jamos-tay:
> > User can now specify a `frontmatter` property in pages:
> 
> Nice. Ensure that if a property is specified multiple times (e.g., via multiple globs) the latest value takes precedence, based on the order given in the site.json.
> 
> > `globalOverride` will override pages, pages will override frontmatter
> 
> An alternative is to rename this as `defaultFrontmatter` and let them be overridden by values specified at page level (but in site.json). That way, the user can specify the typical values as defaults and override only where some pages needs to differ. But in any case, site.json should override frontmatter in individual files, to comply with our top-down overriding policy. What do you think?

Sure, makes sense. It's not too much trouble to swap the priority

src/Page.js
nicholaschuayunzhi: Ah i see, then line 515 and 517 should be done before 512? Otherwise it would override `frontmatterOverride` and `globalOverride` which is inconsistent when front matter is not specified.

src/Page.js
```diff
@@ -507,19 +509,25 @@ Page.prototype.collectFrontMatter = function (includedPage) {
...
   } else {
-    // Page is addressable but no front matter specified
-    this.frontMatter = {
+    const defaultAttributes = {
       src: this.src,
       title: this.title || '',
       layout: LAYOUT_DEFAULT_NAME,
     };
+
+    // Page is addressable but no front matter specified
```
acjh: This comment describes the entire `else` block.

src/Page.js
jamos-tay: Sure, makes sense to move it down for consistency

jamos-tay:
Sorry for the late update, updated and rebased

docs/userGuide/siteConfiguration.md
```diff
@@ -29,7 +29,10 @@ Here is a typical `site.json` file:
       "searchable": "no",
       "externalScripts": [
         "https://cdn.plot.ly/plotly-latest.min.js"
-      ]
+      ],
+      "frontmatter": {
+        "header": "header.md"
+      }
```
yamgent: This part is in the wrong place? Should be inside the specified page.

docs/userGuide/siteConfiguration.md
```diff
@@ -152,6 +159,10 @@ The following properties will apply to `index.md`:
 
 **An array of external scripts to be referenced on all pages.** To reference an external script only on specific pages, `externalScripts` should be specified in `pages` instead. Scripts referenced will be run before the layout script.
 
+#### **`globalOverride`**
+
+**Globally overrides properties in the front matter of all pages.** Any property included in the global override will automatically be merged with the front matter of every single page, and override them if the property exists. Also overrides front matter properties specified in `pages`.
```
yamgent: This documented behaviour is no longer true? `frontMatter` now has a higher overriding priority than `globalOverride` with the latest implementation.

Also update `frontMatter`'s description accordingly.

docs/userGuide/siteConfiguration.md
jamos-tay: Sorry, I'm not sure I get the issue... It seems correct to me

You mean inside the file? We support specifying them in `pages` as well

yamgent: Sorry, my bad, read this part too quickly and got confused, thought it was specified outside of `pages`...

PR #760, opened 2019-03-09
Fix Anti-FOUC styles for Dropdowns

marvinchin:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [x] Bug fix

Fixes #704.

**What is the rationale for this request?**

There were some errors with the dropdown anti-FOUC styles:
- [x] Generation of placeholder span elements for dropdowns produce <span class="undefined temp-dropdown-placeholder">undefined</span> when the dropdown does not have a class and text. 
- [x] Stray `<p>` tags being inserted before the placeholder <span>.
- [x] Anti-FOUC styles not removed upon render for collapsed components.

**What changes did you make? (Give an overview)**

- Handled cases for dropdowns with no class or text attributes
- Changed placeholder element from `span` to `div` to avoid `markdown-it` recognising it as a new paragraph
- Remove anti-FOUC style earlier in the Vue lifecycle (from `mounted` to `created`) so that temporary styles are removed before any other manipulation to the DOM.

**Provide some example code that this change will affect:**

Given Input:
```html
<dropdown></dropdown>
```

Previously:
```html
<span class="undefined temp-dropdown-placeholder">undefined</span>
<dropdown class="temp-dropdown"></dropdown>
```

Now:
```html
<div class="temp-dropdown-placeholder"></div>
<dropdown class="temp-dropdown"></dropdown>
```

**Testing instructions:**
- Dropdown-related pages on netlify preview should be correct: [#1](https://deploy-preview-760--markbind-master.netlify.com/userguide/usingcomponents#dropdowns) [#2](https://deploy-preview-760--markbind-master.netlify.com/userguide/fullsyntaxreference#dropdowns) [#3](https://deploy-preview-760--markbind-master.netlify.com/userguide/readerfacingfeatures#dropdowns)
- Inspecting the HTML output generated for `<dropdowns>` should reflect the above changes. Sample page: `usingComponents.html`

Thanks again to @Xenonym for identifying the issue.

PR #785, opened 2019-03-20
Add algolia-no-index class for hidden content

### LINKED ISSUE #773: Provide an easy way to exclude hidden content from algolia search index
I have managed to try algolia integration with a non-trivial site, https://nus-te3201.github.io/2019/index.html

After some tweaks to the [config file](https://github.com/algolia/docsearch-configs/blob/master/configs/te3201_2019.json), with the help of the Algolia team, the search results are nicely categorized into Admin Info, SE Textbook, Programming Textbook.

The next challenge is to prevent hidden content (e.g., modals, popovers, unselected tabs, unexpanded panels etc.) from being indexed as they cannot be reached by clicking on a search result. Algolia provides a [`selectors_exclude`](https://community.algolia.com/docsearch/config-file.html#selectors_exclude-optional) mechanism for that but we need a way to specify those hidden content.

Is there anything we can do to make it easier? e.g., add a unique class to all such content so that they can be specified in the algolia config file easily?

marvinchin:

Resolves #773.

Note: This PR is rebased on top of #760 as it requires the fix to correctly generate output for `dropdown` components. Will drop this commit once the PR has been merged.

As MarkBind sites are rendered on the client side, the initial HTML includes content that will be hidden to the reader after Vue manipulates the DOM.

The DocSearch crawler indexes all content in this initial HTML, including content that is hidden to the reader. This might be confusing for the reader as he might not be able to find the indexed content immediately on the page.

Let's add the `algolia-no-index` class to content that is known to be hidden by VueStrap components. Authors can then add `algolia-no-index` to the` selectors_exclude attribute` for the site's DocSearch configuration to exclude these content from being indexed.

Input:
```html
<dropdown text="Dropdown">
  <li><a class="dropdown-item" href="/">One</a></li>
  <li><a class="dropdown-item" href="/">Two</a></li>
</dropdown>
```

Output:
```html
<dropdown text="Dropdown" class="algolia-no-index">
  <li><a class="dropdown-item" href="/">One</a></li>
  <li><a class="dropdown-item" href="/">Two</a></li>
</dropdown>
```

- Are there any other components that have content that should not be indexed?

**Testing instructions:**
HTML files generated by MarkBind should have the `algolia-no-index` class added to hidden content in: `dropdown`, `modal`, `panel`, `popover`, `tooltip`, `tabs` components. (see relevant section in the index page of test_site)


test/functional/test_site/expected/index.html
```diff
@@ -596,5 +673,18 @@ <h1 id="heading-in-footer-should-not-be-indexed">Heading in footer should not be
...
+    function initAlgolia() {
+      docsearch({
+        apiKey: "TEST_API_KEY",
+        indexName: "TEST_INDEX_NAME",
+        inputSelector: "#algolia-search-input",
+        algoliaOptions: {},
+        debug: false,
+      });
+    }
+    MarkBind.afterSetup(initAlgolia);
```
Xenonym: Not sure it's a good idea to make API calls to Agolia with invalid credentials when tests are run. Is there a way to test _only_ the `agolia-no-index` class _without_ starting the Agolia script?

damithc:
Suggestion for testing:
* fork se-edu/se-book
* generate site using this feature
* PR the generated site against the gh-pages branch

I'll merge the code to gh-pages branch and update the algolia config file to ignore the hidden content.

test/functional/test_site/expected/index.html
marvinchin: Unfortunately, I can't think of a way to execute only the `postRender` method without executing the rest of the plugin.

Alternatively, we can not do integration tests, and do only unit tests for this plugin - but I'm not sure if this is a good idea either.

marvinchin: On further thought, we can use the demo `apiKey` and `indexName` provided in the DocSearch documentation.

damithc:
> I'll merge the code to gh-pages branch and update the algolia config file to ignore the hidden content.

Done. A quick check suggests the technique works. You can test the [se-book](https://se-education.org/se-book) a bit more to ensure all hidden content are excluded from the index.

marvinchin:
Thanks for the confirmation 🙂 I've taken a look at it and the classes seem to be applied to the correct places.

marvinchin:
Thanks for pointing it out! I've moved it out to a separate test site.

damithc:
@marvinchin Does this feature distinguish between expanded and collapsed panels?

marvinchin:
Yes, expanded panels will not have the ’algolia-no-index’ class added to it, but collapsed panels will.

damithc:
> Yes, expanded panels will not have the ’algolia-no-index’ class added to it, but collapsed panels will.

Nice. We should probably update the documentation about this feature? i.e., how to fine-tune algolia search to avoid indexing hidden content?

marvinchin:
I've added a box mentioning the adding `algolia-no-index` to `selectors_exclude` in the [algolia section](https://markbind.org/userGuide/usingPlugins.html#algolia-enabling-algolia-docsearch).

Do you feel this is sufficient? Or is there more information that I should add?

damithc:
> Do you feel this is sufficient? Or is there more information that I should add?

Ah. That should be enough. 👍

damithc:
@marvinchin another question: Does this exclude tool-tipped keywords from the search index? Ideally, the tooltip text should not be indexed (as it is hidden) but the keyword attached to the tooltip should still be indexed. Same for popovers and modals.

marvinchin:
At the moment the content attached to tooltips and popovers are not indexed (modals are fine). I'll make a separate PR to change this.

PR #796, opened 2019-03-28
Add proposed commit message section to PR Template

### LINKED ISSUE #794: Add Proposed Commit Message section to PR Template
It seems like having the PR author propose a commit message has become part of our development workflow. At the moment, this is often requested in the comments of the PR. 

Perhaps we can include a section for the proposed commit message in the template to remind everyone to provide a commit message when they make the pull request 🙂

marvinchin:

• [x] Other, please explain: GitHub Workflow Update

Resolves #794.

See #794.

Add proposed commit message section to PR template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -37,3 +37,16 @@
 
 
 **Testing instructions:**
+
+
+**Proposed commit message: (wrap lines at 72 characters)**
+
+<!--
+    See this link for more info on how to write a good commit message:
+    https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message
+-->
+
+<!--
+    Some of the responses that you gave to the previous questions might
+    provide you with the information needed to craft your commit message.
+-->


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -37,3 +37,6 @@
 
 
 **Testing instructions:**
+
+
+**Proposed commit message: (wrap lines at 72 characters)**
```
yamgent: Add a comment to mention the rough structure of commit messages (which is stated in the oss-generic documentation):

```html
<!-- 
    See this link for more info on how to write a good commit message:

    https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message 
-->
```

yamgent: Also actually, some of the existing questions in the template could be used to craft the commit message.

In fact there will likely be a repeat of content across the existing questions and the commit message itself. However, I am not sure whether it is a good idea to remove the questions, because it serves as additional guidance for crafting good commit messages.

So, instead of removing the questions, it would be sufficient if we just put this in:

```html
<!--
    Some of the responses that you gave to the previous questions might
    provide you with the information needed to craft your commit message.
-->
```

marvinchin:
@yamgent I've added the comments!

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -37,3 +37,16 @@
...
+
+<!-- 
+    See this link for more info on how to write a good commit message:
+    https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message 
+-->
+
+<!--
+    Some of the responses that you gave to the previous questions might
+    provide you with the information needed to craft your commit message.
+-->
```
yamgent: Need a newline here.

marvinchin: Fixed, sorry about that! 😔

PR #801, opened 2019-04-01
Fix site nav padding

### LINKED ISSUE #800: Site nav padding causes title and scrollbar to be wrongly indented

* **MarkBind Version:** 2.0.0

The title and scrollbar should be correctly positioned.

Screenshot 2019-04-01 at 4 24 23 PM

- The title is indented too far to the right
- The scrollbar is indented too far to the left

marvinchin:
The padding in the site nav is causing the title to be indented further than the list of contents, as well as creating a gap between the scrollbar and the right side of the site nav.

Let's change the padding to fix these issues.

• [x] Bug fix

Fixes #800.

The title in the site nav is misindented, and the scroll bar is mispositioned.

The padding was placed on the outer `<nav>` element, which causes the padding to be applied to both the scrollbar and title. 

Screenshot 2019-04-01 at 4 34 22 PM

- Remove padding from the `<nav>` element and move it to the inner div containing the scrollbar. This way, padding is not applied when calculating scrollbar position and consistent adding is applied to all content within the site nav, including title and content
- Update styles for `site-nav-list` to maintain consistent spacing with other content outside the list
  - Remove `margin-left: -1em` from `.site-nav-list`
  - For outermost `ul`, replace the `px-3` class with `px-0` so that the outermost list uses padding from the parent div
  - For nested `ul`, add `pl-3` to reduce indentation of nested lists (without this, the indentation looks a little too big)
- Remove `indented` class from the title of the `userGuideSections.md` and `devGuideSections.md`

Screenshot 2019-04-01 at 4 37 38 PM

**Testing instructions:**

In the Netlify preview titles in the site nav should appear correctly indented, and scrollbars should be correctly positioned.

Fix site nav padding

The padding in the site nav causes the title to be indented further
right than the list of contents, and creates a gap between the
scrollbar and the right side of the site nav.

Let's move the padding to the inner div that wraps the content
so that the padding is not applied for the scrollbar, and applied
for all content within the site nav.

Let's also update the bootstrap spacing classes for the list
elements inside the site nav to ensure that it is consistent
with the padding of other content in the site nav.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -587,13 +587,14 @@ Page.prototype.insertSiteNav = function (pageData) {
     const siteNavHtml = md.render(siteNavDataSelector('navigation').html().trim().replace(/\n\s*\n/g, '\n'));
     // Add Bootstrap padding class to rendered unordered list
     const siteNavHtmlSelector = cheerio.load(siteNavHtml, { xmlMode: false });
-    siteNavHtmlSelector('ul').first().addClass('px-3');
+    siteNavHtmlSelector('ul').first().addClass('px-0');
+    siteNavHtmlSelector('ul ul').addClass('pl-3');
     const formattedSiteNav = formatSiteNav(siteNavHtmlSelector.html(), this.src);
     siteNavDataSelector('navigation').replaceWith(formattedSiteNav);
   }
   // Wrap sections
   const wrappedSiteNav = `<nav id="${SITE_NAV_ID}" class="navbar navbar-light bg-transparent">\n`
-    + '<div class="sticky-top spacer-top viewport-height-90 scrollable slim-scroll">'
+    + '<div class="sticky-top site-nav-spacer viewport-height-90 scrollable slim-scroll">'
     + `${siteNavDataSelector.html()}\n`
     + '</div>\n'
     + '</nav>';
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -115,8 +115,11 @@ footer {
 
 /* Side Navigation menus */
 
-.spacer-top {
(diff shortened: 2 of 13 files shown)


src/Page.js
```diff
@@ -587,13 +587,14 @@ Page.prototype.insertSiteNav = function (pageData) {
     const siteNavHtml = md.render(siteNavDataSelector('navigation').html().trim().replace(/\n\s*\n/g, '\n'));
     // Add Bootstrap padding class to rendered unordered list
     const siteNavHtmlSelector = cheerio.load(siteNavHtml, { xmlMode: false });
-    siteNavHtmlSelector('ul').first().addClass('px-3');
+    siteNavHtmlSelector('ul').first().addClass('px-0');
+    siteNavHtmlSelector('ul ul').addClass('pl-3');
```
Chng-Zhi-Xuan: Good enhancement to get rid of `margin-left: -1em` from the `site-nav-list` class.

So far I tested with deeply nested and no nesting `site-nav`s and so far works well.

Maybe we should change it to `pl-1` to save some space for text?

marvinchin: Thanks for bringing it up!

Personally, I feel that the indentation for `pl-1` and `pl-2` isn't distinct enough. Here are some images of the different indentation levels for comparison:

`pl-1`:
Screenshot 2019-04-04 at 2 04 37 PM

`pl-2`:
Screenshot 2019-04-04 at 2 04 46 PM

`pl-3`:
Screenshot 2019-04-04 at 2 04 54 PM

Chng-Zhi-Xuan: If there are no more changes, you can remove [WIP] from the PR title.

marvinchin: Done!

PR #804, opened 2019-04-03
Parser: Fix relative src bug involving hostBaseUrl

### LINKED ISSUE #803: {{ baseUrl }} not working with <img> tag when included via div id
v2.1.0

I have the following code in a subsite ``

This used to work before, but now the link is broken.

Output https://nus-te3201.github.io/2019/admin/te3201-admin.html

image

Upon further investigation, it seems the img URL is not working if it is in a file that is included using a div id e.g., 
`<include src="../book/about/usage.md#saving" />`:  Doesn't work
`<include src="../book/about/usage.md" />`:  Works

sijie123:

• [ ] Documentation update
• [x] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Fixes #803 

When including a file from another site, the {{baseUrl}} variables
in src attributes are sometimes rewritten as {{hostBaseUrl}}.
{{hostBaseUrl}} also refers to an absolute path, so we should not
treat it as a relative path and rewrite the URL.

This {{hostBaseUrl}} issue arises when we have a file including content using div ID references from another file within a sub-site.
For example, index.md includes sub_site/inner.md#imageDiv. Then if inner.md#imageDiv references any local resources within the subsite with {{baseUrl}}, this {{baseUrl}} will be rewritten to {{hostBaseUrl}}/sub_site.

Let's teach parser to avoid rewriting URLs that contain
{{hostBaseUrl}}.

See TE3201 website linked from #803 

**Testing instructions:**
Check out TE3201 github page, and build the website using MarkBind from this PR. You can verify that the `chromeSaveAsPdf` images displays correctly.

### DIFF
--- a/test/functional/test_site/expected/sub_site/testReuse._include_.html
+++ b/test/functional/test_site/expected/sub_site/testReuse._include_.html
@@ -19,3 +19,8 @@
   <a href="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">Link to picture</a>
   <a id="namedAnchor">Named Anchor</a>
   <a>Anchor with no attributes</a></p>
+<p>Within DIV tag:</p>
+<div id="imageTest">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+</div>
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -187,7 +187,9 @@ Parser.prototype._preprocess = function (node, context, config) {
   let actualFilePath;
   if (hasSrc && shouldProcessSrc) {
     isUrl = utils.isUrl(element.attribs.src);
-    isAbsolutePath = path.isAbsolute(element.attribs.src) || element.attribs.src.includes('{{baseUrl}}');
+    isAbsolutePath = path.isAbsolute(element.attribs.src)
+                         || element.attribs.src.includes('{{baseUrl}}')
+                         || element.attribs.src.includes('{{hostBaseUrl}}');
     includeSrc = url.parse(element.attribs.src);
     filePath = isUrl
       ? element.attribs.src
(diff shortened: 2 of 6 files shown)


yamgent:
> So, I would need help with figuring out why {{baseUrl}} in the included image was rewritten to be {{hostBaseUrl}}.

If I recall correctly, someone told me that `hostBaseUrl` was needed because sub sites do not have access to the `baseUrl` of the actual site, hence all `baseUrl` needed to be rewritten as `hostBaseUrl` first, before it became available.

You probably would be able to replicate the problem if your webpage is inside `test/functional/test_site/sub_site/`.

sijie123:
Thanks @yamgent.
The issue only arises when we have a file including content using div ID references from another file within a sub-site.
For example, `index.md` includes `sub_site/inner.md#imageDiv`. Then if `inner.md` references any local resources within the subsite with `{{baseUrl}}`, this `{{baseUrl}}` will be rewritten to `{{hostBaseUrl}}`.

I've updated the PR with the fix and test to protect against similar regression.

test/functional/test_site/expected/sub_site/testReuse._include_.html
```diff
@@ -18,4 +18,9 @@
...
   <a href="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">Link to picture</a>
   <a id="namedAnchor">Named Anchor</a>
-  <a>Anchor with no attributes</a></p>
\ No newline at end of file
+  <a>Anchor with no attributes</a></p>
+<p>Within DIV tag:</p>
+<div id="imageTest">
+  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <pic src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png"></pic>
+</div>
```
marvinchin: I think a newline should be placed at the end of this file 🙂

Might be useful to see if your IDE has the option to automatically add new lines at the end of the file when necessary!

sijie123: Thanks for the review!
Actually this file is the auto generated expected file, so having a newline will cause the test to fail...

marvinchin: Ah right, that makes sense! Thanks for pointing it out 🙂

acjh: We should add it during generation then. 🙊

damithc:
Guys, need a hotfix soon as the TE3201 website is affected significantly by this bug.

damithc:
Seems to be fixed now. Thanks all for the quick fix 💯

PR #821, opened 2019-04-07
Change topNav color to match brand color

damithc:
• [x] Documentation update

I tried changing the topNav color to match brand color, mostly to see if it can be done. Not entirely sure if it looks better than the current one. Putting this up just to share with others.

On a related note, we can document how to customize the topNav color once we figure out the best way to do it.

image


yamgent:
Looks a bit too blue for the eyes but that's just my two cents. :P

docs/_markbind/layouts/default/styles.css
```diff
@@ -0,0 +1,3 @@
+.bg-brand {
+    background-color: #00B0F0 !important;
+}
```
yamgent: Need a newline here.

damithc:
> Looks a bit too blue for the eyes but that's just my two cents. :P

If it is about the color itself, we can pick a different color/shade as the brand color. The current brand color is just an arbitrary choice. 

But yes, the buttons and the links are also blue.

The current one (i.e., without the brand color) looks good actually. Doing a good job of 'branding' the website needs more design expertise than what we have I guess :-p Most likely we won't merge this PR. Just sharing only.

damithc:
Closing, as not meant to be merged.

PR #823, opened 2019-04-07
header.md: Use absolute path instead of ../

### LINKED ISSUE #820: Document how to use URLs in _markbind/* files
Previously, we used `{{ baseUrl }}` in `_markbind/*` files. e.g., `_markbind/head/*`. Now, we no longer need to use `{{ baseUrl }}` (is that correct?).

In that case, we need to update the documentation of those files ([example](https://markbind.org/userGuide/tweakingThePageStructure.html)) and also mention what is considered as the 'current directory' when giving URLs in those files.

sijie123:

• [x] Documentation update

Resolves #820 

Header contains link references such as ../index.html. This is so
that the links are correct when referring both the main site and
from a nested directory such as userGuide/. However, this approach
is incorrect if we have more deeply nested directories.
For a more detailed explanation, refer to https://github.com/MarkBind/markbind/issues/820#issuecomment-480574703

Instead of relative references, let's use absolute references w.r.t.
the root directory instead.

**Testing instructions:**
Open the netlify preview site and ensure that there are no broken links.

### DIFF
--- a/docs/_markbind/headers/header.md
+++ b/docs/_markbind/headers/header.md
@@ -1,12 +1,12 @@
 <header>
-  <link rel="stylesheet" href="../css/main.css">
+  <link rel="stylesheet" href="{{baseUrl}}/css/main.css">
   <navbar type="dark">
-    <a slot="brand" href="../index.html" title="Home" class="navbar-brand"><img src="../images/logo-darkbackground.png" height="20" /></a>
-    <li><a href="../index.html" class="nav-link">HOME</a></li>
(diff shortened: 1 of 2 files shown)


Chng-Zhi-Xuan:
Thanks for the PR @sijie123. Tested and works fine.

However, does this PR resolve #820? Since it doesn't update the User Guide to explicitly inform users to use `{{ baseUrl }}` for files within `_/markbind`. 

Do highlight to me if I am missing some information within the comment chain you linked.

sijie123:
Thanks @Chng-Zhi-Xuan. I've added the update to the documentation.

Issue #810, opened 2019-04-04
Refactor the Parser/Site/Page trinity into more comprehensible components


luyangkenneth:
While doing so, also explore the possibility of parsing .md files only once (see #762 for more context)

In cases where complexity cannot be avoided, make sure to add comments to the relevant components/functions and document any unintuitive behaviour. (Related to #602)

ang-zeyu:
- [x] Parser.js
- [ ] Page.js
- [ ] Site.js

PR #826, opened 2019-04-09
Support context-aware includes for images and anchors

### LINKED ISSUE #822: Markdown syntax for images doesn't work without {{ baseUrl }}
`` doesn't work but `` works.

This is a leftover from previous attempts to remove the need for `{{ baseUrl }}` in intra-site links.

Ideally, this should be fixed as it detracts from an otherwise a good fix for a significant flaw that existed before. Minimally, it should be documented so that users are aware of this gotcha.

sijie123:

Fixes #822 

After our parser includes a file, it's context is discarded.
This means that in subsequent passes of parser, we do not know where
the content has been included from, and included content is treated
uniformly as if they were originally part of the file.

This means that content in an included file must be valid wherever
it is included from. For example, resource references need to be
absolute paths, so that they remain valid wherever they are included
from.

This makes it impossible to support relative src references written using Markdown, as markdown code is only parsed after all the includes have been made.

To support context-based references, i.e. the included segments
retain awareness of their original include locations, let's add
a data-included-from meta whenever we include files. To maintain
the existing behaviour and to prevent leaking file structure data,
this meta tag is removed before we finish generating the final
website.

**Breaking changes**
Previously, some relative URLs are allowed. The reference was always made against files which initiate the include. 
Now, any reference is always made against **files using the relative URLs**.
This should make it more intuitive for authors to reuse content, but may break existing code *especially if they are generated using nunjucks/macros*.

Given this directory structure:
```
/
├ userGuide/
├-- subFolder/
├---- content.md
├-- index.md
├-- tutorial.md
```
Assume that we have a tutorial file that simply collates a list of content:
```markdown
<include src="subFolder/content.md" />
```
And assuming we have the existing content file:
```markdown
Topic One
Please refer to [our user guide here](index.html)
```
The existing content file is correct because the generated tutorial.html file will have a correct relative reference to index.html, the homepage of the user guide.
However, after this PR, the existing content file is incorrect. `index.html` will refer to an index file within subFolder. **All relative references to resources and links are now with respect to the file that writes these references**, so in this case, they are made with respect to `content.md`.
content.md should be updated as
```markdown
Topic One
Please refer to [our user guide here](../index.html)
```

As this PR is significantly breaking and likely to be buggy, please help me test as much as possible and point out any bugs so that I can correct them ASAP.
Sites confirmed working:
[x] Our system test site
[x] CS2103T website
[x] TE3201 website

List of known issues:
All known issues to date are resolved.

Progress:
[x] Finish testing against all existing websites and ensure that there are no bugs
[x] Update documentation to reflect change against relative URL references
[x] Determine reason behind unwrapping div bug. **Update**: bug is due to difference between span and div used for inline and non-inline includes. Bug squashed.
[x] Determine why tests aren't correct - and whether to update tests or to find bug. **Update**: bug squashed.

**Testing instructions:**
Please check out this PR and build it against existing, known good sites and verify that the sites are working as expected.

### DIFF
--- a/docs/userGuide/syntax/links.mbdf
+++ b/docs/userGuide/syntax/links.mbdf
@@ -56,12 +56,14 @@ Links should start with {{ showBaseUrlCode }} (which represents the root directo
 1. <code>Click [here]({{ showBaseUrlCode }}/userGuide/reusingContents.html).</code>
 2. `![](`{{ showBaseUrlCode }}`/images/preview.png)`
 
+<box type="important">
+To ensure that links in the <code>_markbind/</code> folder work correctly across the entire site, they should be written as absolute paths, prepended with <code><span>{</span>{ baseUrl }}</code>. 
+</box>
 </div>
 
 Relative paths:
 
 <div class="indented">
-Links to files can also be specified relative to the file that includes it.
 
 {{ icon_example }} Assuming that we have the following folder structure:
 ```
@@ -83,12 +85,14 @@ Within `index.md`, we can also display the image using
 <img src="textbook/image.png" />
 ```
 or by including `subsite.md`:
-```
+```html
 <include src="textbook/subsite.md" />
 ```
 
 <box type="important">
-To ensure that links in the _markbind/ folder work correctly across the entire site, they should be written as absolute paths, prepended with <span>{</span>{ baseUrl }}. 
+    Relative links to resources (e.g. images, hrefs) should be valid relative to the original, included file. In other words, the links should be accessible when traversing starting from the location of the included file.
+    <br>
+    In the example above, <code>image.png</code> is in the same directory as <code>subsite.md</code>. When using relative references, the correct path is <code>image.png</code> and not <code>textbook/image.png</code>.
 </box>
 
 </div>
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -172,24 +172,20 @@ Parser.prototype._preprocess = function (node, context, config) {
   element.attribs = element.attribs || {};
   element.attribs[ATTRIB_CWF] = path.resolve(context.cwf);
 
-  const requiresSrc = ['img', 'pic', 'include'].includes(element.name);
+  const requiresSrc = ['include'].includes(element.name);
   if (requiresSrc && _.isEmpty(element.attribs.src)) {
     const error = new Error(`Empty src attribute in ${element.name} in: ${element.attribs[ATTRIB_CWF]}`);
     this._onError(error);
     return createErrorNode(element, error);
   }
-  const shouldProcessSrc = ['img', 'pic', 'include', 'panel'].includes(element.name);
+  const shouldProcessSrc = ['include', 'panel'].includes(element.name);
   const hasSrc = _.hasIn(element.attribs, 'src');
   let isUrl;
   let includeSrc;
   let filePath;
-  let isAbsolutePath;
   let actualFilePath;
   if (hasSrc && shouldProcessSrc) {
     isUrl = utils.isUrl(element.attribs.src);
-    isAbsolutePath = path.isAbsolute(element.attribs.src)
-                         || element.attribs.src.includes('{{baseUrl}}')
-                         || element.attribs.src.includes('{{hostBaseUrl}}');
     includeSrc = url.parse(element.attribs.src);
     filePath = isUrl
       ? element.attribs.src
@@ -202,7 +198,7 @@ Parser.prototype._preprocess = function (node, context, config) {
       this.boilerplateIncludeSrc.push({ from: context.cwf, to: actualFilePath });
     }
     const isOptional = element.name === 'include' && _.hasIn(element.attribs, 'optional');
-    if (!['img', 'pic'].includes(element.name) && !utils.fileExists(actualFilePath)) {
+    if (!utils.fileExists(actualFilePath)) {
       if (isOptional) {
         return createEmptyNode();
       }
@@ -213,18 +209,6 @@ Parser.prototype._preprocess = function (node, context, config) {
       this._onError(error);
       return createErrorNode(element, error);
     }
-  }
-
-  const shouldProcessHref = ['a', 'link'].includes(element.name);
-  const hasHref = _.hasIn(element.attribs, 'href');
-  if (hasHref && shouldProcessHref) {
-    isUrl = utils.isUrl(element.attribs.href);
-    isAbsolutePath = path.isAbsolute(element.attribs.href) || element.attribs.href.startsWith('{{');
-    includeSrc = url.parse(element.attribs.href);
-    filePath = isUrl
-      ? element.attribs.src
-      : path.resolve(path.dirname(context.cwf), decodeURIComponent(includeSrc.path));
-    actualFilePath = filePath;
   }
 
   if (element.name === 'include') {
@@ -391,16 +375,6 @@ Parser.prototype._preprocess = function (node, context, config) {
     if (element.name === 'body') {
       // eslint-disable-next-line no-console
       console.warn(`<body> tag found in ${element.attribs[ATTRIB_CWF]}. This may cause formatting errors.`);
-    } else if (['img', 'pic'].includes(element.name)) {
-      if (!isUrl && !isAbsolutePath) {
-        const resultPath = path.join('{{hostBaseUrl}}', path.relative(config.rootPath, filePath));
-        element.attribs.src = utils.ensurePosix(resultPath);
-      }
-    } else if (['a', 'link'].includes(element.name)) {
-      if (!isUrl && !isAbsolutePath && hasHref) {
-        const resultPath = path.join('{{hostBaseUrl}}', path.relative(config.rootPath, filePath));
-        element.attribs.href = utils.ensurePosix(resultPath);
-      }
     }
     if (element.children && element.children.length > 0) {
       element.children = element.children.map(e => self._preprocess(e, context, config));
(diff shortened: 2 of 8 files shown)


sijie123:
@Chng-Zhi-Xuan there were some issues with the commit previously - i've updated and rebased. Please pull the changes again. Thanks!

Chng-Zhi-Xuan:
I will be testing in-depth more tomorrow due to time constraints. 

Not immediately after requesting for review from myself :P

yamgent:
A brief comment from a quick glance... commit 11bb1e80 seems to be doing two different things at the same time (adding test + pushing the parsing of img and anchors to a later stage), is it not possible to split it up?

sijie123:
> A brief comment from a quick glance... commit [11bb1e8](https://github.com/MarkBind/markbind/commit/11bb1e8085fa1781a856cb777a158e3e5ee851f0) seems to be doing two different things at the same time (adding test + pushing the parsing of img and anchors to a later stage), is it not possible to split it up?

Oh no. One of the commits disppeared / got squashed together... I'll fix that up in a bit. Thanks for the heads up.

Chng-Zhi-Xuan:
So I've updated the `test_site` files to pass the tests.

Then I've notice the image links are broken for the 2 lines are are replaced here, in `testReuse.md`:
unexpected-behaviour

So the `src` is wrong and there is a `href` attached to those links. 

Here is the line within `testReuse.md`. Can you take a look into this weird bug?
``` HTML
<pic src="images/I'm not allowed to use my favorite tool.png"></pic>
<pic src="{{imgFolder}}/I'm not allowed to use my favorite tool.png"></pic>
```

test/functional/test_site/expected/sub_site/testReuse._include_.html
```diff
@@ -2,14 +2,14 @@
 <p>Some variables:</p>
 <div></div>
 <p>IMG tags:
-  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
-  <img src="/test_site/sub_site/images/I'm not allowed to use my favorite tool.png">
+  <img src="images/I'm not allowed to use my favorite tool.png">
+  <img src="images/I'm not allowed to use my favorite tool.png">
```
yamgent: > * Tests still fail - testReuse.include.html doesn't have "/test_site/sub_site" anymore, but the output site looks correct. Still yet to determine whether this is an intended behaviour or the result of a bug.

Definitely a bug, this seems like a wrong change, because files that include this may not have the correct path references.

Also the output site doesn't look correct :P. I checked the test_site's `index.html`, the panels that include this HTML have broken images:

broken

sijie123:
Thanks @yamgent. That was dangerous - I had thought that the relative URLs in the testReuse.include.html was something expected from the new behaviour.
Lesson learnt: Always trust your tests :P

I've fixed the oversight. Now, included files (via panels) should be processed correctly too.

src/lib/markbind/src/parser.js
```diff
@@ -404,6 +378,53 @@ Parser.prototype._preprocess = function (node, context, config) {
...
+      // Do not rewrite.
+      return;
+    }
+
+    const firstParent = elem.closest('div[data-included-from]');
+    const originalSrc = firstParent.attr('data-included-from') || context;
+    const srcFolder = path.dirname(originalSrc);
+    const absolutePath = `/${path.relative(self.rootPath, srcFolder)}`;
+    $(this).attr('href',
+                 utils.ensurePosix(path.join('{{baseUrl}}', path.resolve(absolutePath, resourcePath))));
```
yamgent: Should be `{{hostBaseUrl}}`?

test/unit/parser.test.js
```diff
@@ -45,10 +45,11 @@ test('includeFile replaces <include> with <div>', async () => {
 
   const expected = [
     '# Index',
-    `<div cwf="${indexPath}" include-path="${includePath}">`,
+    `<div cwf="${indexPath}" include-path="${includePath}">`
+    + `<div data-included-from="${includePath}" cwf="${includePath}">`,
```
yamgent: Is there a reason why we couldn't piggyback the `cwf` attribute for our implementation of relative paths?

src/lib/markbind/src/parser.js
sijie123: Yes. Will change that.

test/unit/parser.test.js
sijie123: I believe we erase this cwf after we do `resolveBaseUrl`, right before we output the file but before we process any post-process plugins. 
I didn't use the existing cwf attribute mainly because I'm not sure what the cwf contains. In most cases, it looks the same as what I put inside data-included-from, but I'm not too sure.
Using cwf also means that I would have to erase the cwf tag at a later stage, and I'm not sure if I will break any existing implementation by doing that. I don't think so, but I was thinking it might be safer to add data-included-from instead.
As a bonus, being standalone, we know that it would be safe to persist this information till much later in the workflow, at least till the post-process plugins, but potentially till the very end of our process if we want to.

yamgent: I am happy with the current implementation, no need to change. Cleanups can be done in the future once we streamlined the Parser in #810.

sijie123:
> Commit "Parser: Support context-aware includes"
> 
> > ```
> > This means that content in an included file must be valid wherever
> > it is included from. For example, resource references need to be
> > absolute paths, so that they remain valid wherever they are included
> > from.
> > ```
> 
> Um, the England here doesn't make sense? :P Are you trying to say that
> paths need to be valid _regardless of_ wherever it is included from?

I was trying to say something like if we have 2 files from different places, e.g.
/a/index.md and /index.md both includes /a/sub-folder/b.md, any reference in b.md needs to remain valid whether it ends up in /a/index.md or /index.md.
For example, if b.md refers to a css file located in /a/file.css, it cannot refer to it as `file.css`, because this reference is invalid from the perspective of /index.md.

I've updated the commit message and fixed the {{hostBaseUrl}} bug. I still don't really like my commit message, but I can't find a clearer way to express the idea above^

yamgent:
I put in the example that you gave, and also invented the word "host file", see if that makes it clearer. The first two paragraphs are modified.

```
After our parser includes a file, its context is discarded.
This means that in subsequent passes of parser, we do not know where
the content has been included from, and included content is treated
uniformly as if they were originally part of the host file.

The path of the references in the included file may not resolve
correctly if it belongs to a different directory from the host file.
For example, if /a/b/c.md refers to a file in '../file.css', and
/index.md is the host file that includes /a/b/c.md, then /index.md
should resolve this path to '/a/b/file.css'. It cannot retain the
original path '../file.css', as that will lead to the wrong css file.

To support context-aware references, i.e. the included segments
retain awareness of their original include locations, let's add
a data-included-from meta whenever we include files. To maintain
the existing behaviour and to prevent leaking file structure data,
this meta tag is removed before we finish generating the final
website.
```

Chng-Zhi-Xuan:
I just pulled the latest branch. Tested and found this:

Both Pic and Img tags test have a stay `C:/` within the `src` |
--- |
bug

sijie123:
> I just pulled the latest branch. Tested and found this:
> 
> Both Pic and Img tags test have a stay `C:/` within the `src`, which results in a broken link
> bug

Confirmed. I'll just have to add a couple more ensurePosix then :(
On a side note, we really have to get our AppVeyor up and running :P

sijie123:
> I put in the example that you gave, and also invented the word "host file", see if that makes it clearer. The first two paragraphs are modified.
> 
> ```
> After our parser includes a file, its context is discarded.
> This means that in subsequent passes of parser, we do not know where
> the content has been included from, and included content is treated
> uniformly as if they were originally part of the host file.
> 
> The path of the references in the included file may not resolve
> correctly if it belongs to a different directory from the host file.
> For example, if /a/b/c.md refers to a file in '../file.css', and
> /index.md is the host file that includes /a/b/c.md, then /index.md
> should resolve this path to '/a/b/file.css'. It cannot retain the
> original path '../file.css', as that will lead to the wrong css file.
> 
> To support context-aware references, i.e. the included segments
> retain awareness of their original include locations, let's add
> a data-included-from meta whenever we include files. To maintain
> the existing behaviour and to prevent leaking file structure data,
> this meta tag is removed before we finish generating the final
> website.
> ```

Yes! The use of "host file" made it much clearer. Thank you so much - I will adopt this commit message :D

I've also attempted a fix for the Windows bug @Chng-Zhi-Xuan by forcing it to use posix notation. I've tested it on a Windows VM and it looks fine now.

sijie123:
Thanks for reviewing @yamgent. Sorry I had to dismiss your review.
I found an issue with authoringContents.md (and some CS2103T pages) that happened as a result of inline includes.
When we do inline includes, we use a span instead of a div. However, my implementation "broke" inline includes because I always use a div. I have now added checks: if it's an inline include, use a span; otherwise, use a div.
(The reason why I don't use span all the time is because somehow spans introduce extra <p></p> that cause the formatting to go wrong)

I've taken a quick glance at the resulting CS2103T and TE3201 websites, they look more fine now :P

I'll still need to update the documentation to reflect the change in our includes. Will do that tomorrow.

Chng-Zhi-Xuan:
Hi @sijie123 Thanks for updating the PR, and doing additional tests.

It will be good to add a `Pending Action` or `To-Do` list either in the PR description or latest comment to inform reviewers that there are still items being added to the PR. 

Just a minor QoL nit :P

sijie123:
Finally! I think I'm done with everything I set out to do...
I'll do one last test on a Windows VM, but in the meantime it's up for review.
Thanks @yamgent @Chng-Zhi-Xuan for your help!

src/lib/markbind/src/parser.js
```diff
@@ -341,7 +325,11 @@ Parser.prototype._preprocess = function (node, context, config) {
         }
         actualContent = self._rebaseReferenceForStaticIncludes(actualContent, element, config);
       }
-      element.children = cheerio.parseHTML(actualContent, true); // the needed content;
+      const inlineWrapper = isInline ? 'span' : 'div';
```
Chng-Zhi-Xuan: The constant name is not consistent with the behaviour since `isInline` can be false, and the resulting wrapper is meant for non-inline includes.

Shall we rename `inlineWrapper` to `wrapper` or `data-included-from-wrapper` ?

docs/userGuide/syntax/links.mbdf
```diff
@@ -83,12 +85,14 @@ Within `index.md`, we can also display the image using
...
-```
+```html
 <include src="textbook/subsite.md" />
 ```
 
 <box type="important">
-To ensure that links in the _markbind/ folder work correctly across the entire site, they should be written as absolute paths, prepended with <span>{</span>{ baseUrl }}. 
+    Relative links to resources (e.g. images, hrefs) should be specified relative to the original, included file. 
+    <br>
+    In the example above, <code>image.png</code> is in the same directory as <code>subsite.md</code>. When using relative references, the correct path is <code>image.png</code> and not <code>textbook/image.png</code>.
```
Chng-Zhi-Xuan: Should we be more explicit here? 

E.G: When using relative references in `subsite.md` after being included in `index.md`, ...

src/lib/markbind/src/parser.js
sijie123: Sometimes I wonder how confused I must have been what I wrote the code... I don't even know why it's named inlineWrapper haha
I'll change it to something like `wrapperType`. Thanks so much.

docs/userGuide/syntax/links.mbdf
sijie123: Hmm, I intended to say something similar to
"When using relative references in subsite.md, links should be valid from the perspective of subsite.md, regardless of whichever file (such as `index.md`) that eventually includes it."

sijie123:
Updated with @Chng-Zhi-Xuan's comments. Thanks :))

PR #827, opened 2019-04-10
Remove sticky-top class from side navigations

### LINKED ISSUE #818: PageNav appears above search results
v2.1.0

The problem:
image

PageNav elements appear over the search results.

marvinchin:

• [x] Bug fix

Fix #818.

The side navigations have unnecessarily high `z-index`, which causes it to be layered on top of other components.

Remove the `sticky-top` bootstrap class which adds the high `z-index`, and use other css classes to retain the same positioning without increasing `z-index`.

**Testing instructions:**
 Build the TE3201 website with this branch, and navigate to `http://127.0.0.1:8080/se-book-adapted/chapters/refactoring.html#refactoring`. Use the searchbar. The page navigation should not appear above the search results.

Screenshot 2019-04-10 at 12 19 32 PM

Screenshot 2019-04-10 at 12 22 10 PM

Remove sticky-top class from side navigations

The sticky-top class increases the z-index of the side navigations to
1030. The high z-index is not necessary for the side navigations, and
has caused it to be layered on top of the dropdowns of other components.

Let's remove the sticky-top class from the side navigations, and use
other css classes to achieve the same positioning without the high
z-index.

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -131,6 +131,10 @@ footer {
     max-height: 90vh;
 }
 
+.position-top {
+    top: 0;
+}
+
 /* Bootstrap small(sm) responsive breakpoint */
 @media (max-width: 767.98px) {
     .dropdown-menu > li > a {
--- a/src/Page.js
+++ b/src/Page.js
@@ -594,7 +594,7 @@ Page.prototype.insertSiteNav = function (pageData) {
   }
   // Wrap sections
   const wrappedSiteNav = `<nav id="${SITE_NAV_ID}" class="navbar navbar-light bg-transparent">\n`
-    + '<div class="sticky-top site-nav-spacer viewport-height-90 scrollable slim-scroll">'
+    + '<div class="position-sticky position-top site-nav-spacer viewport-height-90 scrollable slim-scroll">'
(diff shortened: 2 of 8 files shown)


Chng-Zhi-Xuan:
Are you planning on reverting the changes to dropdown lists in vue-strap?

If so, it should be included in this PR (updated `vue-strap.min.js`).

marvinchin:
Yes, I'm intending to make those changes!

Re-building and updating `vue-strap` is done in our release procedure. Hence I feel like updating it should be done during release for consistency rather than in this PR (since there seems to be 2 logical parts to this - making changes to the `vue-strap` repo, and then updating the `vue-strap` dependency in the `markbind` repo).

What do you think about it?

Chng-Zhi-Xuan:
For this case, as the changes were minimal, we can do the updating of vue-strap dependency on the next release.

For bigger PRs / features. Usually we require the updated `vue-strap.min.js` file within the PR in order for any feature / tests to work properly within MarkBind. 

It is more for the reviewer's convenience, but not necessary for this PR :P

marvinchin:
Understood, thanks for the clarification! 🙂

PR #831, opened 2019-04-11
DevOps: Use stylelint

### LINKED ISSUE #830: Use Stylelint to lint CSS
At the moment, we don't have a linting tool for CSS. Perhaps we can consider adopting [stylelint](https://stylelint.io) in the project to ensure that we adhere to consistent CSS style rules.

marvinchin:

• [x] Other, please explain: developer workflow

Resolves #830.

We don't currently have a linter for CSS to catch formatting mistakes in our CSS files.

Add Stylelint as a dependency, and add it to our pretest scripts.

- I'm still getting a number of diffs in my `package-lock.json` files, even when using `npm v5.8.0`. Would be good if someone can try doing the install as well and see if they get the same diffs too!
- I'm currently linting **all** CSS files in the repository - is this too broad? (e.g. even in generated CSS files in the expected folder)
- Are the current linting rules agreeable to everybody?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -579,6 +579,22 @@
         "@types/yargs": "12.0.10"
       }
     },
+    "@mrmlnc/readdir-enhanced": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz",
+      "integrity": "sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==",
+      "dev": true,
+      "requires": {
+        "call-me-maybe": "1.0.1",
+        "glob-to-regexp": "0.3.0"
+      }
+    },
+    "@nodelib/fs.stat": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
+      "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==",
+      "dev": true
+    },
     "@sindresorhus/slugify": {
       "version": "0.8.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/slugify/-/slugify-0.8.0.tgz",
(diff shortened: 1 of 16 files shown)


.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "stylelint-config-standard",
+  "rules": {
+      "indentation": 4,
+      "no-empty-source": null
```
acjh: Which files violate this?

marvinchin: Some of the CSS files that MarkBind generates upon initialisation are empty (e.g. `docs/_markbind/layouts/default/styles.css`).

yamgent: Might want to document this case in the commit message so that future developers know why this rule `no-empty-source` was added.

package.json
```diff
@@ -61,9 +61,10 @@
     "url": "https://github.com/MarkBind/markbind.git"
   },
   "scripts": {
+    "csslint": "./node_modules/.bin/stylelint '**/*.css' '!**/*.min.css'",
```
yamgent: This didn't work, I modified `asset/css/markbind.css` to violate some rules (e.g. `0` -> `0px`, `#abcdef` -> `#ABCDEF`), but stylelint didn't complain about anything.

marvinchin: Hmm that's odd, it seems to be working for me:

Screenshot 2019-04-12 at 10 40 16 AM

I'll try to see if I can get hold of a windows machine to replicate this.

yamgent: Tried it on another Windows computer, still doesn't work:

csslint

package-lock.json
```diff
@@ -5333,6 +5889,12 @@
       "resolved": "https://registry.npmjs.org/map-cache/-/map-cache-0.2.2.tgz",
       "integrity": "sha1-wyq9C9ZSXZsFFkW7TyasXcmKDb8="
     },
+    "map-obj": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/map-obj/-/map-obj-2.0.0.tgz",
+      "integrity": "sha1-plzSkIepJZi4eRJXpSPgISIqwfk=",
```
marvinchin: I've reinstalled node and npm, and then followed the instructions described here: https://github.com/npm/npm/issues/16938#issuecomment-339863980

The diffs that change the integrity hash of existing packages are gone, but there still seems to be some sha1 hashes used for new packages. Is this expected? (I took a look at the `package-lock.json` file on master, and it seems to have some sha1 hashes as well)

package.json
marvinchin: Seems to be related to https://github.com/stylelint/stylelint/issues/3624. I've replaced the single quotes with double quotes - do you mind giving this another try to see if it works @yamgent? Thanks!

package-lock.json
yamgent: The sha1 hashes should be expected. I tried on your new `package-lock.json`, and there were no changes now. :+1:

package.json
yamgent: Works now.

PR #595, opened 2019-01-19
Revise logic to determine root folder

marvinchin:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [x] Bug fix

Fixes #583.

**What is the rationale for this request?**

When root folder is specified, it should ensure that the specified folder contains the `site.json` config file.

If no root folder is specified, we should search up our parent directories to find the folder containing the `site.json` file, and use it as the root folder.

**What changes did you make? (Give an overview)**

- Move process/fs level logic out of Site class to the CLI handlers
- Update logic to determine root folder to follow expected behaviour

**Provide some example code that this change will affect:**

See #583 

**Testing instructions:**
- `markbind serve nonexistent_folder` should throw error
- `markbind serve` should find the closest ancestor with `site.config` and serve from that folder
- `markbind serve` should throw error if no ancestor contains `site.config`

PR #838, opened 2019-04-12
Support building and deploying from multiple json configs

### LINKED ISSUE #832: Support --site-config option for build command
v2.1.1

Current: MarkBind supports `--site-config` option for the `serve` command

Suggested: support `--site-config` for the `build` command as well

Rationale: This will allow generating multiple sites from the same code base by using different `site.json` files

### LINKED ISSUE #833: Separate UG and DG
Current: both UG and DG are in the same site, deployed to markbind.org
Objective: to separate the two so that DG can kept in sync with the master branch head while UG can kept in sync with the latest release. Furthermore, the two can have two different Aloglia search indexes if we switch to Algolia search in future.

### Suggested plan (up for discussion):

Keep all source code in `markbind` repo, but generate three sites
1. UG+DG
   * This will be used for Netlify preview of any changes to UG or DG
   * specified in `site.json`:  UG+DG. 
2. UG
   * specified in `ug-site.json`: Omits all DG pages
   * Deploy UG site to markbind.org at release
3. DG
   * specified in `dg-site.json`: Omits all UG pages
   * Use Travis to auto-push DG site to to gh-pages markbind.github.io/devdocs when master branch is updated

* File structure
  ```
  site.json
  ug-site.json
  dg-site.json
  index.md
  userguide/
    └─ index.md
  devguide/
    └─ index.md
  ```

### Issues

* **How to build sites using different json files?**
  See #832 

* **How to make the root `index.md` file vary based on which site it is?**
  Use tags? e.g.,
  ```
  <span tags="ug">
    <include src="userguide/index.md" />
  </span>
  <span tags="dg">
    <include src="devguide/index.md" />
  </span>
  ```

sijie123:

• [x] Documentation update
• [x] New feature

Follow-up from #793, fixes #833, fixes #832. Supersedes #799.

MarkBind only supports building and deploying from site.json

As per [this comment](https://github.com/MarkBind/markbind/pull/793#issuecomment-478414584), it is useful to keep our user and developer guides
separate, as they are updated with different release cadences.
While the user guide should only be updated on release, the developer
guide needs to be up-to-date with each commit merged into master.

Thus, we require a mechanism to be able to build with different configurations for different deployments.

| Site | What to build | What to deploy | Where to deploy | Config File | Special | Mechanism |
| ------------- | ---------- | ------ | ----- | --------- | ------ | ---------- |
| markbind.org | UG only | UG only | markbind.github.io.git | ug-site.json | CNAME | Travis |
| markbind.org/devdocs | DG only | DG only | devdocs.git | dg-site.json | Ignore CNAME | Travis |
| Netlify Preview | UG and DG | - | - | site.json | Ignore CNAME | Netlify |
| PR Preview | UG and DG | - | - | site.json | Ignore CNAME | Netlify |

* Update site.json to use the tags plugin, offering support for selectively deploying content based on tags/"environments".
* MarkBind now supports building from multiple json configs, and
* MarkBind now supports deploying from multiple json configs, so that we can build and deploy different versions to different sites.
* Added a new deployment command to our travis config to trigger a deployment to the developer site upon any commit to the master branch.
* Updated index.md to selectively show content of developer guide, user guide, or both, depending on deployment context.

I've tested on the devdocs gh-pages. Looks fine :O
Please help me double-check my site.json and .travis.yml configurations - I don't want Site Update V2 :P

**Testing instructions:**
Create a fork(?) and test with Travis...

### DIFF
--- a/docs/dg-site.json
+++ b/docs/dg-site.json
@@ -1,0 +1,46 @@
+{
+  "baseUrl": "/devdocs",
+  "titlePrefix": "MarkBind",
+  "pages": [
+    {
+      "glob": "**/*.mbd"
+    },
+    {
+      "glob": "*.md"
+    },
+    {
+      "src": "index.md",
+      "searchable": "no"
+    },
+    {
+      "glob": "devGuide/*.md"
+    }
+  ],
+  "headingIndexingLevel": 6,
+  "ignore": [
+    "_markbind",
+    "_site/*",
+    "lib/*",
+    "*.json",
+    "*.md",
+    "*.mbd",
+    "*.mbdf",
+    "*.njk",
+    ".git/*",
+    "*.pptx",
+    "CNAME"
+  ],
+  "plugins" : [
+    "filterTags"
+  ],
+  "pluginsContext" : {
+    "filterTags" : {
+      "tags": ["environment--dg"]
+    }
+  },
+  "deploy": {
+    "message": "Dev Site Update.",
+    "repo": "https://github.com/MarkBind/devdocs.git",
+    "branch": "gh-pages"
+  }
+}
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@ program
   .action((userSpecifiedRoot, options) => {
     let rootFolder;
     try {
-      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot);
+      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot, options.siteConfig);
     } catch (err) {
       handleError(err);
     }
@@ -189,10 +189,11 @@ program
   .alias('d')
   .description('deploy the site to the repo\'s Github pages.')
   .option('-t, --travis [tokenVar]', 'deploy the site in Travis [GITHUB_TOKEN]')
+  .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
   .action((options) => {
     const rootFolder = path.resolve(process.cwd());
     const outputRoot = path.join(rootFolder, '_site');
-    new Site(rootFolder, outputRoot).deploy(options.travis)
+    new Site(rootFolder, outputRoot, undefined, undefined, options.siteConfig).deploy(options.travis)
       .then(() => {
         logger.info('Deployed!');
       })
@@ -205,20 +206,21 @@ program
   .alias('b')
   .option('--baseUrl [baseUrl]',
           'optional flag which overrides baseUrl in site.json, leave argument empty for empty baseUrl')
+  .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
   .description('build a website')
   .action((userSpecifiedRoot, output, options) => {
     // if --baseUrl contains no arguments (options.baseUrl === true) then set baseUrl to empty string
     const baseUrl = _.isBoolean(options.baseUrl) ? '' : options.baseUrl;
     let rootFolder;
     try {
-      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot);
+      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot, options.siteConfig);
     } catch (err) {
       handleError(err);
     }
     const defaultOutputRoot = path.join(rootFolder, '_site');
     const outputFolder = output ? path.resolve(process.cwd(), output) : defaultOutputRoot;
     printHeader();
-    new Site(rootFolder, outputFolder)
+    new Site(rootFolder, outputFolder, undefined, undefined, options.siteConfig)
       .generate(baseUrl)
       .then(() => {
         logger.info('Build success!');
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,20 @@ language: node_js
 node_js:
   - '8'
 deploy:
-  provider: script
-  script: cd docs && ../index.js build && ../index.js deploy --travis
-  skip_cleanup: true
-  on:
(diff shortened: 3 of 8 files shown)


yamgent:
`docs/CNAME` needs to match the correct domain name.

marvinchin:
Since CNAME is used for GitHub Pages, I was thinking if this file could be automatically generated by the `markbind build` or `markbind deploy` method instead.

Could we have a new key in `site.json` that allows the user to specify the custom domain to be used? This would allow each config file to define its own domain name.
```
{
  ...
 "deploy": {
    "domain": "markbind.org".
    "message": "Site Update.",
    "repo": "https://github.com/MarkBind/markbind.github.io.git",
    "branch": "master"
  }
}
```

damithc:
> EDIT: I can't think of a good solution for this right now, but this is a critical problem that needs to be addressed 😕.

Can't we use the `ignore` section of `site.json` to omit the CNAME file?

yamgent:
> Can't we use the `ignore` section of `site.json` to omit the CNAME file?

If we want to deploy to both https://dev.markbind.org/ and https://markbind.org, then we can't just omit the CNAME, because GitHub pages rely on that to do custom domains.

yamgent:
> Could we have a new key in `site.json` that allows the user to specify the custom domain to be used? This would allow each config file to define its own domain name.

We can explore that possibility. Do note that this is a GitHub Pages convention, so the `domain` behavior should be exclusively only for GitHub Pages.

damithc:
> If we want to deploy to both https://dev.markbind.org/ and https://markbind.org, then we can't just omit the CNAME, because GitHub pages rely on that to do custom domains.

The two sites are built based on two site.json files. One can ignore the file while the other keeps the file?

yamgent:
> One can ignore the file

The one that ignores the file loses its custom domain.

damithc:
> The one that ignores the file loses its custom domain.

Ah, got it. But we don't need the custom domain for dev docs. It can be served from markbind.github.io/devdocs ? 

Is there a point in creating a repo `dev.markbind.github.io.git`? I thought there can be only one such 'special' repo in an org?

We can vary content of CNAME file using the tag mechanism, if we remove the restriction on file extensions to process anyfile and also allow specifying the name for the output file.
```json
"src": CNAME
"as": CNAME
```
But this is a separate feature altogether.

yamgent:
> It can be served from markbind.github.io/devdocs ?

Ok, we can do that instead. I.e. Serve from https://markbind.org/devdocs/, by creating a `devdocs` repository and deploying to that with CNAME ignored.

@sijie123 do take note.

> Is there a point in creating a repo `dev.markbind.github.io.git`? I thought there can be only one such 'special' repo in an org?

Yeah there's no point actually, the repo name's has no meaning at all for this particular case.

Xenonym:
Do remember to update the [CLI documentation for `build`](https://markbind.org/userGuide/cliCommands.html#build-command) to document the new `-s/--site-config` flag.

sijie123:
I've updated the configs accordingly to use devdocs.git instead.
I have a question though:
It is possible to configure markbind.org/devdocs to serve from the devdocs.git repository? I'm not sure how this works.
Do I need to modify the baseUrl of the dg-site?

yamgent:
> It is possible to configure markbind.org/devdocs to serve from the devdocs.git repository? I'm not sure how this works.

Yes it is done by the admin side, we just need to enable GitHub pages on devdocs.git and it will automatically tie to that domain (because that's the default behaviour when we don't specify a domain).

> Do I need to modify the baseUrl of the dg-site?

Yes.

docs/dg-site.json
```diff
@@ -0,0 +1,46 @@
...
+  ],
+  "pluginsContext" : {
+    "filterTags" : {
+      "tags": ["environment--dg"]
+    }
+  },
+  "deploy": {
+    "message": "Dev Site Update.",
+    "repo": "https://github.com/MarkBind/devdocs.git",
+    "branch": "master"
```
nicholaschuayunzhi: Should this be gh-pages instead?

damithc: I think so. In fact, it defaults to gh-pages, which means we can omit it?

sijie123: Thanks. Updated. I've set it to gh-pages now, just to make it explicit.

sijie123:
Ready for re-review @yamgent

index.js
```diff
@@ -205,20 +206,21 @@ program
...
           'optional flag which overrides baseUrl in site.json, leave argument empty for empty baseUrl')
+  .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
   .description('build a website')
   .action((userSpecifiedRoot, output, options) => {
     // if --baseUrl contains no arguments (options.baseUrl === true) then set baseUrl to empty string
     const baseUrl = _.isBoolean(options.baseUrl) ? '' : options.baseUrl;
     let rootFolder;
     try {
-      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot);
+      rootFolder = cliUtil.findRootFolder(userSpecifiedRoot, options.siteConfig);
```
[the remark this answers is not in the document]
marvinchin: Yes, seems like in was broken in #595 😅

We will need to pass `options.siteConfig` into the `cliUtil.findRootFolder` call in the `serve` command as well.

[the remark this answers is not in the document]
sijie123: Yes. The serve command actually doesn't find the correct root folder.
Thanks for the catch. I'll fix that up too.

Issue #739, opened 2019-03-02
siteNav: Styles are not being applied for 'layoutName/navigation.md'


amad-person:
**Tell us about your environment**

* **MarkBind Version: 1.19.1**

**What did you do? Please include the actual source code causing the issue.**

1. Create a new MarkBind project (run `markbind init`).
2. Add a new page (e.g. `about.md`) with the following **frontmatter**:

```
<frontmatter>
  layout: default
</frontmatter>
```
3. Update `_markbind/layouts/default/navigation.md` with the site navigation menu to include the new page.

      I copied over the contents of `_markbind/navigation/site-nav.md` and added a new link for my page:
```
<navigation>
* [Home :glyphicon-home:]({{baseUrl}}/index.html)
* [About]({{baseUrl}}/about.html)
</navigation>
```
4. Build the site and serve it.

**What did you expect to happen?**

Expected `about.html` to have styles for `#site-nav` and `#page-content`(specified in `site-nav.css`) applied, as it has a site navigation menu specified in its layout.

**What actually happened? Please include the actual, raw output.**

This is how `about.html` looks like:

![screenshot_2019-03-02 about](https://user-images.githubusercontent.com/22164211/53679902-5bc53c80-3d0e-11e9-9faf-1fde28529b63.png)

DevTools showing CSS rules for ~`#site-nav`~ `#page-content` (but site-nav has no styles too) for `about.html`:

<img width="750" alt="devtools showing css for about.html" src="https://user-images.githubusercontent.com/22164211/53679973-587e8080-3d0f-11e9-94fd-9abd214e4018.png">

Meanwhile, `index.html` looks like this:

![screenshot_2019-03-02 hello world](https://user-images.githubusercontent.com/22164211/53679912-7ac3ce80-3d0e-11e9-86b1-25764052eb89.png)

`index.md` has the following **frontmatter**:

```
<frontmatter>
  title: "Hello World"
  footer: footer.md
  siteNav: site-nav.md
</frontmatter>
```

If there is a siteNav attribute present in the **frontmatter**, the styles are being applied. 
DevTools showing CSS rules for `#site-nav` for `index.html`:

<img width="750" alt="devtools showing css for index.html" src="https://user-images.githubusercontent.com/22164211/53679998-ac896500-3d0f-11e9-9dcb-b9cb7484b31e.png">

[...]

PR #841, opened 2019-04-15
Fix site-nav.css not being added when layout contains site nav

### LINKED ISSUE #528: Specifying layouts in a Page's front matter do not insert required CSS files for site navigation
* **MarkBind Version:** 1.15.2

**Cause**

The problem resides within Page.prototype.prepareTemplateData. If a `siteNav` is not explicitly declared within the `frontmatter` of a page, the associated CSS files will not be inserted into the page.

Bug found in #527

marvinchin:

• [x] Bug fix

Fixes #528.

When a page has a site nav specified in the layout, but not in the frontmatter, the site-nav.css file is not added to the site. This causes the styling of the side nav to break.

- [x] Abstract method to get the path to the site nav file from `insertSiteNav`
- [x] Use abstracted method `getSiteNavPath` to check if a site nav is specified in either layout or frontmatter when preparing template data

~~By default, the default layout contains a blank `navigation.md` file. The site-nav.css file is thus added even though the `navigation.md` file is empty - is this behaviour okay?~~

After updating the logic to use a flag instead, we are able to avoid this issue as we only update the flag after checking that the file is not empty.

**Testing instructions:**

See instructions in #739. The site nav should be styled correctly.

When passing data to the page template, the siteNav variable only checks
for the existence of the site nav specified in the frontmatter. The
site-nav.css file is thus only added when the site nav is specified in
the frontmatter.

When a page has a layout with a site nav, but no site nav is specified
in the frontmatter, the site-nav.css file is not added, which causes the
styling of the site nav to break.

Let's update the logic when forming template data to check for the
existence of a site nav in the template when determining the value of
the siteNav variable.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -87,6 +87,9 @@ function Page(pageConfig) {
   this.keywords = {};
   this.navigableHeadings = {};
   this.pageSectionsHtml = {};
+
+  // Flag to indicate whether this page has a site nav
+  this.hasSiteNav = false;
 }
 
 /**
@@ -231,7 +234,7 @@ Page.prototype.prepareTemplateData = function () {
     markBindVersion: `MarkBind ${CLI_VERSION}`,
     pageNav: this.isPageNavigationSpecifierValid(),
     pageNavHtml: this.pageSectionsHtml[`#${PAGE_NAV_ID}`] || '',
-    siteNav: this.frontMatter.siteNav,
+    siteNav: this.hasSiteNav,
     siteNavHtml: this.pageSectionsHtml[`#${SITE_NAV_ID}`] || '',
     title: prefixedTitle,
     enableSearch: this.enableSearch,
@@ -575,12 +578,15 @@ Page.prototype.insertSiteNav = function (pageData) {
   // Retrieve Markdown file contents
   const siteNavPath = path.join(this.rootPath, siteNavFile);
   if (!fs.existsSync(siteNavPath)) {
+    this.hasSiteNav = false;
     return pageData;
   }
   const siteNavContent = fs.readFileSync(siteNavPath, 'utf8');
   if (siteNavContent === '') {
+    this.hasSiteNav = false;
     return pageData;
   }
+  this.hasSiteNav = true;
   // Set siteNav file as an includedFile
   this.includedFiles[siteNavPath] = true;
   // Map variables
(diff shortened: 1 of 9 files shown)


Chng-Zhi-Xuan:
I briefly looked through, here are some comments

> By default, the default layout contains a blank navigation.md file. The site-nav.css file is thus added even though the navigation.md file is empty - is this behaviour okay?

In my opinion, if the user has a `navigation.md` file in the layouts, we should expect that the user intends to use the `site-nav`. Although we can make it more user friendly by using a better initial template rather than a blank file. 

We might also want to insert notes / tips into the `Layouts` section to inform users to delete certain files if you do not want to use them. 

**Implementation comments**

- Your implementation would require 2 calls to `this.getSiteNavPath()` when generating a page. 

To streamline this, would it be better if you just changed a boolean page variable after `this.getSiteNavPath()` is called, then use the same variable for the `siteNav` value within `prepareTemplateData` ?

marvinchin:
Thanks for pointing that out!

I was trying to avoid the use of a global flag because it introduces some coupling between places that modify the flag, and those that read it (in this case, between `insertSiteNav` and `prepareTemplateData`).

As `getSiteNavPath` involves a (comparatively more costly) fs operation, and I don't forsee any other places where the flag would be modified, it might be more efficient to use a global flag as you suggested. I'll make the changes in a bit! 🙂

marvinchin:
> By default, the default layout contains a blank `navigation.md` file. The site-nav.css file is thus added even though the `navigation.md` file is empty - is this behaviour okay?

After updating the logic to use a flag instead, we are able to avoid this issue as we only update the flag after checking that the file is not empty.

src/Page.js
```diff
@@ -581,6 +584,7 @@ Page.prototype.insertSiteNav = function (pageData) {
   if (siteNavContent === '') {
     return pageData;
   }
+  this.hasSiteNav = true;
```
Chng-Zhi-Xuan: Since you only modify the flag to be true at this point. If you delete the `navigation.md` (Or blank the contents) from the layouts folder, the `site-nav` will still appear in the `markbind serve` preview. Even though MarkBind correctly detected `navigation.md` has changed.

The only correct behaviour in this code is changing `navigation.md` slightly and the changes will reflect in the `markbind serve` preview.

**Suggestion**
Explicitly set the flag to false in the if statements above for it to function properly when in preview mode.

src/Page.js
```diff
@@ -87,6 +87,9 @@ function Page(pageConfig) {
   this.keywords = {};
   this.navigableHeadings = {};
   this.pageSectionsHtml = {};
+
+  // flag to indicate whether this page has a site nav
```
Chng-Zhi-Xuan: A quick search shows majority of inline comments have capitalised starting character (although not all).

marvinchin:
@Chng-Zhi-Xuan I've made the requested changes!

damithc:
Seems to be working. Good work!

PR #852, opened 2019-05-13
Use Sets instead of Javascript objects for Site.baseUrlMap, builtFiles, Page.includedFiles

### LINKED ISSUE #772: Use of Javascript dictionaries for includedFiles

openorclose:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [x] Other, please explain: Coding style "fix"

Resolves #772

Sets are a more idiomatic way of keeping track of present items.

Previously, an object was used:
`object[key] = true` if key exists
and
`object[key]` to check if the key exists

Now, sets are used:
`set.add(key)` if key exists
and
`set.has(key)` to check if the the key exists

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -110,9 +110,9 @@ function calculateNewBaseUrl(filePath, root, lookUp) {
       return undefined;
     }
     const parent = path.dirname(file);
-    if (lookUp[parent] && result.length === 1) {
+    if (lookUp.has(parent) && result.length === 1) {
       return path.relative(root, result[0]);
-    } else if (lookUp[parent]) {
+    } else if (lookUp.has(parent)) {
       return calculate(parent, [parent]);
     }
     return calculate(parent, result);
--- a/src/Page.js
+++ b/src/Page.js
@@ -945,12 +945,11 @@ Page.prototype.resolveDependency = function (dependency, builtFiles) {
     const resultDir = path.dirname(path.resolve(this.resultPath, path.relative(this.sourcePath, file)));
     const resultPath = path.join(resultDir, FsUtil.setExtension(path.basename(file), '._include_.html'));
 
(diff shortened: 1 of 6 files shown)


src/Page.js
```diff
@@ -945,12 +945,12 @@ Page.prototype.resolveDependency = function (dependency, builtFiles) {
     const resultDir = path.dirname(path.resolve(this.resultPath, path.relative(this.sourcePath, file)));
     const resultPath = path.join(resultDir, FsUtil.setExtension(path.basename(file), '._include_.html'));
 
-    if (builtFiles[resultPath]) {
+    if (builtFiles.has(resultPath)) {
       return resolve();
     }
 
     // eslint-disable-next-line no-param-reassign
```
Chng-Zhi-Xuan: This line is not necessary any more

PR #857, opened 2019-05-14
Remove anchor generation dependency on indexing level

### LINKED ISSUE #775: Heading anchors missing in some places
v1.20.0

Problem: heading anchors missing in some places
image

https://nus-te3201.github.io/2019/programming/toc/oop.html#exercise-create-stockitem-class

### LINKED ISSUE #776: Unintended coupling between indexing for built-in search and anchor link generation

* **MarkBind Version:** v1.20.0

Changing also `headingIndexLevel` changes what anchor links are generated. See #775.

Search indexing and anchor generation should not be coupled, since anchors can/should be generated even when built-in search is not enabled.

See #775.

alyip98:
• [X] Bug fix

Fixes #775 and fixes #776 

Anchors were only generated if the indexing level was greater than the
heading level, resulting in missing anchors and inconsistent behaviour

Removed check for `headingIndexingLevel` 
Anchors are generated for all heading tags, except those hidden in panels
Added test pages

There's currently a bug due to how vue handles collapsing panels that make anchors in panels disappear after collapsing and uncollapsing the panel.

**Testing instructions:**
1. Access test_site/testAnchorGeneration.html
2. Check that every heading labeled "should have anchor" indeed have anchors
3. Check that heading s  explicitly labeled "should not have anchor" do not have anchors

Anchor generation was dependent on the heading indexing level as 
defined in site.json, which resulted in missing anchors for some
headings.

Let's decouple anchor generation from heading indexing level, by
enabling anchor generation for all headings regardless of heading 
indexing level.

### DIFF
--- a/src/plugins/default/markbind-plugin-anchors.js
+++ b/src/plugins/default/markbind-plugin-anchors.js
@@ -2,39 +2,24 @@ const cheerio = module.parent.require('cheerio');
 const md = require('./../../lib/markbind/src/lib/markdown-it');
 
 const ANCHOR_HTML = '<a class="fa fa-anchor" href="#"></a>';
-
-/**
- * Generates a heading selector based on the indexing level
- * @param headingIndexingLevel to generate
- */
-function generateHeadingSelector(headingIndexingLevel) {
-  let headingsSelector = 'h1';
-  for (let i = 2; i <= headingIndexingLevel; i += 1) {
-    headingsSelector += `, h${i}`;
-  }
-  return headingsSelector;
-}
-
+const HEADER_TAGS = 'h1, h2, h3, h4, h5, h6';
 /**
  * Adds anchor links to headers
  */
 module.exports = {
-  postRender: (content, pluginContext, frontMatter, pageConfig) => {
+  postRender: (content) => {
     const $ = cheerio.load(content, { xmlMode: false });
-    if (pageConfig.headingIndexingLevel > 0) {
-      const headingsSelector = generateHeadingSelector(pageConfig.headingIndexingLevel);
-      $(headingsSelector).each((i, heading) => {
-        $(heading).append(ANCHOR_HTML.replace('#', `#${$(heading).attr('id')}`));
-      });
-      $('panel[header]').each((i, panel) => {
-        const panelHeading = cheerio.load(md.render(panel.attribs.header), { xmlMode: false });
-        if (panelHeading(headingsSelector).length >= 1) {
-          const headingId = $(panelHeading(headingsSelector)[0]).attr('id');
-          const anchorIcon = ANCHOR_HTML.replace(/"/g, "'").replace('#', `#${headingId}`);
-          $(panel).attr('header', `${$(panel).attr('header')}${anchorIcon}`);
-        }
-      });
-    }
+    $(HEADER_TAGS).each((i, heading) => {
+      $(heading).append(ANCHOR_HTML.replace('#', `#${$(heading).attr('id')}`));
+    });
+    $('panel[header]').each((i, panel) => {
+      const panelHeading = cheerio.load(md.render(panel.attribs.header), { xmlMode: false });
+      if (panelHeading(HEADER_TAGS).length >= 1) {
+        const headingId = $(panelHeading(HEADER_TAGS)[0]).attr('id');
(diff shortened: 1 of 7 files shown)


src/plugins/default/markbind-plugin-anchors.js
```diff
@@ -2,39 +2,25 @@ const cheerio = module.parent.require('cheerio');
...
-  return headingsSelector;
-}
-
+const HEADER_TAGS = 'h1, h2, h3, h4, h5, h6';
 /**
  * Adds anchor links to headers
  */
 module.exports = {
+  // eslint-disable-next-line no-unused-vars
   postRender: (content, pluginContext, frontMatter, pageConfig) => {
```
Chng-Zhi-Xuan: After looking through `filterTags`, I think that we should maintain consistency between our in-built plugins. 

Therefore, we should remove the unused variables `pluginContext`, `frontMatter` and `pageConfig` rather than using the eslint disable next line.

test/functional/test_site/testAnchorGeneration.md
```diff
@@ -0,0 +1,34 @@
...
+<panel header="Not expanded">
+    
+### Header inside panel (non-expanded)
+Should **not** have anchor
+</panel>
+
+
+<panel header="Expanded" expanded>
+    
+### Header inside panel (expanded)
```
Chng-Zhi-Xuan: Should test all 6 levels of headings within a panel as well. Likewise for the included file.

alyip98:
Force pushed with suggested changes

yamgent:
Seems a bit extreme to force all headings to have anchors though... is it possible to support a separate settings just for the anchors? I recall plugins supporting a way to supply settings through pluginsContext.

Chng-Zhi-Xuan:
> Seems a bit extreme to force all headings to have anchors though... is it possible to support a separate settings just for the anchors? I recall plugins supporting a way to supply settings through pluginsContext.

Then we would need to change the phrasing of our user guide a little 

`MarkBind has a set of default plugins that it uses to carry out some of its features. These are enabled by default for every project and should be left alone.`

Since it was originally implied that the default built-in plugins should not be modified, at most turned off. 

Should we open up another issue, to discuss if there is a need to customise anchor generation for different heading levels? I feel having it for all heading levels is fine, and a binary argument to turn it off is sufficient.

damithc:
Shall we try the current approach of generating anchors for _all_ levels and see how that plays out in practice?

yamgent:
> Shall we try the current approach of generating anchors for all levels and see how that plays out in practice?

Ok sure, let's do that.

test/functional/test_site/expected/testAnchorGeneration.html
```diff
@@ -0,0 +1,114 @@
...
+        <h2 id="should-have-anchor-2">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-2"></a></h2>
+        <h3 id="should-have-anchor-3">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-3"></a></h3>
+        <h4 id="should-have-anchor-4">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-4"></a></h4>
+        <h5 id="should-have-anchor-5">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-5"></a></h5>
+        <h6 id="should-have-anchor-6">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-6"></a></h6>
+        <panel header="#### should have anchor<a class='fa fa-anchor' href='#should-have-anchor'></a>">
+          Lorem ipsum
+        </panel>
+        <panel header="Collapsed">
+          <p>Headings in a collapsed-by-default panel should <strong>not</strong> have anchors</p>
```
yamgent: This test case is "should **not** have anchors", but anchors are generated anyway:

`<a class="fa fa-anchor" href="#should-not-have-anchor"></a>`

Same for some other anchors below.

Chng-Zhi-Xuan:
To add on, the anchors for collapsed panels are still there, but not visible. The event listeners (for triggering visibility of anchors) were attached during `setup.js` are removed when the panel collapses due to the conditional rendering of vue.js. 

For now this PR solves the immediate issue present in #775 , although there could be improvements (re-attaching event listeners) and (only generating anchors for expanded panels).

PR #887, opened 2019-06-10
Add support for PlantUML diagrams

### LINKED ISSUE #877: Support embedding PlantUML diagrams
As MarkBind is to be optimized for software documentation, it may be useful to support [PlantUML](http://plantuml.com).

alyip98:
• [x] New feature

Resolves #877 

> As MarkBind is to be optimized for software documentation, it may be useful to support PlantUML.

PlantUML is a tool written in Java that allows users to write diagrams in a simple and intuitive language. Being able to write (and track) UML diagrams in a text format compared to using tools such as Powerpoint or Photoshop allows better version control and ease of use.

Add ability to include UML diagrams via a `<puml>` tag. The diagram itself is written in PlantUML syntax in a separate file. The page includes the diagram using the following syntax: 
`<puml src="diagrams/example.puml"/>`

Due to limitations of htmlparser2, including UML code inline will cause problems as UML syntax violates HTML specifications (e.g. presence of symbols such as `<--` which are misinterpreted as HTML tags). Fixing this will require cheerio to use a forked version of htmlparser2, discussed in #582, which in itself is an open problem.

**Testing instructions:**
The user guide contains an [entry](https://deploy-preview-887--markbind-master.netlify.com/userguide/puml)

Markbind page authors can use diagrams in their sites by linking them
in a `<pic>` tag. This is tedious as it requires the author to generate
the diagrams using an external tool, and the binary files generated by
that tool (e.g. .pptx, .psd) are not version control friendly.

Let's support PlantUML, a tool that allows users to write diagrams in
a simple and intuitive language. Users can then write their diagrams
in a text editor, and Markbind will generate the diagrams using
PlantUML, eliminating the need of an external editor.


Chng-Zhi-Xuan:
> Will it be more advantageous to just use an existing module? node-plantuml is tested, documented, and provides more features. However, not all of its features are relevant to MarkBind (due to it also being a CLI application)

Does your current implementation require the user to already have Java installed? If so it might be problematic to impose installing a JRE / JDK to use this feature.

Having looked through `node-plantuml`, looks like its more suited for MarkBind. 
Seems like we can avoid the *"diagrams generated after MarkBind generates the website"* problem as well by using a custom tag and cheerio to selectively parse the input into `node-plantuml`

```HTML
<plantuml>
Bob->Alice : hello
</plantuml>
```

```Javascript
Page.prototype.generatePlantuml = function (content) {
  const $ = cheerio.load(content);
  const plantumlContent = $("plantuml");
  ...
  ...
}
```

> Is the delay going to be a problem, especially on large sites with numerous diagrams?

Build times is not really a concern (unless it is a significant increase). 

You should take note on how it affects previewing websites for authors. Manually recording some timings on a test page with different number of diagrams & other components can be a good gauge if it is problematic.

**Other comments:**

- Looks like your code crashes if an `out` folder is not created. If a new folder is required, you should place it within `_markbind` as well as creating that folder on `markbind init` by modifying code in `Site.js`.
- Name the folder storing the generated diagrams something better than `out`
- A good enhancement would be to ensure minimal generation of duplicate diagrams (E.G 2 different pages uses the same plantuml markup). Not sure if this is already handled by the plugin/parser.

Although not fully tested, this seems like it is a great feature addition as I can place the plantuml diagrams within MarkBind components.

alyip98:
> Does your current implementation require the user to already have Java installed? If so it might be problematic to impose installing a JRE / JDK to use this feature.

Yeah, PlantUML is supplied as a JAR, there's no way around it. node-plantuml is also a wrapper that runs the JAR which offers an abstracted API. 

> Having looked through node-plantuml, looks like its more suited for MarkBind.
Seems like we can avoid the "diagrams generated after MarkBind generates the website" problem as well by using a custom tag and cheerio to selectively parse the input into node-plantuml

Currently I'm using markdown-it to parse since it's more convenient. But I think I'll try using cheerio and see if its better.

My reservation for not using node-plantuml is because all the extra features are unnecessary, and it presents another potential source of errors.

But now I'm encountering a new problem where certain symbols within the PlantUML language are misinterpreted as HTML tags. For now I'll try using cheerio instead and see if it solves this problem

src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -9,19 +9,35 @@ const JAR_PATH = path.resolve(__dirname, 'plantuml.jar');
...
+  // console.log("original: ", src);
+  // console.log("resolved: ", rawDiagramPath);
+
   // Creates output dir if it doesn't exist
   if (!fs.existsSync(outputDir)) {
-    fs.mkdirSync(outputDir, { recursive: true });
+    try {
+      fs.mkdirSync(outputDir, { recursive: true });
+      // eslint-disable-next-line no-empty
+    } catch (_e) {}
```
Chng-Zhi-Xuan: Is there a reason to catch this error but not throw it?

docs/_markbind/layouts/userGuide/navigation.md
```diff
@@ -7,6 +7,7 @@
   * [Adding Pages]({{baseUrl}}/userGuide/addingPages.html)
   * [MarkBind Syntax Overview]({{baseUrl}}/userGuide/markBindSyntaxOverview.html)
   * [Formatting Contents]({{baseUrl}}/userGuide/formattingContents.html)
+  * [UML]({{baseUrl}}/userGuide/puml.html)
```
Chng-Zhi-Xuan: `Inserting PlantUML Diagrams`?

docs/userGuide/puml.md
```diff
@@ -0,0 +1,84 @@
...
+  layout: userGuide
+</frontmatter>
+
+
+# {{ title }}
+
+
+<span id="overview" class="lead">
+
+**MarkBind offers integration with PlantUML in your pages,** allowing you design and write UML diagrams (and more) using PlantUML syntax, which is then compiled and served as images.
```
Chng-Zhi-Xuan: - Add a newline after the bold text?
- `(and more)` should naturally link to other examples
- You should put a warning at the top of this page indicating that authors need to have a JRE/JDK installed, along with the graphviz installer to use this feature.

test/functional/test_site/testPlantUML.md
```diff
@@ -0,0 +1,22 @@
+## PlantUML Test
+
+### Sequence Diagram
+<puml src="diagrams/sequence.puml" />
```
Chng-Zhi-Xuan: When I generate this page, the `src` will be changed to `test_site/diagrams/abc.png` when the correct one should be without `test_site`.

src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -0,0 +1,76 @@
...
+
+  // Creates output dir if it doesn't exist
+  if (!fs.existsSync(outputDir)) {
+    fs.mkdirSync(outputDir, { recursive: true });
+  }
+
+  // Read diagram file
+  fs.readFile(src, (err, data) => {
+    if (err) {
+      throw err;
```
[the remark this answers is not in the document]
alyip98: Ok, good point

test/functional/test_site/testPlantUML.md
alyip98: Seems to be an issue with the `<pic>` component, will look into it

src/plugins/default/markbind-plugin-plantuml.js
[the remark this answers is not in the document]
Chng-Zhi-Xuan: Documenting in-person discussion:

- Unable to share the same error names as the same name was defined in a higher scope. So we decided to move the verbose error message to the logger file and print out a more meaningful message on the console.

damithc:
Use Github's  'draft PR' feature instead of `WIP` prefix in the PR title.

alyip98:
@Chng-Zhi-Xuan 
> For the graceful error handling, there is room for some improvement:
> 
> * All errors that result in the diagram not being generated should return immediately.
> * All errors should be logged in `logger.debug`, not just replaced with a generic error message.
> * We should only show 1 `logger.error` per diagram (to reduce clutter).
> 
> This would ensure we only get 1 console error message per diagram, and all errors that occurred up until the critical error is logged properly.

Currently it already does this. There are 2 main locations where errors occur - reading the .puml, and running the jar. If an error occurs during the read, the error is logged and the jar is not run. If an error occurs while running the jar, either the `error` _or_ `close` event is triggered, but not both, so it will not be double logged.

alyip98:
@damithc there are some issues with Github's draft PR feature, namely not being able to revert back to it or retroactively mark an opened PR as draft

docs/userGuide/puml.md
```diff
@@ -15,25 +15,31 @@
 
 <box type="warning">
 
-**[Java](https://www.java.com/en/download/) must be installed to use this feature**. _(Any recent version should work)_
-</box>
-<box type="warning">
+**[<tooltip content="Any recent version">Java</tooltip>](https://www.java.com/en/download/) must be installed to use this feature**
 
-**[Graphviz](https://www.graphviz.org/download/) must be installed to generate diagrams**, with the exception of sequence and activity diagrams. (_tested with v2.38_)
+**[<tooltip content="Tested with v2.38">Graphviz</tooltip>](https://www.graphviz.org/download/) must be installed to generate <tooltip content="with the exception of sequence and activity diagrams">diagrams</tooltip>**
```
Chng-Zhi-Xuan: Will there be a case where activity and sequence diagrams don't work if they used a certain syntax for another component (without Graphviz) ?

docs/userGuide/puml.md
```diff
@@ -15,11 +15,11 @@
 
 <box type="warning">
 
-**[Java](https://www.java.com/en/download/) must be installed to use this feature**
+**[Java](https://www.java.com/en/download/) must be installed to use this feature**. _(Any recent version should work)_
```
Chng-Zhi-Xuan: How about "Java must be installed to generate the UML diagrams via a JAR executable"? 

To inform the user why is it needed.

docs/userGuide/puml.md
```diff
@@ -15,11 +15,11 @@
...
 </box>
 <box type="warning">
 
-**[Graphviz](https://www.graphviz.org/download/) must be installed to generate diagrams**, with the exception of sequence and activity diagrams
+**[Graphviz](https://www.graphviz.org/download/) must be installed to generate diagrams**, with the exception of sequence and activity diagrams. (_tested with v2.38_)
```
Chng-Zhi-Xuan: "... to generate *all* PlantUML diagrams"?

docs/userGuide/puml.md
alyip98: Honestly, I don't know. What do you mean by a certain syntax for another component?

damithc: On a separate note, use tooltips only in cases where you want to supply a definition/explanation of a term. Otherwise, the reader will have to retrieve every tooltip to see if there is something relevant in the tooltip. Use dimmed text to provide unimportant details.
`recent version` is not precise enough. Say something like `version 8 or later`.

docs/userGuide/puml.md
alyip98: > * You have not initialised the `diagrams` folder within `_markbind` during `markbind init`. You also have to update the `Site.test.js` code since an additional folder is expected on initialisation. 

The diagram files can be in any folder, not necessarily restricted to the `diagrams` folder. I used `diagrams` in this case as an example, but it could have been any folder.

alyip98:
> * You have not initialised the `diagrams` folder within `_markbind` during `markbind init`. You also have to update the `Site.test.js` code since an additional folder is expected on initialisation. 

The diagram files can be in any folder, not necessarily restricted to the `diagrams` folder. I used `diagrams` in this case as an example, but it could have been any folder.

alyip98:
@Chng-Zhi-Xuan @damithc I've rephrased the requirements section

Chng-Zhi-Xuan:
> You have not initialised the diagrams folder within _markbind during markbind init. You also have to update the Site.test.js code since an additional folder is expected on initialisation. All sharable assets between pages (such as site-nav) should be in the _markbind folder.

@alyip98 You have not addressed the request above.

Chng-Zhi-Xuan:
As discussed in person, since `<puml>` for an `` tag, we should not enforce that diagrams be placed in `_markbind/diagrams` because we don't enforce that for images as well.

Issue #846, opened 2019-05-01
Nunjucks variables not resolved


sijie123:
* **MarkBind Version:**
2.4.0

**What did you do? Please include the actual source code causing the issue.**
Built [CS3281 student website](https://github.com/nus-cs3281/2019) with MarkBind 2.4.0. [Travis Log](https://travis-ci.org/nus-cs3281/2019/builds/526733272?utm_source=github_status&utm_medium=notification)

**What did you expect to happen?**
Expected the site to build as per normal.

**What actually happened? Please include the actual, raw output.**
It appears that the {{students[1]}} variable was not resolved.

Forced npm to use markbind-cli@2.3.0, and the site built successfully. [Travis log](https://travis-ci.org/nus-cs3281/2019/builds/526735520?utm_source=github_status&utm_medium=notification)

I'm not sure if it's an intentional change, but it appears that some variables are not resolved at all...

damithc:
@jamos-tay can take a look?

yamgent:
Nunjucks rendering was deferred to a later stage since #751, which explains the error. We need the render to happen before `_preprocess()` gets called again here:

https://github.com/MarkBind/markbind/blob/4d1101243781742448475993b43b6da2921836ab/src/lib/markbind/src/parser.js#L422

Since MarkBind's user guide is also affected by this (see https://markbind.org/userGuide/formattingContents.html), setting the priority of this to High.

jamos-tay:
Sure I'll take a look over the weekend

damithc:
How's this coming? Need to fix soon as it makes the current MarkBind version unusable :-(

yamgent:
I was looking at the code to see if we can avoid deferring Nunjucks rendering to such a late stage, but I realise that there's going to be a potential clash between supporting this use case and supporting import variables.

Consider this case:

`index.md`

```html
<import src="{{ file }}#var" />
{{ var }}
```

If we do the Nunjucks rendering first before parsing, `{{ var }}` will not be rendered correctly.

If we do the Nunjucks rendering later after parsing, `{{ file }}#var` will not be a valid source path.

Selective Nunjucks rendering could be explored, but it seems that edge cases are going to be quite problematic (e.g. the `src` being entirely a Nunjucks variable could throw off the detection of which variables to render first, and which to render later).

jamos-tay:
Yeah I don't have much ideas either...

[...]

PR #891, opened 2019-06-25
Add importing variables from other files

### LINKED ISSUE #682: Allow accessing page variables from outside

openorclose:

• [x] New feature

Fixes #682 

Syntax introduced:
```xml
<import
  varname1 varname2 ...
  from='filename.md'
  as='namespace'
/>
```
If varname1, varname2... attributes are specified,
the respective variables from filename.md can be used as variables
in the current page normally: `{{varname1}} {{varname2}}.`

If attribute 'as' is specified, ALL page variables from filename.md
can be accessed through the namespace specified in the attribute:
`{{namespace.varname3}}, {{namespace.varname4}}`

**Implementation:**
Extract all imported variables, put them in a namespace if none specified and replace them with themselves during the first nunjucks call:
```
<import var from="file.md" as="namespace" />
```
{{namespace.var}} turns into {{ namespace.var }}.

```
<import var from="file.md" />
```
{{var}} turns into {{ var.var }}. (The generated namespace is the first variable name lexicographically). 

Comments on the syntax and word choice.

Any previous issues that were discussed during the previous implementation, such as the commit message here:

https://github.com/MarkBind/markbind/commit/48b57a18a8bfd68101b163908da4a0541756364a

The issue here: 

https://github.com/MarkBind/markbind/issues/846

Sample use case here: 

https://github.com/MarkBind/markbind/pull/751#issuecomment-469670640

Note that in this use case, storing the file path in a variable does not work out too well because of this line:

https://github.com/MarkBind/markbind/blob/58b23383436cb06615d1399cc496d400a468c3aa/src/lib/markbind/src/parser.js#L146

Markdown converts the file path to a hyperlink, and so can't be used in a src attribute, as can be seen in the error in the preview here: https://5d119ab7d5ca1b0007fe4df5--markbind-master.netlify.com/testing/usecaseincomment.

However, it should still work in spirit, as long as the variable name does not look like a link, shown by the working panel above the error in the previous link.

**Testing instructions:**

I've put some demos into docs/testing so that they show up in the netlify preview. Take a look at the source files in this branch and the resulting output in the preview:

https://5d119ab7d5ca1b0007fe4df5--markbind-master.netlify.com/testing/testimportvariables

https://5d119ab7d5ca1b0007fe4df5--markbind-master.netlify.com/testing/usecaseincomment

**Commit Message:**

```
Let's introduce a new syntax to import variables from other files:

<import
  varname1 varname2 ...
  from='filename.md'
  as='namespace'
/>

If varname1, varname2... attributes are specified,
the respective variables from filename.md can be used as variables
in the current page normally: {{varname1}} {{varname2}}.

If attribute 'as' is specified, ALL page variables from filename.md
can be accessed through the namespace specified in the attribute:
{{namespace.varname3}}, {{namespace.varname4}}

In extractPageVariables, we store all non-imported page variables into
a global lookup table. Also we detect what variables are imported ones,
and let nunjucks replace them with themselves first.
In the above example,
{{namespace.varname1}} would be replaced with {{namespace.varname1}}
in the first nunjucks call in includeFile after extractPageVariables.

After this, and before _preprocess, a new preprocessing step
is introduced that extracts the imported variables,
after which another nunjucks call is made so that the imported variables
can be rendered correctly.
```


src/lib/markbind/src/parser.js
```diff
@@ -566,7 +677,7 @@ Parser.prototype.includeFile = function (file, config) {
         try {
           processed = this._preprocess(d, context, config);
         } catch (err) {
-          err.message += `\nError while preprocessing '${file}'`;
+          err.message += `${err.stack}\nError while preprocessing '${file}'`;
```
Chng-Zhi-Xuan: `err.stack` is not standard and should not be used in production code. 

[Reference here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack)

src/lib/markbind/src/parser.js
```diff
@@ -181,6 +220,73 @@ Parser.prototype._preprocessThumbnails = function (element) {
   return element;
 };
 
+Parser.prototype._prepreprocessInnerVariables = function (content, context, config) {
```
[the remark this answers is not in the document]
Chng-Zhi-Xuan: **pre***pre*process is definitely not a good name 😅

damithc:
@openorclose is this ready for review? Looks like @Chng-Zhi-Xuan has given some comments.

openorclose:
Force pushed with required changes.

openorclose:
Thanks for the review!

> Some things observed while testing this feature:
> 
> If importing a variable from other file with the same name, the variable in the current file is overridden (which defeats the purpose of a namespace).
> 
> `test.md`
> 
> ```
> <variable name="abc">Test's abc.</variable>
> ```
> 
> `index.md`
> 
> ```
> <variable name="abc">Index's abc.</variable>
> <import abc from="test.md" as="testRef" />
> 
> {{abc}}
> {{testRef.abc}}
> 
> 
> 
> ```

If the as attribute for the namespace is specified, then all the variables can be used without explicitly declaring their names :`<import from="test.md" as="testRef"/>` would be enough.

> Namespace the same, but variable names are different. The variables get overridden anyway.
> 
> `alice.md`
> 
> ```
> <variable name="income">$5000</variable>
> ```
> 
> `bob.md`
> 
> ```
> <variable name="job">Clerk</variable>
> ```
> 
> `index.md`
> 
> ```
> <import income from="alice.md" as="dreamJob" />
> <import job from="bob.md" as="dreamJob" />
> 
> {{dreamJob.income}}
> {{dreamJob.job}}
> 
> 
> 
> ```

My implementation requires a different namespace for each import. Alternatively you can do

> ```
> <import income from="alice.md" />
> <import job from="bob.md" />
> 
> {{income}}
> {{job}}
> 
> 
> 
> ```

My intention was that you specify only either a namespace OR the list of variable names. If you specify this list of variable names without a namespace, they can be used like a normal variable {{varname}}. If you specify a namespace, then all variables inside the file can be accessed via {{namespace.varname}}. So if you specify both, you'll be able to use either method to access `varname`.

> Self referencing fails.
> 
> `index.md`
> 
> ```
> <variable name="abc">Abc</variable>
> <import abc from="index.md" as="selfRef" />
> 
> {{selfRef.abc}}
> 
> 
> 
> ```

Should this be possible? It will be impossible to resolve in some cases.

> During live changes: When the actual variable content is updated, the pages referencing it does not reflect the new value.
> 
> `test.md`
> 
> ```
> <variable name="to_update">Update me please!</variable>
> ```
> 
> `index.md`
> 
> ```
> <import to_update from="test.md" as="updateref" />
> 
> {{updateref.to_update}}
> 
> 
> 
> 
> ```
> 
> During live changes: When the actual variable is deeply nested inside another variable, making changes will cause the variable value to disappear (it does not matter if the changes are related to the variable or not).
> 
> `nest1.md`
> 
> ```
> <variable name="nest1">I am deeply nested.</variable>
> ```
> 
> `nest2.md`
> 
> ```
> <import nest1 from="nest1.md" as="nestref" />
> <variable name="nest2">{{nestref.nest1}}</variable>
> ```
> 
> `nest3.md`
> 
> ```
> <import nest2 from="nest2.md" as="nestref" />
> <variable name="nest3">{{nestref.nest2}}</variable>
> ```
> 
> `index.md`
> 
> ```
> <import nest3 from="nest3.md" as="nestref" />
> 
> {{nestref.nest3}}
> 
> 
> 
> 
> 
> ```
> 
> This is not a bug, but rather, a design problem. Importing variables from a non-existant file throws a really ugly error:
> 
> `index.md`
> 
> ```
> <import nosuchvar from="nosuchfile.md" as="ref" />
> 
> {{ref.nosuchvar}}
> ```
> 
> console
> 
> ```
> error:  message=ENOENT: no such file or directory, open 'C:\Users\Tan Wang Leng\Desktop\New folder\nosuchfile.md'
> Missing reference in: undefined, stack=Error: ENOENT: no such file or directory, open 'C:\Users\Tan Wang Leng\Desktop\New folder\nosuchfile.md'
> Missing reference in: undefined
>     at Object.fs.openSync (fs.js:646:18)
>     at Object.fs.readFileSync (fs.js:551:33)
>     at Parser._renderIncludeFile (C:\Users\Tan Wang Leng\Documents\GitHub\yamgent\markbind\src\lib\markbind\src\parser.js:226:40)
> ...
> ```

I will look into the live changes issues.

yamgent:
> If the as attribute for the namespace is specified, then all the variables can be used without explicitly declaring their names :`<import from="test.md" as="testRef"/>` would be enough.

Ok, noted, the problem is with the namespace scoping though, not the variables. :P

I presume you meant that namespaces can be omitted, so the `abc` in `test.md` can be overriding the `abc` in `index.md` in such a situation. That I think is fine. However, this behavior seems less intuitive  when a namespace is specified (I would have expected the imported variable to be scoped to `testRef` namespace, and not affect the `abc` in `index.md`, which is what my test case is about).

Perhaps you can discuss this with Prof further to see if this behavior is ok?

> My implementation requires a different namespace for each import.
> ...
> My intention was that you specify only either a namespace OR the list of variable names. If you specify this list of variable names without a namespace, they can be used like a normal variable {{varname}}. If you specify a namespace, then all variables inside the file can be accessed via {{namespace.varname}}. So if you specify both, you'll be able to use either method to access `varname`.

Ok, understood, same issue in terms of intuition.

Btw, you will also have to document these little details later on in the user guide when the feature is finalized.

> Should this be possible? It will be impossible to resolve in some cases.

Am thinking of this from the point of view of the authors (for example, it is quite possible that some Nunjucks logic, or some boilerplate codes, may result in the "from" being resolved to itself). My main goal here is to reduce confusion for the authors as much as possible, by getting rid of such edge cases.

However, if it is not critical and/or it is difficult to do, I have no issue with leaving it as it is for now.

damithc:
> Btw, you will also have to document these little details later on in the user guide when the feature is finalized.

Yes @openorclose, go ahead and add full documentation. Reading user docs is the only way I can evaluate the feature as I don't have the dev set up on my Computer.

docs/testing/child.md
```diff
@@ -0,0 +1,3 @@
+<variable name = "title">Title</variable>
+<variable name = "child_src">testimportvariables.md</variable>
+<variable name = "child_src_working">testimportvariables</variable>
```
yamgent: Remove the spaces around the `=`.

yamgent: Btw where's the HTML output for all these test cases? :P

docs/userGuide/syntax/variables.mbdf
```diff
@@ -103,6 +103,73 @@ You can specify a default value for a variable, which is displayed when the vari
...
+```html
+<variable name="address">123 Sun Avenue</variable>
+<variable name="name">Mark</variable>
+<variable name="phone">123456789</variable>
+```
+
+`coverpage.md`:
+
+```html
+<import address name from="page.md"/>
```
yamgent: `person.md`

docs/userGuide/syntax/variables.mbdf
```diff
@@ -103,6 +103,73 @@ You can specify a default value for a variable, which is displayed when the vari
...
+
+```
+
+will allow you to access the variables as per normal: <code>{<span></span>{address}}</code>, <code>{<span></span>{name}}</code>, <code>{<span></span>{phone}}</code>.
```
yamgent: You could collapse this by directly using it in `coverpage.md`, saves some words.

src/lib/markbind/src/parser.js
```diff
@@ -27,6 +27,11 @@ const ATTRIB_CWF = 'cwf';
 
 const BOILERPLATE_FOLDER_NAME = '_markbind/boilerplates';
 
+const PREFIX_TO_REDUCE_NAME_CLASH = '$__MARKBIND__';
```
yamgent: This acts like a "global variable prefix", maybe that would be a better name? The fact that it is done to reduce name clash should be a comment.

src/lib/markbind/src/parser.js
```diff
@@ -123,6 +128,18 @@ function extractIncludeVariables(includeElement, contextVariables) {
   return includedVariables;
 }
 
+function getImportedVariableMap(file) {
```
yamgent: Write a header comment for this method.

src/lib/markbind/src/parser.js
openorclose: Done. Named it `IMPORTED_VARIABLE_PREFIX`

src/lib/markbind/src/parser.js
openorclose: Done

docs/userGuide/syntax/variables.mbdf
openorclose: image

I think i'll keep it as it is, the change doesn't look very nice.

docs/userGuide/syntax/variables.mbdf
openorclose: Done

docs/testing/child.md
openorclose: I added text saying there shouldn't be any variables shown.

openorclose:
> I presume you are still working on the live server thing?

It should be working now!

yamgent:
Alright can you move the testing files from documentation to the functional testing sites?

openorclose:
image

What's this error? All the other html files show 2.5.2 but they still work.

docs/testing/child.md
```diff
@@ -0,0 +1,10 @@
+This page contains some variables that are being imported in other files.
+There should be only VARIABLES START and VARIABLES END in red:
+
+<div style="color:red">
+VARIABLES START
+<variable name="title">Title</variable>
+<variable name="child_src">testimportvariables.md</variable>
+<variable name="child_src_working">testimportvariables</variable>
+VARIABLES END
+</div>
```
yamgent: This file needs to be removed as well?

Other than this I have no other comments, I assume that the design is mostly fixed and this PR is no longer WIP.

openorclose:
@yamgent Ready for review!

PR #892, opened 2019-06-26
Fix temp styles not applying to navbars in header files

### LINKED ISSUE #889: Reduce FOUC
MarkBind suffers from [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) due its use of client-side rendering.

It would be good to eliminate or reduce this problem.

alyip98:
• [x] Bug fix

Alleviates #889 

Previously, #664 added temporary styles to `<navbar>` to prevent them from displaying as a list while Vue loads. However, after the layout update, the `insertTemporaryStyles` was called before `insertHeaderFile`, which kept the temporary styles from being applied to `<navbar>` in the header file.

This PR swaps the order of the two calls, such that `<navbar>` in the header file will also get the temporary styles applied.

ezgif-1-423f7783287c

**Testing instructions:**
netlify preview

Previously, a fix was implemented for reducing FOUC by applying
temporary styles for <navbar> and <dropdown>. A later commit broke this
fix by moving the header and footer files to a layout, which caused
the contents of those files to be missed by the temporary style fix.

This commit fixes this by inserting the temporary styles after
the header and footer files are inserted in the page, such that
the temporary styles are applied correctly to the relevant tags.


src/Page.js
```diff
@@ -799,8 +799,8 @@ Page.prototype.generate = function (builtFiles) {
       .then(result => addContentWrapper(result))
       .then(result => this.preRender(result))
       .then(result => this.insertSiteNav((result)))
-      .then(result => this.insertTemporaryStyles(result))
       .then(result => this.insertHeaderFile(result))
+      .then(result => this.insertTemporaryStyles(result))
```
openorclose: Should put this after insertHeaderFile? So that the header and footer are logically grouped together.

alyip98:
A more permanent solution can be explored with pre-rendering Vue components in the future. But for now this minor fix helps alleviate the FOUC issue

damithc:
I was wondering why the navbar was causing a ugly FOUC when it was minimized a few versions ago :-p

src/Page.js
Chng-Zhi-Xuan: Good suggestion, since there is a chance that `navbars` or `dropdown menus` are placed in a footer by the author.

Chng-Zhi-Xuan:
> A more permanent solution can be explored with pre-rendering Vue components in the future. But for now this minor fix helps alleviate the FOUC issue.

Might also want to revisit the original plan of displaying a spinner until the content + vue components have finished loading.

damithc:
> Might also want to revisit the original plan of displaying a spinner until the content + vue components have finished loading.

@marvinchin tried that. We eventually decided against it because it gave a 'slow site' feeling to the user.

Issue #901, opened 2019-06-26
Provide a way to specify relative dates


damithc:
In course websites, often dates are mentioned e.g., assignment submission dates. These websites tend to get reused over multiple offerings of the course in which case the author will have to go and update all the dates manually.

Suggestion: provide a way to specify dates relative to a specific reference date so that all dates can be updated by changing the reference date.

Example: `<date ref="sem_start" week="2" day="Monday" />` will resolve to something like `Monday, January 21` (`sem_start` is defined somewhere else, and given the date of week 1 Monday). Alternatively, it can be a predefined JS function or a nunjucks macro. e.g., `{{ show_date("sem_start", 2, 1) }}`

PR #908, opened 2019-06-27
Add relative date feature

alyip98:
• [x] New feature

Addresses #901 

Adds the ability to insert relative dates

~Add a new markdown-it plugin that looks for `@date Week#day@` tokens and replaces them with 
dates relative to a specified `base` date.~

Changed to a Nunjucks filter extension:
`{{ baseDate | date(format, dayOffset) }}`

Is the proposed syntax suitable, or should we implement it as a tag e.g. `<date>`

**Testing instructions:**
[Netlify preview](https://deploy-preview-908--markbind-master.netlify.com/userguide/fullsyntaxreference#dates)

Add a new nunjucks filter to allow dynamic date calculation given
 a base date and day offset and customizable formatting of the output.

### DIFF
--- a/docs/userGuide/syntax/dates.mbdf
+++ b/docs/userGuide/syntax/dates.mbdf
@@ -1,0 +1,107 @@
+## Dates
+
+{% macro njcode(raw) %}<code>{<a/>{ {{ raw }} }}</code>{% endmacro %}
+{% macro njblock(raw) %}<code>{<a/>% {{ raw }} %}</code>{% endmacro %}
+
+**Markbind supports date formatting and simple calculations** as a Nunjucks [filter](https://mozilla.github.io/nunjucks/templating.html#filters).
+
+**Syntax:** {{ njcode('baseDate | date(format, daysToAdd)') }}
+
+<span id="main-example">
+
+20 days after 1st Jan 2020:
+
+{{ njcode('"2020-01-01" | date("ddd, Do MMM, YYYY", 20) ') }} :glyphicon-arrow-right: {{ "2020-01-01" | date("ddd, Do MMM, YYYY", 20) }}
+</span>
+
+<box type="info">
+
+The baseDate follows the format: `YYYY-MM-DD`
+
+The default output format is `"ddd D MMM"` e.g. Fri 6 Mar
+</box>
+
+### Using variables
+
+{{ njblock('set base1 = "2020-01-01"') }} {% set base1 = "2020-01-01" %} <br/>
+{{ njblock('set format1 = "DD MM YYYY"') }} {% set format1 = "DD MM YYYY" %} <br/>
+{{ njblock('set format2 = "ddd Do MMM (DD/MM/YYYY)"') }} {% set format2 = "ddd Do MMM (DD/MM/YYYY)" %}
+
+{{ njcode('base1 | date') }} :glyphicon-arrow-right: {{ base1 | date }}<br/>
+
+#### Custom formatting
+{{ njcode('base1 | date(format1)') }} :glyphicon-arrow-right: {{ base1 | date(format1) }}<br/>
+
+#### Adding days
+{{ njcode('base1 | date(format2, 0)') }} :glyphicon-arrow-right: {{ base1 | date(format2, 0) }}<br/>
+{{ njcode('base1 | date(format2, 10)') }} :glyphicon-arrow-right: {{ base1 | date(format2, 10) }}<br/>
+
+#### Page variables
+Dates can be supplied using [page variables](../reusingContents.html#variables) for convenience.
+
+Inside `variables.md` or referencing page:
+```
+<variable name="date_pagevar">2020-03-06</variable>
+```
+<variable name="date_pagevar">2020-03-06</variable>
+
+{{ njcode('date_pagevar | date(format2)') }} :glyphicon-arrow-right: {{ date_pagevar | date(format2) }} <br/>
+
+
+
+### Advanced Formatting
+
+The output date can be formatted to suit your needs by specifying a format string as an argument to the date filter.
+
+Default format: `"ddd D MMM"` e.g. Fri 6 Mar
+
+<panel header="**Brief reference**">
+
+Token | Output
+- | -
+D | 1
+Do | 1st
+DD | 01
+M | 1
+MM | 01
+MMM | Jan
+MMMM | January
+YY | 19
+YYYY | 2019
+</panel>
+
+Full formatting reference available [here](https://momentjs.com/docs/#/displaying/format/).
+
+{{ icon_example }}
+<span id="examples" class="d-none">
+<include boilerplate src="outputBox.md">
+<span id="code">
+
+<box><span>
+{{ njblock('set base1 = "2019-08-12"') }}<br/>
+{{ njblock('set format1 = "DD MM YYYY"') }}<br/>
+{{ njblock('set format2 = "ddd Do MM"') }}<br/>
+{{ njcode('base1 | date') }} `<!-- Mon 12 Aug -->`<br/>
+{{ njcode('base1 | date(format1)') }} `<!-- 12 08 2019 -->`<br/>
+{{ njcode('base1 | date(format1, 10)') }} `<!-- 22 08 2019 -->`<br/>
+{{ njcode('base1 | date(format2, 10)') }} `<!-- Thu 22/08 -->`<br/>
+</span></box>
+
+</span>
+<span id="output">
+Mon 12 Aug<br/>
+12 08 2019<br/>
+22 08 2019<br/>
+Thu 22/08
+</span>
+</include>
+
+<span id="short" class="d-none">
+
+<box><span>
+{{ njcode('"2019-08-12" | date("DD.MM.YYYY", 10)') }} `<!-- 22.08.2019 -->`<br/>
+</span></box>
+
+{{ "2019-08-12" | date("DD.MM.YYYY", 10) }}
+</span>
+</span>
--- a/src/Page.js
+++ b/src/Page.js
@@ -2,10 +2,9 @@ const cheerio = require('cheerio');
 const fm = require('fastmatter');
 const fs = require('fs-extra-promise');
 const htmlBeautify = require('js-beautify').html;
-const nunjucks = require('nunjucks');
 const path = require('path');
 const Promise = require('bluebird');
-const nunjuckUtils = require('./lib/markbind/src/utils/nunjuckUtils');
+const njUtil = require('./lib/markbind/src/utils/nunjuckUtils');
 
 const _ = {};
 _.isString = require('lodash/isString');
@@ -409,12 +408,8 @@ class Page {
  * @param pageData a page with its front matter collected
  */
   generateExpressiveLayout(pageData, fileConfig) {
-    const nj = nunjucks;
     const markbinder = new MarkBind({
       errorHandler: logger.error,
-    });
-    nj.configure({
-      autoescape: false,
     });
     const template = {};
     template[LAYOUT_PAGE_BODY_VARIABLE] = pageData;
@@ -440,7 +435,7 @@ class Page {
     // Retrieve Expressive Layouts page and insert content
       fs.readFileAsync(layoutPagePath, 'utf8')
         .then(result => markbinder.includeData(layoutPagePath, result, layoutFileConfig))
-        .then(result => nunjuckUtils.renderEscaped(nj, result, template))
+        .then(result => njUtil.renderRaw(result, template, {}, false))
         .then((result) => {
           this.collectIncludedFiles(markbinder.getDynamicIncludeSrc());
           this.collectIncludedFiles(markbinder.getStaticIncludeSrc());
@@ -481,7 +476,7 @@ class Page {
     // Map variables
     const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
     const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${nunjuckUtils.renderEscaped(nunjucks, headerContent, userDefinedVariables)}\n${pageData}`;
+    return `${njUtil.renderRaw(headerContent, userDefinedVariables)}\n${pageData}`;
   }
 
   /**
@@ -511,7 +506,7 @@ class Page {
     // Map variables
     const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
     const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${pageData}\n${nunjuckUtils.renderEscaped(nunjucks, footerContent, userDefinedVariables)}`;
+    return `${pageData}\n${njUtil.renderRaw(footerContent, userDefinedVariables)}`;
   }
 
   /**
@@ -548,7 +543,7 @@ class Page {
     // Map variables
     const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
     const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    const siteNavMappedData = nunjuckUtils.renderEscaped(nunjucks, siteNavContent, userDefinedVariables);
+    const siteNavMappedData = njUtil.renderRaw(siteNavContent, userDefinedVariables);
     // Convert to HTML
     const siteNavDataSelector = cheerio.load(siteNavMappedData);
     if (siteNavDataSelector('navigation').length > 1) {
@@ -698,12 +693,12 @@ class Page {
       // Map variables
       const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
       const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-      const headFileMappedData = nunjuckUtils.renderEscaped(nunjucks, headFileContent, userDefinedVariables)
+      const headFileMappedData = njUtil.renderRaw(headFileContent, userDefinedVariables)
         .trim();
       // Split top and bottom contents
       const $ = cheerio.load(headFileMappedData, { xmlMode: false });
       if ($('head-top').length) {
-        collectedTopContent.push(nunjuckUtils.renderEscaped(nunjucks, $('head-top')
+        collectedTopContent.push(njUtil.renderRaw($('head-top')
           .html(), {
           baseUrl,
           hostBaseUrl,
@@ -714,7 +709,7 @@ class Page {
         $('head-top')
           .remove();
       }
-      collectedBottomContent.push(nunjuckUtils.renderEscaped(nunjucks, $.html(), {
+      collectedBottomContent.push(njUtil.renderRaw($.html(), {
         baseUrl,
         hostBaseUrl,
       })
@@ -813,7 +808,7 @@ class Page {
           this.addLayoutFiles();
           this.collectHeadFiles(baseUrl, hostBaseUrl);
 
-          this.content = nunjucks.renderString(this.content, {
+          this.content = njUtil.renderString(this.content, {
             baseUrl,
             hostBaseUrl,
           });
@@ -1069,7 +1064,7 @@ class Page {
                                                                               this.baseUrlMap);
           const baseUrl = relative ? `${this.baseUrl}/${utils.ensurePosix(relative)}` : this.baseUrl;
           const hostBaseUrl = this.baseUrl;
-          const content = nunjucks.renderString(result, {
+          const content = njUtil.renderString(result, {
             baseUrl,
             hostBaseUrl,
           });
--- a/src/Site.js
+++ b/src/Site.js
@@ -2,13 +2,12 @@ const cheerio = require('cheerio');
 const fs = require('fs-extra-promise');
 const ghpages = require('gh-pages');
 const ignore = require('ignore');
-const nunjucks = require('nunjucks');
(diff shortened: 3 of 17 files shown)


alyip98:
This is still an early WIP, but I want to get some opinions on where the `base` date should be defined, and whether the syntax is suitable/user friendly

damithc:
Note that the date may need to be embedded in other text. e.g., `Remember to submit by the deadline (Jan 12, 2019) in the right format ...`

It would be nice if a site can support multiple timelines too. The current approach seems to be limited to one timeline per site (because the reference date is fixed to the `base` variable)

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+);
+
+/**
+ * Parses and extracts the week, day offset and format from the raw matched expression
+ * @param raw raw matched expression
+ */
+function parseInput(match) {
+  // Parse week number
+  if (isNaN(match[1])) {
```
openorclose: Will this ever be true, since the regex only matches digits? Also can change the regex for this to `\d\d?` since week should only be 2 digits.

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
```diff
@@ -0,0 +1,76 @@
...
+    return showError(WEEK_NAN, raw);
+  }
+  const week = parseInt(match[1]);
+
+  // Parse day
+  let day = match[2].toLowerCase();
+  if (isNaN(day)) {
```
openorclose: Should check the length instead, 3 character / 1 character. `\w{3}` also matches 3 digits. which may not be intended.

damithc:
Shall we revive this PR @alyip98

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
openorclose: If day is `001` it'll still be correct and interpreted as `1`. Is this intended? Probably not a big deal either way

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
```diff
@@ -0,0 +1,97 @@
...
+  let day = match[3].toLowerCase();
+  if (isNaN(day)) {
+    if (days.includes(day)) {
+      // Maps 'mon' => 1, 'tue' => 2 and so on
+      day = days.indexOf(day) + 1;
+    } else {
+      return formatError(DAY_INVALID, raw);
+    }
+  } else {
+    day = parseInt(day);
```
openorclose: Still recommended to have the base (and i think we do include the base in other code)
```suggestion
    day = parseInt(day, 10);
```

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
```diff
@@ -0,0 +1,97 @@
...
+    if (day < 1 || day > 7) {
+      return formatError(DAY_OUT_OF_RANGE, raw);
+    }
+  }
+
+  let formatPattern = match[4] || DEFAULT_FORMAT;
+  formatPattern = formatPattern === '' ? DEFAULT_FORMAT : formatPattern;
```
openorclose: Should be able to reduce to
```suggestion
  const formatPattern = match[4] === '' ? DEFAULT_FORMAT : formatPattern;
```
?

alyip98: match[4] could be empty string or undefined

src/lib/markbind/src/lib/markdown-it/markdown-it-date.js
alyip98: It defaults to 10, and in this case it's impossible to have 0x in the string so it seems unnecessary

alyip98:
Ready for review!

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,142 @@
...
+<include src="outputBox.md" boilerplate >
+<span id="code">
+
+```
+<variable name="sem_start">2019-08-12</variable>
+<variable name="recess_week">2019-09-23</variable>
+
+The semester starts on @date:{{ sem_start }} W1#1@
+
+Lab 1 is due on @date:{{ sem_start }} W3#5@
```
openorclose: Need to find a way to escape the mustache syntax.

alyip98:
Updated the PR with details regarding the new implemenation, i.e. syntax is now `{{ baseDate | date(format, dayOffset) }}` and it leverages nunjucks rather than adding a new "tag" symbol.

Please let me know if there are any concerns or improvements

test/functional/test_site/testDates.md
```diff
@@ -0,0 +1,27 @@
...
+{{ base1 | date(format2, 10) }} should be Thu 22/08
+
+
+{{ base2 | date }} should be Tue 1 Jan
+
+{{ base2 | date(format1) }} should be 01 01 2019
+
+{{ base2 | date(format1, 10) }} should be 11 01 2019
+
+{{ base2 | date(format2, 10) }} should be Thu 11/01
```
yamgent: 11/01/2019 is a Friday

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,105 @@
...
+
+<span id="main-example">
+<include src="outputBox.md" boilerplate >
+<span id="code">
+
+<box><span>
+{{ njcode('set base1 = "2019-08-12"') }}<br/>
+{{ njcode('set format1 = "DD MM YYYY"') }}<br/>
+{{ njcode('set format2 = "ddd DD/MM"') }}<br/>
+{{ njcode('base1 | date') }} // Mon 12 Aug<br/>
```
yamgent: Minor nit: Use `<!-- -->` instead of `//`

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,105 @@
...
+MMMM | January
+YY | 19
+YYYY | 2019
+
+Full formatting reference available [here](https://momentjs.com/docs/#/displaying/format/).
+
+Example
+<span id="examples" class="d-none">
+<include boilerplate src="outputBox.md">
+<span id="code">
```
yamgent: The code example here did not have the box appearance in the final render, unlike the other one.

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,105 @@
+## Dates
+
+{% macro njcode(raw) %}<code>{<a/>{ {{ raw }} }}</code>{% endmacro %}
```
yamgent: Not a change request, just an enquiry: did you came up with this? This syntax seems to be pretty useful for our other parts of the documentation.

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,105 @@
...
+
+{{ njcode('set format2 = "ddd DD/MM"') }}
+
+{{ njcode('base1 | date') }} // Mon 12 Aug
+
+{{ njcode('base1 | date(format1)') }} // 12 08 2019
+
+{{ njcode('base1 | date(format1, 10)') }} // 22 08 2019
+
+{{ njcode('base1 | date(format2, 10)') }} // Thu 22/08
```
damithc: If the base is defined in a different page, it needs to be imported first as per normal nunjucks practice?
Also, if this `set` is different from the normal nunjucks `set` (`{% set format1 = "DD MM YYYY" %}`), may be use a different keyword to avoid confusion?

docs/userGuide/syntax/dates.mbdf
alyip98: Yup

docs/userGuide/syntax/dates.mbdf
alyip98: It's the normal nunjucks `set`. The base could also be defined in the global `variables.md` or in another page and imported normally as a nunjucks variable. The syntax above is a macro I created to format nunjucks code in Markbind pages.

damithc: What I mean is, the syntax you give here (as shown in the generated UG) is _almost_ the same as the nunjucks syntax to set variables, not not exactly the same.
normal: `{% set format1 = "DD MM YYYY" %}`
vs: `{{ set format1 = "DD MM YYYY" }}`

image

Are they equivalent? Or just a mistake in your documentation?

alyip98: Oh my bad that was a mistake

src/lib/markbind/src/lib/nunjucks-extensions/nunjucks-date.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Calculates the result of adding {days} days to the base date
+ * @param base
+ * @param days
+ */
+function calculateTarget(base, days) {
+  return Moment(base).add(days, 'days');
+}
+
+function filter(baseDate, format = DEFAULT_FORMAT, day = 0) {
```
openorclose: Is it possible to allow the user to set a `DEFAULT_FORMAT`? Maybe somewhere in site.json.

Then we can change the arguments to `filter` to `filter(baseDate, day = 0, format = DEFAULT_FORMAT)`.

This is assuming that the most frequent use case would be the author having only 1 format they want to use, so being able to do `{{ base | date(7*2 + 1) }}` and avoid typing the format all the time might be easier syntax.

alyip98: Fair point, I suppose if there really was a need we could also split the filter into 2 separate filters in a future PR, like `{{ baseDate | addDays(7 * 2 + 1) | formatDate(format) }}`.

Configuring a custom `DEFAULT_FORMAT` isn't too hard but I'm not sure if it will be very useful. I don't think that it matters enough either way, since you could just pass in an extra argument. An alternative is to define a nunjucks macro (`customFormat` defined in `variables.md`)
```
{% macro customDate(baseDate, days) %}
{{ baseDate | date(customFormat, days) }}
{% endmacro %}

Usage: {{ customDate(baseDate, 5) }}
```

damithc:
image

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,98 @@
+## Dates
+
+{% macro njcode(raw) %}<code>{<a/>{ {{ raw }} }}</code>{% endmacro %}
+{% macro njblock(raw) %}<code>{<a/>% {{ raw }} %}</code>{% endmacro %}
+
+**Markbind supports date formatting and simple calculations** as a Nunjucks filter.
```
ang-zeyu: Might be nice to link to https://mozilla.github.io/nunjucks/templating.html#filters

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,98 @@
...
+12 08 2019
+
+22 08 2019
+
+Thu 22/08
+</span>
+</include>
+
+<box type="info">
+The base date follows the format: `YYYY-MM-DD`
```
ang-zeyu: Needs a new line before, or could wrap with `<md>`

src/lib/markbind/src/lib/nunjucks-extensions/nunjucks-date.js
ang-zeyu: Could we still allow this, but without the configuration? :o 
( swapping the `day` and `format` in `format = DEFAULT_FORMAT, day = 0) {` )

src/lib/markbind/src/parser.js
```diff
@@ -1,10 +1,11 @@
 const cheerio = require('cheerio');
 const fs = require('fs');
 const htmlparser = require('htmlparser2'); require('./patches/htmlparser2');
-const nunjucks = require('nunjucks');
 const path = require('path');
 const Promise = require('bluebird');
 const slugify = require('@sindresorhus/slugify');
+const nunjucks = require('nunjucks').configure()
```
ang-zeyu: there's a few usages in `site.js` and `page.js` as well; Perhaps a common `nunjucksUtils` that does this and reexports to the three files could be used ( something like #1049 )

test/unit/parser.test.js
```diff
@@ -4,6 +4,7 @@ const MarkBind = require('../../src/lib/markbind/src/parser.js');
 const { USER_VARIABLES_DEFAULT } = require('./utils/data');
 
 jest.mock('fs');
+jest.mock('../../src/util/logger');
```
ang-zeyu: 👀

src/lib/markbind/src/lib/nunjucks-extensions/nunjucks-date.js
```diff
@@ -0,0 +1,19 @@
...
+  return calculateTarget(baseDate, day)
+    .format(format);
```
ang-zeyu: let's squash this to one line since there's only one level of chaining

docs/userGuide/syntax/dates.mbdf
ang-zeyu: `baseDate`

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,98 @@
...
+
+It can be supplied using [page variables](../reusingContents.html#variables) for convenience.
+</box>
+
+
+### Formatting
+
+The date can be formatted to suit your needs
```
ang-zeyu: Could mention `output date` in particular! Missing a period as well

test/functional/test_site/testDates.md
```diff
@@ -0,0 +1,27 @@
+## Dates
```
ang-zeyu: Could we move some of these to `variables.md` to ensure they work in `site.js`?

src/lib/markbind/src/parser.js
alyip98: We could do nunjucksUtils in a separate PR. For now I will repeat the `configure()` pattern across the different instances.

test/unit/parser.test.js
alyip98: This was needed back when the markdown-it plugin logged errors. The current plugin fails silently and prints "Invalid Date". I'll remove it since it is no longer needed

src/Page.js
```diff
@@ -409,7 +411,7 @@ class Page {
  * @param pageData a page with its front matter collected
  */
   generateExpressiveLayout(pageData, fileConfig) {
-    const nj = nunjucks;
+    const nj = _nunjucks;
```
crphang: Nunjucks instance in expressive layouts should require date filter as well I believe.

test/functional/test_site/testDates.md
ang-zeyu: Could move some of the filters to `variables.md` as well, to ensure the filter works in it

```
{{ base2 | date }} should be Tue 1 Jan

{{ base2 | date(format1) }} should be 01 01 2019
```
maybe

src/lib/markbind/src/parser.js
ang-zeyu: I'm fine with this, #1049 has been merged though, depending on your conflict resolution goes might be nice to do it here instead 😁

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,99 @@
...
+M | 1
+MM | 01
+MMM | Jan
+MMMM | January
+YY | 19
+YYYY | 2019
+
+Full formatting reference available [here](https://momentjs.com/docs/#/displaying/format/).
+
+Example
```
ang-zeyu: could use `{{ icon_example }}`

test/functional/test_site/_markbind/variables.md
```diff
@@ -9,3 +9,5 @@
 <variable name="global_variable_overriding_included_variable">Global Variable Overriding Included Variable</variable>
 <variable name="global_variable">Global Variable</variable>
 <variable name="page_global_variable_overriding_page_variable">Global Variable Overriding Page Variable</variable>
+
+<variable name="base2">2019-01-01</variable>
```
ang-zeyu: the `variables.md` filter test seems to have disappeared 👀

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -7,8 +13,14 @@ function preEscapeRawTags(pageData) {
 }
 
 module.exports = {
-  renderEscaped(nunjucks, pageData, variableMap = {}, options = {}) {
+  renderEscaped(pageData, variableMap = {}, options = {}, env = commonEnv) {
     const escapedPage = preEscapeRawTags(pageData);
-    return nunjucks.renderString(escapedPage, variableMap, options);
+    return env.renderString(escapedPage, variableMap, options);
+  },
+  render(pageData, variableMap = {}, options = {}, env = commonEnv) {
```
ang-zeyu: how about having a separate `renderUnescaped`? Would avoid having to export the `env` configurations which could be kept in this file.

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -7,8 +13,14 @@ function preEscapeRawTags(pageData) {
...
+    return env.renderString(pageData, variableMap, options);
   },
+  nunjucks,
+  commonEnv,
```
ang-zeyu: are there any files using the `nunjucks` / `commonEnv` exports?

src/Site.js
```diff
@@ -120,8 +119,8 @@ class Site {
 
     // Page template path
     this.pageTemplatePath = path.join(__dirname, PAGE_TEMPLATE_NAME);
-    const env = nunjucks.configure({ autoescape: false });
-    this.pageTemplate = nunjucks.compile(fs.readFileSync(this.pageTemplatePath, 'utf8'), env);
+    this.pageTemplate = njUtil.nunjucks.compile(
+      fs.readFileSync(this.pageTemplatePath, 'utf8'), njUtil.unescapedEnv);
```
ang-zeyu: Could add a `compile` method in nunjucksUtils for this to avoid exporting `unescapedEnv` / `nunjucks`

src/lib/markbind/src/utils/nunjuckUtils.js
ang-zeyu: or you could change the fourth option `env` to `autoEscape = false` and do the rendering accordingly

In any case I think its not such a good idea to have a utility file requiring configuration with another one of its exports ( `unescapedEnv` )

alyip98: I was kind of on the fence with this since there are two kinds of escaping (`{%raw%}` vs HTML characters) in here, I also wanted to keep the original signature of accepting env as an argument. I suppose it would be clearer to simply make a separate method that uses the unescapedEnv, following the first suggestion. But we would have to name it something other than `renderUnescaped` since it implies that it is the opposite of `renderEscaped` which it isn't.

ang-zeyu: > To summarize:

Sounds good 👍 I would personally leave `renderUnescaped` as an option in `renderRaw` so that its clearer that we are adding the extra `{% raw/endraw %}`'s in the call though. Since its actually more of `renderRaw+NoAutoEscape`.

alyip98: Good suggestion, it's clearer and less work

test/functional/test_site/_markbind/variables.md
ang-zeyu: Something like ` <variable name=“filtered”> {{ base2 | date(format1, 10) }}</variable> `, then using it in the page
Rationale being that the variables.md variables are resolved separately; would be good to ensure the call to njUtil is there

test/functional/test_site/testDates.md
```diff
@@ -0,0 +1,26 @@
...
+
+{{ base2 | date }} should be Tue 1 Jan
+
+{{ base2 | date(format1) }} should be 01 01 2019
+
+{{ base2 | date(format1, 10) }} should be 11 01 2019
+
+{{ base2 | date(format2, 10) }} should be Fri 11/01
+
+{{ formatted_date }} should be Tue 1 Jan
```
ang-zeyu, line 26: `Fri 11 Jan`

LGTM otherwise, nice work on the njutil refactor 👍

docs/userGuide/syntax/dates.mbdf
```diff
@@ -0,0 +1,99 @@
...
+
+Thu 22/08
+</span>
+</include>
+
+<box type="info">
+
+The baseDate follows the format: `YYYY-MM-DD`
+
+It can be supplied using [page variables](../reusingContents.html#variables) for convenience.
```
damithc, line 41: Can elaborate on this?

alyip98: The format portion or the page variables portion?

damithc: page variable portion. May be give an example?

alyip98:
> Noticed a few more doc nits, could polish it up here or in a separate PR
> 
> * could standardise with the level of subheadings used elsewhere in the page `h2 -> h4 -> h5`
> * `#### Custom formatting` seems to be a repeat of `### Advanced formatting`
> * Could combine `page variables` into `using variables`

Let's do this in a separate PR, it can be a good first issue

ang-zeyu:
> Let's do this in a separate PR, it can be a good first issue

👍, seems good to be merged in that case. Do open up an issue!

damithc:
Seems to be working in production. Good work @alyip98 and reviewers. It will be very useful for the module websites next semester.

PR #919, opened 2019-07-01
Fix PlantUML include bug

### LINKED ISSUE #918: Error message when including a file with a pUML diagram
v2.5.2

Scenario:
```
[root]
- diagram.puml
-topics/page.md
-index.md
```
page.md has a `<puml>` tag for the `diagram.puml`
`index.md` includes `topics/page.md`

Problem: the diagram is generated correctly but there is an error message in the console `error: Error reading [root]\diagram.puml`

Apparently, it's trying to read the page from the wrong location in addition to reading it from the correct location.

alyip98:
• [x] Bug fix

Fixes #918 

Fixes include bug with PlantUML

Resolves `src` with respect to original file location, rather than including file location.

n/a

**Testing instructions:**
I have prepared a test site with 3 layers of folders and nested includes
[puml-test.zip](https://github.com/MarkBind/markbind/files/3343922/puml-test.zip)

PlantUML plugin throws an error while trying to process diagrams 
within an `<include>`. The original implementation resolves 
diagram paths w.r.t. `sourcePath` rather than `cwf`.

`sourcePath` refers to the path of the file being processed
 (in this case the file doing the include) and `cwf` refers to the file
 being included.

This commit fixes that by preferring `cwf` where available (within
included file), and using `sourcePath` when `cwf` is unavailable
(within normal files).


damithc:
> Thanks for the pr! Want to check if this works with boiler plates. For example, if a boilerplate file contains a `puml` tag

Good question. If it doesn't, we'll have to fix that as a separate PR as I can imagine the feature being useful even if it doesn't work in boilerplates.

alyip98:
After looking into it, I realized that there was a bug when a boilerplate containing a `puml` tag is used in a file and said file is included via an `include` tag in another file.

For example:
index.md: `<include src="chapter1/text.md />"`
chapter1/text.md: `<include boilerplate src="puml-boilerplate.md" />`
puml-boilerplate.md: `<puml src="diagram.puml" />`

Expected: the final diagram src refers to `chapter1/diagram.puml`
Actual: src refers to `/diagram.puml`

I will look into the cause of the problem and submit a fix, but it might be later as I am working on another project.

damithc:
> I will look into the cause of the problem and submit a fix, but it might be later as I am working on another project.

Let's do that as a separate PR. The chance of using puml files in a boilerplate is low. Hence, fixing it is not urgent.

damithc:
> Let's do that as a separate PR. 

Created #926

alyip98:
Ok, updated commit message and proposed commit message

alyip98:
Ok, I fixed that bug @yamgent discovered, I got confused by the relative paths and stuff but its all good now, it also resolved the boilerplate issue I mentioned earlier. Thanks for the review!

PR #920, opened 2019-07-01
Add footnotes to userGuide/formattingContents

### LINKED ISSUE #914: Add footnotes to 'Formatting Contents' page
Current: footnotes are explained in https://markbind.org/userGuide/markBindSyntaxOverview.html

Problem: That page is just an overview page, giving some examples only. Furthermore, that page does not come up when searching for `footnotes` keyword.

Suggested: describe the feature in https://markbind.org/userGuide/formattingContents.html instead.

openorclose:

- [x] Enhancement to an existing feature

Fixes #914 

Change fullsyntaxreference such that footnotes appear in the Formatting Contents section.

Better wording of the feature?


openorclose:
> Thanks for the PR! This is a good change but you missed out one part: Since its added as a reader facing feature, foot notes will be featured in the [read facing features page](https://deploy-preview-920--markbind-master.netlify.com/userguide/readerfacingfeatures#footnotes) which requires an example otherwise it will be stated as `this is not updated`
> 
> It may be hard to add an example so potential solutions would be to leave it empty, direct the user to bottom of the page (some how) or just have a dummy example of the foot note

Thanks! I added a short example for footnotes.

PR #922, opened 2019-07-04
Configure nunjucks to always escape variables

### LINKED ISSUE #907: Treat nunjucks variables as safe by default
Current: if a variable has html syntax in it e.g., `<span id="x"><br/></span>`, it needs to be used with the `safe` filter e.g., `{{ x | safe }}`.

Suggestion: set default mode to safe so that the `safe` filter is not required

Relevant nunjucks docs:
image

openorclose:

Fixes #907 

Make nunjucks autoescape all variables.

Configure nunjucks, setting `autoescape` to `true.

https://mozilla.github.io/nunjucks/getting-started.html

If the following is in a nunjucks variable `v`,
```html
<span id="x"><br/></span>,
```
It does not need to be escaped with `{{ v | safe }}` anymore.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -3,8 +3,6 @@ const fm = require('fastmatter');
 const fs = require('fs-extra-promise');
 const htmlBeautify = require('js-beautify').html;
 const nunjucks = require('nunjucks');
-
-nunjucks.configure({ autoescape: true });
 const path = require('path');
 const pathIsInside = require('path-is-inside');
 const Promise = require('bluebird');
@@ -51,6 +49,7 @@ const TEMP_DROPDOWN_PLACEHOLDER_CLASS = 'temp-dropdown-placeholder';
 
 cheerio.prototype.options.xmlMode = true; // Enable xml mode for self-closing tag
 cheerio.prototype.options.decodeEntities = false; // Don't escape HTML entities
+nunjucks.configure({ autoescape: true });
 
 function Page(pageConfig) {
   this.asset = pageConfig.asset;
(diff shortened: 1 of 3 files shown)


yamgent:
Ok, there's something really odd about this PR.

There is a misunderstanding of what the term "escaping" here means. If escaping is enabled, then `<br/>` will always be converted to `&lt;br/&gt;` (that's what escaping means).

With the above definition, `autoescape` should actually be set to `false`, not `true`, in order to disable escaping.

Yet, escaping is disabled anyway even with `autoescape` set to `true`!

This means that the fix in this PR actually won't fix the original problem. In fact, if you use the test case that you posted in the first post, on the `master` branch, you didn't need the `safe` keyword either (which implies that something somewhere else has already disabled the escaping?).

So merging this PR would probably not fix the problem in #907?

yamgent:
To prove my point, you can use this test code to play around with `autoescape`:

```js
const nunjucks = require('nunjucks');

const variableContent = '<a href="https://www.google.com">Test</a> <br/>';

nunjucks.configure({ autoescape: true });   // change this to play around with autoescape
const rendered = nunjucks.renderString('{{variable}}', { variable: variableContent });

console.log(rendered);
```

---

Setting `autoescape` to `true` gives:

```html
&lt;a href=&quot;https://www.google.com&quot;&gt;Test&lt;/a&gt; &lt;br/&gt;
```

Setting `autoescape` to `false` gives:

```html
<a href="https://www.google.com">Test</a> <br/>
```

damithc:
@openorclose any decision on this PR?

openorclose:
Should we close it? Since this PR was created as  I misunderstood how autoescape worked.

damithc:
> Should we close it? Since this PR was created as I misunderstood how autoescape worked.

Sure.

PR #927, opened 2019-07-11
Add temp styles to panel to reduce FOUC

### LINKED ISSUE #921: FOUC caused by collapsed panels
Problem: collapsed panels in a page first load in the expanded form and auto-collapse at some later point in the page loading process, causing a FOUC-like effect

collapsed

I wonder if it is possible to collapse the panel from the start so that there is no jarring change in the page structure

openorclose:

Fixes #921 

Unexpanded panels show their content temporarily until Vue loads
completely.

Let's add some temp styles to panel tags that reduce this FOUC.

View the preview site, see if there is a flash of the unexpanded panel on page load.

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -198,28 +198,31 @@ so we apply temporary styles to fix the layout until Vue correctly updates the D
     display: none;
 }
 
-/* Hide the contents of unexpanded panels*/
-
+/* Hide the contents of unexpanded panels */
+/* stylelint-disable */
 panel:not([expanded]) {
+    /* stylelint-enable */
     visibility: hidden;
     height: 50px;
     display: block;
 }
 
+/* stylelint-disable */
 panel:not([expanded])::before {
+    /* stylelint-enable */
     content: attr(header);
     visibility: visible;
     position: relative;
-    display:block;
+    display: block;
 
     /* Below values are copied from the final rendered styles */
 
     margin-top: 5px;
-    padding: .75rem 1.25rem;
+    padding: 0.75rem 1.25rem;
     margin-bottom: 0;
-    background-color: rgba(0,0,0,.03);
-    border: 1px solid rgba(0,0,0,.125);
-    border-radius: .25rem;
+    background-color: rgba(0, 0, 0, 0.03);
+    border: 1px solid rgba(0, 0, 0, 0.125);
+    border-radius: 0.25rem;
 }
 
 /* Footnote anchor */
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -284,6 +284,11 @@ panel[type=dark]:not([expanded])::before {
 panel[minimized]:not([expanded])::before {
     background-color: var(--light);
 }
+
+/* increasing specificity */
+#flex-body panel:not([expanded]) panel {
+    display: none;
+}
 /* stylelint-enable selector-type-no-unknown */
 
 /* Footnote anchor */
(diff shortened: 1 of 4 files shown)


openorclose:
Seems like this doesn't play well with markdown in the header attribute...

See https://5d26dd360a052f000a7331ae--markbind-master.netlify.com/userguide/fullsyntaxreference

openorclose:
Edited styles: Panels now show `Loading...` instead of their possibly unformatted headers, and minimized and coloured panels are accounted for.

asset/css/markbind.css
```diff
@@ -198,28 +198,31 @@ so we apply temporary styles to fix the layout until Vue correctly updates the D
     display: none;
 }
 
-/* Hide the contents of unexpanded panels*/
-
+/* Hide the contents of unexpanded panels */
+/* stylelint-disable */
```
Chng-Zhi-Xuan: You should use `/* stylelint-disable-next-line */`. Preferably with the specific rule disabled.

Likewise for the other lines below.

[Documentaton here](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#turning-rules-off-from-within-your-css)

Chng-Zhi-Xuan:
Some observations from your current solution:

1. `minimized` panels will have a blank empty square while "loading" with no text
1. `expanded` panels will have their text shown but header is hidden
1. normal panels with large amount of text will have a large blank section with "loading ..."

The third point may cause a "FOUC" as the blank section gets removed and the content from below gets pushed up.

Maybe you could try this approach (if possible):

- Apply the hidden class to the body of the panel but not the header
- Replace the header text with "loading ..." then back after it is loaded.

This should alleviate the blank section problem and the earlier problem of un-rendered markdown header text.

asset/css/markbind.css
openorclose: Ok, I disabled only the selector-type-no-unknown rule.

openorclose:
> 1. `minimized` panels will have a blank empty square with no text

This can't be fixed properly, as the size of the element depends on the rendered content, so will definitely cause FOUC.

> 2. `expanded` panels will have their text shown but header is hidden

I didn't add any classes to prevent this.

> 3. normal panels with large amount of text will have a large blank section with "loading ..."

Ok, fixed it to only take up the space of the header.

yamgent:
> View the preview site, see if there is a flash of the unexpanded panel on page load.

Hmm... the colorful panels in the screenshot below are supposed to be hidden when the page finally loads, however during loading it does show up. That seems to be the same concern stated in #921.

I did not do any scrolling in the screenshot.

fouc1

openorclose:
> > View the preview site, see if there is a flash of the unexpanded panel on page load.
> 
> Hmm... the colorful panels in the screenshot below are supposed to be hidden when the page finally loads, however during loading it does show up. That seems to be the same concern stated in #921.
> 
> I did not do any scrolling in the screenshot.
> 
> fouc1

Ok, I've added another css rule to hide all inner panels from being displayed.

asset/css/markbind.css
```diff
@@ -198,6 +198,99 @@ so we apply temporary styles to fix the layout until Vue correctly updates the D
     display: none;
 }
 
+/* Hide the contents of unexpanded panels */
+/* stylelint-disable selector-type-no-unknown */
+panel:not([expanded]):not([minimized]) {
+    visibility: hidden;
+    display: block;
+    height: 53px;
```
yamgent: Please sort the attributes by alphabetical order.

Chng-Zhi-Xuan:
@openorclose 
Please clean up the commits before merging.

damithc:
> LGTM, just need to clean up the commits before merging.

Squash-merge?

Chng-Zhi-Xuan:
> Squash-merge?

That will work.

PR #929, opened 2019-07-13
Add google analytics support

### LINKED ISSUE #355: Support google analytics
Context: User may need to use Google analytics to track a website
Current: There is no way to add stuff to the `<head>` section of a page
Suggest: provide a way to either add stuff to the head section or direct support for Google analytics

### LINKED ISSUE #902: Support easy integration of Google Analytics
Provide an easy way to add Google Analytics to a site, e.g., simply by configuring site.json

Crphang:

- [X] New feature

Resolves #355
Resolves #902

Remaining Tasks:

- [x] User Guide
- [x] Add test to show that plugin indeed generates google analytics.

To provide easy way for authors to add google analytics into their web built by markbind. This PR adds this support by adding additional Google Analytics plugin.

image

 - Added a Google Analytics plugin

Add to site.json accordingly
```json
{
  "plugins": ["googleAnalytics"],
  "pluginsContext" : {
    "googleAnalytics" : {
      "trackingID": "UA-143800593-1"
    }
  }
}
```

This will be further simplified with #930


Crphang:
Ready for review.

Would require additional help on functional testing portion. Not entirely sure what is the adopted/best practice.

docs/userGuide/plugins/googleAnalytics.mbdf
```diff
@@ -0,0 +1,20 @@
...
+  "plugins": [
+    "googleAnalytics"
+  ],
+  "pluginsContext": {
+    "googleAnalytics": {
+      "trackingID": "YOUR-TRACKING-ID", // replace with your google analytics tracking id.
+    }
+  }
+}
+```
```
yamgent: Missing newline at the end of the file

docs/userGuide/plugins/googleAnalytics.mbdf
```diff
@@ -0,0 +1,20 @@
+#### `Google Analytics`: Enhancing site with Google Analytics
+
+This plugin allows your web pages to be enhanced with [google analytics](https://analytics.google.com/analytics/web/#/) to track, analyse and improve your content.
+
+To add Google Analytics plugin, simply add the following to `site.json`:
```
yamgent: Should there be a basic description on how to get the tracking ID? They can be told to visit this link: https://support.google.com/analytics/answer/1008080.

yamgent: In fact, let's follow the format of Algolia, with a table for the `trackingID` parameter.

docs/userGuide/usingPlugins.md
```diff
@@ -171,6 +171,7 @@ Disabling the `anchors` plugin:
 
 MarkBind has a set of built-in plugins that can be used immediately without installation.
 
+<include src="plugins/googleAnalytics.mbdf />
```
yamgent: Can we have these entries in alphabetical order? Thanks.

test/functional/test_site/testGoogleAnalytics.md
```diff
@@ -0,0 +1,18 @@
...
+<script async src="https://www.googletagmanager.com/gtag/js?id=TRACKING-ID"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+
+  function gtag() {
+    dataLayer.push(arguments);
+  }
+  gtag('js', new Date());
+
+  gtag('config', 'TRACKING-ID');
```
yamgent: I don't think it is necessary to put the content of the script here (and as a matter of fact, this page won't be needed at all). Although it is convenient for reviewers, it also means that future updates to the main script would require additional work in replacing it here as well. New changes would immediately surface on the diff of the expected files anyway, making them quite obvious, so this won't be necessary.

Crphang:
@yamgent made changes. Ready for review. Thanks for reviewing

damithc:
> Hmm interesting, the diagrams seems to change due to it being rendered differently (I presume it is because different OS renders the diagrams differently?)

What do we do about this? The diagrams shouldn't be part of this PR right?

yamgent:
> What do we do about this? The diagrams shouldn't be part of this PR right?

Yes, I am of the view that the changes should be discarded for now in this PR, since the original source code of the diagram did not change.

If the source code changes, then the PR that made the changes should update the diagram (for that case, I think it is OK if the PR author is not using Windows to update them, since pictures need manual verification anyway...)

damithc:
> Yes, I am of the view that the changes should be discarded for now in this PR, since the original source code of the diagram did not change.
> 
> If the source code changes, then the PR that made the changes should update the diagram (for that case, I think it is OK if the PR author is not using Windows to update them, since pictures need manual verification anyway...)

Does that also mean every PR coming from a non-Windows contributor will be required to remove unrelated changes from the PR every time? 😨

yamgent:
> Does that also mean every PR coming from a non-Windows contributor will be required to remove unrelated changes from the PR every time? 😨

A possible way to alleivate this issue for now is to get rid of PlantUML diagrams in the functional tests. 

Ideally, it would be great if someone could try and find ways to make the style consistent across all OS (I haven't research enough to say whether this is feasible with PlantUML though).

Crphang:
@yamgent Removed diagram changes and typo. Ready for review.

PR #934, opened 2019-07-17
Add docs for popover and tooltip hoverable attr

openorclose:

Adds docs for https://github.com/MarkBind/vue-strap/pull/114

Document the new attribute `hoverable` available for popovers and tooltips.

Added a new entry to the attributes table, and gave an example on how it works.


docs/userGuide/syntax/popovers.mbdf
```diff
@@ -135,6 +137,22 @@ More about <trigger for="pop:trigger_id">trigger</trigger>.
...
+
+<include src="outputBox.md" boilerplate >
+<span id="code">
+
+```html
+<popover content="You can select me!" hoverable>This popover's content can be interacted with.</popover>
+```
+</span>
+<span id="output">
+<popover content="You can select me!" hoverable>This popover's content can be interacted with.</popover>
```
damithc: I'm wondering, why can't this be the default/only behavior? Is there any downside to making all popovers hoverable?

yamgent: I left a reason [here](https://github.com/MarkBind/vue-strap/pull/115#issuecomment-512628980). It is a small problem, but still tolerable enough to make it the only behavior.

yamgent:
Doesn't seem to be relevant anymore after MarkBind/vue-strap#115 is merged.

Issue #861, opened 2019-05-20
Update 'init' command to create a site given in markbind/init-typical


damithc:
Suggestion: update the `markbind init` command to produce the site given in markbind/init-typical so that the user gets a more fleshed out starter site.

Example output https://markbind-init-typical.netlify.com

damithc:
Also see [this comment](https://github.com/MarkBind/markbind/issues/502#issuecomment-450076594)

Issue #519, opened 2019-01-01
Improve the init command


damithc:
Some suggestions for improving the `init` command (can be broken down to multiple PRs):

The default starter site can be enhanced to make it more useful to the user as a starting point. e.g., create a site that has all typical features (i.e., top nav, site nav, footer, search, a bunch of pages typically used such as home, contact, docs etc.) so that the user can go from there to a production website with least amount of effort.

Perhaps we can have other options?

`--minimal`: create an empty site 
`--convert`: converts an existing site to a MarkBind site. This can be an 'intelligent' conversion e.g., deduce siteNav items based on page titles
`--restore`: restores the folder to the previous state (i.e., reverses the `init` command)

Chng-Zhi-Xuan:
I would say `--minimal` and `--restore` is reasonable to implement, as well as revamping the default init `index.md`. 

However, `--covert` could prove to be challenging as existing html code could have components that MarkBind does not support (from other static site generators). A straightforward implementation could be using `jQuery` or `Cheerio` to extract the `<body>` of the previous site and place it within the `index.md` file. Any rendering or formatting issues would have to be manually fixed by the user. An advantage of this is that any of the `<frontmatter>` components can be used immediately without changes to existing code base.

damithc:
> However, `--convert` could prove to be challenging as existing html code could have components that MarkBind does not support (from other static site generators).

We can limit this feature to simple markdown-based document repositories, such as documentation in GitHub repos/wikis. The main aim is to give an easy way for GitHub repos to convert their documentation into a MarkBind site.

damithc:
@amad-person this is somewhat related to your current work.

damithc:
related #861

yash-chowdhary:
> `--restore`: restores the folder to the previous state (i.e., reverses the `init` command)

Would this simply cover deleting the files and directories that `init` created (`contents`, `site.json`, `index.md` and `_markbind`) or should it be the case that this command can _only_ be run immediately after `init` is executed (I'm thinking of the case where `serve` or `build` has been executed and then `--restore` is executed, or where it's run before `init`)?

damithc:
> `--restore`: restores the folder to the previous state (i.e., reverses the `init` command)

If we assume most users will be using git, we might not even need a restore option.

lhw-1:
There are four possible improvements brought up in this issue: `--minimal`, `--convert`, `--restore`, and the updates to the starter site.

As of March 2023:

[...]

PR #935, opened 2019-07-28
Refactor initialization code and update typical init

Crphang:
Things Left To Do:

- [X] Refactor based on comments on the architecture
- [X] Update the `default` theme to actually reflect what is required in #861 
- [X] Add defensive code to validate given theme
- [x] Refactor Tests. This portion is rather major. A design doc will be updated in this thread on how unittests are refactored

- [X] Enhancement to an existing feature

Fixes: #861 
Related to: #519

This PR updates current initialization to illustrate a more complete set of features. On top of that, it also addresses the issue that editing our scaffold markbind or providing more initialization options is too complex in our current codebase which uses const hardcoded defaults.

There were two main architecture that I explored in this PR on how to refactor this initialization code:

Copying the entire theme:
image

Creating all the files and directories, and copying from the theme if file exist for the theme:
image

The code below illustrates the change in Site.initSite based on the first architecture.

```js
/**
 * Static method for initializing a markbind site.
 * Generate the site.json and an index.md file.
 *
 * @param rootPath
 */
Site.initSite = function (rootPath, theme) {
  const themePath = path.join(__dirname, TEMPLATE_ROOT_FOLDER_NAME, theme);

    fs.accessAsync(rootPath)
      .catch(() => fs.copy(themePath, rootPath))
      .then(resolve)
      .catch(reject);
  });
};
```

Design | Pros | Cons
--- | --- | ---
*First Design which copies everything* | WYSIWYG nature can make it easy to work with for maintainers to understand the code of initialization because it is simple | The codebase will have many repetition as we need to provide empty directories for all themes/templates even if it is empty
*Second Design which creates directories first and only copies required or provided files* | Only need to provide the files that is required. <br> When adding a new directory, we don't have to go into each of those themes to add those empty directory if those themes don't require. | Similar to current codebase, we have to go through the entire list of file generation to understand what files are created.

This commit illustrates the first design, but I am more *for* the second design which is basically adding 1 additional line to each promise blocks of the previous code. Would appreciate some discussion here.

Example second design:

```js
function createAndCopyFileFromTheme(generateFileOrDirectory, objectPath, themePath) {
  generateFileOrDirectory
  .then(() => {
    if (!fs.existsSync(themePath)) {
      return Promise.resolve();
    }
    return fs.copyAsync(themePath, objectPath)
  })
  .catch(() => {
    return Promise.reject();
  })
}

Site.initSite = function (rootPath, theme) {
  const themePath = path.join(__dirname, TEMPLATE_ROOT_FOLDER_NAME, theme);
  const boilerplatePath = path.join(rootPath, BOILERPLATE_FOLDER_NAME);

    fs.accessAsync(boilerplatePath)
      .catch(() => {
        if (fs.existsSync(boilerplatePath)) {
          return Promise.resolve();
        }
        const boilerplateThemePath = path.join(themePath, BOILERPLATE_FOLDER_NAME);

        return createAndCopyFileFromTheme(fs.mkdirp(boilerplatePath), boilerplatePath, boilerplateThemePath);
      })
      .then(resolve)
      .catch(reject);
  });
}
```


Crphang:
Request for comments 

cc: @damithc

damithc:
> cc: @damithc

As this is an internal design improvement, I'm not in a position to say much. I'll wait for senior devs to weigh in.

yamgent:
I am leaning towards the second design, although I feel that both designs will start to show their flaws when we need to rename existing directories.

> **Second Design Pros:**
> Only need to provide the files that is required. 
> When adding a new directory, we don't have to go into each of those themes to add those empty directory if those themes don't require.

Actually, is there strictly a need to follow a directory structure for the template in the first place? I feel that this requirement will make it harder to maintain as we now have to ensure that both the code and the template directory structure are in sync with each other, which isn't desirable either.

Crphang:
@yamgent very good points raised there, definitely there are some restrictions on both designs. 

On those points, I think a mixture of first and second design can provide both flexibility and proper safeguards. I treat the `code` directories as a form of Interface where all templates in the future need, these directories should be minimal. If these `code` directories need to be renamed or changed, the impact will be minimized.

We can then follow the first design to initialize the site if those basic directories exist (passes validation) by copying the entire template files.

What do you think?

yamgent:
> On those points, I think a mixture of first and second design can provide both flexibility and proper safeguards. I treat the `code` directories as a form of Interface where all templates in the future need, these directories should be minimal. If these `code` directories need to be renamed or changed, the impact will be minimized.
> 
> We can then follow the first design to initialize the site if those basic directories exist (passes validation) by copying the entire template files.
> 
> What do you think?

Sure, let's evolve the design to that direction and see how it goes.

Crphang:
Test changes:

1. Included validation in unit test to show test for validation.
1. Leverage more on MemFS to perform template mocking.

Previous  design:
image

Current  design:

image

Main differences lies in Site initialization process. Previous site initialization does not require user input, so the unit test was resembling an integration test. Current site initialization requires user input and the output depends on the template that was chosen. Since this is a Unit Test, user input does not matter for us (should be covered by integration tests for future templates added), we mock the template input accordingly.

Crphang:
@yamgent wondering whether you have any insight on the failing test. It is passing locally for me, seems it could be because of the node_modules cache.

yamgent:
> @yamgent wondering whether you have any insight on the failing test. It is passing locally for me, seems it could be because of the node_modules cache.

Were you running npm 5.8.0 when you update package-lock.json? I recall the newer version of npm handling versions differently, which may cause certain dependencies to be the wrong versions and could result in different test outputs generated.

Crphang:
Ready for review. Thanks for your help @yamgent so far! I reverted to node 8 and npm 6 to get this working. Definitely some issue dependency issues there when using node 10 and upgrading our dependencies. It will cause deploy/netlify stage to fail I would attempt to reproduce in a fresh issue.

The architecture and design are stated in previous comments on how I approached this issue. Hopefully that will help in reviewing. I adopted open-close principal in allowing for developers to create new initialization templates, and also this design made it easier to integrate with remote templates in the future.

package-lock.json
```diff
@@ -9587,4 +9587,4 @@
       }
     }
   }
-}
+}
```
yamgent: This change should not happen.

src/Site.js
```diff
@@ -231,119 +153,17 @@ function setExtension(filename, ext) {
...
-  const headFolderPath = path.join(rootPath, HEAD_FOLDER_PATH);
-  const indexPath = path.join(rootPath, INDEX_MARKDOWN_FILE);
-  const siteNavPath = path.join(rootPath, SITE_NAV_PATH);
-  const siteLayoutPath = path.join(rootPath, LAYOUT_FOLDER_PATH);
-  const siteLayoutDefaultPath = path.join(siteLayoutPath, LAYOUT_DEFAULT_NAME);
-  const siteDefaultLayoutScriptsPath = path.join(siteLayoutDefaultPath, LAYOUT_SCRIPTS_PATH);
-  const sitePluginPath = path.join(rootPath, PROJECT_PLUGIN_FOLDER_NAME);
-  const userDefinedVariablesPath = path.join(rootPath, USER_VARIABLES_PATH);
-  // TODO: log the generate info
+Site.initSite = function (rootPath, templatePath = 'default') {
```
yamgent: The default value is not necessary since the caller in index.js already supplies the default value.

src/template/template.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  return true;
+}
+
+function generateSiteWithTemplate(rootPath, templatePath) {
+  return new Promise((resolve, reject) => {
+    fs.accessAsync(rootPath)
+      .catch(() => fs.emptydirSync(rootPath))
```
yamgent: I don't think it is appropriate to wipe the rootPath directory should generation fail, it may already have some existing content before MarkBind was invoked, and it is not a good idea to delete those files.

src/template/template.js
```diff
@@ -0,0 +1,55 @@
...
+      .then(() => fs.copyAsync(templatePath, rootPath))
+      .then(() => {
+        resolve();
+      })
```
yamgent: The three lines above can be simplified to:

```js
        .then(resolve)
```

Same for the other places in the code.

src/template/default/site.json
```diff
@@ -0,0 +1,30 @@
...
+      "glob": "**/index.md"
+    },
+    {
+      "glob": "**/*.+(md|mbd)"
+    }
+  ],
+  "deploy": {
+    "message": "Site Update."
+  }
+}
```
yamgent: Add newline

src/template/template.js
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error('Template validation failed. Required files does not exist');
+  }
+
+  return new Promise((resolve, reject) => {
+    generateSiteWithTemplate(this.rootPath, templatePath)
+      .catch((err) => {
+        reject(err);
+      })
+      .then(() => resolve())
+      .catch(() => reject());
```
yamgent: Can be simplified to:

```js
        .catch(reject);
```

src/template/template.js
```diff
@@ -0,0 +1,55 @@
...
+  return new Promise((resolve, reject) => {
+    fs.accessAsync(rootPath)
+      .catch(() => fs.emptydirSync(rootPath))
+      .then(() => fs.copyAsync(templatePath, rootPath))
+      .then(() => {
+        resolve();
+      })
+      .catch((err) => {
+        reject(err);
+      });
```
yamgent: Does `.catch(reject)` not work for this case? I don't think we need to explicitly specify `err` unless we want to use the value (e.g. if we want to log the error message with the logger).

test/functional/test_site/expected/testImportVariables.html
```diff
@@ -62,6 +62,19 @@ <h2 id="trying-to-access-an-imported-variable-via-namespace">Trying to access an
...
+<script>
+  window.dataLayer = window.dataLayer || [];
+
+  function gtag() {
+    dataLayer.push(arguments);
+  }
+  gtag('js', new Date());
+
+  gtag('config', 'TRACKING-ID');
+</script>
```
yamgent: These changes seem to be unrelated to the PR, I wonder how come our tests didn't fail. :O

Crphang: Shouldn't this be added as part of previous PR 🤔. Why didnt previous PR fail if it does not have this.

src/template/template.js
Crphang: Hmm i'll change

src/template/template.js
Crphang: I thought that was the behaviour, it creates the directory if accessAsync fails as it assumes that the directory does not exist.

yamgent: Perhaps `fs.mkdirSync()` would be more appropriate for implementing that behaviour instead?

test/functional/test_site/expected/testImportVariables.html
yamgent: I think it is because another separate PR was also merged together with your previous PR, and the separate PR added in some new files. Never mind, let's put the fix in this PR, but as a standalone commit.

Crphang:
Made changes accordingly @yamgent. Thanks for helping to review!

I was wondering where would be most appropriate to place `page.ejs` since `template` dir is being hijacked :laughing:.

Based on my understanding, our current logic doesn't allow multiple forms of base template which means the current structure of `page.ejs` is quite coupled with our design. In that direction of thought, I would think that placing it at top level together with Page.js and Site.js would be appropriate. To limit the scope of this PR, I feel that it would be better to refactor this in a separate PR. What do you think?

Crphang:
Great looking out there!

Added a fix and a unit test to check that behavior

src/util/fsUtil.js
```diff
@@ -30,4 +52,6 @@ module.exports = {
     path.dirname(filePathWithExt),
     path.basename(filePathWithExt, path.extname(filePathWithExt)),
   ),
+
+  copySyncWithOptions,
```
yamgent: Can just put the code for `copySyncWithOptions()` here directly, since other methods also do it the same way for this file, let's try to be consistent.

Crphang:
Changed according to comments. Ready for review @yamgent. Thanks a lot!

yamgent:
> You might want to rename the first commit to follow the same as the PR title.

Apologies for being unclear, when I stated this, I meant removing the "[861-update-init-command]" part, so the title should only be "Refactor initialization code and update typical init". We don't put branch names in commit titles here. :P

Otherwise this looks pretty good. :+1:

Crphang:
Updated the commit name. Also merged the third commit into the first. Thanks a lot :smile:

PR #938, opened 2019-09-12
Provide additional minimal initialization option for users

Crphang:

Part of #519 to provide minimal initialization template. This helps to serve users who just want to get started authoring contents quickly.

The other portion of #519 is completed by #935 that provides a more feature filled initialization.

To do:

- [X] Add minimal template
- [X] Update User Guide
- [X] Add integration test to ensure that minimal template is correct for bash script
- [x] Add integration test to ensure that minimal template is correct for batch script

To discuss:

Would appreciate comments on the content on the minimal template I provided. Trying to balance between minimal and useful is rather tricky, what I did was to point them to the user guide in `index.md`


yamgent:
I would still create the files inside `_markbind/`, such as `variables.md` and `footers/footer.md`, so that the proper structure of the website is still there (but their contents can be empty unlike the default template). 

With the proper structure in place, authors do not have to figure out the structure from scratch, and can start filling up the content immediately.

I think providing the best setup without the placeholder content should be the goal for "minimal", hence the suggestion above.

Crphang:
Ready for review.

Added integration tests for initialization templates. Also updated the minimal initialization according to review comments, which makes it easier for user to start using markbind with helper directories.

src/template/minimal/_site/index.html
```diff
@@ -0,0 +1,51 @@
+<!DOCTYPE html>
```
yamgent: Er these files are not supposed to be in the commit right? The generated sites shouldn't be in the template.

docs/userGuide/templates.md
```diff
@@ -0,0 +1,22 @@
...
+</frontmatter>
+
+# Templates
+
+Markbind allows you to start off your new project with different templates.
+
+During initialization you can add a flag `--template <template-key>` to select a different template to initialize with. For example:
+
+```
+markbind init --template default
```
yamgent: Maybe use minimal as an example instead? default template shouldn't need to be specified.

docs/userGuide/gettingStarted.md
```diff
@@ -59,6 +59,11 @@ You can add the `--help` flag to any command to show the help screen. <br>
 
 Navigate to the project {{ tooltip_root_directory }}.
 
+</tab>
+<tab header="Starting with different templates">
+
+By default, you will initialize markbind with a feature filled [template](https://markbind-init-typical.netlify.com/). By adding a flag `--template <template-name>` you can select a different template to initialize with. Learn about other [templates](templates.html).
```
yamgent: `markbind` -> `MarkBind`.

test/functional/test.sh
```diff
@@ -44,7 +44,7 @@ do
     cp -r $site_convert/non_markbind_site/_site $site_convert
 
     # run our test script to compare html files
-    node testUtil/test.js $site_convert
+    node testUtil/test.js compare $site_convert
```
yamgent: `test/functional/testUtil/test.js` doesn't seem to be using `compare` as the command keyword to run the tests, so how does this work? 🤔

Crphang: I extended the functionality `test.js` to take in actions and flags like file path.

Crphang:
Updated. Thanks so much for reviewing!

test/functional/test.sh
yamgent: I understand your purpose of extending the `test.js`. What I am saying is that your extended `test.js` is using the keyword `test`, not `compare`:

https://github.com/MarkBind/markbind/blob/7f07b5ce9db10d3912783ccd4300b1711ad09d9f/test/functional/testUtil/test.js#L14-L15

So nothing in `test.js` is actually getting executed. Calling a non-existent command also returns 0, so we don't know whether the tests work or not. There is also another issue with `test.js`, which I highlighted in another comment...

test/functional/testUtil/test.js
```diff
@@ -10,41 +11,46 @@ function readFileSync(...paths) {
...
+          throw new Error('Site data does not match with the expected file.');
+        }
+      }
     }
-  }
-}
 
-if (error) throw new Error('Diffs found in .html files');
+    if (error) throw new Error('Diffs found in .html files');
+  });
```
yamgent: There should be a `program.parse(process.argv);`, otherwise the arguments are not getting parsed and the `test` command won't work.

test/functional/test.sh
```diff
@@ -59,6 +59,35 @@ do
...
+    node ../../index.js init $site/tmp --template $flag
+    node ../../index.js build $site/tmp
+
+    # run our test script to compare html files
+    node testUtil/test.js compare $site --expectedPath expected --sitePath tmp/_site
+    
+    # remove generated files
+    rm -rf $site/tmp
+
+    if [ $? -ne 0 ]
```
yamgent: This is checking the return code of `rm`, but we should be checking the return code of `node testUtil/test.js`.

test/functional/test.sh
```diff
@@ -59,6 +59,35 @@ do
     trap - EXIT
 done
 
+for siteInfo in $(cat test_template_sites);
+do
+    trap cleanup_initialized EXIT
+
+    flag=$(cut -d',' -f1 <<<"$siteInfo")
+    site=$(cut -d',' -f2 <<<"$siteInfo")
```
yamgent: `test_sites` and `test_convert_sites` don't use `,` as delimiters, instead each entry is a single line. Let's also do that for `test_template_sites`.

docs/userGuide/templates.md
```diff
@@ -0,0 +1,22 @@
+<variable name="title" id="title">Templates</variable>
+<frontmatter>
+  title: "User Guide: {{ title }}"
+  layout: userGuide
+</frontmatter>
+
+# Templates
+
+Markbind allows you to start off your new project with different templates.
```
yamgent: Should be capital B, Mark**B**ind.

docs/userGuide/templates.md
```diff
@@ -0,0 +1,22 @@
...
+
+```
+markbind init --template minimal
+```
+
+## Supported Templates
+
+Name    | Template key | Description
+----    | -------      | ------
+Default | `default`    | Default template filled with Markbind features to guide you to author better content
```
yamgent: Can mention in the description that "(this is the template used by default if `--template` is not specified)"

docs/userGuide/templates.md
```diff
@@ -0,0 +1,22 @@
...
+Minimal | `minimal`    | Minamalistic template that gets you started quickly
```
yamgent: `Minamalistic` -> `Minimalistic `

test/functional/test.sh
Crphang: I am not very sure about this. I see it as 1 test case is 1 single line. 

Suppose I remove `,`  I either derive the flag from the path. For example:

1. `test_site_templates/test_minimal` -> minimal
1. Or derive the path `minimal` -> `test_site_templates/test_minimal`.

Between these two, I would prefer the second one, but when writing this, I was thinking that being explicit between the flag and the test path would have been better. What do you think?

yamgent:
I will try and get back to you by Sunday night!

test/functional/test.sh
yamgent: Ah icic, had some misunderstanding about this part of the code, I am fine with this approach. :+1:

yamgent:
Ok, content wise I am satisfied with the changes. We can now reorganize the commits in the PR, suggested structure:

* Add minimal template
* Add functional tests for template initialization
* Add documentation for initializing different templates

Crphang:
@yamgent Thanks a lot for the reviews. Made the changes accordingly.

yamgent:
For first commit: Please put the second sentence as the commit message body. Generally commit messages should only have one proper English sentence without periods (see [this page](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message) for formatting commit messages).

Something like this:

```
Add minimal template

The template provides a minimalistic view of markbind for authors.
```

Crphang:
Changed accordingly @yamgent!

PR #944, opened 2019-11-24
Support for expressive layouts with page.md

### LINKED ISSUE #631: Layouts: give a way to insert default content at top/bottom of the page body
Current: Users can use the layouts feature to configure various defaults for a page e.g., footer, header

Suggested: Give a way to specify some default content to be at the top and bottom of the page body.

Justification: In some cases, the user may want to wrap body content in a div e.g., `<div id="main">` or insert a TopNav in every page.

Some options:
1. Support a `top.md` and `bottom.md` within a layout
2. Support a `page.md` within a layout e.g.,

```html
<include src="topNav.md" />
<div id="main">
{{ body }}
<hr>
</div>
```
The advantage of option 2 is that we can even support default `<frontmatter>`

Crphang:

• [X] New feature

Resolves #631 

Currently this is a WIP but the Proof of Concept generally working. The idea is for user to define a main page structure to allow users to define wrapper content around their main content.

For instance:

Code:
image

Result:
image

Definitely a lot more work to be done:

- [x] Because page need to be pre-processed (prior to parser), we have yet to process frontmatter which is required to pass in layout information. Figure out a way to process/wrap page after parser.
- [x] I currently hacked into Parser to make this work. Refactor some part of parser to take in data instead of file.
- [x] Layouts are currently not refreshed during hot reloading.


yamgent:
No comments so far for the current features that the prototype has, looking forward to the final implementation. :+1:

Btw, if there are a lot of refactoring to be done, and those refactoring might also benefit other features (not necessary just for expressive layout), you might want to consider handling the refactoring in separate PRs, so that this PR doesn't get too big.

Crphang:
Keeping things super simple for now to ensure that this PR does not go out of hand:

- [X] Fix problem with parsing and undoing refactor
- [] User Guide update
- [] Update test

Important functional and non-functional requirements that this PR takes care of:

1. Allowing user to define any layout style and embed main body content - Core Feature
1. Backward compatibility of existing users if they upgrade.

Many more work to be done but will be separated to different PRs. A non-exhaustive list include:

1. Refactoring layouts logic from insertion of header, footer, ...
1. Removing inference of layouts from frontmatter. Currently the code takes frontmatter as source of truth instead of the Page property. This needs to be updated but it is not that simple.
1. Expressive Layouts Validation. If users change the keyword of expressive layout from MAIN_CONTENT_BODY to something else, pages will not work. We should have validation (for templates created by developers) and soft failure (for user input changes) to skip expressive layouts if user make wrong changes.

Crphang:
Ready for review

test/functional/test_site_convert/expected/contents/topic1.html
```diff
@@ -61,10 +61,11 @@
         </div>
       </nav>
       <div id="content-wrapper">
-        <br>
+        <br cwf="/home/crphang/markbind/test/functional/test_site_convert/non_markbind_site/contents/topic1.md">
```
yamgent: A quick review in passing... seems like this should not be appearing as part of the output?

Crphang: Yes. I realized that so reverted "ready for review" :sweat:. The ordering of how when the parser is invoke matter. I changed it a little and this happened, even though functionally it works.

I'll see how far I could work around the parser on this diff. If things would start to look really messy, I would see if some refactor of the parser have to be done to make this work better.

test/functional/test_site_templates/test_default/expected/index.html
```diff
@@ -89,9 +89,9 @@ <h1 id="heading-1">Heading 1<a class="fa fa-anchor" href="#heading-1"></a></h1>
         </ul>
         <p><strong>A <code>code</code> example:</strong></p>
         <div class="code-block">
-          <pre><code class="hljs html"><span class="hljs-tag">&lt;<span class="hljs-name">foo</span>&gt;</span>
-  <span class="hljs-tag">&lt;<span class="hljs-name">bar</span> <span class="hljs-attr">type</span>=<span class="hljs-string">"name"</span>&gt;</span>goo<span class="hljs-tag">&lt;/<span class="hljs-name">bar</span>&gt;</span>
-<span class="hljs-tag">&lt;/<span class="hljs-name">foo</span>&gt;</span>
+          <pre><code class="hljs html">&amp;lt;foo&amp;gt;
```
Crphang: Hmm. Everything else seems okay, but what happened here :thinking:

yamgent:
Just to let you know that I will take a look at this soon!

src/Page.js
```diff
@@ -788,11 +821,13 @@ Page.prototype.generate = function (builtFiles) {
     userDefinedVariablesMap: this.userDefinedVariablesMap,
   };
   return new Promise((resolve, reject) => {
-    markbinder.includeFile(this.sourcePath, fileConfig)
+    fs.readFileAsync(this.sourcePath)
       .then((result) => {
         this.collectFrontMatter(result);
         return this.removeFrontMatter(result);
```
yamgent: ⭐⭐

The rearranging of the processing order may break some use cases. For example, in `docs/userGuide/markBindSyntaxOverview.md`, the frontmatter actually contains some nunjucks syntax:

```html
<frontmatter>
  title: "User Guide: {{ title }} "
  layout: userGuide
</frontmatter>
```

This caused a problem with the page title being rendered incorrectly.

`includeFile()` also calls `preprocess()`, but in the new case it seems to be not called before this line? Because this does contain precisely the nunjucks rendering that allows the above to work.

Is it possible to obtain the layout name without having to re-order the sequence?

`docs/userGuide/index.md` is pretty broken too. Not sure what is going on there, the frontmatter seems to be totally left behind.

Crphang: Thanks so much for the review :smile:. I added a fix on this. Will address the rest once we confirmed the approach for this.

The new sequence is this:

(1) Markbind preprocessing
(2) Frontmatter
(3) Expressive Layouts

Within expressive layouts, I instantiated a markbinder and parsed the additional wrapper content within layouts.

yamgent: > The new sequence is this:
> 
> (1) Markbind preprocessing
> (2) Frontmatter
> (3) Expressive Layouts
> 
> Within expressive layouts, I instantiated a markbinder and parsed the additional wrapper content within layouts.

Seems fine for this sequence at the moment, can't think of major issues for this.

src/Page.js
```diff
@@ -505,6 +507,42 @@ Page.prototype.removeFrontMatter = function (includedPage) {
   return $.html();
 };
 
+/**
+ * Produces expressive layouts by inserting page data into pre-specified layout
+ * @param pageData a page with its front matter collected
+ */
+Page.prototype.generateExpressiveLayout = function (pageData, fileConfig) {
```
yamgent: It seems like we can't use variables inside `page.md`, so for example if we do:

```
{{ example }}

{{ MAIN_CONTENT_BODY }}

Bottom Content
```

The `{{ example }}` variable won't get rendered.

Crphang: Yes. Was wondering about that. It is okay to support page and global user defined variables. But things can start to get weird when we think about imported/inner variables for this situation.

For example, since the layout and corresponding imports is embedded into all pages, does imported variables come from the layout page or is it part of the actual page? Implementation wise, there should be some additional steps required to prune the import elements from the `page.md` before it is merged with the other content to prevent import error later on.

yamgent: So do you mean that you are worried that in this case, `index.html` will show "This is from b" rather than "This is from c"?

`page.md`
```
<import name from "b.md"/>
{{ name }}

{{ MAIN_CONTENT_BODY }}
```

`index.md`
```
<import name from "c.md"/>
```

`b.md`
```
<variable name="name">This is from b.</variable>
```

`c.md`
```
<variable name="name">This is from c.</variable>
```

Crphang: Hmm, not exactly. The above is resolved. 

It is more of a minor user experience nit about making it clear whether the variable imported from the `page.md` is relative to the `page.md` file while the others are from the actual `index.md`.

Let me enhance the tests to reflect that.

docs/userGuide/syntax/pageLayouts.mbdf
```diff
@@ -95,3 +101,41 @@ afterSetup(() => {
...
+
+#### Title 1
+
+Content of your page
+
+#### Title 2
+
+Additional Content
+
+<form action="https://www.paypal.com/cgi-bin/webscr" method="post" target="_top">
```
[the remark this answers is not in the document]
Crphang: Updated it to be showing statistics formula instead :smile:. This should make it more realistic and applicable to our current users.

Crphang:
Ready for review

docs/userGuide/syntax/pageLayouts.mbdf
```diff
@@ -95,3 +101,37 @@ afterSetup(() => {
...
+<box>
+  {{ mainContentBody }}
+  ```html
+    <panel header="Statistics Formula for the class" type="primary">
+      <img src="path_to_your_formula.png" />
+    </panel>
+  ```
+</box>
+
+Here's how the above content would appear:
```
yamgent: You might also want to mention the content of `index.md` separately, so that it is easier for the reader to understand that the `#### Main content of your page` is actually coming from `index.md`.

test/functional/test_site_expressive_layout/_markbind/layouts/default/page.md
```diff
@@ -0,0 +1,14 @@
+<variable name="layoutVar">Variable from layout</variable>
+<import importedVar from="imported.md"/>
+
+{{ layoutVar }}
+{{ importedVar }}
```
yamgent: This variable doesn't seem to work? Unless I missed it, I don't see it in the output. :X

docs/userGuide/syntax/pageLayouts.mbdf
Crphang: Added the change

test/functional/test_site_expressive_layout/_markbind/layouts/default/page.md
Crphang: Thank you so much for the review :smile:! 

I fixed this by adding a namespace. Seems like without assigning a namespace importedVar wouldn't work, this is the case even for current release 2.9.0.

Sidenote: I don't think this is the intended behaviour but could investigate in a separate issue and PR.

docs/userGuide/syntax/pageLayouts.mbdf
```diff
@@ -95,3 +101,50 @@ afterSetup(() => {
...
+
+{{ icon_example }}Adding statistics formula to the bottom of every page
+
+<box>
+<i>index.md</i>
+<br><br>
+
+#### Main content of your page
+
+Content of your page
```
yamgent: The rendering of the example looks very odd for both `index.md` and `page.md` because of some inconsistencies in the style:

image

Possible to show the source code of `index.md` and `page.md` instead? `index.html` can still show the rendered result.

docs/userGuide/syntax/pageLayouts.mbdf
```diff
@@ -95,3 +101,50 @@ afterSetup(() => {
...
+
+In the `page.md` file of your layouts, it should come with the following reserved variable:
+
+<box>
+  {{ mainContentBody }}
+</box>
+
+which injects the actual page content in every page. This allows you to build layouts in different ways.
+
+{{ icon_example }}Adding statistics formula to the bottom of every page
```
yamgent: Minor nit: Space between the `}}` and `Adding`.

docs/userGuide/syntax/pageLayouts.mbdf
```diff
@@ -95,3 +101,52 @@ afterSetup(() => {
...
+
+#### Using Expressive Layout Templates
+
+In the `page.md` file of your layouts, it should come with the following reserved variable:
+
+<box>
+  {{ mainContentBody }}
+</box>
+
+which injects the actual page content in every page. This allows you to build layouts in different ways.
```
damithc: Reads better this way?

> The `page.md` file of a layout should come with the reserved variable `{{ MAIN_CONTENT_BODY }}` which injects the actual page content in every page. This allows you to build layouts in different ways.

Alternatively, can be written this way (i.e., avoid breaking sentences across boxes):
> The `page.md` file of a layout should come with the following reserved variable which injects the actual page content in every page. 
>```
>{{ MAIN_CONTENT_BODY }}
>```
>This allows you to build layouts in different ways.

PR #946, opened 2019-12-02
Fix expected help message to show header

### LINKED ISSUE #644: -h/--help option does not output the banner

I ran `markbind -h/--help` /  `markbind <option> -h/--help`.

Like other commands, `-h/--help` should output with the header:
```
  __  __                  _      ____    _               _
 |  \/  |   __ _   _ __  | | __ | __ )  (_)  _ __     __| |
 | |\/| |  / _` | | '__| | |/ / |  _ \  | | | '_ \   / _` |
 | |  | | | (_| | | |    |   <  | |_) | | | | | | | | (_| |
 |_|  |_|  \__,_| |_|    |_|\_\ |____/  |_| |_| |_|  \__,_|

 v1.17.1
 
  Usage: markbind <command>


  Options:

    -V, --version  output the version number
    -h, --help     output usage information


  Commands:

    init|i [root]                      init a markbind website project
    serve|s [options] [root]           build then serve a website from a directory
    deploy|d [options]                 deploy the site to the repo's Github pages.
    build|b [options] [root] [output]  build a website
```
The help information was printed without a header:
```
PS C:\Users\pzy5a\Desktop\testbind> markbind -h

  Usage: index  <command>


  Options:

    -V, --version  output the version number
    -h, --help     output usage information


  Commands:

    init|i [root]                      init a markbind website project
    serve|s [options] [root]           build then serve a website from a directory
    deploy|d [options]                 deploy the site to the repo's Github pages.
    build|b [options] [root] [output]  build a website
```
This was also the case for option help eg. `markbind serve -h/--help`.
Also, the usage information is improperly formatted; it should be `markbind <command>` instead of `index  <command>` (note the extraneous space).

madanalogy:

• [X] Bug Fix

Resolves #644 

Improves consistency of markbind's CLI interface.

I added a function to handle help messages for use with commander. The function prints the banner and is used in commander's ```helpOption()``` method.

```
function handleHelp(txt) {
  printHeader();
  logger.log('');
  return txt;
}
program
  .helpOption('-h, --help', handleHelp('output usage information'));
```

Updating of the commander module to latest version, whether it was done properly.

**Testing instructions:**

From the main working directory, run ```node index.js -h```

Fixes CLI help command.

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,11 +37,21 @@ function handleError(error) {
   process.exitCode = 1;
 }
 
+// We want to customize the help message to print MarkBind's header,
+// but commander.js does not provide an API directly for doing so.
+// Hence we override commander's outputHelp() completely.
+program.defaultOutputHelp = program.outputHelp;
+program.outputHelp = function (cb) {
+  printHeader();
+  this.defaultOutputHelp(cb);
+};
+
 program
   .allowUnknownOption()
   .usage(' <command>');
 
 program
+  .name('markbind')
   .version(CLI_VERSION);
 
 program
@@ -236,7 +246,6 @@ if (!program.args.length
   if (program.args.length) {
     logger.warn(`Command '${program.args[0]}' doesn't exist, run "markbind --help" to list commands.`);
   } else {
-    printHeader();
     program.help();
   }
 }


madanalogy:
Thank you for the information and suggestion :)

I have reverted the changes to the json files and implemented an override for outputHelp (instead of my earlier initial implementation). I will also be editing the PR title, thanks again for pointing me in the right direction!

index.js
```diff
@@ -37,11 +37,27 @@ function handleError(error) {
   process.exitCode = 1;
 }
 
+// The following code block overrides commander's outputHelp()
```
yamgent: It would be better if the comment mentions the _why_ (the reason for overriding) as well.

You should mention that we want to customize the help message with MarkBind's header, but commander.js doesn't provide any API to customize the help message.

```js
// We want to customize the help message to print MarkBind's header,
// but commander.js does not provide an API directly for doing so.
// Hence we override commander's outputHelp() completely.
```

index.js
```diff
@@ -37,11 +37,27 @@ function handleError(error) {
...
+  if (!cb) {
+    // eslint-disable-next-line no-param-reassign
+    cb = function (passing) {
+      return passing;
+    };
+  }
+  process.stdout.write(cb(this.helpInformation()));
+  this.emit('--help');
+};
+module.exports = program;
```
yamgent: Is this line needed? I don't think anybody outside `index.js` uses the `program` variable inside here.

index.js
```diff
@@ -37,11 +37,27 @@ function handleError(error) {
...
+delete program.outputHelp;
+program.outputHelp = function (cb) {
+  printHeader();
+  if (!cb) {
+    // eslint-disable-next-line no-param-reassign
+    cb = function (passing) {
+      return passing;
+    };
+  }
+  process.stdout.write(cb(this.helpInformation()));
```
yamgent: I was wondering whether it would be even better if we just shadow the original method, and utilise it in our new method. This way, we do not have to duplicate the original method's logic at all, in case there's an update to the commander's logic when we move to a new version:

```js
program.defaultOutputHelp = program.outputHelp;
program.outputHelp = function (cb) {
  printHeader();
  this.defaultOutputHelp(cb);
};
```

madanalogy:
Thank you for the suggestions! I have changed the implementation to shadow instead of replace and updated the code comment. I have also rebased and squashed my commits.

PR #955, opened 2019-12-15
Allow live reloading for plugin defined sources

### LINKED ISSUE #913: Support live-preview for .puml files
Suggestion: capture changes to .puml files in the live preview

ang-zeyu:

• [x] Documentation update

Resolves #913 

To allow plugins to define their own source file types and corresponding
source files that will be watched and updated by the live preview appropriately.

- Plugins can a new attribute in its exports: `getSources`
  - `getSources`: Allows plugins to return an array of source file paths to watch
- Similar to `Page.prototype.preRender`, `Page.prototype.getSources` runs `getSources` all plugins just before
`preRender`.
- Updated relavant test files

```js
(Plant uml plugin file)
module.exports = {
  ...,
  getSources: (content) => {
    // Add all src attributes in <puml> tags to watch list
    const $ = cheerio.load(content, { xmlMode: true });

    return $('puml').map((i, tag) => tag.attribs.src).get();
  },
```

Whether a callback style solution would be better ( a 5th parameter to `preRender / postRender` that contains callbacks plugins can use ), since in the long run one may find plugins requiring more and more functionality.

**Testing instructions:**
I added a `<puml>` tag, and an `<include>` tag ( including testPlantUML.md )
to the test site. Changing any of the source files ( e.g. `activity.puml` ) should
cause the affected pages to reload when serving the page.

Allow live reloading for plugin defined sources

This allows plugins to define a custom list of source files and file
types for the page to be watched, allowing live reload to rebuild
the affected page.

### DIFF
--- a/docs/userGuide/usingPlugins.md
+++ b/docs/userGuide/usingPlugins.md
@@ -143,6 +143,34 @@ This will add the following link and script elements to the page:
 - `<script src="SCRIPT_LINK"></script>`
 - `<script>alert("hello")</script>`
 
+#### Live reload
+
+By default, MarkBind treats `.html`, `.md`, `.mbd`, and `.mbdf` as source files, and will rebuild any
+pages changed when serving the page.
+
+During the `preRender` and `postRender` stages however, plugins may do custom processing using some other
+source file types, as parsed from the raw Markdown, typically requiring rebuilding the site.
+
+Hence, to add custom source files to watch, you can implement the `getSources()` method.
+- `getSources(content, pluginContext, frontMatter)`: Returns an array of source file paths to watch. Called **before** a Markdown file's `preRender` function is called.
+  - `content`: The raw Markdown of the current Markdown file (`.md`, `.mbd`, etc.).
+  - `pluginContext`: User provided parameters for the plugin. This can be specified in the `site.json`.
+  - `frontMatter`: The frontMatter of the page being processed, in case any frontMatter data is required.
+
+Example usage of `getSources` from the PlantUML plugin:
+
+```js
+{
+  ...
+  getSources: (content) => {
+    // Add all src attributes in <puml> tags to watch list
+    const $ = cheerio.load(content, { xmlMode: true });
+
+    return $('puml').map((i, tag) => tag.attribs.src).get();
+  },
+}
+```
+
 ### Advanced: Default plugins
 
 MarkBind has a set of default plugins that it uses to carry out some of its features. These are enabled by default for every project and should be left alone.
--- a/src/Page.js
+++ b/src/Page.js
@@ -81,6 +81,7 @@ function Page(pageConfig) {
   this.headings = {};
   this.headingIndexingLevel = pageConfig.headingIndexingLevel;
   this.includedFiles = new Set();
+  this.pluginSourceFiles = new Set();
   this.keywords = {};
   this.navigableHeadings = {};
   this.pageSectionsHtml = {};
@@ -794,6 +795,7 @@ Page.prototype.generate = function (builtFiles) {
       })
       .then(result => removePageHeaderAndFooter(result))
       .then(result => addContentWrapper(result))
+      .then(result => this.collectPluginSources(result))
       .then(result => this.preRender(result))
       .then(result => this.insertSiteNav((result)))
       .then(result => this.insertHeaderFile(result))
@@ -859,6 +861,21 @@ Page.prototype.getPluginConfig = function () {
     includedFiles: this.includedFiles,
     resultPath: this.resultPath,
   };
+};
+
+/**
+ * Collects file sources provided by plugins for the page
+ */
+Page.prototype.collectPluginSources = function (content) {
+  Object.entries(this.plugins).forEach(([pluginName, plugin]) => {
+    if (plugin.getSources) {
+      const sources = plugin.getSources(content, this.pluginsContext[pluginName] || {},
+                                        this.frontMatter, this.getPluginConfig());
+      sources.forEach(src => this.pluginSourceFiles.add(path.resolve(ensurePosix(src))));
+    }
+  });
+
+  return content;
 };
 
 /**
--- a/index.js
+++ b/index.js
@@ -119,7 +119,7 @@ program
     const addHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (fsUtil.isSourceFile(filePath)) {
+        if (fsUtil.isSourceFile(filePath) || site.isPluginSourceFile(filePath)) {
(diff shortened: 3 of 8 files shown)


src/Page.js
```diff
@@ -861,6 +863,25 @@ Page.prototype.getPluginConfig = function () {
...
 
+/**
+ * Collect page sources provided by plugins
+ */
+Page.prototype.collectPluginSources = function (content) {
+  Object.entries(this.plugins).forEach(([pluginName, plugin]) => {
+    if (plugin.getSources) {
+      const sources = _.castArray(
+        plugin.getSources(content, this.pluginsContext[pluginName] || {},
+                          this.frontMatter, this.getPluginConfig()));
```
acjh: Why `_.castArray`? `getSources` should return an array.

ang-zeyu: So that a single source file could be returned as well,
should I enforce returning an array instead?

acjh: Yes, since `getSources` is plural.

ang-zeyu:
Updated `getSources` to require returning array instead as per method naming

ang-zeyu:
Updated `getSources` to require returning array instead as per method naming

docs/userGuide/usingPlugins.md
```diff
@@ -143,6 +143,34 @@ This will add the following link and script elements to the page:
...
+
+```js
+{
+  ...
+  getSources: (content) => {
+    // Add all src attributes in <puml> tags to watch list
+      const $ = cheerio.load(content, { xmlMode: true });
+    
+      return $('puml').map((i, tag) => tag.attribs.src).get();
+    },
```
yamgent: The indentation of the method body is off. Only the comment line was in the correct indentation

docs/userGuide/usingPlugins.md
```diff
@@ -143,6 +143,34 @@ This will add the following link and script elements to the page:
...
 
+#### Live reload
+
+By default, MarkBind treats '.html', '.md', '.mbd', and '.mbdf' as source files, and will rebuild any
+pages changed when serving the page.
+
+During the `preRender` and `postRender` stages however, plugins may do custom processing using some other
+source file types, as parsed from the raw Markdown, typically requiring rebuilding the site.
+
+Hence, to add custom source files to watch, MarkBind allows defining an additional method in the plugin.
```
yamgent: `Hence, to add custom source files to watch, `~~`MarkBind allows defining an additional method in the plugin.`~~` implement the ``getSources`` method:`

Prefer being more direct in documentations so that it is easier for the reader to process.

docs/userGuide/usingPlugins.md
```diff
@@ -143,6 +143,34 @@ This will add the following link and script elements to the page:
...
+- `getSources(content, pluginContext, frontMatter, config)`: Called before the `preRender` function to retrieve an array of source file paths to watch.
+  - `content`: The raw Markdown of any Markdown file (`.md`, `.mbd`, etc.).
```
yamgent: `any` -> `the current`? The use of "any" makes it seems like this call is not tied to any pages.

docs/userGuide/usingPlugins.md
```diff
@@ -143,6 +143,34 @@ This will add the following link and script elements to the page:
...
+#### Live reload
+
+By default, MarkBind treats '.html', '.md', '.mbd', and '.mbdf' as source files, and will rebuild any
+pages changed when serving the page.
+
+During the `preRender` and `postRender` stages however, plugins may do custom processing using some other
+source file types, as parsed from the raw Markdown, typically requiring rebuilding the site.
+
+Hence, to add custom source files to watch, MarkBind allows defining an additional method in the plugin.
+- `getSources(content, pluginContext, frontMatter, config)`: Called before the `preRender` function to retrieve an array of source file paths to watch.
```
yamgent: **`Returns an array of source file paths to watch. `**`Called before a `**`Markdown file's`**`preRender function`**` is called.`**

src/Page.js
```diff
@@ -861,6 +863,21 @@ Page.prototype.getPluginConfig = function () {
   };
 };
 
+/**
+ * Collect page sources provided by plugins
```
yamgent: Seems a bit wrong to say "page sources", since puml files are not really pages? Maybe "file sources"?

ang-zeyu:
Hi, thanks for looking through it! I've updated the relavant lines.

On another note, I noticed that `usingPlugins.md` does not have documentation for the fourth parameter, `config`, even though it is used in the `preRender` and `postRender` functions. I've correspondingly also omitted it from `getSources` for now.

Was this an intentional decision to hide some of the implementation details? If not, I could update it as well.

yamgent:
> Was this an intentional decision to hide some of the implementation details? If not, I could update it as well.

It was probably an oversight. Feel free to add it in.

ang-zeyu:
Updated, will squash if its fine

ang-zeyu:
Updated, will squash if its fine!

acjh:
>> I noticed that `usingPlugins.md` does not have documentation for the fourth parameter, `config`, even though it is used in the `preRender` and `postRender` functions. I've correspondingly also omitted it from `getSources` for now.
>>
>> Was this an intentional decision to hide some of the implementation details? If not, I could update it as well.
> 
> It was probably an oversight. Feel free to add it in.

Actually, it was intentionally undocumented by @jamos-tay as of https://github.com/MarkBind/markbind/pull/714#discussion_r258418526:

> I needed access to `Page.headingIndexingLevel`, so I added a 4th parameter, page which is the `Page` object itself... Other plugins might need access to page config so this could be helpful
>
> I'm thinking we can use this for internal plugins, but don't document it in user docs, since we don't want to expose authors to the implementation

ang-zeyu:
> Actually, it was intentionally undocumented by @jamos-tay as of [#714 (comment)]

Thanks for the heads up!

Since @jamos-tay ended up going with the approach @marvinchin suggested, there isn't too much to document though. Regarding the original concern for the change ( passing the entire `Page` being dangerous ), I could include a warning in the docs about `includedFiles` since it is used in `Site.proto.updateSiteData()`. Still, its a subsection to add into the docs which means more details for the user to read. Other properties aside though, `rootPath`, `sourcePath` and `resultPath` seem to be quite useful for plugins. 

What's the take on this? 😮

acjh:
> Since @jamos-tay ended up going with the approach @marvinchin suggested, there isn't too much to document though. ... Other properties aside though, `rootPath`, `sourcePath` and `resultPath` seem to be quite useful for plugins.

The concern isn't that it is tedious to document. Rather, ideally, plugins should purely process content rather than use variables from and depend on the (filesystem-aware) implementation of Markbind CLI.

ang-zeyu:
Noted on the intended usage of plugins. Thanks!

Reverted the 'documentation updates' and squashed

PR #957, opened 2019-12-16
Add language indicator to code blocks

### LINKED ISSUE #646: Indicate the language of a code block

**Rationale:** Developer documentation sometimes includes code to do one thing in multiple languages (eg. API use in multiple languages). It can be helpful to indicate the language of a code block in the block itself.

**Example:** From [Travis CI documentation](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci):

code_language

nbriannl:

Fixes #646 

Developer documentation sometimes includes code to do one thing in multiple languages (eg. API use in multiple languages). It can be helpful to indicate the language of a code block in the block itself.

- In `lib\markdown-it\index.js`, I rewrote the rule for a markdown fenced code block, to add a div that will show the name of the language of that code block (if a language is specified). 
- Because I had to position the div relative to the code block, I had to wrap the code block as a parent div so that I can position the child div according to the parent. `markbind.css` 

Here is how it looks like

image

~~This is a WIP, I believe I am failing some tests ('Test test_site Failed').~~ And will be mentioned later, I'm open to a better solution. ~~Plus, I will remove a particular `console.log` asap.~~ ~~I have yet to update UserGuide.~~

I am not sure if I need to update User Guide, as a deployment of the Markbind docs will now show the output code blocks with the code language labels. So I'll remove [WIP], I'll put it back to WIP if there a outstanding changes to be made.

https://gist.github.com/nbriannl/755a2d05145a27e3dc7e5797c25f4a5f 
Include the markdown code from this gist and `markbind serve` to see the changes.

Or even simply view the netifly preview and view `/userguide/fullsyntaxreference`

First time contributing, I am open if there is a more elegant solution to obtain the language of the code block. I can also consider looking into the direction #629 did, since it is regarding adding a copy button instead of a language to a code block. 

Add language indicator to code blocks

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -24,6 +24,26 @@ pre > code.hljs {
     border-radius: 5px;
 }
 
+.code-block {
+    position: relative;
+}
+
+.code-block > .code-block-lang {
+    background-color: #b4b4b9;
+    border-radius: 0.25rem;
+    color: #f8f8ff;
+    display: inline-block;
+    font-size: 75%;
+    line-height: 1;
+    padding: 0.25em 0.4em;
+    position: absolute;
+    right: 0.3em;
+    text-align: center;
+    top: 0.4em;
+    vertical-align: baseline;
+    white-space: nowrap;
+}
+
 kbd {
     background-color: #fafbfc;
     border: 1px solid #c6cbd1;
--- a/src/lib/markbind/src/lib/markdown-it/index.js
+++ b/src/lib/markbind/src/lib/markdown-it/index.js
@@ -20,11 +20,11 @@ const slugify = require('@sindresorhus/slugify');
 // markdown-it plugins
 markdownIt.use(require('markdown-it-mark'))
   .use(require('markdown-it-ins'))
-  .use(require('markdown-it-anchor'), {slugify: (str) => slugify(str, { decamelize: false })})
-  .use(require('markdown-it-imsize'), {autofill: false})
+  .use(require('markdown-it-anchor'), { slugify: (str) => slugify(str, { decamelize: false }) })
+  .use(require('markdown-it-imsize'), { autofill: false })
   .use(require('markdown-it-table-of-contents'))
-  .use(require('markdown-it-task-lists'), {enabled: true})
-  .use(require('markdown-it-linkify-images'), {imgClass: 'img-fluid'})
+  .use(require('markdown-it-task-lists'), { enabled: true })
+  .use(require('markdown-it-linkify-images'), { imgClass: 'img-fluid' })
   .use(require('markdown-it-attrs'))
   .use(require('./markdown-it-dimmed'))
   .use(require('./markdown-it-radio-button'))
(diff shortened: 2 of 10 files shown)


src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -3,14 +3,17 @@ const markdownIt = require('markdown-it')({
   html: true,
   linkify: true,
   highlight: function (str, lang) {
+    console.log(str)
```
nbriannl: Will remove this

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -3,14 +3,17 @@ const markdownIt = require('markdown-it')({
...
     if (lang && hljs.getLanguage(lang)) {
+      console.log('yes');
```
nbriannl: Add this and the others

asset/css/markbind.css
```diff
@@ -24,6 +24,26 @@ pre > code.hljs {
...
+    display: inline-block;
+    padding: 0.25em 0.4em;
+    font-size: 75%;
+    line-height: 1;
+    text-align: center;
+    white-space: nowrap;
+    vertical-align: baseline;
+    border-radius: 0.25rem;
+    color: #f8f8ff;
+    background-color: #b4b4b9;
```
yamgent: Attributes should be sorted in alphabetical order.

damithc:
@nbriannl UG needs to be updated too.

asset/css/markbind.css
nbriannl: Noted.

nbriannl:
> @nbriannl UG needs to be updated too.

image
I have updated the UG as well.

yamgent:
Thanks, it looks good now. Can you please squash your commits into one? Also, please use imperative verbs when writing the titles for both the PR and the commits (i.e. "Add language indicator to code blocks")

nbriannl:
> Thanks, it looks good now. Can you please squash your commits into one? Also, please use imperative verbs when writing the titles for both the PR and the commits (i.e. "Add language indicator to code blocks")

Noted on squash-rebase workflow as well as imperative naming of the squashed commit and PR.

yamgent:
@nbriannl Thanks for your contributions!

damithc:
Good work on this feature @nbriannl 
Works as intended. However, I noticed an unanticipated problem as there is no way to turn the feature off or override the language label shown. Sometimes, we use a closely-matching language when the actual language is not supported. Here's an example:
image
Showing the language in such cases can be misleading. Is there any workaround we can implement?

yamgent:
Hmm we probably need to improve on the feature to support manipulating the label manually.

1. One possible way could be to amend the markdown syntax to allow specifying the actual language. This would make it incompatible with standard markdown syntax though.

```html
```html (MarkBind)
<box>
This is actually MarkBind, not html.
</box>
``
```

2. Alternatively, perhaps we can support a tag that changes the label's language for the corresponding code block. This should be parsed by the MarkBind's parser such that the final html output will not have these tags.

```html
<code-label language="MarkBind"/>
```html
<box>
...
</box>
``
```

And also provide a way to turn it off completely:

```html
<code-label off/>
```html
<box>
...
</box>
``
```

3. The last possible way would be to create a brand new code block component, for
advanced use. This is probably the best way, as if we want to implement a "copy" button as well, this design is probably the cleanest. However, this will take the longest to implement.

damithc:
> 1. The last possible way would be to create a brand new code block component, for
>    advanced use. This is probably the best way, as if we want to implement a "copy" button as well, this design is probably the cleanest. However, this will take the longest to implement.

We might have to go this way because we have other advanced features in mind too, such as the ability to highlight lines/words, add tool tips etc.

BTW, here's another problem I noticed.
image

I wonder if we should revert this feature for the time being.

acjh:
> 1. One possible way could be to amend the markdown syntax to allow specifying the actual language. This would make it incompatible with standard markdown syntax though.
>
> ` ```html (MarkBind)`

At least it appears to be compatible with GFM (GitHub Flavored Markdown).

Alternatively, ` ```html {.MarkBind}` based on the syntax of `markdown-it-attrs` used for inline code syntax highlighting (#557).

---

@yamgent Aside: To show Markdown fences in Markdown, use longer fences for the outer block.

`````
````html
```html (MarkBind)
<box>
This is actually MarkBind, not html.
</box>
```
````
`````

...becomes:

````html
```html (MarkBind)
<box>
This is actually MarkBind, not html.
</box>
```
````

yamgent:
> I wonder if we should revert this feature for the time being.

Yes, we will get this reverted then, the next PR should try with the new design in mind so that it can be scaled to include other things as well.

> @yamgent Aside: To show Markdown fences in Markdown, use longer fences for the outer block.

Thanks for the tip!

damithc:
> @yamgent Aside: To show Markdown fences in Markdown, use longer fences for the outer block.

TIL. Nice one @acjh

damithc:
> Yes, we will get this reverted then, the next PR should try with the new design in mind so that it can be scaled to include other things as well.

Yes, let's do that. I just found that the following problem happens a lot in one of my MarkBind sites that I need to use next semester.

> BTW, here's another problem I noticed.
> image

nbriannl:
Sorry for the late reply. Somehow I did not get notified on this in my email. Anyways, I see that my changes had caused some problems hence the revert. All noted. (Y)

PR #974, opened 2020-01-10
Fix plugin live reloading for relative file paths

### LINKED ISSUE #913: Support live-preview for .puml files

ang-zeyu:

• [x] Documentation update
• [x] Bug fix

Fixes #913 

As mentioned in #913, relative file paths for sources defined by plugins are not correctly resolved
against the page being processed. If the source is also part of an included file, then it should be resolved against the included file.

Urls are also added to the page's `pluginSourceFiles`, which has no effect, but is incorrect

Let's fix this and simplify the api for defining sources.

`getSources` now returns
```
{
  tagMap: [ ['tag name', 'src attribute name' ], ['tag name', 'src attribute name' ], ... ]
  sources: as before, which is an array of absolute / relative file paths to watch
}
```

From a plugin author's perspective, resolving said sources against the page being processed
/ the included file is tricky. This moves all such logic to `Page.js`, exposing a simpler api.

The old return syntax is deprecated, but still supported as well.

```js
getSources: () => ({
  tagMap: [['puml', 'src']],
}),
```

**Testing instructions:**
Simply use `npm run test` to run the new unit tests,

For manual testing of live reloading, the test_site has some
sample pumls at the bottom.

Fix plugin live reloading for relative file paths

Relative file paths provided by plugins are not resolved against the
processed page or included file.

This results in live reloading failing for pages not in the root
directory.

Let's fix this, and also rewrite plugin source collection, allowing
plugins to simply define tags and corresponding source attributes
to watch.

### DIFF
--- a/docs/userGuide/usingPlugins.md
+++ b/docs/userGuide/usingPlugins.md
@@ -152,22 +152,28 @@ During the `preRender` and `postRender` stages however, plugins may do custom pr
 source file types, as parsed from the raw Markdown, typically requiring rebuilding the site.
 
 Hence, to add custom source files to watch, you can implement the `getSources()` method.
-- `getSources(content, pluginContext, frontMatter)`: Returns an array of source file paths to watch. Called **before** a Markdown file's `preRender` function is called.
-  - `content`: The raw Markdown of the current Markdown file (`.md`, `.mbd`, etc.).
-  - `pluginContext`: User provided parameters for the plugin. This can be specified in the `site.json`.
-  - `frontMatter`: The frontMatter of the page being processed, in case any frontMatter data is required.
 
-Example usage of `getSources` from the PlantUML plugin:
+`getSources(content, pluginContext, frontMatter)`: Called _before_ a Markdown file's `preRender` function is called.
+- `content`: The raw Markdown of the current Markdown file (`.md`, `.mbd`, etc.).
+- `pluginContext`: User provided parameters for the plugin. This can be specified in the `site.json`.
+- `frontMatter`: The frontMatter of the page being processed, in case any frontMatter data is required.
+
+It should return an object, consisting of _at least one of the following fields_:
+- `tagMap`: An array consisting of `['tag name', 'source attribute name']` key value pairs.
+  - MarkBind will automatically search for matching tags with the source attributes, and watch them.
+  - For relative file paths, _if the tag is part of some included content_ ( eg. `<include />` tags ), it will be resolved against the included page. Otherwise, it is resolved against the page being processed.
+- `sources`: An array of source file paths to watch, where relative file paths are resolved only against the page being processed.
+- You can also directly return an array of source file paths to watch. ( ie. the `sources` field ) ___(deprecated)___
+
+Example usage of `getSources` from the PlantUML plugin, which allows insertion of PlantUML diagrams using `<puml src="..." >` tags.
+This allows files specified by the `src` attributes of `<puml>` tags to be watched:
 
 ```js
 {
   ...
-  getSources: (content) => {
-    // Add all src attributes in <puml> tags to watch list
-    const $ = cheerio.load(content, { xmlMode: true });
-
-    return $('puml').map((i, tag) => tag.attribs.src).get();
-  },
+  getSources: () => ({
+    tagMap: [['puml', 'src']],
+  })
 }
 ```
 
--- a/test/unit/Page.test.js
+++ b/test/unit/Page.test.js
@@ -1,3 +1,8 @@
+const path = require('path');
+const {
+  COLLECT_PLUGIN_SOURCES,
+  COLLECT_PLUGIN_TEST_PLUGIN,
+} = require('./utils/pageData');
 const Page = require('../../src/Page');
 
 test('Page#collectIncludedFiles collects included files from 1 dependency object', () => {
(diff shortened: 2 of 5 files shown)


docs/userGuide/usingPlugins.md
```diff
@@ -152,22 +152,32 @@ During the `preRender` and `postRender` stages however, plugins may do custom pr
...
+    - MarkBind will automatically search for matching tags with the source attributes, and watch them.
+    - For relative file paths, _if the tag is part of some included content_ ( eg. `<include />` tags ), it will be resolved against the included page.
+      Otherwise, it is resolved against the page being processed.
+  - `sources`: An array of source file paths to watch, where relative file paths are resolved only against the page being processed.
+- It is called _before_ a Markdown file's `preRender` function is called.
+- You can also directly return an array of source file paths to watch, similar to the `sources` field. ___(deprecated)___
+- **Parameters:**
+    - `content`: The raw Markdown of the current Markdown file (`.md`, `.mbd`, etc.).
+    - `pluginContext`: User provided parameters for the plugin. This can be specified in the `site.json`.
+    - `frontMatter`: The frontMatter of the page being processed, in case any frontMatter data is required.
```
yamgent: Any reason for deviating from the original ordering that the others follow (i.e. parameters, then what it returns)? Seems like the existing format can still be used here, even if it has more information compared to the rest like `getLinks()` and `getScripts()`.

test/unit/Page.test.js
```diff
@@ -20,3 +26,29 @@ test('Page#collectIncludedFiles collects nothing', () => {
...
+  expect(page.pluginSourceFiles.has(
+    path.resolve('/except/absolute/sources.c'))).toEqual(true);
+  // tagMap
+  expect(page.pluginSourceFiles.has(
+    path.resolve('/root/images/sample1.png'))).toEqual(true);
+  expect(page.pluginSourceFiles.has(
+    path.resolve('/root/subdir/images/sample2.png'))).toEqual(true);
+  expect(page.pluginSourceFiles.has(
+    path.resolve('/root/images/sample3.png'))).toEqual(true);
+  expect(page.pluginSourceFiles.has(
```
yamgent: Seems that the `pluginSourceFiles` variable should just be a `Set`.

test/unit/utils/pageData.js
```diff
@@ -0,0 +1,29 @@
+module.exports.COLLECT_PLUGIN_SOURCES = `
+<custom-tag-one src="images/sample1.png">
+  Lorem ipsum dolor sit amet, Ut enim ad minim veniam, 
+  <div data-included-from="/root/subdir/includedpage.md">
+    consectetur adipiscing elit, 
+    <div>
+      <custom-tag-two srcattr="images/sample2.png">
+        sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+      </custom-tag-two>
```
yamgent: For a sanity check, maybe have another include inside the include to ensure that it resolves correctly even if it involves 2 levels deep?

ang-zeyu:
Updated, thanks for looking through it!

docs/userGuide/usingPlugins.md
```diff
@@ -152,22 +152,28 @@ During the `preRender` and `postRender` stages however, plugins may do custom pr
...
+
+It should return an object, consisting of _at least one of the following fields_:
+- `tagMap`: An array consisting of `['tag name', 'source attribute name']` key value pairs.
+  - MarkBind will automatically search for matching tags with the source attributes, and watch them.
+  - For relative file paths, _if the tag is part of some included content_ ( eg. `<include />` tags ), it will be resolved against the included page.
+    Otherwise, it is resolved against the page being processed.
+- `sources`: An array of source file paths to watch, where relative file paths are resolved only against the page being processed.
+- You can also directly return an array of source file paths to watch ( ie. the `sources` field ) ___(deprecated)___
 
 Example usage of `getSources` from the PlantUML plugin:
```
yamgent: I feel that the reader might have a bit of difficulty understanding the example without context. Maybe an addition like this will make it clearer?

"Example usage of `getSources` from the PlantUML plugin, which will watch files from tags such as `<puml src="..." />`:"

docs/userGuide/usingPlugins.md
```diff
@@ -152,22 +152,28 @@ During the `preRender` and `postRender` stages however, plugins may do custom pr
...
+- `pluginContext`: User provided parameters for the plugin. This can be specified in the `site.json`.
+- `frontMatter`: The frontMatter of the page being processed, in case any frontMatter data is required.
+
+It should return an object, consisting of _at least one of the following fields_:
+- `tagMap`: An array consisting of `['tag name', 'source attribute name']` key value pairs.
+  - MarkBind will automatically search for matching tags with the source attributes, and watch them.
+  - For relative file paths, _if the tag is part of some included content_ ( eg. `<include />` tags ), it will be resolved against the included page.
+    Otherwise, it is resolved against the page being processed.
+- `sources`: An array of source file paths to watch, where relative file paths are resolved only against the page being processed.
+- You can also directly return an array of source file paths to watch ( ie. the `sources` field ) ___(deprecated)___
```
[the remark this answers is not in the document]
ang-zeyu: Hmm, I actually meant only to deprecate the old syntax, not the functionality of returning some manually specified file paths (where relative ones are resolved according to the file processed).

This can be useful if a plugin needs live reload for a file that isn't specified as some `src` attribute in the dom. E.g. a plugin where you can have `file_name`+`_page_extension` that adds some dynamic content to the relavant page based on that file's contents.

ang-zeyu:
Updated, save for the deprecated syntax parts

docs/userGuide/usingPlugins.md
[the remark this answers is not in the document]
yamgent: Hmm ok, the sentence standalone feels a bit difficult to comprehend without your additional explanation (because of the `(ie. the sources field)` portion), but I am fine with leaving it as it is for the moment.

Issue #983, opened 2020-01-16
Reading progress indicator 


yash-chowdhary:
**Is your request related to a problem?**

This is a nice-to-have feature. It's a nice way to show how far along the user is on a page. The page navigation doesn't quite capture the same thing, as it only shows the headers and sub-headers of the content.

**Describe the solution you'd like**

The author can have the option to add a progress indicator to a page. 

Currently, I started by adding the following markup to the page - 
```html
<div class="progress"></div>
```

The CSS and the JS will already be available when `markbind init` is executed should the author ever need to use this feature.

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/21968718/72519949-596ec100-3893-11ea-86ac-452386002eb3.gif)

**Describe alternatives you've considered**

Perhaps it can be implemented through _Front Matter_ as a boolean flag? This would avoid having to add the markup in every page that the progress indicator is desirable. This would also ensure that the authors don't need to deal with HTML at all. 

I am yet to try this alternative approach. I'd like to get some feedback first.

**Additional context**

None.

yamgent:
> Perhaps it can be implemented through _Front Matter_ as a boolean flag? This would avoid having to add the markup in every page that the progress indicator is desirable. This would also ensure that the authors don't need to deal with HTML at all.

Seems like #944 will solve the problem of having to add the markup at every page. In fact the author can also use `header.md` to add the markup without having to repeat themselves. Also authors are already dealing with HTML when using vue components.

Perhaps the syntax can be a bit friendlier to the author, so that it is as if it is another vue component? Just a thought.

yash-chowdhary:
>  In fact the author can also use header.md to add the markup without having to repeat themselves.

Ah yes, you're right. 

> Perhaps the syntax can be a bit friendlier to the author, so that it is as if it is another vue component? Just a thought.

This could be done too. However, since it is just a one-liner, wouldn't it be a bit overkill to create a vue component for it?

nbriannl:
An update to this issue should anyone come across this
Based on the discussion in #985, it is generally concluded that 'there's very little value in having this in' MarkBind.

yash-chowdhary:
Closing - the reason is mentioned above.

PR #985, opened 2020-01-18
Add progress indicator feature

yash-chowdhary:

• [x] New feature

This PR adds the progress indicator as outlined in #983. This gives authors the choice to add a reading progress indicator to pages.

Added `.progress` CSS class in `markbind.css` that styles the progress bar. 
Added logic in `setup.js` to get the scroll percentage in a webpage.
In order to enable the progress bar in the pages, the following markup can be added to the header file:
```html
<div class="progress"></div>
```

```js
function setupReadingProgress() {
  const h = document.documentElement;
  const b = document.body;
  const st = 'scrollTop';
  const sh = 'scrollHeight';
  const progress = document.querySelector('.progress');
  let scroll;

  document.addEventListener('scroll', () => {
    scroll = Number(h[st] || b[st]) / Number((h[sh] || b[sh]) - h.clientHeight);
    scrollPercentage = scroll * 100;
    progress.style.setProperty('--scroll', `${scrollPercentage}%`);
  });
}
```

**Testing instructions:**
* Run `markbind init` 
* Add the markup to the header file (`_markbind/headers/header.md`)
* Run `markbind serve` to generate the site
* Scroll to see the progress indicator

Add reading progress indicator


asset/js/setup.js
```diff
@@ -95,6 +95,21 @@ function setupSiteNav() {
   );
 }
 
+function setupReadingProgress() {
+  const h = document.documentElement;
```
marvinchin: Could we use more descriptive variable names here?

asset/js/setup.js
```diff
@@ -95,6 +95,21 @@ function setupSiteNav() {
...
+  const b = document.body;
+  const st = 'scrollTop';
+  const sh = 'scrollHeight';
+  const progress = document.querySelector('.progress');
+  let scroll;
+
+  document.addEventListener('scroll', () => {
+    scroll = Number(h[st] || b[st]) / Number((h[sh] || b[sh]) - h.clientHeight);
```
marvinchin: Is there a reason why we need to use `Number` here?

asset/js/setup.js
```diff
@@ -95,6 +95,21 @@ function setupSiteNav() {
   );
 }
 
+function setupReadingProgress() {
+  const h = document.documentElement;
+  const b = document.body;
+  const st = 'scrollTop';
+  const sh = 'scrollHeight';
+  const progress = document.querySelector('.progress');
+  let scroll;
```
marvinchin: Please correct me if I'm wrong, but could we make this a `const` within the event listener instead? We don't seem to need the previous scroll value to compute the new one.

asset/js/setup.js
```diff
@@ -95,6 +95,21 @@ function setupSiteNav() {
   );
 }
 
+function setupReadingProgress() {
+  const h = document.documentElement;
+  const b = document.body;
+  const st = 'scrollTop';
```
marvinchin: Is there a reason we are assigning these constant field names to a variable? Why not just do `document.body.scrollTop` etc. instead? It might be more verbose, but I think that is more readable compared to using ultra-short variable names.

asset/js/setup.js
marvinchin: I think it might also be worth splitting this assignment up into several steps. There is a lot of logic going on here that is not immediately obvious.

asset/js/setup.js
yash-chowdhary: I've removed them altogether. They're not really required.

asset/js/setup.js
yash-chowdhary: In the previous iteration, I had set `scroll` to the expression which is now separated into 2 lines - defining `scroll` and `scrollPercentage`. This caused some lint issues with the operators. That's why I used `Number`. 
I have now removed the `Number` typecasting.
I have also switched to the `.` notation for accessing the properties.

asset/js/setup.js
yash-chowdhary: Yes, you're right. :) Resolved.

asset/js/setup.js
yash-chowdhary: Resolved. :)

yash-chowdhary:
Thanks for your review, @marvinchin :) 
Refactored the code and added the progress bar to `test_site`.

asset/js/setup.js
```diff
@@ -95,6 +95,18 @@ function setupSiteNav() {
   );
 }
 
+function setupReadingProgress() {
+  const progress = document.querySelector('.progress');
+
+  document.addEventListener('scroll', () => {
+    const amountScrolled = (document.documentElement.scrollTop || document.body.scrollTop);
+    const pageHeight = ((document.documentElement.scrollHeight || document.body.scrollHeight)
```
marvinchin: Just a nit - I don't think you need the outermost brackets.

docs/userGuide/syntax/headers.mbdf
```diff
@@ -23,6 +23,18 @@ In the page that you want to include the header:
 ```
 </div>
 
+You can also choose to add a <trigger for="pop:progress-indicator">reading progress indicator</trigger> to let the reader know how far along they are in the page.
```
yamgent: Due to how our current search works, your feature won't be discoverable if the authors attempt a search on our documentation website. I suggest giving a heading to your section:

`### Adding progress indicators`

yash-chowdhary: Sure, will do 👍

asset/js/setup.js
```diff
@@ -95,6 +95,18 @@ function setupSiteNav() {
   );
 }
 
+function setupReadingProgress() {
+  const progress = document.querySelector('.progress');
+
+  document.addEventListener('scroll', () => {
```
marvinchin: Oops, one more thing! We should only add the event listener when the progress indicator is on the page. Otherwise, `progress` might not be defined.
Screenshot 2020-01-20 at 5 04 47 PM

openorclose:
This causes quite a big jump when the content height is changed (such as with panels or on resize).

Is it possible to respond to these events immediately rather than waiting for a scroll event to occur?
prog

asset/js/setup.js
yash-chowdhary: Fixed

yash-chowdhary:
Thanks for the observation @openorclose. :) 
I've added a hacky feature that will periodically check for document changes (as outlined [here](https://stackoverflow.com/questions/14866775/detect-document-height-change)), as there is no straightforward solution to checking this.
I've also added a listener for window resizing. 
Ready for review.

damithc:
@yash-chowdhary this is a very low-value feature IMO (note most other sites don't have such a feature). Therefore, make sure there is no performance cost on those who don't use the feature and the future code maintenance cost is VERY low. If that's not possible, the right thing to do is to shelve this feature.

yash-chowdhary:
> make sure there is no performance cost on those who don't use the feature

@damithc There are no performance issues for websites/pages that don’t use the progress indicator as the listeners are only added if and when the indicator is enabled.

damithc:
> @damithc There are no performance issues for websites/pages that don’t use the progress indicator as the listeners are only added if and when the indicator is enabled.

That's good. The amount of code changes in this PR is still a bit worrying though. Feels like the cost is not worth the benefit. All this code needs to be maintained in future. To clarify, I feel the value of the feature is low because the vertical scrollbar of the browser also serves as a similar indicator of progress through the page. I didn't object to this feature at first because I thought it is going to be simple to implement. What do you guys think?

ang-zeyu:
> What do you guys think?

I think we should keep it, in the future we could even extend this with something similar to https://bootstrap-vue.js.org/docs/directives/scrollspy/ to provide more value over the browser scrollbar.
( e.g. a popover that appears on scrolling indicating the current 'topic' being viewed, then fades away after a while so it dosen't affect readability )

> I've added a hacky feature that will periodically check for document height changes (as outlined [here](https://stackoverflow.com/questions/14866775/detect-document-height-change)), as I couldn't find a straightforward solution to this.
 
How about setting a simple transition duration for the width? Then instead of modifying `--scroll` you would simply modify the width. https://www.w3schools.com/howto/howto_js_scroll_indicator.asp

It's a compromise for performance though, since it would only smooth out the sharp transition once the user scrolls after opening a panel. But personally I would prefer this over the performance cost of `setTimeout`

One of my mice is rather hard to scroll, and I use a larger lines-per-scroll setting to compensate for that. It may help to even out the transition for such cases as well.

yash-chowdhary:
> How about setting a simple transition duration for the width? Then instead of modifying `--scroll` you would simply modify the width. https://www.w3schools.com/howto/howto_js_scroll_indicator.asp
 
I've tested it out with different values. This would help minimize the jump.

yash-chowdhary:
@marvinchin @yamgent any thoughts on how to proceed?

yamgent:
> @marvinchin @yamgent any thoughts on how to proceed?

Besides the vertical scrollbar cons that Prof has mentioned, the bar might also be confused with the content loading bar (e.g. the loading bar on the top that YouTube uses when navigating between pages). The content loading bar is also a common visual design on mobile phones. [This article](https://uxdesign.cc/pros-and-cons-of-progress-indicator-as-a-scroll-bar-345f19967cb6) provides some arguments about the pros and cons of a progress indicator.

I am inclined to agree with Prof that this feature might not see widespread use, so there's very little value in having this in.

yash-chowdhary:
Closing as a result of the above discussion.

PR #991, opened 2020-01-20
Support line numbers for code blocks

### LINKED ISSUE #473: Support annotations in code examples
Current: code snippets can be syntax colored

Suggested: add the ability to add more information into code snippets, for example, 
* highlight specific parts of the code
* add a comment about a specific part of a code
* show line numbers

### LINKED ISSUE #1023: markdown-it-attrs not working for code fences

* **MarkBind Version:**

````md
```js {.someClass }
const js = 1;
```
````

```html
<pre><code class="hljs js someClass"><span class="hljs-keyword">const</span> js = <span class="hljs-number">1</span>;
</code></pre>
```

```html
<pre><code class="hljs js"><span class="hljs-keyword">const</span> js = <span class="hljs-number">1</span>;
</code></pre>
```

Markdown-it-attrs are not being rendered, `someClass` did not get added.

openorclose:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Partially fixes #473 (only adding support for line nos),

Fixes #1023 

# Small bugfix: 

Our code fences (` ``` `) do not support [markdown-it-attrs](https://github.com/arve0/markdown-it-attrs) even though one of the examples there shows it should. 

Turns out this is due to our overriding of the syntax highlighting.

image

We copied the second example from markdown-it, so that we could add our own `hljs` class to the `<pre>` tag, but this breaks markdown-it-attrs since the attrs there do not get added to our hardcoded `pre` tag.

So instead we provide our own renderer to render the attributes parsed by markdown-it-attrs.

# Actual implementation

There is no way to directly use CSS and determine when line breaks happen, so we change `highlightLines` in markdown-it to split every `\n` and wrap a `span` around them.

Then, we use some css to insert the line numbers to it.

Line numbers will be shown by default, and authors can disable it by adding the class `no-line-numbers`.

````
```js
function f(n) {
  if (n < 1) {
    return 1;
  } else {
    return f(n - 1) + f(n - 2);
  }
}
f(10); // returns 12345;
```
````

Before and After:

image

If line numbers are not wanted:

````
```js {.no-line-numbers}
function f(n) {
  if (n < 1) {
    return 1;
  } else {
    return f(n - 1) + f(n - 2);
  }
}
f(10); // returns 12345;
```
````

Check out the 2103 textbook here, where all code snippets have line numbers:
(the search on top doesn't work, it will bring you to the actual 2103 book site)
https://openorclose.github.io/se-book/

These pages have many code snippets:
- https://openorclose.github.io/se-book/codeQuality/
- https://openorclose.github.io/se-book/oop/
- https://openorclose.github.io/se-book/testing/
- https://openorclose.github.io/se-book/cppToJava/

**Testing instructions:**
Verify that `<span>`s get added to html of code fences, and manually check generated code fences for line numbers.

````
Add line numbers for code blocks

Let's wrap each line of a code block in a `<span>`,
and then add relevant css styles to display line numbers beside.

Line numbers will be displayed by default.

To turn off line numbers, add the class `no-line-numbers`
to a code block via markdown-it-attrs:

```js {.no-line-numbers}
const js = 1;
```
````
// TODO when design finalised

### CODE REVIEW

**Comment by damithc** [comment]
Nice. Can we make it the default? i.e., all fenced code blocks will have line numbers by default.

**Comment by openorclose** [comment]
> Nice. Can we make it the default? i.e., all fenced code blocks will have line numbers by default.

Sure, then users can disable it similarly (or add any other classes they want) by doing `{.no-line-numbers}`?

**Comment by yamgent** [comment]
> Sure, then users can disable it similarly (or add any other classes they want) by doing `{.no-line-numbers}`?

Yes please do provide the option to disable it.

**Comment by openorclose** [comment]
Ok changes made,

Now all code fences that specify a language will have numbered lines.

Add the class `no-line-numbers` to disable.

File: src/lib/markbind/src/lib/markdown-it/index.js [hunk 1]
```diff
@@ -2,17 +2,16 @@ const hljs = require('highlight.js');
...
-      } catch (__) {
-      }
+        str = hljs.highlight(lang, str).value;
+      } catch (__) {}
     }
-
-    return '<pre><code class="hljs">' + markdownIt.utils.escapeHtml(str) + '</code></pre>';
+    const lines = str.split('\n');
+    lines.pop(); // last line is always a single '\n' newline, so we remove it
+    return `<span>${lines.join('</span><span>')}</span>`; //wrap all lines with <span> so we can number them
```
marvinchin: I personally think using `map` to wrap each line and then joining it would be more readable 🙂 
```
lines.map((line) => `<span>${line}</span>`).join('')
```

openorclose: Sure!

asset/css/markbind.css
```diff
@@ -297,3 +297,28 @@ panel[minimized]:not([expanded])::before {
 li.footnote-item:target {
     background-color: #eee;
 }
+
+pre {
```
alyip98: Should this rule be changed to apply on `.hljs` instead? I doubt it will have an observable effect but it would be better practice.

alyip98:
What do you think of supporting arbitrary starting line numbers? I could see some use cases where the author wants to include a short code snippet without including the whole file. Implementation should be easy with the `counter-set` property.

yamgent:
This PR is breaking the rendering of some code snippets, and also some scrollbars just appearing randomly:

image

openorclose:
> This PR is breaking the rendering of some code snippets, and also some scrollbars just appearing randomly:
> 
> image

When a language is not specified, the .hljs class is not applied by markdown it, which is probably why the old solution was used in the first place to add the .hljs class to code fences with no language. Bc

But reverting the change would break markdown-it-attrs for code fences.

yamgent:
> But reverting the change would break markdown-it-attrs for code fences.

Is it possible for MarkBind to add the class back in after all markdown syntax has been rendered (i.e. some post-rendering processing)?

openorclose:
> Is it possible for MarkBind to add the class back in after all markdown syntax has been rendered (i.e. some post-rendering processing)?

I made code highlighting a new renderer rule instead of using the `highlight` property from markdown-it, so it works fine now.

image.

This allows us to manually render the attributes parsed from `markdown-it-attrs`.




> What do you think of supporting arbitrary starting line numbers? I could see some use cases where the author wants to show a short code snippet without showing the whole file.

This would require a separate markdown plugin, as it is not possible (i.e. not nice) with pure css.
I would do this along with the rest of #473 by introducing some new syntax like markdown-it-attrs in a separate PR.

On a separate note, why do inline styles not work with markdown it attrs?

````
```js {data=123 style="counter-set:90;"}
function f(n) {
  if (n < 1) {
    return 1;
  } else {
    return f(n - 1) + f(n - 2);
  }
}
f(10); // returns 12345;
```
````

I tried doing this,

image

The inline style does get parsed and rendered in renderAttrs, but gets stripped out in the final html:

image


I tried it with `**hi** {style = "color:red"}` too and the the style doesn't work there too.

**Comment by damithc** [comment]
This page seems to be broken https://deploy-preview-991--markbind-master.netlify.com/userguide/reusingcontents

**Comment by openorclose** [comment]
> This page seems to be broken https://deploy-preview-991--markbind-master.netlify.com/userguide/reusingcontents

Forgot to manually escape code blocks without language, Thanks!

File: asset/css/markbind.css (hunk 2 above)
Comment (openorclose) [reply asset/css/markbind.css]: You mean `pre.hljs` instead?

**Comment by openorclose** [comment]
Ready for review!

**Comment by damithc** [comment]
Good work so far; looking forward to using this feature.
In some cases, there's a lot of space between the numbers and the code?
image

**Comment by openorclose** [comment]
> Good work so far; looking forward to using this feature.
> In some cases, there's a lot of space between the numbers and the code?
> image

I set it to to the size of 2 digits, since that's probably enough for most purposes:

This is what it looks like if I set the width to 2 digits:
image

What it looks like now:
image

To solve the space issue when there is only one digit, I changed the text to align right instead:
image

File: asset/css/markbind.css [hunk 3]
```diff
@@ -298,7 +298,7 @@ li.footnote-item:target {
     background-color: #eee;
 }
 
-pre {
+pre.hljs {
```
alyip98: The class `.hljs` is applied to the `<code>` inside of `<pre>`

openorclose: Fixed! thanks

openorclose:
> Some parts of the documentation have line numbers, but we don't really want that because they are not code (e.g. https://deploy-preview-991--markbind-master.netlify.com/userguide/). Let's disable them.

Thanks! I've edited them, looked for instances of ` ``` ` and couldn't find other similar pages.

openorclose:
Fixes #1017, too with the added 
```
pre {
    white-space: pre-wrap;
}
```
suggested in that issue

yamgent:
There are some additional rendering problems in the documentation, can you please take a look at these:

(Both of these pictures are in `userGuide/formattingContents.html`, left side is this PR, right side is the correct version)

blockquotes

rules

openorclose:
> There are some additional rendering problems in the documentation, can you please take a look at these:
> 
> (Both of these pictures are in `userGuide/formattingContents.html`, left side is this PR, right side is the correct version)
> 
> blockquotes
> 
> rules

Thanks!

This is due to hljs's rendering into `<span>`s sometimes containing line breaks:
image
```
*****
-----
```
gets rendered as one single span, so we can't just split the lines after highlighting.


So I decided to split the lines, then perform highlighting instead.

This causes a minor discrepancy with blockquotes (the single `>` is not highlighted.
image


Horizontal and vertical spaces are fixed
image

damithc:
Works in production. Nice! 👍 Thanks @openorclose and reviewers. Let's push ahead and make our code blocks the _best of class_!

PR #1000, opened 2020-01-27
Add support for Octicons

### LINKED ISSUE #998: Support Octicons
Explore if it is possible to support https://octicons.github.com/

dvrylc:

• [X] New feature

Implements and closes #998. 

[GitHub's Octicons](https://octicons.github.com/) are a set of icons that represent Git and GitHub-specific features, useful when writing technical documents (about software engineering for example). 

- Add support for Octicons (with the `:octicon-(name):` syntax) using the official [Octicons Node library](https://github.com/primer/octicons/tree/master/lib/octicons_node)
- Add unit tests for Octicons
- Update the functional test sites with the Octicon stylesheet (via `npm run updatetest`)
- Update the user guide with examples for Octicon usage

Example: 
Screen Shot 2020-01-27 at 10 26 22 PM

```js
// Extends existing icon parsing logic
const octicons = require('@primer/octicons');

    /:(fa[brs]|glyphicon|octicon)-([a-z-]+):/,
    (match, utils) => {
        let iconFontType = match[1];
        let iconFontName = match[2];

            return octicons[iconFontName].toSVG();
        } else { // If icon is a Font Awesome icon
            return `<span aria-hidden="true" class="${iconFontType} fa-${iconFontName}"></span>`;
        }
    }
);
```

`Site.js#Site.prototype.copyOcticonsAsset`

The Octicons Node library distributes an unminified stylesheet (identical to [this](https://github.com/primer/octicons/blob/master/lib/octicons_node/index.scss)) that uses 2 spaces for indentation. If the stylesheet is copied directly to the output site (similar to how it's done for Font Awesome and Glyphicons), it causes stylelint errors for the test sites after executing `npm run updatetest` because stylelint expects 4 spaces for indentation. 

Due to this, I had to manually "minify" the CSS file by stripping out all whitespace during the copy. 

**Testing instructions:**
- Use the `:octicon-(name):` syntax to display any Octicon
- Verify that the unit and functional tests are correct, and that `npm run test` is passing

Add support for Octicons

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -235,7 +235,7 @@ class Site {
                                   path.join(this.siteAssetsDestPath, 'glyphicons', 'css',
                                             'bootstrap-glyphicons.min.css')),
         octicons: path.relative(path.dirname(resultPath),
-                                path.join(this.siteAssetsDestPath, 'css', 'octicons.css')),
+                                path.join(this.siteAssetsDestPath, 'css', 'octicons.min.css')),
         highlight: path.relative(path.dirname(resultPath),
                                  path.join(this.siteAssetsDestPath, 'css', 'github.min.css')),
         markbind: path.relative(path.dirname(resultPath),
@@ -874,9 +874,18 @@ class Site {
   copyOcticonsAsset() {
     const octiconsRootSrcPath = path.join(__dirname, '..', 'node_modules', '@primer', 'octicons', 'build');
     const octiconsCssSrcPath = path.join(octiconsRootSrcPath, 'build.css');
-    const octiconsCssDestPath = path.join(this.siteAssetsDestPath, 'css', 'octicons.css');
+    const octiconsCssDestPath = path.join(this.siteAssetsDestPath, 'css', 'octicons.min.css');
 
-    return fs.copyAsync(octiconsCssSrcPath, octiconsCssDestPath);
+    const minifiedCss = fs.readFileSync(octiconsCssSrcPath).toString().replace(/\s/g, '');
+
+    return new Promise((resolve, reject) => {
+      try {
+        fs.writeFileSync(octiconsCssDestPath, minifiedCss);
+        resolve();
+      } catch (error) {
+        Site.rejectHandler(reject, error, []);
+      }
+    });
   }
 
   /**
--- a/src/Site.js
+++ b/src/Site.js
@@ -234,6 +234,8 @@ class Site {
         glyphicons: path.relative(path.dirname(resultPath),
                                   path.join(this.siteAssetsDestPath, 'glyphicons', 'css',
                                             'bootstrap-glyphicons.min.css')),
+        octicons: path.relative(path.dirname(resultPath),
+                                path.join(this.siteAssetsDestPath, 'css', 'octicons.css')),
         highlight: path.relative(path.dirname(resultPath),
                                  path.join(this.siteAssetsDestPath, 'css', 'github.min.css')),
         markbind: path.relative(path.dirname(resultPath),
@@ -526,6 +528,7 @@ class Site {
         .then(() => this.buildSourceFiles())
         .then(() => this.copyMarkBindAsset())
         .then(() => this.copyFontAwesomeAsset())
+        .then(() => this.copyOcticonsAsset())
         .then(() => this.copyLayouts())
         .then(() => this.updateSiteData())
         .then(() => {
@@ -863,6 +866,17 @@ class Site {
     const faFontsDestPath = path.join(this.siteAssetsDestPath, 'fontawesome', 'webfonts');
 
     return fs.copyAsync(faCssSrcPath, faCssDestPath).then(fs.copyAsync(faFontsSrcPath, faFontsDestPath));
+  }
+
+  /**
+   * Copies Octicon assets to the assets folder
+   */
+  copyOcticonsAsset() {
+    const octiconsRootSrcPath = path.join(__dirname, '..', 'node_modules', '@primer', 'octicons', 'build');
+    const octiconsCssSrcPath = path.join(octiconsRootSrcPath, 'build.css');
+    const octiconsCssDestPath = path.join(this.siteAssetsDestPath, 'css', 'octicons.css');
+
+    return fs.copyAsync(octiconsCssSrcPath, octiconsCssDestPath);
   }
 
   /**
--- a/src/Site.js
+++ b/src/Site.js
@@ -235,7 +235,7 @@ class Site {
                                   path.join(this.siteAssetsDestPath, 'glyphicons', 'css',
                                             'bootstrap-glyphicons.min.css')),
         octicons: path.relative(path.dirname(resultPath),
-                                path.join(this.siteAssetsDestPath, 'css', 'octicons.min.css')),
+                                path.join(this.siteAssetsDestPath, 'css', 'octicons.css')),
         highlight: path.relative(path.dirname(resultPath),
                                  path.join(this.siteAssetsDestPath, 'css', 'github.min.css')),
         markbind: path.relative(path.dirname(resultPath),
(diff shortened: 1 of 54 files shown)


src/Site.js
```diff
@@ -894,6 +897,26 @@ Site.prototype.copyFontAwesomeAsset = function () {
   return fs.copyAsync(faCssSrcPath, faCssDestPath).then(fs.copyAsync(faFontsSrcPath, faFontsDestPath));
 };
 
+/**
+ * Minfies and copies Octicon assets to the assets folder
```
marvinchin: I think you misspelt minifies 😛

dvrylc: 🤣 fixed!

dvrylc:
> I just wonder whether the value we get from having so many different icon packs is worth the additional cost of having to load all the assets. 

In the case of Octicons, the additional cost is actually quite minimal because the extra stylesheet is tiny and the icons are inline SVG - ensuring that only icons that are used are included in the output. 

But I agree with your point that we might need a smarter way of loading assets, especially if we intend to support more large icon libraries in the future. 

Also side note: the Travis build seems to be failing now even though the most recent commit only changed a comment 🤔 I can't reproduce this on my local machine so I suspect it might a bug on Travis, are you able to restart the build?

marvinchin:
@dvrylc I've restarted the build, but it still failed. Do you want to investigate it further?

dvrylc:
Turns out that tests were failing because of a new test site introduced in bec1916c9a086fefa318a2780edc972481311da1. Rebased and updated the new test site with the Octicon stylesheet. 

@marvinchin should be good to go now 👍

src/Site.js
```diff
@@ -894,6 +897,26 @@ Site.prototype.copyFontAwesomeAsset = function () {
...
 
+/**
+ * Minifies and copies Octicon assets to the assets folder
+ */
+Site.prototype.copyOcticonsAsset = function () {
+  const octiconsRootSrcPath = path.join(__dirname, '..', 'node_modules', '@primer', 'octicons', 'build');
+  const octiconsCssSrcPath = path.join(octiconsRootSrcPath, 'build.css');
+  const octiconsCssDestPath = path.join(this.siteAssetsDestPath, 'css', 'octicons.min.css');
+
+  const minifiedCss = fs.readFileSync(octiconsCssSrcPath).toString().replace(/\s/g, '');
```
marvinchin: Sorry I missed this out previously.

I don't think we should be manually minifying CSS like this. It might be fine _for now_, but it might introduce some very hard to find errors if any whitespace sensitive CSS properties are ever used in the Octicon CSS file. 

If there is no minified CSS file included in the library, I think we can just use the un-minified CSS for now since it's not very big. Perhaps we should look into programatically minifying all of our CSS and JS assets during the build phase.

dvrylc: That's actually the approach I initially went with. The problem is - The unminified CSS uses 2 spaces for indentation, which causes stylelint violations (expects 4 spaces) when the test sites are updated with the Octicon CSS file (via `npm run updatetest`). 

I guess here we have a few options: 

1. If we're worried about self-minification being fragile, we can consider bringing in a library to help with this - [`clean-css`](https://github.com/jakubpawlowicz/clean-css) for example. 

2. Add rules for stylelint to ignore the functional test sites.

marvinchin: Ah that's unfortunate. Applying our own linting rules to external dependencies will probably always be an exercise in frustration. I'm leaning towards making stylelint ignore the functional test sites. 

@yamgent do you have any thoughts on this?

dvrylc: @marvinchin how should we proceed from here?

marvinchin: Let's go ahead and disable the lint rules for the test sites (lets try to make the excluded paths as specific as possible). We should do that in a separate PR, and then rebase this PR on top of that.

dvrylc: @marvinchin I was thinking, we could simply name the Octicon CSS file `octicons.min.css` when copying without actually minifying it. 

This allows stylelint to ignore it (because it ends with `min.css`, just like all our other external libraries) without having to modify the project configuration just for one rare case of a library that doesn't distribute minified files. It also allows us to continue linting the test sites without having to add library-specific ignore rules. What do you think?

marvinchin: @dvrylc I’d prefer not to do that because the CSS file is not actually minified, so naming it as such would be confusing. (It would also be achieving the same outcome of “disabling” the linter for the file, but in a more hacky way)

Not linting generated files seem to make sense to me because some of these generated files might be minified, dependent on external assets etc. and would not satisfy our linting rules (it wouldn’t make sense to lint node_modules for similar reasons!). We should only be applying linting rules to the source files that we are working on.

Also, to clarify - I’m suggesting excluding all generated files from the linter, and not specifically this file in particular.

Happy to discuss this further if you have any other thoughts or concerns!

damithc:
1000th post (issue/pr) in the MarkBind issue tracker 🎉 😀

src/Site.js
dvrylc: Ah I see your point, sounds good! I'll rebase and update this once #1011 is merged.

damithc:
Good work @dvrylc 
Is there any noteworthy performance cost? If there is, should we make this a plugin?

dvrylc:
> Is there any noteworthy performance cost? If there is, should we make this a plugin?

@damithc the following is a breakdown of the possible extra costs. In my opinion, they are not significant and the reasons are listed below 👇 

1. Extra [`octicon.css`](https://github.com/primer/octicons/blob/master/lib/octicons_node/index.scss) file

    This extra CSS file is actually very small, I doubt it will cause any performance impact. 

2. Inline SVG icons

    Unlike Font Awesome and Glyphicons that are implemented by [including a webfont](https://github.com/MarkBind/markbind/blob/31dfb1ac62086b6dd5a46c9eb243f5e9091e53c4/src/Site.js#L855-L866) and then using [`<span>`](https://github.com/MarkBind/markbind/blob/31dfb1ac62086b6dd5a46c9eb243f5e9091e53c4/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js#L8) elements to display the icon, Octicons are implemented by [inlining the desired icon's SVG code directly into the page](https://github.com/MarkBind/markbind/blob/f813145178c9178c6f2e7427697bcff522211f12/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js#L11-L12). 

    This means that a page using Octicons will have a slightly larger source file based on how many Octicons it uses. The largest benefit of this approach is that instead of loading an entire webfont regardless of whether the page uses those icons, only the ones that are used will be included and nothing else. And since the individual chunks of SVG are much smaller than a webfont, I think this is a reasonable approach.

damithc:
> @damithc the following is a breakdown of the possible extra costs. In my opinion, they are not significant and the reasons are listed below 👇

Thanks for the investigation @dvrylc 

@yamgent I'm OK to go ahead with this.

damithc:
Tried, and seem to work. Nice work @dvrylc. Some followup issues:
* They seem to have [three modes dark/light/grid](https://octicons.github.com/icon/bug/). Is there any way to support those modes? Not important though.
* Doesn't seem to work with the [icon-size property of boxes](https://markbind.org/userGuide/usingComponents.html#boxes).

damithc:
> * They seem to have [three modes dark/light/grid](https://octicons.github.com/icon/bug/). Is there any way to support those modes? Not important though.
> * Doesn't seem to work with the [icon-size property of boxes](https://markbind.org/userGuide/usingComponents.html#boxes).

@dvrylc any updates about the above?

dvrylc:
@damithc Apologies for missing out on this for so long! Got lost in the heaps of GitHub notifications. 

> * They seem to have [three modes dark/light/grid](https://octicons.github.com/icon/bug/). Is there any way to support those modes? Not important though.

The page that you linked shows some examples of how the icon can be used - on dark and light backgrounds, and the grid shows how the icon is constructed. Octicons don't have "modes" unlike Font Awesome, and all octicons are actually just SVG paths that can be styled by setting CSS properties like `color` and `height` on them. Seems like there's already some work being done on this in #1122. 

> * Doesn't seem to work with the [icon-size property of boxes](https://markbind.org/userGuide/usingComponents.html#boxes).

I'll look into this and submit a PR to vue-strap. Sorry again about the delay!

damithc:
Thanks for the follow up @dvrylc

PR #1002, opened 2020-01-29
Allow using 'none' footer attribute in frontmatter

### LINKED ISSUE #638: Allow using an empty footer attribute in frontmatter to omit the default footer
The following should make the default footer (inherited via the default layout) to be omitted from the page? In general, an empty `x` attribute should indicate `no x`?
```
<frontmatter>
footer:
</frontmatter>
```

yash-chowdhary:

- [x] Enhancement to an existing feature

Fixes #638 

Enables users to exclude the footer that is derived/inherited from the layout specified in the `<frontmatter>`.

Added a check for when the `footer` attribute is `none` while inserting the footer file into a `Page`.

**Testing instructions:**
* Specify a layout in a page's `<frontmatter>`
* Add `footer: none` to `<frontmatter>`
* Build and serve site

Expected output: Footer of the derived layout is omitted

Allow using 'none' footer attribute in frontmatter

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -500,7 +500,7 @@ class Page {
     if (footer === FRONT_MATTER_NONE_ATTR) {
       return pageData;
     }
-    
+
     let footerFile;
     if (footer) {
       footerFile = path.join(FOOTERS_FOLDER_PATH, footer);
--- a/src/Page.js
+++ b/src/Page.js
@@ -37,6 +37,7 @@ const {
   NAVIGATION_FOLDER_PATH,
   CONTENT_WRAPPER_ID,
   FRONT_MATTER_FENCE,
+  FRONT_MATTER_NONE_ATTR,
   PAGE_NAV_ID,
   PAGE_NAV_TITLE_CLASS,
   SITE_NAV_ID,
@@ -496,6 +497,10 @@ class Page {
    */
   insertFooterFile(pageData) {
     const { footer } = this.frontMatter;
+    if (footer === FRONT_MATTER_NONE_ATTR) {
+      return pageData;
+    }
+    
     let footerFile;
     if (footer) {
       footerFile = path.join(FOOTERS_FOLDER_PATH, footer);
--- a/src/Page.js
+++ b/src/Page.js
@@ -683,6 +683,8 @@ class Page {
   collectHeadFiles(baseUrl, hostBaseUrl) {
     const { head } = this.frontMatter;
     if (head === FRONT_MATTER_NONE_ATTR) {
+      this.headFileTopContent = '';
+      this.headFileBottomContent = '';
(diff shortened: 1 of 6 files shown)


src/Page.js
```diff
@@ -594,6 +594,12 @@ Page.prototype.insertHeaderFile = function (pageData) {
 Page.prototype.insertFooterFile = function (pageData) {
   const { footer } = this.frontMatter;
   let footerFile;
+
+  if (footer === 'none') {
+    // avoid adding any footer
```
crphang: comment seems redundant

src/Page.js
```diff
@@ -594,6 +594,12 @@ Page.prototype.insertHeaderFile = function (pageData) {
 Page.prototype.insertFooterFile = function (pageData) {
   const { footer } = this.frontMatter;
   let footerFile;
+
+  if (footer === 'none') {
```
crphang: if 'none' is a keyword, put it as a constant.

src/Page.js
yash-chowdhary: will remove it 👍

src/Page.js
yash-chowdhary: will make the change 👍

damithc:
> Even though it might not be in the scope of this PR, wondering whether this functionality should only be for footer? @damithc

There was a suggestion to extend this to other elements. I have no objections to that. I'll leave the decision to senior devs.

yash-chowdhary:
> There was a suggestion to extend this to other elements. I have no objections to that. I'll leave the decision to senior devs.

yash-chowdhary:
> > Even though it might not be in the scope of this PR, wondering whether this functionality should only be for footer? @damithc
> 
> There was a suggestion to extend this to other elements. I have no objections to that. I'll leave the decision to senior devs.

@yamgent @acjh any suggestions?

yamgent:
> @yamgent @acjh any suggestions?

Sure, feel free to add it for header, site-nav, etc...

src/Page.js
```diff
@@ -669,6 +682,10 @@ class Page {
 
   collectHeadFiles(baseUrl, hostBaseUrl) {
     const { head } = this.frontMatter;
+    if (head === FRONT_MATTER_NONE_ATTR) {
```
crphang: I believe you have to set

```
this.headFileTopContent = '';
this.headFileBottomContent = '';
```

If not this, there could be an issue when changing the file on live reload in this sequence:

1. Add layout
2. Add `head: none`

But the `headFileTopContent` is not removed.

yash-chowdhary: Yes, you're right. Will make the change 👍

yash-chowdhary:
Ready for review

PR #1005, opened 2020-01-31
Remove extra blank lines from page.njk output

ang-zeyu:

• [x] Others

Remove extra blank lines from page.njk output

Removed blank lines from page.njk output using nunjucks's whitespace trimming `{%-`

na

**Testing instructions:**
`npm run test` should pass

Remove extra blank lines from page.njk output

### DIFF
--- a/src/page.njk
+++ b/src/page.njk
@@ -13,32 +13,26 @@
     <link rel="stylesheet" href="{{ asset.glyphicons }}">
     <link rel="stylesheet" href="{{ asset.highlight }}">
     <link rel="stylesheet" href="{{ asset.markbind }}">
-    {%- if asset.pluginLinks %}
-    {%- for link in asset.pluginLinks %}
+    {%- if asset.pluginLinks -%}
+    {%- for link in asset.pluginLinks -%}
         {{ link }}
-    {%- endfor %}
-    {%- endif %}
+    {%- endfor -%}
+    {%- endif -%}
     <link rel="stylesheet" href="{{ asset.layoutStyle }}">
-    {%- if siteNav %}
-        <link rel="stylesheet" href="{{ asset.siteNavCss }}">
-    {%- endif %}
-    {%- if pageNav %}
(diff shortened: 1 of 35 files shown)


yamgent:
Thanks for this. Yes please update the expected files here, otherwise I will still have to do something like 
2ad11f5 again anyway 😛.

ang-zeyu:
> Thanks for this. Yes please update the expected files here, otherwise I will still have to do something like
> [2ad11f5](https://github.com/MarkBind/markbind/commit/2ad11f593dbfd5329c020056d7e04e238d042b53) again anyway 😛.

By updating tests I meant doing what 2ad11f5 does exactly actually 😅 
Just to confirm, since its updated already, this pr wouldn't have any use right?

yamgent:
> By updating tests I meant doing what [2ad11f5](https://github.com/MarkBind/markbind/commit/2ad11f593dbfd5329c020056d7e04e238d042b53) does exactly actually 😅 ( made this pr before it )

I see, I mistakenly assumed that this PR was removing additional blank lines generated by the html output. If you are interested you can rework this PR to focus on that end goal instead.

So for example, given the following html output below, the blank lines at line 22 and 25 can actually be removed, if you replace `{{` with `{{-` for the Nunjucks blocks in `page.njk`.

https://github.com/MarkBind/markbind/blob/2ad11f593dbfd5329c020056d7e04e238d042b53/test/functional/test_site_algolia_plugin/expected/index.html#L16-L29

The blank line at line 20 is also not supposed to be there, some html outputs do not have that blank line.

ang-zeyu:
> I see, I mistakenly assumed that this PR was removing additional blank lines generated by the html output. If you are interested you can rework this PR to focus on that end goal instead.

Misunderstood your intention as well. Thanks for the suggestion!

I've updated the test files and trimmed a few more blank lines.
The commit message is also changed to reflect the purpose better.

ang-zeyu:
Rebased on latest, no changes

Issue #994, opened 2020-01-23
Improve default search UI organization


ang-zeyu:
**Is your request related to a problem?**

Currently, searches for a term that has multiple occurences in a document display like so:

![searchcurrent](https://user-images.githubusercontent.com/3306138/72987979-673dbc80-3e26-11ea-9c57-1c2f3cad8d9c.png)

**Describe the solution you'd like**
This can be improved by organizing search results that belong to the same page, like how algolia does:

![searchdesired](https://user-images.githubusercontent.com/3306138/72988045-850b2180-3e26-11ea-97de-a279f65bef60.png)

We can follow this intuitive design, appending keywords to the right side if any
```
--page name--
  | --heading 1-- | --heading 1--
                    --keyword 1--
                    --keyword 2--
                         ...
  | --heading 2-- | --heading 2--
  | --heading 3-- | --heading 3--
  | --heading 4-- | --heading 4--
```

As an improvement, the typeahead can also only highlight the matched words on the right side, instead of both sides as seen in the image.

**Describe alternatives you've considered**
na

PR #1006, opened 2020-01-31
Restructure generated site data for searchbar ui

ang-zeyu:

Resolves MarkBind/markbind#994
Requires https://github.com/MarkBind/vue-strap/pull/126 ( searchbar ui for vue-strap )

Generated site data for the searchbar needs to be restructured to sift out the keywords from the heading deterministically.

Generated site data:
- `headings` now store only the `id:'heading text...'` pair instead of `id:'heading text... | concatenated keywords ...'`
- keywords for headings are now stored in the `headingKeywords` field
- Removed `concatenateHeadingsAndKeywords` which is no longer needed

New site data:
```js
const siteData = {
    enableSearch: this.siteConfig.enableSearch,
    pages: this.pages.filter(page => page.searchable)
      .map(page => ({
        ...page.frontMatter,
        headings: page.headings,
        headingKeywords: page.keywords,
      })),
  };
```

na

**Testing instructions:**
`npm run test` should pass

Restructure generated site data for searchbar ui

Keywords tied to a heading are appended to the heading with a prepending
'|' character.
This is so that the '|' character would serve as a visual separator
between the heading and the keyword.

However, it is thus not possible to extract a keyword from its heading
deterministically, as a heading can also include the '|' character.

Let's move the keywords to its own field, allowing the searchbar to
extract a heading's keyword from it.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -344,17 +344,6 @@ class Page {
   }
 
   /**
-   * Concatenates keywords in this.keywords to heading text in this.headings
-   */
-  concatenateHeadingsAndKeywords() {
-    Object.keys(this.keywords)
-      .forEach((headingId) => {
-        const keywordString = this.keywords[headingId].join(', ');
-        this.headings[headingId] += ` | ${keywordString}`;
-      });
-  }
-
-  /**
    * Records the dynamic or static included files into this.includedFiles
    * @param dependencies array of maps of the external dependency and where it is included
    */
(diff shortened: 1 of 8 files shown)


ang-zeyu:
Rebased on latest, no conflicts

ang-zeyu:
Rebased on the latest master, no conflicts / changes otherwise

PR #1011, opened 2020-02-01
Allow stylelint to ignore test sites

dvrylc:

• [X] Other, please explain:

This PR allows stylelint to ignore CSS files in the generated test sites. Refer to https://github.com/MarkBind/markbind/pull/1000#discussion_r371920889 for more details.

As @marvinchin explained -
> Not linting generated files seem to make sense to me because some of these generated files might be minified, dependent on external assets etc. and would not satisfy our linting rules (it wouldn’t make sense to lint node_modules for similar reasons!). We should only be applying linting rules to the source files that we are working on.

Add an ignore file for stylelint to exclude CSS files in `test/functional/**`. 

```
**/*.min.css
test/functional/**/*.css
```

–

**Testing instructions:**
Check that files that end with `min.css` and CSS files in `test/functional/**` are not linted by stylelint. 

Allow stylelint to ignore test sites

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,2 @@
+**/*.min.css
+test/functional/**/*.css
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,2 +1,2 @@
 **/*.min.css
-test/functional/**/*.css
+test/functional/**/expected/**/*.css
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
   },
   "scripts": {
     "autolint": "npm run lintfix && npm run csslintfix",
-    "csslint": "./node_modules/.bin/stylelint **/*.css !**/*.min.css",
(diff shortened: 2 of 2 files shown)


.stylelintignore
```diff
@@ -0,0 +1,2 @@
+**/*.min.css
+test/functional/**/*.css
```
marvinchin: Correct me if I'm wrong, but it seems like the contents of the `expected` folder are the only files that are generated programatically (when we expect a diff, we run `markbind build` on the test site and replace `expected` with the output).

Can we further restrict this to `test/functional/**/expected/**/*.css` to make it more specific? This way, the css files that are written by hand for the test sites can still be linted.

dvrylc: Makes sense! Even though we currently don't have any hand-written CSS files for the test sites, this is still a good idea in case we decide to in the future.

Issue #1003, opened 2020-01-29
Update documentation for embedded HTML in dropdown component


le0tan:
**Is your request related to a problem?**

This resolves https://github.com/MarkBind/vue-strap/pull/122#issuecomment-578146475

le0tan:
@yamgent can you kindly check if the changes from https://github.com/MarkBind/vue-strap/pull/122 are merged into MarkBind's current master? When I was trying to write the documentation the new feature doesn't work, but my own build of vue-strap works.

PR #1013, opened 2020-02-03
Update documentation on using raw HTML in dropdown

le0tan:

• [X] Documentation update

Resolves https://github.com/MarkBind/markbind/issues/1003

### DIFF
--- a/docs/userGuide/syntax/dropdowns.mbdf
+++ b/docs/userGuide/syntax/dropdowns.mbdf
@@ -96,6 +96,77 @@
 </span>
 </include>
 
+**You can also use raw HTML as the content of dropdown menus.**
+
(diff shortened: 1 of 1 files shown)


le0tan:
https://github.com/MarkBind/markbind/pull/1029
@yamgent I made the suggested implementation, if it looks OK I'll write documentation to that in the new PR and you may close this PR.

ang-zeyu:
> I am wondering whether we should seek to implement markdown in dropdown, and introduce slots to allow additional customization

Sounds good; 👍

I don't think we should include any extra slots, since the design of dropdowns aren't suited for containing much other than inline markdown. Though, I may've yet to think of a use case for it.

le0tan:
PR closed due to more updated version at #1029

PR #1021, opened 2020-02-04
UserGuide: Document the use of icon-size in boxes

### LINKED ISSUE #986: <box>: support different icon sizes
e.g., `icon-medium` `icon-large` etc.?

yash-chowdhary:

• [x] Documentation update

Resolves #986 

Update documentation for supporting different icon sizes in a box

### DIFF
--- a/docs/userGuide/syntax/boxes.mbdf
+++ b/docs/userGuide/syntax/boxes.mbdf
@@ -160,6 +160,37 @@
 </span>
 </include>
 
+**You can also use icons and resize them accordingly.**
+
+<include src="outputBox.md" boilerplate >
+<span id="code">
+```html
+<box type="warning" icon=":fas-camera:" icon-size="2x" heading="Tip box heading">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+</box>
+<box type="definition" icon-size="2x" heading="Tip box heading" dismissible>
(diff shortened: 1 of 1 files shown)


docs/userGuide/syntax/boxes.mbdf
```diff
@@ -169,6 +200,7 @@ border-left-color | `String` | `null` | Overrides border-color for the left bord
 color | `String` | `null` | Color of the text.
 dismissible | `Boolean` | `false` | Adds a button to close the box to the top right corner.
 icon | `String` | `null` | Inline MarkDown text of the icon displayed on the left.
+icon-size | `String` | `null` | Resizes the icon. Supports integer-scaling of the icon dimensions e.g. `2x`, `3x`, `4x`, etc.
```
openorclose: Would `1x` be a better default?

yash-chowdhary: I am not sure that would be better. If `1x` were documented as the default, it could confuse the user to interpret it as though the class `fa-1x` would be set for the icon by default - which is not the case. 
Setting the default as `1x` in the `TipBox.vue` component would be redundant as well because by default the icon's size is always `1x`.

docs/userGuide/syntax/boxes.mbdf
```diff
@@ -160,6 +160,37 @@
 </span>
 </include>
 
+**You can also use icons and resize them accordingly.**
+
+<include src="outputBox.md" boilerplate >
+<span id="code">
+```html
+<box type="warning" icon=":fas-camera:" icon-size="2x" heading="Tip box heading">
```
yamgent: Can we have the same tipbox type with the same content for all three tipboxes? This would make it easier to see the contrast in size (both visually and in code), and I don't think we need the `heading` or `dismissible` attributes for these examples.

The size showcased should be no-resize, "2x" and "3x".

PR #1026, opened 2020-02-05
Refactor preprocess and url processing functions

ang-zeyu:

• [x] Other, please explain:

Part of #810 

Another step to modularising parser functions, enhancing code maintainability and readability.

- Break down `_preprocess` and move related functions to a separate file, `componentPreprocessor`
- Refactor common ( to `componentPreprocessor` and `parser` ), stateless url processing functions to `utils/urls`
- Enhance in-code documentation for the above along the way
- Several smaller code refactors for the above ( reducing nested ifs, separating long functions when appropriate, etc. )

```
// directory structure from parser.js
- parser
  - utils
    - index
    - urls
  - preprocessors
    - componentPreprocessor
...
```

1. Directory structure of the refactor

2. Of note is that `componentPreprocessor` and `parser` is still rather coupled. ( `parser` is passed as a parameter to the entry point in `componentPreprocessor` )
  - Ideally, only `parser` should depend on `componentPreprocessor`, and not the other way around at all if possible.
  - I initially tried making `componentPreprocessor` a class which would subsume required properties from the main `parser`. This was not feasible at the moment however, as many other `parser` functions have 'side effects' on `parser`'s instance variables. It would be counterproductive to add code to synchronize the variables between the two.
  - Hence, we can proceed with a smaller refactor first, and gradually aim to remove this coupling if possible.

**Testing instructions:**
`npm run test` should pass

Refactor preprocess and url processing functions

The parser class houses a large number of functions.
This can be daunting for newer developers to the project, and decreases
maintainability of the code.

Let’s start by modularising the preprocess and url processing functions
into separate files.
Let’s also enhance some in-code documentation for these functions.
Some of these functions can also be abstracted into smaller units of
functionality, increasing the maintainability of the code.

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -4,10 +4,9 @@ const htmlparser = require('htmlparser2'); require('./patches/htmlparser2');
 const nunjucks = require('nunjucks');
 const path = require('path');
 const Promise = require('bluebird');
-const url = require('url');
-const pathIsInside = require('path-is-inside');
 const slugify = require('@sindresorhus/slugify');
 const componentParser = require('./parsers/componentParser');
+const componentPreprocessor = require('./preprocessors/componentPreprocessor');
 
 const _ = {};
 _.clone = require('lodash/clone');
@@ -15,11 +14,10 @@ _.cloneDeep = require('lodash/cloneDeep');
 _.hasIn = require('lodash/hasIn');
 _.isArray = require('lodash/isArray');
 _.isEmpty = require('lodash/isEmpty');
-_.pick = require('lodash/pick');
 
-const CyclicReferenceError = require('./handlers/cyclicReferenceError.js');
 const md = require('./lib/markdown-it');
 const utils = require('./utils');
+const urlUtils = require('./utils/urls');
 
 cheerio.prototype.options.xmlMode = true; // Enable xml mode for self-closing tag
 cheerio.prototype.options.decodeEntities = false; // Don't escape HTML entities
@@ -27,7 +25,6 @@ cheerio.prototype.options.decodeEntities = false; // Don't escape HTML entities
 const {
   ATTRIB_INCLUDE_PATH,
   ATTRIB_CWF,
-  BOILERPLATE_FOLDER_NAME,
   IMPORTED_VARIABLE_PREFIX,
 } = require('./constants');
 
@@ -129,21 +126,6 @@ class Parser {
     return _.clone(this.missingIncludeSrc);
   }
 
-  static _preprocessThumbnails(element) {
-    const isImage = _.hasIn(element.attribs, 'src') && element.attribs.src !== '';
-    if (isImage) {
-      return element;
-    }
-    const text = _.hasIn(element.attribs, 'text') ? element.attribs.text : '';
-    if (text === '') {
-      return element;
-    }
-    const renderedText = md.renderInline(text);
-    // eslint-disable-next-line no-param-reassign
-    element.children = cheerio.parseHTML(renderedText);
-    return element;
-  }
-
   _renderIncludeFile(filePath, element, context, config, asIfAt = filePath) {
     try {
       this._fileCache[filePath] = this._fileCache[filePath]
@@ -153,20 +135,19 @@ class Parser {
       const missingReferenceErrorMessage = `Missing reference in: ${element.attribs[ATTRIB_CWF]}`;
       e.message += `\n${missingReferenceErrorMessage}`;
       this._onError(e);
-      return Parser.createErrorNode(element, e);
+      return utils.createErrorNode(element, e);
     }
     const fileContent = this._fileCache[filePath]; // cache the file contents to save some I/O
-    const { parent, relative } = Parser.calculateNewBaseUrls(asIfAt, config.rootPath, config.baseUrlMap);
+    const { parent, relative } = urlUtils.calculateNewBaseUrls(asIfAt, config.rootPath, config.baseUrlMap);
     const userDefinedVariables = config.userDefinedVariablesMap[path.resolve(parent, relative)];
     // Extract included variables from the PARENT file
     const includeVariables = Parser.extractIncludeVariables(element, context.variables);
     // Extract page variables from the CHILD file
     const pageVariables
       = this.extractPageVariables(asIfAt, fileContent, userDefinedVariables, includeVariables);
-    const content
-      = nunjucks.renderString(fileContent,
-                              { ...pageVariables, ...includeVariables, ...userDefinedVariables },
-                              { path: filePath });
+    const content = nunjucks.renderString(fileContent,
+                                          { ...pageVariables, ...includeVariables, ...userDefinedVariables },
+                                          { path: filePath });
     const childContext = _.cloneDeep(context);
     childContext.cwf = asIfAt;
     childContext.variables = includeVariables;
@@ -197,10 +178,7 @@ class Parser {
       // Render inner file content
       const { content: renderedContent, childContext, userDefinedVariables }
         = this._renderIncludeFile(filePath, element, context, config);
-      if (!Parser.PROCESSED_INNER_VARIABLES.has(filePath)) {
-        Parser.PROCESSED_INNER_VARIABLES.add(filePath);
-        this._extractInnerVariables(renderedContent, childContext, config);
-      }
+      this.extractInnerVariablesIfNotProcessed(renderedContent, childContext, config, filePath);
       const innerVariables = this.getImportedVariableMap(filePath);
       Parser.VARIABLE_LOOKUP.get(filePath).forEach((value, variableName, map) => {
         map.set(variableName, nunjucks.renderString(value, { ...userDefinedVariables, ...innerVariables }));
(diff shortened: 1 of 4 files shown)


ang-zeyu:
Ready for review

Some quirks I noticed which I'm not so sure about:

1. Boilerplate include sources are added to the parser's `boilerplateIncludeSrc` array. However, it is also added to `dynamicIncludeSrc` afterward, and is functionally the same for the purposes of live reloading. Should we remove `boilerplateIncludeSrc`? ( left as is for now )

2. `_.hasIn` is used quite a few times to check whether some attribute exists in an element. I've changed these to `_.has`, since attributes are direct properties of `element.attribs`, we do not want to check for indirect attributes.

3. Whether preprocessing logic for thumbnails ( moved to `_preprocessThumbnail` ) should be moved to `componentParser` instead ( left as is for now )

4. For `include` tags, there is some specific processing for `context.mode`. However, the `_preprocess` was never called with anything other than `context.mode = 'include'`. Should we remove this? ( left as is for now )

5. We only call `_rebaseReferenceForStaticIncludes` for `<include>`'s ( left as is )
  5.1. which's includeSrc is an md file,
  5.2. a `src` that has some hash `#...`.
  Is this intentional? ( though, changing it include `5.2` or both `5.2` and `5.1` and running `updatetest` didn't change any output test files )

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -0,0 +1,394 @@
...
+function _preprocessVariables() {
+  return utils.createEmptyNode();
+}
+
+function _preprocessBody(node) {
+  // eslint-disable-next-line no-console
+  console.warn(`<body> tag found in ${node.attribs[ATTRIB_CWF]}. This may cause formatting errors.`);
+}
+
+function preProcessComponent(node, context, config, parser) {
```
ang-zeyu: Entry point to the original `_preprocess` function, the respective switch cases are the respective control flows of `_preprocess` for different types of elements

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -0,0 +1,394 @@
...
+  if (isIncludeSrcMd) {
+    if (context.mode === 'include') {
+      actualContent = isInline ? actualContent : `\n\n${actualContent}\n`;
+    } else {
+      // TODO do we need this branch?
+      actualContent = md.render(actualContent);
+    }
+
+    // TODO clarify why this is only done if there is a hash, or fix it if it is unintended
+    if (hash) _rebaseReferenceForStaticIncludes(actualContent, element, config);
```
ang-zeyu: Is this intentional? ( point 5 )

src/lib/markbind/src/parser.js
```diff
@@ -197,193 +178,19 @@ class Parser {
...
-      filePath = isUrl
-        ? element.attribs.src
-        : path.resolve(path.dirname(context.cwf), decodeURIComponent(includeSrc.path));
-      actualFilePath = filePath;
-      const isBoilerplate = _.hasIn(element.attribs, 'boilerplate');
-      if (isBoilerplate) {
-        element.attribs.boilerplate = element.attribs.boilerplate || path.basename(filePath);
-        actualFilePath
-          = Parser.calculateBoilerplateFilePath(element.attribs.boilerplate, filePath, config);
-        this.boilerplateIncludeSrc.push({ from: context.cwf, to: actualFilePath });
```
ang-zeyu: Is `boilerplateIncludeSrc` necessary?
For both panels and includes we eventually still push either `dynamicIncludeSrc` or `staticIncludeSrc`

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -0,0 +1,394 @@
...
+  const element = node;
+
+  // We do this since element.attribs is undefined if it does not exist
+  element.attribs = element.attribs || {};
+
+  element.attribs[ATTRIB_CWF] = path.resolve(context.cwf);
+}
+
+// TODO move this to componentParser
+function _preProcessThumbnail(node) {
```
ang-zeyu: Should this be moved?

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -0,0 +1,394 @@
...
+  } else {
+    actualContent = (fileContent && isTrim) ? fileContent.trim() : fileContent;
+  }
+
+  if (isIncludeSrcMd) {
+    if (context.mode === 'include') {
+      actualContent = isInline ? actualContent : `\n\n${actualContent}\n`;
+    } else {
+      // TODO do we need this branch?
+      actualContent = md.render(actualContent);
```
ang-zeyu: A quick usage find for `_preprocess` reveals that it is never called with anything other than `context.mode = 'include'`

src/lib/markbind/src/parser.js
```diff
@@ -197,193 +178,19 @@ class Parser {
...
-      return Parser.createEmptyNode();
-    } else {
-      if (element.name === 'body') {
-        // eslint-disable-next-line no-console
-        console.warn(`<body> tag found in ${element.attribs[ATTRIB_CWF]}. This may cause formatting errors.`);
-      }
-      if (element.children && element.children.length > 0) {
-        element.children = element.children.map(e => self._preprocess(e, context, config));
-      }
+  extractInnerVariablesIfNotProcessed(content, childContext, config, filePathToExtract) {
```
ang-zeyu: To allow `componentPreprocessor` and `_extractInnerVariables` to share this common code

src/lib/markbind/src/parser.js
```diff
@@ -685,15 +429,71 @@ class Parser {
...
+        const boilerplateFilePath = urlUtils.calculateBoilerplateFilePath(path.basename(file), file, config);
         if (utils.fileExists(boilerplateFilePath)) {
           actualFilePath = boilerplateFilePath;
         }
       }
 
-      this.preprocess(actualFilePath, pageData, context, config)
-        .then(resolve)
-        .catch(reject);
+      const currentContext = context;
```
ang-zeyu: This chunk is moved from the `preprocess` ( not `_preprocess` ) function, which was added in the expressive layouts pr,
to remove naming ambiguity.

A future PR may look into whether expressive layouts can be introduced in similar fashion to the page `header`, `sitenav`, etc., as `includeData` ( the function this code is under ) is very similar to `includeFile`. ( to reduce code duplication )

src/lib/markbind/src/parser.js
```diff
@@ -804,34 +604,32 @@ class Parser {
     // rebase current element
     if (element.attribs[ATTRIB_INCLUDE_PATH]) {
       const filePath = element.attribs[ATTRIB_INCLUDE_PATH];
-      let newBase = Parser.calculateNewBaseUrls(filePath, this.rootPath, this.baseUrlMap);
-      if (newBase) {
-        const { relative, parent } = newBase;
+      let newBaseUrl = urlUtils.calculateNewBaseUrls(filePath, this.rootPath, this.baseUrlMap);
```
ang-zeyu: The only irrelevant refactor to remove the nested `ifs`, and slightly better variable naming; Let me know if this should be removed!

acjh:
> changing it ... and running `updatetest` didn't change any output test files

Instead of relying on non-exhaustive functional tests, for extensive refactoring, let's check that there is no diff to the generated `_site` of [nus-cs2103/website-base](https://github.com/nus-cs2103/website-base).

ang-zeyu:
> Instead of relying on non-exhaustive functional tests, for extensive refactoring, let's check that there is no diff to the generated `_site` of [nus-cs2103/website-base](https://github.com/nus-cs2103/website-base).

Thanks for the suggestion! Will take note of it for upcoming refactors.

I've removed the specific processing for `context.mode`, after confirming the output did not change between the latest master and my branch for the 2103 site. I added a `console.log` in the branch previously as well, and confirmed no statements were printed.

Interestingly, `_rebaseReferenceForStaticIncludes` with any combination of the conditions ( including removing it completely ) did not change the output.
I've left it in for now, it may be better to investigate this in more detail in a seperate PR.

ang-zeyu:
Rebased on latest, no changes

src/lib/markbind/src/preprocessors/componentPreprocessor.js
yamgent: Whether it should be moved or not will depend on how specialised componentProcessor will eventually get. I noted that there's some coupling between this and the parser that still needs to be resolved in future PRs, so I believe the answer will naturally come by as the refactoring process goes.

Also noted about the weird quirks that you have found. Refactoring by itself is usually quite a tough task by itself; verifying additional changes on top of it is going to make it even more tough, so always welcome to settle those problems in separate PRs (and good to open these in the issue tracker too).

Since my time is now quite restricted as compared to the past, I am no longer able to manually verify that everything works. So these green ticks approval that I give from now on would basically be just a glance through your code to see that there's no major source of issues in terms of design and implementation. You guys might want to test out each other's code from time to time to see if there's any problems.

With all that said, will be merging this. :+1:

yamgent:
> Should this be moved?

Whether it should be moved or not will depend on how specialised componentProcessor will eventually get. I noted that there's some coupling between this and the parser that still needs to be resolved in future PRs, so I believe the answer will naturally come by as the refactoring process goes.

Also noted about the weird quirks that you have found. Refactoring by itself is usually quite a tough task by itself; verifying additional changes on top of it is going to make it even more tough, so always welcome to settle those problems in separate PRs (and good to open these in the issue tracker too).

Since my time is now quite restricted as compared to the past, I am no longer able to manually verify that everything works. So these green ticks approval that I give from now on would basically be just a glance through your code to see that there's no major source of issues in terms of design and implementation. You guys might want to test out each other's code from time to time to see if there's any problems.

With all that said, will be merging this. :+1:

Issue #1032, opened 2020-02-08
Plugins not used on included content


openorclose:
**What did you do? Please include the actual source code causing the issue.**

```html
//index.md
<panel src="inc.md"></panel>
<panel><span heading><b>header</b></span> Content of panel</panel>



//inc.md
<panel><span heading><b>header</b></span> Content of panel</panel>
```

**What did you expect to happen?**

Both the contentful panels should look the same.

**What actually happened? Please include the actual, raw output.**
![image](https://user-images.githubusercontent.com/3646725/74080508-a05d6a00-4a7f-11ea-82be-d043b49173e7.png)

This happens because plugins are not rendered on included files.

Should this be documented, or is this a bug?

ang-zeyu:
To be more precise, it's not rendered on dynamic includes ( `<panel>` or `<include dynamic>` ),
but for static `<include>`s it is

damithc:
> Should this be documented, or is this a bug?

I don't think that is intentional. Dynamic panels should look the same as other panels except the content will not be loaded until they are expanded.

PR #1033, opened 2020-02-08
Migrate to bootstrap-vue popovers

### LINKED ISSUE #814: Migrate all MarkBind components from vue-strap to bootstrap-vue
The end goal here is to remove our dependency on vue-strap altogether. This is a major investment, so the team will have to make the choice of whether (or better yet, _when_) to prioritize this over feature work.

Related discussion:
- #411 illustrates the inconveniences of working with vue-strap
- #245 discusses the incentives of such a migration and what work it will involve

### LINKED ISSUE #873: Popovers and Tooltips: migrate to bootstrap-vue
https://bootstrap-vue.js.org/docs/components/popover/
https://bootstrap-vue.js.org/docs/components/tooltip/

These components already exist and work better than the ones we currently have.

openorclose:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Fixes #873 
Fixes #814

**Updated PR, modals, tooltips, and popovers are all working**

## Some breaking changes:

Contextmenu (right click) no longer a trigger

bootstrap-vue modals don't support the width attribute
But their modals have a whole lot of other options, see them here:
https://bootstrap-vue.js.org/docs/components/modal/

Popovers now don't have the zoom effect. I tried to reimplement the zoom effect, but was only successful for modals. Will try to add back in later.

I have updated the docs for these.

----

We are still reliant on vue-strap for popovers, modals, and tooltips. This aims to complete the migration to bootstrap-vue, by using their versions of tooltips, popovers, and modals.

Updated our bootstrap-vue js and css to the latest version. This gives us better versions of popovers and tooltips.

bootstrap-vue's trigger mechanism is very different from ours:

Each popover must target a single element on page load, so our "many triggers for one popover" implementation does not work.

So, we take all `<trigger>` elements and turn it into a <span> with a [v-b-popover directive](https://bootstrap-vue.js.org/docs/directives/popover), and pass in functions that allow us to return a HTML string of the desired popover.
Taken from the docs:
```md
More about <trigger for="pop:trigger_id">trigger</trigger>.
<popover id="pop:trigger_id" content="This popover is triggered by a trigger"></popover>
<br>
This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.

```
Old:
image

New:
image

Functionality of the popovers.

Note that modals and tooltips are broken for now, but the solution for them should be similar
**Testing instructions:**


openorclose:
https://deploy-preview-1033--markbind-master.netlify.com/userguide/usingcomponents#popovers

All popovers trigger only on click for now for better testing (can open multiple at once, inspect element more easily).

ang-zeyu:
> Whether duplicating the popovers is an acceptable solution.

Since the typical use case for a popover is to include really small tidbits of info, I don't think the cost for this is significant

----------

I like the idea of moving the processing work to /markbind, but why a plugin though?
This feels like something that should be inherent to markbind.

Could we try integrating this into `componentParser` instead?
It would be in line with where we do custom processing for specific components, and reduce a round of parsing the entire `content` with cheerio.

The tricky part would be that `componentParser` does not have access to the entire dom, which is needed for `$([target='${popoverId}'])`. Perhaps during `_preprocess` you could whitelist popovers and add their content to a parser instance variable map for access later.

That said, that's just my two cents, but at the very least this should be a stage in the `page.generate()` chain, not a plugin I think.

----------

Another possible issue is the way `popoverId` is done. There will likely be duplicate popover ids between a page and its external dependencies. This will not cause issues when the page is initially loaded, but may after the dynamic dependency has loaded. ( assuming #1032 is implemented ) A simple fix may be to prepend something unique to the page ( file path hash? ) before the id though

openorclose:
Thanks for the review!

> > Whether duplicating the popovers is an acceptable solution.
> 
> Since the typical use case for a popover is to include really small tidbits of info, I don't think the cost for this is significant
> 
> I like the idea of moving the processing work to /markbind, but why a plugin though?
> This feels like something that should be inherent to markbind.
> 
> Could we try integrating this into `componentParser` instead?
> It would be in line with where we do custom processing for specific components, and reduce a round of parsing the entire `content` with cheerio.
> 
> The tricky part would be that `componentParser` does not have access to the entire dom, which is needed for `$([target='${popoverId}'])`. Perhaps during `_preprocess` you could whitelist popovers and add their content to a parser instance variable map for access later.

This also wouldn't work. Consider a trigger, followed by the actual popover element. When the trigger component is parsed, the popover is not yet parsed. Later on, there's no way to "go back" and edit it back in. If we were to go that route then we would need two maps, one storing all the triggers, and another storing all the popovers, and each time we parse either we need to check if they have been referenced by another popover/trigger. Maybe I'll try this.

> That said, that's just my two cents, but at the very least this should be a stage in the `page.generate()` chain, not a plugin I think.

Yes, the main reason for it being a plugin is so that it has access to the entire DOM.

Yeah and noted regarding the dependent files too, could probably have a counter counting the number of included files too.

ang-zeyu:
> This also wouldn't work. Consider a trigger, followed by the actual popover element. When the trigger component is parsed, the popover is not yet parsed. Later on, there's no way to "go back" and edit it back in. If we were to go that route then we would need two maps, one storing all the triggers, and another storing all the popovers, and each time we parse either we need to check if they have been referenced by another popover/trigger. Maybe I'll try this.

Hmm, there shouldn't be a need to 'go back'.
To clarify what I meant :

Pseudo code:
```
// preprocess
if (node.name === 'popover')
  this.popoverContentMap.set(popover.id, popover's content and title attribute)

// componentParser
case 'trigger':
  search popoverContentMap for popover indicated by our for attribute
  rename element as popover
  set content and title attribute to equal the content and title retrieved

// front end popover component
use the v-b-popover directive, and set the content and title attribute accordingly

```

openorclose:
> > This also wouldn't work. Consider a trigger, followed by the actual popover element. When the trigger component is parsed, the popover is not yet parsed. Later on, there's no way to "go back" and edit it back in. If we were to go that route then we would need two maps, one storing all the triggers, and another storing all the popovers, and each time we parse either we need to check if they have been referenced by another popover/trigger. Maybe I'll try this.
> 
> Hmm, there shouldn't be a need to 'go back'.
> To clarify what I meant :
> 
> Pseudo code:
> 
> ```
> // preprocess
> if (node.name === 'popover')
>   this.popoverContentMap.set(popover.id, popover's content and title attribute)
> 
> // componentParser
> case 'trigger':
>   search popoverContentMap for popover indicated by our for attribute
>   rename element as popover
>   set content and title attribute to equal the content and title retrieved
> 
> // front end popover component
> use the v-b-popover directive, and set the content and title attribute accordingly
> ```

Yeah I got that, but the problem is in `case trigger`. The `popoverContentMap` might not contain our needed element yet, because a trigger can appear after a popover, but a trigger can also appear before a popover.

Like
```
<trigger for=pop> click me</trigger>
<popover id=pop content = "i will appear">/popover>
```
or
```
<trigger for=pop> click me</trigger>
<popover id=pop content = "i will appear">/popover>
```

I believe it still can be a solution if we use two maps, a `triggerMap` and a `popoverContentMap`. Not sure if that will be too messy.

ang-zeyu:
> Yeah I got that, but the problem is in `case trigger`. The `popoverContentMap` might not contain our needed element yet, 

Hmm, isn't `_preprocess` ( `includeFile` ) quite a few stages before `_parse` ( `renderFile` )?
This shouldn't occur

openorclose:
> > Yeah I got that, but the problem is in `case trigger`. The `popoverContentMap` might not contain our needed element yet,
> 
> Hmm, isn't `_preprocess` ( `includeFile` ) quite a few stages before `_parse` ( `renderFile` )?
> This shouldn't occur

Ah yes, you're right. Thanks!

openorclose:
```
// preprocess
if (node.name === 'popover')
  this.popoverContentMap.set(popover.id, popover's content and title attribute)

// componentParser
case 'trigger':
  search popoverContentMap for popover indicated by our for attribute
  rename element as popover
  set content and title attribute to equal the content and title retrieved

// front end popover component
use the v-b-popover directive, and set the content and title attribute accordingly
```

I realised this doesn't work well also:

The popover's content and title might contain other stuff that need to be processed if slots are used,

like 
```html
<popover>
<div slot="content><include src="file.md"/></div> Hover me </popover>
```

but then the map wouldn't have access to the processed popover.


I can add the not-yet-parsed popover into a map during the preprocess stage no problem, but that will not give access to the actual parsed popover.

ang-zeyu:
> I can add the not-yet-parsed popover into a map during the preprocess stage no problem, but that will not give access to the actual parsed popover.

True. In that case, I can't think of a better implementation than this at the moment.

An alternative would be to do this on the front end, by creating b-popovers as many times as there are triggers. But this is likely to be more heavy on the client, which we should try to avoid.

alyip98:
**Plugin**
I like the plugin implementation, as it consolidates all popover related functionality rather than sprinkling them all over the `then` train. Another benefit of doing everything in `postRender` is we avoid tricky situations with includes or dynamic panels.

**Duplication**
I think that html duplication for popovers and tooltips isn't an issue as they are meant for light content. Modals are often heavier, but even then a typical page would not have too many trigger-modal pairs.

openorclose:
> **Plugin**
> I like the plugin implementation, as it consolidates all popover related functionality rather than sprinkling them all over the `then` train. Another benefit of doing everything in `postRender` is we avoid tricky situations with includes or dynamic panels.

The plugin _works_, but we'd run into issues along the way: the footnotes plugin uses popovers, so we'd want the plugins to render footnotes into popovers, and then transform all popovers into bootstrap-vue components. So we need to control the order in which plugins are executed. Is there a way to do this now?

ang-zeyu:
> The plugin _works_, but we'd run into issues along the way: the footnotes plugin uses popovers, so we'd want the plugins to render footnotes into popovers, and then transform all popovers into bootstrap-vue components. So we need to control the order in which plugins are executed. Is there a way to do this now?

> but at the very least this should be a stage in the page.generate() chain, not a plugin I think.

We could insert this a seperate stage before `postRender`, something like `processBvComponents`, without much changes needed to this implementation, if any.
This consolidates all functionality as well, and prevents the user from misconfiguring the plugin config to turn this off, which we likely don't want since this is a core feature.

We could also use this for same-page id conflict resolution as well, e.g. a `bvPopoversIdMap` page variable, which avoids having to append something unique to the page before the id, reducing clutter in the output. ( and avoid passing `bvPopoversIdMap` into a plugin )

ang-zeyu:
On another note though, was using `popper.js` ( the library bootstrap-vue uses for popover positioning ) considered?

It seems like bootstrap-vue popovers & tooltips aren't suited to our use case ( multiple triggers for one popover ).
I've looked into popper.js briefly, and it seems to support creating multiple 'popper instances' with just one content element.
It's also pretty lightweight ( ~3kb ), if we were to include it as a standalone library over bootstrap-vue. ( ~200kb )

That said if the long term goal is to exploit the large amount of features bootstrap-vue has, and not just for popovers / tooltips, then I don't think there's an issue here.

alyip98:
> So we need to control the order in which plugins are executed. Is there a way to do this now?

I'm not sure, does the current implementation respect load order? If not we could make a separate issue and PR to improve plugins, possibly also adding more hooks throughout the `generate()` chain.

openorclose:
Okay after some days I realised a better solution: v-b-popover directives allow us to pass in an object, so that we can provide functions that return strings for content and title:
https://bootstrap-vue.js.org/docs/directives/popover#directive-syntax-and-usage

image

This allows us to have the best of both words: no duplication of any popovers and the subjectively better positioning offered by bootstrap-vue.

**One downside is that we won't have support for the `contextmenu` trigger (right click) for now**, will that be a problem @damithc? 

View the popovers here:
https://deploy-preview-1033--markbind-master.netlify.com/userguide/usingcomponents#popovers

This solution is a lot simpler than my previous solution, and does not require the iffy generation of ids for popovers at all.

damithc:
> **One downside is that we won't have support for the `contextmenu` trigger (right click) for now**, will that be a problem @damithc?

I think it's an acceptable compromise.

openorclose:
Triggers are done!

https://5e456a9f4a4d7e000d455324--markbind-master.netlify.com/userguide/usingcomponents#tooltips

openorclose:
Ready for review!

Some breaking changes:

Contextmenu (right click) no longer a trigger

bootstrap-vue modals don't support the width attribute
But their modals have a whole lot of other options, see them here:
https://bootstrap-vue.js.org/docs/components/modal/

Popovers now don't have the `zoom` effect. I tried to reimplement the zoom effect, but was only successful for modals. Will try to add back in later.

ang-zeyu:
Took a brief look at the code, looks good 👍.

Functionality wise, noticed some sort of flickering when the placement is `top/bottom`, where the 'arrow' isn't completely flush with the popover / tooltip's content. Could the flickering and positioning be fixed?

flickering

Is it possible to use an empty `modal-footer` slot to override the ok and cancel buttons, while preserving the user's ability to use a `modal-footer` slot as well? ( i'm fine with it being there though, although I think its use case is more for forms contained in modals which markbind dosen't have )

modal

openorclose:
I've removed the cancel button (just added the `ok-only` attribute to b-modal).

Seems like bootstrap-vue has fixed the positioning thing!

I updated bootstrap-vue.js and bootstrap-vue.css to the latest versions and they work better. They even have support for the hovering over of popovers, so that you can select text inside them.

asset/css/bootstrap-vue.min.css
```diff
@@ -1 +1,3 @@
...
\ No newline at end of file
+/*!
```
marvinchin: I think it'd be preferable to update the bootstrap-vue assets in a separate PR

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -42,6 +42,26 @@ function _parseAttributeWithoutOverride(element, attribute, isInline, slotName =
   delete el.attribs[attribute];
 }
 
+/**
+ * Takes an element, looks for direct elements with slots and transforms to avoid Vue parsing.
+ * @param element Element to transform
+ */
+function _transformSlottedComponents(element) {
+  element.children.forEach((c) => {
+    const child = c;
```
marvinchin: Might be missing something here, but could we just rename the `c` parameter to `child` directly instead? 😅

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,24 +139,61 @@ function _assignPanelId(element) {
   }
 }
 
+/*
+ * Triggers
+ */
+
+function _parseTrigger(element) {
+  const el = element;
```
marvinchin: Same for `el` and `element`

marvinchin: Also, in the above we seem to be favouring explicitness (`child` instead of `c`), but here we seem to be favouring terseness. Is there a reason for this change?

test/functional/test_site_algolia_plugin/expected/markbind/js/setup.js
```diff
@@ -171,6 +171,35 @@ function setupWithSearch() {
...
+function makeHtmlGetterFor(componentType, attribute) {
+  return (element) => {
+    const contentWrapper = document.getElementById(element.attributes.for.value);
+    return contentWrapper.dataset.mbComponentType === componentType
+      ? makeInnerGetterFor(attribute)(contentWrapper) : '';
+  };
+}
+
+
+/* eslint-disable no-unused-vars */
```
marvinchin: Let's leave some comments explaining why these unused variables are needed.

test/functional/test_site_algolia_plugin/expected/markbind/js/setup.js
```diff
@@ -171,6 +171,35 @@ function setupWithSearch() {
...
+}
+
+function makeHtmlGetterFor(componentType, attribute) {
+  return (element) => {
+    const contentWrapper = document.getElementById(element.attributes.for.value);
+    return contentWrapper.dataset.mbComponentType === componentType
+      ? makeInnerGetterFor(attribute)(contentWrapper) : '';
+  };
+}
+
```
marvinchin: There seems to be an additional newline here.

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,24 +139,61 @@ function _assignPanelId(element) {
   }
 }
 
+/*
+ * Triggers
+ */
+
+function _parseTrigger(element) {
```
marvinchin: I think it would also be great if we could leave come comments to give future developers some context as to why we are doing these transformations.

src/lib/markbind/src/parsers/componentParser.js
openorclose: I'll change the child function then, to keep it more consistent.

src/lib/markbind/src/parsers/componentParser.js
openorclose: Yeah, It's because of our eslint rule `no-param-reassign`

asset/css/bootstrap-vue.min.css
openorclose: Ok! I'll revert the change.

openorclose:
> The approach seems fine, but I think it would be great if we can leave more context for people who might come across this in the future and wonder why we are doing all of these transformations. Just a couple of questions about variables!
> 
> Also, are the issues you mentioned in the PR regarding modals and tooltips resolved? I gave it a try on the netlify site and it seems to work.

Thanks for the review! Yeah modals and tooltips all work now, updated the PR.

I'll add in more documentation on the transformations do too

marvinchin:
@openorclose there seems to be an issue with the changes that causes code blocks to line wrap when mousing over it:

code-line-wrap

The markbind.org version seems fine, so I suspect it might be related to some changes here?

openorclose:
> ~@openorclose there seems to be an issue with the changes that causes code blocks to line wrap when mousing over it:~
> 
> code-line-wrap
> 
> ~The markbind.org version seems fine, so I suspect it might be related to some changes here?~
> 
> I think this is a bug on master. Other netlify builds that have been rebased on top of master has the same issue.

It's a new feature in #991, when the user hovers over too-long code it will automatically wrap

ang-zeyu:
> It's a new feature in #991, when the user hovers over too-long code it will automatically wrap

Hmm is this suppose to occur? ( its not tied to this PR though, just using your netlify preview for convenience )
https://deploy-preview-1033--markbind-master.netlify.com/userguide/makingthesitesearchable#search-bars
( Hover over the first code block and observe the width of the main container )

Issue #645, opened 2019-01-30
Include a file name for code blocks


Xenonym:
**Tell us about your environment**

* **MarkBind Version:** 1.17.1

**Rationale:** Developer documentation usually includes not only code excerpts, but also sample code files. It can be a nice visual reminder for the reader if we include the file name for a given sample code block, especially if that code file needs to have a fixed name (eg. configuration files like `.travis.yml`).

**Example:** From [Travis CI documentation](https://docs.travis-ci.com/user/tutorial/#to-get-started-with-travis-ci):

![code_filename](https://user-images.githubusercontent.com/1782590/51967655-d6e6c900-24aa-11e9-9ab0-d645f10e5d18.png)

damithc:
> How about in the lower right corner? Having the file name in the same position with the text might mess up copy-pasting though.

Perhaps it is OK to do it as in the screenshot, as the filename is going to be optional and will not be present in most cases.

> Generating a download via JS is technically possible, see [FileSaver.js](https://github.com/eligrey/FileSaver.js). Unlike #629, this should definitely be optional, since there probably isn't any value in downloading small code excerpts.

Yes, that should be optional too, only for cases where the user is meant to save the file locally and do something with it. In fact, probably should be a very low priority feature, if we are going to do it at all.

nbriannl:
Hi, I'm interested in picking this up. Can I clarify if by 'code block' you mean https://markbind.org/userGuide/formattingContents.html#code ? This also applies to #646

damithc:
> Hi, I'm interested in picking this up. Can I clarify if by 'code block' you mean https://markbind.org/userGuide/formattingContents.html#code ? This also applies to #646

Yes for both.

damithc:
Also, should we generalize this beyond filenames? There might be other information the author might want to show e.g., `before` and `after`.

damithc:
If we make this a general `heading` feature (similar to box headings), we might not even need #646 
![image](https://user-images.githubusercontent.com/1673303/73627258-dd43ee00-4686-11ea-93ad-ee7f948b3e29.png)

While it is a bit of extra work, authors can show the code language by simply adding a heading.

yamgent:
> If we make this a general `heading` feature (similar to box headings), we might not even need #646

Good idea, @nbriannl what do you think? Since with #646 authors must specify whether to have the indicator enabled or not, I think we might as well just let the authors be able to customize the content of the heading entirely.

nbriannl:
[...]

PR #1034, opened 2020-02-09
Add heading and line highlighting to code blocks 

### LINKED ISSUE #646: Indicate the language of a code block

nbriannl:
~**Note: Will take out of WIP when #991 is approved**~

~**Note: This branch is a rebase over #991 in order to work on top of his changes. Will rebase over master once #991 is merged**~

Code headings fixes #646 #645
Partially fixes the requirements of line highlighting in #473

For MarkBind to be a developer documentation tool, can be helpful to indicate the language of a code block in the block itself. To make the usage more general, any heading test can be optionally added to a code block, be it a file name & extension, the language used, or a generic heading. 

Also, it would be good to be able to highlight specific lines of the code block, so to allow the author to emphasis certain lines.

Because the following are markdown-it-attrs, you can use the original fenced code syntax if you do not want to use these feature. 

We overwrite the markdown-it rule for fenced code blocks, to add a heading (which is a div) if a heading is specified as an attribute. 

As for the highlight-lines attribute, firstly, the values are parsed into ranges. Secondly, we add upon the implementation of the `map` function in #991 that maps each lines into spans. What this addition is is that if the current line is within any of the ranges that is intended to be highlighted, it is then given the CSS class "highlighted". This also means that it is ok to give an invalid ranges, i.e. a line number that is out of bounds of the number of lines in the code, because the lines are checked against the ranges and not the other way around.   

CSS changes were also made to enable styling and position of the heading div with respect to the div containing the code.

The headings are meant to be additional (and possibly unimportant) info. That's why they are on the right rather than left, away from the readers main path of reading.

Hence, the following design, here is how it looks like

image

image

Add the following syntax to any test site and serve

````
```xml
<foo>
  <bar type="name">goo</bar>
</foo>
```
```xml {heading="html"}
<foo>
  <bar type="name">goo</bar>
</foo>
```
```nginx {heading="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."}
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
}
```
```nginx {heading="This is like a box heading" highlight-lines="1,4-7,8,11-55"}
user  www www;
worker_processes  2;
pid /var/run/nginx.pid;
error_log  /var/log/nginx.error_log  debug | info | notice | warn | error | crit;

events {
    connections   2000;
    use kqueue | rtsig | epoll | /dev/poll | select | poll;
}

server { # simple reverse-proxy
    listen       80;
    server_name  domain2.com www.domain2.com;
    access_log   logs/domain2.access.log  main;
} 
# It is ok to specify a range greater than the lines of the code block
```
````

Please comment if the styling of the component can be improved.

Add heading and line highlighting to code blocks

### CODE REVIEW

**Comment by alyip98** [comment]
I think it looks weird to have the heading box on the right, its usually on the left.

I personally prefer the look of the previous implementation more than the heading box


And we could add a bit of CSS to solve the previously mentioned issues
```css
.code-block:hover > .code-block-lang {
    display: none;
}
```

damithc:
> I think it looks weird to have the heading box on the right, its usually on the left.

Heading is a misnomer; these are meant to be additional (and possibly unimportant) info. That's why they are on the right rather than left, away from the readers main path of reading.

> And we could add a bit of CSS to solve the previously mentioned issues

What about when the page is converted to pdf?

damithc:
In terms of saving vertical space, perhaps we can make the heading overlap with the box about half a line height? Not sure how it will look though. Can try it out.
image

nbriannl:
>Heading is a misnomer

Then, should it be given another name, like 'alt-text'?

damithc:
> Then, should it be given another name, like 'alt-text'?

I think we should. `alt-text` might not work though, as it already is used for a different purpose (in HTML).

alyip98:
> What about when the page is converted to pdf?

We could use the `.d-print-none` class provided by Bootstrap to hide it from print.

damithc:
> We could use the `.d-print-none` class provided by Bootstrap to hide it from print.

The user may not want to hide it from the print.

nbriannl:
image

>In terms of saving vertical space, perhaps we can make the heading overlap with the box about half a line height? Not sure how it will look though. Can try it out.

This is how it looks like

nbriannl:
image

As discussed, this is the final one.

asset/css/markbind.css
```diff
@@ -25,6 +25,31 @@ pre > code.hljs {
...
 
+pre > code.hljs[heading] {
+    border-top-right-radius: 0;
+}
+
+.code-block {
+    position: relative;
+}
+
+.code-block > .code-block-heading {
```
marvinchin: `.code-block-heading` sounds specific enough without having to further qualify it with `.code-block >`. Let's remove the unnecessary qualifiers.

docs/userGuide/syntax/code.mbdf
```diff
@@ -47,9 +61,111 @@ As seen above, line numbers are automatically provided for a code block. To hide
...
+
+```xml {heading="Heading title"}
+<foo>
+  <bar type="name">goo</bar>
+</foo>
+```
+</span>
+</include>
+
+##### Using them altogether
```
marvinchin: I feel that this section isn't very necessary, since combining selectors should be fairly self explanatory. I'd lean towards removing this section to avoid bloating the user guide.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -45,17 +45,44 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (!highlighted) {
     str = markdownIt.utils.escapeHtml(str);
   }
+  let highlightLines = token.attrGet('highlight-lines')
+  let lineNumbers = []
+  if (highlightLines) {
+    lineNumbers = highlightLines.split(',').map(v => v.split('-').map(v => parseInt(v, 10)))
```
marvinchin: Some comments here to describe what the input format should look like would be helpful! It would make maintaining this code easier in the future. 

Splitting up this line into several steps with well-named intermediate variables could help as well.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -45,17 +45,44 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+  str =  lines.map((line, index) => {
+    const currentLineNumber = index + 1
+    const inRange = lineNumbers.some(([start, end]) =>{
+      if (start && end) {
+        return currentLineNumber >= start && currentLineNumber <= end
+      }
+      return currentLineNumber === start
+    })
+    if (inRange) {
+      return `<span class="highlighted">${line || '&#x200B;'}</span>`;
```
marvinchin: I think we can abstract the section `${line || '&#x200B;'}` into a separate variable, to make it clearer that this is identical across both cases.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -45,17 +45,44 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (!highlighted) {
     str = markdownIt.utils.escapeHtml(str);
   }
+  let highlightLines = token.attrGet('highlight-lines')
```
marvinchin: You have missing `;`s in this file.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -45,17 +45,44 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
   if (highlighted) {
     token.attrJoin('class', lang);
   }
-  return `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
+
+  const heading = token.attrGet('heading')
+  if (heading) {
+    return `<div class='code-block'>`
+      + `<div class='code-block-heading'><span>` + heading + `<span></div>` 
+      + `<div class='code-block-content'><pre><code ${slf.renderAttrs(token)}>${str}</code></pre></div>`
```
marvinchin: Same as above, lets abstract the common part into it's own variable.

nbriannl: Done

src/lib/markbind/src/lib/markdown-it/index.js
nbriannl: Done. Turns out eslint ignores `src/lib/markbind/src/lib/markdown-it/*` 😱

src/lib/markbind/src/lib/markdown-it/index.js
nbriannl: Addressed.

src/lib/markbind/src/lib/markdown-it/index.js
nbriannl: Edited.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -42,6 +30,72 @@ markdownIt.renderer.rules.table_close = (tokens, idx) => {
...
+
+  const lines = str.split('\n');
+  lines.pop(); // last line is always a single '\n' newline, so we remove it
+  // wrap all lines with <span> so we can number them
+  str = lines.map((line, index) => {
+    // if a line is empty we put a 0 width non breaking space
+    const content = line || '&#x200B;';
+    const currentLineNumber = index + 1;
+    // check if there is at least one range or line number that matches the current line number
+    const inRange = lineNumbersAndRanges.some(([start, end]) => {
```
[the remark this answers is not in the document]
nbriannl: I referenced the algorithm off [markdown-it-highlight-lines](https://github.com/egoist/markdown-it-highlight-lines/blob/master/src/index.js)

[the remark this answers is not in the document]
marvinchin: That's a good point. You could bring it down to O(lines) with a more efficient algorithm here, but I don't think it would make much of a performance difference unless the user has some really massive files.

I'd towards a simpler implementation that's easier to reason about for the purpose of maintainability. If we do find this a bottleneck at some point in time, we can optimize it then 🙂

nbriannl:
> I was playing around with the feature and found an issue when the heading is too long:
> Screenshot 2020-02-13 at 11 10 50 PM
> 
> Perhaps we could either split it across several lines, or truncate it after a certain length?

`overflow-wrap: break-word;` does the [trick](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap).

src/lib/markbind/src/lib/markdown-it/index.js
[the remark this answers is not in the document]
nbriannl: Very well then, is there a way we can let future devs know that this is a possible area of optimization? A //todo perhaps?

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -45,17 +45,55 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (!highlighted) {
     str = markdownIt.utils.escapeHtml(str);
   }
+  
+  const linesToHighlight = token.attrGet('highlight-lines');
```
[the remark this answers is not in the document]
nbriannl: Looks good to me.

src/lib/markbind/src/lib/markdown-it/index.js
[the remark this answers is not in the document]
marvinchin: It might be worth leaving a comment explaining why we chose to go with this approach for now (simplicity). I wouldn't necessarily label it a TODO because there's no definite or expected need for us to do anything to change this.

nbriannl:
#991 has been merged. Ready for review.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -64,18 +64,57 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
   token.attrJoin('class', 'hljs');
   if (highlighted) {
     token.attrJoin('class', lang);
   }
-  return `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
+
+  const heading = token.attrGet('heading');
+  const codeBlockContent = `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
+  if (heading) {
+    return `${'<div class=\'code-block\'>'
```
marvinchin: The nested string interpolation feels a little strange to me. I normally do prefer string interpolation to concatenation, but for multi-line strings it gets difficult to read. What do you think of this?
```
return '<div class="code-block">'
      + `<div class="code-block-heading"><span>${heading}<span></div>`
      + `<div class="code-block-content">${codeBlockContent}</div>`
      + '</div>';
```

marvinchin: Sorry if this diff was already included when I last reviewed it and I didn't catch it then 🙏

nbriannl: Done 👍

yamgent:
@nbriannl will need you to update the css files, thanks.

nbriannl:
CSS files updated. @yamgent

damithc:
@nbriannl the netlify preview of this PR has some hover over glitches. Are they being addressed elsewhere?
image

nbriannl:
@damithc what is the hover over glitch specifically?

damithc:
The page width seem to change as I scroll the page.  In the top half of the page, the middle column seem to be wider than intended, which pushes the pageNav gets out of view (as seen from the screenshot) but things go back to normal in the 2nd half of the page. Scroll that specific page in the user guide from top to bottom while watching the width of the middle column. Let me know if you can't reproduce it.

nbriannl:
I can reproduce it in this PR. https://github.com/MarkBind/markbind/issues/1077 has been raised regarding this.

damithc:
> I can reproduce it in this PR. #1077 has been raised regarding this.

Got it. Seems severe enough problem that needs to be fixed before the next release though. Are you working on it?

nbriannl:
I was not working on it and this is the first time I'm aware regarding this. I'm investigating it now.

docs/userGuide/syntax/code.mbdf
nbriannl: Turns out we do need it 😅

PR #1035, opened 2020-02-09
Fix frontmatter not live-reloading on header/footer update

### LINKED ISSUE #1001: Inconsistent reloading of site on updating frontmatter
* **MarkBind Version:**
MarkBind 2.9.1

* Created a MarkBind site using the default template: `markbind init --template default`
* Edited `index.md` to use the default layout on that page by replacing `header: header.md` with `layout: default`. 
* Executed `markbind serve`
* `index.html` w/o header live-reloads correctly.
* Replaced default layout(`layout: default`) with header in frontmatter (`header: header.md`) 
* `index.html` live-reloads with the header, as expected
* Retried using the default layout on that page by replacing `header: header.md` with `layout: default`

(same results with `footer: footer.md`)

```html
<frontmatter>
  layout: default
  pageNav: 2
  pageNavTitle: "Chapters of This Page"
  siteNav: site-nav.md
</frontmatter>
```

The default layout to take over and the original header/footer in `header.md`/`footer.md` to not take effect.

The header/footer doesn't get removed.
However, if I disconnect the sever and re-run `markbind build` / `markbind serve`, the change is reflected i.e. the `default` layout is rendered instead of the header in `header.md`.

If this behaviour is intentional, perhaps we can document it under `Frontmatter` or `Page Layouts`?

Parcly-Taxel:

• [X] Bug fix

This one-line update fixes #1001 using the suggested fix provided [here](https://github.com/MarkBind/markbind/issues/1001#issuecomment-579854673).

This is also my first pull request to this project.

**Testing instructions:**
See the issue opener on #1001.

Fix reload inconsistency when updating frontmatter


yash-chowdhary:
Thanks for your PR @Parcly-Taxel! 
As mentioned in #1001's comments - [1](https://github.com/MarkBind/markbind/issues/1001#issuecomment-579854673) and [2](https://github.com/MarkBind/markbind/issues/1001#issuecomment-582775751), we can also go ahead with the other suggested fix of filling the `header.md` and `footer.md` files to make sure they're not blank.

Parcly-Taxel:
> As mentioned in #1001's comments, we can also go ahead with the other suggested fix of filling the `header.md` and `footer.md` files to make sure they're not blank.

@yamgent I've made the suggested fix. Please review.

With the current diff utilities it is impossible to ignore timestamps like the one in MarkBind's default footer. I had to change `diffHtml.js`: see [here](https://github.com/MarkBind/markbind/pull/1035/files#diff-8efa11e479e524fcdc0cab45a29fd839L12-R14).

(Should I also ignore the version number?)

test/functional/testUtil/diffHtml.js
```diff
@@ -9,7 +9,9 @@ const DiffPrinter = require('./diffPrinter');
  * @returns {bool} if diff was found
  */
 const diffHtml = (expected, actual, filePathName) => {
-  const diffParts = jsdiff.diffWords(expected, actual);
+  // ignore default template's footer
+  const timeRegex = /\[Generated by .*? on .*?\]/;
```
ang-zeyu: Could we use an empty footer instead?
This change may cause text explicitly typed by the user that matches the regex to not be detected.

Looks good otherwise 👍

Parcly-Taxel: If that line is not present, five of the tests fail.

ang-zeyu: Yes, because `{{ timestamp }}` is present in the footer, those tests would fail.
You could make the footer empty ( but keep the `<footer>` tag ) to fix it

ang-zeyu: Just noticed `src/template/minimal` has the same `footer/header/navgation.md` files too, do update those as well

Parcly-Taxel: @ang-zeyu I only needed to take out one line to make all the tests pass (i.e. the footer remains non-empty). Please re-review.

And by the way, what files am I supposed to change to make the footer empty?

ang-zeyu: > And by the way, what files am I supposed to change to make the footer empty?

That would be the same `footer.md` file you already edited, but removing the `{{ timestamp }}` is fine as well 👍 

Do make the changes for the minimal template's files too ( `src/template/minimal` )

Parcly-Taxel:
@ang-zeyu how about now?

src/template/minimal/_markbind/layouts/default/footer.md
```diff
@@ -0,0 +1 @@
+<footer/>
```
ang-zeyu: Could we use
```
<footer/header/navigation>
---empty line---
</footer/header/navigation>
```

in the minimal template files instead?
So if the user wants to add content it'll be slightly simpler

Parcly-Taxel: @ang-zeyu Yeah, that will do.

ang-zeyu:
Ok looks good 👍 

Do squash your commits

As for the commit message `Default template: fill header and footer`, perhaps the subject line should instead address the issue being fixed here. The body could then explain why it occurs, and then the change you made and how it fixes that.
You can browse some of the merged PRs for samples

Parcly-Taxel:
>Do squash your commits

How do I do that?

Parcly-Taxel:
>Do squash your commits

@ang-zeyu How do I do that?

ang-zeyu:
> @ang-zeyu How do I do that? I've made a pull from master during this pull request, so I can't just rewind all the way, right?

It should be possible to rename your master branch to something else then rebase on top of it, though that will likely mean this PR has to be closed and a new one opened from the renamed branch. You can link back to this PR for some context if that's the case.

This repo practices squash-rebase workflow, so your master branch should track this repo's master;
Your PRs should also hence be from a separate branch, and conflicts should be resolved by rebasing instead of merging. ( you could still merge in the interim though, then rebase once its ready, but its problematic here since your PR is from `master` )

Issue #1027, opened 2020-02-06
Better documentation and contribution guide


le0tan:
**Is your request related to a problem?**

Currently not all functions in MarkBind are well-documented, and in the devguide there's no overview of how the generation works - which may make it difficulty for a newcomer to get started with contributing to MarkBind.

**Describe the solution you'd like**

- [ ]  JSDoc for all functions
- [x] Introduction to the architecture of MarkBind
- [x] Contribution guide to components (adding new components or adding features to existing components)

**Describe alternatives you've considered**

na

**Additional context**

na

nbriannl:
I agree, it would be really good to end off the semester with good documentation for future MarkBind devs.

nbriannl:
Related to #816 #1139

jovyntls:
> * [ ]   JSDoc for all functions

Could I check what the scope of this would be (do we want _all_ functions to have JSDoc)? 
Some cases to consider:
- for groups of functions (such as [processXXXAttributes](https://github.com/MarkBind/markbind/blob/1bd81ff03b9eecf142deed8832e4fe35ec2af823/packages/core/src/html/MdAttributeRenderer.js#L103-L145)) that perform a similar job - may be better to have one overarching JSDoc comment for the group since a JSDoc for each function may be repetitive and not provide additional value to the reader
- simpler functions that might be self-explanatory, e.g. [`isMarkdownFileExt`](https://github.com/MarkBind/markbind/blob/1bd81ff03b9eecf142deed8832e4fe35ec2af823/packages/core/src/utils/fsUtil.js#L18-L20), might not need JSDocs?

tlylt:
> > * [ ]    JSDoc for all functions
> 
> Could I check what the scope of this would be (do we want _all_ functions to have JSDoc)? Some cases to consider:

I feel that this scope is too broad and perhaps a little impractical.

Also given that we are moving to TypeScript in the future, JSDocs could be less useful in general (the input, output return type part).

kaixin-hc:
I think the most confusing part of the js section is the Site Generation file, and the confusion mainly comes from how large it is rather than the functions being undocumented. Perhaps if #1756 can be achieved we can close this issue?

ang-zeyu:
@jovyntls @tlylt agreed. Jsdocs could still be useful for providing high level descriptions of classes / functions.

To provide some context, at this point https://github.com/MarkBind/markbind/pull/1039 we haven't really been considering typescript yet (was suggested in the thread briefly). 

[...]

PR #1039, opened 2020-02-11
Update JSDoc type and param info for Page/Site.js

le0tan:

• [X] Other, please explain:

Related to #1027 

* Add typedefs for configuration objects in `Page.js`.
* Add property typedefs for Page object.
* Updated some outdated JSDoc params in both `Site.js` and `Page.js`.

na

Whether the documentation is accurate.

**Testing instructions:**

No functional changes. No testing needed.

TODO


le0tan:
@alyip98 @nbriannl thanks for the input! I will request another review after settling some of the comments.

> Another PR can be made to flesh out the documentation even further, which will require someone who understands what's going on in Site and Page.

I think @alyip98 showed us some documentation on this before?

src/Page.js
```diff
@@ -55,48 +55,212 @@ cheerio.prototype.options.xmlMode = true; // Enable xml mode for self-closing ta
...
+   * @property {Set<any>} baseUrlMap
+   * @property {string} content
+   * @property {string} faviconUrl
+   * @property {Object<string, any>} frontmatter
+   * @property {string} layout
+   * @property {string} layoutsAssetPath
+   * @property {string} rootPath
+   * @property {boolean} enableSearch
+   * @property {boolean} globalOverride
+   * @property {Array} plugins
```
marvinchin, line 72: Should we label the type of the content in the array?

src/Page.js
```diff
@@ -847,8 +1027,21 @@ class Page {
...
 
+  /**
+   * A plugin configuration object.
+   * @typedef {Object<string, any>} PluginConfig
+   * @property {number} headingIndexingLevel
+   * @property {boolean} enableSearch
+   * @property {boolean} searchable
+   * @property {string} rootPath
+   * @property {string} sourcePath
+   * @property {Set} includedFiles
```
marvinchin, line 1038: Same here, should we label the contents of the set?

alyip98:
> I think @alyip98 showed us some documentation on this before?

It's rather outdated and I don't have plans to update it, so feel free to take over if you'd like 
[link](https://github.com/alyip98/markbind/blob/812-documentation/docs/devGuide/activity.md)
[pdf](https://github.com/MarkBind/markbind/files/4432139/Markbind.Page.and.Site.Reference.pdf)

src/Page.js
le0tan: The content in `plugin` array is the `require`d plugin objects (e.g. `plugins['name']=require('name');`)

PR #1043, opened 2020-02-15
Add copy code-block plugin

### LINKED ISSUE #562: Give a copy button in code blocks

yash-chowdhary:

• [x] New feature

Closes #562 

Lets users copy code in a given code block.
Since one of MarkBind's major use cases is project documentation, copying code blocks would be very beneficial to users.

* Picked up from where @amad-person left off with #629.
* Addressed comments brought up in #629 (making it a plugin, showing status of code being copied).
* Converted this feature into a plugin.
* Made use of the [Docusaurus style](https://docusaurus.io/docs/en/installation#verifying-installation) of copying code blocks. 
* Used styling as in #957 for alignment of the button.

**Demo**

ezgif com-video-to-gif-2

N/A

**Testing instructions:**
* Since this is a plugin, add the following to `site.json` :
  ```JSON
  ...
  "plugins": [
      "codeBlockCopyButtons"
  ]
  ...
  ```

Add copy code-block plugin


openorclose:
What's the reason to use clipboard.js, a rather large file, over a shorter code snippet?

Is it for browser compatibility?

yash-chowdhary:
In part, yes. I was also facing some difficulty in using `document.execCommand("copy")` as it requires the `id` of the element and our `<pre>` elements don't have `id`s.

openorclose:
> In part, yes. I was also facing some difficulty in using `document.execCommand("copy")` as it requires the `id` of the element and our `<pre>` elements don't have `id`s.

The button is located inside the `<pre>` element, so you should be able to access it with `event.target.parentElement`.

yash-chowdhary:
I've removed the dependency on `clipboard.js`. Updated the PR description accordingly.

src/constants.js
```diff
@@ -64,6 +64,15 @@ module.exports = {
   ALGOLIA_JS_URL: 'https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js',
   ALGOLIA_INPUT_SELECTOR: '#algolia-search-input',
 
+  // src/plugins/copyCode.js
+  COPIED_TO_CLIPBOARD: 'Copied!',
```
marvinchin: Since we use these text only for display purposes, and will not be reusing them anywhere else, is it necessary for us to put this in the constants file?

docs/userGuide/plugins/copyCode.mbdf
```diff
@@ -0,0 +1,20 @@
+#### `copyCode`: Enabling code-blocks to be copied
```
marvinchin: I feel like the name `copyCode` might not be explicit enough. What do you think of `codeBlockCopyButtons`?

asset/css/markbind.css
```diff
@@ -302,3 +302,50 @@ panel[minimized]:not([expanded])::before {
 li.footnote-item:target {
     background-color: #eee;
 }
+
+/* "Copy" code block button */
+pre {
+    position: relative;
+}
+
+pre > .copy-btn-icon {
```
marvinchin: IMO `.copy-btn-icon` is specific enough without requiring the `pre >` to be specified.

marvinchin: Is there a reason why `copy-btn-icon` and `copy-btn` are split? They seem to only be applied together on the same element. `copy-btn-icon` also seems to be setting the styles for the button itself, and not the icon.

asset/css/markbind.css
```diff
@@ -302,3 +302,50 @@ panel[minimized]:not([expanded])::before {
...
+    padding: 0.25em 0.4em;
+    position: absolute;
+    right: 0.3em;
+    text-align: center;
+    top: 0.4em;
+    vertical-align: baseline;
+    white-space: nowrap;
+}
+
+pre .copy-btn-icon:hover {
```
marvinchin: Is there a reason why the above has a has the child selector `>` but not this?

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+    </button>`;
+  return html;
+}
+
+function getCopyCodeBlockScript() {
+  return `<script>
+    document.onclick = ((event) => {
+        let element = event.target;
+        const elementClassList = [...element.classList];
+        const listOfSelectors = ['copy-btn-label', 'copy-btn', 'copy-btn-icon', 'copy-btn-body', 'cpy'];
```
marvinchin: This should be declared outside the `onClick` handler. Also, a more descriptive name (e.g. `copyButtonSelectors`) might help to explain what exactly these selectors are for.

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+        let parents = [];
+        let copyButton = null;
+        while (element) {
+            parents.unshift(element);
+            let classList = element.classList;
```
marvinchin: `classList` doesn't seem to be re-used anywhere. Can we make it `const`?

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+                        copyButton = element;
+                    }
+                });
+            }
+            element = element.parentNode;
+        }
+        
+        const copyButtonClicked = listOfSelectors.some(sel => elementClassList.includes(sel));
+        if (copyButtonClicked) {
+            var pre;
```
marvinchin: Let's use `let` instead of `var`.

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+            for (const obj of parents) {
+                    if (obj['outerHTML'] && obj['outerHTML'].indexOf('<pre>') === 0) {
+                    pre = obj;
+                }
+            }
+            let codeElement = pre.querySelector('code');
```
marvinchin: Similarly, this isn't changed anywhere. Let's make it `const`.

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+    ${COPY_ICON}
+    <strong class="copy-btn-label">${COPY_TO_CLIPBOARD}</strong>
+    </div>
+    </button>`;
+  return html;
+}
+
+function getCopyCodeBlockScript() {
+  return `<script>
+    document.onclick = ((event) => {
```
marvinchin: It feels really strange to have an `onclick` handler for the entire document, then check to see if it's hitting the button. Can we just add an `onclick` handler to the button directly instead?

https://www.w3schools.com/jsref/event_onclick.asp

src/plugins/copyCode.js
```diff
@@ -0,0 +1,77 @@
...
+        if (copyButtonClicked) {
+            var pre;
+            for (const obj of parents) {
+                    if (obj['outerHTML'] && obj['outerHTML'].indexOf('<pre>') === 0) {
+                    pre = obj;
+                }
+            }
+            let codeElement = pre.querySelector('code');
+            const textElement = document.createElement('textarea');
+            textElement.value = codeElement.textContent;
```
marvinchin: I wonder if we can just select the codeElement directly and do the copy without having to create the dummy text area. If that is not possible, let's add some comments here to explain why we are creating the dummy text area.

yash-chowdhary: The reason I used the dummy text area is because `select()` is defined for `<textarea>` and `<input>` elements. I can add comments to explain the usage 👍

docs/userGuide/plugins/copyCode.mbdf
yash-chowdhary: I don't have any objection to it. I just tried to make it short, but you're right - it doesn't convey the entire purpose. Will make the change 👍

src/constants.js
yash-chowdhary: Fair enough. Will move them out 👍

asset/css/markbind.css
yash-chowdhary: Resolved. No reason - forgot to remove the `>` after debugging.

src/plugins/copyCode.js
yash-chowdhary: Removed it entirely since the click event listener is added to the `Copy` button.

src/plugins/copyCode.js
yash-chowdhary: Oh yeah, you're right 😅 Made the change 👍

src/plugins/codeBlockCopyButtons.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+function getCopyCodeBlockScript() {
+  return `<script>
+    function copyCodeBlock() {
+        let element = event.target;
+        const elementClassList = [...element.classList];
+        let parents = [];
+        let copyButton = null;
+        while (element) {
```
openorclose: Since now the onclick is on the button directly, we don't need this part anymore.

src/plugins/codeBlockCopyButtons.js
```diff
@@ -0,0 +1,79 @@
...
+                        copyButton = element;
+                    }
+                });
+            }
+            element = element.parentNode;
+        }
+        
+        let pre;
+        for (const obj of parents) {
+                if (obj['outerHTML'] && obj['outerHTML'].indexOf('<pre>') === 0) {
```
openorclose: Similarly, we know that the pre is the direct parent of the button, so we don't need this part too.

src/plugins/codeBlockCopyButtons.js
```diff
@@ -0,0 +1,79 @@
...
+  const html = `<button onclick="copyCodeBlock()" class="copy-btn" type="button">
+    <div class="copy-btn-body">
+    ${COPY_ICON}
+    <strong class="copy-btn-label">${COPY_TO_CLIPBOARD}</strong>
+    </div>
+    </button>`;
+  return html;
+}
+
+function getCopyCodeBlockScript() {
```
openorclose: This function could be a string, that's all it does.

asset/css/markbind.css
```diff
@@ -302,3 +302,46 @@ panel[minimized]:not([expanded])::before {
...
+    color: #f8f8ff;
+    display: inline-block;
+    font-size: 75%;
+    line-height: 1;
+    padding: 0.25em 0.4em;
+    position: absolute;
+    right: 0.5em;
+    text-align: center;
+    top: 0.4em;
+    vertical-align: baseline;
```
marvinchin: Is `vertical-align: baseline;` necessary? It seems to be the default value.

https://www.w3schools.com/cssref/pr_pos_vertical-align.asp

yash-chowdhary: Resolved 👍

yash-chowdhary:
Ready for review

yash-chowdhary:
Rebased on master. Resolved conflicts.

Issue #984, opened 2020-01-18
Strong support for diagrams


damithc:
As project documentation is one of our target use cases, we should be able to say we have the best support for diagramming.
* Strengthen support for PlantUML
* Possibly, add support for other type of diagramming tools (?)

crphang:
Some more diagram support with libraries like https://github.com/mermaid-js/mermaid could help.

Further abstraction the work already done for inline PUML: #968 can allow integration with these diagrams to be done more easily.

ang-zeyu:
I'm almost done with the core support for this as mentioned in #968 
( providing an api to let htmlparser2 and markdown-it ignore everything in special tags )

but am stuck at a design decision for the markdown-it patch

Specifically, the _standard_ behaviour of markdown-it is such that something like this dosen't work:
```
<div>
<script>
alert("hello!")

alert("world")
</script>
</div>
```
The script ( special ) tags will not work at all, unless
- you leave an empty line between the first opening `<div>` and the first opening `<script>` tag, 
- or remove the empty line between the two `alert`s.

This actually adheres to commonmark spec, hence, do we want to follow this behaviour?

The best user behaviour for any custom tag I have in mind would of course instead be that _the special tag works even without a `<htmltag>` directly on top of it._ 
( it would work with the empty line between the alerts, or any number of empty lines, although removing the empty line is a user solution to the problem )

The full list of html blocks ( the `<div>s` ) that would cause this ( point 6 [here](https://spec.commonmark.org/0.28/#html-blocks) )

damithc:
Adding support for this is useful/feasible? https://visjs.org/

damithc:
Let's continue to try to integrate Mermaid, as it can bring in a significant value to our diagrams support.

yucheng11122017:
Mermaid supported through plugin now :)

PR #1047, opened 2020-02-17
Implement an api to ignore text in certain tags

ang-zeyu:

• [x] New feature

Groundwork for #984 

To provide a convenient interface, for plugins / core markbind code to blacklist certain tags to ignore in parsers.

This should open the doors to custom components that can have all sorts of strange syntaxes inside them, without using markdown code blocks. e.g. `puml` #968 

- Patch an existing _rule_ in markdown-it, exposing an interface that allows injecting tags to ignore before registering the patch. The patch allows these additional tags to be treated the same way `<script/pre/style>` tags are treated in markdown-it currently.
- Patch many of htmlparser2 state handlers and added some state variables to the tokenizer, exposing a similar interface.

```js
function injectIgnoreTags(tagsToIgnore) {
	Tokenizer.prototype.specialTagNames = Tokenizer.prototype.specialTagNames.concat(...tagsToIgnore)
}
```

https://github.com/MarkBind/markbind/issues/984#issuecomment-586812974 - degree of commonmark compliance we should follow

**Testing instructions:**
- `npm run test` should work
- rebuilding the 2103 site should show no diffs, save for timestamps and log output

Implement an api to ignore content in special tags

The parsers MarkBind uses parses content in html tags as html
or markdown respectively.
This makes it difficult to add custom components that utilize
conflicting syntax.

Let’s add an interface to ignore content in certain special tags.
This interface is also exposed to plugins as the getSpecialTags option.


ang-zeyu:
@crphang 
The interface is exposed through the two inject methods, called under `collectPluginSpecialTags` once during `Site` generation.
Plugins implement the `getSpecialTags: () => [ 'tagnames' , '...', ...]` method to provide these tag names.
Let me know if this works for what you have in mind, thanks!

crphang:
Let me branch off this and test for inline puml.

ang-zeyu:
> Let me branch off this and test for inline puml.

👍 , the content in the inline puml should be exposed as a text node to cheerio

ang-zeyu:
> Let me branch off this and test for inline puml

Ready for review,
Not much changed from the first commit apart from adding a functional test and cleaning some things up.

crphang:
Sorry for the delay in reviewing, will take a deeper look into the PR shortly.

For now, I've branched off this PR https://github.com/crphang/markbind/tree/inline-puml-plugin and tested inline puml, works really well :+1:

ang-zeyu:
> Sorry for the delay in reviewing, will take a deeper look into the PR shortly.
> 
> For now, I've branched off this PR https://github.com/crphang/markbind/tree/inline-puml-plugin and tested inline puml, works really well 👍

No worries; Look forward to the finished version 😄

ang-zeyu:
> Not a complete review, but can i suggest a change to the specialcharacter parsing logic?

Hmm, I went with the same char-by-char approach as htmlparser2 mainly for performance, which is essential for something used throughout markbind.

I definitely see some areas in the two functions which can be made simpler after looking at it again though.
Will update shortly, let me know if its better 😁

openorclose:
> > Not a complete review, but can i suggest a change to the specialcharacter parsing logic?
> 
> Hmm, I went with the same char-by-char approach as htmlparser2 mainly for performance, which is essential for something used throughout markbind.
> 
> I definitely see some areas in the two functions which can be made simpler after looking at it again though.
> ~Will update shortly~ Updated, let me know if its better grin

Yup this looks easier to understand

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -129,6 +148,189 @@ Tokenizer.prototype._stateText = function(c){
...
+
+	this._nextSpecialTagMatchIndex = 0;
+	return NO_MATCH;
+};
+
+/**
+ * Changes the Tokenizer state to BEFORE_SPECIAL_END if the token matches one of
+ * the first character of the currently matched special tag.
+ */
+Tokenizer.prototype._stateBeforeCloseingTagName = function(c) {
```
crphang: Minor: `closingTagName`

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -129,6 +148,189 @@ Tokenizer.prototype._stateText = function(c){
...
+	if (whitespace(c));
```
crphang: What do we do if it is whitespace?

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -129,6 +148,189 @@ Tokenizer.prototype._stateText = function(c){
...
+};
+
+/**
+ * Processes the _matchingSpecialTags array, filtering out previous match indexes
+ * that do not match the current token.
+ * If one of the previous matches successfully matched, the match index is returned.
+ */
+Tokenizer.prototype.processSpecialFunctions = function(c) {
+	const matchingSpecialTags = [];
+	const numMatchingTags = this._matchingSpecialTags.length;
```
crphang: Minor: Could `_matchingSpecialTags` be renamed to `_matchingSpecialTagsIndex`? It makes the operation below a little less confusing.

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -95,20 +99,35 @@ var i = 0,
...
 
-    SPECIAL_NONE              = j++,
-    SPECIAL_SCRIPT            = j++,
-    SPECIAL_STYLE             = j++;
+	/*
+	 Non parser-state constants
+	 */
+
+	// processSpecialFunctions, processSpecialClosingTagCharacter
+	HAS_MATCHING              = -1,
```
crphang: Do you think it is possible to follow the previous enumerations for these

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: It's a patched method from htmlparser2

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: This line is from html parser2 as well; Its to allow something like `</    tagname>` to work

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: Needed negative numbers here as the positive ones are 'taken' by the indexes

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: Updated

ang-zeyu:
Added injection to js-beautify as well ( line 762-764 of site.js ), so the input to special tags appears as is in the browser with no formatting changes ( indentations in particular ); Js-beautify already has a 'blacklist' option thankfully

crphang:
@marvinchin could you help with reviewing this? 🙏. Have a few PRs I'm hoping to rebase on this. 🙇

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -95,20 +99,35 @@ var i = 0,
...
+	NO_MATCH                  = -2,
+	HAS_MATCHED               = -3;
+
+function whitespace(c) {
+	return c === " " || c === "\n" || c === "\t" || c === "\f" || c === "\r";
```
marvinchin: Should we use a regex whitespace check instead? I'm just a little concerned about doing an explicit check because there are so many different ways to get whitespace :P

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -129,6 +148,189 @@ Tokenizer.prototype._stateText = function(c){
...
+	const lowerCaseChar = c.toLowerCase();
+	for (let j = 0; j < numSpecialTags; j += 1) {
+		if (lowerCaseChar === this.specialTagNames[j][0]) {
+			this._matchingSpecialTagIndexes.push(j)
+		}
+	}
+
+	if (this._matchingSpecialTagIndexes.length > 0) {
+		this._nextSpecialTagMatchIndex = 1;
+		return true
```
marvinchin: Missing `;` here

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-escape-special-tags.js
```diff
@@ -0,0 +1,95 @@
...
+    [ /^<!--/,        /-->/,   true ],
+    [ /^<\?/,         /\?>/,   true ],
+    [ /^<![A-Z]/,     />/,     true ],
+    [ /^<!\[CDATA\[/, /\]\]>/, true ],
+    [ new RegExp('^</?(' + block_names.join('|') + ')(?=(\\s|/?>|$))', 'i'), /^$/, true ],
+    [ new RegExp(HTML_OPEN_CLOSE_TAG_RE.source + '\\s*$'),  /^$/, false ]
+  ];
+
+
+  function escape_special_tags(state, startLine, endLine, silent) {
```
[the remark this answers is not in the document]
marvinchin: I also had some difficulty trying to find the specific part of the markdown-it documentation to understand how exactly this interacts with markdown-it. Do you think it would be helpful to link the relevant documentation in a comment?

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -129,6 +148,189 @@ Tokenizer.prototype._stateText = function(c){
...
+	const numSpecialTags = this.specialTagNames.length;
+	const lowerCaseChar = c.toLowerCase();
+	for (let j = 0; j < numSpecialTags; j += 1) {
+		if (lowerCaseChar === this.specialTagNames[j][0]) {
+			this._matchingSpecialTagIndexes.push(j)
+		}
+	}
+
+	if (this._matchingSpecialTagIndexes.length > 0) {
+		this._nextSpecialTagMatchIndex = 1;
```
marvinchin: Just wondering - why is this set to `1`? What exactly does the `nextSpecialTagMatchIndex` refer to?

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-escape-special-tags.js
```diff
@@ -0,0 +1,95 @@
+const markdownIt = require('../markdown-it');
+
+/*
+ Custom patch for the api to escape content in certain special tags
+ Adapted from the default markdown-it html_block rule and replaces it.
```
ang-zeyu: Apologies, should have really highlighted this.
The specific file is `\node_modules\markdown-it\lib\rules_block\html_block`

75% of this file ( all of the comments ) is essentially the same as it.
The patch here essentially just uses the collected special tags to form a new regex to replace the first one in the original rule.

The htmlparser2 patch has a lot of changes though, 'thankfully' 😅

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: It's a method from htmlparser2

src/lib/markbind/src/patches/htmlparser2.js
ang-zeyu: It's the next character index to match for the special tags
Here it is `1` since it matched the `0`th character

ang-zeyu:
> Thanks for explaining it - this looks okay to the best of my understanding :P

Thanks for looking through this as well!

I've made just a few more method name changes in the htmlparser2 patch:
`isFirstCharacterSpecialTagCharacter` -> `_matchSpecialTagsFirstCharacters`
`processSpecialFunctions` -> `_matchSpecialTagsNextCharacters`
`processSpecialTagClosingCharacter` -> `_matchNextSpecialTagClosingCharacter`

Issue #762, opened 2019-03-10
Provide a way to escape nunjucks syntax


damithc:
Current: any nunjucks-like syntax e.g., `{{ x }}` in MarkBind files will be processed by nunjucks, even if it is in a ~~blockquote~~ code block.

Problem: Others will not be able to use MarkBind for documenting projects that use such syntax. 

Suggested: Provide a way to escape nunjucks processing for source segments so that nunjucks-like syntax can be shown in the generated pages.

acjh:
> even if it is in a *blockquote*.

*code block*?

damithc:
> _code block_?

Yup. Corrected :-p

Xenonym:
The `jinja2` docs (of which Nunjucks is a port) [already has methods for escaping syntax](http://jinja.pocoo.org/docs/2.10/templates/#escaping), by wrapping braces in a variable (`{{ '{{' }}`/`{{ '}}' }}`) or using `{% raw %}`/`{% endraw %}` to mark a block to not be interpreted as Nunjucks.

However, these methods don't work in MarkBind because we run source files through Nunjucks multiple times, and any escaping present in one round of Nunjucks rendering gets interpreted as syntax in the next round. 

I wonder if it would be possible to retain this syntax (eg. by only rendering Nunjucks once) because it would be more user-friendly to have MarkBind use existing syntax for escaping Nunjucks instead of inventing new syntax.

yamgent:
> I wonder if it would be possible to retain this syntax (eg. by only rendering Nunjucks once)

The issue for just rendering once, is that some things need to be called multiple times before it can render correctly. For example, variables referencing other variables:

---

`variable.md`
```
<variable name="text">
Some text here.
</variable>

<variable name="example">
{{ '{{' }} example {{ '}}' }}: {{ text }}
</variable>

<variable name="howToUseVariables">
Specify the variable name with double curly-brace. For example: {{ example }}.
</variable>
```

`index.md`
```
{{ howToUseVariables }}
```

What `index.md` currently renders:
```
Specify the variable name with double curly-brace. For example: : Some text here.
```

What `index.md` may possibly render if we only do rendering once:
```
Specify the variable name with double curly-brace. For example: {{ '{{' }} example {{ '}}' }}: {{ text }}
```

The ideal output for `index.md`:
```
Specify the variable name with double curly-brace. For example: {{ example }}: Some text here.
```

luyangkenneth:
[...]

PR #1049, opened 2020-02-20
Escape Nunjucks for special tags

crphang:
Initial attempt:

It escapes the tokens `}`, `}}`, `{{` in a single pass by adding {%raw%} {%endraw%}.

In the event of `{{`, it attempts to check whether the content is a variable or not, if it is, it ignores it and goes to the next token. If not, it would escape content to the next token (Always safe to escape).

P.S.

This PR allows `{%raw%}` block tags that are used for escaping nunjuck syntax to be used for special tags. This allows special tag plugins that rely on these syntax to escape these syntax by providing a way to escape repeatedly called nunjucks render through markbind's repository.

Suggested commit message:

Provide a way to escape mustache syntax for repeatedly called nunjucks render for special tags.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -815,11 +815,10 @@ class Page {
           this.addLayoutFiles();
           this.collectHeadFiles(baseUrl, hostBaseUrl);
 
-          this.content = nunjucks.renderString(this.content,
-                                               {
-                                                 baseUrl,
-                                                 hostBaseUrl,
-                                               });
+          this.content = nunjucks.renderString(this.content, {
+            baseUrl,
+            hostBaseUrl,
+          });
 
           this.collectAllPageSections();
           this.buildPageNav();
@@ -1079,11 +1078,10 @@ class Page {
           const newBaseUrl = Page.calculateNewBaseUrl(file, this.rootPath, this.baseUrlMap);
           const baseUrl = newBaseUrl ? `${this.baseUrl}/${newBaseUrl}` : this.baseUrl;
           const hostBaseUrl = this.baseUrl;
-          const content = nunjucks.renderString(result,
-                                                {
-                                                  baseUrl,
-                                                  hostBaseUrl,
-                                                });
+          const content = nunjucks.renderString(result, {
+            baseUrl,
+            hostBaseUrl,
+          });
           return fs.outputFileAsync(resultPath, htmlBeautify(content, Page.htmlBeautifyOptions));
         })
         .then(() => {
--- a/src/Page.js
+++ b/src/Page.js
@@ -6,6 +6,7 @@ const nunjucks = require('nunjucks');
 const path = require('path');
 const pathIsInside = require('path-is-inside');
 const Promise = require('bluebird');
+const nunjuckUtils = require('./lib/markbind/src/utils/nunjuckUtils');
 
 const _ = {};
 _.isString = require('lodash/isString');
@@ -440,7 +441,7 @@ class Page {
     // Retrieve Expressive Layouts page and insert content
       fs.readFileAsync(layoutPagePath, 'utf8')
         .then(result => markbinder.includeData(layoutPagePath, result, layoutFileConfig))
-        .then(result => nj.renderString(result, template))
+        .then(result => nunjuckUtils.renderEscaped(nj, result, template))
         .then((result) => {
           this.collectIncludedFiles(markbinder.getDynamicIncludeSrc());
           this.collectIncludedFiles(markbinder.getStaticIncludeSrc());
@@ -481,7 +482,7 @@ class Page {
     // Map variables
     const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
     const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
-    return `${nunjucks.renderString(headerContent, userDefinedVariables)}\n${pageData}`;
+    return `${nunjuckUtils.renderEscaped(nunjucks, headerContent, userDefinedVariables)}\n${pageData}`;
   }
 
   /**
@@ -511,7 +512,7 @@ class Page {
     // Map variables
     const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
     const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
-    return `${pageData}\n${nunjucks.renderString(footerContent, userDefinedVariables)}`;
+    return `${pageData}\n${nunjuckUtils.renderEscaped(nunjucks, footerContent, userDefinedVariables)}`;
   }
 
   /**
@@ -548,7 +549,7 @@ class Page {
     // Map variables
     const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
     const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
-    const siteNavMappedData = nunjucks.renderString(siteNavContent, userDefinedVariables);
+    const siteNavMappedData = nunjuckUtils.renderEscaped(nunjucks, siteNavContent, userDefinedVariables);
     // Convert to HTML
     const siteNavDataSelector = cheerio.load(siteNavMappedData);
     if (siteNavDataSelector('navigation').length > 1) {
@@ -698,12 +699,12 @@ class Page {
       // Map variables
       const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
       const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
-      const headFileMappedData = nunjucks.renderString(headFileContent, userDefinedVariables)
+      const headFileMappedData = nunjuckUtils.renderEscaped(nunjucks, headFileContent, userDefinedVariables)
         .trim();
       // Split top and bottom contents
       const $ = cheerio.load(headFileMappedData, { xmlMode: false });
       if ($('head-top').length) {
-        collectedTopContent.push(nunjucks.renderString($('head-top')
+        collectedTopContent.push(nunjuckUtils.renderEscaped(nunjucks, $('head-top')
           .html(), {
           baseUrl,
           hostBaseUrl,
@@ -714,7 +715,7 @@ class Page {
         $('head-top')
           .remove();
       }
-      collectedBottomContent.push(nunjucks.renderString($.html(), {
+      collectedBottomContent.push(nunjuckUtils.renderEscaped(nunjucks, $.html(), {
         baseUrl,
         hostBaseUrl,
       })
@@ -798,6 +799,7 @@ class Page {
         .then(result => this.insertFooterFile(result))
         .then(result => Page.insertTemporaryStyles(result))
         .then(result => markbinder.resolveBaseUrl(result, fileConfig))
+        .then(result => nunjuckUtils.removeNunjucksEscapes(result))
         .then(result => fs.outputFileAsync(this.tempPath, result))
         .then(() => markbinder.renderFile(this.tempPath, fileConfig))
         .then(result => this.postRender(result))
@@ -814,10 +816,11 @@ class Page {
           this.addLayoutFiles();
           this.collectHeadFiles(baseUrl, hostBaseUrl);
 
-          this.content = nunjucks.renderString(this.content, {
-            baseUrl,
-            hostBaseUrl,
-          });
+          this.content = nunjucks.renderString(this.content,
+                                               {
+                                                 baseUrl,
+                                                 hostBaseUrl,
+                                               });
 
           this.collectAllPageSections();
           this.buildPageNav();
@@ -1063,6 +1066,7 @@ class Page {
           isDynamic: true,
           dynamicSource: source,
         }))
+        .then(result => nunjuckUtils.removeNunjucksEscapes(result))
         .then(result => fs.outputFileAsync(tempPath, result))
         .then(() => markbinder.renderFile(tempPath, {
           baseUrlMap: this.baseUrlMap,
@@ -1077,10 +1081,11 @@ class Page {
           const newBaseUrl = Page.calculateNewBaseUrl(file, this.rootPath, this.baseUrlMap);
           const baseUrl = newBaseUrl ? `${this.baseUrl}/${newBaseUrl}` : this.baseUrl;
           const hostBaseUrl = this.baseUrl;
-          const content = nunjucks.renderString(result, {
-            baseUrl,
-            hostBaseUrl,
-          });
+          const content = nunjucks.renderString(result,
+                                                {
+                                                  baseUrl,
+                                                  hostBaseUrl,
+                                                });
           return fs.outputFileAsync(resultPath, htmlBeautify(content, Page.htmlBeautifyOptions));
         })
         .then(() => {
--- a/src/Page.js
+++ b/src/Page.js
@@ -799,7 +799,6 @@ class Page {
         .then(result => this.insertFooterFile(result))
         .then(result => Page.insertTemporaryStyles(result))
         .then(result => markbinder.resolveBaseUrl(result, fileConfig))
-        .then(result => nunjuckUtils.removeNunjucksEscapes(result))
         .then(result => fs.outputFileAsync(this.tempPath, result))
         .then(() => markbinder.renderFile(this.tempPath, fileConfig))
         .then(result => this.postRender(result))
@@ -1066,7 +1065,6 @@ class Page {
           isDynamic: true,
           dynamicSource: source,
         }))
-        .then(result => nunjuckUtils.removeNunjucksEscapes(result))
         .then(result => fs.outputFileAsync(tempPath, result))
         .then(() => markbinder.renderFile(tempPath, {
           baseUrlMap: this.baseUrlMap,
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -146,10 +146,14 @@ class Parser {
     // Extract page variables from the CHILD file
     const pageVariables
       = this.extractPageVariables(asIfAt, fileContent, userDefinedVariables, includeVariables);
-    const content
-      = nunjuckUtils.renderEscaped(nunjucks, fileContent,
-                                   { ...pageVariables, ...includeVariables, ...userDefinedVariables },
-                                   { path: filePath });
+    const content = nunjuckUtils.renderEscaped(nunjucks,
+                                               fileContent,
+                                               {
+                                                 ...pageVariables,
+                                                 ...includeVariables,
+                                                 ...userDefinedVariables,
+                                               },
+                                               { path: filePath });
     const childContext = _.cloneDeep(context);
     childContext.cwf = asIfAt;
     childContext.variables = includeVariables;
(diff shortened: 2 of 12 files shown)


crphang:
I changed the approach from when I first started. This only partially address #762. The main value of this PR is that together with #1047 that escapes tags, we can completely escape content that are meant to be parsed by external libraries (diagrams, latex) which commonly uses `>` and `}` as tokens.

Previously, the attempt was more ambitious. I attempted to tokenize and capture nunjuck variable tags, and conditionally escape those if they are not a variable. This works generally well for user-defined variables within Markbind. 

The issue arrives when the user defines variables that are not within Markbind, for example in Markbind [user guide](https://github.com/MarkBind/markbind/blob/master/docs/userGuide/authoringContents.md). These variables are very hard to detect, and are not explicitly handled within Nunjucks itself see https://github.com/mozilla/nunjucks/issues/557.

Currently, we allow the user to define and escape it's own `{%raw%} content {%endraw%}` tags in places they want to escape. However, the escaping is only for plugins and will not be displayed until we managed to deal with other syntax clashes from Vue and Markdown-it-attr. For now, I explicitly remove all the raw tags before we call `render` because of syntax clash with `markdown-it-attrs` and how newlines are handled.

For instance:

```javascript

const MarkdownIt = require('markdown-it');
const md = new MarkdownIt().use(require('markdown-it-attrs'));

console.log(md.render('{%raw%}content{%endraw%}')); // <p %endraw%="">{%raw%}content</p>

// For new lines
console.log(md.render(`<div>{%raw%}

content

{%endraw%}</div>`)); 

// <p %raw%="">&lt;div&gt;</p>
// <p>content</p>
// <p>{%endraw%}&lt;/div&gt;</p>
```

test/unit/nunjuckUtils.test.js
```diff
@@ -0,0 +1,30 @@
...
+  expect(escapedContent).toBe(escapedString);
+});
+
+test('Escaping multiple nunjucks raw tags', () => {
+  const escapedString = 'Multiple escapes: {%raw%} first {%endraw%} {%raw%} second {%endraw%}';
+  const escapedContent = nunjuckUtils.renderEscaped(nunjucks, escapedString);
+
+  expect(escapedContent).toBe(escapedString);
+});
+
```
marvinchin: Do you think this nested escapes are a case we need to account for?

```
test('Escaping nested nunjucks raw tags', () => {
  const escapedString = 'This is a content with nested escaped data {%raw%} {%raw%} CONTENT {%endraw%} {%endraw%}';
  const escapedContent = nunjuckUtils.renderEscaped(nunjucks, escapedString);
  expect(escapedContent).toBe(escapedString);
});
```

I can imagine this might be useful for documenting how to use to use the escape tags in the first place

src/lib/markbind/nunjuckUtils.js
```diff
@@ -0,0 +1,28 @@
+const START_ESCAPE_STR = '{% raw %}';
+const END_ESCAPE_STR = '{% endraw %}';
+const regex = new RegExp('({%( *)raw( *)%})(.*?)({%( *)endraw( *)%})', 'gs');
+
+function addEscapeTags(match) {
+  return START_ESCAPE_STR + match + END_ESCAPE_STR;
```
marvinchin: I personally prefer string interpolation to explicit concatenation:
`${START_ESCAPE_STR}${match}${END_ESCAPE_STR}`;

crphang:
The scope of this PR will not address rendering of `{{ }}`, but would allow plugins that rely of such syntax to parse properly. would continue the discussion of deconflicting markbind-it-attr and Vue in the issue itself.

I would rebase this PR on top of #1047, will be adding an integration test to ensure that nunjuck syntax are properly escaped for plugins.

ang-zeyu:
> The scope of this PR will not address rendering of displaying `{{ }}`, but would allow plugins that rely on such syntax to parse properly. would continue the discussion of deconflicting markbind-it-attr and Vue in the issue itself.
> 
> I would rebase this PR on top of #1047, will be adding an integration test to ensure that nunjuck syntax are properly escaped for plugins.

#1047 deals with markbind-it-attr already, since the entire special tag is processed as a html block.

For vue, perhaps after this PR we could extend #1047's api with this PR's functionality.

The regex would then also comb for these additional patterns and append `{% raw/endraw %}` to them.
e.g. for markdown code blocks, we comb for triple backticks and append said tags, then we modify the renderer to add a `v-pre` [attribute](https://vuejs.org/v2/api/#v-pre) to the resulting `<pre>` tag. 

For general use, we could create a special component, something like `<escape></escape>` that acts the same as a `<span v-pre></span>` in addition to appending {% raw %} {% endraw %}

The same could be done for plugin defined components sparingly.
( I don't think we currently have plugins needing `{{ ... }}` at the moment though )

crphang:
> #1047 deals with markbind-it-attr already, since the entire special tag is processed as a html block.

Yep. Thanks to good work in #1047, special tags are isolated from markdown-it and markdown-it-attr. I was considering about how to handle other cases when the nunjuck escapes are processed by `md.render`. 

> ( I don't think we currently have plugins needing `{{ ... }}` at the moment though )

Right now we probably don't, but potentially plugins like latex will start having `{{ }}` when we have more complicated compositions.


> For general use, we could create a special component, something like `<escape></escape>` that acts the same as a `<span v-pre></span>` in addition to appending {% raw %} {% endraw %}

Something we could certainly explore

crphang:
Ready for review.

Integration test added hints at how this can be used with special tags, it naively replaced all mustaches with `success`. In a more common use case, it helps to enable proper escaped special tag which relies of mustaches such as latex or other syntax.

src/lib/markbind/src/parser.js
```diff
@@ -392,13 +397,18 @@ class Parser {
...
-          = nunjucks.renderString(
-            data,
-            { ...pageVariables, ...userDefinedVariables },
-            { path: actualFilePath });
+          = nunjuckUtils.renderEscaped(nunjucks,
+                                       data,
+                                       {
+                                         ...pageVariables,
+                                         ...userDefinedVariables,
+                                         path: actualFilePath,
```
ang-zeyu: `renderEscaped` does not seem to pass the third `options` argument with `path` to `nunjucks.renderString`;
Let's keep the `{ path: ... }` object separate as well since the user might have a variable named `path`

crphang:
> On how this can be used with special tags, would it be better to keep it to a separate PR since it hasn't been implemented?

For now, the test focuses on ensuring that a special tag plugin could retrieve an untouched unparsed mustache syntax. Before we have the special component, it would be good to ensure that there will be no regression by having the integration test.

> For general use, we could create a special component, something like <escape></escape> that acts the same as a <span v-pre></span> in addition to appending {% raw %} {% endraw %}

Would be great to have something like that. When we do, we could add other tests that checks this or even replace this current one.

crphang:
> My bad, realised I edited your comment instead of making a new one

No worries :smile:.

Thank you for your review! I reviewed some of the existing `nunjucks.RenderString` that exists, and I agree that it will get confusing when there are 2 ways of using nunjucks. Actually, there is no difference between `nunjucks.RenderString` and `nunjuckUtils.renderEscaped` after `nunjucksUtils.removeNunjucksEscapes` is called.

I have added additional tests to ensure that the behavior of rendering nunjucks to remove brackets is retained even when raw tags is used. This helps to prevent backward compatibility issues where we actually need the mustaches to be removed so that Vue can work properly until we have the component mentioned above.

test/functional/test_site/testNunjuckUtils.md
```diff
@@ -0,0 +1,9 @@
+{%raw%}
+
+The standard trick to render braces still works {<span>{ content }</span>}.
+
+Nonexistent variables should still be removed: {{ unknown_variable }}. In the rare event that user wrapped an unknown variable inside a raw tag, it should retain its current behaviour and not be produced to ensure Vue does not break.
```
ang-zeyu: Shouldn't it be un-removed? ( isn't this PR to let `{% raw %} ... {% endraw %}` work across multiple nunjucks renders? )

test/functional/test_site/testNunjuckUtils.md
```diff
@@ -0,0 +1,9 @@
...
+
+Random braces should be removed: { }
```
ang-zeyu: Isn't this due to `markdown-it-attrs`?

ang-zeyu: e.g. something like 
```
<div>
lorem { ipsum }
</div>
```
should still have the braces afterward

crphang: Removed this test, as this is not really related to the PR

test/functional/test_site/testNunjuckUtils.md
crphang: The scope of this PR is only to escape nunjucks for special tags. This PR currently does not help to render `{{ unknown_variable }}` if it is not wrapped in a special tag. That is currently the purpose of `nunjucks.removeNunjucksEscapes`, as we still have to ensure that compatibility is maintained until the following is dealt with:

1. markdown-it-attr which also uses `{`, `}`. `md.render` would actually remove all the raw tags and add it as a HTML attribute.
1. We still need to escape `{{` and `}}` in Vue. A special component mentioned that uses `v-pre` might help.

crphang:
> On that note, and `{{` still getting erased, shouldn't the final nunjucks call for the page / external dependency's output be the raw call to `nunjucks.renderString` still?

True 👍 , we should remove any additional `{% raw %}` if it still exists in the final call.

> Perhaps `nunjucks.removeNunjucksEscapes` is not necessary in this sense, as nunjucks already removes `{% raw %}... {% endraw %}`

We would still require this. As explained above, we still need to ensure `{% raw %}` tags does not change current behavior by interfering with markdown-it-attr, now that it can be "retained" until the parser stage.

test/functional/test_site/testNunjuckUtils.md
ang-zeyu: Hmm, I still don't follow 😅 
When we use your version of `nunjucks.removeNunjucksEscapes` for _all calls except the last_, shouldn't `{{ xxx }}` be preserved no matter?

> We still need to escape {{ and }} in Vue. A special component mentioned that uses v-pre might help.

> For general use, we could create a special component, something like <escape></escape> that acts the same as a <span v-pre></span> in addition to appending {% raw %} {% endraw %}

When I mentioned this, it's only a passing remark for a future PR, such that `{{ xxx }}` would be skipped in vue compilation and show up in the browser.

Before that though, shouldn't `{{ xxx }}` show up in the html output though? ( `markdown-it-attrs` does not erase `{{ xxx }}`, nor `{ xx }` if it is not placed appropriately. https://npm.runkit.com/markdown-it-attrs )

ang-zeyu: > We would still require this. As explained above, we still need to ensure {% raw %} tags does not change current behavior by interfering with markdown-it-attr, now that it can be "retained" until the parser stage.

Ahh my bad, I see now that `{% raw %}` is the offender, not `{{ xxx }}`

crphang: Yea :sweat:. 

For that, since it is not possible to change delimiters for both nunjucks and markdown-it-attr, I am separately testing whether it is possible to ignore nunjuck raw tags for markdown-it-attr: https://github.com/arve0/markdown-it-attrs/blob/a40f59a4ee690e2464c553b14a2e57e48d5418ae/utils.js#L101

src/lib/markbind/src/parser.js
```diff
@@ -145,9 +146,10 @@ class Parser {
     // Extract page variables from the CHILD file
     const pageVariables
       = this.extractPageVariables(asIfAt, fileContent, userDefinedVariables, includeVariables);
-    const content = nunjucks.renderString(fileContent,
-                                          { ...pageVariables, ...includeVariables, ...userDefinedVariables },
-                                          { path: filePath });
+    const content
```
ang-zeyu: Should we preserve the style here? :o

test/functional/test_site_special_tags/_markbind/plugins/testSpecialTag.js
```diff
@@ -17,11 +23,20 @@ function preRender(content) {
     $(testElement).text(wrappedText);
   });
 
+  const escapedNunjucks = $('mustache');
+  escapedNunjucks.each((index, element) => {
+    const unwrappedText = $(element).text();
+    const unescapedText = unwrappedText.replace(ESCAPE_REGEX, removeEscapeTags);
+    const transformedText = unescapedText.replace(/{/g, '!success').replace(/}/g, 'success!');
```
ang-zeyu: Shouldn't the text transformation ( `!success ... success!` ) be done in `removeEscapeTags`?
Text without `{% raw %}` would still be transformed otherwise

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -0,0 +1,31 @@
...
+    return nunjucks.renderString(escapedPage, variableMap, options);
+  },
+
+  /**
+   * RemoveNunjucksEscapes removes raw tags from page data when processing
+   * Downstream calls to nunjuckUtils.renderEscaped will be essentially the same as nunjucks.renderString
+   * This ensures that we remove {{ and }} from the final output which prevents problem with Vue
+   */
+  removeNunjucksEscapes(pageData) {
+    return pageData.replace(REGEX, removeEscapeTags);
```
ang-zeyu: Just realised there's the `$&` replacement pattern; Could help make this even more succinct 😄 
I'm fine with using a function as well though. ( likewise for `preEscapeRawTags` )

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter

test/functional/test_site_special_tags/_markbind/plugins/testSpecialTag.js
crphang: The previous statement is to remove `{% raw %}` for this test. This can be simplified if we have some attr like `escape` for plugins which automatically removes these raw statements. This line is to replace the mustaches `{{` to success statements for this specific test. In normal circumstance, text with/without `{% raw %}` outside of special tags would not have `{{` available during post-render since `{% raw %}` would be removed by markdown-it-attr.

crphang:
:+1: . I removed `removeNunjuckEscape` here and moved the test to `postRender` to test that special tags still works under more rounds of parsing. 

> Once we have a patch for `markdown-it-attrs` and `markdown-it`, we could then reintroduce a similar test that tests the functionality of `{% raw %}...{% endraw %}` when not wrapped by a special tag, without the need for `removeNunjucksEscapes`.

Yea we could do that. Removing `removeNunjucksEscapes` right now means that when user does `{% raw %}`, they will be parsed as classes due to markdown-it-attr but they are previously not. This is because raw tags are retained until `md.render`. There will be no key differences in user experience and this can be addressed later on.

test/functional/test_site_special_tags/_markbind/plugins/testSpecialTag.js
ang-zeyu: Ah I see, how about wrapping it with `unescapedText = unwrappedText.replace(ESCAPE_REGEX, 'raw$1endraw')` to ensure {% raw %} was present at the time of `postRender`? ( since `replace` would return the original string if there's no match as well )

crphang: That makes things clearer :+1:

test/unit/nunjuckUtils.test.js
marvinchin: Nesting with stuff in between doesn't seem to work correctly 😅 
```
test('Escaping nested nunjucks raw tags', () => {
  const escapedString = 'Nested escaped data: {% raw %}BEFORE{% raw %} CONTENT {% endraw %}AFTER{% endraw %}';
  const escapedContent = nunjuckUtils.renderEscaped(nunjucks, escapedString);
  expect(escapedContent).toBe(escapedString);
});
```
```
Expected: "Nested escaped data: {% raw %}BEFORE{% raw %} CONTENT {% endraw %}AFTER{% endraw %}"
Received: "Nested escaped data: {% raw %}BEFORE{% raw %} CONTENT {% endraw %}{% endraw %}AFTER"
```

src/Page.js
```diff
@@ -814,10 +815,11 @@ class Page {
           this.addLayoutFiles();
           this.collectHeadFiles(baseUrl, hostBaseUrl);
 
-          this.content = nunjucks.renderString(this.content, {
-            baseUrl,
-            hostBaseUrl,
-          });
+          this.content = nunjucks.renderString(this.content,
```
marvinchin: Not sure if this is personal preference, but I find the original indentation (aligned to the left column) easier to read than aligning with the left opening bracket.

test/unit/nunjuckUtils.test.js
marvinchin: It feels like as long as we go with the opentag/closetag approach, we would always run into this issue:
`$OPEN$ $CLOSE$ $CLOSE$` <- how do we match the `$OPEN` to the right `$CLOSE`?

Do you think having a special tag that escapes the next character would help? This would "break" the syntax so the original special tag wont be treated as such.

crphang:
@marvinchin Thanks for your review. I updated it and fixed some of the styles to make it consistent. I removed the test that was supporting nested tags because there is no need for the user to input nested escapes in any case. It is much better to remove the test than to suggest to further developers that we are officially supporting and maintaining an unnecessary interaction.

> Do you happen to know if we currently have any other way to escape the `{{ VARIABLE }}` syntax?

There are two main approaches for it:

1. Automated detection of `{{` and `}}`. 
1.1 If it is a variable, we simply leave as is. 
1.2 If it is not a variable, we could use `{% raw %}` tags to escape those by default. Alternatively we could try escaping using `nunjucks.RenderString`, but we inject the text in between `{{` and `}}` as a variable. However, this method is quite buggy as `renderString` does not work when the text has spaces and it is very limiting. `{% raw %}` tags is still the most reliable and users who are well-acquainted with nunjucks would be familiar with them.
1.3 Automated detection has a lot of edge cases to cover. For example, on top of having to build to cover edge cases like non-wellformed `{{` and `}}` brackets where the numbers are unequal, we also need to know what are all the variables defined in the page. For example, not all variables are defined within markbind. There are some that uses nunjucks syntax to define variable directly: https://mozilla.github.io/nunjucks/templating.html#for which some users could adopt (our own userguide). Fetching those variables to ignore escaping is not officially supported by nunjucks: https://github.com/mozilla/nunjucks/issues/329.

2. Manual user input of `{% raw %}`. Users know best what they want to ignore and we leave it as such. (1.2) describes why other ways is not that simple to escape and `{%raw%}` is simpler and more reliable.
2.1 The current "decorator" approach of supporting `{%raw%}` tags still allow us to go to automatic escaping (if we actually want). It helps to maintain the `{%raw%}` tags throughout if we are able to detect, but as discussed throughout the PR.

Hope that helps to explain!

crphang:
@marvinchin ready for review! Sorry for the double ping 🙇

damithc:
@crphang what's the outcome of this PR, from author POV?
I thought this syntax should work now, but doesn't seem to.
```
{% raw %}
{{ abc }}
{% endraw %}
```

crphang:
@damithc This works when used in specialTags, such as puml.

This allows curly braces `{{ content }}` to be retained throughout Parse, which can be seen in the integration test added, which allows later plugins that require such syntax to not be conflicted. 

To actually show the curly braces: `{{ content }}`, we need an addition as described: https://github.com/MarkBind/markbind/pull/1049#issuecomment-590297046.

> For general use, we could create a special component, something like `<escape></escape>` that acts the same as a `<span v-pre></span>` in addition to appending {% raw %} {% endraw %}

This could be a separate plugin (which can be registered as special tag) that also notifies Vue that it is raw with `v-pre`.

crphang:
From an author POV, nothing should change right now, but it allows us to define more components that could consist of conflicting syntax, in this case curly braces.

I would raise a seperate issue to track this, for creating a new plugin that allows author to render `{{ }}` syntax

damithc:
> From an author POV, nothing should change right now, but it allows developers to define more components that could consist of conflicting syntax, in this case curly braces.
> 
> I would raise a seperate issue to track this, for creating a new plugin that allows author to render `{{ }}` syntax

Got it. 👍

PR #1052, opened 2020-02-21
Use path.join instead of string interpolation

alyip98:
Assuming forward slashes may cause errors on windows systems

• [x] Bug fix

This line was causing a error in tests as `createDir` uses string splitting based on the `path.sep` native to the OS.

```
function createDir(pathArg) {
  const { dir, ext } = path.parse(pathArg);
  const dirNames = (ext === '')
    ? pathArg.split(path.sep)
    : dir.split(pathArg.sep);
```

Changed a path concatenation to use `path.join` instead of string interpolation.

It was the only instance I could find in the entire project that was using string interpolation. 

Should not have any changes

-

**Testing instructions:**
-

Assuming forward slashes may cause errors on windows systems. 
Let's use path.join to maintain cross platform compatibility.

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -920,7 +920,7 @@ class Site {
       }
       const filteredFiles = files.filter(file => _.includes(file, '.') && !_.includes(file, '.md'));
       const copyAll = Promise.all(filteredFiles.map(file =>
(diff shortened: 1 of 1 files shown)


openorclose:
LGTM 🚀

PR #1055, opened 2020-02-26
Update Test for Version 2.11.0

nbriannl:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [X] Other, please explain: Build is failing due to test file not updated. 

Build is failing. Updating test reveals one file is not updated.

Update file

Update expected output for test


yamgent:
Thanks for spotting this. I am leaving it to @marvinchin to merge so that he can test out his new merging powers. :P

On an unrelated note, this probably should be fixed in the long run by mocking the version number, so that we don't have to update the tests every release (this should be done in a separate PR).

PR #1056, opened 2020-02-26
Unify markdown-it parser variants

ang-zeyu:

• [x] Documentation update
• [x] Other, please explain:

As mentioned here: https://github.com/MarkBind/markbind/pull/975#discussion_r368469434

To standardise the two parser variants used for attributes in components and all other cases,
which should reduce the potential matainence needed, and standardises the feature set
across the two locations.

As a result:

##### Two markdown-it plugins vue-strap has but MarkBind dosen't were added:
- markdown-it-sub ( uses `~...~` )
- markdown-it-sup ( uses `^...^` )

##### Three markdown-it plugins MarkBind has but vue-strap dosen't were added to vue attributes:
- markdown-it-attrs ( uses `{ ... }` )
- markdown-it-footnotes ( uses `[^footnoteidentifier] or ^[inline footnote]` )
  - the caveat here is that non-inline footnotes don't work here as they are parsed in seperate runs, although inline ones still do.
- markdown-it-linkify-images ( no syntax changes )

##### Markdown-it plugins MarkBind has but vue-strap dosen't, but would not be used in attributes anyway, and dosen't conflict with existing syntax:
- markdown-it-table-of-contents ( uses `[[toc]]` )
- markdown-it-task-lists ( uses `1./*/- [x]` )
- markdown-it-block-embed ( uses `@[youtube](lJIrF4YjHfQ)` )
- markdown-it-radio-button ( uses `- ( ) Item 1` )

- Merged the two parser variants into one
- Corresponding documentation was updated for the above changes

na

**Testing instructions:**
- `npm run test` should pass
- `build` on the 2103 site should have no diffs, save for timestamps and logs

Unify markdown-it parser variants

The markdown-it parser used for attributes in components
uses a different set of plugins from the one used throughout markbind.

Let's unify the two variants to reduce the potential amount of
maintainence needed, and standardise the usable markdown
syntax throughout markbind.


openorclose:
LGTM

Unsure if we need to be concerned that authors can use markdown syntax that don't make sense in attributes though.

ang-zeyu:
> Unsure if we need to be concerned that authors can use markdown syntax that don't make sense in attributes though.

I would think its better we leave that to the author's discretion, since its better to have things work where expected as far as markdown goes.
E.g. since our docs states that box headers 'support markdown text', one would expect block level markdown ( e.g. markdown-it-task-lists ) to work in there as well, although the use cases for it are probably zero.

ang-zeyu:
Moved `markdown-it-escape-special-tags` from #1047 back to the `markdown-it` folder as well

PR #1057, opened 2020-02-26
Add warnings for conflicting/deprecated component attribs

### LINKED ISSUE #1048: Implement warning logging when inconsistent or deprecated attributes are used

Discussion here https://github.com/MarkBind/vue-strap/pull/118#discussion_r378933607

We can warn the user at compile time of inconsistent attributes / slots used in components simultaneously.

Implement warning logging in for various components

- Boxes
  - `light` vs `seamless` option
  - `no-xx` vs `xx` attributes ( e.g. `no-icon` + `icon` )
  - `no-background/border` and `seamless` option
- All components
  - warn about deprecated attributes
  - warn about inconsistent `slot` + `attribute` ( that also translates to the slot ) choices ( e.g. using `content` attribute and `content` slot for popovers at the same time )

nbriannl:

• [X] Other, please explain: Add warning for bad markbind attribute usage

Fixes #1048

Discussion here MarkBind/vue-strap#118 (comment)

Warning the user at compile time of the usage of conflicting/deprecated attributes in components is good for author usability.

When parsing component attributes, check for presence of inconsistent, and deprecated attributes and warn accordingly.

Used console.log instead of logger utility because `parser.test.js` fails. This might possibly be a bug regarding how `logger.js` resolves the directory. See #1060 

No documentation needed since it is a warning on the CLI side.

The code in this gist will now produce warnings
https://gist.github.com/nbriannl/5c44b3ee8a8af78c45a7a4f3659ce196

Instead of writing another method to warn for deprecated slot name (which is incurring an addtional 1 pass search on the element's children), I wanted to change `renameSlot` to `processSlot` which will do both checking for deprecated slot name, and renaming. But this assumes that every rename of a slot is for a deprecated name. Therefore, I stuck with making two separate functions.

**Testing instructions:**
`markbind serve` the example code in the gist.
https://gist.github.com/nbriannl/5c44b3ee8a8af78c45a7a4f3659ce196

You should see
image

Add warnings for conflicting/deprecated component attribs

### DIFF
--- a/src/lib/markbind/src/parsers/componentParser.js
+++ b/src/lib/markbind/src/parsers/componentParser.js
@@ -299,7 +299,7 @@ function _renameAttribute(node, originalAttribute, newAttribute) {
 function _parseModalAttributes(node) {
   _warnDeprecatedAttributes(node, { title: 'header' });
   _warnDeprecatedSlotNames(node, { 'modal-header': 'header', 'modal-footer': 'footer' });
-  
+
   _parseAttributeWithoutOverride(node, 'header', true, 'modal-title');
   _parseAttributeWithoutOverride(node, 'title', true, 'modal-title');
 
--- a/src/lib/markbind/src/parsers/componentParser.js
+++ b/src/lib/markbind/src/parsers/componentParser.js
@@ -134,6 +134,45 @@ function _assignPanelId(node) {
   }
 }
 
+/**
+ * Check and warns if element has conflicting attributes.
+ * Note that attirbutes in `attrsConflictingWith` are not in conflict with each other,
+ * but only towards `attribute`
+ * @param node Root element to check
+ * @param attribute An attribute that is conflicting with other attributes
+ * @param attrsConflictingWith The attributes conflicting with `attribute`
+ */
+function _warnConflictingAttributes(node, attribute, attrsConflictingWith) {
+  if (!(attribute in node.attribs)) {
+    return;
+  }
+  attrsConflictingWith.forEach((conflictingAttr) => {
+    if (conflictingAttr in node.attribs) {
+      // TODO: Use logger here instead of console.warn. See issue #1060
+      // eslint-disable-next-line no-console
+      console.warn(`warn: Usage of conflicting ${node.name} attributes: `
+        + `'${attribute}' with '${conflictingAttr}'`);
+    }
+  });
+}
+
+/**
+ * Check and warns if element has a deprecated attribute.
+ * @param node Root element to check
+ * @param attributeNamePairs Object of attribute name pairs with each pair in the form deprecated : correct
+ */
+function _warnDeprecatedAttributes(node, attributeNamePairs) {
+  Object.entries(attributeNamePairs).forEach(([deprecatedAttrib, correctAttrib]) => {
+    if (deprecatedAttrib in node.attribs) {
+      // TODO: Use logger here instead of console.warn. See issue #1060
+      // eslint-disable-next-line no-console
+      console.warn(`warn: ${node.name} attribute '${deprecatedAttrib}' `
+        + `is deprecated and may be removed in the future. Please use '${correctAttrib}'`);
+    }
+  });
+}
+
+
 /*
  * Triggers
  *
@@ -171,9 +210,10 @@ function _parseTrigger(node) {
  */
 
 function _parsePopover(node) {
+  _warnDeprecatedAttributes(node, { title: 'header' });
+
   _parseAttributeWithoutOverride(node, 'content', true);
   _parseAttributeWithoutOverride(node, 'header', true);
-  // TODO deprecate title attribute for popovers
   _parseAttributeWithoutOverride(node, 'title', true, 'header');
 
   node.name = 'span';
@@ -184,6 +224,32 @@ function _parsePopover(node) {
     = 'popoverInnerGenerator';
   node.attribs.class = node.attribs.class ? `${node.attribs.class} trigger` : 'trigger';
   _transformSlottedComponents(node);
+}
+
+/**
+ * Check and warns if element has a deprecated slot name
+ * @param element Root element to check
+ * @param namePairs Object of slot name pairs with each pair in the form deprecated : correct
+ */
+
+function _warnDeprecatedSlotNames(element, namePairs) {
+  if (!(element.children)) {
+    return;
+  }
+  element.children.forEach((child) => {
+    if (!(_.has(child.attribs, 'slot'))) {
+      return;
+    }
+    Object.entries(namePairs).forEach(([deprecatedName, correctName]) => {
+      if (child.attribs.slot !== deprecatedName) {
+        return;
+      }
+      // TODO: Use logger here instead of console.warn. See issue #1060
+      // eslint-disable-next-line no-console
+      console.warn(`warn: ${element.name} slot name '${deprecatedName}' `
+        + `is deprecated and may be removed in the future. Please use '${correctName}'`);
+    });
+  });
 }
 
 /*
@@ -207,7 +273,8 @@ function _parseTooltip(node) {
 
 function _renameSlot(node, originalName, newName) {
   if (node.children) {
-    node.children.forEach((child) => {
+    node.children.forEach((c) => {
+      const child = c;
       if (_.has(child.attribs, 'slot') && child.attribs.slot === originalName) {
         child.attribs.slot = newName;
       }
@@ -230,11 +297,12 @@ function _renameAttribute(node, originalAttribute, newAttribute) {
  */
 
 function _parseModalAttributes(node) {
+  _warnDeprecatedAttributes(node, { title: 'header' });
+  _warnDeprecatedSlotNames(node, { 'modal-header': 'header', 'modal-footer': 'footer' });
+  
   _parseAttributeWithoutOverride(node, 'header', true, 'modal-title');
-  // TODO deprecate title attribute for modals
   _parseAttributeWithoutOverride(node, 'title', true, 'modal-title');
 
-  // TODO deprecate modal-header, modal-footer attributes for modals
   _renameSlot(node, 'header', 'modal-header');
   _renameSlot(node, 'footer', 'modal-footer');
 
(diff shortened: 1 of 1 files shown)


src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -4,6 +4,7 @@ const _ = {};
 _.has = require('lodash/has');
 
 const vueAttrRenderer = require('../lib/vue-attribute-renderer');
+const logger = require('../../../../util/logger');
```
nbriannl: Requiring logger here fails `parser.test.js`

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -178,13 +212,18 @@ function _parseTabAttributes(element) {
  */
 
 function _parseBoxAttributes(element) {
+  _warnInconsistentAttributes(['light', 'seamless'], element.attribs);
+  _warnInconsistentAttributes(['no-background', 'background-color'], element.attribs);
+  _warnInconsistentAttributes(['no-background', 'seamless'], element.attribs);
+  _warnInconsistentAttributes(['no-border', 'border-color'], element.attribs);
+  _warnInconsistentAttributes(['no-border', 'border-left-color'], element.attribs);
+  _warnInconsistentAttributes(['no-border', 'seamless'], element.attribs);
+  _warnInconsistentAttributes(['no-icon', 'icon'], element.attribs);
```
nbriannl: [LFC] 
I considered changing 
```
function _warnInconsistentAttributes(inconsistentAttribs, elemAttribs) {
  const hasInconsistentAttribs = inconsistentAttribs.every(
    inconsistentAttrib => inconsistentAttrib in elemAttribs);

  if (hasInconsistentAttribs) {
    // TODO: Use logger here instead of console.log. See issue #1060
    // logger.warn(`Usage of inconsistent attributes: ${inconsistentAttribs.join(', ')}`);
    // eslint-disable-next-line no-console
    console.log(`warn: Usage of inconsistent attributes: ${inconsistentAttribs.join(', ')}`);
  }
}
```

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,6 +121,41 @@ function _assignPanelId(element) {
...
+      if (attrsInconsistentWith[i] in element.attribs) {
+        // TODO: Use logger here instead of console.log. See issue #1060
+        // eslint-disable-next-line no-console
+        console.log(`warn: Usage of inconsistent ${element.name} attributes: `
+          + `'${attribute}' with '${attrsInconsistentWith[i]}'`);
+        return;
+      }
+    }
+  }
+}
```
nbriannl: I wrote the function like so because i noticed that some attributes are inconsistent with a set of attributes. 
Like A-B, A-C, A-D
At first I wanted to check for the pairs individually, but I felt that cutting short the search at the absence of A would be good.

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,6 +119,42 @@ function _assignPanelId(element) {
...
+  if (!(attribute in element.attribs)) {
+    return;
+  }
+  for (let i = 0; i < attrsInconsistentWith.length; i += 1) {
+    if (attrsInconsistentWith[i] in element.attribs) {
+      // TODO: Use logger here instead of console.log. See issue #1060
+      // eslint-disable-next-line no-console
+      console.log(`warn: Usage of inconsistent ${element.name} attributes: `
+        + `'${attribute}' with '${attrsInconsistentWith[i]}'`);
+      return;
```
openorclose: Could we not `return` here? Then if there are multiple inconsistent attributes authors can see all warnings at once.

nbriannl: Oh, thanks for noticing.

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -196,10 +254,10 @@ function _parseDropdownAttributes(element) {
   const slotChildren = el.children && el.children.filter(child => _.has(child.attribs, 'slot'));
   const hasHeaderSlot = slotChildren && slotChildren.some(child => child.attribs.slot === 'header');
 
+  _warnDeprecatedAttribute(element, 'text', 'header');
   // If header slot is present, the header attribute has no effect, and we can simply remove it.
   if (hasHeaderSlot) {
     delete el.attribs.header;
```
openorclose: Should we warn here too? 

If user has header slot and header/text attr

nbriannl: Wasn't part of the original issue but good to include in the fix. 👍

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,6 +119,41 @@ function _assignPanelId(element) {
   }
 }
 
+/**
+ * Check and warns if element has conflicting attributes.
+ * @param element Root element to check
+ * @param attribute An attribute that is conflicting with other attributes
+ * @param attrsConflictingWith The attributes conflicting with `attribute`
+ */
+function _warnConflictingAttributes(element, attribute, attrsConflictingWith) {
```
marvinchin: Could we just take in a set of conflicting attributes and ensure that at most one attribute from the set is applied to the element? I'm thinking of the case where we might have more than two attributes that are conflicting (e.g. `light`, `seamless`, `dark`).

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,6 +119,41 @@ function _assignPanelId(element) {
...
+  if (!(attribute in element.attribs)) {
+    return;
+  }
+  for (let i = 0; i < attrsConflictingWith.length; i += 1) {
```
marvinchin: Can we use `i++`?

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -119,6 +119,41 @@ function _assignPanelId(element) {
...
+  }
+}
+
+/**
+ * Check and warns if element has a deprecated attribute.
+ * @param element Root element to check
+ * @param attribute An attribute that is deprecated.
+ * @param attrsInconsistentWith The attribute to be used instead
+ */
+function _warnDeprecatedAttribute(element, deprecatedAttrib, correctAttrib) {
```
marvinchin: What do you think of taking as input an object of `{deprecated: correct}` instead? This will allow us to handle multiple deprecated attributes in one function instead of having to call this multiple times.

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -127,7 +162,7 @@ function _assignPanelId(element) {
 function _parsePopoverAttributes(element) {
   _parseAttributeWithoutOverride(element, 'content', true);
   _parseAttributeWithoutOverride(element, 'header', true);
-  // TODO deprecate title attribute for popovers
+  _warnDeprecatedAttribute(element, 'title', 'header');
```
marvinchin: Let's put all warnings at the top/bottom so it's easier to see.

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -139,6 +174,26 @@ function _parseTooltipAttributes(element) {
...
+/*
+ * Warning for deprecated slot names
+ */
+
+function _warnDeprecatedSlotName(element, deprecatedName, correctName) {
+  if (!(element.children)) {
+    return;
+  }
+  element.children.forEach((c) => {
+    const child = c;
```
marvinchin: Can we just name the parameter `child` instead?

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -139,6 +174,26 @@ function _parseTooltipAttributes(element) {
   _parseAttributeWithoutOverride(element, 'content', true, '_content');
 }
 
+/*
+ * Warning for deprecated slot names
+ */
+
+function _warnDeprecatedSlotName(element, deprecatedName, correctName) {
```
marvinchin: Similar to above, can we make this take an object of deprecated/correct names instead?

src/lib/markbind/src/parsers/componentParser.js
marvinchin: Hmm I think having a separate function for this kind of warnings (`ignoredAttribute` maybe?) might be useful

src/lib/markbind/src/parsers/componentParser.js
nbriannl: We have [no-plusplus](https://eslint.org/docs/rules/no-plusplus) in our linter.

src/lib/markbind/src/parsers/componentParser.js
nbriannl: Sorry, I don't understand. 

`_warnDeprecatedAttribute(element, 'title', 'header');`
becomes
`_warnDeprecatedAttribute(element, {'title', 'header'});` ?
You cannot have a string as a object key. Unless you mean `title` will be the name of the key itself?

How does a method call of multiple deprecated attributes look like for your case?

marvinchin: Consider we have: the attribute `title` to be replaced with `header` and the attribute `some-other-key` to be replaced with `corrected-key`.

I imagine the function call would look something like this:

`_warnDeprecatedAttribute(element, {title: 'header', 'some-other-key': 'corrected-key'});`

https://stackoverflow.com/a/24243895 might give a little more context about why object keys are strings.

src/lib/markbind/src/parsers/componentParser.js
marvinchin: @nbriannl thanks for pointing out that we have some attributes where conflicts are not transitive. Since we don't have a specific use case for dealing with a set of mutually conflicting attributes for now, I'm happy to stick with this approach for now.

src/lib/markbind/src/parsers/componentParser.js
marvinchin: What do you think of using `attrsConflictingWith.forEach` instead then?

src/lib/markbind/src/parsers/componentParser.js
nbriannl: I still feel it's overkill, the case for this is too specific to abstract a function out. i.e. that if there is a X slot, therefore attribute Y has no effect.

yamgent:
@nbriannl resolve the conflicts

PR #1075, opened 2020-03-02
Allow changing parameter properties

### LINKED ISSUE #1065: Override eslint rule to "no-param-reassign": ["error", { "props": false }]
Because of our `no-param-reassign` rule, a lot of redeclaration of a function parameter is done, e.g. in parser.js:

```js
 _preprocess(node, context, config) {
    const element = node; //redecl
    const self = this;
...
}
```

Why this rule is recommended is because reassigning the function parameter can cause the `arguments` object to be weird:
```js
var makePerson = function(favoriteColor, name, age) {
  if (arguments.length < 3) {
    favoriteColor = "green";
    name = arguments[0];
    age = arguments[1];
  }

  return {
    name: name,
    age: age,
    favoriteColor: favoriteColor
  };
};

var person = makePerson("Joe", 18);
console.log(JSON.stringify(person));
// => {"name":"green","age":"green","favoriteColor":"green"}
```
https://spin.atomicobject.com/2011/04/10/javascript-don-t-reassign-your-function-arguments/

This won't happen if we are just changing properties on the parameter!

(Describe your proposed solution here.)

Let's override the eslint rule with `"no-param-reassign": ["error", { "props": false }]`, so that we can change properties on parameters, and avoid the unnecessary redeclarations of parameters.

A nonexhaustive list of places this is done:

https://github.com/MarkBind/markbind/blob/3591bc564bfd6a3a6cb81465b8b51b1ef40ae4f1/src/lib/markbind/src/parser.js#L212

https://github.com/MarkBind/markbind/blob/3591bc564bfd6a3a6cb81465b8b51b1ef40ae4f1/src/lib/markbind/src/parser.js#L452

https://github.com/MarkBind/markbind/blob/3591bc564bfd6a3a6cb81465b8b51b1ef40ae4f1/src/lib/markbind/src/parser.js#L792

and many others in `componentParser.js`

openorclose:

Fixes #1065 

Improve code quality, so that we don't need to unnecessarily redeclare parameters. 

Override the eslint rule to allow for parameter properties to be reassigned.

```js
function setName(o, name) {
o.name = name;
}
```
The above is now allowed.

Ensure that parameters still can't be reassigned.

**Testing instructions:**

```
Allow function param property reassign

Our eslint rule doesn't allow for
properties on parameters to be changed.

This leads to a lot of boilerplate:

function setName(obj, newName) {
    const sameObj = obj;
    sameObj.name = newName;
}

Let's eliminate this by allowing parameter properties
to be reassigned via the eslint rule:

"no-param-reassign": ["error", { "props": false }]

```


ang-zeyu:
LGTM, should we standardise all such names to `node` at the same time?

src/lib/markbind/src/parser.js
```diff
@@ -194,17 +194,23 @@ class Parser {
...
+      const originalSrc = utils.ensurePosix(firstParent.attr('data-included-from') || context);
+      const originalSrcFolder = path.posix.dirname(originalSrc);
+      const fullResourcePath = path.posix.join(originalSrcFolder, relativeResourcePath);
+      const resolvedResourcePath = path.posix.relative(utils.ensurePosix(rootPath), fullResourcePath);
+      return path.posix.join('{{hostBaseUrl}}', resolvedResourcePath);
+    }
+
     $('img, pic, thumbnail').each(function () {
       const elem = $(this);
-      if (elem[0].name === 'thumbnail' && elem.attr('src') === undefined) {
```
ang-zeyu: Reverted 204-207 #1092

PR #1079, opened 2020-03-02
WIP: Adding mermaid integration for support of more diagrams

crphang:
Addresses: #984 

WIP, here's an example.

image

Problems that still needs to be addressed before it is usable:

- [ ] Currently I installed mermaid CLI globally to make it work on my own machine. This should generally not be required, we could assess whether we want to this dependency as part of Markbind, because mermaid requires puppeteer which is quite a big dependency.
- [ ] Allow user to pass some critical parameters like height and width
- [ ] Rebase this on #1047 and have it working inline as well.
- [ ] Tests
- [ ] User guide


damithc:
For the cost/benefit analysis, we can also consider if it brings in enough extra diagrams that plantUML doesn't.

damithc:
> For the cost/benefit analysis, we can also consider if it brings in enough extra diagrams that plantUML doesn't.

I think Mermaid's GitGraph feature https://mermaid.js.org/syntax/gitgraph.html is good enough justification to support it in MarkBind.

tlylt:
Closing to consolidate the more recent discussion in #2052

PR #1087, opened 2020-03-04
Simplify baseUrl resolving process

### LINKED ISSUE #1083: Investigate the baseUrl resolving process as a whole

https://github.com/MarkBind/markbind/pull/1026#issuecomment-583320666

Removing `_rebaseReferenceForStaticIncludes` which is used on static `<includes>` during the preprocess stage seems to have no effect on the generated output.

Inspecting said code and `_rebaseReference` in the main parser, it seems both functions do the same thing, in that it renders `{{baseUrl}}` according to the location ( site ) that the included file is in.

Let's investigate whether we could unify the `baseUrl` resolving process to either the preprocess stage, or `resolveBaseUrl` only, which should make for better code readability and maintainability.

ang-zeyu:

• [x] Other, please explain: Code maintainability / readability

Resolves #1083 

To remove unnecessary complexity in the baseUrl resolving process and refactor some methods to simplify things.

- Added a simpler version of `calculateNewBaseUrls` that directly returns the immediate parent site path instead of `{ parent, relative }`, as is the use case for `calculateNewBaseUrls` in many cases.
- Removed some unnecessary complexity ( `foundBase`, see below ) and refactored `_rebaseReference` to use guard clauses instead of nested ifs.
- Removed `_rebaseReferenceForStaticIncludes`, which is made redundant by `_rebaseReference`.

What is the purpose of `foundBase` in `_rebaseReference`, if any?
The old `_rebaseReference` code calculated the new relative url from `combinedBases[bases[0]]`, where `combinedBases` was the result of `{ ...childrenBase, ...parentBase }`.

The purpose of this seems unclear, and also indeterministic. ( no issues were presented thus far as adding a `console.log` to `combinedBases` reveals that `combinedBases` is always of length `1` at most )

For example, suppose we have
```
<include src="file from main site" />
<include src="file from a sub (x2) site" />
```
that are siblings of one another. After we process the first `include`, `foundBase[root] = ''` would be present.
After processing the second `include` from a doubly nested sub site, we would have `foundBase[subsite1] = 'subsite1 to 2 relative path'`.
Since `Object.keys()` maintains the insertion order, we would wrongly retrieve the relative path of `''` for the second include, causing `{{baseUrl}}` to be resolved wrongly.

Another irregularity was the need for `childrenBase` in `combinedBases`.
When we find the relative base url of the included file, the only concern when resolving said url is `what site/subsite does the included file belong in?`.
It can also cause the same insertion order problem if we repeat the above example, but shift the second include as the child of the first include.

ie.
```
<include src="file from main site"> // in this case the relative url for the parent include is resolved wrongly
  <include src="file from a sub (x2) site" /> // that was a result of the parent include.
</include
```

**Testing instructions:**
`npm run test`, and 2103 site diff should have no diffs.

Simplify baseUrl resolving process

BaseUrls for static includes are rendered once during pre-processing
and another time during resolveBaseUrls.
The rebaseReference subroutine used during resolveBaseUrls also has
some unnecessary complexity.
In addition, in calculateNewBaseUrls, the relative baseUrl for doubly
nested subsites are not resolved correctly.

Let’s remove the rendering during pre-processing, unifying where baseUrl
is resolved for statically included content.
Let’s also refactor the rebaseReference method, simplifying its logic.
At the same time, we can standardise usages of the calculateNewBaseUrls
method to calculateImmediateParentSitePath, and fix the relative baseUrl
returned for doubly nested subsites.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -4,7 +4,6 @@ const fs = require('fs-extra-promise');
 const htmlBeautify = require('js-beautify').html;
 const nunjucks = require('nunjucks');
 const path = require('path');
-const pathIsInside = require('path-is-inside');
 const Promise = require('bluebird');
 const nunjuckUtils = require('./lib/markbind/src/utils/nunjuckUtils');
 
@@ -20,6 +19,7 @@ const logger = require('./util/logger');
 const MarkBind = require('./lib/markbind/src/parser');
 const md = require('./lib/markbind/src/lib/markdown-it');
 const utils = require('./lib/markbind/src/utils');
+const urlUtils = require('./lib/markbind/src/utils/urls');
 
 const CLI_VERSION = require('../package.json').version;
 
@@ -480,8 +480,8 @@ class Page {
     // Set header file as an includedFile
     this.includedFiles.add(headerPath);
     // Map variables
-    const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
-    const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
+    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
+    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
     return `${nunjuckUtils.renderEscaped(nunjucks, headerContent, userDefinedVariables)}\n${pageData}`;
   }
 
@@ -510,8 +510,8 @@ class Page {
     // Set footer file as an includedFile
     this.includedFiles.add(footerPath);
     // Map variables
-    const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
-    const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
+    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
+    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
     return `${pageData}\n${nunjuckUtils.renderEscaped(nunjucks, footerContent, userDefinedVariables)}`;
   }
 
@@ -547,8 +547,8 @@ class Page {
     // Set siteNav file as an includedFile
     this.includedFiles.add(siteNavPath);
     // Map variables
-    const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
-    const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
+    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
+    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
     const siteNavMappedData = nunjuckUtils.renderEscaped(nunjucks, siteNavContent, userDefinedVariables);
     // Convert to HTML
     const siteNavDataSelector = cheerio.load(siteNavMappedData);
@@ -697,8 +697,8 @@ class Page {
       // Set head file as an includedFile
       this.includedFiles.add(headFilePath);
       // Map variables
-      const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap) || '';
-      const userDefinedVariables = this.userDefinedVariablesMap[path.join(this.rootPath, newBaseUrl)];
+      const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
+      const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
       const headFileMappedData = nunjuckUtils.renderEscaped(nunjucks, headFileContent, userDefinedVariables)
         .trim();
       // Split top and bottom contents
@@ -808,8 +808,9 @@ class Page {
         .then((result) => {
           this.content = htmlBeautify(result, Page.htmlBeautifyOptions);
 
-          const newBaseUrl = Page.calculateNewBaseUrl(this.sourcePath, this.rootPath, this.baseUrlMap);
-          const baseUrl = newBaseUrl ? `${this.baseUrl}/${newBaseUrl}` : this.baseUrl;
+          const { relative } = urlUtils.getParentSiteAbsoluteAndRelativePaths(this.sourcePath, this.rootPath,
+                                                                              this.baseUrlMap);
+          const baseUrl = relative ? `${this.baseUrl}/${utils.ensurePosix(relative)}` : this.baseUrl;
           const hostBaseUrl = this.baseUrl;
 
           this.addLayoutFiles();
@@ -1026,7 +1027,6 @@ class Page {
    * @param builtFiles set of files already pre-rendered by another page
    */
   resolveDependency(dependency, builtFiles) {
-    const source = dependency.from;
     const file = dependency.asIfTo;
     return new Promise((resolve, reject) => {
       const resultDir = path.dirname(path.resolve(this.resultPath, path.relative(this.sourcePath, file)));
@@ -1061,8 +1061,6 @@ class Page {
         .then(result => markbinder.resolveBaseUrl(result, {
           baseUrlMap: this.baseUrlMap,
           rootPath: this.rootPath,
-          isDynamic: true,
-          dynamicSource: source,
         }))
         .then(result => fs.outputFileAsync(tempPath, result))
         .then(() => markbinder.renderFile(tempPath, {
@@ -1075,8 +1073,9 @@ class Page {
         .then(result => markbinder.processDynamicResources(file, result))
         .then((result) => {
           // resolve the site base url here
-          const newBaseUrl = Page.calculateNewBaseUrl(file, this.rootPath, this.baseUrlMap);
-          const baseUrl = newBaseUrl ? `${this.baseUrl}/${newBaseUrl}` : this.baseUrl;
+          const { relative } = urlUtils.getParentSiteAbsoluteAndRelativePaths(file, this.rootPath,
+                                                                              this.baseUrlMap);
+          const baseUrl = relative ? `${this.baseUrl}/${utils.ensurePosix(relative)}` : this.baseUrl;
           const hostBaseUrl = this.baseUrl;
           const content = nunjucks.renderString(result, {
             baseUrl,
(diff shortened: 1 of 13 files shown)


src/Page.js
```diff
@@ -1059,8 +1058,6 @@ class Page {
         .then(result => markbinder.resolveBaseUrl(result, {
           baseUrlMap: this.baseUrlMap,
           rootPath: this.rootPath,
-          isDynamic: true,
```
ang-zeyu: these variables in `resolveDependency` are not necessary, as in the `resolveBaseUrl` process, `attribs[ATTRIB_CWF] = context.cwf` is the same assigned to `parser.dynamicIncludeSrc.push({ from: context.cwf, ... })`

src/lib/markbind/src/parser.js
```diff
@@ -552,30 +553,17 @@ class Parser {
...
           reject(error);
           return;
         }
-        const nodes = dom.map((d) => {
-          const node = d;
-          const childrenBase = {};
-          if (this.isDynamic) {
-            // Change CWF for each top level element
-            if (node.attribs) {
-              node.attribs[ATTRIB_CWF] = this.dynamicSource;
```
ang-zeyu: see the comment for `resolveDependency`

src/lib/markbind/src/parser.js
```diff
@@ -588,54 +576,53 @@ class Parser {
     });
   }
 
-  _rebaseReference(node, foundBase) {
+  /**
+   * Pre-renders the baseUrl of the provided node and its children according to
+   * the site they belong to, such that the final call to render baseUrl correctly
+   * resolves baseUrl according to the said site.
+   */
+  _rebaseReference(node) {
```
ang-zeyu: Two changes here essentially:
- replace nested ifs with guard clauses
- removal of `foundBase` ( see the explanation in the OP )

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -181,28 +181,6 @@ function _preProcessPanel(node, context, config, parser) {
  * Includes
  */
 
-
-function _rebaseReferenceForStaticIncludes(pageData, element, config) {
```
ang-zeyu: See #1083

src/lib/markbind/src/utils/urls.js
```diff
@@ -44,12 +44,42 @@ function calculateNewBaseUrls(filePath, root, lookUp) {
   return calculate(filePath, undefined);
 }
 
+/**
+ * Returns the absolute path of of the immediate parent site of the specified file path.
+ */
+function calculateImmediateParentSitePath(filePath, root, lookUp) {
```
ang-zeyu: Simpler version of `calculateNewBaseUrls` that just returns the immediate parent site path, which is what's needed for the majority of use cases

src/lib/markbind/src/utils/urls.js
```diff
@@ -44,12 +44,42 @@ function calculateNewBaseUrls(filePath, root, lookUp) {
...
+      return file;
+    }
+
+    const parent = path.dirname(file);
+    return lookUp.has(parent)
+      ? parent
+      : calculate(parent);
+  }
+
+  return calculate(filePath, undefined);
```
crphang: Calculate defined here takes only 1 parameter.

src/lib/markbind/src/parser.js
```diff
@@ -588,54 +576,53 @@ class Parser {
...
     delete element.attribs[ATTRIB_CWF];
+    delete element.attribs[ATTRIB_INCLUDE_PATH];
+
+    // Skip rebasing for non-includes
+    if (!includedSourceFile || !element.children) {
+      return element;
+    }
+
+    // Where the children are included
+    const newBase = urlUtils.calculateNewBaseUrls(includedSourceFile, this.rootPath, this.baseUrlMap);
```
crphang: Loving the review that you did previously, really helped a lot. Will be good if you could help explain this part as well.

For newBase, is it possible to use `calculateImmediateParentSitePath`?

src/lib/markbind/src/parser.js
```diff
@@ -588,54 +576,53 @@ class Parser {
...
+    // Where the include was preprocessed
+    const currentBase = urlUtils.calculateImmediateParentSitePath(includeSourceFile,
+                                                                  this.rootPath, this.baseUrlMap);
+
+    // Only re-render if include src and content are from different sites
+    if (currentBase !== path.resolve(newBase.parent, newBase.relative)) {
```
crphang: Then we can have currentBase !== newBase

ang-zeyu: This is one of the use cases for `calculateNewBaseUrls` in particular
`relative` is needed for the resolving the new relative baseurl
`baseUrl: {{hostBaseUrl}}/${newBase.relative}`

src/lib/markbind/src/parser.js
ang-zeyu: Made the comments slightly clearer

ang-zeyu:
> Then we can have currentBase !== newBase

I've removed `calculateNewBaseUrls`, but moved it to `calculateImmediateParentSitePath` as an option as its still necessary sometimes to have the relative base url.

Also fixed a bug with doubly nested subsites that existed before with baseurls ( `baseurl` simply resolves wrongly for it )
`Page.calculateNewBaseUrl` is also standardised to use `calculateImmediateParentSitePath` now instead.

src/lib/markbind/src/utils/urls.js
```diff
@@ -8,48 +8,59 @@ const {
...
  * @param filePath The absolute file path to look up that is nested inside the root directory
  * @param root The base directory from which to terminate the look up
  * @param lookUp The set of urls representing the sites' base directories
- * @return An object containing
- *         1. The parent site's path of that the immediate parent site of the specified filePath,
- *            or the root site's path if there is none
- *         2. The relative path from (1) to the immediate parent site of the specified filePath
+ * @param needRelative Whether the relative file path from the root site to the immediate parent site
+ *                     should be returned as well. Defaults to false if unspecified
+ * @return Either
```
marvinchin: Is it possible/inexpensive enough for us to just calculate and return both the absolute and relative path every time? I think having to deal with two possible return types make this function more complicated, and can easily lead to errors especially without a type system in place.

src/lib/markbind/src/parser.js
```diff
@@ -581,53 +570,52 @@ class Parser {
...
-      }
-      delete node.attribs[ATTRIB_INCLUDE_PATH];
-    }
+
+    // Rebase children elements
+    node.children.forEach(el => this._rebaseReference(el));
+
+    const includeSourceFile = node.attribs[ATTRIB_CWF];
+    const includedSourceFile = node.attribs[ATTRIB_INCLUDE_PATH];
     delete node.attribs[ATTRIB_CWF];
```
marvinchin: Just wondering - should we delete these attributes here? It feels a little odd since they were added in `_preprocess` (at least, I think), and it isn't immediately intuitive that these attributes would be deleted in this function.

If it is necessary for us to delete it here, could we leave a comment explaining why we need to do this? And maybe add a comment in `_preprocess` to notify that this variable will be deleted here.

test/functional/test_site/expected/markbind/css/markbind.css
```diff
@@ -124,7 +124,7 @@ code.hljs:hover {
 header.header-fixed {
     position: fixed;
     width: 100%;
-    z-index: 9999;
+    z-index: 1000;
```
marvinchin: Hmm this diff doesn't seem to be related to the changes in this PR. 

Seems to be related to https://github.com/MarkBind/markbind/pull/1070. Let's fix this in another PR? Also, it seems like CSS diffs aren't being checked by our tests 😅

src/lib/markbind/src/parser.js
```diff
@@ -581,53 +570,52 @@ class Parser {
...
-    if (node.attribs[ATTRIB_INCLUDE_PATH]) {
-      const filePath = node.attribs[ATTRIB_INCLUDE_PATH];
-      let newBaseUrl = urlUtils.calculateNewBaseUrls(filePath, this.rootPath, this.baseUrlMap);
-      if (newBaseUrl) {
-        const { relative, parent } = newBaseUrl;
-        // eslint-disable-next-line no-param-reassign
-        foundBase[parent] = relative;
-      }
-      // override with parent's base
-      const combinedBases = { ...childrenBase, ...foundBase };
```
marvinchin: This seems to be the commit that added the `foundBase` change:
https://github.com/MarkBind/markbind/commit/0faa1b731a9d4e4675b4f88d36f411f54f2a7580

src/lib/markbind/src/utils/urls.js
ang-zeyu: I went with two possible return types to avoid e.g. `{ absolute: parentSite } = urlUtils.calculate...` all over. I merged it into one now though, will revert if preferred!

src/lib/markbind/src/parser.js
ang-zeyu: The original function already deletes both

ang-zeyu: It's not necessary to the best of my knowledge, but not having `<div cwf="...">` all over the output is a nice touch

test/functional/test_site/expected/markbind/css/markbind.css
ang-zeyu: 125ac543da90753aa13f8da9a7caaffa34cf617a, was indeed #1070's test file updates

> Seems to be related to #1070. Let's fix this in another PR? Also, it seems like CSS diffs aren't being checked by our tests 😅

I checked the test script briefly, it seems that only `siteData.json` and `.html` files are diffed. 
The likely original reason was to avoid diffing large library files ( eg `bootstrap.min.css` ); Perhaps a better solution would be to exclude these files explicitly. Will open as an issue!

src/lib/markbind/src/parser.js
ang-zeyu: Thanks for this! Unfortunately the PR dosen't really reveal the workings of `combinedBases`, but

> What is the rationale for this request?
> When including files belonging to an inner website, the computation of baseUrl is wrong if the depth of the inclusion is only 1.

Going by this, it should be sufficient to resolve the baseurl only when the include's file and included file belongs to different sites. The likely cause of the issue was wrongly resolving against `ATTRIB_CWF` instead of `ATTRIB_INCLUDE_PATH`, and perhaps the above mentioned irregularities with `foundBase`.

I was looking at https://github.com/MarkBind/markbind/commit/64b2ff6d93a457869fe965d896995dd7d1d8083e previously, where `foundBase` was included with the function right from the start. 
It dosen't reveal much as well unfortunately, perhaps @Gisonrg could comment!

src/lib/markbind/src/parser.js
marvinchin: Ah noted I didn't notice that these outputs would be included in the output!

src/lib/markbind/src/utils/urls.js
marvinchin: I personally would lean towards avoiding hard to spot errors over having slightly shorter code, but I'm open to the input of others.

What do you think about splitting them up into two functions? Would help make this more explicit while reducing unnecessary destructuring at the call site? (i.e. `calculateAbsoluteParentSitePath` and `calculateAbsoluteAndRelativeParentSitePaths`)

ang-zeyu: Sorry for the late reply! I've changed it back to two functions

PR #1088, opened 2020-03-04
Fix baseUrl not working in include src attribute

### LINKED ISSUE #928: {{baseUrl}} in boilerplates is not behaving as an absolute URL
v2.5.2

Context: `{{ baseUrl }}` should resolve to the root of the site, which gives us a way to specify an absolute URL within the site.

Expectation: In a boilerplate file, a URL is resolved relative to where the file is included. However, sometimes one needs to specify a fixed URL inside a boilerplate. Suppose the following two lines are in a boilerplate file.
```html
<include src="index.md" />
<include src="{{ baseUrl }}/index.md" />
``` 
The first one should resolve to different files based on where the boilerplate is used whereas the second one should resolve to the same file no matter where the boilerplate is used.

Problem: both URLs seem to vary based on where the boilerplate is included.

ang-zeyu:

• [x] Bug fix

Fixes #928 
~Requires #1087 ( removal of `_rebaseReferenceForStaticIncludes` ( neccessary ) and adding `getParentSiteAbsolutePath` ( standardisation ) )~ merged

To fix the case that the user wants to use `{{baseUrl}}` inside the `src` attribute of a `<include/panel>`.

- When resolving the source file path of the included path, in addition to checking whether the source file is url, we check if the `src` has a `{{\s*baseUrl\s*}}[/\\]` regex.
If so, we recalculate the include source file path using the baseUrl of the current working file.
- Added some functional tests to check for such cases
- Added `Markbind.unwrapIncludeSrc` to the dynamic include build chain since it was missed before and my functional tests require it ( otherwise `data-included-from` would have my system's absolute file path, causing tests to fail )

```js
...
if (baseUrlRegex.test(includePath)) {
  // The baseUrl has not been resolved during pre-processing, but we need the source file path
  const parentSitePath = urlUtils.getParentSiteAbsolutePath(context.cwf, config.rootPath, config.baseUrlMap);
  filePath = path.resolve(parentSitePath, includePath.replace(baseUrlRegex, ''));
} else {
...
```

na

**Testing instructions:**
- `npm run test` should pass
- 2103 site should differ for `**._include_.html` files as `Markbind.unwrapIncludeSrc` was added. ( the wrapper div and spans with `data-included-from` should be deleted. )

Fix baseUrl not working in include src attribute

While pre-processing includes or panels, we have not yet resolved
the baseUrl of the src attribute.
This causes markbind to fail in finding the source file, since baseUrl
remains in the src attribute.

Let’s resolve the baseUrl for such cases first, allowing the user to use
the baseUrl attribute in include and panel src attributes.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -1060,6 +1060,7 @@ class Page {
         .then(result => this.postRender(result))
         .then(result => this.collectPluginsAssets(result))
         .then(result => markbinder.processDynamicResources(file, result))
+        .then(result => MarkBind.unwrapIncludeSrc(result))
         .then((result) => {
           // resolve the site base url here
           const { relative } = urlUtils.getParentSiteAbsoluteAndRelativePaths(file, this.rootPath,
--- a/src/lib/markbind/src/preprocessors/componentPreprocessor.js
+++ b/src/lib/markbind/src/preprocessors/componentPreprocessor.js
@@ -108,9 +108,23 @@ function _getSrcFlagsAndFilePaths(element, context, config) {
   // We do this even if the src is not a url to get the hash, if any
   const includeSrc = url.parse(element.attribs.src);
 
-  const filePath = isUrl
-    ? element.attribs.src
-    : path.resolve(path.dirname(context.cwf), decodeURIComponent(includeSrc.path));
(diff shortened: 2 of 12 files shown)


ang-zeyu:
Rebased this on top of #1087 instead of making the same required changes instead; Hopefully its easier to review this way!

PR #1089, opened 2020-03-04
bugfix: Fix hrefs for headings with angular brackets

### LINKED ISSUE #1076: Improper linking of headings that have part of text enclosed within "<>"

* **MarkBind Version:** 2.11.0

There are instances where part of the header text that is enclosed within "<>". 
Referring to such headings using hrefs is not straightforward.
For example, the heading - 
# Specifying Variables in an `<include>`
will have to be linked using the following href - `#specifying-variables-in-an-and-lt-include-and-gt`

The href should be `#specifying-variables-in-an-include`

The href was `#specifying-variables-in-an-and-lt-include-and-gt`, as the string passed as input to slugify is `Specifying variables in an &lt; include &gt;`

There are a couple of instances in the user guide where linking [such headers](https://markbind.org/userGuide/reusingContents.html#specifying-variables-in-an-include) is broken ([this](https://markbind.org/userGuide/reusingContents.html#specifying-variables-in-an-and-lt-include-and-gt) is the correct link).

**Suggested Fix(es)**
For this particular case, I have tried removing the "[-]?and-lt"  and "[-]?and-gt" substrings from the slugified text like so - 
```js
headerId = headerId.replace(/[-]?and-lt/g, '');
headerId = headerId.replace(/[-]?and-gt/g, '');
```

yash-chowdhary:

• [x] Bug fix

Fixes #1076 

Presence of angular brackets in headings creates undesirable href links.

Removed occurrences of the strings `'and lt'` and `'and gt'` that are passed to `slugify`. 

~~customReplacements: [~~
      ~~['and lt', ''],~~
      ~~['and gt', ''],~~
    ~~],~~
~~});~~

New fix: 

```js
let cleanedContent = textContent.replace(/&lt;/g, '');
cleanedContent = cleanedContent.replace(/&gt;/g, '');
```

n/a

```
The hrefs of headings with angular brackets are incorrect because
markdown-it escapes the characters to "&lt;" and "&gt;" and the
subsequent slugify operation converts it to "and-lt" or "and-gt".

Let's fix this by removing the occurrences of the strings 
"&lt;" and "&gt;" before the slugify operation
```

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -263,7 +263,13 @@ class Parser {
 
     if ((/^h[1-6]$/).test(node.name) && !node.attribs.id) {
       const textContent = utils.getTextContent(node);
-      const slugifiedHeading = slugify(textContent, { decamelize: false });
+      const slugifiedHeading = slugify(textContent, {
+        decamelize: false,
+        customReplacements: [
+          ['and lt', ''],
+          ['and gt', ''],
+        ],
+      });
 
       let headerId = slugifiedHeading;
       const { headerIdMap } = config;
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -263,13 +263,9 @@ class Parser {
 
     if ((/^h[1-6]$/).test(node.name) && !node.attribs.id) {
       const textContent = utils.getTextContent(node);
-      const slugifiedHeading = slugify(textContent, {
-        decamelize: false,
-        customReplacements: [
-          ['and lt', ''],
-          ['and gt', ''],
-        ],
-      });
+      let cleanedContent = textContent.replace(/&lt;/g, '');
+      cleanedContent = cleanedContent.replace(/&gt;/g, '');
+      const slugifiedHeading = slugify(cleanedContent, { decamelize: false });
 
       let headerId = slugifiedHeading;
       const { headerIdMap } = config;
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -263,8 +263,8 @@ class Parser {
 
     if ((/^h[1-6]$/).test(node.name) && !node.attribs.id) {
       const textContent = utils.getTextContent(node);
-      let cleanedContent = textContent.replace(/&lt;/g, '');
-      cleanedContent = cleanedContent.replace(/&gt;/g, '');
+      // remove the '&lt;' and '&gt;' symbols that markdown-it uses to escape '<' and '>'
+      const cleanedContent = textContent.replace(/&lt;|&gt;/g, '');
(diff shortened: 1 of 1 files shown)


yash-chowdhary:
@ang-zeyu could I get your review on this?

ang-zeyu:
Realised that the con of this fix is that you cant type `and lt` 😧 

`headerId = headerId.replace(/[-]?and-lt/g, '');`
Perhaps your original fix, but placed before the `slugify` call and adapted to only handle `&lt/gt;` would be better?

yash-chowdhary:
> Realised that the con of this fix is that you cant type `and lt`

Good catch 😅 

> Perhaps your original fix, but placed before the `slugify` call and adapted to only handle `&lt/gt;` would be better?

Yup, I've now changed it to use the previously suggested fix. This will remove all occurrences of `&lt;` and `&gt;` from the headings after `markdown-it`'s parsing.

src/lib/markbind/src/parser.js
```diff
@@ -263,7 +263,9 @@ class Parser {
 
     if ((/^h[1-6]$/).test(node.name) && !node.attribs.id) {
       const textContent = utils.getTextContent(node);
-      const slugifiedHeading = slugify(textContent, { decamelize: false });
+      let cleanedContent = textContent.replace(/&lt;/g, '');
+      cleanedContent = cleanedContent.replace(/&gt;/g, '');
```
ang-zeyu: >Yup, I've now changed it to use the previously suggested fix.

👍 

Why not combine it into one though `&lt;|&gt;` :o
A comment here explaining why we clean `&lt;&gt;` in particular might be nice too

nbriannl:
@yash-chowdhary i think you may see if this PR is ready to be merged according to https://markbind.org/devdocs/devGuide/maintainerGuide.html

yash-chowdhary:
> @yash-chowdhary i think you may see if this PR is ready to be merged according to https://markbind.org/devdocs/devGuide/maintainerGuide.html

I suppose that this PR is trivial enough to be merged after 1 approval

nbriannl:
sorry for deleting comment.

> > @yash-chowdhary i think you may see if this PR is ready to be merged according to https://markbind.org/devdocs/devGuide/maintainerGuide.html
> 
> I suppose that this PR is trivial enough to be merged after 1 approval

yeah, i don't really know if anything will regress after this. No clean unit test for this because these functionality lies within  `_parse` function.

If anything, the fact that this link
https://deploy-preview-1089--markbind-master.netlify.com/userGuide/reusingContents.html#specifying-variables-in-an-include
now directs the page correctly means that your implementation is correct. 😄

yash-chowdhary:
> If anything, the fact that this link
> https://deploy-preview-1089--markbind-master.netlify.com/userGuide/reusingContents.html#specifying-variables-in-an-include
> now directs the page correctly means that your implementation is correct.

Yup, it can be merged. 🙂

PR #1092, opened 2020-03-07
Fix refactor to processDynamicResources

ang-zeyu:

• [x] Bug fix

Apologies, missed this out while reviewing #1075.
See https://github.com/MarkBind/markbind/pull/1075#discussion_r389262255 for original code

CI passes on travis because `ensurePosix` only runs if `path.sep !== /` ( non-unix ).
Otherwise `undefined` is passed to it which causes tests to fail. This is one such case #1040 could be really useful

Reintroduced a `undefined` check for the the element's `src` before passing it to `utils.ensurePosix` to prevent said error.

Fix refactor to processDynamicResources

### DIFF
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -211,11 +211,10 @@ class Parser {
 
     $('img, pic, thumbnail').each(function () {
       const elem = $(this);
-      const resourcePath = utils.ensurePosix(elem.attr('src'));
-      if (resourcePath === undefined || resourcePath === '') {
-        // Found empty img/pic resource in resourcePath
+      if (!elem.attr('src')) {
         return;
       }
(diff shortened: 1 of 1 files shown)


src/lib/markbind/src/parser.js
```diff
@@ -211,6 +211,9 @@ class Parser {
 
     $('img, pic, thumbnail').each(function () {
       const elem = $(this);
+      if (elem[0].name === 'thumbnail' && elem.attr('src') === undefined) {
```
openorclose:
```suggestion
      if (elem.attr('src') === undefined) {
```

Should we do this then?

Even though pics and imgs must have the src attribute, the author could still forget to include them.

ang-zeyu: Updated

Issue #1099, opened 2020-03-08
Special Tags have parsing issues when closed with <tag /> instead of <tag></tag>


crphang:
**Tell us about your environment**
Latest Master

* **MarkBind Version:**
2.11.0
**What did you do? Please include the actual source code causing the issue.**

For any plugin, register a special tag:
```
getSpecialTags: () => ['testtag'],
```

In any markbind document, we could write normally as:

```
<testtag />

<testtag>
Some document
</testtag>
```

**What did you expect to happen?**

When running markbind build: The produced result remain as:

```
<testtag />

<testtag>
Some document
</testtag>
```
regardless of whether it is a special tag.

**What actually happened? Please include the actual, raw output.**

The second testtag's closing bracket is removed, leading to some parsing error downstream from cheerio.

![image](https://user-images.githubusercontent.com/4986717/76160072-ce5aca80-6161-11ea-924d-2d67581420bf.png)

** More examples **

See #1100:
In the case of using puml as a special tag, existing usages of having:
```
<puml src="a.puml" />

<puml src="b.puml" />

<puml src="c.puml" />
```

would not work as only the other puml tags will not be parsed accordingly from cheerio.

ang-zeyu:
Hmm, this is actually the standard behaviour for `script/style` tags ( spec does not allow `<script ... />` ).

I could either
- allow this for script / style tags as well ( less performance heavy, but it would break once in the browser )
- or include more code to allow special self closing tags for the injected tags only.
- or necessitate opening/closing tags for special tags, e.g. inline puml would have another `<puml-inline>` tag

Which should we go with? @crphang

crphang:
Thanks for quick response @ang-zeyu  :tada: 

Option (2) seems like the best. Would be great if you already have an approach for it.

Option (3) can be done, but I would suppose it could get awkward when we would have to handle inline and src differently for all other plugins going forward

ang-zeyu:
I went with option 2 #1101; The approach also turned out to be less problematic than I expected! ( though a little hard-coding was involved ( but documented ))

PR #1100, opened 2020-03-08
Inline puml

### LINKED ISSUE #912: Allow inline plantUML code when possible

crphang:

• [X] New feature

Resolves #912, continuation of #968.

Added inline PUML support which allows user to write PUML inline.

1. Registered `puml` as a special tag which allows markdown-it to ignore the content (escaping it accordingly).
1. Added additional plugin utils that extracts escape content or from source accordingly.
1. Generates md5 file name when user does not provide it, which allows the file name to be consistent for the same content and prevents regeneration if already exists.

The following tag would produce puml documents.
```
<puml>
@startuml
object Object01
object Object02
object Object03
object Object04
object Object05
object Object06
object Object07
object Object08

object user {
  name = "Dummy"
  id = 123
}

Object01 <|-- Object02
Object03 *-- Object04
Object05 o-- "4" Object06
Object07 .. Object08 : some labels
@enduml

</puml>
```

This feature relies on special tags and would require #1099 to be resolved enabling having multiple puml tags:

``` 
<puml src="a.png" /> 

<puml> 
...inline puml
</puml>
```

Switching PUML as a special tag now would break existing behavior. To illustrate this: I have switched all puml tags in tests to:

```diff
-<puml src="a.png" />
+<puml src="a.png"></puml>

-<puml src="b.png" />
+<puml src="b.png"></puml>
```


crphang:
@ang-zeyu could you review this?

This is a much simplified version of previous PR. Tests has worked after I rebased on #1099 :tada:.

src/util/pluginUtil.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const hashedContent = cryptoJS.MD5(content).toString();
+    return `${hashedContent}.png`;
+  },
+
+  /**
+   * Returns the string content of the plugin.
+   */
+  getPluginContent: ($, element, cwf, config) => {
```
ang-zeyu: Could we expand the scope of this PR a little?
- I realised `cwf` will always be available after refactoring `componentPreprocessor`, let's remove `config` from the entire plantuml plugin, and consequently the entire `config` ( I don't see any other plugins using this, but you could double confirm ). This should be a good step for https://github.com/MarkBind/markbind/pull/714#discussion_r258503158 ( which is no longer necessary with the prior refactors ).

- We currently don't have any plugins that use `config` save for this anymore, so let's repurpose it for plugin utils, as calling `const pluginUtil = require('../../util/pluginUtil');` on every plugin is dependent on the plugin's location. E.g. if the plugin is in `markbind`, one would have to know `pluginUtil`'s location relative to the parent module or in the file system.
  - Let's then expose these pluginUtils. Since we only have one plugin using this now, we can leave it undocumented for a while until we have more and the implementation / requirements becomes more final.

I think its a good step to rectify the solution that was used in https://github.com/MarkBind/markbind/pull/714#discussion_r258503158, and since `pluginUtils` as I see it here isn't parser/page dependent, I think exposing it eventually after a few more rounds of plugin development would be a good option. Might be good to have @marvinchin 's opinion since he was involved though 😁.

crphang: Hey, could you elaborate more on point 2? Not too sure I am following well

ang-zeyu: I'm thinking of
`preRender: (content, pluginContext, frontMatter, useful safe interfaces (pluginUtil) that we want to expose to plugins) => ...` 
instead of depending on
`preRender: (content, pluginContext, frontMatter, secret unsafe details) => ...` 
or having to rely on `require('../../util/pluginUtil')` where the path may differ from plugin to plugin

crphang: I see. However, this introduces a breaking change if we remove config from pre/post-render, especially if there are users who have defined their own plugins.

ang-zeyu: It was intentionally undocumented because it exposed too much implementation details https://github.com/MarkBind/markbind/pull/714#discussion_r258418526 ( the above comment ); should be fine here

src/util/pluginUtil.js
```diff
@@ -0,0 +1,54 @@
...
+   * Return based on given name if provided, or it will be based on src.
+   * If both are not provided, a md5 generated hash is provided for consistency.
+   */
+  getFileNameForPlugin: (tagAttribs, content) => {
+    if (tagAttribs.name !== undefined) {
+      return `${fsUtil.removeExtension(tagAttribs.name)}.png`;
+    }
+
+    if (tagAttribs.src !== undefined) {
+      const fileName = fsUtil.removeExtension(tagAttribs.src);
```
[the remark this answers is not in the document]
crphang: I believe we want to have it as `fsUtil.removeExtension`. For example, if the `src='diagrams/activity.puml`, this way the output fileName will be `src='diagrams/activity.png'`. 

With `path.basename(tagAttribs.src)`, it becomes `activity.png`

crphang:
@ang-zeyu thanks for your review! Thanks for pointing out that cwf always exist, I refactored slightly to make the code simpler. I also added the changes to user guide.

Currently I would like to scope this PR smaller! Wanted to point out that PlantUML plugin is currently still using `config` for `resultPath` and it is quite a common use case. In the direction that you suggested, I believe we could initialize those properties in `pluginUtil` and expose them as functions within `pluginUtil` instead.

src/util/pluginUtil.js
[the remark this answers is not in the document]
ang-zeyu: Hmm I meant `fsUtil.removeExtension(path.basename(tagAttribs.src))`, but I realised omitting basename keeps the output file in the same directory as the input file.
In that case how about changing `fileName` -> `filePath`?

src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -6,92 +6,73 @@
...
 } = require('../../constants');
 
 // Tracks diagrams that have already been processed
 const processedDiagrams = new Set();
 
 /**
- * Generates diagram and replaces src attribute of puml tag
- * @param src src attribute of the puml tag
- * @param cwf original file that contains the puml tag, we resolve relative src to this file
+ * Generates diagram and returns the file name of the diagram
```
ang-zeyu: `file path`

src/util/pluginUtil.js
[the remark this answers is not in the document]
ang-zeyu: To keep the behaviour when ` src` is specified consistent, if a `name` is also specified,
should we instead return the file path that is the result of replacing the basename of `src` with `name`?

src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -103,14 +84,21 @@ module.exports = {
     $('puml').each((i, tag) => {
       // eslint-disable-next-line no-param-reassign
       tag.name = 'pic';
-      const { src, cwf } = tag.attribs;
+      const { cwf } = tag.attribs;
+      const pumlContent = pluginUtil.getPluginContent($, tag, cwf);
+
+      const fileName = pluginUtil.getFileNameForPlugin(tag.attribs, pumlContent);
```
ang-zeyu: `filePath` correspondingly

crphang:
@ang-zeyu thanks for your review! Indeed filePath would've been more appropriate, updated accordingly. :+1: 

> should we instead return the file path that is the result of replacing the basename of src with name?

I thought a little about this. However, I believe letting `name` be the source of truth is better considering both inline and referenced puml inputs. For example:

```html
<puml name="output/activity">
activity diagram
</puml>

<puml name="output/object" src="input/object.puml" /> 
```

Here, an author would like both output to be in `output` directory for inline and referenced PUML, instead of having no such option if `src` is specified.

src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -103,14 +84,21 @@ module.exports = {
     $('puml').each((i, tag) => {
       // eslint-disable-next-line no-param-reassign
       tag.name = 'pic';
-      const { src, cwf } = tag.attribs;
+      const { cwf } = tag.attribs;
+      const pumlContent = pluginUtil.getPluginContent($, tag, cwf);
+
+      const filePath = pluginUtil.getFilePathForPlugin(tag.attribs, pumlContent);
+
       // eslint-disable-next-line no-param-reassign
```
ang-zeyu: I think this and the above one could this be deleted - there was a PR a while back that changed this

nbriannl:
Hi, thanks for the changes. Are you able to ping for reviewers in this repo?

crphang:
@marvinchin, could you help to review this? Thanks :bowing_man: !

nbriannl:
Hi Chun Rong @crphang 

It seems that updating tests on Windows `npm run updatetestwin` produces a different set of expected results.

image

damithc:
Seems to be working in production. Good work @crphang and reviewers.

PR #1101, opened 2020-03-08
Allow special tags to be self-closing

### LINKED ISSUE #1099: Special Tags have parsing issues when closed with <tag /> instead of <tag></tag>

ang-zeyu:

• [x] Documentation update

Resolves #1099

Allow injected special tags to be self-closing as well, while preserving the contrary for `script/style` tags ( according to html5 spec ).

- Simple 2 line patch to self-closing tag state handler to allow said behaviour
```js
// self-closing tag state handler
...
this._special = SPECIAL_NONE;
...
```
- Update tests to test for self-closing injected special tags as well
- Small documentation update to reflect this information

na

**Testing instructions:**
- `npm run test`

Allow special tags to be self-closing

Unlike the html script and style special tags, plugin authors may want
injected special tags to be able to be self-closing as well.
This can help unify the syntax used in the plugin, which makes for
better author usability of the plugin.

Let’s allow injected special tags to be self-closing.
Let’s also do so for the script and style tags, which would be
expanded to their normal form after rendering.

### DIFF
--- a/src/lib/markbind/src/patches/htmlparser2.js
+++ b/src/lib/markbind/src/patches/htmlparser2.js
@@ -262,6 +262,29 @@ Tokenizer.prototype._stateBeforeSpecial = function(c) {
 };
 
 /**
+ * Patched self closing tag state handler that removes the special state
+ * if the special tag was self-closed.
+ */
+Tokenizer.prototype._stateInSelfClosingTag = function(c) {
+	if (c === ">") {
+		this._cbs.onselfclosingtag();
+		this._state = TEXT;
+		this._sectionStart = this._index + 1;
+		/*
+		 Allow all special tags to be self-closed.
+		 Script and style tags are also allowed to be self-closed,
+		 which breaks from the default html spec-compliant of behaviour of htmlparser2.
+		 We allow this as such tags would be expanded upon re-rendering the html anyway.
+		 ie. '<script ... />' would end up as '<script ...></script>'
+		 */
+		this._special = SPECIAL_NONE;
+	} else if (!whitespace(c)) {
+		this._state = BEFORE_ATTRIBUTE_NAME;
+		this._index--;
+	}
+};
+
+/**
  * Processes the _special flag and _nextSpecialTagMatchIndex state variable,
  * returning a flag indicating whether the current special tag has finished matching or not.
  */
--- a/test/functional/test_site_special_tags/_markbind/plugins/testSpecialTag.js
+++ b/test/functional/test_site_special_tags/_markbind/plugins/testSpecialTag.js
@@ -28,7 +28,7 @@ function preRender(content) {
   without interference from other dependencies
 */
 function postRender(content) {
-  const $ = cheerio.load(content);
+  const $ = cheerio.load(content, { xmlMode: false });
   const escapedNunjucks = $('mustache');
   escapedNunjucks.each((index, element) => {
     const unwrappedText = $(element).text();
@@ -44,5 +44,5 @@ function postRender(content) {
 module.exports = {
   preRender,
   postRender,
-  getSpecialTags: () => ['testtag', 'mustache'],
+  getSpecialTags: () => ['testtag', 'testselfclosingtag', 'mustache'],
 };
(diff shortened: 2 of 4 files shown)


src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -261,6 +261,28 @@ Tokenizer.prototype._stateBeforeSpecial = function(c) {
...
+ * Removes the special state if the special tag was self-closed.
+ */
+Tokenizer.prototype._stateInSelfClosingTag = function(c) {
+	if (c === ">") {
+		this._cbs.onselfclosingtag();
+		this._state = TEXT;
+		this._sectionStart = this._index + 1;
+		if (this._special > 2) {
+			/*
+			 We intentionally fail if the special tag in question is
```
crphang: Minor: Would really help to document/comment what "fail" imply when you set `this._special = SPECIAL_NONE`.

ang-zeyu: Updated

ang-zeyu:
> Since there's no such thing as self closing tags in the HTML specs, should we just allow script and style tags to be self closing too?
> 
> In the generated HTML they would all be expanded out to the valid `<script></script>` and `<special></special>` forms anyway.

I'm open to either; What do you think? @crphang

crphang:
> > Since there's no such thing as self closing tags in the HTML specs, should we just allow script and style tags to be self closing too?
> > In the generated HTML they would all be expanded out to the valid `<script></script>` and `<special></special>` forms anyway.
> 
> I'm open to either; What do you think? @crphang

Looks fine to me, not 100% sure if there will be no issues, but seems like it will make the code simpler here

ang-zeyu:
Updated to allow `script / style` tags as well;
It should be side effect free as htmlparser2 dosen't do special processing for script/style tags outside of its tokenizer

crphang:
@marvinchin could you help with the review of this? 🙇

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -261,6 +261,22 @@ Tokenizer.prototype._stateBeforeSpecial = function(c) {
 	this._special = result + 1;
 };
 
+/**
+ * Removes the special state if the special tag was self-closed.
```
marvinchin: Can we explicitly state that this is an altered version of the `_stateInSelfClosingTag` defined in `htmlparser`? I think it might help other developers new to this part of the codebase understand what's going on better.

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -261,6 +261,22 @@ Tokenizer.prototype._stateBeforeSpecial = function(c) {
...
+ */
+Tokenizer.prototype._stateInSelfClosingTag = function(c) {
+	if (c === ">") {
+		this._cbs.onselfclosingtag();
+		this._state = TEXT;
+		this._sectionStart = this._index + 1;
+		// Allow all special tags (including script & style tags) to be self-closed
```
marvinchin: Let's explain why we allow this despite the spec not allowing it. I think just stating the fact that the original htmlparser allows it is sufficient.

ang-zeyu: To clarify, did you mean the `original htmlparser disallows it` ( that is the case, we're breaking away from the parser's default behaviour here )?

ang-zeyu:
> (even for re-reviews!)

Will do 👍 Thanks for looking through this!

ang-zeyu:
Resolved minor conflict with #1049

src/lib/markbind/src/patches/htmlparser2.js
marvinchin: Oops, I actually thought that the original htmlparser allowed it, since the changes we made didn't seem to affect any of the logic specific to script/style tags. After taking a closer look at the rest of the changes in our patches, it seems like you are right 🙂

ang-zeyu:
> broken currently

Fixed, updated #1049's cheerio test site call with `xmlMode: false`.
Needed because in xmlMode `<abc></abc>` is recognised as an empty tag, hence it renders back to `<abc />`, which breaks once passed to the next plugin where `xmlMode = false`

PR #1110, opened 2020-03-09
Fix lazy reload for urls without index

### LINKED ISSUE #1109: Lazy live preview not working for links ending with /
v2.12.0
e.g., http://127.0.0.1:8080/2020/schedule/week8/ fails to trigger the building of the corresponding index.html page.

ang-zeyu:

• [x] Bug fix

Fixes #1109 

To normalize url requests that reference `index.html` implcitly ( eg `hostname/userGuide`, `hostname/userGuide/` ), such that a page build is triggered ( if needed ).

Added some code to normalize urls that reference `index.html` implicitly

```js
// Maps hostname/userGuide/  to hostname/userGuide/index
normalizedUrl = (hasNoExtension || hasEndingSlash)
  ? path.posix.join(normalizedUrl, 'index')
  : normalizedUrl;
```

na

**Testing instructions:**
lazy reload for urls that reference `index.html` implicitly ( eg. `userGuide/`, `userGuide` ) should now correctly trigger a page build ( if necessary )

Fix lazy reload for urls without index

Lazy live reloading uses the file extension of the request url to
determine whether the request is for a html page, and not from a
dynamic include.
Hence, requests for index.html with urls that do not end with
index.html fail to be detected.

Let’s add extra processing to detect such requests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -179,20 +179,37 @@ program
 
         if (onePagePath) {
           const lazyReloadMiddleware = function (req, res, next) {
-            const isHtmlFile = req.url.endsWith('.html');
+            const urlExtension = path.posix.extname(req.url);
 
-            if (isHtmlFile) {
-              const isDynamicIncludeHtmlFile = req.url.endsWith('._include_.html');
+            const hasEndingSlash = req.url.endsWith('/');
+            const hasNoExtension = urlExtension === '';
+            const isHtmlFileRequest = urlExtension === '.html' || hasEndingSlash || hasNoExtension;
 
-              if (!isDynamicIncludeHtmlFile) {
-                const urlWithoutBaseUrl = req.url.replace(config.baseUrl, '');
-                const urlWithoutExtension = fsUtil.removeExtension(urlWithoutBaseUrl);
+            if (!isHtmlFileRequest || req.url.endsWith('._include_.html')) {
+              next();
+              return;
+            }
 
-                const didInitiateRebuild = site.changeCurrentPage(urlWithoutExtension);
-                if (didInitiateRebuild) {
-                  req.url = ensurePosix(path.join(config.baseUrl || '/', LAZY_LOADING_SITE_FILE_NAME));
-                }
+            if (hasNoExtension && !hasEndingSlash) {
+              // Urls of type 'host/userGuide' - check if 'userGuide' is a raw file or does not exist
+              const diskFilePath = path.resolve(rootFolder, req.url);
+              if (!fs.existsSync(diskFilePath) || !fs.isDirectorySync(diskFilePath)) {
+                // Request for a raw file
+                next();
+                return;
               }
+            }
+
+            const urlWithoutBaseUrl = req.url.replace(config.baseUrl, '');
+            // Map 'hostname/userGuide/' and 'hostname/userGuide' to hostname/userGuide/index.
+            const urlWithIndex = (hasNoExtension || hasEndingSlash)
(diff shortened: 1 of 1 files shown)


ang-zeyu:
@nbriannl could I get your review here?

Aim of the changes here being to normalize `host/userGuide/` or `host/userGuide` ( if `host/userGuide` is not a raw file ) to `host/userGuide/index` ( intentionally extensionless )

nbriannl:
If i understand correctly, 
If I `markbind serve`
Changing `docs\userGuide\index.md` will trigger a live reload on both
http://127.0.0.1:8080/userGuide/ and http://127.0.0.1:8080/userGuide/index.html

Before this change, only http://127.0.0.1:8080/userGuide/index.html, would live reload?

ang-zeyu:
> If i understand correctly,
> With this change, when `markbind serve` and change `docs\userGuide\index.md`, I will trigger a live reload on both http://127.0.0.1:8080/userGuide/ and http://127.0.0.1:8080/userGuide/index.html
> 
> Before this change, only http://127.0.0.1:8080/userGuide/index.html, would live reload?

Yup

index.js
```diff
@@ -179,21 +179,43 @@ program
...
 
-              if (!isDynamicIncludeHtmlFile) {
-                const urlWithoutBaseUrl = req.url.replace(config.baseUrl, '');
-                const urlWithoutExtension = fsUtil.removeExtension(urlWithoutBaseUrl);
+            if (!isHtmlFileRequest || req.url.endsWith('._include_.html')) {
+              next();
+              return;
+            }
+
+            let normalizedUrl = req.url.replace(config.baseUrl, '');
```
marvinchin: We don't use `normalizedUrl` until L208. Could we define `normalizedUrl` closer to it's usage?

Also, I generally prefer using `const` and naming intermediate variables explicitly. I find that this makes it easier to follow what each step is doing.

ang-zeyu: Good catch 😅, thanks! Updated some of the comments as well

index.js
```diff
@@ -179,21 +179,42 @@ program
...
+                next();
+                return;
               }
             }
+
+            // Remove baseUrl
+            let normalizedUrl = req.url.replace(config.baseUrl, '');
+
+            // Map 'hostname/userGuide/' and 'hostname/userGuide' to hostname/userGuide/index.
+            normalizedUrl = (hasNoExtension || hasEndingSlash)
```
marvinchin: Should we do:
```
const urlWithoutBase = ...
const urlWithIndex = ...
const urlWithoutExtension = ...
```

I think we should avoid mutation where possible. In this case, we can break this transformation down into individual independent steps with variables names that make each step self-explanatory.

index.js
```diff
@@ -179,21 +179,42 @@ program
 
         if (onePagePath) {
           const lazyReloadMiddleware = function (req, res, next) {
-            const isHtmlFile = req.url.endsWith('.html');
+            const requestExtension = path.posix.extname(req.url);
```
marvinchin: I think just `extension` or `urlExtension` is fine? 😛 `req` itself is an object that has a bunch of other stuff in it, so `requestExtension` doesn't feel very correct to me.

index.js
marvinchin: The logical flow is fairly straightforward here, but in other more complex cases variable mutation can lead to some hard to trace errors. I have a strong preference for `const` over `let` whenever possible to avoid this, but I'm happy to discuss if you (or anyone else) has a different preference :)

ang-zeyu:
Updated the variable naming

PR #1111, opened 2020-03-09
Show pointer and use underline dashed for click trigger

### LINKED ISSUE #1053: Tool tips and popovers should be indicated by dotted lines, not dashed lines
v2.11.0
image

Problem: tooltips/popovers should be indicated by dotted underlines while dashed underlines are to be used for indicating modal windows.

openorclose:

• [X] Bug fix

Fixes #1053 

Regression from #1033, the migration to bootstrap-vue.

There were two styles for triggers depending on if it was a `click` trigger or not.

During the migration, only one style was used for all triggers.

Implemented the two different styles depending on type of trigger.

```html
<trigger trigger="click" for="id">Pointer cursor, underline dashed</trigger>
<trigger for="id">Normal cursor, underline dotted</trigger>
```

**Testing instructions:**

Check that click triggers have the proper style

```
Show pointer and use underline dashed for click trigger

A regression from #1033

Let's add back the two styles of triggers:
- click: show mouse pointer to signal that it is clickable
  and use underline dashed
- not click: no special cursor, and use underline dotted
```


ang-zeyu:
Seems to work fine 👍 

Might be good to change the `contextMenu` examples left over in the docs to click triggers though, and the corresponding example code

damithc:
@openorclose this doesn't seem to be working in the latest release. This hover tooltip is shown as a dashed underline (should be a dotted underline?)

image

PR #1115, opened 2020-03-10
Add start from line number functionality to code blocks

### LINKED ISSUE #1114: Code blocks: support specifying a starting line number
Suggestion: for code blocks, support specifying a starting line number.
Reason: 
* When there are multiple code blocks in a near vicinity, this feature will allow the a code block line numbers to continue from the previous code block. e.g., first code block has lines 1-5, the second one can be set to start from 6. 
* Also, it is useful when quoting a subset of lines from a longer code block that is already in the page (the sub set of lines can be set to follow the same line numbers they had in the longer code block.

openorclose:

Fixes #1114 

Authors may want to include a code snippet from a larger file, so they may prefer to have the original line numbers.

If the `start-from` attribute it set, add an inline style that sets the counter to the specified value -1.

```js {start-from=6}
function add(a, b) {
    return a + b;
}
```

Outcome:
image

How should this interact with line highlighting? I didn't change that part, so line highlighting still always starts from line 1.

Should we make line highlighting follow the `start-from` attribute too?

**Testing instructions:**

```
Add start from line number functionality to code blocks

Let's do this by checking for the `start-from` attribute on a code block
and then set an inline style that sets the `line` counter to the 
appropriate value.
```

### DIFF
--- a/src/lib/markbind/src/lib/markdown-it/index.js
+++ b/src/lib/markbind/src/lib/markdown-it/index.js
@@ -82,7 +82,7 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
 
   if (startFromOneBased > 1) {
     // counter is incremented on each span, so we need to subtract 1
-    token.attrJoin('style', `counter-set: line ${startFromZeroBased};`);
+    token.attrJoin('style', `counter-reset: line ${startFromZeroBased};`);
   }
   
   const highlightLinesInput = getAttributeAndDelete(token, 'highlight-lines');
--- a/src/lib/markbind/src/lib/markdown-it/index.js
+++ b/src/lib/markbind/src/lib/markdown-it/index.js
@@ -64,6 +64,13 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (!highlighted) {
     lines = markdownIt.utils.escapeHtml(str).split('\n');
   }
+
+  const startFrom = Number(token.attrGet('start-from'));
+
+  if (startFrom) {
+    // counter is incremented on each span, so we need to subtract 1
+    token.attrJoin('style', `counter-set: line ${startFrom - 1};`);
+  }
   
   const highlightLinesInput = token.attrGet('highlight-lines');
   let lineNumbersAndRanges = [];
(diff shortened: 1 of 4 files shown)


damithc:
Nice.

> Should we make line highlighting follow the `start-from` attribute too?

Feels like we should as I think authors will look at the live preview to figure out which lines to highlight.

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -64,6 +64,13 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (!highlighted) {
     lines = markdownIt.utils.escapeHtml(str).split('\n');
   }
+
+  const startFrom = Number(token.attrGet('start-from'));
+
+  if (startFrom) {
+    // counter is incremented on each span, so we need to subtract 1
+    token.attrJoin('style', `counter-set: line ${startFrom - 1};`);
```
alyip98:
```suggestion
    token.attrJoin('style', `counter-reset: line ${startFrom - 1};`);
```

[counter-reset](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-reset) has better browser compatibility than [counter-set](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-set)

test/functional/test_site/expected/testCodeBlocks.html
```diff
@@ -32,6 +32,8 @@
         <pre><code class="hljs js"><span><span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs-title">fourEmptyLinesBelow</span>(<span class="hljs-params"></span>) </span>{</span><span>&#x200B;</span><span>&#x200B;</span><span>&#x200B;</span><span>&#x200B;</span><span>} <span class="hljs-comment">// four empty lines above</span></span></code></pre>
         <p><strong>hljs span spanning multiple lines</strong></p>
         <pre><code class="hljs markdown"><span><span class="hljs-strong">*****</span></span><span>-----</span></code></pre>
+        <p><strong>start-from attr causes inline style to be set</strong></p>
+        <pre><code start-from="30" style="counter-set: line 29;" class="hljs markdown"><span><span class="hljs-strong">*****</span></span><span>-----</span></code></pre>
```
alyip98: We could also remove the `start-from` attribute from the tag

openorclose: Thanks! I'll remove the highlight lines attr also then

damithc:
This feature in action https://nus-te3201.github.io/2020/schedule/week11/topics-py.html
Good work @openorclose and reviewers

PR #1116, opened 2020-03-10
Upgrade js-beautify and provide option to turn it off

### LINKED ISSUE #1067: Upgrade js-beautify and provide a site configuration option to turn it on

Performance of markbind

Build times for 2103 site:

Before: **98s**
Removing js-beautify: **76s**
Upgrading js-beautify: **85s**

There's a significant benefit to build times in either option.

If we remove it, editing html post build would be less friendly for the user ( do we want them to edit it post build though? @damithc ), and on the other hand inspecting test output when updating test files may also be harder for us.
The performance benefits are clear in this case though, and its also one less tool in the page build process to consider when maintaining.

Upgrading it would be the other option here, which apart from performance benefits ( also quite significant ) has a few bug fixes. ( https://github.com/beautify-web/js-beautify/blob/master/CHANGELOG.md ) ( we are on `1.7.5` currently )

**Edit:** Third option [first comment](https://github.com/MarkBind/markbind/issues/1067#issuecomment-592910308)

Upgrade js-beautify, but also provide a configuration option in `site.json` to turn on js-beautify for html files, which would be off by default.
For our test_sites, we could keep it turned on to make it easier to review changes to expected test files in PRs.

Which should we go with?

na

**Additional context**
Inspecting the annotations and commit history for `js-beautify`, it seems this was added in the very early stages of markbind.
Over time a lot of functionality has been added though, which increased the overhead of beautifying html files significantly.

If we don't want editing html files post build to be as user friendly as possible ( mainly the indents, picture below ), removing it at this stage may be the better option.

indentgone

yash-chowdhary:

Resolves #1067 

We use an old version of `js-beautify` (`v1.7.5` as opposed to the latest `v1.10.3`). 
As documented in #1067, `js-beautify` affects the performance of our build process. It would be beneficial for MarkBind to turn off `js-beautify`, at the same time allowing users to choose to turn it on if needed. 

* Upgrades `js-beautify` to `v1.10.3`
* Adds support for a `jsBeautify` option in `site.json`, which when set to `true` will make use of `js-beautify` on HTML files

Added a check in multiple places similar to this -

```js
this.content = this.jsBeautify 
               ? htmlBeautify($.html(), Page.htmlBeautifyOptions)
               : $.html();
```

Since I've updated the `js-beautify` library, there have been changes in the expected output of the test sites according to the new changes/rules added in the library. 
Also, the test_site_convert and test_site_template run `init`, which means they will have `js-beautify` turned off. Just a heads up in case anyone was wondering why the formatting is off for the HTML files in those test sites.

**Testing instructions:**
`npm run test`

```
Upgrade js-beautify and provide users with an option to toggle it on.

MarkBind currently runs js-beautify v1.7.5, although later versions
of the library have been released (the latest version as of this PR
is v1.10.3).

Turning off js-beautify can improve the performance of MarkBind's 
build process, so let's make that the default setting.

Let's upgrade js-beautify to the latest version.
Let's also give the user the option to turn on js-beautify (whenever
necessary) through an option in site.json.
```


yash-chowdhary:
@ang-zeyu can I get your review on this?

docs/userGuide/siteConfiguration.md
```diff
@@ -199,6 +199,11 @@ The example above uses tags as an example of configuring plugin settings, refer
 
 **Specifies that the website should use MarkBind's search functionality.** Default: `true`. See [User Guide: Making the Site Searchable](makingTheSiteSearchable.html) for more details.
 
+#### **`jsBeautify`**
```
ang-zeyu: Maybe `htmlBeautify` would be clearer

package.json
```diff
@@ -34,7 +34,7 @@
     "highlight.js": "^9.14.2",
     "htmlparser2": "^3.10.1",
     "ignore": "^5.1.4",
-    "js-beautify": "~1.7.5",
+    "js-beautify": "^1.10.1-rc2",
```
ang-zeyu: Are there breaking changes from `1.10.1-rc2` to `1.10.3`?

package.json
```diff
@@ -46,11 +46,11 @@
     "markdown-it-linkify-images": "^1.1.1",
     "markdown-it-mark": "^2.0.0",
     "markdown-it-regexp": "^0.4.0",
+    "markdown-it-sub": "^1.0.0",
+    "markdown-it-sup": "^1.0.0",
```
ang-zeyu: 😮

src/Page.js
```diff
@@ -61,6 +61,7 @@ class Page {
     this.content = pageConfig.content || '';
     this.faviconUrl = pageConfig.faviconUrl;
     this.frontmatterOverride = pageConfig.frontmatter || {};
+    this.jsBeautify = pageConfig.jsBeautify;
```
ang-zeyu: As with above, `htmlBeautify` might be more semantic

src/Page.js
```diff
@@ -658,14 +659,23 @@ class Page {
...
-        = htmlBeautify(`<nav id="${PAGE_NAV_ID}" class="navbar navbar-light bg-transparent">\n`
-        + '<div class="border-left-grey nav-inner position-sticky slim-scroll">\n'
-        + `${pageNavTitleHtml}\n`
-        + '<nav class="nav nav-pills flex-column my-0 small no-flex-wrap">\n'
-        + `${pageNavHeadingHTML}\n`
-        + '</nav>\n'
-        + '</div>\n'
-        + '</nav>\n', Page.htmlBeautifyOptions);
+        = this.jsBeautify
+          ? htmlBeautify(`<nav id="${PAGE_NAV_ID}" class="navbar navbar-light bg-transparent">\n`
```
ang-zeyu: I wonder if these calls were actually unnecessary, since `htmlBeautify` is called on the entire html file at the end of page generation.
Could remove these calls if so for better performance

Will mark the other calls that could be investigated as well 🙋‍♂

src/Page.js
```diff
@@ -756,10 +766,12 @@ class Page {
     if (pageSection.length === 0) {
       return;
     }
-    this.pageSectionsHtml[section] = htmlBeautify($.html(section), Page.htmlBeautifyOptions)
-      .trim();
+    this.pageSectionsHtml[section]
+      = this.jsBeautify
+        ? htmlBeautify($.html(section), Page.htmlBeautifyOptions).trim()
```
ang-zeyu: 🙋‍♂

src/Page.js
```diff
@@ -756,10 +766,12 @@ class Page {
...
+        : $.html(section);
     pageSection.remove();
-    this.content = htmlBeautify($.html(), Page.htmlBeautifyOptions);
+    this.content = this.jsBeautify ? htmlBeautify($.html(), Page.htmlBeautifyOptions) : $.html();
```
ang-zeyu: 🙋‍♂

src/Page.js
```diff
@@ -805,7 +817,7 @@ class Page {
         .then(result => markbinder.processDynamicResources(this.sourcePath, result))
         .then(result => MarkBind.unwrapIncludeSrc(result))
         .then((result) => {
-          this.content = htmlBeautify(result, Page.htmlBeautifyOptions);
+          this.content = this.jsBeautify ? htmlBeautify(result, Page.htmlBeautifyOptions) : result;
```
ang-zeyu: 🙋‍♂

src/Page.js
```diff
@@ -822,10 +834,13 @@ class Page {
           this.collectAllPageSections();
           this.buildPageNav();
 
-          return fs.outputFileAsync(this.resultPath, htmlBeautify(
-            this.template.render(this.prepareTemplateData()),
-            Page.htmlBeautifyOptions,
-          ));
+          const renderedTemplate = this.template.render(this.prepareTemplateData());
+          const outputTemplateHTML
```
ang-zeyu: Would 
```
const outputTemplateHTML = this.jsBeautify
  ? ...
  : ...
```
be better?

src/Page.js
```diff
@@ -1081,7 +1096,11 @@ class Page {
             baseUrl,
             hostBaseUrl,
           });
-          return fs.outputFileAsync(resultPath, htmlBeautify(content, Page.htmlBeautifyOptions));
+          const outputContentHTML
+            = this.jsBeautify
```
ang-zeyu: likewise

src/Site.js
```diff
@@ -236,6 +236,7 @@ class Site {
       faviconUrl: config.faviconUrl,
       frontmatter: config.frontmatter,
       globalOverride: this.siteConfig.globalOverride || {},
+      jsBeautify: this.siteConfig.jsBeautify,
```
ang-zeyu: `htmlBeautify`

src/Page.js
```diff
@@ -822,10 +834,13 @@ class Page {
...
           this.buildPageNav();
 
-          return fs.outputFileAsync(this.resultPath, htmlBeautify(
-            this.template.render(this.prepareTemplateData()),
-            Page.htmlBeautifyOptions,
-          ));
+          const renderedTemplate = this.template.render(this.prepareTemplateData());
+          const outputTemplateHTML
+            = this.jsBeautify
+              ? htmlBeautify(renderedTemplate, Page.htmlBeautifyOptions)
```
ang-zeyu: https://github.com/beautify-web/js-beautify/blob/master/js/src/html/options.js
The `htmlBeautifyOptions` will need to be updated with `textarea` as well.

package.json
yash-chowdhary: upgraded to `1.10.3`.

package.json
yash-chowdhary: that's strange.. i don't remember installing these. uninstalled them now 👍

yash-chowdhary: ~that's strange.. i don't remember installing these. uninstalled them now~

I hadn't rebased my code on to master

src/Page.js
yash-chowdhary: just to clarify, do you mean bringing the "`= this.jsBeautify`" to the previous line, or to combine lines 837 and 838 to only create one `const` value as well?

package.json
yash-chowdhary: rectified

docs/userGuide/siteConfiguration.md
```diff
@@ -199,6 +199,11 @@ The example above uses tags as an example of configuring plugin settings, refer
...
+
+**Toggles-on [js-beautify](https://github.com/beautify-web/js-beautify) for html files generated by MarkBind.** By default, MarkBind turns off `js-beautify` to improve the performance of the MarkBind `build` process.
```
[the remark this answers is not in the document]
yash-chowdhary: As in, leave out `to improve the performance of the MarkBind `build` process`?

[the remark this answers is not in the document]
yash-chowdhary: because the description in the previous line states that it's only for HTML

test/functional/test_site/expected/test_md_fragment.html
```diff
@@ -1,40 +1,32 @@
 <!DOCTYPE html>
 <html>
-
 <head>
-  <meta name="default-head-top">
```
ang-zeyu: Seems like some of the calls are actually necessary; Likely the two that follow `this.pageSections[...] = htmlBeautify(...`

docs/userGuide/siteConfiguration.md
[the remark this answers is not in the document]
ang-zeyu: My bad was a little unclear here. The latter 2 references to `js-beautify` could be changed to reference the option instead ( e.g. `...by default markbind turns this off`, `toggling on html beautification allows...` ) of `js-beautify`. The heading could be changed to `htmlBeautify` to match with the option name as well.

test/functional/test_site/expected/test_md_fragment.html
yash-chowdhary: I'm not sure I understand.
The effect of `this.pageSections[...] = htmlBeautify(... ` seems to be that beautification is only done on sections like `<nav>`, `<header>`, `<footer>` and  `#site-nav`. 
It will not result in `<head>` beautification.
All these sections (mentioned above) will anyway be beautified as they are part of the main HTML.

Or is it that we _want_ beautification of these sections by default, and only allow the main content (as in the `.md` files excluding the options in frontmatter) to be beautified through  the toggle?

test/functional/test_site/expected/testAnchorGeneration.html
```diff
@@ -1,96 +1,85 @@
...
-        <h1 id="root-file">Root file<a class="fa fa-anchor" href="#root-file"></a></h1>
-        <h1 id="should-have-anchor">should have anchor<a class="fa fa-anchor" href="#should-have-anchor"></a></h1>
-        <h2 id="should-have-anchor-2">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-2"></a></h2>
-        <h3 id="should-have-anchor-3">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-3"></a></h3>
-        <h4 id="should-have-anchor-4">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-4"></a></h4>
-        <h5 id="should-have-anchor-5">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-5"></a></h5>
-        <h6 id="should-have-anchor-6">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-6"></a></h6>
-        <panel id="should-have-anchor-7"><template slot="_header"><h4 id="should-have-anchor-7">should have anchor<a class="fa fa-anchor" href="#should-have-anchor-7"></a></h4></template> Lorem ipsum
-        </panel>
-        <panel><template slot="_header"><p>Collapsed</p></template>
```
ang-zeyu: Not specifically to do with the head, some other areas lost their indent as well.

test/functional/test_site/expected/test_md_fragment.html
ang-zeyu: Placed the comment at a bad location, see https://github.com/MarkBind/markbind/pull/1116#discussion_r395998678

ang-zeyu: Might not be an issue with `this.pageSections[...] = htmlBeautify(...` as well, it seems to occur in quite a few areas 😕

test/functional/test_site/expected/testAnchorGeneration.html
yash-chowdhary: Fixed it. The option hadn't been updated to `htmlBeautify` from `jsBeautify`.

docs/userGuide/siteConfiguration.md
```diff
@@ -199,6 +199,18 @@ The example above uses tags as an example of configuring plugin settings, refer
 
 **Specifies that the website should use MarkBind's search functionality.** Default: `true`. See [User Guide: Making the Site Searchable](makingTheSiteSearchable.html) for more details.
 
+#### **`htmlBeautify`**
+
+**Toggles-on html beautification done by [js-beautify](https://github.com/beautify-web/js-beautify) for html files generated by MarkBind.** By default, MarkBind turns this option off to improve the performance of the MarkBind `build` process.
```
marvinchin: I personally lean towards defaulting to having `htmlBeautify` on, and allowing users to turn it off if they really wish to get the performance improvement.

The reason behind this are:
- The build process generally isn't very time critical or performance sensitive for most use cases
- Standardisation of our generated files make it easier to ignore diffs for non-material changes (i.e. markbind might make changes internally that changes the indendentation of files, but these changes don't actually change the markup of the site)
- Having inconsistently formatted files by default isn't really a good look for the product

src/Site.js
```diff
@@ -922,7 +923,7 @@ class Site {
     injectMarkdownItSpecialTags(tagsToIgnore);
     Page.htmlBeautifyOptions = {
       indent_size: 2,
-      content_unformatted: ['pre', ...tagsToIgnore],
+      content_unformatted: ['pre', 'textarea', ...tagsToIgnore],
```
marvinchin: What's the rationale for adding this new `textarea` option?

docs/userGuide/siteConfiguration.md
ang-zeyu: > The build process generally isn't very time critical or performance sensitive for most use cases

Hmm, although there's `markbind serve -o` now, it could still really help when writing the page ( ~20% performance gains ).

> Having inconsistently formatted files by default isn't really a good look for the product

Good point though

> Standardisation of our generated files make it easier to ignore diffs for non-material changes (i.e. markbind might make changes internally that changes the indendentation of files, but these changes don't actually change the markup of the site)

The test sites which don't have `htmlBeautify` turned on are only the `template sites` and `test_site_convert`.

In this sense, how about changing the one for the template sites to be `on` instead? This should solve the problem of presenting a good first impression, while defaulting to `off` if it is not specified.

`test_site_convert` would be an acceptable compromise here I think, but a future pr could always tackle this 😄

src/Site.js
ang-zeyu: html beautify was updated with this as well since our last version; unfortunately the options don't merge with the defaults, so a little manual inspection has to be done here

https://github.com/beautify-web/js-beautify/blob/master/js/src/html/options.js

docs/userGuide/siteConfiguration.md
marvinchin: > Hmm, although there's markbind serve -o now, it could still really help when writing the page ( ~20% performance gains ).

Good point. I wonder what the performance cost for turning beautify on is. Would anyone happen to have some metrics on that?

> In this sense, how about changing the one for the template sites to be on instead? This should solve the problem of presenting a good first impression, while defaulting to off if it is not specified.

I think that the build assets that we generate for the user matters as well. For the typical user, build performance is probably not a big concern. Users who are performance conscious are also generally more willing to tinker with their tools. But this, of course, is difficult to know for certain without further user research :P

I feel that the defaults should be set in a way such that a typical user needs minimal configuration to get an output they are happy with.

marvinchin: Maybe as part of user research we can ask @damithc to see if he happens to have any thoughts on this? 😃

ang-zeyu: > Good point. I wonder what the performance cost for turning beautify on is. Would anyone happen to have some metrics on that?

Yup, its in the original issue #1067 

> I think that the build assets that we generate for the user matters as well. For the typical user, build performance is probably not a big concern.

Wouldn't adding `htmlBeautify: on` to the template solve this though? ( since these are the sites the user starts with during `markbind init )

> I feel that the defaults should be set in a way such that a typical user needs minimal configuration to get an output they are happy with.

This does feel cleaner, my worry here is that performance of `markbind serve` also affects ux ultimately though. Not too sure which of the two we should go with either 😬

marvinchin: > Yup, its in the original issue #1067

Thanks for pointing me to it :)

> Wouldn't adding htmlBeautify: on to the template solve this though? ( since these are the sites the user starts with during `markbind init )

Ah okay, I see what you mean there. I still think that it makes logical sense for the default behaviour and the default configs we provide to be aligned, but changing the template `site.json` might be an acceptable compromise.

ang-zeyu: @damithc could we get your opinion on whether performance / neatness of generated source files should be prioritised in the default behaviour?

damithc: Perhaps safer to prioritize neatness as basic users might not need the performance gain (which will be tiny for small web sites) but might get a shock if they to take a look at the generated code? I am OK to go either way though.

yash-chowdhary: In that case, we can shift to enabling it by default with an option to turn it off.

docs/userGuide/siteConfiguration.md
```diff
@@ -202,6 +202,18 @@ The example above uses tags as an example of configuring plugin settings, refer
 
 **Specifies that the website should use MarkBind's search functionality.** Default: `true`. See [User Guide: Making the Site Searchable](makingTheSiteSearchable.html) for more details.
 
+#### **`disableHtmlBeautify`**
+
+**Turn off html beautification done by [js-beautify](https://github.com/beautify-web/js-beautify) for html files generated by MarkBind.** By default, MarkBind beautifies the HTML files it generates, which helps in user-friendly viewing and editing.
+Users can choose to turn this option off to improve the performance of the MarkBind `build` process.
```
ang-zeyu, line 208: ~Users~ -> You

PR #1117, opened 2020-03-11
Support variables to be defined in JSON

### LINKED ISSUE #319: Support global variables in nunjucks format
v1.8.4

Current: markbind variables are declared in the `variables.md` as the body of a `div` element and can be accessed globally. This has some downsides:
* Certain html fragments cannot be used as a variable value. e.g., `<div class="indented">` cannot be used as a variable value because it will break the nested div structure of the page due to missing `</div>` tag.
* Only flat string values can be a variable value. It may be useful to have more structured data values e.g., arrays of strings

Suggested: provide a way to declare variables in an alternate syntax that allows the above. We can use nunjucks variables syntax. However, markbind should raise such variables into the global scope, similar to normal markbind variables.

crphang:

Resolves #319

Allow variable to be defined in JSON. This allows us to define variable in different HTML fragments such as:

Added an additional logic to handle variable defined in JSON.

```
<variable type="json">
{
    "front": "<div> front",
    "back": "back </div>"
}
</variable>

{{ front }} {{ back}}
```

One of the attempts before relying on a common syntax like JSON was to enable nunjucks variables. However, there is no easy way to retrieve variables from nunjucks because of it's design: https://github.com/mozilla/nunjucks/issues/329.

**Testing instructions:**

`npm run test`

Support variables to be defined in JSON.

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -503,10 +503,10 @@ class Site {
     this.baseUrlMap.forEach((base) => {
       const userDefinedVariables = {};
       Object.assign(userDefinedVariables, markbindVariable);
-
+      const userDefinedVariablesPath = path.resolve(base, USER_VARIABLES_PATH);
+      const userDefinedVariablesDir = path.dirname(userDefinedVariablesPath);
       let content;
       try {
-        const userDefinedVariablesPath = path.resolve(base, USER_VARIABLES_PATH);
         content = fs.readFileSync(userDefinedVariablesPath, 'utf8');
       } catch (e) {
         content = '';
@@ -521,9 +521,27 @@ class Site {
       const $ = cheerio.load(content);
       $('variable,span').each(function () {
         const name = $(this).attr('name') || $(this).attr('id');
-        // Process the content of the variable with nunjucks, in case it refers to other variables.
-        const html = nunjuckUtils.renderEscaped(nunjucks, $(this).html(), userDefinedVariables);
-        userDefinedVariables[name] = html;
+        const variableSource = $(this).attr('from');
+
+        if (variableSource !== undefined) {
+          try {
+            const variableFilePath = path.resolve(userDefinedVariablesDir, variableSource);
+            const jsonData = fs.readFileSync(variableFilePath);
+            const varData = JSON.parse(jsonData);
+            Object.entries(varData).forEach(([varName, varValue]) => {
+              // Process the content of the variable with nunjucks, in case it refers to other variables.
+              const variableValue = nunjuckUtils.renderEscaped(nunjucks, varValue, userDefinedVariables);
+
+              userDefinedVariables[varName] = variableValue;
+            });
+          } catch (err) {
+            logger.warn(`Error ${err.message}`);
+          }
+        } else {
+          // Process the content of the variable with nunjucks, in case it refers to other variables.
+          const html = nunjuckUtils.renderEscaped(nunjucks, $(this).html(), userDefinedVariables);
+          userDefinedVariables[name] = html;
+        }
       });
     });
   }
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -8,6 +8,7 @@ const slugify = require('@sindresorhus/slugify');
 const componentParser = require('./parsers/componentParser');
 const componentPreprocessor = require('./preprocessors/componentPreprocessor');
 const nunjuckUtils = require('./utils/nunjuckUtils');
+const logger = require('../../../util/logger');
 
 const _ = {};
 _.clone = require('lodash/clone');
@@ -67,6 +68,7 @@ class Parser {
    */
   // eslint-disable-next-line class-methods-use-this
   extractPageVariables(fileName, data, userDefinedVariables, includedVariables) {
+    const fileDir = path.dirname(fileName);
     const $ = cheerio.load(data);
     const pageVariables = {};
     Parser.VARIABLE_LOOKUP.set(fileName, new Map());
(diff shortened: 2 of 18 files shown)


openorclose:
Is there any other place where `<variable>` defines multiple variables?

Trying to read 

```html
<variable type="json">
{
    "front": "<div> front",
    "back": "back </div>"
}
</variable>
```

sounds like "I'm creating a variable of type 'json' " to me, and not "I'm creating multiple variables whose names and values are defined in this json object".

damithc:
How about this?
```html
<variables>
{
    "front": "<div> front",
    "back": "back </div>"
}
</variables>
```
`variables` by default should be json syntax. Braces can be made optional (?)
```html
<variables>
    "front": "<div> front",
    "back": "back </div>"
</variables>
```

* main benefit of the feature: can be used to define multiple variables in one place.
* a side benefit: can be used to put non-well-formed values into variables
* caveat: clunky if the value is long and has a lot of symbols

acjh:
How about introducing a variables.json file for arbitrary strings?
I think this actually makes sense for a lot of those in [nus-cs2103/website-base's variables.md](https://raw.githubusercontent.com/nus-cs2103/website-base/c6fe7eae75282d99e98edc18c37d610503c01263/_markbind/variables.md).

I suspect why we had variables.md was to support valid HTML.
The example given would actually be interpreted in HTML as:

```html
<variable type="json">
    " { "front": ""
    <div> front", "back": "back </div>
    "" } "
</variable>
```

A limitation is that we have to define an order for which file is processed first (likely variables.json), which restricts which file can use variables in the other.
@damithc Is this limitation compatible with your use case?

damithc:
> @damithc Is this limitation compatible with your use case?

variables.json would have the global scope right? While having one would be good, it doesn't serve cases where the variable needs to be in local scope. e.g., an ugly partial html fragment that needs to be repeated several times in a particular page but is not used anywhere else.

acjh:
(Maybe I forgot something) but isn't variables.md scoped similarly?

damithc:
> (Maybe I forgot something) but isn't variables.md scoped similarly?

It is. I thought this new feature is meant to be used in both variables.md and as page variables. @crphang does it work like that?

crphang:
> "I'm creating a variable of type 'json'

True. Perhaps renaming the attribute might help such as `from=json`. Alternatively we could interpret content on the fly, by using value produced by JSON parse if it is valid and defaults to regular string expressions.

> It is. I thought this new feature is meant to be used in both variables.md and as page variables.

Yes that was the intention. However, since the file is still regarded as a `.md` and we are retrieving the content from cheerio, we might not be completely immune of issues regarding interference of syntax even though the tests I have done suggests that it alright.

ang-zeyu:
If the aim is to allow any content inside `<variable>`, how about adding it to the list of special tags?
Dosen't solve the second part of the issue though, so maybe we could add this to that

>How about introducing a variables.json file for arbitrary strings?

acjh:
> even though the tests I have added suggests that it alright.

@crphang The test that you have added for HTML happens to close itself.
Try swopping the order or only having the one with the opening tag.

crphang:
Letting variables be defined as Json within the page will cause issue if the HTML fragments are not self-closing as @acjh suggests.

I took another approach to allow user to json variables define in separate json file.

1. It supports variables.json as a separate file, but will still be backward compatible `variables.md` as well as page variables defined inline. This approach supports both page variables and global user defined variables. 
1. I also added this in initialization, to make it easy for new users to choose between defining variables inline or in a separate json file.

We could further add in user guide in [tips and tricks section](https://markbind.org/userGuide/tipsAndTricks.html#tips-and-amp-tricks) about allowing malformed html fragments in json variables if user requires.

damithc:
Remember to update user docs too.

crphang:
Updated user guide. Ready for review.

crphang:
@yamgent could you help with the review of this? 🙇

crphang:
@marvinchin could you help review this? :bowing_man:

src/Site.js
```diff
@@ -521,9 +521,27 @@ class Site {
...
+        const variableSource = $(this).attr('from');
+
+        if (variableSource !== undefined) {
+          try {
+            const variableFilePath = path.resolve(userDefinedVariablesDir, variableSource);
+            const jsonData = fs.readFileSync(variableFilePath);
+            const varData = JSON.parse(jsonData);
+            Object.entries(varData).forEach((variableEntry) => {
+              const [varName, varValue] = variableEntry;
+              const variableValue = nunjuckUtils.renderEscaped(nunjucks, varValue, userDefinedVariables);
```
marvinchin: I think it might be helpful to include the same comment as below:
`Process the content of the variable with nunjucks, in case it refers to other variables.`

src/Site.js
```diff
@@ -521,9 +521,27 @@ class Site {
...
-        const html = nunjuckUtils.renderEscaped(nunjucks, $(this).html(), userDefinedVariables);
-        userDefinedVariables[name] = html;
+        const variableSource = $(this).attr('from');
+
+        if (variableSource !== undefined) {
+          try {
+            const variableFilePath = path.resolve(userDefinedVariablesDir, variableSource);
+            const jsonData = fs.readFileSync(variableFilePath);
+            const varData = JSON.parse(jsonData);
+            Object.entries(varData).forEach((variableEntry) => {
```
marvinchin: I think we can do `forEach((varName, varValue) => { ... })` to avoid having to manually destructure each entry.

src/lib/markbind/src/parser.js
```diff
@@ -95,17 +97,40 @@ class Parser {
...
-        pageVariables[variableName] = variableValue;
-        Parser.VARIABLE_LOOKUP.get(fileName).set(variableName, variableValue);
+      const variableSource = $(this).attr('from');
+      if (variableSource !== undefined) {
+        try {
+          const variableFilePath = path.resolve(fileDir, variableSource);
+          const jsonData = fs.readFileSync(variableFilePath);
+          const varData = JSON.parse(jsonData);
+          Object.entries(varData).forEach((variableEntry) => {
+            const [varName, varValue] = variableEntry;
```
marvinchin: There seems to be some repetition between the json and non-json cases. Should we abstract this out into a separate function? That will make what we are trying to do a little more explicit too.

crphang:
@marvinchin thanks for reviewing! Updated the changes :smile:

src/lib/markbind/src/parser.js
```diff
@@ -95,17 +97,32 @@ class Parser {
...
+        } catch (err) {
+          logger.warn(`Error ${err.message}`);
+        }
+      } else {
+        if (!variableName) {
+          // eslint-disable-next-line no-console
+          console.warn(`Missing 'name' for variable in ${fileName}\n`);
+          return;
+        }
+        const variableValue = nunjuckUtils.renderEscaped(
```
marvinchin: Do you feel it makes sense to add this into the abstracted function too?
```
const variableValue = nunjuckUtils.renderEscaped(
          nunjucks, md.renderInline(variableElement.html()), {
            ...importedVariables, ...pageVariables, ...userDefinedVariables, ...includedVariables,
          });
```

marvinchin: Or perhaps, would it make sense to define an inline function? This way we can get the benefits of abstraction while still being able to access the variables we need directly. We could do something like:
```
const processVariable(variableName, rawVariableValue) {
  const otherVariables = { ...importedVariables, ...pageVariables, ...userDefinedVariables, ...includedVariables };
  const variableValue = nunjckUtils.renderEscaped(nunjucks, rawVariableValue, otherVariables);
  if (!pageVariables[variableName]) {
      pageVariables[variableName] = variableValue;
      Parser.VARIABLE_LOOKUP.get(fileName).set(variableName, variableValue);
    }
}
```

crphang:
@marvinchin Thanks for the suggestion, I refactored to make the logic cleaner as suggested. :+1:

PR #1122, opened 2020-03-14
Add class support for octicons

### LINKED ISSUE #1095: Octicons: support different modes
Octicon seems to have three modes dark/light/grid. Is there any way to support those modes? Not important though.

le0tan:

• [X] New feature

Fixes #1095 

Currently there's no way of controlling the style of icons except for using the icon pack's built-in modes. By supporting the class property on Octicons we can customize the size, color, etc. of these icons.

Appending `~class-name` to the octicon name will add the corresponding `class-name` to the `class` property of the generated svg.

Add ~class-name support to Octicons.


ang-zeyu:
Since `markdown-it-attrs` already allows adding classes eg. `:octicon-git-pull-request:{.my-5}`, how about what was mentioned here https://github.com/MarkBind/markbind/issues/1095#issuecomment-599013598?
Something like `octicond-git-pull-request` ( `d` instead of `s` ) 
Would be a really nice touch on top of what octicons offers

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
```diff
@@ -1,15 +1,22 @@
 const octicons = require('@primer/octicons');
 
 module.exports = require('markdown-it-regexp')(
-    /:(fa[brs]|glyphicon|octicon)-([a-z-]+):/,
+    /:(fa[brs]|glyphicon|octicon|octiconlight)-([a-z-]+)~?([a-z-]+)?:/,
```
ang-zeyu: how about `:(fa[brs]|glyphicon|octicon|octiconlight)-([a-z-]+)(~[a-z-]+)?:` so that `?` won't have to be repeated twice?

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
```diff
@@ -1,15 +1,22 @@
...
     (match, utils) => {
         let iconFontType = match[1];
         let iconFontName = match[2];
+        let iconClass = match[3];
 
         if (iconFontType === 'glyphicon') {
             return `<span aria-hidden="true" class="glyphicon glyphicon-${iconFontName}"></span>`;
         } else if (iconFontType === 'octicon') {
+          if (iconClass) {
+            return octicons[iconFontName].toSVG({"class": iconClass});
```
ang-zeyu: I personally prefer ternary expressions instead of 1-line `if-else`s especially since this is nested, but i'm fine eitherway 😄

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
```diff
@@ -1,15 +1,22 @@
...
+          } else {
             return octicons[iconFontName].toSVG();
+          }
+        } else if (iconFontType === 'octiconlight') {
+          return octicons[iconFontName].toSVG({"style": "color: #fff;"});
```
ang-zeyu: Should we add class support here as well?

src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
ang-zeyu: Realised `match[3]` needs to be updated as well here; Probably better to revert it to the original

ang-zeyu:
@le0tan there's some conflicts

src/lib/markbind/src/lib/markdown-it/markdown-it-icons.js
```diff
@@ -1,15 +1,20 @@
...
         let iconFontType = match[1];
         let iconFontName = match[2];
+        let iconClass = match[3];
 
         if (iconFontType === 'glyphicon') {
             return `<span aria-hidden="true" class="glyphicon glyphicon-${iconFontName}"></span>`;
         } else if (iconFontType === 'octicon') {
-            return octicons[iconFontName].toSVG();
+            return iconClass ? octicons[iconFontName].toSVG({"class": iconClass})
+                             : octicons[iconFontName].toSVG();
```
ang-zeyu, line 14: Should we follow the normal ternary style in other places instead?

.......
  ? ...
   : ...

ang-zeyu: dosen't seem to be resolved @le0tan; The reason this works here is that the `markdown-it` folder is ignored in eslint.
Its a very minor nit though, could resolve before merging if you have the time!

Looks good otherwise 👍

ang-zeyu:
@le0tan do refer here https://markbind.org/devdocs/devGuide/maintainerGuide.html for the merging procedures for future prs.
Use squash and merge for prs with only one logical commit, making sure the commit title is appropriately set. Likewise for #1039 

@acjh @yamgent do we force push master to clean it up in such a case?

Untitled

damithc:
Working in production. Good work @le0tan and reviewers.

PR #1123, opened 2020-03-14
Update documentation on icon slot for boxes

### LINKED ISSUE #1030: <box>: Allow slot support for icons
Maybe we should add `slot` support for `icon` property? That way, we can
* change color of the icon
* use thumbnails as icons

le0tan:

• [X] Documentation update

Please replace the vue-strap file with https://github.com/MarkBind/vue-strap/pull/136

Fixes #1030

### DIFF
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -72,9 +72,31 @@ Using the normal syntax, you are only able to use markdown formatting on heading
 <br>
 
 **Box Slot Options:**
-Slot name | Default class | Notes
---- | --- | --- 
-_icon | (depends on box's `type` attribute) | Example: `<i slot="_icon" class="fas fa-times"></i>` will replace the icon with a cross sign in spite of the box's type.
+
+Slot name | Default class |
+--- | --- |
+icon | (depends on box's `type` attribute) |
+
+*Example 1*: Override the default icon for a certain type of box.
+
+```html
+<box type="info">
+    <span slot="icon" class="text-danger"><md>:fas-home:</md></span>
+    info
+</box>
+```
+
+<box type="info">
+    <span slot="icon" class="text-danger"><md>:fas-home:</md></span>
+    info
+</box>
+
+*Example 2*: Use pictures (or even gifs) as icon for a box.
+
+<box type="info">
+    <img slot="icon" src="https://icons8.com/vue-static/landings/animated-icons/icons/cloud/cloud.gif"></img>
+    some very useful info
+</box>
 
 **Panel Slot Options:**
 Slot name | Default class | Notes
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -71,6 +71,11 @@ Using the normal syntax, you are only able to use markdown formatting on heading
 </tip-box>
 <br>
 
+**Box Slot Options:**
+Slot name | Default class | Notes
+--- | --- | --- 
+_icon | (depends on box's `type` attribute) | Example: `<i slot="_icon" class="fas fa-times"></i>` will replace the icon with a cross sign in spite of the box's type.
+
 **Panel Slot Options:**
 Slot name | Default class | Notes
 --- | --- | --- 
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -75,9 +75,9 @@ Using the normal syntax, you are only able to use markdown formatting on heading
 
 Slot name | Default class |
 --- | --- |
-icon | (depends on box's `type` attribute) |
+icon | depends on box's `type` attribute |
 
-*Example 1*: Override the default icon for a certain type of box.
+{{ icon_example }} Override the default icon for a certain type of box.
 
 ```html
 <box type="info">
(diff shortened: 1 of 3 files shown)


docs/userGuide/components/advanced.md
```diff
@@ -71,6 +71,11 @@ Using the normal syntax, you are only able to use markdown formatting on heading
 </tip-box>
 <br>
 
+**Box Slot Options:**
+Slot name | Default class | Notes
+--- | --- | --- 
+_icon | (depends on box's `type` attribute) | Example: `<i slot="_icon" class="fas fa-times"></i>` will replace the icon with a cross sign in spite of the box's type.
```
damithc: `<i slot="_icon" class="fas fa-times"></i>` is not how I normally use the `slot` technique. 
I was expecting to use it like `<span slot="icon" class="text-danger">:fas-home:</span>`. Would that work?

le0tan: You can use `<span slot="_icon" class="text-danger"><md>:fas-home:</md></span>`:
image
The only differences are: using `_icon` instead of `icon`, using `<md>` to enforce the markdown parsing inside embedded HTML.

damithc: Nice.
I suppose not easy to change to `icon` instead of `_icon`? If not, this is going to be the first time we expose a underscore prefixed name in the user guide.

ang-zeyu: > Nice.
> I suppose not easy to change to `icon` instead of `_icon`? If not, this is going to be the first time we expose a underscore prefixed name in the user guide.

Should be pretty simple. Having an underscore does not change vue's slot behaviour in anyway.
`_xxx` is just the naming convention for slots we don't want to be exposed to the user https://github.com/MarkBind/vue-strap/pull/120#discussion_r366824610

damithc: Let's use `icon` then, as we are now exposing it to the users?

damithc: Also, are there any limitations as to what can be put in the slot? e.g., can I put a markbind thumnail or an image in the icon slot?

le0tan: > Let's use `icon` then, as we are now exposing it to the users?

Sure, will create corresponding vue-strap PR shortly.

> Also, are there any limitations as to what can be put in the slot? e.g., can I put a markbind thumnail or an image in the icon slot?

There shouldn't be any limitations as it's just replacing the placeholder in the template. Just the formatting could be messed up and need manual config.

damithc: Good. Can try some reasonable use cases (e.g., using thumbnails, images, small animated gifs) as icons and give them as examples in the documentation.

docs/userGuide/components/advanced.md
```diff
@@ -71,6 +71,33 @@ Using the normal syntax, you are only able to use markdown formatting on heading
...
+    info
+</box>
+
+*Example 2*: Use pictures (or even gifs) as icon for a box.
+
+<box type="info">
+    <img slot="icon" src="https://icons8.com/vue-static/landings/animated-icons/icons/cloud/cloud.gif"></img>
+    some very useful info
+</box>
+
```
damithc: Can post a screenshot? I can't see the final look from the PR preview.

le0tan: 63a0b9f4-8658-4f6d-9077-d8102c6152b2

damithc: For the animated gif, can find one without background (or make the box background white) so that it blends in nicely? We should try to make the examples as good looking as possible.

Can also add an example using a thumbnail as an icon (i.e., using MarkBind thumbnails feature), possibly a thumbnail of a letter or a number?

docs/userGuide/components/advanced.md
```diff
@@ -71,6 +71,33 @@ Using the normal syntax, you are only able to use markdown formatting on heading
 </tip-box>
 <br>
 
+**Box Slot Options:**
+
+Slot name | Default class |
+--- | --- |
+icon | (depends on box's `type` attribute) |
```
ang-zeyu: Could remove the brackets `(` `)`

docs/userGuide/components/advanced.md
```diff
@@ -71,6 +71,33 @@ Using the normal syntax, you are only able to use markdown formatting on heading
...
+
+*Example 1*: Override the default icon for a certain type of box.
```
ang-zeyu: Could use `{{ icon_example }}` to be consistent with the rest of the page here

damithc:
@le0tan perhaps this enhancement got affected by MarkBind/vue-strap#131 ?
image

le0tan:
@damithc looks like so. As the width of the icon wrapper is fixed, it makes the slot for icon less useful - `1em` width is usually not enough for pictures. @yash-chowdhary can you take a look?

yash-chowdhary:
> @le0tan perhaps this enhancement got affected by [MarkBind/vue-strap#131]

This issue can be fixed by making use of our `icon-size` attribute.
Here's the original (w/o `icon-size`) -
Screenshot 2020-04-02 at 9 54 43 AM

Here's the box with `icon-size`= "2x" - 
Screenshot 2020-04-02 at 9 55 19 AM
 
We can still maintain the width of the icon-wrappers with this enhancement.

damithc:
> For @damithc 's example on using thumbnails -
> Screenshot 2020-04-04 at 1 56 02 PM

The thumbnail should be circle.

BTW, if we have to make a choice, I'd rather sacrifice fixed alignment to gain better aspect ratio. Hopefully, we can have both.

yash-chowdhary:
The solution mentioned in #1172 would fix this.
Vuestrap [PR link](https://github.com/MarkBind/vue-strap/pull/142)

PR #1128, opened 2020-03-16
Octicons: support icon-size property

### LINKED ISSUE #1096: Octicons: support icon-size property
Octicons don't seem to work with the `icon-size` property of boxes.

hcwong:
• [x] Bug fix

Fixes #1096 

The bug occured because there was the style property currently for icons only works for favicons. We do some HTML manipulation in the parser to detect if theres magnification and adjust the size accordingly. We cannot do this in the vue component as to change a svg's size, we need to change the element directly and its surrounding divs.

I haven't really played with the parser code, so if its not in line with the current style, do let me know. Also, I'm not sure if this octicon icon with icon-size is used anywhere else. If so, I might want to generalize this solution. 

**Testing instructions:**
Run the code with octicons and use the icon-size property. Regardless the icon-size property is there or not, it should not break the code and the octicon should vary in size

Support icon-size property for octicons.

### DIFF
--- a/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
+++ b/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
@@ -1,7 +1,7 @@
 const octicons = require('@primer/octicons');
 
 module.exports = require('markdown-it-regexp')(
-  /:(fa[brs]|glyphicon|octicon)-([a-z-]+):/,
+    /:(fa[brs]|glyphicon|octicon)-([a-z-]+):/,
     (match, utils) => {
         let iconFontType = match[1];
         let iconFontName = match[2];
--- a/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
+++ b/src/lib/markbind/src/lib/markdown-it-shared/markdown-it-icons.js
@@ -1,20 +1,15 @@
 const octicons = require('@primer/octicons');
 
 module.exports = require('markdown-it-regexp')(
-    /:(fa[brs]|glyphicon|octicon)-([a-z-]+):([0-9])x/,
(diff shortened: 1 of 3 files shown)


hcwong:
@le0tan can I tag you as reviewer haha?

dvrylc:
@hcwong I actually did a bit of digging on this problem (before I realized you were working on it) and found this: 

The way `icon-size` currently works is that it sets classes like `fa-2x` (for example) on the icon elements. Font Awesome's CSS then provides the styles for these classes by setting the `font-size` property on them. This approach doesn't work for octicons because their size is controlled by height and width (as you already know) and I managed to fix this by writing a bit of CSS: ef4f0b585e971a798281851db5c3f811994cf270. 

Not sure if this is useful to you, but just leaving it here as an idea.

hcwong:
> @hcwong I actually did a bit of digging on this problem (before I realized you were working on it) and found this:
> 
> The way `icon-size` currently works is that it sets classes like `fa-2x` (for example) on the icon elements. Font Awesome's CSS then provides the styles for these classes by setting the `font-size` property on them. This approach doesn't work for octicons because their size is controlled by height and width (as you already know) and I managed to fix this by writing a bit of CSS: [ef4f0b5](https://github.com/MarkBind/markbind/commit/ef4f0b585e971a798281851db5c3f811994cf270).
> 
> Not sure if this is useful to you, but just leaving it here as an idea.

Not sure about this, but do we need to modify the svg height and width attribute directly?

dvrylc:
> Not sure about this, but do we need to modify the svg height and width attribute directly?

Not quite sure what you mean but the `.octicon` class is set on the SVG element, so by modifying the class attributes we're setting the SVG's height and width. Was that your question?

PR #433, opened 2018-09-18
Show anchor links for each heading

jamos-tay:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] New feature

Fixes #426 

**What is the rationale for this request?**

Users may want to share a link to a specific heading with others.

**What changes did you make? (Give an overview)**

Allow users to automatically generate anchor links for each heading that direct the URL to that heading.

**Provide some example code that this change will affect:**

Users must include an `anchors` property in the `frontmatter`:

`anchors: true`

Functions similar to https://developers.timekit.io/v2/reference#section-root-endpoint-versions-and-protocol

- Hovering over a header shows an anchor icon
- Clicking on the icon links the URL to the header

out2

**Is there anything you'd like reviewers to focus on?**

Should we make the behavior different from the example site (e.g. change the icon)?

PR #1129, opened 2020-03-16
Add plugin local asset collection

### LINKED ISSUE #839: Allow plugins to add assets
At the moment, plugins do not have the ability to add custom assets to be served. Plugins authors are hence unable to use custom assets for their plugins (e.g. CSS stylesheets, images).

Plugin authors should be able to add custom assets to be served from as part of the MarkBind site. This allows MarkBind plugins to be more versatile.

ang-zeyu:

• [x] Documentation update

Resolves #839

#442 - fixed with `stopPropagation` instead of relying on runtime `window.location.href` shifting
#433's hiding/displaying of anchor icons - done using css instead

There is some degree of logic within the markbind core code that is related to plugins.
This is because the plugin `getLinks` and `getScripts` methods only allow linking to external sources.

We can enhance the API hence, allowing us to encapsulate plugin logic within itself, which should lead to better code readability.

Commit organization:
1. Add local plugin asset collection, and small update to docs accordingly
2. Encapsulate `markbind-plugin-anchors` using this new feature
3. Add and update tests for (1)
4. Update tests for (2)

Was there a reason #433 was done using runtime jQuery?

**Testing instructions:**
- `npm run test` should pass

Add local asset collection for plugins

Plugins might want to package their own assets for use instead of
relying on external sources only.
The main markbind code also has some amount of logic and styles
relating to the anchor plugin.

Let's add this, allowing the getLinks and getScripts methods to return
link and script elements that have a relative or absolute file path as
its src or href attributes.
Let's encapsulate the anchor plugin's logic and styles within the
plugin itself, which should lead to better code readability.
This also allows the user to turn off the plugin without risk of side
effects.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -40,6 +40,7 @@ const {
   FRONT_MATTER_NONE_ATTR,
   PAGE_NAV_ID,
   PAGE_NAV_TITLE_CLASS,
+  PLUGIN_SITE_ASSET_FOLDER_NAME,
   SITE_NAV_ID,
   SITE_NAV_LIST_CLASS,
   TITLE_PREFIX_SEPARATOR,
@@ -1119,6 +1120,45 @@ class Page {
   }
 
   /**
+   * Resolves a resource specified as an attribute in a html asset tag
+   * (eg. '<script>' or '<link>') provided by a plugin, and copies said asset
+   * into the plugin's asset output folder.
+   * Does nothing if the resource is a url.
+   * @param assetElementHtml The asset element html, as a string, such as '<script src="...">'
+   * @param tagName The name of the resource tag
+   * @param attrName The attribute name where the resource is specified in the tag
+   * @param plugin The plugin object from which to retrieve its asset src and output paths
+   * @param pluginName The name of the plugin, used to determine a unique output path for the plugin
+   * @return String html of the element, with the attribute's asset resolved
+   */
+  getResolvedAssetElement(assetElementHtml, tagName, attrName, plugin, pluginName) {
+    const $ = cheerio.load(assetElementHtml, { xmlMode: false });
+    const el = $(`${tagName}[${attrName}]`);
+
+    el.attr(attrName, (i, assetPath) => {
+      if (!assetPath || utils.isUrl(assetPath)) {
+        return assetPath;
+      }
+
+      const srcPath = path.resolve(plugin._pluginAbsolutePath, assetPath);
+      const srcBaseName = path.basename(srcPath);
+
+      fs.existsAsync(plugin._pluginAssetOutputPath)
+        .then(exists => exists || fs.mkdirp(plugin._pluginAssetOutputPath))
+        .then(() => {
+          const outputPath = path.join(plugin._pluginAssetOutputPath, srcBaseName);
+          fs.copyAsync(srcPath, outputPath, { overwrite: false });
+        })
+        .catch(err => logger.error(`Failed to copy asset ${assetPath} for plugin ${pluginName}\n${err}`));
+
+      return path.posix.join(this.baseUrl || '/', PLUGIN_SITE_ASSET_FOLDER_NAME, pluginName, srcBaseName);
+    });
+
+    return $.html();
+  }
+
+
+  /**
    * Collect page content inserted by plugins
    */
   collectPluginsAssets(content) {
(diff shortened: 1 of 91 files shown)


src/Page.js
```diff
@@ -983,6 +984,28 @@ class Page {
    * Collect page content inserted by plugins
    */
   collectPluginsAssets(content) {
+    const getResolvedAssetElement = (html, tagName, attrName, plugin, pluginName) => {
```
openorclose, line 987: only one suggestion:

The implementation of `utils.buildScript` and `buildStyleSheet` plus this `getResolvedElement` seems rather redundant, where build* produces `<script src =...></script>`, and getResolvedElement later reparses it. If buildScript could instead return something like

```js
{ type: 'scriptsrc', src: 'file.js' }
```

then we wouldn't need to use cheerio at all, just convert the src first and then return  `<script src =...></script>`.

openorclose:
Nice refactor!

LGTM, only one small suggestion.

src/Page.js
openorclose: One con with this suggestion would be that the user cannot manually return `<script src = path></script>`.

Maybe it's better to leave it in so the user has more choices?

ang-zeyu: Reworked the `buildScript/buildStylesheet` a little with the new `pluginUtils` from #1100, its the same from a plugin author's standpoint though ( changes are limited to the first of the 4 commits ).

```
{ type: 'scriptsrc', src: 'file.js' }
```
for this, I think returning `<script/link>` directly is more intuitive from a plugin standpoint, since we represent html in html and not in json. This way the author can also add other variables to the tag as well easily, e.g. `<link prefetch ... />`.

openorclose: Ah, but now the user has to use buildScript if they want an external link.

So they can't add other attributes like `<link prefetch href = "externa.css" />`.

The first implementation is probably better

ang-zeyu: Good point, reverted that. ( `getResolvedAssetElement` is able to stay in `pluginUtils` though )

ang-zeyu:
@marvinchin could I get your thoughts here? since you opened up #839 =)

The main change here being letting plugins add local assets instead of just ones from external cdns ( first of the 4 commits )

marvinchin:
Sorry for the delays 🙏 I'll look at this tomorrow!

ang-zeyu:
> Sorry for the delays 🙏 I'll look at this tomorrow!

No rush, not in a hurry to get this merged! 😄

src/Site.js
```diff
@@ -892,6 +893,17 @@ class Site {
...
+      if (!this.plugins[plugin].getLinks && !this.plugins[plugin].getScripts) {
+        return;
+      }
+
+      // For resolving plugin asset source paths later
+      this.plugins[plugin]._pluginAbsolutePath = path.dirname(require.resolve(pluginPath || plugin));
+      this.plugins[plugin]._pluginAssetOutputPath = path.resolve(this.outputPath,
+                                                                 PLUGIN_SITE_ASSET_FOLDER_NAME, plugin);
+
+      fs.mkdirSync(this.plugins[plugin]._pluginAssetOutputPath, { recursive: true });
```
marvinchin, line 906: Even if plugins have assets, the sources might be absolute right? In that case, there might not be any assets copied to the output folder. Should we defer the creation of the folder until the point that we are copying it over?

marvinchin:
Also just to note, this PR could have been split into two parts - one adding support for local assets, and then the other one using it to fix `markbind-plugin-anchors` 🙂 This way each PR could be focused and easier to review. Not necessary to split them up now though!

src/util/pluginUtil.js
```diff
@@ -50,4 +53,42 @@ const pluginUtil = {
   },
 };
 
-module.exports = pluginUtil;
+/**
```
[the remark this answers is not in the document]
ang-zeyu: moved it back to `page`

test/functional/test_site/expected/markbind/js/setup.js
```diff
@@ -13,63 +13,47 @@ function scrollToUrlAnchorHeading() {
   }
 }
 
-function flattenModals() {
```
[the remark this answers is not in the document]
ang-zeyu: yup, its a leftover from the bootstrap-vue PR

ang-zeyu:
@nbriannl need you to remove the change request =X

PR #1132, opened 2020-03-17
Fix octicon sizing

### LINKED ISSUE #1096: Octicons: support icon-size property

dvrylc:
• [X] Bug fix

Fixes #1096. 

Currently, icons are sized by using CSS classes (`fa-2x` for example) that set the `font-size` property. Since octicons are not font-based but rather SVG-based, this style does not affect them, causing them to be incorrectly sized. 

Screen Shot 2020-03-17 at 11 50 59 PM

Screen Shot 2020-03-18 at 11 30 36 AM

This PR sets the height and width for octicons to `1em`, allowing them to inherit their parent container's font size as its dimensions. 

Screen Shot 2020-03-17 at 11 51 40 PM

Screen Shot 2020-03-18 at 11 29 59 AM

**Testing instructions:**

1. Use an octicon where you would expect it to be a different size. Some examples are: 
    - Headings
    - `Box` icon with a custom `icon-size`
2. The size of the octicon should match where it's used

**Credits**
- @ang-zeyu for the generalized solution (in the comments below)
- @hcwong for the existing work on #1128 (which this PR fixes as well) 

Fix octicon sizing

### DIFF
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -442,33 +442,8 @@ pre {
     outline: none;
 }
 
-/* Octicon heading sizes */
-h1 .octicon {
-    height: 48px;
-    width: 48px;
-}
-
-h2 .octicon {
-    height: 38px;
-    width: 38px;
-}
-
-h3 .octicon {
-    height: 33px;
-    width: 33px;
-}
-
-h4 .octicon {
-    height: 28px;
-    width: 28px;
-}
-
-h5 .octicon {
-    height: 24px;
-    width: 24px;
-}
-
-h6 .octicon {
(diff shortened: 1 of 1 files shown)


ang-zeyu:
How about simply 

```
.octicon {
  width: 1em,
  height: 1em,
}
```
?

This would allow it to be responsive in any container instead of just headings.
It would likely fix #1128 as well

dvrylc:
> ```
> .octicon {
>   width: 1em,
>   height: 1em,
> }
> ```

Oh right that makes sense, thanks!

PR #1134, opened 2020-03-18
Remove OK button from modals

### LINKED ISSUE #1126: Remove OK button from modals
v2.12.0
Looks like an `OK` button has gotten into modals (by accident?). As we already have an `X` button, do we need the `OK` button too?

image

openorclose:

• [x] Bug fix

Fixes #1126

Remove the redundant OK button from modals, only show it if the author specifies `ok-title`.

If there is no `ok-title` attribute, and there is no footer slot, hide the footer.

**Testing instructions:**
Added a test and a docs example to show the case where `ok-text` is set.

```
Remove OK button from modals

Let's remove the redundant OK button from modals,
and only show it if the `ok-title` attribute is present.
```

### DIFF
--- a/docs/userGuide/syntax/modals.mbdf
+++ b/docs/userGuide/syntax/modals.mbdf
@@ -22,6 +22,15 @@ This is the same <trigger for="modal:loremipsum">trigger</trigger> as last one.
   Centered
 </modal>
 
+<trigger for="modal:ok-text">This is a trigger for a modal with a custom OK button</trigger>.
+
+<modal header="OK button visible!" id="modal:ok-text" ok-text="Custom OK">
+  Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore
+  magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+  consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+  Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+</modal>
+
 ```
 </span>
 <span id="output">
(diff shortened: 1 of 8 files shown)


src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -311,7 +311,18 @@ function _parseModalAttributes(node) {
...
   _renameAttribute(node, 'center', 'centered');
 
-  node.attribs['ok-only'] = ''; // only show OK button
+  const hasOkTitle = _.has(node.attribs, 'ok-title');
+  const hasFooter = node.children.some(child =>
+    _.has(child.attribs, 'slot') && child.attribs.slot === 'modal-footer');
+
+  if (!hasFooter && !hasOkTitle) {
+    // markbind doesn't show the footer by default
+    node.attribs['hide-footer'] = '';
```
ang-zeyu: does setting it to `true` erase the `=""` portion in `hide-footer=""`? might be nice if so, same for `ok-only`

docs/userGuide/syntax/modals.mbdf
```diff
@@ -22,6 +22,15 @@ This is the same <trigger for="modal:loremipsum">trigger</trigger> as last one.
...
+<modal header="OK button visible!" id="modal:ok-text" ok-text="Custom OK">
+  Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore
+  magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+  consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+  Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+</modal>
+
 ```
 </span>
 <span id="output">
```
ang-zeyu: should `ok-text` be `ok-title` instead? ( the props table )

src/lib/markbind/src/parsers/componentParser.js
openorclose: Nope, i just get this

image

docs/userGuide/syntax/modals.mbdf
openorclose: Should be fine as `ok-text`. since it's not really a title of anything.

ang-zeyu: > Remove the redundant OK button from modals, only show it if the author specifies `ok-title.`

Ahh ok, thought the attribute from the user standpoint here was `ok-title`

PR #1137, opened 2020-03-21
Code blocks: Implement inline markdown support for heading

### LINKED ISSUE #1119: code blocks: support inline markdown for heading
Current: `{heading="Heading title"}`
Suggested: support cases such as `{heading="**Heading** title"}`

Not sure what the cost is, but if it is not too high, it's nice to support inline markdown formatting as we support it in other places.

nbriannl:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Fixes #1119 

It would be nice to support inline markdown formatting as we support it in other places.

Render `heading` markdown-it-attr as inline markdown using `markdownIt.renderInline(heading)`

If heading contains rendered inline markdown, increase line height because default line-height will make the headings crowded. As seen below

_with `line-height: 1.5`_
image

_with `lint-height: 1`_
image

Added tests as well.

````
```html {heading="### Heading level 3"}
<foo></foo>
```

```html {heading="Line breaks<br>Are possible<br>[And links!](https://markbind.org)"}
<foo></foo>
```

```html {heading="**Bold**, _Italic_, ___Bold and Italic___, ~~Strike through~~, :+1: :exclamation: :x: :construction:"}
<foo></foo>
```

```html {heading="****Super Bold****, ++Underline++, ==Highlight==, %%Dim%%"}
<foo></foo>
```

```html {heading="<blockquote>Using blockquote tag</blockquote>"}
<foo></foo>
```

```html {heading="![](https://markbind.org/images/logo-lightbackground.png)"}
<foo></foo>
```

```html {heading="@[youtube](http://www.youtube.com/watch?v=v40b3ExbM0c)"}
<foo></foo>
```

- Can't support `inline code` text style (because it would break markdown-it attrs)
- Dim does not really work.
- Can't do headers, lists, paragraphs, footnotes, embeds
````

**Testing instructions:**

Try the test code on your own or check out 
https://nbriannl.github.io/markbind-plain-site/

Code blocks: Implement inline markdown support for heading

### DIFF
--- a/docs/userGuide/syntax/code.mbdf
+++ b/docs/userGuide/syntax/code.mbdf
@@ -149,6 +149,25 @@ To add a heading, add the attribute `heading` with the heading text as the value
 </span>
 </include>
 
+Headings support inline Markdown, except for `Inline Code` and %%Dim%% text styles.
+
+<include src="outputBox.md" boilerplate >
+<span id="code">
+
+```` {.no-line-numbers}
+```{heading="**Bold**, _Italic_, ___Bold and Italic___, ~~Strike through~~, ****Super Bold****, ++Underline++, ==Highlight==, :+1: :exclamation: :x: :construction:<br>We support page breaks"}
+<foo></foo>
+```
+````
+</span>
+<span id="output">
+
+```{heading="**Bold**, _Italic_, ___Bold and Italic___, ~~Strike through~~, ****Super Bold****, ++Underline++, ==Highlight==, :+1: :exclamation: :x: :construction:<br>We support page breaks"}
+<foo></foo>
+```
+</span>
+</include>
+
 ##### Using multiple features
 You can also use multiple features together, as shown below.
 
--- a/src/lib/markbind/src/lib/markdown-it/index.js
+++ b/src/lib/markbind/src/lib/markdown-it/index.js
@@ -136,8 +136,10 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   const heading = token.attrGet('heading');
   const codeBlockContent = `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
   if (heading) {
+    const renderedHeading = markdownIt.renderInline(heading);
+    const headingStyle = (renderedHeading === heading) ? 'code-block-heading' : 'code-block-heading inline-markdown-heading';
     return '<div class="code-block">'
-      + `<div class="code-block-heading"><span>${heading}</span></div>`
+      + `<div class="${headingStyle}"><span>${renderedHeading}</span></div>`
       + `<div class="code-block-content">${codeBlockContent}</div>`
       + '</div>';
   }
--- a/asset/css/markbind.css
+++ b/asset/css/markbind.css
@@ -46,6 +46,10 @@ pre > code.hljs[heading] {
     text-align: right;
 }
 
+.inline-markdown-heading {
+    line-height: 1.5;
+}
+
 .code-block-content {
     clear: both;
     display: block;
--- a/test/functional/test_site/expected/markbind/css/markbind.css
+++ b/test/functional/test_site/expected/markbind/css/markbind.css
@@ -46,6 +46,10 @@ pre > code.hljs[heading] {
     text-align: right;
 }
 
+.inline-markdown-heading {
+    line-height: 1.5;
+}
+
 .code-block-content {
     clear: both;
     display: block;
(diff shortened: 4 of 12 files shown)

### CODE REVIEW

File: src/lib/markbind/src/lib/markdown-it/index.js [hunk 1]
```diff
@@ -109,8 +109,10 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
   const codeBlockContent = `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
   if (heading) {
-    return '<div class="code-block">'
-      + `<div class="code-block-heading"><span>${heading}</span></div>`
+    const renderedHeading = markdownIt.renderInline(heading);
+    const headingStyle = (renderedHeading === heading) ? 'code-block-heading' : 'code-block-heading-with-inline';
+      return '<div class="code-block">'
+      + `<div class="${headingStyle}"><span>${renderedHeading}</span></div>`
       + `<div class="code-block-content">${codeBlockContent}</div>`
       + '</div>';
```
nbriannl: is there a better way than this?

should i just make a `.increased-line-height` class and apply it after I apply the base `.code-block-heading` class

ang-zeyu: Could try this https://stackoverflow.com/questions/6078051/html-how-to-keep-spans-with-padding-and-borders-from-overlapping/38270183

nbriannl: With
image

image

The height of non-markdown headings have been increased as well. investigating further.

ang-zeyu: It should be `.code-block-heading * { display: ... }`

nbriannl: Hm, nevermind, i found an alternative.

docs/userGuide/syntax/code.mbdf
```diff
@@ -125,6 +125,25 @@ To add a heading, add the attribute `heading` with the heading text as the value
 </span>
 </include>
 
+Headings support inline MarkDown some exceptions (i.e. inline code, dim text style, block quotes, headers, lists, paragraphs, footnotes and embeds).
```
ang-zeyu: `with some exceptions`
Could only mention the inline exclusions ( inline code / dim )

src/lib/markbind/src/lib/markdown-it/index.js
```diff
@@ -109,8 +109,10 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   const heading = token.attrGet('heading');
   const codeBlockContent = `<pre><code ${slf.renderAttrs(token)}>${str}</code></pre>`;
   if (heading) {
-    return '<div class="code-block">'
-      + `<div class="code-block-heading"><span>${heading}</span></div>`
+    const renderedHeading = markdownIt.renderInline(heading);
+    const headingStyle = (renderedHeading === heading) ? 'code-block-heading' : 'code-block-heading inline-markdown-heading';
+      return '<div class="code-block">'
```
ang-zeyu: indent here should be 2 to the left, is github displaying it wrongly?

nbriannl: No you're right. The pains of not having eslint consider this file. 😞

docs/userGuide/syntax/code.mbdf
```diff
@@ -125,6 +125,25 @@ To add a heading, add the attribute `heading` with the heading text as the value
 </span>
 </include>
 
+Headings support inline MarkDown, except for `Inline Code` and %%Dim%% text styles.
```
openorclose: I think we use 'Markdown' in all other places.
```suggestion
Headings support inline Markdown, except for `Inline Code` and %%Dim%% text styles.
```

nbriannl: I see, thank you!

damithc:
> Just wondering - what if the user wants to use symbols that are part of markdown syntax in the heading? (i.e. user wants *Title* rather than **Title**)

In fact, that scenario is not limited to headings, but applies to anywhere markdown syntax is used, right?

marvinchin:
> In fact, that scenario is not limited to headings, but applies to anywhere markdown syntax is used, right?

Yup, I think that should be the case.

I did some digging around and it seems like we already have a page on escaping markdown - my bad for not noticing it before:
https://markbind.org/userGuide/tipsAndTricks.html#tip-escaping-characters

I think my main concern here would be whether or not introducing this change would lead to "breaking" existing users' code block headings which use characters from the markdown syntax.

nbriannl:
Noting conflicts. I'll resolve it asap.

nbriannl:
Rebased on master, ready for merge

Issue #967, opened 2019-12-28
Unexpected behaviours of heading indexing in panels


ang-zeyu:
**Tell us about your environment**

* **MarkBind Version:** 2.8.0

**What did you do? Please include the actual source code causing the issue.**

Is the current behaviour of header indexing in panels intended?

When there is no `header` specified, and there is a `slot="header"` specified with some level of markdown heading inside, there are two cases:
- `expanded` is specified : the heading is indexed, as expected, but there are now two identical `ids` in the dom (one for the `cardContainer`, the other from the `slot`. ( not that its really an issue ) ( point 1 )
- `expanded` is not specified: the slot heading is not indexed ( This leads to point `2` below ), and there are still two identical `ids`.

On the other hand, if a `header` is specified for the `panel`, then the code to search for another header in the `headerWrapper` will not execute ( second branch in `mounted()` ).

`mounted()` in PanelBase.js
```js
...
} else if (this.$refs.headerWrapper.innerHTML) {
  const header = jQuery(this.$refs.headerWrapper.innerHTML).wrap('<div</div>').parent().find(':header');
  if (header.length > 0) {
    this.$refs.cardContainer.setAttribute('id', header[0].id);
  }
}
```

- If said `header` is not a md header ( `### ...` ), then the code will not execute ( with no implications, as far as I can see ) ( point 1 )
- The behaviour for the `slot header` is similar to above, barring the duplicate `id` ( since the second branch in `mounted()` is no longer executed ).
  - if `unexpanded`, the heading will not be indexed
  - if `expanded` is specified, and the `header="..."` is a valid md header, then the slot header, along with the overwritten `header` are indexed. ( point 3 )

**What did you expect to happen?**

1. It seems that the second branch is unnecessary, and it would only result in a duplicate id currently. ( I might be missing something )

Sample code producing the issue ( no `header` attribute, so the second branch is executed ):
```html
<panel type="info">
  <div slot="header">
  
  # some header
  </div>
  .........
</panel>
```

Result (duplicate ids) (removing the second branch in `mounted()` only removes the id from the cardContainer, which dosen't break the anchor since there is still one in the heading.

![duplicateIdsDom](https://user-images.githubusercontent.com/3306138/71552508-76a13400-2a39-11ea-9235-df32782575b6.png)

2. The slot heading is not indexed when `expanded` is not specified.

Result from the same code for point 1: ( adding `expanded` resolves it )
[...]

PR #1145, opened 2020-03-25
Fix anchor navigation in panel headers

### LINKED ISSUE #1071: Anchor navigation in panel not working properly with fixed header

* **MarkBind Version:**2.11.0

```html
<panel header="## Hello">
</panel>
```

### LINKED ISSUE #1104: Heading hidden when clicking pageNav
pageNav-glitch

le0tan:

• [X] Documentation update
• [X] Bug fix

Fixes #1104 Fixes #1071 

Re-implement the anchor navigation fix for fixed headers using pseudo selectors, so that no runtime generation of dummy spans is no longer needed.

Instead of adding a invisible dummy span, we introduce the empty space using `::before` css selector. For the special case of panel headers, we add `::before` to the container so that when the `scrollToUrlAnchorHeading` function locates the target as the container, the offset is correctly applied.

This change should only affect the behavior when `class="header-fixed"` is set in `header.md`'s `<header>`. Test if there are unexpected issues of anchor navigation.

**Important**: I applied the css to `.card-container`, not sure if there are any other vue-strap components that uses this class name and is affected unexpectedly.

**Testing instructions:**

Replace the vue-strap file from https://github.com/MarkBind/vue-strap/pull/141

To avoid generating lots of dummy spans at runtime when fixed header is
enabled, and to fix the bug where anchor navigation doesn't work when
the panel is minimized, we re-implement the anchor nav fix using
::before selector.

Let's introduce the invisible empty space by applying styles to
::before of h tags and card-containers.

Let's disable the pointer-events of the card-container to prevent users
from clicking on the invisible empty space; and force the
pointer-events to all on the button space of the panel. (See the
companion vue-strap commit)


ang-zeyu:
Have you tried the solution adopted here? ( notice the `::before` )
https://getbootstrap.com/docs/4.1/getting-started/introduction/

I tried it briefly on the inspector, it seems to work well as a css-only solution for both minimized and expanded panel forms:
panel-id

Could this be used instead of creating spans at runtime?

le0tan:
@ang-zeyu you were adding `::before` to the card container instead of the `h` tag - but that's indeed a good idea. I'm trying to work out an impl using it.

ang-zeyu:
> @ang-zeyu you were adding `::before` to the card container instead of the `h` tag

This is due to how we assign the id of the header to the root of the panel ( `card-container` ) as well, which is necessary as a panel's `alt` may not necessarily be the `header` as well. 

You can refer to #967 for a more in depth disccussion.

> Edit: I added `::before` to both `<h>` tags and `span` tags. Can you kindly review before I resolve conflicts in the test sites?

👍 will take a look in a while

ang-zeyu:
Seems good, for headings inside panels, you would also need to add `pointer-events: none` css ( in vue-strap ) or something of the sort though, to prevent the user from being able to click the `::before` element and expand the panel.

le0tan:
> Seems good, for headings inside panels, you would also need to add `pointer-events: none` css ( in vue-strap ) or something of the sort though, to prevent the user from being able to click the `::before` element and expand the panel.

Good catch! Will rectify this shortly. Thanks!

Issue #1141, opened 2020-03-23
Search shows duplicate for User Guide: Getting Started


Tejas2805:
**Tell us about your environment**
MacOS

* **MarkBind Version:**
Latest master (for developers)

**What did you do? Please include the actual source code causing the issue.**

N.A.

**What did you expect to happen?**
Expected it to appear only once. 

**What actually happened? Please include the actual, raw output.**
![Screen Shot 2020-03-24 at 2 46 07 AM](https://user-images.githubusercontent.com/35946746/77352283-eb0a1b80-6d79-11ea-8de0-f4ff27b31a48.png)

Tejas2805:
Soled in #1147

PR #1108, opened 2020-03-09
Fix bug in search for dev guide

Tejas2805:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [x] Bug fix

Fixes #1107 

**What is the rationale for this request?**
It becomes confusing to see the same results twice. 

**What changes did you make? (Give an overview)**

Changed the navigation layout for developer guide and maintainer guide

**Provide some example code that this change will affect:**

```html
<navigation>

<span class="lead">**Developer Guide**</span>

* [Maintainer Guide]({{baseUrl}}/devGuide/maintainerGuide.html)
</navigation>
```

**Is there anything you'd like reviewers to focus on?**
N.A.

**Testing instructions:**
Can check deployed

**Proposed commit message: (wrap lines at 72 characters)**

PR #1147, opened 2020-03-25
Bug in Search for UG and DG

### LINKED ISSUE #1107: Search shows duplicate for Developer Guide
MacOS

* **MarkBind Version:**
Latest master (for developers)

N.A.

Expected it to appear only once. 

Screen Shot 2020-03-09 at 3 35 38 PM

Tejas2805:

• [x] Bug fix

Fixes #1107 and #1141 

It becomes confusing to see the same results twice. 

Changed the navigation layout for developer guide and maintainer guide

```html
<navigation>

<span class="lead">**Developer Guide**</span>

* [Maintainer Guide]({{baseUrl}}/devGuide/maintainerGuide.html)
</navigation>
```

N.A.

**Testing instructions:**
Can check deployed


docs/devGuide/index.md
```diff
@@ -1 +1,3 @@
-<include src="devGuide.md" />
\ No newline at end of file
+<script>
+  window.location.href = "devGuide.html"
+</script>
```
Tejas2805: This seems to work.

But the issue is: The back button on the browse doesn't work on this page anymore. Every time you press the back button it brings you to the same page. This happens due to re-direction I feel.

@ang-zeyu Your POV on this.

ang-zeyu: Hmm what browser is this? It seems to work fine for me on chrome, edge and firefox.
Did a little reading, `location.href =` should not add a new entry if it is not user initiated.

Otherwise you could try a more explicit version https://developer.mozilla.org/en-US/docs/Web/API/Location/replace

Tejas2805: I tried on Google Chrome too. But wasn't going back for some reason.

Tejas2805: I tried on Google Chrome again after closing the browser. Works okay!

Tejas2805:
@marvinchin Can I get your review on this?

Tejas2805:
> It seems like the root cause of the problem is that we have 2 separate pages with exactly the same content. Should we just consolidate them into one page instead? Maybe we should remove either the index page or the getting started page (I think I'd prefer to keep the index page)?

Can do. This was the initial approach we had in mind. Then we decided to go ahead with re-direction. Will remove the other pages and keep the index page only.

marvinchin:
> Can do. This was the initial approach we had in mind. Then we decided to go ahead with re-direction. Will remove the other pages and keep the index page only.

Hmm do you remember what the rationale for deciding on redirection was? Just wondering in case I didn't consider that.

Tejas2805:
I think the Prof recommended it, so we decided to go ahead with it. No other reason as such. 

You can look into this #1108 for a more detailed conversation.

Do let me know how you want me to proceed.

yamgent:
I think that a redirection will ensure that old links to either one of the pages will still work.

PR #1155, opened 2020-03-28
Parse popovers for footnotes

### LINKED ISSUE #1154: Footnotes don't show popovers

* **MarkBind Version: 2.12.0**

```md

**Normal footnotes:**
Here is a footnote reference,[^1] and another.[^longnote]
​
[^1]: Here is the footnote. Footnotes will appear at the bottom of the page.
​
[^longnote]: Here's one with multiple blocks.
​
    Subsequent paragraphs are indented to show that they
belong to the previous footnote.


**Inline footnotes:**
Here is an inline note.^[Inlines notes are easier to write, since
you don't have to pick an identifier and move down to type the
note.]

```

The footnotes could be hovered over to get a popover.

https://markbind.org/userGuide/formattingContents.html#footnotes

Popovers don't appear.

This is because the converting from markbind popovers to bootstrap-vue happens before plugins, and the footnotes popovers do not get parsed then.

openorclose:

• [x] Bug fix

Fixes #1154 

Specifically called the componentParser on the `popover`s in the footnotes plugin

```md

**Normal footnotes:**
Here is a footnote reference,[^1] and another.[^longnote]
​
[^1]: Here is the footnote. Footnotes will appear at the bottom of the page.
​
[^longnote]: Here's one with multiple blocks.
​
    Subsequent paragraphs are indented to show that they
belong to the previous footnote.


**Inline footnotes:**
Here is an inline note.^[Inlines notes are easier to write, since
you don't have to pick an identifier and move down to type the
note.]

```

This is the simple bug fix for now. Long term solution to fix footnotes inside dynamic panels is still in the works.

**Testing instructions:**

Check that the popovers appear.

```
During the migration to bootstrap-vue, popovers had to be parsed into bootstrap-vue syntax.

This is done before the plugins stage, where footnotes are implemented, so the popovers don't work there.

Let's add the parsing of popovers to the footnotes plugin to solve this.
```


test/functional/test_site/expected/index.html
```diff
@@ -536,25 +536,27 @@ <h6 class="always-index" id="level-6-header-outside-headingsearchindex-with-alwa
...
-              </p>
-            </div>
-          </popover>
+          <span id="pop:footnote3" data-mb-component-type="popover" v-b-popover.hover.top.html="popoverInnerGenerator" class="trigger">
+          <div data-mb-html-for="content">
+            <p>Inlines notes are easier to write, since
+you don't have to pick an identifier and move down to type the
+note.</p>
+          </div>
+        </span>
```
nbriannl: weird that the indents are incorrect, but i assume this is what `updatetest` generated, so can't really argue with this, can I?

openorclose: Yeah..

yamgent:
@openorclose resolve conflicts

openorclose:
> @openorclose resolve conflicts

Done!

PR #1175, opened 2020-04-01
Add SiteConfig model

ang-zeyu:

• [x] Other, please explain: Code maintainability

To consolidate all default value initialisation of the site config into a new model

- Added a new SiteConfig model
- Moved all lazy 'default value' initialisations to this model

na

**Testing instructions:**
- `npm run test` should pass

Add SiteConfig model

The undefined properties of the site configuration object of Site are
lazily initialised throughout site generation.
This decreases the maintainability of site configuration logic.

Let’s consolidate all such initialisation to a new SiteConfig model.

### DIFF
--- a/src/Site.js
+++ b/src/Site.js
@@ -33,6 +33,7 @@ const delay = require('./util/delay');
 const FsUtil = require('./util/fsUtil');
 const logger = require('./util/logger');
 const Page = require('./Page');
+const SiteConfig = require('./SiteConfig');
 const Template = require('./template/template');
 
 const CLI_VERSION = require('../package.json').version;
@@ -44,7 +45,6 @@ const {
   CONFIG_FOLDER_NAME,
   FAVICON_DEFAULT_PATH,
   FOOTER_PATH,
-  HEADING_INDEXING_LEVEL_DEFAULT,
   INDEX_MARKDOWN_FILE,
   LAYOUT_DEFAULT_NAME,
   LAYOUT_FOLDER_PATH,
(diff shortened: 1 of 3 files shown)


src/Site.js
```diff
@@ -197,21 +197,17 @@ class Site {
     return false;
   }
 
-  readSiteConfig(baseUrl) {
-    return new Promise((resolve, reject) => {
+  async readSiteConfig(baseUrl) {
```
openorclose, line 200: Don't think there's an async operation going on here, can we remove it?

ang-zeyu: The `serve` command in `index.js` expects the async nature unfortunately, didn't want to touch that here =X

le0tan:
This was what I wanted to do to improve the type checking of the code base (actually the branch name for https://github.com/MarkBind/markbind/pull/1039 is `config-object` because I used to extract those type definitions as separate classes =)

ang-zeyu:
> This was what I wanted to do to improve the type checking of the code base (actually the branch name for #1039 is `config-object` because I used to extract those type definitions as separate classes =)

Ah I see, didn't notice the branch name 😆

I only intended for this to be a consolidation of where site config properties are initialised, instead of separating them all over `Site.js`. Will place the appropriate jsDoc for the model once #1039 is merged

ang-zeyu:
Ready for review again, updated with jsDocs

src/Site.js
```diff
@@ -128,7 +128,10 @@ class Site {
     this.baseUrlMap = new Set();
     this.forceReload = forceReload;
     this.plugins = {};
-    this.siteConfig = {};
+    /**
+     * @type {undefined | SiteConfig}
```
le0tan, line 132: Is it actually possible for the siteConfig to be `undefined`?

ang-zeyu: upon initial initialization:
> `this.siteConfig = undefined;`

this is done to defend against using an empty `siteConfig`, as it is lazily initialised. ( ensure a call to `readSiteConfig` )

Since we can't have async methods in constructors, a future PR can deal with this, adopting a factory method approach to constructing `Site` that does all the io boilerplate, ensuring `siteConfig` is initialised when constructing `Site`

PR #1176, opened 2020-04-02
Use <br> to separate lines of code block

### LINKED ISSUE #1125: Copying code in code blocks not working properly

* **MarkBind Version: 2.12**

Ubuntu 18.04

````
line1
line2
line3
````

then `markbind serve` (or copy any of the code examples from https://markbind.org/userGuide/fullSyntaxReference.html)
When pasting into a text editor, line breaks are preserved.

```
line1line2line3
```

### LINKED ISSUE #1173: Extra unprintable chars in code blocks
When I copy-paste python code from this section, intellij indicates there are some unprintable chars.
https://nus-te3201.github.io/2020/schedule/week10/topics-py.html#appendix-d-unit-testing

image

I wonder if they have been added by MarkBind. The code in the source file doesn't seem to have the same chars. https://github.com/nus-te3201/2020/blob/te3201/programming/unittesting/text.md

### PR DESCRIPTION
**Pull request by openorclose**

• [x] Bug fix

Fixes #1173 
Fixes #1125 

In #991 , to handle empty lines, an unprintable character was added to make the span take up space. This causes problems for code copying.

Instead of that, let's make the span be `inline-block` and use `<br>` to put them on new lines.

````
```
some code





new lines above
```
````

This should now display correctly, and not have any new line characters inside.

**Testing instructions:**

Check that multiple lines work, and that copying them also works

```
Use <br> and inline-block to separate spans.

Adding extra unprintable characters cause code copying to fail.

Let's not use them, and manually separate
them by using <br> instead.
```


ang-zeyu:
Different fix may be needed:

code fence

https://deploy-preview-1176--markbind-master.netlify.com/userguide/formattingcontents#line-highlighting

openorclose:
> Different fix may be needed:
> 
> code fence
> 
> https://deploy-preview-1176--markbind-master.netlify.com/userguide/formattingcontents#line-highlighting

Oh what browser is that?

I'm using firefox 73 and it looks ok:

image

ang-zeyu:
> Oh what browser is that?

chrome v80, seems to be broken on edge ( I'm on a pretty old version though ) as well

openorclose:
> > Oh what browser is that?
> 
> chrome v80, seems to be broken on edge ( I'm on a pretty old version though ) as well
> 
> Looks fine on firefox for me as well

Yeah, have to think of a different fix...

Changing empty lines to a single space probably wouldn't work also, since these extra spaces might affect how the program runs.

alyip98:
@openorclose does this [fix](https://github.com/alyip98/markbind/tree/fix-copying) work?

openorclose:
Yeah seems to work. Thanks!

openorclose:
> OK. Just a non-blocking question, can't remember whether this was asked: why did we use `<span>` instead of `<div>`?

I chose it rather arbitrarily, but if i were to give a reason it would be that `<span>` can only contain inline elements which seems more appropriate for showing a line of code.

jamessspanggg:
@openorclose code copying for safari:

Kapture 2020-04-10 at 23 33 06

PR #1187, opened 2020-04-10
Fix anchor navigation in panel headers

### LINKED ISSUE #1071: Anchor navigation in panel not working properly with fixed header

### LINKED ISSUE #1104: Heading hidden when clicking pageNav

le0tan:

• [X] Documentation update
• [X] Bug fix

Replaces #1145 
Fixes #1104 Fixes #1071

Move span generation to a plugin during build time. Add `::before` to panel card containers to make navigation work properly on panel headers.

na

Test if the panel functions are broken in any way.

**Testing instructions:**

na

To avoid dynamically injecting lots of dummy spans in a document with
many headings, let's generate these spans using a plugin, post
rendering.

To make anchor navigation function properly on panel headers, let's
introduce ::before element to `card-container` to make space for the
fixed top nav.


src/plugins/fixedHeader.js
```diff
@@ -0,0 +1,19 @@
+const cheerio = module.parent.require('cheerio');
+
+module.exports = {
```
ang-zeyu, line 3: don't think we should do this as a plugin since it is a core markbind feature.
You could look to `insertHeaderFile` and `_parse` to integrate this directly

le0tan: I can do that in `postRender` - the same place where plugins are applied. Will there be extra concerns by doing that?

ang-zeyu: mainly of code organization and performance secondly ( as putting it there means you are parsing the entire `content` one more time, keeping the number of parses to a minimum would be nice )

Unless absolutely necessary ( like in the case of id resolution in popovers ), I think each stage of page generation should maintain a single unit of functionality as is currently. `postRender` is for plugins - you could certainly add it immediately outside of `postRender`, but adding one entire stage of parsing just for fixed navbars seems a little bit of waste, and there is already similar parsing going on in `_parse/insertHeaderFile`. Would make sense to put it there from both standpoints

le0tan: If we're to implement this inside `_parse`, fixed header would be a config option in `site.json` I suppose?

ang-zeyu: you could still 'detect' it during `insertHeaderFile` - though just to confirm the current implementation is `<header class="header-fixed">` right? 😮

Issue #931, opened 2019-07-14
Cannot use nunjucks in a sub-site


damithc:
Problem: if a page in a subsite uses nunjucks, the scripts will not work if that page is used in the outer site.

ang-zeyu:
Hmm what script produced this?

I've tested with macros, if-else's, for loops and they seem to work fine

damithc:
This was just a guess, because nunjucks import statements need to specify the path from the root of the site. Perhaps it's already taken care of.
Are you able to get the script to work in the sub-site (i.e., when you generate the site for the sub-site without the outer site) and also when the sub-site included in the outer site.

ang-zeyu:
> This was just a guess, because nunjucks import statements need to specify the path from the root of the site. Perhaps it's already taken care of.

Will test imports out as well

> Are you able to get the script to work in the sub-site (i.e., when you generate the site for the sub-site without the outer site) and also when the sub-site included in the outer site.

Yup, I used a markbind `<include>` though, will experiment with nunjucks includes.

ang-zeyu:
Tested the following: ✅ 
- markbind `<include>` a sub site's md/html file that uses a mix of for loops, macros on its own
- `{% import %}` ing a sub_site's md/html from the main site, and then using it
- `{% include %}` ing a sub_site's md / html file from the main site that uses a mix of for loops, ...
  - `baseUrl` resolves to the main site unlike markbind includes as expected
  - output is as expected
- `{% import %}` from a sub site's file, then `<include>` / `{% include %}` it in the main site

What fails is the following as mentioned:
> script to work in the sub-site (i.e., when you generate the site for the sub-site without the outer site) and also when the sub-site included in the outer site.

ie. `{% include/import xx %}` in a sub site, where xx is from the root of the sub site.
It works when building the sub_site alone, but fails when including it in the main site as `xx` is resolved from the main site.

Will see if it can't be fixed

damithc:
possibly related #684

damithc:
This is actually a serious flaw that went undetected because our only sub-site (se-edu/se-book) hasn't uses nunjucks up to now. It's good if we an fix it soon.

ang-zeyu:
> possibly related #684

The solution used in it is related, but instead of configuring one nunjucks instance per page, it should be one per (sub)site.

[...]

Issue #894, opened 2019-06-26
Allow includes in variables.md


damithc:
Suggestion: allow the `_markbind/variables.md` to use `include` directives.

Rationale: `variables.md` of the outer site can `include` the `variables.md` files of sub-sites, instead of having to define similar variables again. This is especially useful when the set of icons used by the site is defined as variables and the outer site wants to use the same set of icons, for consistency.

PR #1189, opened 2020-04-10
Reorganise variable processing

ang-zeyu:

• [x] Other, please explain:

Yet another part of #810 

Prelude to #931, #894 type of prs - my solution for #931 added quite some boilerplate, which I don't really think is sustainable anymore -- this should be a good point to refactor it

- variable processing is rather decentralized and unmaintainable currently, some examples:
  - `Site.js` requires the entire `Parser` abstraction to reset site variables during live reload
  - massive 70-100+ line functions like `renderIncludeFile/extractPageVariables/...`
  - a lot of boilerplate code setup with minor variations e.g.

```js
const userDefinedVariables = config.userDefinedVariablesMap[parentSitePath];
// Extract included variables from the PARENT file
const includeVariables = Parser.extractIncludeVariables(node, context.variables);
// Extract page variables from the CHILD file
const pageVariables = Parser.extractPageVariables(asIfAt, fileContent,
                                                       userDefinedVariables, includeVariables);
const content = njUtil.renderRaw(fileContent, {
  ...pageVariables,
  ...includeVariables,
  ...userDefinedVariables,
}, {
  path: filePath,
});
```
    

- Introduced a new abstraction, `variablePreprocessor`, which will be responsible for all variable storage and rendering

This doodle explains the intended endgame better (pink lines are read as 'uses'):

There are two logical parts to this refactor

The first (commit):

Completes and unifies where site variables are accessed and stored ( the first, second pink lines, and part of the fourth pink line )
- Some misc refactors like SLAP, guard clauses, better in code documentation, etc. are also made
- methods relating to functions 3 / 4 are simply moved from `parser` to `variablePreprocessor` without much change ( just some misc refactors like SLAP, etc per above )

The second (commit):

Completes and logically changes functions 3 / 4.

The variable processing flow before was such that `<import>`s and locally declared `<variable>`s were extracted and rendered separately. This inevitably results in more boilerplate ( e.g. note double extraction in the code above ), and uses dirty hacks like `Proxy` to force nunjucks to output namespaced variables back into themselves.

This commit unifies the two processes hence (see `extractPageVariables` / `renderPage` ), extracting and rendering both in one go while respecting the variable declaration order still. It also removes the need for `Proxy`s, and has some modest performance benefits due to the reduced number of passes.

**End goals:**

1. The main goal is to reduce repeated boilerplate in the example code above throughout and centralise all rendering, note how the above is turned into simply 1 / 2 lines. ( mostly 1 )

```
this.variablePreprocessor.extractPageVariables(file, content);
const renderedContent = this.variablePreprocessor.renderPage(file, content, {}, additionalVariables);
```

2. Misc
   - Better separation of concerns #810 
   - More unit focused code
   - Better in code documentation ( results in most of the added lines )
   - Better performance ( in part 2 - consistently about 10% due to the variable processing logic changes as mentioned above )

The main motive for this is to solve #931 type of issues easily, and hopefully make variable processing development in the future easier.
With this in mind, could the architecture be further improved?

as an example for #931, with all site variable rendering now centralised, by intended solution is to let `variablePreprocessor` hold multiple nunjucks instances in the future, each configured with its root path to be the respective root paths of the (sub)sites.
This should allow any nunjucks tags that uses paths to resolve correctly across (sub)sites

**Testing instructions:**
- 2103 site should have no diffs, save for logs and timestamps
- all functional / unit tests should pass

(pending review of the commit structure)

Reorganise variable processing (part 1)

Processing variables involves a lot of boilerplate code with minor
variations. This decentralizes variable processing, decreasing code
maintainability and readability.

Let's embark on a two stage refactor to improve variable processing.
In this first refactor, we create the variablePreprocessor abstraction,
and completely migrate the use of userDefinedVariablesMap for site
variables to variablePreprocessor, reducing boilerplate code relating
to site variables.
Additionally, we move all page variable processing related methods to
variablePreprocessor from Parser, abstracting such methods into
smaller units of functionality where possible, and enhancing in code
documentation.

In the second part, we will similarly refactor page variable processing
methods to reduce more boilerplate, increasing code maintainability.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -80,7 +80,7 @@ class Page {
    * @property {string} pageTemplate template used for this page
    * @property {string} title
    * @property {string} titlePrefix https://markbind.org/userGuide/siteConfiguration.html#titleprefix
-   * @property {Object<string, any>} userDefinedVariablesMap
+   * @property {VariablePreprocessor} variablePreprocessor
    * @property {string} sourcePath the source file for rendering this page
    * @property {string} tempPath the temp path for writing intermediate result
    * @property {string} resultPath the output path of this page
@@ -170,9 +170,9 @@ class Page {
      */
     this.titlePrefix = pageConfig.titlePrefix;
     /**
-     * @type {Object<string, any>}
+     * @type {VariablePreprocessor}
      */
-    this.userDefinedVariablesMap = pageConfig.userDefinedVariablesMap;
+    this.variablePreprocessor = pageConfig.variablePreprocessor;
 
     /**
      * The source file for rendering this page
@@ -631,10 +631,9 @@ class Page {
     }
     // Set header file as an includedFile
     this.includedFiles.add(headerPath);
-    // Map variables
-    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${njUtil.renderRaw(headerContent, userDefinedVariables)}\n${pageData}`;
+
+    const renderedHeader = this.variablePreprocessor.renderSiteVariables(this.sourcePath, headerContent);
+    return `${renderedHeader}\n${pageData}`;
   }
 
   /**
@@ -661,10 +660,9 @@ class Page {
     const footerContent = fs.readFileSync(footerPath, 'utf8');
     // Set footer file as an includedFile
     this.includedFiles.add(footerPath);
-    // Map variables
-    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${pageData}\n${njUtil.renderRaw(footerContent, userDefinedVariables)}`;
+
+    const renderedFooter = this.variablePreprocessor.renderSiteVariables(this.sourcePath, footerContent);
+    return `${pageData}\n${renderedFooter}`;
   }
 
   /**
@@ -695,10 +693,7 @@ class Page {
     }
     this.includedFiles.add(siteNavPath);
 
-    // Render variables
-    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    const siteNavMappedData = njUtil.renderRaw(siteNavContent, userDefinedVariables);
+    const siteNavMappedData = this.variablePreprocessor.renderSiteVariables(this.sourcePath, siteNavContent);
 
     // Check navigation elements
     const $ = cheerio.load(siteNavMappedData);
@@ -879,11 +874,9 @@ class Page {
       const headFileContent = fs.readFileSync(headFilePath, 'utf8');
       // Set head file as an includedFile
       this.includedFiles.add(headFilePath);
-      // Map variables
-      const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-      const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-      const headFileMappedData = njUtil.renderRaw(headFileContent, userDefinedVariables)
-        .trim();
+
+      const headFileMappedData = this.variablePreprocessor.renderSiteVariables(this.sourcePath,
+                                                                               headFileContent).trim();
       // Split top and bottom contents
       const $ = cheerio.load(headFileMappedData, { xmlMode: false });
       if ($('head-top').length) {
@@ -941,7 +934,7 @@ class Page {
    * @typedef {Object<string, any>} FileConfig
    * @property {Set<string>} baseUrlMap the set of urls representing the sites' base directories
    * @property {string} rootPath
-   * @property {Object<string, any>} userDefinedVariablesMap
+   * @property {VariablePreprocessor} variablePreprocessor
    * @property {Object<string, number>} headerIdMap
    * @property {boolean} fixedHeader indicates whether the header of the page is fixed
    */
@@ -949,15 +942,15 @@ class Page {
   generate(builtFiles) {
     this.includedFiles = new Set([this.sourcePath]);
     this.headerIdMap = {}; // Reset for live reload
-
-    const markbinder = new MarkBind();
+    const markbinder = new MarkBind({
+      variablePreprocessor: this.variablePreprocessor,
+    });
     /**
      * @type {FileConfig}
      */
     const fileConfig = {
       baseUrlMap: this.baseUrlMap,
       rootPath: this.rootPath,
-      userDefinedVariablesMap: this.userDefinedVariablesMap,
       headerIdMap: this.headerIdMap,
       fixedHeader: this.fixedHeader,
     };
@@ -1264,11 +1257,12 @@ class Page {
        * We create a local instance of Markbind for an empty dynamicIncludeSrc
        * so that we only recursively rebuild the file's included content
        */
-      const markbinder = new MarkBind();
+      const markbinder = new MarkBind({
+        variablePreprocessor: this.variablePreprocessor,
+      });
       return fs.readFileAsync(dependency.to, 'utf-8')
         .then(result => markbinder.includeFile(dependency.to, result, {
           baseUrlMap: this.baseUrlMap,
-          userDefinedVariablesMap: this.userDefinedVariablesMap,
           rootPath: this.rootPath,
           cwf: file,
         }))
--- a/src/Site.js
+++ b/src/Site.js
@@ -6,7 +6,7 @@ const path = require('path');
 const Promise = require('bluebird');
 const ProgressBar = require('progress');
 const walkSync = require('walk-sync');
-const MarkBind = require('./lib/markbind/src/parser');
+const VariablePreprocessor = require('./lib/markbind/src/preprocessors/variablePreprocessor');
 const njUtil = require('./lib/markbind/src/utils/nunjuckUtils');
 const injectHtmlParser2SpecialTags = require('./lib/markbind/src/patches/htmlparser2');
 const injectMarkdownItSpecialTags = require(
@@ -134,7 +134,9 @@ class Site {
      */
     this.siteConfig = undefined;
     this.siteConfigPath = siteConfigPath;
-    this.userDefinedVariablesMap = {};
+
+    // Site wide variable preprocessor
+    this.variablePreprocessor = undefined;
 
     // Lazy reload properties
     this.onePagePath = onePagePath;
@@ -271,7 +273,7 @@ class Site {
       title: config.title || '',
       titlePrefix: this.siteConfig.titlePrefix,
       headingIndexingLevel: this.siteConfig.headingIndexingLevel,
-      userDefinedVariablesMap: this.userDefinedVariablesMap,
+      variablePreprocessor: this.variablePreprocessor,
       sourcePath,
       resultPath,
       asset: {
@@ -513,6 +515,7 @@ class Site {
       .map(x => path.resolve(this.rootPath, x));
 
     this.baseUrlMap = new Set(candidates.map(candidate => path.dirname(candidate)));
+    this.variablePreprocessor = new VariablePreprocessor(this.rootPath, this.baseUrlMap);
 
     return Promise.resolve();
   }
@@ -523,12 +526,9 @@ class Site {
   collectUserDefinedVariablesMap() {
     // The key is the base directory of the site/subsites,
     // while the value is a mapping of user defined variables
-    this.userDefinedVariablesMap = {};
-    const markbindVariable = { MarkBind: MARKBIND_LINK_HTML };
+    this.variablePreprocessor.resetUserDefinedVariablesMap();
 
     this.baseUrlMap.forEach((base) => {
-      const userDefinedVariables = {};
-      Object.assign(userDefinedVariables, markbindVariable);
       const userDefinedVariablesPath = path.resolve(base, USER_VARIABLES_PATH);
       const userDefinedVariablesDir = path.dirname(userDefinedVariablesPath);
       let content;
@@ -539,15 +539,17 @@ class Site {
         logger.warn(e.message);
       }
 
-      // This is to prevent the first nunjuck call from converting {{baseUrl}} to an empty string
-      // and let the baseUrl value be injected later.
-      userDefinedVariables.baseUrl = '{{baseUrl}}';
-      this.userDefinedVariablesMap[base] = userDefinedVariables;
+      /*
+       This is to prevent the first nunjuck call from converting {{baseUrl}} to an empty string
+       and let the baseUrl value be injected later.
+       */
+      this.variablePreprocessor.addUserDefinedVariable(base, 'baseUrl', '{{baseUrl}}');
+      this.variablePreprocessor.addUserDefinedVariable(base, 'MarkBind', MARKBIND_LINK_HTML);
 
       const $ = cheerio.load(content);
-      $('variable,span').each(function () {
-        const name = $(this).attr('name') || $(this).attr('id');
-        const variableSource = $(this).attr('from');
+      $('variable,span').each((index, element) => {
+        const name = $(element).attr('name') || $(element).attr('id');
+        const variableSource = $(element).attr('from');
 
         if (variableSource !== undefined) {
           try {
@@ -555,18 +557,13 @@ class Site {
             const jsonData = fs.readFileSync(variableFilePath);
             const varData = JSON.parse(jsonData);
             Object.entries(varData).forEach(([varName, varValue]) => {
-              // Process the content of the variable with nunjucks, in case it refers to other variables.
-              const variableValue = njUtil.renderRaw(varValue, userDefinedVariables, {}, false);
-
-              userDefinedVariables[varName] = variableValue;
+              this.variablePreprocessor.renderAndAddUserDefinedVariable(base, varName, varValue);
             });
           } catch (err) {
             logger.warn(`Error ${err.message}`);
           }
         } else {
-          // Process the content of the variable with nunjucks, in case it refers to other variables.
-          const html = njUtil.renderRaw($(this).html(), userDefinedVariables, {}, false);
-          userDefinedVariables[name] = html;
+          this.variablePreprocessor.renderAndAddUserDefinedVariable(base, name, $(element).html());
         }
       });
     });
@@ -691,7 +688,7 @@ class Site {
     const filePathArray = Array.isArray(filePaths) ? filePaths : [filePaths];
     const uniquePaths = _.uniq(filePathArray);
     logger.info('Rebuilding affected source files');
-    MarkBind.resetVariables();
+    this.variablePreprocessor.resetVariables();
     return new Promise((resolve, reject) => {
       this.regenerateAffectedPages(uniquePaths)
         .then(() => fs.removeAsync(this.tempPath))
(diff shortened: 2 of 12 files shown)


src/Site.js
```diff
@@ -131,7 +131,9 @@ class Site {
     this.plugins = {};
     this.siteConfig = {};
     this.siteConfigPath = siteConfigPath;
-    this.userDefinedVariablesMap = {};
+
+    // Site wide variable preprocessor
+    this.variablePreprocessor = undefined;
```
ang-zeyu, line 136: `userDefinedVariablesMap` is removed - note all variable preprocessing will be centered around `variablePreprocessor`

src/lib/markbind/src/parser.js
```diff
@@ -415,27 +246,21 @@ class Parser {
...
-          path: actualFilePath,
-        });
-        this._extractInnerVariables(fileContent, context, config);
-        const innerVariables = this.getImportedVariableMap(context.cwf);
-        fileContent = njUtil.renderRaw(fileContent, {
-          ...userDefinedVariables, ...innerVariables,
-        });
+        const { variablePreprocessor } = config;
+
+        const outerFileContentRendered = variablePreprocessor.renderOuterVariables(data, file);
```
ang-zeyu, line 251: preview of the next part of this PR: reducing boilerplate for rendering site and page variables together.
hopefully this breaks the pr into two more reviewable components

Only site variables are completed for now, but the aim is to get this chunk down even further to 1 - 3 lines of code

src/lib/markbind/src/parser.js
```diff
@@ -449,14 +274,6 @@ class Parser {
     context.cwf = config.cwf || file; // current working file
 
     return new Promise((resolve, reject) => {
-      let actualFilePath = file;
```
ang-zeyu, line 452: removed this as 
- confirmed that for expressive layouts, the file will always exist ( ie. the layout file ) ( there is also a explicit check to abort in `generateExpressiveLayout` if the file does not exist )
- it adds confusion to the changes below, since `actualFilePath` will always `== file`

src/lib/markbind/src/parser.js
```diff
@@ -480,37 +297,28 @@ class Parser {
         resolve(cheerio.html(nodes));
       });
 
+      const { additionalVariables, variablePreprocessor } = config;
+
+      const outerFileContentRendered = variablePreprocessor.renderOuterVariables(pageData, file,
```
ang-zeyu, line 302: similar preview boilerplate reduction

src/Page.js
```diff
@@ -628,8 +628,7 @@ class Page {
     this.includedFiles.add(headerPath);
     // Map variables
     const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${njUtil.renderRaw(headerContent, userDefinedVariables)}\n${pageData}`;
+    return `${this.variablePreprocessor.renderSiteVariables(parentSite, headerContent)}\n${pageData}`;
```
ang-zeyu, line 631: boilerplate reduction for `site` variables ( formerly `userDefinedVariablesMap` ) which is completed in this PR

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -307,6 +302,9 @@ function _preprocessInclude(node, context, config, parser) {
   element.children = cheerio.parseHTML(childrenHtml, true);
 
   if (element.children && element.children.length > 0) {
+    childContext.source = isIncludeSrcMd ? 'md' : 'html';
+    childContext.callStack.push(context.cwf);
```
ang-zeyu, line 306: moved this down here to where its actually needed, so the variable processing code above is cleaner

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+
+    /*
+     * Page level
+     */
+
+    /**
+     * Map of files to another object containing its variable names matched to values
+     * @type {Object<string, Object<string, string>>}
+     */
+    this.variableLookup = {};
```
ang-zeyu, line 55: these 3 variables are the static variables originally from `Parser`

I changed these from `Map<string, Map<string, string>>` to `Object<...>` for the following reasons:
- familiarity - objects are being used mostly everywhere else in markbind, if not all other places.
- order matters, but not in the way of retrieval (get/set)
  - when we extract variables are sensitive to declaration order - this is still done the same way as before, which is the order that say `$('variable')` iterates through the elements. This is also the only order that matters
  - when we retrieve variables, briefly ( checked by scrutinizing the usages of `variableLookup/fileAliasesMap` ), order does not matter, because variable names are already unique after extraction
- the key value pairs are all primitives that don't require use of a `Map` ( performance )

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+  ATTRIB_CWF,
+  IMPORTED_VARIABLE_PREFIX,
+} = require('../constants');
+
+
+/**
+ * Variable preprocessor to recursively extract and render pages' variables.
+ * Houses a site-wide variable cache.
+ */
+class VariablePreprocessor {
```
ang-zeyu, line 27: centerpoint of variable processing; There's generally 4 categories of methods here ( demarcated with dividers below )
1. site variable storage (get/set) type of methods
2. site variable-only rendering type of methods ( using (1) )
3. page variable storage ( the 3 static variables originally from `Parser` - this is likely still to be revamped in part 2 ) - unlike (1), this also includes 'extraction type' methods ( methods that to sift out the outer page variables and inner page variables ) - here, not much is changed for now ( will annotate where important ), but will be in part 2
4. combined rendering methods (new) - the main benefit of this part 1 pr, along with (2) - which allows reducing boilerplate (main pr post for example). Still a work in progress, with part 2 aiming to reduce this even further, but much less than (3).

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+   * Renders the supplied content using only the site variables belonging to the specified base.
+   */
+  renderSiteVariables(siteBase, content) {
+    return njUtil.renderRaw(content, this.userDefinedVariablesMap[siteBase]);
+  }
+
+
+  /*
+   * --------------------------------------------------
+   * Page level variable storage methods
```
ang-zeyu, line 125: these will likely be heavily revamped in part 2.

For now, generally, miscallaneous refactors are made to prepare for it ( SLAP, better/more comments, better spacing and reorganisation in methods for logical grouping of functionality, etc )

will point out any outliers

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  /**
+   * Extracts inner variables of the content, resolving new imports with respect to
+   * the supplied contentFilePath.
+   * @param content to extract from
+   * @param contentFilePath of the content
+   * @return {[]} Array of files that were imported
+   */
+  extractInnerVariables(content, contentFilePath) {
```
ang-zeyu, line 261: - reduction of a parameter ( `config` ): due to site variable processing being centered in here now
- `context` -> `contentFilePath` ( let's not pass in what's more than necessary )

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+
+  /**
+   * Renders an include file with the supplied context, returning the rendered
+   * content and new context with respect to the child content.
+   * @param filePath of the included file source
+   * @param node of the include element
+   * @param context object containing the variables for the current context
+   * @param asIfAt where the included file should be rendered from
+   */
+  renderIncludeFile(filePath, node, context, asIfAt = filePath) {
```
ang-zeyu, line 390: one parameter reduced, similar to the above

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+
+  /**
+   * Extracts the content's page variables, then renders it.
+   * @param content to render
+   * @param contentFilePath of the content to render
+   * @param additionalVariables if any
+   */
+  renderOuterVariables(content, contentFilePath, additionalVariables = {}) {
```
ang-zeyu, line 428: new combined rendering methods ( `renderInner/OuterVariables` ) - aim to get this down to 1 in part 2

ang-zeyu:
on hold pending #1225, so that less changes are observed here

src/Page.js
```diff
@@ -658,10 +657,8 @@ class Page {
     const footerContent = fs.readFileSync(footerPath, 'utf8');
     // Set footer file as an includedFile
     this.includedFiles.add(footerPath);
-    // Map variables
-    const parentSite = urlUtils.getParentSiteAbsolutePath(this.sourcePath, this.rootPath, this.baseUrlMap);
-    const userDefinedVariables = this.userDefinedVariablesMap[parentSite];
-    return `${pageData}\n${njUtil.renderRaw(footerContent, userDefinedVariables)}`;
+
+    return `${pageData}\n${this.variablePreprocessor.renderSiteVariables(this.sourcePath, footerContent)}`;
```
ang-zeyu, line 661: example of site variable processing boilerplate reduction (the second arrow), similar ones follow throughout in `page.js`

src/Site.js
```diff
@@ -539,34 +539,31 @@ class Site {
         logger.warn(e.message);
       }
 
-      // This is to prevent the first nunjuck call from converting {{baseUrl}} to an empty string
-      // and let the baseUrl value be injected later.
-      userDefinedVariables.baseUrl = '{{baseUrl}}';
-      this.userDefinedVariablesMap[base] = userDefinedVariables;
+      /*
```
ang-zeyu, line 542: Example of arrow number 1 in the diagram (site variable extraction), similar ones follow in this file

src/lib/markbind/src/parser.js
```diff
@@ -27,107 +26,16 @@ cheerio.prototype.options.decodeEntities = false; // Don't escape HTML entities
...
-    this._options = options || {};
-    this._fileCache = {};
+  constructor(config) {
+    this.variablePreprocessor = config.variablePreprocessor;
     this.dynamicIncludeSrc = [];
     this.staticIncludeSrc = [];
     this.missingIncludeSrc = [];
   }
 
-  /**
```
ang-zeyu, line 42: Most of the methods relating to page variable extraction and rendering are completely changed (there's a high level overview of the changes in the original post).
Do let me know if you prefer me to point out the characteristics (as per the overview) in the original code!

src/lib/markbind/src/parser.js
```diff
@@ -401,29 +235,15 @@ class Parser {
...
-        ...additionalVariables,
-      });
-      this._extractInnerVariables(fileContent, context, config);
-      const innerVariables = this.getImportedVariableMap(context.cwf);
-      fileContent = njUtil.renderRaw(fileContent, {
-        ...userDefinedVariables,
-        ...additionalVariables,
-        ...innerVariables,
-      });
+      const renderedContent = this.variablePreprocessor.renderPage(file, content, additionalVariables);
```
ang-zeyu, line 238: huge boilerplate reduction, due to the process overhaul and method extraction.
The left shows the lengthy 2 step process as mentioned in the overview in the op.

This is also the first of 2 methods belonging to function 4 I changed. (the second being `renderIncludeFile` (annotated below shortly))

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -241,16 +240,11 @@ function _preprocessInclude(node, context, config, parser) {
...
-
-  const innerVariables = parser.getImportedVariableMap(filePath);
-  const fileContent = njUtil.renderRaw(content, {
-    ...userDefinedVariables, ...innerVariables,
-  });
+  const { variablePreprocessor } = config;
+  const {
+    renderedContent,
+    childContext,
+  } = variablePreprocessor.renderIncludeFile(actualFilePath, element, context, filePath);
```
ang-zeyu, line 247: Similar boilerplate reduction here, and also the second method of function 4.

The difference here (and also per the original code) is that the child context is returned which is necessary to render nested includes with the parent include's `<variable>`s wrapped by `<include>`s.(https://markbind.org/userGuide/reusingContents.html#specifying-variables-in-an-include)

The 'page' (included content) is also rendered with the extracted variables wrapped inside the include

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -318,6 +315,16 @@ function _preprocessVariables() {
   return utils.createEmptyNode();
 }
 
+function _preprocessImports(node, parser) {
+  if (node.attribs.from) {
+    parser.staticIncludeSrc.push({
```
ang-zeyu, line 320: previously this is done inside the variable processing methods, hence `variablePreprocessor` needs to know of `parser`. I moved it here to remove that dependency

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,421 @@
...
+  /**
+   * Renders an <include> file with the supplied context, returning the rendered
+   * content and new context with respect to the child content.
+   * @param filePath of the included file source
+   * @param node of the include element
+   * @param context object containing the parent <include> (if any) variables for the current context,
+   *        which have greater priority than the extracted include variables of the current context.
+   * @param asIfAt where the included file should be rendered from
+   */
+  renderIncludeFile(filePath, node, context, asIfAt) {
```
ang-zeyu, line 364: The gist of the revamp is summarised in this method and the one below as mentioned.
The main highlights are:
- the process is made pure - no static `Parser` variables are needed any longer
- no hacky workarounds like `Proxy` due to removal of the 2-step extraction/rendering into one
- better performance 😁 

The differences in this and render page are:
- we need to render `<include>` files with the `<include **var-xx="..."**>**<variable>...</variable>**</include>`s
- the child context is returned for processing of nested `<include>`s

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -162,236 +145,187 @@ class VariablePreprocessor {
...
-          const jsonData = fs.readFileSync(variableFilePath);
-          const varData = JSON.parse(jsonData);
-          Object.entries(varData).forEach(([varName, varValue]) => {
-            setPageVariableIfNotSetBefore(varName, varValue);
-          });
-        } catch (err) {
-          logger.warn(`Error ${err.message}`);
-        }
+    // NOTE: Selecting both at once is important to respect variable/import declaration order
+    $('variable, import[from]').not('include > import').each((index, elem) => {
```
marvinchin, line 269: Should this be `include > variable` instead, to avoid recognizing include variables as page variables?

src/Site.js
```diff
@@ -691,7 +688,6 @@ class Site {
     const filePathArray = Array.isArray(filePaths) ? filePaths : [filePaths];
     const uniquePaths = _.uniq(filePathArray);
     logger.info('Rebuilding affected source files');
-    MarkBind.resetVariables();
```
marvinchin, line 694: Is there a reason why we no longer need to reset the variables? What if a variable has been deleted? Would the old value associated with the variable also be cleared when rebuilding the page?

src/lib/markbind/src/parser.js
```diff
@@ -27,107 +26,16 @@ cheerio.prototype.options.decodeEntities = false; // Don't escape HTML entities
 const {
   ATTRIB_INCLUDE_PATH,
   ATTRIB_CWF,
-  IMPORTED_VARIABLE_PREFIX,
 } = require('./constants');
 
 class Parser {
-  constructor(options) {
-    this._options = options || {};
-    this._fileCache = {};
```
marvinchin, line 36: It seems like our new implementation no longer caches included files. Should we add this to the `variablePreprocessor`?

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,495 @@
...
+
+      pageVariables[variableName] = njUtil.renderRaw(rawVariableValue, {
+        ...importedVariables,
+        ...pageVariables,
+        ...userDefinedVariables,
+        ...includedVariables,
+      }, {}, false);
+    };
+
+    $('variable').each((index, node) => {
```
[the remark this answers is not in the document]
ang-zeyu: yup its revamped

src/lib/markbind/src/preprocessors/variablePreprocessor.js
```diff
@@ -0,0 +1,495 @@
...
+    const pageVariables = this.extractPageVariables(asIfAt, fileContent, includeVariables);
+
+    // Render the included content with all the variables
+    const content = njUtil.renderRaw(fileContent, {
+      ...pageVariables,
+      ...includeVariables,
+      ...userDefinedVariables,
+    });
+
+    const includedSrces = this.extractInnerVariablesIfNotProcessed(content, asIfAt, asIfAt);
```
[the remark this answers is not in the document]
ang-zeyu: this is revamped as well (moved to `componentPreprocessor`)

test/unit/parser.test.js
```diff
@@ -278,15 +275,10 @@ test('includeFile replaces <include src="include.md#doesNotExist"> with error <d
   fs.vol.fromJSON(json, '');
   const baseUrlMap = new Set([ROOT_PATH]);
 
-  const markbinder = new MarkBind({
-    errorHandler: (e) => {
```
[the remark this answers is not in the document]
ang-zeyu: it's a leftover from a #1207, I needed to pass in `variablePreprocessor` so I removed that here as well.

src/Site.js
ang-zeyu: To add, the process before essentially involved 2x extraction and 2x rendering, storing the intermediate extracted variables in `Parser`'s static variables.
The process is now pure since its just one step, so we have no need for that.

ang-zeyu:
Resolved a minor conflict with #1222

test/functional/test_site/expected/testImportVariables._include_.html
```diff
@@ -6,7 +6,7 @@
 <p>Test import variables that itself imports other variables:</p>
 <h2 id="trying-to-access-a-page-variable">Trying to access a page variable:<a class="fa fa-anchor" href="#trying-to-access-a-page-variable" onclick="event.stopPropagation()"></a></h2>
 <p>There should be something red below:</p>
-<div style="color:red">This is a page variable from <a href="http://variablestoimport.md">variablestoimport.md</a></div>
```
marvinchin, line 9: Is the removal of the anchor tags expected?

marvinchin: It feels like there was a bug with our previous implementation that got `variablestoimport.md` to be rendered as a link. The new version seems to be the correct one.

ang-zeyu: yup, this is the source `<variable name="pagevar">This is a page variable from variablestoimport.md</variable>`, it was there at first due to the removed line 125 of `parser.js` which does `md.renderInline` on it.
All md rendering is done in the `render` stage, so this isn't necessary.It gets removed because in the source file, this is the code:
```
<div style='color:red'>{{pagevar}}</div>
```

by commonmark spec `pagevar` which contains the anchor dosen't get markdown treatment

ang-zeyu: > Just to confirm, I assume you've verified this with the new changes? 🙂

just checked again on the latest master (clean rebase below), no changes save for logs & timestamps!

ang-zeyu: I'll get this in then :grin:

PR #1191, opened 2020-04-11
[v2] Fix anchor navigation in panel headers

### LINKED ISSUE #1071: Anchor navigation in panel not working properly with fixed header

### LINKED ISSUE #1104: Heading hidden when clicking pageNav

le0tan:

• [X] Documentation update
• [X] Bug fix

Replaces #1145 #1187 
Fixes #1104 Fixes #1071

Move span generation to a plugin during build time. Add `::before` to panel card containers to make navigation work properly on panel headers.

na

Test if the panel functions are broken in any way.

**Testing instructions:**

na

To avoid dynamically injecting lots of dummy spans in a document with
many headings, let's generate these spans at build time.

To make anchor navigation function properly on panel headers, let's
introduce ::before element to `card-container` to make space for the
fixed top nav.

### DIFF
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -20,8 +20,8 @@ function insertCss(cssCode) {
 }
 
 function setupAnchorsForFixedNavbar() {
-  const headerSelector = jQuery('header');
-  const isFixed = headerSelector.filter('.header-fixed').length !== 0;
+  const headerSelector = jQuery('header[fixed]');
+  const isFixed = headerSelector.length !== 0;
   if (!isFixed) {
     return;
   }
--- a/asset/js/setup.js
+++ b/asset/js/setup.js
@@ -13,61 +13,45 @@ function scrollToUrlAnchorHeading() {
   }
 }
 
-function flattenModals() {
-  jQuery('.modal').each((index, modal) => {
-    jQuery(modal).detach().appendTo(jQuery('#app'));
-  });
-}
-
 function insertCss(cssCode) {
   const newNode = document.createElement('style');
   newNode.innerHTML = cssCode;
   document.getElementsByTagName('head')[0].appendChild(newNode);
 }
 
-function setupAnchors() {
+function setupAnchorsForFixedNavbar() {
   const headerSelector = jQuery('header');
   const isFixed = headerSelector.filter('.header-fixed').length !== 0;
+  if (!isFixed) {
+    return;
+  }
+
   const headerHeight = headerSelector.height();
   const bufferHeight = 1;
-  if (isFixed) {
-    jQuery('.nav-inner').css('padding-top', `calc(${headerHeight}px)`);
-    jQuery('#content-wrapper').css('padding-top', `calc(${headerHeight}px)`);
-    insertCss(
-      `span.anchor {
-      position: relative;
-      top: calc(-${headerHeight}px - ${bufferHeight}rem)
-      }`,
-    );
-    insertCss(`span.card-container::before {
-          display: block;
-          content: '';
-          margin-top: calc(-${headerHeight}px - ${bufferHeight}rem);
-          height: calc(${headerHeight}px + ${bufferHeight}rem);
-        }`);
-  }
+  jQuery('.nav-inner').css('padding-top', `calc(${headerHeight}px)`);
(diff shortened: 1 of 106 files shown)


src/lib/markbind/src/parser.js
```diff
@@ -353,6 +353,11 @@ class Parser {
 
     componentParser.postParseComponents(node, this._onError);
 
+    // If a fixed header is applied to this page, generate dummy spans as anchor points
+    if (config.fixedHeader && (/^h[1-6]$/).test(node.name) && node.attribs.id) {
```
ang-zeyu, line 357: since `(/^h[1-6]$/).test(node.name)` is used twice in the method, perhaps we could extract that out into a variable `isHeaderTag`

src/Page.js
```diff
@@ -927,7 +940,12 @@ class Page {
         .then(result => this.insertFooterFile(result))
         .then(result => Page.insertTemporaryStyles(result))
         .then(result => markbinder.resolveBaseUrl(result, fileConfig))
-        .then(result => markbinder.render(result, this.sourcePath, fileConfig))
+        .then((result) => {
+          // this.fixedHeader may be updated in previous step insertHeaderFile
+          // we need to reflect the update and pass it to the renderer, if necessary
+          fileConfig.fixedHeader = this.fixedHeader;
```
ang-zeyu, line 946: should we pass `fileConfig` into `insertHeaderFile` instead? Best to keep related code close together, would remove the need for the comment as well

ang-zeyu:
Some conflicts @le0tan

le0tan:
@ang-zeyu looks like you removed some pointer events from `setupAnchors`. This function is necessary in all cases as it injects the events where hovering over the heading or anchor will make the anchor icon visible. Do tell me if I misunderstood anything.

ang-zeyu:
> @ang-zeyu looks like you removed some pointer events from `setupAnchors`. This function is necessary in all cases as it injects the events where hovering over the heading or anchor will make the anchor icon visible. Do tell me if I misunderstood anything.

Refer to the second commit in #1129. It's moved to plugin css ( new feature )

le0tan:
@ang-zeyu got it and I changed my code accordingly, can you take a look?

ang-zeyu:
@le0tan there's still a large amount of conflicts, mostly from previous versions

docs/userGuide/syntax/headers.mbdf
```diff
@@ -23,6 +23,17 @@ In the page that you want to include the header:
 ```
 </div>
 
+You can fix the header to be always on the top by add `class="header-fixed"` attribute to the `<header>` tag of your header file:
+
+{{ icon_example }}
+**`_markbind/headers/`**`header.md`:
```
ang-zeyu, line 29: Since we are documenting this now, would be a good point to get @damithc's opinion on fixed navbar usage.
( maybe something simpler like `<header fixed>`? 😮  )

asset/js/setup.js
```diff
@@ -32,20 +32,18 @@ function setupAnchorsForFixedNavbar() {
   jQuery('#content-wrapper').css('padding-top', `calc(${headerHeight}px)`);
```
ang-zeyu, line 32: would it be possible to just insert the padding for the main container and not its 3 sub containers?
fixedheader

docs/userGuide/syntax/headers.mbdf
damithc: Yes I prefer a simpler `fixed` if there are no other drawbacks.

asset/js/setup.js
le0tan: As far as I'm concerned it doesn't work...

ang-zeyu: ok, might be better to leave it this way as well: `.css('nav-inner')` needs to be offsetted with the original padding

navbarpadding

le0tan: just did that, wondering if we need to add `1rem` to `content-wrapper` as well, but probably better leave it as a future first timer issue?

ang-zeyu: Good idea, would make it seem less squeezed to the top. Let's leave it as a first timer issue then, though if you want to include it here it'd be nice as well

PR #1168, opened 2020-03-30
Remove duplicate code in syntax doc

nbriannl:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Enhancement to an existing feature
• [ ] Other, please explain:

Fixes #1159
~~Requires #1193~~ (merged) 

**What is the rationale for this request?**

Currently developers need to write the same code twice for syntax and component documentation: one for the code block in the "code" section and one to be rendered by MarkBind in the "output" section. However this duplication should have been avoided. And developers may forget to make sure that both sections are on parity with each other.

**What changes did you make? (Give an overview)**

Created a new boilerplate `codeAndOuput.md` that takes in MarkBind code assigned to the `code` variable declared in the main markdown file.
And renders two things: (1) the code block containing the code, and (2) the rendered output of the code

**Provide some example code that this change will affect:**

**Is there anything you'd like reviewers to focus on?**

**Testing instructions:**

Compare the following pages, and see if any of the examples regress or render wrongly, be it either the source code or the rendered output.

https://deploy-preview-1168--markbind-master.netlify.com/userguide/formattingcontents
https://markbind.org/userGuide/formattingContents.html

https://deploy-preview-1168--markbind-master.netlify.com/userguide/usingcomponents
https://markbind.org/userGuide/usingComponents.html

**Proposed commit message: (wrap lines at 72 characters)**
Remove duplicate code in syntax documentation

Right now, we have to type example code in our User Guides twice,
once as example syntax, and once to be rendered by Markbind.

This practice is prone to errors as we have to ensure the two code
are the same. And it makes the user documentation quite lengthy
in terms of the source md and mbdf files. 

Let's make a boilerplate that takes the example code, 
and renders it once as a code block containing the code
and once as the rendered MarkBind output of the code.
And use it in the User Guide.

PR #1193, opened 2020-04-11
Allow variable tags to contain malformed xml

ang-zeyu:

• [x] Bug fix

Fixes the first point mentioned here https://github.com/MarkBind/markbind/pull/1168#issuecomment-612330854

Simple patch to allow `<variable>` tags to hold any content built on #1047 

na

**Testing instructions:**
`npm run test` should pass

Allow variable tags to contain malformed xml

With the special tags patch, variables are able to contain custom
content possibly containing malformed xml.

Let’s add variable to the list of special tags to do so.

### DIFF
--- a/src/lib/markbind/src/lib/markdown-it/markdown-it-escape-special-tags.js
+++ b/src/lib/markbind/src/lib/markdown-it/markdown-it-escape-special-tags.js
@@ -10,7 +10,7 @@ function escape_plugin(md, tagsToIgnore) {
   const HTML_OPEN_CLOSE_TAG_RE = require('markdown-it/lib/common/html_re').HTML_OPEN_CLOSE_TAG_RE;
 
   const specialTagsRegex = Array.from(tagsToIgnore)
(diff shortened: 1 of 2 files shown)


nbriannl:
Thanks for this fix, would anything regress now that variables can contain malformed xml? I don't see any reason why anything would regress.

ang-zeyu:
> Thanks for this fix. Would anything regress now that variables can contain malformed xml? I don't see any reason why anything would regress. Test seem fine, update tests seem ok too.

Yup, 2103 site had no diffs either. If anything it should fix edge cases of variables currently in use that had slightly incompatible syntax

damithc:
@ang-zeyu should we update the user guide somewhere to reflect this enhancement?

ang-zeyu:
> @ang-zeyu should we update the user guide somewhere to reflect this enhancement?

Will do, didn't notice this limitation was documented

damithc:
Tested in production. Good work @ang-zeyu and reviewers.

PR #1203, opened 2020-04-16
Update variables documentation

ang-zeyu:

• [x] Documentation update for #1193 and some reformatting for variable documentation

Update variable documentation

The limitation of variables not being able to contain incomplete or
malformed xml has been patched.
Some areas of the variable documentation can also be formatted with
newly introduced features to be more visually appealing. 

Let’s update the user guide to reflect the patch, and reformat some
areas of the variable documentation.


damithc:
Do we have an example of using ill-formed xml as a variable? Some might assume the value needs to be well-formed even if it was never mentioned.

ang-zeyu:
> Do we have an example of using ill-formed xml as a variable? Some might assume the value needs to be well-formed even if it was never mentioned.

ok I see the point. I didn't document this as I felt that it should be the expected behaviour.
Noticed some other areas the variables docs can be tidied up in as well. Will expand the scope of this.

PR #1204, opened 2020-04-16
Update the build steps in the maintainer guide

marvinchin:

• [x] Documentation update

When releasing `v2.14.0`, there was a build error because dependencies were not updated. Let's add a step in the build instructions to remind maintainers to update their dependencies before generating the build.

Also, https://github.com/MarkBind/vue-strap/pull/130 split the `build` script into `build` and `builddev`. Let's update the maintainer guide to reflect this change.

Added the `npm install` step before building vue-strap and markbind
Updated the `npm run build` step for vue-strap

Update the build steps in the maintainer guide

When releasing `v2.14.0`, there was a build error because
dependencies were not updated. Let's add a step in the build
instructions to remind maintainers to update their
dependencies before generating the build.

Also, the vue-strap `build` script has been split into `build`
and `builddev`. Let's update the maintainer guide to reflect
this change.


damithc:
Related to releases:
When writing release notes, should we write the normal release note targeting authors (i.e., include only things that make sense to authors) and put all 'internal' improvements (i.e., changes that authors will not understand) at the bottom in a separate section? Alternatively, we can have a 'User-facing' improvements as a separate section at the top and the rest can remain the same.

docs/devGuide/maintainerGuide.md
```diff
@@ -64,11 +64,17 @@ Ensure that:
...
-1. Ensure that you are on the root directory on vue-strap, then execute:
+```
+$ npm install
+```
+
+2. Build the assets by executing:
 
 ```
 $ npm run build
+$ npm run builddev
```
ang-zeyu, line 77: just one nit: this shouldn't be needed, it builds the unminified version

ang-zeyu:
> Related to releases:
> When writing release notes, should we write the normal release note targeting authors (i.e., include only things that make sense to authors) and put all 'internal' improvements (i.e., changes that authors will not understand) at the bottom in a separate section? Alternatively, we can have a 'User-facing' improvements as a separate section at the top and the rest can remain the same.

Could move `### Code quality, dependencies` under this. Maybe also add `devOps, plugin API, miscellaneous` sections as well. Think we could do this in a separate PR though

docs/devGuide/maintainerGuide.md
```diff
@@ -77,30 +83,35 @@ You should see changes in `vue.min.js`, `vue-strap.js.map` and `vue-strap.min.js
 
 * **Note:** Do take a glance at the diff for `vue.min.js` to see if there's any strange changes (e.g. the version of jQuery changes even though no one upgraded it).
 
-2. Commit the new changes.
+3. Commit the new changes.
```
le0tan, line 86: Can actually use `1.` for all numbered lists in markdown: according to the [CommonMark](https://spec.commonmark.org/0.29/#ordered-list) specification, the subsequent ordered list numbers are disregarded.

marvinchin: This doesn't seem to work for this file. I suspect it's because we have a bunch of other stuff between bullet points.

docs/devGuide/maintainerGuide.md
```diff
@@ -64,43 +64,53 @@ Ensure that:
...
 ```
 $ npm run build
 ```
 
-You should see changes in `vue.min.js`, `vue-strap.js.map` and `vue-strap.min.js`.
+You should see changes in `vue-strap.min.js`.
 
 * **Note:** Building vue-strap is optional if there's no changes since the last release of MarkBind. If there are no changes to the three files mentioned above, skip to the "Building MarkBind" section.
 
-* **Note:** Do take a glance at the diff for `vue.min.js` to see if there's any strange changes (e.g. the version of jQuery changes even though no one upgraded it).
```
marvinchin, line 78: We don't actually get our `vue.min.js` file from `vue-strap`

docs/devGuide/maintainerGuide.md
```diff
@@ -333,6 +355,6 @@ test: Add diff printing for easier debugging (#632)
 
 f. Click "Publish release".
 
-11. Finally, announce the new release on our Slack channel (replace `A.B.C` with the new version). Congrats!
+12. Finally, announce the new release on our Slack channel (replace `A.B.C` with the new version). Congrats!
```
damithc, line 358: If this is a Markdown list, we can use `1.` for every time, which Markdown will convert into a proper numbered list. That way, easier to insert items into the list in future.

docs/devGuide/maintainerGuide.md
le0tan: Yes, the numbering will reset if the list is interrupted.

docs/devGuide/maintainerGuide.md
marvinchin: Unfortunately, because we have a bunch of other stuff in between each item, the numbering will reset for each point if we do that.

PR #1207, opened 2020-04-19
Replace console.warn with logger.warn: componentParser.js

### LINKED ISSUE #1060: Importing logger from componentParser.js cause parser.test.js to fail

In MarkBind Version: 2.11.0, I have an issue when importing logger from componentParser.js cause parser.test.js to fail.
For example, calling it accordingly as here https://github.com/MarkBind/markbind/blob/8005582e60e8f917c2629d5f81742b103133659f/src/lib/markbind/src/parsers/componentParser.js#L7

will cause `parser.test.js` to fail as shown below. 
image

As @ang-zeyu has pointed out, this occurs for parser.js as well, which is likely the reason logger isn't used in there as well.

The issue is that parser.test.js mocks the whole fs module, which causes the DailyRotateFile constructor to fail automatically in finding the directory.

A simple fix would be to only mock the specific functions of fs relavant to parser.test.js

(Describe your proposed solution here.)

(Write your answer here.)

**Additional context**

(Write your answer here.)

nbriannl:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [X] Other, please explain: Finishing up a todo from #1057 

This is a follow up to #1057 where `console.warn` is used instead of `logger.warn` due to a bug as raised in #1060. The test no long fail in #1117 by mocking logger in the `parser.test.js`. Although the approach could be improved.

Replaced all `console.warn` with `logger.warn` as indicated by the TODO signposts.

**Testing instructions:**
Run tests to check if anything regressed

`markbind serve` the example code in the gist.
https://gist.github.com/nbriannl/5c44b3ee8a8af78c45a7a4f3659ce196

You should now see
image

Replace console.warn with logger.warn,

A follow up to #1057.
Let us also remove console.error as the error handler 
And directly set the error handler to be logger during init

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -568,9 +568,7 @@ class Page {
    * @param {FileConfig} fileConfig
    */
   generateExpressiveLayout(pageData, fileConfig) {
-    const markbinder = new MarkBind({
-      errorHandler: logger.error,
-    });
+    const markbinder = new MarkBind();
     const template = {};
     template[LAYOUT_PAGE_BODY_VARIABLE] = pageData;
     const { layout } = this.frontMatter;
@@ -916,9 +914,7 @@ class Page {
     this.includedFiles = new Set([this.sourcePath]);
     this.headerIdMap = {}; // Reset for live reload
 
-    const markbinder = new MarkBind({
-      errorHandler: logger.error,
-    });
+    const markbinder = new MarkBind();
     /**
      * @type {FileConfig}
      */
@@ -1236,9 +1232,7 @@ class Page {
        * We create a local instance of Markbind for an empty dynamicIncludeSrc
        * so that we only recursively rebuild the file's included content
        */
-      const markbinder = new MarkBind({
-        errorHandler: logger.error,
-      });
+      const markbinder = new MarkBind();
       return markbinder.includeFile(dependency.to, {
         baseUrlMap: this.baseUrlMap,
         userDefinedVariablesMap: this.userDefinedVariablesMap,
--- a/src/lib/markbind/src/parser.js
+++ b/src/lib/markbind/src/parser.js
@@ -33,8 +33,6 @@ const {
 class Parser {
   constructor(options) {
     this._options = options || {};
-    // eslint-disable-next-line no-console
-    this._onError = this._options.errorHandler || console.error;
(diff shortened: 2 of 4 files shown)


ang-zeyu:
should we fix the root cause of the tests failing first instead? ( mocking fs for `winston` )

nbriannl:
> should we fix the root cause of the tests failing first instead? ( mocking fs for `winston` )

Even if we fix the root cause, we would still have to do this change. Therefore, I thought of doing it now. 😄 Since the bug is fixed, albeit not as ideally.

Should I raise a new issue or rewrite the issue?

ang-zeyu:
Ahh ok, I'm fine with either. If that's the case perhaps could replace the usages in `parser.js` as well?

*edited your comment when quoting by accident

nbriannl:
> Ahh ok, I'm fine with either. If that's the case perhaps could replace the usages in `parser.js` as well?
> 
> *edited your comment when quoting by accident

Sounds good, I can do that here.

nbriannl:
Should i be changing this line as well? https://github.com/MarkBind/markbind/blob/68eaefb2c9bcbdd5b2708343d792df4d912cc02a/src/lib/markbind/src/parser.js#L37

ang-zeyu:
> Should i be changing this line as well?

Yes, think we could remove it completely though, and all instances of `errorHandler` being passed as an option into the parser constructor. It was likely a workaround for the root problem

nbriannl:
> > Should i be changing this line as well?
> 
> Yes, think we could remove it completely though, and all instances of `errorHandler` being passed as an option into the parser constructor. It was likely a workaround for the root problem

Hm, not entirely sure what you mean. `_onError` is also used in part relating to `componentPreprocessor.preProcessComponent()`

ang-zeyu:
> Hm, not entirely sure what you mean. `_onError` is also used in part relating to `componentPreprocessor.preProcessComponent()`

It's a workaround for not being to import `logger` directly in `parser`
- when we the unit test is run, `_onError` defaults to `console.error`
- when running anything else from a user standpoint ( e.g. `markbind build` ), `errorHandler` is set to `logger.error`. Should be unncessary with the changes here now

PR #1211, opened 2020-04-19
Add DG to markbind.org

### LINKED ISSUE #1161: Link to DevGuide in markbind.org

Currently we have no easy access to the developer's guide from markbind.org. It would be better if developer's guide along with contribution guide can be easily accessible from the official website.

Tejas2805:

• [x] Documentation update
• [x] Bug fix

Fixes #1161 

The site markbind.org doens't have a developer guide.

Added developer guide to the site.

N.A.

N.A.

**Testing instructions:**
Not sure. 

```
The site markbind.org doesn't provide access to
developer guide.

Let's add the developer guide to the site, so that the
developer guide along with the contribution guide is
accessible from official website.
```


damithc:
Is [this requirement](https://github.com/MarkBind/markbind/issues/1161#issuecomment-614434564) satisfied?

Tejas2805:
> Is [this requirement](https://github.com/MarkBind/markbind/issues/1161#issuecomment-614434564) satisfied?

I think so. Just waiting for one reviewer to review and confirm the same.

Tejas2805:
Based on what I tried locally, this works!

nbriannl:
> Keep in mind the UG should follow the latest released version while DG should follow the master branch.

Is this satisfied? It seems the current implementation is building the DG pages as part of ug-site generation. This would mean that the DG would only follow the latest released version.

I might be wrong however.

le0tan:
From what I’ve seen in [markbind.org repo](https://github.com/MarkBind/markbind.github.io), the site generation only happens at every release. So basically DG is not updated more frequently than UG unless we change the trigger for Travis.

Tejas2805:
> From what I’ve seen in [markbind.org repo](https://github.com/MarkBind/markbind.github.io), the site generation only happens at every release. So basically DG is not updated more frequently than UG unless we change the trigger for Travis.

But won't this affect UG too then?

nbriannl:
For this PR, you could update the Navbar and SiteNav to contain links to the DG hosted at https://markbind.org/devdocs/

damithc:
The current solution: we only generate DG in netlify version because netlify version is always up to date with master branch. DG is omitted in the version deployed to markbind.org because that version can be behind the master branch. Hence, developers looking to read the DG should be directed to the netlify version even if the link to the DG is given in the markbind.org.

Anyway, that's the current solution. Not wedded to it if there is a better alternative.

damithc:
> The current solution: we only generate DG in netlify version because netlify version is always up to date with master branch. DG is omitted in the version deployed to markbind.org because that version can be behind the master branch. Hence, developers looking to read the DG should be directed to the netlify version even if the link to the DG is given in the markbind.org.
> 
> Anyway, that's the current solution. Not wedded to it if there is a better alternative.

Not sure if things changed recently though. The above was the situation sometime back. @yamgent might remember better.

Tejas2805:
@nbriannl Do you mean something along these lines?

nbriannl:
> @nbriannl Do you mean something along these lines?

Yeah. This isn't complete though, because I believe you need to do `userGuide/header.md`

Tejas2805:
> Yeah, along this line! This isn't complete though, because I believe you need to do `userGuide/header.md`

I was just confirming if you had something like this in your mind. Have implemented the rest.

marvinchin:
Thanks for trying to address this! Sorry I haven't been able to respond sooner.

After looking at the concerns raised above, I think fiddling with the links might not be worth the effort, and it might make the site harder to maintain. Perhaps we can just add a page that directs potential contributors to our GitHub page (this can be a hard link), and update the README to point them to the netlify version with the DG. This would help direct potential contributors to the necessary resources without much added complexity.

damithc:
Agree with @marvinchin 
Given DG updates are rare, it may be OK for the links to be broken in netlify preview, if the alternative solutions are costly.  Netlify previews are used by the reviewers only.

PR #1216, opened 2020-04-27
Use border dotted for hover triggers

openorclose:

• [x] Bug fix

Fixes https://github.com/MarkBind/markbind/issues/1053

Use dotted instead of dashed for border property


ang-zeyu:
needs another `updatetest`

PR #1218, opened 2020-05-02
Fix several b-modal selector names

### LINKED ISSUE #1217: Some pageNav links don't work, possibly due to the use of colon
Steps:
1. Go to https://nus-cs2113-ay1920s2.github.io/website/admin/tp-deliverables.html
2. Click on one of these links
   image

Problem: the page doesn't scroll to the relevant heading. No such problem with the other links in the PageNav. Is it something to do with the use of `:` in the heading?

ang-zeyu:

• [x] Bug fix

Fixes #1217

Change leftover `'modal'` selectors from the bootstrap vue migration to `'b-modal'`

**Testing instructions:**
- `npm run test` should pass

Fix several b-modal selector names

The bootstrap vue migration causes modal elements to be transformed into
b-modal.

Let’s change the leftover selector names to match the new name.

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -338,7 +338,7 @@ class Page {
       return;
     }
     const $ = cheerio.load(content);
-    $('modal').remove();
+    $('b-modal').remove();
     this._collectNavigableHeadings($, $.root()[0], elementSelector);
   }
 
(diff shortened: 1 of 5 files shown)


nbriannl:
Not very familiar, so I need to make some clarifications before I approve. Basically, https://github.com/MarkBind/markbind/pull/1033 migrated modals to b-modals, but there were some parts of the code, namely the files you edited, that still used the previous name?

ang-zeyu:
it was missed out in #1033, and only surfaced now. Could double check with @openorclose if there's a reason it wasn't changed though (unlikely)

openorclose:
> it was missed out in #1033, and only surfaced now. Could double check with @openorclose if there's a reason it wasn't changed though (unlikely)

No reason, I missed it out. Thanks for the fix!

ang-zeyu:
ok will merge it in then, thanks for looking through it @nbriannl @openorclose!

PR #1208, opened 2020-04-19
Enhance Json Variable Doc with scope example

crphang:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [X] Documentation update

Resolves #1202 

**What is the rationale for this request?**

Elaborate on how JSON variable is scoped.

**What changes did you make? (Give an overview)**

Added documentation.

**Provide some example code that this change will affect:**

N/A

**Is there anything you'd like reviewers to focus on?**

Clarity of documentation.

**Testing instructions:**

N/A
**Proposed commit message: (wrap lines at 72 characters)**

Enhance Json Variable Doc with scope example

PR #1220, opened 2020-05-03
Allow using {% (end)raw %} tags

### LINKED ISSUE #762: Provide a way to escape nunjucks syntax

### LINKED ISSUE #1206: Allow authors to use render syntax with curly braces `{{ content }}`
Currently authors have no way to render `{{ content }}` easily due to Vue. The only workaround is adding appropriate span tags

`<span>{</span>{ content <span>}</span>}`. 

Rendering such syntax could be required for certain authors attempting to teach various Domain Specific Language that could contain such syntax, one example is Latex.

With work on #1049 and #1047, MarkBind currently allows special tags that retains curly braces instead of being removed from Nunjucks over repeated rendering if `{% raw %} {% endraw%}` is being used. However, due to Vue, we are still unable to render `{{ }}` as it is being used for interpolation. See: https://github.com/vuejs/vue/issues/4223.

A possible experience to overcome that would be to introduce a new plugin:

```
<escape>
{%raw%}
{{ any content }} 
{%endraw%}
</escape>
```

Would be even better if `<escape>` automatically adds required nunjucks escaping.

References and potential solution:

https://github.com/MarkBind/markbind/pull/1049#issuecomment-590297046

> For general use, we could create a special component, something like <escape></escape> that acts the same as a <span v-pre></span> in addition to appending {% raw %} {% endraw %}

ang-zeyu:
~### *netlify node_version needs to be updated to the current lts~

• [ ] Documentation update
• [x] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Resolves #762
Resolves #1206 

Allow usage of {% raw/endraw %} tags

- Patch a _soft_ syntax conflict with markdown-it-attrs, allowing `{% raw/endraw %}` to be used
- Auto add `v-pre` to code fences and inline code and `<code>` elements for convenience
  - usage of `{% raw/endraw %}` is still necessary to display `{{ content }}` - hence variables can still be used inside code fences / etc.
  - for all other cases ( `<div/span/table/etc>` ), one would need to add `v-pre` attribute on top of `{% raw/endraw %}` to disable vue compilation for that element

na

**Testing instructions:**
- {% raw/endraw %} should now work anywhere

Allow using {% (end)raw %} tags

Nunjuck's {% raw %} tags don’t work as expected due to a soft syntax
conflict with markdown-it-attrs.
In addition, variable interpolation syntax cannot be output in
code elements due to vue's variable interpolation.

Let's patch markdown-it-attrs to allow the slight syntax conflicts,
and automatically assign v-pre to code elements to skip vue compilation.


src/lib/markbind/src/lib/markdown-it/patches/markdown-it-attrs-nunjucks.js
```diff
@@ -0,0 +1,78 @@
...
+      start = slice === options.leftDelimiter ? 0 : -1;
+      slice = str.slice(str.length - options.rightDelimiter.length);
+      end = slice === options.rightDelimiter ? str.length - options.rightDelimiter.length : -1;
+      break;
+    }
+
+    /*
+     Simple patch here - abort if the delimiters wrap around % %
+     */
+    const isCharAfterStartPercent = str.charAt(start + 1) === '%';
```
ang-zeyu, line 66: *the only functional changes in this file are these 5 lines

marvinchin: Just to confirm, this won't break the other nunjucks templating syntax that use `{% ...%}` right?

ang-zeyu: yup, if anything it should 'fix' it!

e.g.
```
{% for x in y %}
{{ x }}
{% endfor %}
```

this previously works (and still does) because nunjucks rendering is the first step and long precedes markdown rendering, so the `{% for/endfor %}` is 'erased' long before markdown-it-attrs touches it.
It dosen't work in the case of {% raw/endraw %} because it is still present at the time of markdown rendering though

So this should give more flexibility in future restructuring of the rendering order (if ever needed)

marvinchin: That makes sense, thanks for clarifying!

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -6,10 +6,49 @@ const unescapedEnv = nunjucks.configure({ autoescape: false }).addFilter('date',
...
+const RAW_TAG_REGEX = new RegExp('{% *(end)?raw *%}', 'g');
 
+/**
+ * Pads the outermost {% raw %} {% endraw %} pairs with {% raw %} {% endraw %} again.
+ * This allows variables and other nunjuck syntax inside {% raw %} {% endraw %} tags
+ * to be ignored by nunjucks until the final renderString call.
+ */
 function preEscapeRawTags(pageData) {
-  return pageData.replace(REGEX, `${START_ESCAPE_STR}$&${END_ESCAPE_STR}`);
+  const tagMatches = pageData.matchAll(RAW_TAG_REGEX);
```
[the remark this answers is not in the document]
ang-zeyu: ahh I didn't notice the supported versions. In fact node v10 is still listed under 'maintainence' [here](https://github.com/nodejs/Release#nodejs-release-working-group).

Will update shortly with a more verbose but compatible solution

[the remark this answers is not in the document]
ang-zeyu: updated with the solution suggested here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec, reverted the node version changes in `travis.yml` as well

ang-zeyu:
Thanks for looking through this! @marvinchin

damithc:
@marvinchin @ang-zeyu we might have to do a release because the previous release version is no longer compatible with our own docs after this PR was merged. i.e., I can't build docs in the master branch using the previous released version of MarkBind.

marvinchin:
> the previous release version is no longer compatible with our own docs after this PR was merged. i.e., I can't build docs in the master branch using the previous released version of MarkBind.

@damithc I think that it is reasonable for the previous release to be incompatible with the docs in master branch. Any new feature that is introduced to master that involves any syntax changes to the docs would not be compatible with the previous release. It should be sufficient that the docs on master is compatible with the markbind build on master.

That being said, it has been some time since we have done a release. Perhaps we can do one this week 🙂  @ang-zeyu shall we find some time to do the next release together?

damithc:
> @damithc I think that it is reasonable for the previous release to be incompatible with the docs in master branch. Any new feature that is introduced to master that involves any syntax changes to the docs would not be compatible with the previous release. It should be sufficient that the docs on master is compatible with the markbind build on master.

I agree @marvinchin , it's the normal case. Just that I want to do a doc PR and I don't have the dev setup.

damithc:
But not _very_ urgent. I can PR based on the last commit that is backward compatible. Release at your convenience.

ang-zeyu:
> But not _very_ urgent. I can PR based on the last commit that is backward compatible. Release at your convenience.

Might want to hold off a little bit, let's discuss #1236 first which potentially breaks a lot of things

> That being said, it has been some time since we have done a release. Perhaps we can do one this week 🙂 @ang-zeyu shall we find some time to do the next release together?

Sure! If its fine we could merge in #1189 and #1208  before doing so as well

ang-zeyu:
I'm thinking of getting https://github.com/MarkBind/vue-strap/pull/147 merged as well for now, so the release process is a tiny bit easier.

Not too sure if the repo merging process I mentioned there is the right/best way to do it though (aim being to preserve both repo's histories)

PR #1225, opened 2020-05-20
Remove expressive layouts code duplication

ang-zeyu:

• [x] Other, please explain: small code refactor

Merge the logic for expressive layouts in `parser` into `includeFile`

- merge said logic ( the only difference was `additionalVariables`, which is introduced to `includeFile` )
- move file I/O of `includeFile` into `Page.js` instead which was the case for `includeData` for expressive layouts ( small change, but I think its cleaner to move the initial I/O into `Page.js` )

na

**Testing instructions:**
- `npm run test`

Remove expressive layouts code duplication

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -566,11 +566,9 @@ class Page {
    * Produces expressive layouts by inserting page data into pre-specified layout
    * @param pageData a page with its front matter collected
    * @param {FileConfig} fileConfig
+   * @param {Parser} markbinder instance from the caller, for adding the seen sources.
    */
-  generateExpressiveLayout(pageData, fileConfig) {
-    const markbinder = new MarkBind();
-    const template = {};
-    template[LAYOUT_PAGE_BODY_VARIABLE] = pageData;
+  generateExpressiveLayout(pageData, fileConfig, markbinder) {
     const { layout } = this.frontMatter;
     const layoutPath = path.join(this.rootPath, LAYOUT_FOLDER_PATH, layout);
     const layoutPagePath = path.join(layoutPath, LAYOUT_PAGE);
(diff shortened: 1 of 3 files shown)


src/Page.js
```diff
@@ -566,43 +566,32 @@ class Page {
    * Produces expressive layouts by inserting page data into pre-specified layout
    * @param pageData a page with its front matter collected
    * @param {FileConfig} fileConfig
+   * @param {Parser} markbinder instance from the caller, for adding the seen sources.
    */
-  generateExpressiveLayout(pageData, fileConfig) {
-    const markbinder = new MarkBind();
-    const template = {};
-    template[LAYOUT_PAGE_BODY_VARIABLE] = pageData;
+  generateExpressiveLayout(pageData, fileConfig, markbinder) {
```
ang-zeyu, line 571: passing the parser (`markbinder`) isn't very clean, but will be dealt with in #1189 in full (making it an instance variable)

ang-zeyu:
> The refactor looks good! Just a couple of thoughts about lifting the I/O part up to `Page.js`:
> 
> * Leaving it in will reduce duplication of the `fs.readFileAsync` call we have to make before every `includeFile` call.
> * It kinda makes sense semantically (to me at least) that `includeFile` does the reading of the file. Also, since there should be no other reason for us to use the file content outside of this function, I think not having to pass the file content to `includeFile` will make the call site a little cleaner.
> 
> What are your thoughts about this?

Hmm, my viewpoint was from a perspective of concerns - I think `parser` should be concerned with as little file I/O as possible, because it only 'parses' ( and also 'renders' ).
That said many other areas of `parser` still necessarily do file i/o ( includes / imports ), and I don't see it going away unless we implement a file cache of sorts ( I think there was an issue about this that fizzled though since the benefits were marginal ( modern ssds are too fast 🌚 ) ).

Its a really small thing though, I'm fully open to reverting it. What do you think?

marvinchin:
That makes sense as well. I'm happy to go with this and move towards removing I/O concerns from the parser. Let's get the conflicts resolved and we can get this in!

ang-zeyu:
> That makes sense as well. I'm happy to go with this and move towards removing I/O concerns from the parser. Let's get the conflicts resolved and we can get this in!

Done!

PR #1226, opened 2020-05-22
Refactor parser into a page instance variable

ang-zeyu:

• [x] Other, please explain: small code refactor, originally part of #1189 ( decided to extract this out as well so less changes are observed over there )

We use `parser` ( `markbinder` ) in several places in `Page.js` now. Passing it around as a parameter isn't really ideal (#1225), hence, let's make it an instance variable instead.

- change `markbinder` variable name to `parser` 
- extract parser as an instance variable, and tweak the logic for creating the `resolveDependency` promises accordingly.
- remove `includedFiles` collection in `resolveDependency` ( we can just collect it once in `page.generate()` now, since there's only one parser )

na

**Testing instructions:**
`npm run test` should pass

Refactor parser into a page instance variable

### DIFF
--- a/src/Page.js
+++ b/src/Page.js
@@ -15,7 +15,7 @@ const CyclicReferenceError = require('./lib/markbind/src/handlers/cyclicReferenc
 const { ensurePosix } = require('./lib/markbind/src/utils');
 const FsUtil = require('./util/fsUtil');
 const logger = require('./util/logger');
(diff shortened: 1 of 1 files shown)


ang-zeyu:
shelving this for now, although not as clean, realised there's some benefits to constructing a new instance for each generation, especially with regards to plugin hooks. Perhaps a better solution will soon be found.

Issue #1151, opened 2020-03-27
Restructure plugin documentation


ang-zeyu:
**Is your request related to a problem?**

There's been quite a few developer side plugin apis added to the `using plugins` page, which pertain more to `writing plugins` than using them.
Since plantUML is also a plugin, but is the top site nav layer, we could restructure plugin documentation to be like so for better organization:

**Describe the solution you'd like**

```
- userGuide.md
  - ...
  - authoring contents
  - Using plugins
    - plantUML plugin
    - default plugins ( anchors, etc )
    - algolia
    - ...
- devGuide.md
  - Developer's guide
    - Writing plugins
```

**Describe alternatives you've considered**

na

**Additional context**

na

PR #1228, opened 2020-05-28
Add plugin hooks

### LINKED ISSUE #1188: Add more hooks for plugins

Work done here for a rough idea: #968 

Plugins modify the content currently only through 2 hooks, `pre/postRender`.
This is very expensive however, as it involves parsing each page an additional time for each plugin there is.

Since element processing in `componentParser/Preprocessor` is now split cleanly on a per-element basis, and these elements are presented in a familiar format that is from `htmlParser2`, we could provide hooks in these places to let plugins tap directly into the processes(s) here.

Cons:
- It's still a less clean solution than strictly abiding by passing the string `content`, but
  - Reparsing `content` for every plugin there is isn't really scalable in achieving #984 
- `componentPreprocessor` has some amount of cyclic dependencies with `Parser` still that shouldn't need to be exposed to plugins ideally. We could find a solution to this first, or only do it for `componentParser`

na

**Additional context**

na

ang-zeyu:

• [x] Other, please explain: plugin scalability (performance)

Resolves #1188 as a first, there's still much room for polish in the plugin api

> Plugins modify the content currently only through 2 hooks, pre/postRender.
> This is very expensive however, as it involves parsing each page an additional time for each plugin there is.

- Added two new hooks, `preRenderNode` and `postRenderNode`. These tap into the same processes as `pre/postRender` at their respective stages, but directly through the recursive dom handling process.
- Convert 3 plugins as a start to test the performance improvements ( `~10%` ) in three separate commits
  - anchors plugin ⚓ 
  - plant uml
  - `<span heading>` shorthand plugin for panels ( converts to `<div slot="header">` )

Hence instead of reparsing the content for every plugin (`cheerio.load`), one can write a simple plugin through the following (shorthand plugin)
```js
if (node.name === 'panel' && node.children) {
  node.children.forEach((n) => {
  if (n.name === 'span' && n.attribs.heading !== undefined) {
    n.attribs.slot = 'header';
    n.attribs.class = n.attribs.class ? `${n.attribs.class} card-title` : 'card-title';
    delete n.attribs.heading;
  }
});
return node;
```

na, I'm leaving this undocumented for now pending #1151, would also be good if anything needs to be changed in the immediate future

**Testing instructions:**
`npm run test` should pass

Add more plugin hooks

There are only 2 interfaces for plugins to modify a page's content,
preRender and postRender. These plugin hooks do so by passing the
processed page content thus far to the plugin, requiring the plugin to
reparse the entire page content, often only to perform simple
operations.

Let's add 2 more plugin hooks to allow plugins to tap into the
recursive dom handling processes at these respective stages.

In the following commits, we adapt some of the plugins to use these new
hooks, and observe a significant performance improvement of ~10%, which
should likely scale as we add and adapt more plugins.


ang-zeyu:
resolved conflicts and converted the codeBlockCopy plugin as well

ang-zeyu:
implemented in #1403

Issue #958, opened 2019-12-20
Unexpected behaviour of components when immediately prepended by GFMD


ang-zeyu:
**Tell us about your environment**

* **MarkBind Version:** 2.70
* also tested on the latest master, briefly

**What did you do? Please include the actual source code causing the issue.**

In general, any markdown + html with the following format causes the issue
```html
markdown pre text **immediately** prepending a component, without an empty line after
<box or panel>
line 1 which behaves as per normal ( empty line still counts )
--empty line--
lines 3 onward are 'placed' outside of the component
...
...
</box or panel>
```
Only tested with box and panels, since that is likely the only places multiple paragraphs would be used.

Exceptions for the markdown pre text ( as far as I have found, exhaustively tested from this list https://markbind.org/userGuide/formattingContents.html ):
- markdown headings ( # ... )
- code blocks ( ``` ... ``` )
- markdown tables
- video embeds ( but the video disappears, leaving the link remaining instead )
- lists / blockquotes without content ( e.g. `>` does not produce the issue but `> some text` does. )

_Example 1 (boxes):_

![example1](https://user-images.githubusercontent.com/3306138/71280734-0a6a6480-2397-11ea-86cc-01dbdb2b5d56.png)

_Example 2 (panels):_

![example2](https://user-images.githubusercontent.com/3306138/71280982-bca22c00-2397-11ea-9620-b5cfa0f971bd.png)

**What did you expect to happen?**

The multiple paragraphs should be chucked into the component itself, like so

_Example 2:_ ( result achieved by leaving an empty line between the pre text and component )
![expected2](https://user-images.githubusercontent.com/3306138/71281068-fbd07d00-2397-11ea-9df0-ce8035d9da7d.png)

**What actually happened? Please include the actual, raw output.**

The subsequent paragraphs after the first are appended after the component

_Example 1:_
![result1](https://user-images.githubusercontent.com/3306138/71280741-0e968200-2397-11ea-916b-f87941598ed5.png)

_Example 2:_
![result2](https://user-images.githubusercontent.com/3306138/71280995-c0ce4980-2397-11ea-824f-d8f1f7684478.png)

wxwxwxwx9:
@ang-zeyu I've looked into this issue and here's what I discovered so far. 

From my investigation, the problem stems from `markdown-it` parsing. 

[...]

PR #1239, opened 2020-06-10
Refactor baseUrl into variablePreprocessor

### LINKED ISSUE #907: Treat nunjucks variables as safe by default

### LINKED ISSUE #1236: {{ xx | safe }} filter is always being "applied"

* **MarkBind Version:** `2.14.1`

```
// Suppose you have
<variable name="xx">
<span>... any more html ...</span>
... html
</variable>

{{ xx }}
```

Should output the escaped version of the html `&lt; ... &gt; ....`. 

However the unescaped version is outputted. (`<span>...</span>`)

---

This occurs because during the resolveBaseUrl stage, `decodeEntities` is turned on, which turns previously escaped characters back to the unescaped versions. (noticed this in an upcoming pr that inadvertedly fixes this).

Is this intended? (note that this goes against standard nunjucks behaviour)

Our docs have also been relying on this behaviour a lot, as does the 2103 site.
While it is convenient from a user standpoint, once we start adding features that pipe in data/variables from potentially unsafe sources, it poses a security problem.

On the other hand, containing html-in-variables is really quite convenient.
If we do decide to preserve the current behaviour formally (turning off `autoEscape` in nunjucks variable rendering), we should provide an `unsafe` filter of sorts

ang-zeyu:

• [x] Other, please explain: More refactoring work

Resolves #1236 Resolves #907 as well

Less code, moderately (10%) better performance, more cohesion

essentially a smaller follow up to the variable refactoring work, although I ended up needing to resolve the `xmlMode` issues as well (not much changes on the code side though) 😞 

Moved `baseUrl` rendering into `variablePreprocessor`, removing the need for an entirely separate stage of baseUrl processing which requires edge case baseUrl resolving like in #1088.

the changes in the test output.

there are a few _blank_ `<p></p>` addition / removals; These are expected, and will be annotated.

the 2103 site has a lot of similar expected `<p></p>` additions for the same reason, and I've manually verified each of those are non-breaking.

**Testing instructions:**

Refactor baseUrl into variablePreprocessor

BaseUrl processing is done in a separate stage involving repeated and
recursive parsing / rendering of the content.

This decreases cohesiveness of variable processing, and also
performance due to the repeated parsing and rendering.
It also necessitates edge-case solutions such as that in #1088 when we
need to resolve the baseUrl before the resolveBaseUrl stage has been
reached.

With a framework for variable processing now, let's move baseUrl
processing into it, solving the above said problems.

Furthermore, rendering of other variables containing html is dependent
on the extra htmlparser call in resolveBaseUrl.

Let's formally remove the need for this by using only the unescaped
nunjucks environment to render variables.


test/functional/test_site/expected/index.html
```diff
@@ -251,6 +251,7 @@ <h1 id="panel-without-heading-with-keyword">Panel without heading with keyword<a
             <h1 id="keyword-should-be-tagged-to-this-heading-not-the-panel-heading">Keyword should be tagged to this heading, not the panel heading<a class="fa fa-anchor" href="#keyword-should-be-tagged-to-this-heading-not-the-panel-heading" onclick="event.stopPropagation()"></a></h1>
             <p><span class="keyword">panel keyword</span></p>
           </panel>
+          <p></p>
```
ang-zeyu, line 254: This occurs due to html not needing a closing `<p>` tag.
Previously, this is the _intermediate_ output after markdown-it renders the html.
It renders invalid html because it is subtly invalid markdown - `panel` is not a whitelisted html block tag and cannot terminate the paragraph rule (also resulting in #958):

```
<p><span class="keyword">panel keyword</span>
</panel>
</p>
```

when rerendering the html after having passed it through cheerio, the bottom lone `</p>` is invalid in `xmlMode`, and gets erased. when rendering without `xmlMode` however, it gets expanded into an empty `p` tag by htmlparser2's forgiving nature (there is no spec for this afaik). (line 226 of htmlparser2's Parser.js)

Ultimately, the extra `<p></p>` is due to miswritten markdown;
```
<span class="keyword">
panel keyword</span>
```

would be the correct way of not inserting any additional `<p>`'s at all (note the newline). (I've left this in for now as to explain the many similar cases in the 2103 site)

test/functional/test_site/expected/index.html
```diff
@@ -356,11 +357,11 @@ <h1 id="path-within-the-boilerplate-folder-is-separately-specified">Path within
             <p>Like static include, pages within the site should be able to use files located in folders within boilerplate.</p>
             <p>Also, the boilerplate file name (e.g. <code v-pre="">inside.md</code>) and the file that it is supposed to act as (<code v-pre="">notInside.md</code>) can be different.</p>
             <p>This file should behaves as if it is in the <code v-pre="">requirements</code> folder:</p>
-            <panel src="/test_site/requirements/NonFunctionalRequirements._include_.html"><template slot="_header">
-                <p>Tested with the folllowing include</p>
-              </template>
-              <p></p>
-            </panel>
+            <p>
+              <panel src="/test_site/requirements/NonFunctionalRequirements._include_.html"><template slot="_header">
```
ang-zeyu, line 361: due to the added `/` in the source file

test/functional/test_site/expected/index.html
```diff
@@ -716,6 +715,7 @@ <h6 class="always-index" id="level-6-header-outside-headingsearchindex-with-alwa
 
 
 
+
```
ang-zeyu, line 718: due to the intermediary `endraw` removal, likewise for the above `<p></p>`'s

test/functional/test_site_special_tags/expected/index.html
```diff
@@ -84,6 +76,7 @@ <h2 id="so-far-as-to-comply-with-the-commonmark-spec">So far as to comply with t
             */
           </p>
         </abc>
+        <p></p>
```
ang-zeyu, line 79: the two `<p></p>` additions in this file occur due to miswritten markdown as well.
In the first case, the `/* ...` indicates the start of a paragraph ending after the `</abc>`
In the second case, the `if ...` does the same

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -1,73 +1,13 @@
...
-const START_ESCAPE_STR = '{% raw %}';
-const END_ESCAPE_STR = '{% endraw %}';
-const RAW_TAG_REGEX = new RegExp('{% *(end)?raw *%}', 'g');
-
-/**
- * Pads the outermost {% raw %} {% endraw %} pairs with {% raw %} {% endraw %} again.
- * This allows variables and other nunjuck syntax inside {% raw %} {% endraw %} tags
- * to be ignored by nunjucks until the final renderString call.
- */
-function preEscapeRawTags(pageData) {
```
ang-zeyu, line 16: I guess the real solution to #762 was to make nunjucks a single pass, removing the need for preescaping `raw/endraw`. A little unfortunate that the work done here is removed almost as soon as it was introduced #1220 #1049, but we could always revert it in the future if there's ever a need for multiple nunjucks passes.

ang-zeyu: @crphang could I get your opinion on this?

ang-zeyu:
rebased on latest, no changes

src/Site.js
```diff
@@ -1,4 +1,4 @@
-const cheerio = require('cheerio');
+const cheerio = require('cheerio'); require('./lib/markbind/src/patches/htmlparser2');
```
ang-zeyu, line 1: added this in all files to ensure unit tests load the patch, which is more than ever important with the extra 2 "patches" for `recognizeSelfClosing` and `lowerCaseAttributeNames`

src/Site.js
```diff
@@ -540,13 +539,17 @@ class Site {
       }
 
       /*
-       This is to prevent the first nunjuck call from converting {{baseUrl}} to an empty string
-       and let the baseUrl value be injected later.
+       We retrieve the baseUrl of the (sub)site by appending the relative to the configured base url
+       i.e. We ignore the configured baseUrl of the sub sites.
        */
-      this.variablePreprocessor.addUserDefinedVariable(base, 'baseUrl', '{{baseUrl}}');
+      const siteRelativePathFromRoot = utils.ensurePosix(path.relative(this.rootPath, base));
```
ang-zeyu, line 545: This is the main functional change, the rest are just necessary changes due to the tight coupling of `xmlMode`/rendering order/using it to unescape escaped html from removing `resolveBaseUrl`

We simply add the correct baseUrl for each (sub)site as a variable directly into `variablePreprocessor` (previously `userDefinedVariablesMap`)

src/Site.js
```diff
@@ -540,13 +539,17 @@ class Site {
...
+      const siteBaseUrl = siteRelativePathFromRoot === ''
+        ? this.siteConfig.baseUrl
+        : path.posix.join(this.siteConfig.baseUrl || '/', siteRelativePathFromRoot);
+      this.variablePreprocessor.addUserDefinedVariable(base, 'baseUrl', siteBaseUrl);
       this.variablePreprocessor.addUserDefinedVariable(base, 'MarkBind', MARKBIND_LINK_HTML);
 
-      const $ = cheerio.load(content);
+      const $ = cheerio.load(content, { decodeEntities: false });
```
ang-zeyu, line 552: `{ decodeEntities: false }` here as this file dosen't have the `cheerio.prototype.options.decodeEntities = false;` line like the others, and is the only `cheerio.load` call in the file

src/lib/markbind/src/constants.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   // src/lib/markbind/src/parser.js
-  ATTRIB_INCLUDE_PATH: 'include-path',
```
ang-zeyu, line 3: previously used for `_rebaseReference`

src/lib/markbind/src/parser.js
```diff
@@ -48,20 +39,16 @@ class Parser {
...
-    const { rootPath } = this;
     function getAbsoluteResourcePath(elem, relativeResourcePath) {
       const firstParent = elem.closest('div[data-included-from], span[data-included-from]');
       const originalSrc = utils.ensurePosix(firstParent.attr('data-included-from') || context);
       const originalSrcFolder = path.posix.dirname(originalSrc);
       const fullResourcePath = path.posix.join(originalSrcFolder, relativeResourcePath);
-      const resolvedResourcePath = path.posix.relative(utils.ensurePosix(rootPath), fullResourcePath);
-      return path.posix.join('{{hostBaseUrl}}', resolvedResourcePath);
+      const resolvedResourcePath = path.posix.relative(utils.ensurePosix(config.rootPath), fullResourcePath);
+      return path.posix.join(config.baseUrl || '/', resolvedResourcePath);
```
ang-zeyu, line 51: it's already rendered at this point, so we need to change it correspondingly to the actual baseUrl

src/lib/markbind/src/parser.js
```diff
@@ -70,7 +57,7 @@ class Parser {
         return;
       }
       const resourcePath = utils.ensurePosix(elem.attr('src'));
-      if (utils.isAbsolutePath(resourcePath) || utils.isUrl(resourcePath)) {
+      if (path.isAbsolute(resourcePath) || utils.isUrl(resourcePath)) {
```
ang-zeyu, line 60: previously `utils.isAbsolutePath` also checks for the existence of the `{{ baseUrl }}` substring. This is no longer necessary, so we could call `path.isAbsolute` directly

src/lib/markbind/src/parser.js
```diff
@@ -139,32 +123,12 @@ class Parser {
...
       node.children = cheerio.parseHTML(md.renderInline(cheerio.html(node.children)), true);
-      cheerio.prototype.options.xmlMode = true;
       break;
     case 'markdown':
       node.name = 'div';
-      cheerio.prototype.options.xmlMode = false;
       node.children = cheerio.parseHTML(md.render(cheerio.html(node.children)), true);
-      cheerio.prototype.options.xmlMode = true;
-      break;
-    case 'panel': {
```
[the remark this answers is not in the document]
ang-zeyu: this is moved to `componentPreprocessor` L149 - L155.

This chunk does the rewriting of `src="topic1.html"` into `src="topic1._include_.html"`, so it's still necessary

The main difference is the removal of the `fileExists` check, which already wasn't necessary before (there's a similar check done in `componentPreprocessor`)
The second difference would be `path.dirname(path.join('{{hostBaseUrl}}', path.relative(config.rootPath, src)))` using `config.baseUrl` directly instead of `{{ hostBaseUrl }}`, since we no longer have / render `{{ hostBaseUrl }}`

src/lib/markbind/src/parsers/componentParser.js
```diff
@@ -483,6 +486,10 @@ function postParseComponents(node) {
   } catch (error) {
     logger.error(error);
   }
+
+  if (node.attribs) {
+    delete node.attribs[ATTRIB_CWF];
```
ang-zeyu, line 491: this was previously done in `_rebaseReference`

src/lib/markbind/src/patches/htmlparser2.js
```diff
@@ -1,4 +1,23 @@
...
+ * 1. Defaulting to self closing tag recognition
+ * 2. Disabling automatic attribute name conversion to lower case
+ * 3. Recognising text in markdown inline code / code blocks as raw text
+ * 4. Ability to inject/whitelist certain tags to be parsed like script/style tags do. (special tags)
+ */
+
+const { Tokenizer, Parser } = require('htmlparser2');
+
+// Self closing tag recognition patch
+Parser.prototype.onselfclosingtag = function () {
```
ang-zeyu, line 13: https://github.com/fb55/htmlparser2/wiki/Parser-options

Enables `recognizeSelfClosing` and disables `lowerCaseAttributeNames` so there isn't a need to keep passing in those options.
Not sure if passing it in may be better on the other hand though, in the case of a new developer very familiar with "anonymous self closing tags" in `html`, they would have to find this file and the comment above if they're wondering why it behaves as such

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -81,19 +80,17 @@ function _getSrcFlagsAndFilePaths(element, context, config) {
...
   } else {
     const includePath = decodeURIComponent(includeSrc.path);
 
-    if (baseUrlRegex.test(includePath)) {
-      // The baseUrl has not been resolved during pre-processing, but we need the source file path
-      const parentSitePath = urlUtils.getParentSiteAbsolutePath(context.cwf, config.rootPath,
-                                                                config.baseUrlMap);
-      filePath = path.resolve(parentSitePath, includePath.replace(baseUrlRegex, ''));
+    if (path.posix.isAbsolute(includePath)) {
+      // {{ baseUrl }} (or simply '/...' if baseUrl === '') was used in this src attribute,
```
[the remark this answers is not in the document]
ang-zeyu: yup, will reword the comment a little

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -123,39 +120,42 @@ function _getSrcFlagsAndFilePaths(element, context, config) {
  * and adds the appropriate include.
  */
 function _preProcessPanel(node, context, config, parser) {
-  const element = node;
```
ang-zeyu, line 126: irrelevant style change =C, let me know if you want to move it to another PR!

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -123,39 +120,42 @@ function _getSrcFlagsAndFilePaths(element, context, config) {
...
 
   if (!isUrl && hash) {
-    element.attribs.fragment = hash.substring(1);
+    node.attribs.fragment = hash.substring(1);
   }
 
-  element.attribs.src = filePath;
+  const { fragment } = node.attribs;
+  const resultPath = utils.setExtension(path.relative(config.rootPath, filePath), '._include_.html');
+  const baseUrlPrependedPath = path.posix.join(`${config.baseUrl}/`, utils.ensurePosix(resultPath));
```
ang-zeyu, line 151: as mentioned in the intermediate processing comment, the `fileExists` check is done above in `_getFileExistsNode`, and `{{ hostBaseUrl }}` is replaced with `config.baseUrl`

src/lib/markbind/src/utils/nunjuckUtils.js
```diff
@@ -1,73 +1,13 @@
...
-
 module.exports = {
-  renderRaw(pageData, variableMap = {}, options = {}, autoescape = true) {
-    const escapedPage = preEscapeRawTags(pageData);
-    if (autoescape) return commonEnv.renderString(escapedPage, variableMap, options);
-    return unescapedEnv.renderString(escapedPage, variableMap, options);
-  },
-  renderString(pageData, variableMap = {}, options = {}, env = commonEnv) {
-    return env.renderString(pageData, variableMap, options);
+  renderRaw(pageData, variableMap = {}) {
```
ang-zeyu, line 7: a little strange to call it `renderRaw` now that we don't preescape it, but I'll be touching on this shortly in the next PR that fixes #931

test/unit/parser.test.js
```diff
@@ -121,7 +121,7 @@ test('includeFile replaces <include src="doesNotExist.md" optional> with empty <
 
   const expected = [
     '# Index',
-    '<div/>',
```
ang-zeyu, line 124: consequence of using `xmlMode` previously; `<div></div>` is rendered to `<div/>` which is invalid

src/lib/markbind/src/patches/htmlparser2.js
marvinchin: I think we can add some comments here to provide the context as to why we are patching this. I think it makes sense to simplify usage at the call site, and this file is the natural first place to look if anyone is wondering why the behaviour might not be as they expected.

src/lib/markbind/src/preprocessors/componentPreprocessor.js
```diff
@@ -123,39 +120,42 @@ function _getSrcFlagsAndFilePaths(element, context, config) {
...
   }
 
   if (!isUrl && hash) {
-    element.attribs.fragment = hash.substring(1);
+    node.attribs.fragment = hash.substring(1);
   }
 
-  element.attribs.src = filePath;
+  const { fragment } = node.attribs;
+  const resultPath = utils.setExtension(path.relative(config.rootPath, filePath), '._include_.html');
```
marvinchin, line 150: Would it make sense to rename `resultPath` -> `relativePath` and `baseUrlPrependedPath` -> `absolutePath`? I feel like that might better reflect what they are.

ang-zeyu:
Thanks for the review again @marvinchin!

PR #1246, opened 2020-06-13
Refactor developer docs

damithc:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [x] Documentation update

**What is the rationale for this request?**, **What changes did you make? (Give an overview)**

See the proposed commit message.

**Testing instructions:**

Read the DG via the Netlify preview of the PR.

**Proposed commit message: (wrap lines at 72 characters)**

```
Refactor developer docs

Let's refactor the dev docs in the following ways to improve the
overall organization:
* Split up into separate pages such as setting up, design, workflow etc.
* Refer to se-education.org/guides when possible
* Improve phrasing, layout, formatting

This reorganization also aims to increase the consistency with other
sister projects such as RepoSense
```

Issue #903, opened 2019-06-26
Support a way to set the primary color of a site


damithc:
Inspired by https://yakworks.github.io/mkdocs-material-components/getting-started/#primary-colors

tlylt:
> Inspired by [yakworks.github.io/mkdocs-material-components/getting-started/#primary-colors](https://yakworks.github.io/mkdocs-material-components/getting-started/#primary-colors)

the above link is dead... it's now available at this [link](https://yakworks.github.io/docmark/getting-started/#primary-colors) instead.
Screenshot:
![image](https://user-images.githubusercontent.com/41845017/152343582-9b31e8c4-8ff9-4202-82c2-1ab1d894a7e1.png)

ang-zeyu:
This won't be accomplished anytime soon, but to list the key issues in the various linked discussions above:
- bootstrap **themes** seem to conflict with this both from an implementation and usage standpoint.
- I'm personally for removing official support for the 15+ bootstrap themes, and developing our components with bootstrap, which have proven very difficult to maintain properly (there are many, many issues if you dig around).
- (personally) everything looks very bootstrap-ey
- that said, implementing a completely custom theming scheme isn't easy, I haven't exactly thought of a concrete proposal either (which might flesh out more issues =X) We should probably limit the official support to **color-only** (bootstrap themes do much more than that) theming for a start to make it feasible
  - A pretty solid example direction I think, is mdBook's https://github.com/rust-lang/mdBook/blob/master/src/theme/css/variables.css

wdyt @jonahtanjz @ryoarmanda @damithc?

ps: starting the discussion now as it affects #1702

jonahtanjz:
> * I'm personally for removing official support for the 15+ bootstrap themes, and developing our components with bootstrap, which have proven very difficult to maintain properly (there are many, many issues if you dig around).
> * (personally) everything looks very bootstrap-ey

Just to clarify, do you mean removing bootstrap support in general (and have our own styling for our components) or only removing the support for themes but still maintain the use of bootstrap for our components?

ang-zeyu:
In general, to remove the "bootstrap look", which includes removing bootstrap themes completely. We can certainly still use bootstrap for our components (e.g. `d-none`), barring things like `bg-dark` / `alert-warning` (which would then be customisable instead under a coloring system). Bootstrap will also still be available to the author.

[...]

PR #1248, opened 2020-06-14
Merge /vue-strap into /markbind

### LINKED ISSUE #411: Explore the feasibility of merging vue-strap into the MarkBind repository
Currently, we maintain the vue-strap repository independent of the MarkBind repository.

There are some inconveniences and issues:

* A bit of labour when reviewing PRs. After building `vue-strap.min.js` on the vue-strap repo (the "binary" file of vue-strap), we must manually copy it into our MarkBind repo, before we can test the components.
* Because the documentation for our components are now situated in MarkBind rather than vue-strap, we must open two PRs (one in each repo) whenever we update a component
    * For example, when adding a new attribute to the component, we need to open one PR in vue-strap to update the logic, another PR in MarkBind to update the documentation.
* The original vue-strap repo is dead. It seems like there is no reason for keeping it separate anymore, since we aren't likely to receive any new updates from upstream. Also, a lot of components were purged during the migration, so our vue-strap is completely serving the interest of MarkBind, and it seems no longer appropriate to view it as a derivation of vue-strap.

Explore whether it is feasible to merge the code of vue-strap into the MarkBind repository.

Alternatively, we can also solve some of the issues above by:
* Creating a playground functionality for vue-strap.
* Creating and maintaining vue-strap as a proper npm package under MarkBind, and encourage developers to use `npm link` to avoid having to manually copy and paste `vue-strap.min.js`.
* Rebranding vue-strap. This might be a good reason to relocate the documentation back on vue-strap, while allowing us to use MarkBind syntax in those documentations.

### LINKED ISSUE #981: Request to rebrand vuestrap as an independent npm package under MarkBind org

This is an extension of issue #411 , since we've been independently maintaining vue strap for over a year now, we should probably treat our fork of vue-strap as our own library of Vue components. As described in the previous issue, it's quite troublesome to build and copy `vue-strap.min.js` when making changes to the Vue components of MarkBind.

As mentioned in #411:

> Creating and maintaining vue-strap as a proper npm package under MarkBind, and encourage developers to use `npm link` to avoid having to manually copy and paste `vue-strap.min.js`.

We can probably name it VueBind?

Migrate current components from vuestrap to bootstrapvue. However, although bv is actively maintained, it's probably too complicated if we are to maintain a fork of it. MarkBind only makes use of a small subset of bv's huge component library.

**Additional context**

na

ang-zeyu:

• [x] Other, please explain:

Resolves #411 
Resolves #981 

Merge /vue-strap repo into /markbind

I aimed for minimal changes for now, but one thing lead to another as explained in the commit messages (particularly the last one), and had to make quite a few changes still.

na

**Testing instructions:**
- `npm run build` should build the minified `vue-strap` as per normal
-  `npm run test/lint` should pass, and the snapshot tests should also be run

Merge /vue-strap into /markbind

Markbind's vue component definitions are stored in the vue-strap fork.
This has some implications on development and maintainer workflow (see
issue #411 for more detail), most notably needing to manually copy the
minified bundle to test changes in /vue-strap.

An alternative was raised in issue #981 to alleviate this, that is to
make /vue-strap a scoped package. This still has several drawbacks as
compared to a complete repo merge as outlined in #411, such as needing
to run tests on both repos.

Hence, let's change the directory structure and configurations of
/markbind to accomodate the linting, testing and building processes
from /vue-strap.


ang-zeyu:
@acjh one thing led to another (as explained particularly in the last commit message), and I needed to make quite a few changes to ensure the build, linting and testing processes didn't break minimally.

Most notably I ended up deciding to remove `package.json` of /vue-strap (preserving the readme/license notices). Not too sure if there are benefits to retaining it at this point, but it caused some of the issues above generally because of how eslint / jest resolves some things with respect to the closest package.jsons ( details in commit messages ).

* ci currently fails, but it works on windows systems. still investigating the cause.

Future plans:
- Update dev guide after #1246 
- Move up webpack config files to `frontend/` directory in preparation for the below point, and then change CI/CD to make the release process easier
- Merge `assets` into the webpack build process
- Setup hot/live reload for frontend development
- ...

ang-zeyu:
Thanks for the thorough review! Wasn't expecting one so quick. Apologies for the late reply as well, needed some time to give it more thought.

> Thanks for this attempt.
> 
> However, we may be moving in the wrong direction by merging the tests and dependencies.
> We can provide a single helper command in package.json to run both tests, but not like this.

To provide more context, on top of the reasons mentioned in the commit messages, 
I aim to eventually streamline the whole frontend build process in light of many eventual goals:

- move assets into the webpack build process as well
  - allowing us do some tree shaking of bootstrap-vue
  - maintain a consistent transpiling / minification / ... process to save some headaches
- eventually migrate to sass stylesheets (also inside the vue component `<style>`s themselves) in order to facilitate #903 using sass variables
- live/hot (i'm using this interchangably as there will always be cases you need to do a manual refresh) reload for frontend ( components + assets folder )
- SSR - I think this should be the eventual goal particularly given the nature of education sites (high amount of components, also why FOUC is pretty noticeable)
- ... probably more I forgot about here and there, but these should be the main ones =C

I haven't thought of the specifics of course, more problems will always arise when actually implementing these, but just on a quick glance I think going for a mono-package structure for now would help tremendously in these goals, since many of /vue-strap's dependencies are necessary for the above.

On the other hand, sub packages do provide a clear seperation of dependencies. On this note and the above raised points, perhaps we could eventually create a new sub-package for the whole of `frontend/`?
Removing the package for now (providing more flexibility) then reintroducing it later may also be easier than migrating later on, since we can't really iron out the challenges that will be faced in attaining above said goals.

TLDR: perhaps we can shift back when we start to have a ton of dependencies lumped together and it's not clear what is used where.

src/Page.js
```diff
@@ -346,7 +346,6 @@ class Page {
     this._collectNavigableHeadings($, $.root()[0], elementSelector);
   }
 
-
```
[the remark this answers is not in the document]
ang-zeyu: this occurs due to upgrading to /vue-strap's version of eslint; likely a bug fix

acjh:
> To provide more context, on top of the reasons mentioned in the commit messages,
I aim to eventually streamline the whole frontend build process in light of many eventual goals:

Even so, we shouldn't make those changes (that should be assessed separately) in this PR.
Let's aim for a clean merge and stick to the scope of #411 that considers the existing needs.

> Removing the package for now (providing more flexibility) then reintroducing it later may also be easier than migrating later on

It's very messy now. I'm not sure why you phrase it differently, you are just "migrating" now.

ang-zeyu:
> Even so, we shouldn't make those changes (that should be assessed separately) in this PR.
Let's aim for a clean merge and stick to the scope of #411 that considers the existing needs.

Noted. The commit organization is quite messy. Will split this into 2/3 prs. This PR should be able to retain up to the first 4 commits though

> I'm not sure why you phrase it differently, you are just "migrating" now.

I guess I meant "sticking to the existing structure for now then migrating later" 😅

frontend/components/README.md
```diff
@@ -0,0 +1,5 @@
+# vue-strap
+*Many of the components under this folder was from markbind's forked project of VueStrap.*
+
+## License
```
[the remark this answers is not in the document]
ang-zeyu: Looking at this list again, was there a reason you seperated "ported" and "modified" actually?
Iirc at least 90% (haven't looked at the others) of the components had markbind contributors in `git annotate`; We could just lump them together on that note

ang-zeyu:
updated @acjh, just 5 commits now
- copy and paste /vue-strap
- update ignore configs (eslint, stylelint gitignore)
- remove unnecessary files
- update README.md
- update package.json

frontend/components/README.md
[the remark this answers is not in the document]
acjh: I didn't mean to separate "ported" and "modified" (both lists mention "ported"), but to separate components that did not originate from [yuche/vue-strap](https://github.com/yuche/vue-strap).

frontend/components/package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "scripts": {
+    "build": "webpack --progress --hide-modules --config  webpack.build.js",
+    "test": "jest",
+    "updatesnapshot": "jest --updateSnapshot",
+    "builddev": "webpack --progress --hide-modules --config  webpack.dev.js",
+    "lint": "./node_modules/.bin/eslint . --ext .js,.vue; exit 0",
+    "lintwin": "./node_modules/.bin/eslint . --ext .js,.vue",
+    "lintfix": "./node_modules/.bin/eslint . --fix --ext .js,.vue"
+  },
```
acjh, line 71: Do we want to add helper commands (e.g. `vuebuild`, `vuelint`, `lintall`) that `cd frontend/components` and then run these in the `"scripts"` of the root package.json?
Or do we plan to do it in a separate PR?

package.json
```diff
@@ -81,7 +81,8 @@
   "jest": {
     "testPathIgnorePatterns": [
       "/node_modules/",
-      "test/functional"
+      "test/functional",
+      "frontend"
```
acjh, line 85: Alphabetical order?

frontend/components/README.md
```diff
@@ -0,0 +1,37 @@
+# MarkBind Vue components
```
acjh, line 1: Much of the content removed in https://github.com/MarkBind/markbind/pull/1248/commits/99245c49ac7c9e573cf91bfb775e3f8d87b04449 is still relevant and should be restored.

frontend/components/package.json
ang-zeyu: if we go with removing `package.json` and use nested eslintrcs that won't be necessary, but either way a separate PR would be better.
Right now i'm aiming for "copy and paste + make ci work + edit whatever documentation dosen't make sense anymore" (I'll add in devdocs as well once #1246 gets merged)

frontend/components/README.md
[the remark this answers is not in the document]
ang-zeyu: ahhh ok, I'll reseparate it under that with the additional custom directives heading

frontend/components/README.md
```diff
@@ -0,0 +1,37 @@
...
+### Components modified from [Markbind/vue-strap](https://github.com/MarkBind/vue-strap)
+
+- Pic.vue
+- Question.vue
+- Retriever.vue
+- Searchbar.vue
+- SearchbarPageItem.vue
+- Thumbnail.vue
+- TipBox.vue
+- Affix.vue
```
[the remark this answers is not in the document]
ang-zeyu: resolved per https://github.com/MarkBind/markbind/pull/1248#discussion_r439834040

frontend/components/.eslintignore
```diff
@@ -0,0 +1,48 @@
...
+src/Popover.vue
+src/Progressbar.vue
+src/Radio.vue
+src/Select.vue
+src/Slider.vue
+src/Spinner.vue
+src/Tab.vue
+src/TabGroup.vue
+src/Tabset.vue
+src/Thumbnail.vue
```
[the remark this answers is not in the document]
ang-zeyu: > Actually, I ignored .vue files in MarkBind/vue-strap#29 to avoid large code style changes

I'll open an issue for this sometime, but it's rather unrelated at the moment.
Perhaps we could use git annotate to see the % of lines changed to decide which to unignore, but as we migrate more components to bootstrap-vue this shouldn't be necessary 😄

ang-zeyu:
@acjh ready for another review, resolved your comments and updated with a 6th commit updating the dev guide as well

ang-zeyu:
@acjh on the multi package topic, I was curious how much larger projects handled the amount of subpackages they had - and found this https://lerna.js.org/

This potentially resolves all the concerns I had with duplicated dependencies / `npm installs` I had previously. I've yet to test the specifics though (like whether jest, babel-jest, etc. plays nice with this which gave quite a few issues on its own already)  \*crosses fingers 🤞\*

It even has some features like lerna publish to publish multiple packages with the same version simultaneously, which can help improve multi package release processes

If all goes well (not having compatibilty issues with nested babel configs, etc.), I feel this would be the next logical step to take before attempting any more of the plans https://github.com/MarkBind/markbind/pull/1248#issuecomment-643739166.

What do you think of it?

docs/devGuide/projectManagement.md
```diff
@@ -109,18 +109,18 @@ For general best practices, refer to the guide [_Working with PRs_ @SE-EDU](http
...
-   $ cp <VUE_STRAP_REPO>/dist/vue-strap.min.js <MARKBIND_REPO>/asset/js/vue-strap.mins.js
-   # for each test site's expected folder:
-   $ cp <VUE_STRAP_REPO>/dist/vue-strap.min.js <MARKBIND_REPO>/test/functional/<TEST_SITE_NAME>/expected/markbind/js/vue-strap.min.js
+   $ cp frontend/components/dist/vue-strap.min.js asset/js/vue-strap.min.js
    ```
+   
+   Run `npm updatetest` to and ensure only the minified bundle is copied to each of the expected test site's folders as well.
 
-   Commit the new vue-strap assets (replace `XYZ` with version number).
+   Commit only the new vue-strap assets (replace `XYZ` with version number). (if there are other changes from `npm updatetest`, discard them for now - refer to **Rebuild the test files**)
```
acjh, line 123:
```suggestion
   Commit only the new vue-strap assets (replace `XYZ` with version number).
```

frontend/components/README.md
```diff
@@ -0,0 +1,37 @@
+# MarkBind Vue components
+
+This folder contains MarkBind's [vuejs](https://vuejs.org/) UI components.
```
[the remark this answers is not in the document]
acjh: Not resolved.

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+Most of the styles are based on Bootstrap's markup and CSS. As a result no dependency Bootstrap's JavaScript is required. The only required dependencies are:
```
acjh, line 6:
```suggestion
Most of the styles are based on Bootstrap's markup and CSS, but no dependency on Bootstrap's JavaScript is required. The only required dependencies are:
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+* [Vue.js](http://vuejs.org/) (required ^v2.x.x, test with v2.5.16).	They were modified to suit MarkBind's needs for educational and documentation websites.
```
acjh, line 8: The additional line seems out of place here.

```suggestion
* [Vue.js](http://vuejs.org/) (required ^v2.x.x, test with v2.5.16).
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+* [Bootstrap CSS](http://getbootstrap.com/) (required 4.x.x, test with 4.1.1). VueStrap doesn't depend on a very precise version of Bootstrap.
+
+
+## Components / directives
+
+Many of the components under this folder were originally from [MarkBind's forked project](https://github.com/MarkBind/vue-strap) of [VueStrap](https://github.com/yuche/vue-strap), and were modified to suit MarkBind's needs for educational and documentation websites.
```
acjh, line 14:
```suggestion
Many of the components under this folder were originally from [MarkBind/vue-strap](https://github.com/MarkBind/vue-strap), forked from [yuche/vue-strap](https://github.com/yuche/vue-strap), and modified to suit MarkBind's needs for educational and documentation websites.
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+Some custom directives were also added for MarkBind's own use.
```
acjh, line 17:
```suggestion
Some custom directives were also added for MarkBind's use.
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+- TipBox.vue
+
+
+### Custom directives ported from [Markbind/vue-strap](https://github.com/MarkBind/vue-strap)
+
+- Closeable.js
+- Float.js
+
+
+### VueStrap components modified for use in MarkBind; ported from [Markbind/vue-strap](https://github.com/MarkBind/vue-strap), a fork of [yuche/vue-strap](https://github.com/yuche/vue-strap):
```
acjh, line 37: Headings shouldn't be this long. And already mentioned above.

```suggestion
### VueStrap components modified for use in MarkBind
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Tab.vue
+- TabGroup.vue
+- Tabset.vue
+- Typeahead.vue
+
+
+## Installation
+
+### Browser globals
+The `dist` folder contains `vue-strap.js` and `vue-strap.min.js` with all components exported in the <code>window.VueStrap</code> object.
```
acjh, line 53:
```suggestion

The `dist` folder contains `vue-strap.js` and `vue-strap.min.js` with all components exported in the `window.VueStrap` object.
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+```html	- Pic.vue
```
acjh, line 55:
```suggestion
```html
```

frontend/components/README.md
```diff
@@ -0,0 +1,89 @@
...
+See the [documentation](https://markbind.org/userGuide/usingComponents.html) with live editable examples.	
+
+
+## Local Setup
+* Install with `npm install`.
+* Build with `npm run build`.	
+
+
+## Lint the code for any code and style errors using ESLint
+* For Unix: `npm run lint`
```
acjh, line 83: Do check the diff when committing: https://github.com/MarkBind/markbind/commit/949d9097eba63b37fdb1e7ab721c9ad5778f380c

```suggestion
* For Unix: `npm run lint`
```

frontend/components/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "node-sass": "^4.13.1",
+    "sass-loader": "^8.0.2",
+    "style-loader": "^1.1.3",
+    "vue-hot-reload-api": "^2.3.4",
+    "vue-jest": "^3.0.5",
+    "vue-template-compiler": "^2.6.11",
+    "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.10.3"
+  },
+  "scripts": {
```
acjh, line 64: Let's further adjust the order based on https://docs.npmjs.com/files/package.json.

```diff
  "name": "vue-strap",
  "version": "2.14.1",
  "description": "Bootstrap components built with Vue.js",
+ "keywords": [
+ "homepage": "https://markbind.org",
+ "license": "MIT",
+ "files": [
  "main": "dist/vue-strap.js",
  "repository": {
- "homepage": "https://markbind.org",
- "license": "MIT",
+ "scripts": {
  "dependencies": {
- "files": [
- "keywords": [
  "devDependencies": {
- "scripts": {
```

package.json
```diff
@@ -80,6 +80,7 @@
   },
   "jest": {
     "testPathIgnorePatterns": [
+      "frontend",
```
acjh, line 83:
```suggestion
      "/frontend/",
```

if we're placing before `/node_modules/`.

acjh:
> What do you think of it?

I don't think we currently need Lerna. MarkBind has way fewer packages than several of the repos — only 3 (we haven't adopted my [npm plugin proposal](https://github.com/MarkBind/markbind/pull/474#issuecomment-448994052)) — and we currently only publish one.

> On second thought, I'm still not sure if there's much point in maintaining a seperate package just for the vue components though. To reiterate,

If we develop a feature that adds significant value and absolutely requires combining the dependencies, then we will cross that bridge.

Separation of concerns has value. It's not something we should abandon for convenience.

frontend/components/package.json
ang-zeyu: will do the ones for `markbind/markbind-cli` in a seperate PR as well

ang-zeyu:
Thanks for the thorough suggestions! @acjh Rectified accordingly

> Separation of concerns has value.

Definitely, I just don't think that having the formalities of a `package.json` indicating a subpackage helps much with this. SoC should be primarily achieved with when looking at the code changes during PR reviews imo. A semantic directory structure should also help in achieving SoC in place of `package.json`.
Simply having a `package.json` dosen't add any value to me in this sense, only overhead in dependency management.

That said, I don't have a lot of experience working with multi package environments; I'm willing to see where this goes for now, if need be we can always eliminate it in the future. 

> I don't think we currently need Lerna. MarkBind has way fewer packages than several of the repos — only 3 (we haven't adopted my npm plugin proposal) — and we currently only publish one.

👍, could reconsider if we have more / actually restructure `markbind/markbind-cli` for their respective purposes ( reason being we might want to move unit tests in there, and want to avoid running 3 npm run tests )

I just found out that `npm install <folder>` [hoists](https://docs.npmjs.com/cli/install) dependencies as well. Minimally we should use that to avoid duplicated dependency bloat. (in fact we should probably be using this for `markbind/markbind-cli` as well to avoid stating dependencies in `markbind-cli` which aren't even directly used by it)

frontend/components/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "version": "2.14.1",
+  "description": "Bootstrap components built with Vue.js",
+  "keywords": [
+    "bootstrap",
+    "vue-bootstrap",
+    "vue-component",
+    "vue",
+    "markbind"
+  ],
+  "homepage": "https://markbind.org",
```
acjh, line 12: should this be https://markbind.org/userGuide/usingComponents.html?

frontend/components/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "license": "MIT",
+  "files": [
+    "LICENSE",
+    "README.md",
+    "CHANGELOG.md",
+    "src",
```
acjh, line 18: seems there was never `CHANGELOG.md` and it's not necessary to package `src`.

frontend/components/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "main": "dist/vue-strap.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MarkBind/markbind.git"
+  },
+  "scripts": {
+    "build": "webpack --progress --hide-modules --config  webpack.build.js",
+    "test": "jest",
+    "updatesnapshot": "jest --updateSnapshot",
+    "builddev": "webpack --progress --hide-modules --config  webpack.dev.js",
```
acjh, line 30: Alphabetical order?

frontend/components/package.json
ang-zeyu: hmm might be better to just use `markbind.org` in case we shift `usingComponents` next time. it contains some irrelevant things like the page/sitenav too

frontend/components/package.json
ang-zeyu: resolved. perhaps we could add https://www.npmjs.com/package/sort-package-json as a dev dependency at another time 😅

frontend/components/package.json
acjh: Ok.

acjh:
Btw, I suggest using `git commit --fixup` + `git rebase --autosquash` flow instead of constantly force-pushing for easier review with clearer diff.

https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash

ang-zeyu:
> Btw, I suggest using `git commit --fixup` + `git rebase --autosquash` flow instead of constantly force-pushing for easier review with clearer diff.
> 
> https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash

thanks for the review and tip!

I don't quite get the force pushing though - since the commits after the fixup-ed commit have their hashes changed, wouldn't that still require a force push? (or do you mean to do the `git rebase --autosquash` part after the reviewer has approved the fixup commit?)

acjh:
Yup.

PR #1253, opened 2020-06-15
Hoist markbind package dependencies

ang-zeyu:

• [x] Other, please explain: Dependency maintainability

- Reduce dependency management overhead in managing the duplicated dependencies in both packages, many of which `markbind-cli` does not use as well.
- Update all relative `markbind` imports to use the `markbind` package import instead to discourage writing long relative imports

- removed all dependencies which are being used in markbind but not markbind-cli, then running a `npm install && npm install src\lib\markbind
- added a missing markdown-it dependency for `markbind`
- use the patched version of markdown from the `markbind` package for one of the tests instead of `markdown-it` to remove the `markdown-it` dependency; We have quite a few patches / added plugins for the behaviour in `markdown-it` which may interfere with the the plugin being tested itself as well.
- use package imports for `markbind-cli`

Hoist markbind core package dependencies

Markbind-cli lists many packages under as its dependencies which are
used by the core markbind package but not markbind-cli.
Importing files from markbind from markbind-cli also requires writing
long relative imports.

Let's use npm install <folder> instead to hoist these packages to
markbind-cli's node_modules.
Let's also use package imports for importing the core markbind library
files.


.eslintignore
```diff
@@ -1,4 +1,5 @@
 *.min.*
+node_modules
```
ang-zeyu, line 2: added this as there are still some unhoisted dependencies (`moment`) that will be installed in `markbind` (lib)

__mocks__/fs-extra-promise.js
```diff
@@ -135,7 +135,6 @@ fs.copySync = (src, dest) => {
  */
 fs.readJsonSync = filePath => JSON.parse(fs.readFileSync(filePath, 'utf8'));
 
-
```
ang-zeyu, line 138: these occur due to dependency upgrades of eslint

package.json
```diff
@@ -71,8 +53,9 @@
     "ciwin:all": "npm ci && npm run install:components",
     "csslint": "stylelint **/*.css",
     "csslintfix": "stylelint **/*.css --fix",
-    "install:all": "npm install && npm run install:components",
+    "install:all": "npm install && npm run install:markbind && npm run install:components",
     "install:components": "cd frontend/components && npm install",
+    "install:markbind": "npm install src/lib/markbind",
```
ang-zeyu, line 58: hence, for example if you only updated the markbind core package's dependencies, simply run `install:markbind`.
If multiple packages had their dependencies  upgraded, run `install:all` instead. (documented under **workflow** under dev guide)

src/Page.js
```diff
@@ -4,20 +4,21 @@ const fs = require('fs-extra-promise');
...
+const { ensurePosix } = require('markbind/src/utils');
+const MarkBind = require('markbind/src/parser');
+const md = require('markbind/src/lib/markdown-it');
+const utils = require('markbind/src/utils');
+
 const logger = require('./util/logger');
-const MarkBind = require('./lib/markbind/src/parser');
-const md = require('./lib/markbind/src/lib/markdown-it');
-const utils = require('./lib/markbind/src/utils');
+const FsUtil = require('./util/fsUtil');
```
acjh, line 21: Move to before `const logger`?

test/unit/markdown-it-icons.test.js
```diff
@@ -1,7 +1,6 @@
 const expectedOcticon = require('@primer/octicons')['git-pull-request'].toSVG();
 
-const markdownIt = require('markdown-it')()
-  .use(require('../../src/lib/markbind/src/lib/markdown-it/markdown-it-icons'));
+const markdownIt = require('markbind/src/lib/markdown-it/index');
```
acjh, line 3:
```suggestion
const markdownIt = require('markbind/src/lib/markdown-it');
```

src/Site.js
```diff
@@ -6,12 +6,12 @@ const path = require('path');
...
 const ProgressBar = require('progress');
 const walkSync = require('walk-sync');
-const VariablePreprocessor = require('./lib/markbind/src/preprocessors/variablePreprocessor');
-const njUtil = require('./lib/markbind/src/utils/nunjuckUtils');
-const injectHtmlParser2SpecialTags = require('./lib/markbind/src/patches/htmlparser2');
-const injectMarkdownItSpecialTags = require(
-  './lib/markbind/src/lib/markdown-it/markdown-it-escape-special-tags');
-const utils = require('./lib/markbind/src/utils');
+
+const injectHtmlParser2SpecialTags = require('markbind/src/patches/htmlparser2');
```
acjh, line 10: Can remove the import on Line 1.

src/plugins/default/markbind-plugin-footnotes-popovers.js
```diff
@@ -1,5 +1,5 @@
 const cheerio = module.parent.require('cheerio');
-const { parseComponents } = require('../../lib/markbind/src/parsers/componentParser.js');
+const { parseComponents } = require('markbind/src/parsers/componentParser.js');
```
acjh, line 2:
```suggestion
const { parseComponents } = require('markbind/src/parsers/componentParser');
```

test/unit/parser.test.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path');
 const fs = require('fs');
-const Parser = require('../../src/lib/markbind/src/parser.js');
-const VariablePreprocessor = require('../../src/lib/markbind/src/preprocessors/variablePreprocessor');
+const Parser = require('markbind/src/parser.js');
```
acjh, line 3:
```suggestion
const Parser = require('markbind/src/parser');
```

docs/devGuide/workflow.md
```diff
@@ -125,3 +125,17 @@ ESLint has [integrations with popular editors](https://eslint.org/docs/user-guid
 There are several versions of the _test and linting scripts_ in the main `package.json` that execute for only one of the main or sub packages,
 feel free to look into `package.json` and use them as you see fit!
 </box>
+
+## Updating dependencies
+
+We use `npm install <package folder>` to <tooltip content="take a look under the command in the link!">[hoist](https://docs.npmjs.com/cli/install)</tooltip> the markbind core package's dependencies to the root `node_modules`, avoiding dependency duplication between the `markbind-cli` and `markbind` packages where possible.
+
+Hence, when updating dependencies of the `markbind` core package, be sure to run the `npm install src/lib/markbind` command, or simply the `npm run install:markbind` script. If you upgraded the dependencies of multiple packages, simply run `npm run install:all` instead.
```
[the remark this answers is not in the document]
ang-zeyu: > Can consider moving src/lib/markbind to packages/core similar to Angular:

Agree with this, `src/lib` seems a little awkward considering the main package is supposed to be concerned with `cli` only.

I tested a simple directory restructure as well, seems like the diffs are pretty manageable. Shall we deal with this in this pr?

docs/devGuide/workflow.md
```diff
@@ -125,3 +125,17 @@ ESLint has [integrations with popular editors](https://eslint.org/docs/user-guid
...
+
+If the dependency <tooltip content="i.e. also listed under the root `package.json`">is also used</tooltip> in the root package, make sure to update it's <tooltip content="as listed in the `package.json` file">version number</tooltip> as well!
+
+<box type="warning">
+
+Conversely, when updating dependencies of the *root* package, be sure to still check whether any subpackages list this as a dependency!
+If so, follow the above process as well.
```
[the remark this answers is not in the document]
ang-zeyu: > this encourages us to ensure that the core package works

Yup, I think this should be a good direction to move in, and is sort of the second main point of this pr. Otherwise it somewhat defeats the purpose of having the `markbind/markbind-cli` separation. (this also applies if `frontend/components` and `markbind-cli` were to have any duplicated dependencies)

Admiteddly it is a little awkward, which is why I suppose tools like lerna exist (seems like `lerna add` does this for all packages in there) that help keep dependency versions consistent across all packages. I'm not too sure if there is a purely npm solution as well though. 🤔

> Actually, do we still need to list packages in package.json if they are present in src/lib/markbind/package.json?

There's an eslint rule necessitating this, it's also good practice I would think, as it makes clear the dependencies of the main (`cli`) package.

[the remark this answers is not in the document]
ang-zeyu: > Actually, do we still need to list packages in package.json if they are present in src/lib/markbind/package.json?

But this shouldn't really be applicable once we move all the other non-cli things like `Page.js/Site.js` into `markbind`

src/Page.js
ang-zeyu: 😅

src/Site.js
ang-zeyu: not too sure what you meant here,

`const cheerio = require('cheerio'); require('./lib/markbind/src/patches/htmlparser2');`?
`cheerio` is used in one instance below, the patch is necessary to ensure unit tests load with the patch

ang-zeyu: changed `require('./lib/markbind/src/patches/htmlparser2');` to `require('markbind/src/patches/htmlparser2');` though if that's what you meant

ang-zeyu:
Thanks for the review @acjh! Pushed the relevant fixups

src/Site.js
acjh: We're doing the same import on this line 🙊 

But if it's clearer that we're doing the patch, then I'm ok with the duplicate import.

docs/devGuide/workflow.md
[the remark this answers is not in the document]
acjh: Let's do that in a subsequent PR since it's a significant change that is distinct from hoisting the dependencies.

src/lib/markbind/package.json
```diff
@@ -27,6 +27,7 @@
     "markdown-it-emoji": "^1.4.0",
     "markdown-it-imsize": "^2.0.1",
     "markdown-it-ins": "^2.0.0",
+    "markdown-it-linkify-images": "^1.1.1",
```
acjh, line 30: Shall we adjust the order of the sections in both package.json in this PR?
Or will it significantly mess up the diff since we're modifying packages too?

Ref: https://github.com/MarkBind/markbind/pull/1248#discussion_r439851727

src/Page.js
```diff
@@ -4,20 +4,21 @@ const fs = require('fs-extra-promise');
...
 
 const _ = {};
 _.isString = require('lodash/isString');
 _.isObject = require('lodash/isObject');
 _.isArray = require('lodash/isArray');
 
-const CyclicReferenceError = require('./lib/markbind/src/handlers/cyclicReferenceError.js');
-const { ensurePosix } = require('./lib/markbind/src/utils');
-const FsUtil = require('./util/fsUtil');
+const CyclicReferenceError = require('markbind/src/handlers/cyclicReferenceError.js');
```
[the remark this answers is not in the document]
acjh: Also in componentPreprocessor.js.

src/lib/markbind/package.json
ang-zeyu: Hmm perhaps the one that would move `src/lib/markbind` into `packages/markbind`? Would be more ~suitable~ relevant I think
The diffs wouldn't be too big if we do it here either though

acjh: I'm ok with either, if it doesn't significantly mess up the diff in this PR.

ang-zeyu: 👍 I'll give it a try, will stash it if it's too much

src/Site.js
ang-zeyu: > But if it's clearer that we're doing the patch

🤦‍♂️ 🤦‍♂️ I'll leave it in then!

src/lib/markbind/package.json
ang-zeyu: The diffs seem ok (Fixup hoist markbind core package dependencies commit); I extracted out the `jest` config into a separate file for consistency as well

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  testPathIgnorePatterns: [
+    '/frontend/',
+    '/node_modules/',
+    'test/functional',
```
acjh, line 5: Since we're modifying this line..

```suggestion
    '/test/functional/',
```

src/lib/markbind/package.json
```diff
@@ -1,17 +1,21 @@
...
     "reusable"
   ],
-  "author": "Jiang Sheng <gisonrg@gmail.com>",
+  "homepage": "https://markbind.org",
   "license": "MIT",
+  "main": "src/parser.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MarkBind/markbind.git"
+  },
```
acjh, line 18: Let's add:

```suggestion
    "directory": "src/lib/markbind"
  },
```

src/lib/markbind/package.json
acjh: https://github.com/MarkBind/markbind/pull/1253/commits/4da36b4f85ab37ac0439cf0122cdb394673ddd3a should be its own commit instead of a fixup.
Otherwise it messes up the diff since it's no longer obvious that we removed several packages.

ang-zeyu:
> 4da36b4 should be its own commit instead of a fixup.

Done, the new order:

Untitled

ang-zeyu:
Thanks again for the review @acjh! 😄

Will be dealing with the `packages/markbind` next, on that note, should we:
1. move `frontend/components` into `packages/components` as well?
2. improve `markbind/markbind-cli` separation by moving `assets` into the markbind core package `packages/markbind`

acjh:
> Will be dealing with the `packages/markbind` next, on that note, should we:
> 
> 1. move `frontend/components` into `packages/components` as well?

Sounds good.

Alternatives for the record:
- `src/lib/markbind` → `packages/markbind` or `packages/core`
- `frontend/components` → `packages/components` or `packages/vue-components`

> 2. improve `markbind/markbind-cli` separation by moving `assets` into the markbind core package `packages/markbind`
> 
> an alternative to (2) being to create another package for `assets`, but it seems a little awkward having a package just for that

Looks like `asset/` is only used by Site.js, which is currently in `markbind-cli`?

ang-zeyu:
> Alternatives for the record:

I'll go with the latter options, they look more semantic 😁

> Looks like asset/ is only used by Site.js, which is currently in markbind-cli?

yup, with the eventual goal being to move `Site/Page.js` + all other non-cli functionality into `packages/markbind` or `packages/core`. We could deal with this in one go in a later PR though, suggesting it now since we're touching on "frontend"

acjh:
> > Looks like `asset/` is only used by Site.js, which is currently in `markbind-cli`?
> 
> yup, with the eventual goal being to move `Site/Page.js` + all other non-cli functionality into `packages/markbind` or `packages/core`. We could deal with this in one go in a later PR though, suggesting it now since we're touching on "frontend"

Let's do it in a later PR.

PR #1254, opened 2020-06-15
Fix padding and font weight for custom list items

### LINKED ISSUE #1249: SiteNav: formatting difference for linked and unlinked items
Version: current master

image

See above. One item is linked to a page while the other is not (it serves as a section heading). Can they be formatted the same way? At least indentation and font weight?

ang-zeyu:

• [x] Bug fix

Fixes #1249

Simply adds the custom styles used for default list items to custom list items, doing some class renaming to better differentiate the two.

**Testing instructions:**
custom list classes should now get the font-weight and padding styling, but without the highlight-on-hover

---

Code:
Untitled

---

Output:
Untitled

---

Fix padding and font weight for custom list items

Padding and font weight formatting are not applied for custom list
items without nested lists or anchors.

Let's add these, and rename the classes used for better naming
consistency.

### DIFF
--- a/asset/css/site-nav.css
+++ b/asset/css/site-nav.css
@@ -30,7 +30,7 @@
     margin: 0 -12px;
 }
 
-.site-nav-list-item-content {
+.site-nav-default-list-item {
     display: flex;
     padding: 0.5rem 0 0 2.8rem;
     transition: background-color 0.08s;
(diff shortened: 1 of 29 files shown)


ang-zeyu:
clean rebase above

@damithc I left the right padding as is, you could add it using the bootstrap classes easily though `{ .p-1/2/3/4/5 }`, or overwrite the `site-nav-custom-list-item` class, sounds good?

damithc:
> @damithc I left the right padding as is, you could add it using the bootstrap classes easily though `{ .p-1/2/3/4/5 }`, or overwrite the `site-nav-custom-list-item` class, sounds good?

Yes. Right-padding is not that important.

ang-zeyu:
> Yes. Right-padding is not that important.

ok i'll merge this in then, since it hasn't been released and just shifts some styles around

PR #1263, opened 2020-06-28
Fix nunjucks path resolving

### LINKED ISSUE #931: Cannot use nunjucks in a sub-site

ang-zeyu:

• [x] Bug fix

Fixes #931 

To be filled

**Testing instructions:**
- npm run test should pass
- nunjucks paths should now resolve from the closest parent (sub)site ( tests added )

Fix nunjucks path resolving

Nunjucks' features that use paths resolve from the specified template
directory by design.
In MarkBind's case, the template directories are the respective root
paths of the various sites and subsites of the project.

Hence, a separate nunjucks environment with the template directory
configured as such is needed for each site or subsite.

Let's introduce a wrapper abstraction for this, VariableRenderer, to
wrap over such a singular nunjucks environment.
The VariableRenderer instances will be managed by VariableProcessor
(previously VariablePreprocessor), which already has a framework set up
to render variables belonging to each site or subsite.

In addition, {% raw %} tags inside a layouts file do not work due to
two nunjucks passes being made.
Let's revert the removal of the helper functions to keep {% raw %} tags
in the first pass to fix this, while searching for a solution that
would allow a single pass for layouts.


damithc:
This will be a big step forward if we manage to pull it off. 👍 
I believe sub-site support is one of our unique features and this was a vital missing piece in that feature.

ang-zeyu:
It's pretty much ready, just need to update some more test files. I'm putting it on hold as we are doing quite a bit of directory restructuring as suggested by @acjh here https://github.com/MarkBind/markbind/pull/1253#discussion_r446224358. Would be easier to review then.

> I believe sub-site support is one of our unique features and this was a vital missing piece in that feature.

The next step would be layouts such as header/footer/expressive layouts and such, I'm not sure if the current behaviour is intended though (layout files from the root site can currently be used in subsites, unlike variables)
And then we have other things like respecting configuration of individual `site.json`s (not sure if the current behaviour is intended as well though).

All of this should be able to be easily achieved by building the sites individually (the work so far still being necessary even if so though), and "blocking parent sites from building sub site pages".

we should probably flesh out the requirements for sub sites first in an issue or document the current behaviour in ug (if intended)

ang-zeyu:
> This looks fine to me! Just to check - I assume this has been tested on sites with nunjucks (2103T maybe?)? I don't see anything that would change existing working behaviour, but just want to be sure 🙂

yup!

> Just a suggestion regarding structuring commits - I think it would be nice to have renaming `variablePreprocessor` -> `variableProcessor` to be in its own commit. This way the reviewer can filter the commits to focus only on the material changes.

That makes sense, my bad 😅 I reseparated it - although there's some renaming in the first commit as well. This should be tremendously easier to look at in the history still though.

I also moved the variable rendering portion in `includeFile` out of it so there's no need to pass the 2 intermediate parameters (`additionalVariables` and `keepPercentRaw`) to `includeFile` which shouldn't know about it.

Also more clearly separates the variable rendering from the "including" part of the process. (retested with 2103 site again too)

packages/core/src/Page/index.js
```diff
@@ -590,16 +589,20 @@ class Page {
     // Set expressive layout file as an includedFile
     this.includedFiles.add(layoutPagePath);
     return fs.readFileAsync(layoutPagePath, 'utf8')
-      // Include file but with altered cwf (the layout page)
-      // Also render MAIN_CONTENT_BODY back to itself
+      /*
+       Render {{ MAIN_CONTENT_BODY }} and {% raw/endraw %} back to itself first,
+       which is then dealt with in the call below to {@link renderSiteVariables}.
+       */
+      .then(result => this.variableProcessor.renderPage(layoutPagePath, result, {
```
ang-zeyu, line 596: variable rendering is moved out of `includeFile`, also in `page.generate() / resolveDependency()`

packages/core/src/variables/VariableProcessor.js
```diff
@@ -396,8 +415,9 @@ class VariablePreprocessor {
    * @param content to render
    * @param highestPriorityVariables to render with the highest priority if any.
    *        This is currently only used for the MAIN_CONTENT_BODY in layouts.
+   * @param keepPercentRaw whether to reoutput {% raw/endraw %} tags, also used only for layouts.
```
marvinchin, line 418: Since we have some specific behaviour for layouts, do you think it's worth having a `renderLayout` function?

ang-zeyu: that makes sense! I think there's quite a bit of opportunity for refactor regarding layouts though, apart from this. (site nav, page layouts, header, footer takes up a sizeable portion of `Page.js`) Perhaps we could defer it to that pr?

marvinchin: That sounds fine too, let's fix that in the next PR then 🚀

Issue #987, opened 2020-01-18
<box>: support seamless style


damithc:
Suggestion: we can support this style for boxes (inspired by asciidoc), (perhaps using an attribute `seamless`?)

![image](https://user-images.githubusercontent.com/1673303/72659101-7a4f2780-39f5-11ea-8a68-cc1df25df1d8.png)

alyip98:
Can I clarify that a seamless box is similar to a normal box, just without background color?

damithc:
> Can I clarify that a seamless box is similar to a normal box, just without background color and border?

Yes, but further tweaks might be needed to make it look nice. We can make it similar to the asciidoc example given.

le0tan:
![image](https://user-images.githubusercontent.com/28982486/73235867-1da8f500-41cb-11ea-94e5-9f4e136d5cc7.png)

This is the prototype that I make, is this what's expected?

damithc:
> This is the prototype that I make, is this what's expected?

* icon can be bigger
* icon can follow the type color (green for type="tip" etc.)
* divider line can follow the type color (or a variant of it)

le0tan:
![image](https://user-images.githubusercontent.com/28982486/73334169-f58ec400-42a6-11ea-8bee-6ef13f8f7560.png)

I made some requested changes. However the color of the vertical divider is a bit tricky to inherit from the icon color as it's not a font-color property anymore.

My attempt is shown below:
```css
td.content {
    padding-left: 1em;
    border-left: 3px solid #dddddd; <---change the color here
}
```

damithc:
Perhaps use the bright colors used by the `light` style?
The divider can remain grey if hard to change.
Should the icon align to the top of the box, like this?
![image](https://user-images.githubusercontent.com/1673303/73334562-0d1a7c80-42a8-11ea-897d-bb88bc0752c8.png)

acjh:
For icon size, please leave it to https://github.com/MarkBind/vue-strap/pull/124.

le0tan:
![image](https://user-images.githubusercontent.com/28982486/73335191-9da58c80-42a9-11ea-96ed-3edd8983956e.png)

I made the requested changes.

le0tan:
> For icon size, please leave it to [MarkBind/vue-strap#124](https://github.com/MarkBind/vue-strap/pull/124).

[...]

PR #1268, opened 2020-07-04
Tweak box styles and revamp box code

### LINKED ISSUE #538: <box> icon: document support for font icons
Current: `<box>` allows specifying an emoji as the icon
image

Suggested: allow font icons i.e., fontawesome or glyphicon to be used as the box icon too, possibly allowing custom color for the icon as well.

Perhaps this can be achieved by allowing the icon to be specified using a `slot` element?

### LINKED ISSUE #1233: Seamless boxes: use black for text color
Received some [user feedback](https://github.com/reposense/RepoSense/pull/1281#issuecomment-640008391) that colored text in `seamless` boxes are hard on the eyes. May be use black text, similar to `light` boxes?

### LINKED ISSUE #1234: Box: do not apply the color property to the icon
Current: the `color` property of boxes apply to the icon as well as the text.

Suggestion: apply it to the text only (or give a way to adjust the two separately).

Reason: Icon color goes with the type of the box while text color may need to be adjusted for readability reasons. e.g., bright red for the icon to catch the eye but black for the text for easier readability.

ang-zeyu:

• [x] Documentation update
• [x] Bug fix
• [x] Other, please explain: code ~refactor~ revamp

Fixes #538 
Fixes #1233 
Fixes #1234 

This started as a code refactor to remove cascaded styling for our numerous box styles which introduces a lot of coupling / redundant classes (see commit message).
In doing so, it was almost entirely rewritten, so I tackled some other minor issues as well.

Refer to commit message for details:
- revamp box code
- tweak box styling
- enable linting for it
- give the docs some more style variations

I suggest simply going to the past code and observing the coupling between the cascaded classes (some even redundant), and checking out the new one to see if it is satisfactory since it's almost a rewrite

need an opinion on the 2nd (`primary` coloring for `definition` boxes) and 4th (https://github.com/MarkBind/markbind/issues/987#issuecomment-579621614 - lighter colors for seamless boxes) style tweaks, not sure if it was an intentional design choice @damithc 

**Testing instructions:**
- everything in the docs (added more variations as well) displays as is before, save for the styling tweaks mentioned below

```
Tweak box styles and revamp box code

The box component has numerous possible styles which relies heavily on
highly cascaded classes' styles.
This introduces a high degree of coupling between the classes. Also, as
a result of the cascading, some classes are rendered but completely
overwritten by others, making them redundant.

Let's revamp the box template and code to fix this.
We centralize the dynamic styles of each element into computed
properties, and extract helper methods to make these computed
properties as lean as possible.

Also, the component was already heavily rewritten prior to this. Hence,
let's remove it from the eslint ignore list to achieve consistent code
styling.

While revamping the code, let's tweak the box styles a little as well:
- Add some more padding for the header wrapper to better accomodate
markdown taking up a lot of space.
- Use the bootstrap 'primary' coloring for definition type boxes,
providing better visual diversity.
- Separate the applying of colors to the content wrapper and icon
wrapper. In so doing, the color attribute applies only to the content
(#1234), and seamless boxes' content correctly inherit color (#1233).
- Use the lighter `text-xx` bootstrap classes for seamless boxes'
coloring instead which are more suitable against a common white
background than the overly dark `alert-xx` classes.
```

### DIFF
--- a/docs/userGuide/syntax/boxes.mbdf
+++ b/docs/userGuide/syntax/boxes.mbdf
@@ -1,6 +1,6 @@
 ## Boxes
 
-**Box comes with different built-in types.**
+**Boxes come with different built-in types.**
 
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">html</variable>
@@ -33,10 +33,12 @@
     dismissible info
 </box>
 <box type="success" header="#### Header :rocket:" icon-size="2x">
-    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
-    <box type="warning" header="You can use **markdown** here! :pizza:" dismissible>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
-    </box>
+
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+
+  <box type="warning" header="You can use **markdown** here! :pizza:" dismissible>
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+  </box>
 </box>
 </variable>
 </include>
@@ -46,6 +48,9 @@
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">html</variable>
 <variable name="code">
+<box light>
+    default light
+</box>
 <box type="info" light>
     info light
 </box>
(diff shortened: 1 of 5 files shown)


damithc:
Good work! I'm generally OK with the changes, but my feedback is only from the user POV. Some points to consider.

* This icon position seems to be off. Shouldn't it be aligned to the top?
  image
* This leaves out the ability to change the color of the icon. If cannot be done here, may be create a separate issue?

ang-zeyu:
> This icon position seems to be off. Shouldn't it be aligned to the top?

Oops, mistakenly thought it was supposed to be in the middle. Will fix

> This leaves out the ability to change the color of the icon. If cannot be done here, may be create a separate issue?

👍 I'll create an issue after this is in, since it's a feature addition

ang-zeyu:
rebased first two commits on latest and swapped all computed properties for methods instead

docs/userGuide/syntax/boxes.mbdf
```diff
@@ -1,6 +1,6 @@
 ## Boxes
 
-**Box comes with different built-in types.**
+**Boxes comes with different built-in types.**
```
marvinchin, line 3:
```suggestion
**Boxes come with different built-in types.**
```

ang-zeyu:
updated, squashed and merged :) thanks! @marvinchin @damithc

damithc:
@ang-zeyu I'm using the latest master (pulled the master and ran `npm link`) and yet I get this. I thought the definition boxes should look different?

image

It could be I'm still running the old version but not sure how to verify.

ang-zeyu:
@damithc we are still sticking with updating our components bundle at release to avoid large diffs in prs and get a more accurate line diff count. You could do `npm run build:components` (in the same directory) after `npm link` to get the latest updates though

damithc:
Yup, that worked. Should we mention it in this page? https://markbind-master.netlify.app/devguide/settingup

ang-zeyu:
> Yup, that worked. Should we mention it in this page? https://markbind-master.netlify.app/devguide/settingup

the frontend dev flow will be reworked in an upcoming pr for hot reload, for now it's as if we never merged our `vue-strap` fork into here

PR #1272, opened 2020-07-10
Move lots of things to the core package

ang-zeyu:

• [x] Other, please explain: package content restructure

devGuide:
```
The command-line interface (CLI) application, which accepts commands from users and then uses the core library to parse and generate web pages, resides in the root.
```

let's move the `markbind-cli` package content to `@markbind/core` to further conform to this

All that can be moved has been moved to `@markbind/core` has been moved.

Functional tests aren't moved since they are more of e2e tests concerning the commands in `markbind-cli`

In each commit, what is stated in the commit message is first moved, then:
- it's dependencies are also moved (the `require(...)`'s)
- dependencies as stated by the `markbind-cli` and `@markbind/core` `package.json`s are removed and added accordingly
- any other necessary modifications (notably file paths) to ensure nothing breaks at any particular commit

- `Progress` (progress bar) dosen't seem like something that should be moved to `@markbind/core` considering it exists for `cli`. That said it shouldn't break `@markbind/core` as a standalone package, if it makes sense we can extract it out again in the future. (0c07b86)
- this PR can be broken up at any point for easier reviewing 😬 (**RFC**), some changes in earlier commits are overwritten by later ones though

Move non-cli functionalities to @markbind/core

The cli package houses numerous key functionalities in site generation,
such as the Site and Page models.

Let's move all non-cli related functionality to the @markbind/core
package instead.
This encourages `markbind-cli` to focus solely on the `cli` aspect of
the application.


ang-zeyu:
@acjh do you have time to review this? It's really just tedious work here, I don't think there's much for discussion. (if anything maybe just the dev guide update commit)
If not I'll merge this in a while

Issue #1078, opened 2020-03-02
Allow vue named slots shorthand and start deprecating old syntax


ang-zeyu:
**Describe the solution you'd like**

Vue has a shorthand for named slots `v-slot:header`/`slot="header"` -> `#nameofslot`.

However, this is not usable for certain components where markbind does extra processing on the slots currently. ( such as indexing the `header` slot of panels )

We could add logic to these locations to also search for the corresponding name slot shorthand, allowing the user to use a friendlier syntax.

At the same time, we could start deprecating the old syntax so that we can eventually remove all logic in markbind that searches for `slot="header"`, and rewrite corresponding vue templates to either `v-slot:slotname` or `#slotname` especially since it is a deprecated syntax in vue itself.

Warnings could also be shown if the old syntax is in use ( #1048 ).

**Describe alternatives you've considered**
na

**Additional context**
https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand

https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax

le0tan:
I will take a look into this~

Tejas2805:
@le0tan Are you still working on this? Or can I try this issue out?

wxwxwxwx9:
Let me take a look at this :-)

wxwxwxwx9:
As discussed in #1483, using `#example` breaks the CommonMark spec. Thus, if we want to implement a shorthand slot syntax for users, we have to look for an alternative syntax. 

However, there are not many other shorthand syntax prefix options available. The only ones that I found to be usable are:
- `_example`
- `:example`

I tested the possible syntax using [CommonMark's demo.](https://spec.commonmark.org/dingus/?text=%3Ca%20%23ref%3E%0A). 

@ang-zeyu also suggested something like `s-example`. 

@damithc would you have any recommendation for a suitable syntax?

damithc:
> @damithc would you have any recommendation for a suitable syntax?

No strong views. We already have `var-foo` syntax for specifying variable values for `<include>` https://markbind.org/userGuide/reusingContents.html#specifying-variables-in-an-include.  So, the `s-example` has some level of consistency although `slot-example` is more consistent with the `var-foo` syntax.

wxwxwxwx9:
@damithc thanks for your input! If we were to use `slot-example`, I think there would not be much benefits in terms of "shorter syntax". But it does provide the benefit of a more consistent syntax than `slot="example"` I suppose. 

[...]

PR #1274, opened 2020-07-13
Revamp question component

### LINKED ISSUE #1184: Support various types of question components

Question components currently only have one design 

```
wrapper
  question text
  hint panel
    hint
  answer panel
    answer
```

- panels are often used for large sections of content, which affects the visual distinctiveness of questions - at the very least we could have a separate type of panel designed for questions
- content like mcq questions are not integrated, and have to be manually inserted, there is also no 'answer' checking mechanism for this

As a static site, the aim isn't to replace fully fledged learning platforms, but at the very least, we could make questions more interactive and visually distinct:

- [ ] Redesign hint / answer to be more distinct ( e.g. clicking a button to make a hint popup )
- [ ] Integrate mcq ( both radio / multiple checkboxes style ), and mcq answer checking
- [ ] Provide a well designed and easily inserted `<textarea>` for open ended questions ( perhaps answer checking could be done here as well by looking for simple "keywords" in the answer )
- [ ] ❔ 

na

**Additional context**

na

ang-zeyu:

• [x] New feature

Resolves #1184

Total redesign / rewrite of question components, while preserving backward compatibility (though somewhat awkwardly since questions are always marked as 'correct').

External libraries: Suprisingly there aren't any ready made vue quiz components that suits our requirements (though an abundance of tutorials), also backwards syntax compatibility is a must

Added various `<question type="">`s:
- Mcq
- Checkbox
- Text (or old components with `hasInput`)

Added `<quiz>`, which strings multiple `<question>`s together

_to be done: (WIP)_
- polish usage / design
- snapshot tests
- investigate updating vue bundle to 2.6 in a separate PR before this one (test files `vue.min.js` aren't updated here as well so ci fails for now)
- add deprecation warnings in another PR after this one

- any suggestions / comments on design / usage / docs / bugs
- would avoid looking at the implementation since it's a huge rewrite, but if you have the time any comments are welcome ^^

**Testing instructions:**
- `npm run build:components` on this branch then `markbind s` to play around with it
- netlify preview https://deploy-preview-1274--markbind-master.netlify.app/userguide/usingcomponents#questions-and-quizzes

Revamp question component

Question components serve as a thin syntax wrapper on top of panel
components, and don't provide any form of answer checking.

Let's rewrite and redesign the question component, using bootstrap
cards as the design basis.
Let's also provide 3 types of questions - mcq, checkbox and text, which
allow for simple answer checking.

Educational websites may also want to build mini quizzes consisting of
multiple questions.
Let's provide the quiz component for this, which can string multiple
questions together with a simple scoring mechanism.


ang-zeyu:
could use some opinions here 😁 @damithc @acjh @marvinchin 
https://deploy-preview-1274--markbind-master.netlify.app/userguide/usingcomponents#questions-and-quizzes

damithc:
> could use some opinions here 😁 @damithc @acjh @marvinchin
> https://deploy-preview-1274--markbind-master.netlify.app/userguide/usingcomponents#questions-and-quizzes

Looks nice. The questions feature has value in terms of nicer presentation. It can have even more value if we separate _check_ from _show answer_ so that the reader can try multiple times. Can add a reset button too.

I'm not sure if the quiz feature has enough value to make it worth the cost. One can always list the questions one after the other to form a quiz. Adding up the correct answers doesn't add a lot of value unless the quizzes are long.

Side note: In my current use of MarkBind I use LumiNUS for all my quiz needs. The questions I currently have in my MarkBind sites were added before I moved to using LumiNUS for quizzes. Hence, my low enthusiasm for this feature (wouldn't want to spend resources on a feature I'm not going to use much).

Perhaps we strengthen the question feature but put the quiz feature on hold for now?

ang-zeyu:
> I'm not sure if the quiz feature has enough value to make it worth the cost. One can always list the questions one after the other to form a quiz. Adding up the correct answers doesn't add a lot of value unless the quizzes are long.

Cost in terms of maintainability? I separated how a `<question>` functions with / without a `<quiz>` wrapping it into an easily unpluggable file actually (`QuizQuestionMixin.js`), it should be pretty decoupled. I may be a little biased here though, would be nice to have a few more eyes on it

In terms of client side rendering performance it would actually improve it as opposed to listing the questions one after the other, as the dom and virtual dom for unseen questions is not rendered.

> It can have even more value if we separate _check_ from _show answer_ so that the reader can try multiple times. Can add a reset button too.

These functions are provided in the quiz already, I avoided putting in `<question>`s as well as:
- it encourages spamming / brute forcing the answer especially for mcq / checkbox questions, which defeats the purpose
  - using a series of questions in a `<quiz>` in contrast reinforces the learning points better as the reader is forced to remember the answers to each of the questions when clicking `retry`
- for `text` questions it can easily lead to frustration when the user keeps retrying due to spelling errors / using synonyms, which again encourages simply jumping to `show` instead of `check`
- it should be left for more gamified learning platforms (not luminus), where each question poses considerable difficulty

This is just imo though, it's highly subjective to individual learning styles. I'm not opposed to having it as an opt-in thing like `<question has-check>` or the reverse either

> Side note: In my current use of MarkBind I use LumiNUS for all my quiz needs. The questions I currently have in my MarkBind sites were added before I moved to using LumiNUS for quizzes. Hence, my low enthusiasm for this feature (wouldn't want to spend resources on a feature I'm not going to use much).
> 
> Perhaps we strengthen the question feature but put the quiz feature on hold for now?

Yup the aim isn't to replace fully fledged learning platforms with a backend, but simply to provide a more interactive experience, as well as
1. a more performance friendly way to display a whole bunch of questions
2. a simple scoring mechanism to encourage retrying the quiz to reinforce the learning points

If it helps, my intended use case for the quiz component was somewhat of the following:
- `OOP.md`
  - quiz on oop at the very bottom
- `Gathering Requirements.md`
  - quiz on methods of gathering requirements

damithc:
> Cost in terms of maintainability?

Yes, plus the effort needed to go from here to production., which could be a lot.

> This is just imo though, it's highly subjective to individual learning styles. I'm not opposed to having it as an opt-in thing like `<question has-check>` or the reverse either

As this is self-study mode without any oversight from instructor, we have to assume the reader will choose whatever works for her. We simply provide all the possibilities. The reader choose correctly, we reveal everything. If the answer is wrong, the reader can try again or give up and ask for the answer. I don't think there is any downside of giving that choice to the reader.

ang-zeyu:
> Yes, plus the effort needed to go from here to production., which could be a lot.

Just need some snapshot tests and it should be good to go. Also any other minor design / feature tweaks and additions.
Code wise, again, it should be pretty decoupled. Rather, the below feature request likely adds even more coupling between `<q-option>/<question>` than `<quiz>/<question>`

> I don't think there is any downside of giving that choice to the reader.

Presentation wise there may be value in having a clean ui, especially if the page is already heavy

Also for mcq questions, you very quickly exhaust all the possible options (even if you're not intending to brute force out the answer) by just clicking `check`. Having `show` just feels like feature bloat / duplication to me. 

For checkbox questions, I feel that it quickly becomes a tiresome guessing game if we don't show which option(s) were correct / wrong, since there are 2^n possible answers. If we show which option(s) were correct / wrong, it leads to inconsistent behaviour with mcq questions, and also it is pointless - because unchecked answers can be correct we need to show whether all options were correct / wrong. (`check` would just become `show`)

How about just having it for `text` questions? It may be difficult to get the correct keyword sometimes.

How it looks from some in browser edits:

image

damithc:
Consider this analogy:

Scenario 1:
1. Teacher asks a question.
2. Student gives an answer.
3. Teacher says that answer is incorrect. Possibly also explain why it's incorrect.
4. Student can either try again, or say "I give up, give me the right answer"

Scenario 2:
1. Teacher asks a question.
2. Student gives an answer.
3. Teacher says if the answer is correct or wrong and also gives the right answer.

The value of this process comes from the student _thinking_ about the question and trying to find the answer. This is why teachers don't give answers right away and instead try to guide the student towards the answer by giving more and more hints. The answer itself is less important than the attempts to get at the answer. The more attempts the better. Of course this doesn't work if the student is randomly answering without thinking about it. In that case might as well give away the answer.

Consider these buttons:
* `check answer`: only reveals if the answer is correct. For MCQ, can also reveal the explanation of the currently chosen option as an additional information.
* `show answer`: reveals all info. If an answer is given, also shows if the answer is correct or wrong.

The reader can choose scenario 1 or 2 by clicking on one of the two buttons. This is something the old version cannot do. 

Having said that, I'm fine with the current version too as it behaves the same way as existing questions and look nicer. Try not to break the legacy syntax because I might not have time to migrate to the new syntax right away.

ang-zeyu:
> The value of this process comes from the student thinking about the question and trying to find the answer.

On second thought, that probably has more priority than trying to make the behaviour consistent from an author standpoint, or minor feature duplication. To reconfirm what you meant (the `1`'s):
a1. Mcq questions - `check` (also equals `retry`) - explanation for the selected answer, correct or wrong, is displayed, and a `show` button

or

a2. Mcq questions - `check` (also equals `retry`) - explanation for the selected answer, correct or wrong, is displayed, _no_ `show` to avoid minor feature bloat

b1. Checkbox questions - `check` (also equals `retry`) - no explanations are displayed at all unless all options are correct, this is necessary since unchecked answers can be correct, and `show` 

or 

b2. Checkbox questions - `check` (also equals `retry`) - explanations for checked answers (correct or wrong) are displayed, and `show` 

c. Text questions - `check` doesn't display anything, `show` does


When `<question>`s are wrapped by `<quiz>`s however, the current netlify behaviour is followed, except `check` is renamed to `show` for consistency:
a. Mcq questions - only a `show`, however all explanations are immediately displayed
b. Checkbox questions - only a `show`, however all explanations are immediately displayed as well
c. Text questions - only a `show` here

Is there a better naming for `show` and `check` to reflect the purpose better as well? (since `show` also 'checks' the answer)
`show` -> `check`
`check` -> `try` / `??`

> Having said that, I'm fine with the current version too as it behaves the same way as existing questions and look nicer. Try not to break the legacy syntax because I might not have time to migrate to the new syntax right away.

The only weird part currently would be the addition of the `check` button which always ends up being correct. I'll remove it for `<question>`s without types or text questions without keywords (`hasInput`), should be fully compatible otherwise.

On syntax, would it be better to just have `<template #answer>` for consistency? Although it's somewhat of a misnomer for mcq / checkbox questions.

damithc:
Haven't looked at the syntax yet. Let's nail down the behavior first.

a1, to allow the reader to give up up after checking
b1. no need to show explanations

text questions can have keywords that should be in the answer and a separate explanation, which means it too can have both buttons.

What do you think of this alternative?

Only a `check` button is present. If the answer is correct, everything is revealed. If the answer is wrong, a dialog prompts the user with `retry` and `show answer` options.

ang-zeyu:
> text questions can have keywords that should be in the answer and a separate explanation, which means it too can have both buttons.

yup

> What do you think of this alternative?

good idea, this would make the initial UI more consistent with questions in quizzes as well.
by dialog, do you mean a popup dialog? how about just transitioning the buttons out for less disruption? (the current transition between `check` and `next` in quizzes)

I'm ok with `retry`. `show answer` seems a little long for a button though, should we just stick with `show`?

extrabutton

damithc:
> by dialog, do you mean a popup dialog? how about just transitioning the buttons out for less disruption? (the current transition between `check` and `next` in quizzes)

Yes, button transition better.


> I'm ok with `retry`. `show answer` seems a little long for a button though, should we just stick with `show`?

Sure.

damithc:
On a separate matter, is it smoother to make the hint appear just above the button rather than far away from the button? Less travelling for the eye.

ang-zeyu:
> On a separate matter, is it smoother to make the hint appear just above the button rather than far away from the button? Less travelling for the eye.

like this? works too. 

extrabutton

damithc:
> like this? works too.

Yes, let's do that then.

damithc:
Regarding the syntax, possible to avoid introducing a `#` to the syntax? It's not used anywhere else in MarkBind.

ang-zeyu:
the old `slot="..."` syntax works too (in the deprecation notes).

eventual deprecation of all `slot="xx"`s to a more succinct `#xx` is almost a must though (#1078), as vue 3.0 is dropping it as well (https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax)

If preferred I could use it in the examples for now still instead, to avoid inconsistency with the other components

damithc:
> the old `slot="..."` syntax works too (in the deprecation notes).
> 
> eventual deprecation of all `slot="xx"`s to a more succinct `#xx` is almost a must though (#1078), as vue 3.0 is dropping it as well (https://vuejs.org/v2/guide/components-slots.html#Deprecated-Syntax)
> 
> If preferred I could use it in the examples for now still instead, to avoid inconsistency with the other components

I see. I guess we'll use the slot syntax for now until we changeover to # syntax later.

ang-zeyu:
Made the changes, I also changed `<template slot="answers">` to `<template slot="body">` for all question types.

Placing `<q-option>` into any of these slots is entirely optional, it can be placed as a direct child of `<question>` as well.
In this sense, should we instead make `body` the default slot (ie. no slot) and `header` an optional slot?

The downside would be needing to manually format the question everytime, ie.,

```
<question type="checkbox">

Which of the following is true?

<q-option correct>
  20 + 20 = 2020
  <template slot="reason">
    lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum
  </template>
</q-option>
<q-option>
  1 / 0 = infinity
  <template slot="reason">
    Division by zero is <md>**undefined**!</md>
  </template>
</q-option>
<q-option correct>
11 / 11 = 1
</q-option>
<template slot="hint">
This is the question hint, you can use markdown here as well.
</template>
</question>
```

results in a really plain question text

Untitled

damithc:
> In this sense, should we instead make `body` the default slot (ie. no slot) and `header` an optional slot?
> 
> The downside would be needing to manually format the question everytime, ie.,

Yes, I think it's fine. It's actually better to let the author format the question in any way rather than force a specific format. Bolding the question by default is not appropriate in some cases e.g., when the questions is very long.

Pros and cons of using simpler tags `<option>`, `<reason>` instead of slots?

ang-zeyu:
> Pros and cons of using simpler tags `<option>`, `<reason>` instead of slots?

`<option>` is a html element already https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

For `<reason>`, I think it's inconsistent with our syntax in other components. How about adding:
- `reason` prop for `<q-option>`
- `header` prop for `<question>`s
- `hint` prop for `<question>`s
- `answer` prop for _text_ `<question>`s

all of which allows markdown like our other components?

damithc:
> For `<reason>`, I think it's inconsistent with our syntax in other components. How about adding:
> 
> * `reason` prop for `<q-option>`
> * `header` prop for `<question>`s
> * `hint` prop for `<question>`s
> * `answer` prop for _text_ `<question>`s
> 
> all of which allows markdown like our other components?

Do you mean these can be specified as either properties (supports inline markdown) or using the slot syntax (if block markdown is needed)? This this?
```
<q-option reason="that _works_">
  20 + 20 = 2020
</q-option>
```
```
<q-option>
  20 + 20 = 2020
  <template slot="reason">

#### reason heading
* sub-reason 1
* sub-reason 2

  </template>
</q-option>
```

ang-zeyu:
> Do you mean these can be specified as either properties (supports inline markdown) or using the slot syntax (if block markdown is needed)? Like this?

Yup much like most other components

damithc:
> Yup much like most other components

Sounds fair. In the documentation, give a simple example first (i.e., using props) and mention that slot syntax can be used too (with a smaller example). Otherwise the syntax can appear too verbose at first glance. Alternatively, use slot syntax only for one option of the example.

ang-zeyu:
Done, features and behaviour should be more or less locked down. Also added tests for transforming the mentioned attributes into slots.
I'll add snapshot tests this weekend as they take a while to inspect

In the meantime, any comments / suggestions on the docs are welcome 😄

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
-  </div>
-  <div slot="answer">
-    Question answer. <md>:pencil:</md>
-  </div>
+#### Introduction
+
+Question components (`<question>`) can be one of the following types: **Mcq**, **Checkbox** or **Text**.
+
+<div class="px-5">
+<question type="checkbox" header="Which of the following is true? (**Card Header**)" hint="This is the question hint, you can use **markdown** here as well.">
```
damithc, line 10: `Card` is a bootstrap term, not used in the MarkBind vocabulary (from author POV)?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
+</q-option>
 </question>
-</variable>
-</include>
+</div>
+
+
+In all cases, note that content directly inserted in between `<question>...</question>` will be inserted into the card's body.
 
+To output your markup to the header, or insert content into the hint tip box, simply insert your markup inside the `header` and `hint` attributes respectively:
```
damithc, line 29: `To output your markup to the header`: rephrase in a non-programmer friendly way?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
+  <template slot="hint">
+
+  This is the question hint, you can use **markdown** here as well.
+
+  Paragraph **2**
+
+  </template>
 </question>
 </variable>
 </include>
```
damithc, line 68: image

Perhaps something like `Expected keywords: [none provided]`? Otherwise it looks like the heading for the hint.

Also, we haven't explained text questions yet at this point.

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
```
damithc, line 69: It's harder but better if the example used is a real question that illustrates all the points.

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
+  <!-- Optionally, you may use a reason slot instead of a reason attribute. -->
+  <template slot="reason">
+  If you test all invalid test inputs together, you will not know if each one of the invalid inputs are handled
+  correctly by the SUT.
+  This is because most SUTs return an error message upon encountering the first invalid input.
+  </template>
+</q-option>
+<template slot="hint">
+How do you figure out which inputs are wrong? (or correct)
+</template>
```
damithc, line 114: image
Will this confuse the reader? This reminded me, do we want to support having multiple correct answers to an MCQ question.

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
+<question type="text" header="Which country did the Hawaiian pizza originate from?"
+          keywords="hawaii" fuzzy="1" threshold="0.5" answer="It originated from Hawaii!">
+  <template slot="hint">
+
+  Watch some pizza commercials!
+
+  :pizza: :pizza: :pizza: :pizza: :pizza: :pizza: :pizza:
+
+  </template>
+</question>
```
damithc, line 176: Any mention of whether _all_ keywords need to match?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,325 @@
...
+```
+
+```html { heading="Quiz" }
+<quiz>
+<question type="mcq">...</question>
+<question type="checkbox">...</question>
+<question type="text">...</question>
+</quiz>
+```
```
damithc, line 318: Questions used as examples earlier can be reused here via `<include>`?

I'm not sure the current slide-in animation is a good fit. In this page itself, it goes over the site-nav menu -- feels like the question is overstepping its play area.

docs/userGuide/syntax/questions.mbdf
ang-zeyu: dosen't this defeat the purpose of having an introduction section (not having touched on text / mcq / checkbox yet)? chucking it all into mcq and checkbox questions seems a little heavy

docs/userGuide/syntax/questions.mbdf
ang-zeyu: it's already supported, but undocumented.

Fixed the ticks too

docs/userGuide/syntax/questions.mbdf
ang-zeyu: added a mention, there's a `threshold` option to configure it as well

docs/userGuide/syntax/questions.mbdf
ang-zeyu: > I'm not sure the current slide-in animation is a good fit. In this page itself, it goes over the site-nav menu -- feels like the question is overstepping its play area.

fixed, it's a bug. It's supposed to only slide in from as far as the root container goes.

docs/userGuide/syntax/questions.mbdf
ang-zeyu: I forgot to mention the checking algorithm, currently it just does case normalization + a naive substring search, so if you have a keyword `xx`, `yxxy` would also match.

Not sure if it should be more restrictive:
1. splitting by whitespace + case normalization
2. remove leading + trailing punctuation only: `\[](){}&|\'\"<>#:;~_^=\\-‑+*/\\\‘’“”，。《》…—‐•\`
3. then checking each token against the keyword

So far I can't really think of a use case that has leading and trailing punctuations, but to be safe, we could also provide an `exact` prop that skips step 2.

What do you think?

docs/userGuide/syntax/questions.mbdf
ang-zeyu: I truncated the code though, and kept the `q-option`s in the output (since mcq/checkbox hasn't been explained yet), if that's what you meant

damithc: I meant the example should be a real-word example, not a meta example.
e.g., Instead of `Paragraph **2**` (which will never appear in a real-question), use an answer that actually has two real paragraphs. Coming up with such illustrative real-world examples are harder though. I think some of the examples given match that criterion, but some don't.

docs/userGuide/syntax/questions.mbdf
damithc: This is not easy to do -- not sure if we should even try. Validation of text questions is not going to be that useful for most cases. Given the risk of false negatives (as well as false positives), as an author I'm likely to just give the answer in the explanation and let the reader figure out. 
Something related: It can be misleading to mark answer as correct when no keywords are specified. The reader might think the answer was validated when it was not.

ang-zeyu: > I'm likely to just give the answer in the explanation and let the reader figure out.

I suspect this would be the majority use case as well, for questions that have long answers. Hence it's somewhat ok for the matching to be naive.

I was thinking more of mathematical / numerical questions with a single definite text answer, which the current algorithm may break (imagine a `keywords="1.99"` and the answer provided is `1.999` -- still correct by substring matching). Using the `1,2,3` above would fix this, but also makse the checking for long answers slightly stricter.

I'm not sure if we should advertise `text` questions as supporting mathematical / numerical questions as such, or defer it entirely to another pr and just use substring matching here 😮 

> It can be misleading to mark answer as correct when no keywords are specified. The reader might think the answer was validated when it was not.

fixed. it will still count toward the score in quizzes though.

docs/userGuide/syntax/questions.mbdf
ang-zeyu: ahh my bad. will try to find an interesting one 😏

docs/userGuide/syntax/questions.mbdf
damithc: > I suspect this would be the majority use case as well, for questions that have long answers. Hence it's somewhat ok for the matching to be naive.

Let's add a note in the documentation to explain for which kind of cases the validation works and recommend to use it only for such cases. We don't want to give false hopes.

ang-zeyu: Should we still at least perform tokenization + punctuation removal though? Since if that's the case, I think there's no harm in improving the algorithm to suit the majority of answers (which I think we can assume to be well formed) for text question validation.

ang-zeyu: @damithc

damithc: I'm not quite sure.  'Works for most cases' isn't very reassuring. We need to specify precisely the class of answers that works 100% and inform users to use the feature only for that class. We can't expect authors to experiment trial-and-error to find out whether to use the feature for each question.

For example, something like the following is safe:
> The validation works only if the expected answer is a precise phrase e.g., an equation (case insensitive, white space variations ignored)

Anything else that works 100%? If there is, we can add it to the above.

Supporting percentage matches is problematic because `not rewriting` and `rewriting` can both match the keyword `rewriting`.

ang-zeyu: > The validation works only if the expected answer is a precise phrase e.g., an equation (case insensitive, white space variations ignored)

this dosen't work actually, in either algorithm, it's not the intended case

> I'm not sure if we should advertise text questions as supporting mathematical / numerical questions as such, or defer it entirely to another pr and just use substring matching here 😮

I think we should defer those types of questions to another type of question, with a smaller input box.

I intend for the matching to be only correct 80-90% of the time. It's difficult by nature to match this 100% like you mentioned, and to reach 99% of matches we likely need some other complex language processing which is out of scope. There's also the `show` button as a fallback so readers can always view the answer.

I was wondering if a simple substring match that matches potentially more things but also risks more false positives (`xx` -> `yxxy`) would be better or a slightly more restrictive one with tokenization and punctuation removal as mentioned above as such would be better in achieving that 80-90%, but,
having just realised there's also the issue of non whitespace tokenized languages 😱, I think we could just stick with the former.

ang-zeyu: to further clarify I intend to add something like that to the docs:

> Warning: Keywords are validated by simply looking for the keyword as a pattern in the user's answer!
> This works well for some cases but not all. (and then a popover giving one good and one bad example)

damithc: Can go ahead.

ang-zeyu:
Added tests, and resolved the above documentation nits. @damithc anymore comments about the documentation?

I'll give the tests one more look tonight before merging this if not, then I think we can get started with the release 😄 @marvinchin

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
+<question ... header="Which of the following is correct? (**Header**)" hint="Think out of the box! :fas-box:">
+
+  <!-- Anything you place directly under a question not in a slot is inserted into the question body! -->
+  <small>Adapted from [Daily Mail](https://www.dailymail.co.uk/femail/article-4702868/Can-pass-intelligence-test.html)
+  </small>
+  </pic>
+
+  (**Question Body**)
+
+  <!-- Several hidden checkbox q-option components explained later -->
```
damithc, line 27: Let's remove these to keep the examples _pure_. They can be even be misinterpreted as part of the question display template precisely because a reader doesn't expect them to be added by the author.
image

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
-## Questions
+## Questions and Quizzes
 
-**Question component consists of a question body, a hint and an answer.**
+Question and quiz components provide an easy way to test readers on the relevant content topic in the page.
 
-<include src="codeAndOutput.md" boilerplate >
+#### Introduction
+
+Question components (`<question>`) can be one of the following types: **Mcq**, **Checkbox** or **Text**.
```
damithc, line 7: `Mcq` -> `MCQ`?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
-<question>
-  The question body. <md>:grey_question:</md>
-  <div slot="hint">
-    Question hint. <md>:crystal_ball:</md>
-  </div>
-  <div slot="answer">
-    Question answer. <md>:pencil:</md>
-  </div>
+<!-- Insert markdown into the header and hint using the respective attributes -->
+<question ... header="Which of the following is correct? (**Header**)" hint="Think out of the box! :fas-box:">
```
damithc, line 18: Should this be an mcq question instead?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
 </question>
 </variable>
 </include>
+</panel>
+
+<box type="tip" seamless class="mt-3">
 
-**Use the minimized Panel component to create and group questions inline.**
+Placing the question into the header is entirely optional. You may also wish to include the question directly in the question body, omitting the header entirely.
+</box>
```
damithc, line 138: So, is there any value in supporting a header attribute.

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
+
+
+****Common `<question>` Options and Slots****
+Name | Type | Default | Description
+--- | --- | --- | ---
+header | `String` | `''` | The markup to insert into the question header. The header is omitted if this is not provided.
+header | Slot | empty | Functions similar to the `header` option, and overrides it if both are present.
+hint  | `String` | `''` | The content to display in the hint box.
+hint  | Slot | empty | Functions similar to the `hint` option, and overrides it if both are present.
```
damithc, line 147: I think we should some other way to indicate which attributes support the alternative slot syntax, rather than add one row for each. e.g., use a header<sup>[s]</sup>

docs/userGuide/syntax/questions.mbdf
damithc: What's this `...` here?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
+</panel>
+
+<box type="tip" seamless class="mt-3">
 
-**Use the minimized Panel component to create and group questions inline.**
+Placing the question into the header is entirely optional. You may also wish to include the question directly in the question body, omitting the header entirely.
+</box>
+
+
+****Common `<question>` Options and Slots****
```
damithc, line 141: _Common_ here can be misinterpreted as _commonly used_ `Options common to all question types`?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
+    <template slot="reason">
+    If you test all invalid test inputs together, you will not know if each one of the invalid inputs are handled
+    correctly by the SUT.
+    This is because most SUTs return an error message upon encountering the first invalid input.
+    </template>
+  </q-option>
+  <template slot="hint">
+  How do you figure out which inputs are wrong? (or correct)
+  </template>
+</question></variable>
```
damithc, line 184: We don't use `template` anywhere else. Why not `<div slot="...">`?

docs/userGuide/syntax/questions.mbdf
```diff
@@ -1,130 +1,343 @@
...
+
+
+****Common `<question>` Options and Slots****
+Name | Type | Default | Description
+--- | --- | --- | ---
+header | `String` | `''` | The markup to insert into the question header. The header is omitted if this is not provided.
+header | Slot | empty | Functions similar to the `header` option, and overrides it if both are present.
+hint  | `String` | `''` | The content to display in the hint box.
+hint  | Slot | empty | Functions similar to the `hint` option, and overrides it if both are present.
+
```
damithc, line 148: Why `type` is not mentioned in this table?

docs/userGuide/syntax/questions.mbdf
damithc, line 18: Alternatively, the intro section can simply give an example of each question type without code?

ang-zeyu: I made this checkbox for the purposes of the later more elaborate example with slots (`Challenge: Try to get all the options correct in your first try!`).

changed `...` -> `type="checkbox"`

docs/userGuide/syntax/questions.mbdf
ang-zeyu: For formatting, it automatically bolds the text and has a more prominent background. (rather subtle in the default bootstrap theme but some other themes make it stand out more)

docs/userGuide/syntax/questions.mbdf
ang-zeyu: hmm, I was thinking we should actually introduce slots upfront in the components section and not all the way below in `advanced tips and tricks`, since they are widely used in the examples. (don't think we should forward reference in user guides). How about cleaning and standardising this in that PR?

docs/userGuide/syntax/questions.mbdf
ang-zeyu: updated, note that it introduces an additional wrapper `<div>` though.

perhaps we could build a shorthand syntax feature for `<template slot="xx">` -> `<t slot="xx>`? (noting that vue 3.0 already shortens `slot="xx"` to `#xx`

docs/userGuide/syntax/questions.mbdf
ang-zeyu: It feels a little repetitive as questions are pretty big

docs/userGuide/syntax/questions.mbdf
damithc: No need for forward reference. e.g., we can make it a modal/popover. Define in one place and `<include>` everywhere needed. No matter where the reader encounters it first, the definition is just a hover/click away.

docs/userGuide/syntax/questions.mbdf
damithc: Roger that.

docs/userGuide/syntax/questions.mbdf
damithc: In that question, the question itself can be rephrased to fit a check-box type. The current question is an equation that can only have one answer. Hence, the reader will wonder why we are using check boxes.

ang-zeyu: not sure I follow, isn't the snippet above the same question with two answers?:

Untitled

docs/userGuide/syntax/questions.mbdf
ang-zeyu: Sounds good. I'll add it in a pr after the one fixing the rest of the comments here

docs/userGuide/syntax/questions.mbdf
damithc: My bad. I didn't read the question closely. In fact, I read the last line only and thought it's a simple addition question for primary kids.

Issue #1285, opened 2020-07-26
v2.15.0 fails to install entirely


ang-zeyu:
**Tell us about your environment**

* **MarkBind Version:** the next (unpublished) release

**What actually happened? Please include the actual, raw output.**

![Untitled](https://user-images.githubusercontent.com/3306138/88471387-76b7d280-cf3b-11ea-8b60-a240c9443c89.png)

**What did you do? Please include the actual source code causing the issue.**

Suspected cause: we moved towards using relative file specifiers in `package.json` (`@markbind/core: "file:packages/core"`) since the last release. I've confirmed that subpackage `package.json`s are indeed published, however, when doing `npm install -g`, npm attempts to resolve the root `package.json`'s dependencies first without having downloaded the entire package (including the subpackage's `package.json`). This causes the above error.

**Attempted Solutions** (from posts and articles)
- `npm cache clean --force` + `npm cache verify` - dosen't work, and users shouldnt have to do this to upgrade.
- using npm [shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap), which is essentially a publishable `package-lock.json`. This causes a `maximum call stack size exceeded` error with a not very useful debug log either.

**Proposed solution**

We opted for the current structure (a root `markbind-cli` package with 2 subpackages) for reasons of simplicity for now. From searching up similar issues, the working solutions _all_ tended towards:
```
_private_ (unpublished) root package.json managed by lerna / yarn workspaces (simply a meta file, possibly containing dev dependencies, but not actually a package)
packages (all published independently)
  - package A
  - package B
  - package C
  - ...
```

which is also the structure many well known projects use (babel, react, gatsby, ...). 
The devtools often used here are lerna (+ yarn workspaces).

_Solution_: Let's move towards this more conventional structure, and publish each package independently and separately using lerna as a high level helper (yarn workspaces can come later if really needed).
This also paves the way for future packages https://github.com/MarkBind/markbind/pull/1248#issuecomment-643800784.

1. Move markbind-cli into `packages/cli` (but retain package naming for "backwards naming compatibility") (PR 1)
2. setup lerna, edit the devguide sections accordingly (PR 2)
3. publish

**Reservations**

It's a little strange for `package.json`s to allow relative `file:...` specifiers but have them break when published. I haven't been able to find much on this (most solutions just tended towards lerna + yarn) so far either.

[...]

PR #1291, opened 2020-07-27
Setup lerna

ang-zeyu:

• [x] Other, please explain:

Resolves https://github.com/MarkBind/markbind/issues/1285

See #1285

Commit order:
- Move all cli files to `packages/cli` 5a56a69
  - and update `.gitignore` ignore path accordingly
- Add`.npmignore` for core package d7ca8bc
- Setup lerna (gitignore update + `lerna init` + `lerna bootstrap`) 7979698
- Update font awesome test files (the previous step caused some package-lock versions to change, which caused font awesome to update 605a4e2
- Adapt test scripts f7c16e3
- Adapt linting configs, scripts c3c22bc
  - this commit also unifies the linting into one call to `eslint .`, and ditches separately running `eslint .` in all packages. This allows vue-components' eslint to properly extend the root config
- Move vue-components script to root private package 823e9ec
- Adapt CI and setup (install) scripts b0d33ca
- Update devguide bbe4bec (wip)
  - the publishing process is yet to be tested here

* bbe4bec 's publishing process is yet to be dummy tested

**Testing instructions:**
- no change -- `npm run test`
  - under the hood this now runs eslint & stylelint, then the `test` npm script in all packages separately

The cli and core packages uses relative file specifiers to link
dependencies in the mono repo. This causes a published cli package to
fail to install.

Besides site generation, the core package also houses several other
things like templates, assets and plugins, which can be moved into
separate packages for better separation of concerns.

Let's extract the main cli package into packages/cli as such, resulting
in a flat package structure.
The multi package structure implies publishing all packages
independently, which fixes the usage of relative file specifiers.

Let's also setup lerna, which is a multi-package monorepo development
and dependency management helper.
This greatly reduces the amount of work needed to manage multiple
independent packages.


.eslintignore
```diff
@@ -2,6 +2,33 @@
 node_modules
 packages/core/src/lib/markdown-it/*
 packages/core/src/lib/markdown-it-shared/*
-packages/vue-components
 
 !.eslintrc.js
+
+# vue-components
```
marvinchin, line 8: Just wondering - is there any reason why we don't just use `packages/vue-components` here?

docs/devGuide/settingUp.md
```diff
@@ -32,14 +32,16 @@ We recommend the **WebStorm IDE** for working with MarkBind code.
 ## Setting up the dev environment
 
 1. **Fork and clone** the MarkBind repo.
-1. **Install dependencies** by running `npm run ci:all`
-   <popover header="Why **2** `ci` commands?" content="Our components subpackage depends on `node-sass` indirectly, which depends on `fsevents`.<br><br> `fsevents` is not needed nor compatible with Windows, and `npm ci` does not respect this since it uses `package-lock.json`.">(`ciwin:all` if you are on a Windows machine)</popover>
-   in the root folder of your cloned repo.
-1. **To bind your cloned version of MarkBind to your console** (instead of the released version of MarkBind), run `npm link` in the root folder of the cloned repo.
+1. **Install dependencies** by running <popover content="Under the hood, this calls `npm ci` and `lerna bootstrap`">`npm run setup`</popover> in the root folder of your cloned repo.
```
marvinchin, line 35: I had some issues due to existing `node_modules` directories in the existing packages. Do you think it's worth adding a note here for existing markbind devs to first run `lerna clean` before running `npm run setup`?

docs/devGuide/settingUp.md
```diff
@@ -32,14 +32,16 @@ We recommend the **WebStorm IDE** for working with MarkBind code.
...
+  <box type="info" seamless>
+  MarkBind uses <md>[lerna](https://github.com/lerna/lerna)</md>, a popular multi-package development tool, to manage it's dependencies. It is essentially a high level wrapper over node and npm's functionalities
+  </box>
+  
+1. **To bind your cloned version of MarkBind to your console** (instead of the released version of MarkBind), run `npm link` in the `packages/cli` folder of the cloned repo.
```
marvinchin, line 40: Does this work for you now? I'm getting errors because our internal packages haven't been published yet.

ang-zeyu: got that too, you could get around it by temporarily removing the `@markbind/core` dependency and `npm link` then undoing it

docs/devGuide/settingUp.md
ang-zeyu: `lerna clean` won't work unless you have it installed globally or have run `npm ci` (which `npm run setup` does) first unfortunately. I think the devguide should be written in the perspective of new developers always though. Experienced developers could always drop an issue / ping, and have a likely have a better understanding of node already 😁

.eslintignore
ang-zeyu: 👍 fixed

ang-zeyu: > Some questions/thoughts:
> - Do you know why the FontAwesome version was bumped? Ideally, we should retain the original version in this PR to keep these concerns separate and avoid unnecessary diffs.

we previously installed the core package transitively via the cli package, so there was no `package-lock` there. so the latest semantic ver gets installed.

I'll get out a separate pr to update it though, then rebase this one on top of that

> - I feel that we should keep the `README.md` in the root. This can help new devs who first visit our repo some information on the project and how to get started at first glance. Maybe we should set up a new README for `cli` and each of the other leaf packages, with a short explanation on what each package does.

🤦‍♂️ unintentional. Will revert and add some brief readmes for `cli/core` after the `npmignore` commit during the rebase above

ang-zeyu:
Updated:
- reverted README.md movement 502041c
- add README for cli / core package 6360743
- dropped font awesome commit

---

https://github.com/lerna/lerna/blob/master/doc/hoist.md

Also switched over to `lerna bootstrap --hoist` in `npm run setup` (`package-lock` changes in 502041c & d6b62c3 only), which is compatible with the current setup and should install faster + take up less space.

the second disadvantage mentioned isn't applicable either as `eslint-plugin-import` guards against it

docs/devGuide/settingUp.md
marvinchin: ah fair enough!

packages/cli/package-lock.json
```diff
@@ -0,0 +1,4683 @@
+{
```
marvinchin, line 1: I presume that these package-lock changes are done by `lerna bootstrap`? Ideally we'd be able to verify that our dependency versions are still the same, but these diffs seem a little too large to manually verify 🙁

packages/vue-components/src/Searchbar.vue
```diff
@@ -21,9 +21,9 @@ export default {
       }
       const pages = [];
       const regexes = this.value.split(' ')
-        .filter((searchKeyword) => searchKeyword !== '')
```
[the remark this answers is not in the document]
ang-zeyu: yup the vue eslintrc now properly extends the previous root config

packages/cli/package-lock.json
ang-zeyu: yep, I think that if our functional tests stay the same we should be good though, since `package-lock` is only effective in development environment

if there are changes (like the font awesome bundles), we can then investigate them and deal with the diff in a separate pr

ang-zeyu:
Updated @marvinchin, seems to work. Travis seems to be broken on all prs though, likely unrelated to the changes here

494ba03 - rename the packages with `markbind-` prefix
4dc983d - changed `lerna publish` to -> `lerna publish from-git` ( tells lerna to use the version commit from `lerna version` instead of running `lerna version` again )

marvinchin:
This is a big PR with multiple steps. I feel that we should do a merge commit, rather than squashing the commit history here.

I also think we should add some detail to commit 494ba03 explaining why we are doing the renaming.

ang-zeyu:
> This is a big PR with multiple steps. I feel that we should do a merge commit, rather than squashing the commit history here.

yup that's the plan

> I also think we should add some detail to commit [494ba03](https://github.com/MarkBind/markbind/commit/494ba03d56ff767a9d32f88d03c06a5f37efb983) explaining why we are doing the renaming.

Just received npm permissions from @yamgent, will revert back to `@markbind/xx`!

ang-zeyu:
reverted to `@markbind`, should be ready to merge now. Thanks @marvinchin @yamgent!

Issue #567, opened 2019-01-09
Remove stray keyword in user guide's search index


damithc:
Current: 
![image](https://user-images.githubusercontent.com/1673303/50880356-204a7800-1419-11e9-9b42-0b24a83deeb3.png)

Problem: there should not be a search result for `regressions`

This is likely a mishap in the code that explains the search keyword feature.

![image](https://user-images.githubusercontent.com/1673303/50880450-60a9f600-1419-11e9-81c8-0a695c546b75.png)

acjh:
This isn't a bug, since it's intentional to render the example as-is.

Perhaps a more appropriate example such that it fits the section?

damithc:
> This isn't a bug, since it's intentional to render the example as-is.

Yes, but perhaps OK to tweak the code to make it _look_ exactly like how it would look when rendered but not _behave_ the same way? In fact, I think I did that for the heading `Developer Testing` which looks like a heading but not behave like one (that's why it doesn't turn up in search results) but forgot to do the same for the keyword.

Current code:
```html
<box>

<big>**Developer Testing**</big><br>

This is good for catching <span class="keyword">regressions</span>.
</box>
</div>
```

> Perhaps a more appropriate example such that it fits the section?

Can't think of an example that will give the correct search behavior, as in, a user will not be surpised to reach this location by searching for that keyword neither will she be surpised to see it in search results. Any ideas?

acjh:
> perhaps OK to tweak the code to make it *look* exactly like how it would look when rendered but not *behave* the same way?

I don't think it's good to do that. I'd expect an example to work properly and that header to be `<h4>`.

Alternatively, we can use a screenshot.

> Can't think of an example that will give the correct search behavior, as in, a user will not be surpised to reach this location by searching for that keyword neither will she be surpised to see it in search results. Any ideas?

We could use `<span class="keyword">keywords</span>`.

damithc:
> I don't think it's good to do that. I'd expect an example to work properly and that header to be `<h4>`.

I agree.

> We could use `<span class="keyword">keywords</span>`.

It feels a bit 'too clever'. Going meta like that can make it slightly harder for the reader to understand. 

> Alternatively, we can use a screenshot.

Yes, that works, as this is a static example. But the look and feel will be slightly different based on the OS, right?

damithc:
> The idea of the example working properly, while not having it in the search index sounds like a contradiction to me. :P

[...]

PR #1294, opened 2020-07-30
Misc doc updates

### LINKED ISSUE #339: Make markbind styles cumulative on top of bootstrap styles
v1.9.0

Example: Bootstrap allows `<p class="lead"> some text </p>` to be used to show leading text of an article in a different formatting.
image

This normally works in markbind, but as soon as we apply some markbind formatting on the text, it overrides the bootstrap formatting. e.g., `<p class="lead"> **some** text </p>` will show the text in normal formatting + bolding

Suggestion: in general, it's nice if our own styles such as bold/underline/tooltip etc. can be applied on top of styles applied by bootstrap.

### LINKED ISSUE #611: <tooltip> example code uses undocumented option `effect`

In our [documentation for the `<tooltip>` component](https://markbind.github.io/markbind/userGuide/usingComponents.html#tooltip), we use the `effect="scale"` option in the example code, which is not documented.

The `effect` option should be documented, or be removed depending on what should be the actual effect of `effect` (I tried `effect="fade"` like other components, but it didn't seem to work).

### LINKED ISSUE #888: Update docs for relative intra-site links support
Current:
* This works: ``
* This does not: ``
  (but this works ``)

Suggestion: support relative URLs for the latter as well.

Note: this is a left-over from https://github.com/MarkBind/markbind/pull/677

UPDATE: as mentioned in the thread below, the issue has been resolved. Just need to update docs a bit.

### LINKED ISSUE #993: Content Insertion into <head> does not respect file types

* **MarkBind Version: 2.9.0**

According to the [UG](https://markbind.org/userGuide/tweakingThePageStructure.html#page-head), adding raw JS and CSS files to the `<head>` is supported. 

I attempted to insert a custom CSS file into the head of a page.

**index.md**
```html
<frontmatter>
  header: header.md
  pageNav: 2
  pageNavTitle: "Chapters of This Page"
  siteNav: site-nav.md
  head: customStyles.css
</frontmatter>
```

**customStyles.css**
```css
.test {
	color: red;
}
```

I expected the contents of `customStyles.css` to be wrapped in a `<style>` tag and inserted. 

The contents of `customStyles.css` is inserted raw, i.e. without being wrapped in a `<style>` tag.

This issue also affects JS files inserted this way.

**Proposed solution:**
Edit `Page.js` to wrap contents in `<style>` and `<script>` tags for CSS and JS files respectively.

### LINKED ISSUE #1106: Document multiple markdown-it-attrs for users

MacOS

* **MarkBind Version:**
Latest master

````html
```xml {heading="Heading title", highlight-lines="2"}
<foo>
  <bar type="name">goo</bar>
</foo>
```
````

The `,` shouldn't have appeared in the heading title.

Screen Shot 2020-03-09 at 3 24 34 PM

Can check PR #1105

### LINKED ISSUE #1151: Restructure plugin documentation

### LINKED ISSUE #1230: Document the -s flag
A `-s` flag was added to the CLI in #440 
This flag doesn't seem to be documented in user docs.

### LINKED ISSUE #1258: Dot in variable names cause problem inside <frontmatter>
See build error https://travis-ci.org/github/reposense/RepoSense/jobs/700289052

Caused by this code: 
```html
<variable name="title">`run.sh` format</variable>
<frontmatter>
  title: "{{ title }}"
</frontmatter>
```

Specifically, it's the `.` in `run.sh` that's causing the problem. `@` is similarly problematic but not `:`.
This problem is present in v2.14.1 but no longer present in the current master.

Sharing here in case it is useful.

### LINKED ISSUE #1276: Document syntax features missing from the user guide
Realized that these features are supported but are missing in our documentation. Some of which are shown in https://markdown-it.github.io/

`~subscript~`
`^superscript^`

Starting a list with a given number:
```
52. item
1. item
1. item
```

ang-zeyu:

• [x] Documentation update

Resolve a couple more documentation issues before release

810244d
https://github.com/MarkBind/markbind/pull/1274#discussion_r460495295

ac13f89
Resolves #1106 - review would be nice here, not sure if it's confusing @damithc 
Should resolve #339 as well - `<md>` tags are already documented
Resolves #888 
Resolves #1276 

06e1907
Resolves #1258 

2377dd4
Resolves #1151 - not too sure where to put this in the site nav, looks a little out of place 🤔

ae7a0bb
Resolves #611 

3c5b75b
Resolves #1230 

50c92a1
Resolves #993 

Let's do a rebase merge here


docs/userGuide/syntax/attributes.mbdf
```diff
@@ -0,0 +1,65 @@
...
+<variable name="code">
+Apply classes, attributes, identifiers to block level markdown (eg. paragraphs, headings)
+by leaving a space before '{' {.text-success #attribute-example}
+
+#### heading {.text-info}
+
+--- {.border-danger}
+
+Apply the same to inline markdown (eg. bold text) by
+omitting the **space**{.text-primary .bg-light attribute=value}
```
damithc, line 20: image

mistake?

docs/userGuide/syntax/attributes.mbdf
```diff
@@ -0,0 +1,65 @@
...
+</variable>
+</include>
```
damithc, line 22: Side question: what if I want to apply a style to multiple words inside a paragraph that doesn't have any markdown symbol around them? No shorthand syntax for that I guess.

docs/userGuide/components/advanced.md
```diff
@@ -1,81 +1,66 @@
...
-  As shown in this panel, using the header slot allows you to customize the Panel's header using HTML.
+  As shown in this panel, using the header slot
+  allows you to customize the Panel's header using HTML.
 </panel>
+</variable>
+<variable name="highlightStyle">html</variable>
+</include>
+</div>
+  
+<modal header="Richer formatting of attributes using slots" id="on-slots">
```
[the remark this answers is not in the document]
ang-zeyu: 👀 never knew this existed, it's undocumented as well. added a commit to document this too

docs/userGuide/syntax/attributes.mbdf
ang-zeyu: html attributes don't require the `"quotes"` if there aren't any spaces in the value

docs/userGuide/syntax/attributes.mbdf
ang-zeyu: yep don't think there's any shorthand for that unfortunately

docs/_markbind/layouts/devGuide/navigation.md
```diff
@@ -6,6 +6,7 @@
 * [Setting up]({{baseUrl}}/devGuide/settingUp.html)
 * [Workflow]({{baseUrl}}/devGuide/workflow.html)
 * [Design]({{baseUrl}}/devGuide/design.html)
+* [Writing Plugins]({{baseUrl}}/devGuide/writingPlugins.html)
```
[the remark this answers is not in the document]
ang-zeyu: Added. I think the main content should stay in the devguide though, since the large majority of users won't be writing such plugins.

docs/userGuide/syntax/attributes.mbdf
damithc: I didn't mean the quotes. That whole thing shouldn't be there, right?

ang-zeyu: it's to show an example of attributes (vs classes and ids), but since these aren't visual, you'd have to use inspect element to see it.

damithc: Can use a real or real-looking attribute name and a value instead? Otherwise it looks like a generic placeholder.

ang-zeyu: added a comment to use inspect element

ang-zeyu:
also:

Resolves #567

PR #1297, opened 2020-08-01
Use js for tests

### LINKED ISSUE #829: Use Javascript for developer scripts instead of having both bash and batch files
At the moment, we have both bash and bat files for developer scripts (e.g. `test.sh` vs `test.bat`). This makes it difficult for developers to update both files as they need both OS'es to test the two different versions of the scripts.

Perhaps we can consider moving the scripts to Javascript instead - since we need Javascript for the test regardless.

ang-zeyu:

• [x] Other, please explain: Standardisation of test / update test scripts

Resolves #829 

Standardise scripts to reduce maintainence from 2 sets of files (bash + sh)

- Use the existing `.sh` script as a reference, creating test / update test scripts with js
- Remove existing `win` version of CI scripts (and in dev guide)

Any additional test cases I missed out below

**Testing instructions:**
For each category of test sites ( conversion, normal test sites, template test sites ),
- change something in the source file -> tests should fail
- change something in the source file ->  update tests -> test files update as expected

Use javascript for tests

Test and test update scripts are written using bash and shell scripts,
which increases maintainence due to having to maintain two sets of OS
dependent scripts.

Let's rewrite these scripts using javascript instead.


ang-zeyu:
Retested the above test cases again, which should cover everything our batch and shell scripts currently does.

@marvinchin let me know if this isn't what you had in mind! :)

marvinchin:
@ang-zeyu sorry for the delay! This definitely fits what I had in mind. Let me take a closer look at this tomorrow, and we can address any comments (if any!) in a future PR.

ang-zeyu:
No worries, thanks @marvinchin!

I merged this in just for convenience as it has a few conflicts I'd like to resolve with #1306 (which is more pressing since we have a few new contributors coming in - I'm targeting to add live / hot reload by this week). But neither are critical; We can always correct things with reviews later on if needed :-)

packages/cli/test/functional/test.js
```diff
@@ -0,0 +1,69 @@
...
+    execSync(`node ../../index.js build ${siteName}`, execOptions);
+    compare(siteName);
+  } catch (err) {
+    printFailedMessage(err, siteName);
+    process.exit(1);
+  }
+});
+
+testConvertSites.forEach((siteName) => {
+  console.log(`Running ${siteName} tests`);
```
marvinchin, line 37: I feel like we can further break these into separate `testXYZSite` functions.

packages/cli/test/functional/test.js
```diff
@@ -0,0 +1,69 @@
...
+    printFailedMessage(err, siteName);
+    cleanupConvert(path.resolve(__dirname, siteName));
+    process.exit(1);
+  }
+  cleanupConvert(path.resolve(__dirname, siteName));
+});
+
+testTemplateSites.forEach((templateAndSitePath) => {
+  const flag = templateAndSitePath.split(',')[0];
+  const sitePath = templateAndSitePath.split(',')[1];
```
marvinchin, line 53: Since we reuse the split path, perhaps we can store it in a separate variable?

packages/cli/test/functional/testSites.js
```diff
@@ -0,0 +1,19 @@
+const testSites = [
+  'test_site',
+  'test_site_algolia_plugin',
+  'test_site_expressive_layout',
+  'test_site_special_tags',
+];
+
+const testConvertSites = ['test_site_convert'];
+
+const testTemplateSites = [
```
marvinchin, line 10: Let's leave a comment explaining the syntax to express a template site test.

packages/cli/test/functional/testUtil/compare.js
```diff
@@ -0,0 +1,73 @@
+const fs = require('fs');
+const path = require('path');
+const ignore = require('ignore');
+const walkSync = require('walk-sync');
+const { isBinary } = require('istextorbinary');
+const diffChars = require('./diffChars');
+
+const _ = {};
+_.isEqual = require('lodash/isEqual');
```
marvinchin, line 9: We don't seem to use `isEqual` in this file. Is this import necessary?

packages/cli/test/functional/testUtil/compare.js
```diff
@@ -0,0 +1,73 @@
...
+      .replace(CRLF_REGEX, '\n');
+
+    if (!NULL_WHITELIST.ignores(expectedFilePath) && isBinary(null, expected)) {
+      // eslint-disable-next-line no-console
+      console.warn(`Unrecognised file extension ${expectedFilePath} contains null characters, skipping`);
+      continue;
+    }
+
+    const hasDiff = diffChars(expected, actual, expectedFilePath);
+    error = error || hasDiff;
```
marvinchin, line 64: I think we should leave a note to future contributors explaining that not terminating early in the case of an error is expected behaviour.

packages/cli/test/functional/updatetest.js
```diff
@@ -0,0 +1,67 @@
...
+  try {
+    execSync(`node ../../index.js build ${siteName} ${siteName}/expected`, execOptions);
+  } catch (err) {
+    printFailedMessage(err, siteName);
+    process.exit(1);
+  }
+});
+
+testConvertSites.forEach((siteName) => {
+  console.log(`Updating ${siteName}`);
```
marvinchin, line 35: Similarly, I think we can abstract this into a separate function.

marvinchin: There's also quite a bit of overlap with the functions used to generate the output in `test.js`. Perhaps we should generalize the functions that performs the build. We can pass in the optional argument for where the built files should be output. I think that would allow us to use it in both `test` and `updatetest`.

packages/cli/test/functional/test.js
ang-zeyu:
```
(siteName) => {
  console.log(`Running ${siteName} tests`);
  try {
    execSync(`node ../../index.js build ${siteName}`, execOptions);
    compare(siteName);
  } catch (err) {
    printFailedMessage(err, siteName);
    process.exit(1);
  }
}
```

you mean to reuse this for all three test types? Sounds good!

packages/cli/test/functional/testUtil/compare.js
ang-zeyu: Nice catch, I'll see if I can make the eslint ignore more specific as well

packages/cli/test/functional/test.js
marvinchin: Hmm I meant having separate `testConvertSite`, `testTemplateSite` and `testSite` functions, with each function doing what is currently done in the anonymous function we currently have. Not sure if this is what you meant!

ang-zeyu: Hmm I meant something like this:
```
genericTestFunction = (siteName, cleanup) => {
  console.log(`Running ${siteName} tests`);
  try {
    execSync(`node ../../index.js build ${siteName}`, execOptions);
    compare(siteName);
  } catch (err) {
    printFailedMessage(err, siteName);
    process.exit(1);
    cleanup()
  }
  cleanup()
}

testConvertSites.forEach((siteName) => {
  genericTestFunction(siteName, cleanupCallback, ...)
})
```

Did you mean to just extract the anonymous functions as named functions?

PR #1309, opened 2020-08-02
Extract component preprocessor

ang-zeyu:

• [x] Other, please explain:

Another part of https://github.com/MarkBind/markbind/issues/810 (leaving `render` and a few other utility methods to refactor out in `Parser.js`)

- make `componentPreprocessor` a class to avoid passing configuration parameters all throughout its methods
- move `includeFile` into `componentPreprocessor` to consolidate functionality
- adapt tests

na

**Testing instructions:**
- `npm run test` passes

Extract componentPreprocessor into a class

The includeFile process depends on heavily on configuration parameters
being passed repeatedly throughout its functions (in
componentPreprocessor).
Let's make componentPreprocessor a class instead, passing these
configurations as instance properties.

The Page model's included sources are collected in instance properties
of Parser and then reconsolidated in page generation.
Let's create a PageSources model to hold these properties instead, and
correspondingly move them to the ComponentPreprocessor class.
This clearly defines the schema and purpose of a Page's sources.

The entry point to ComponentPreprocessor resides in the includeFile
method, which resides in Parser.
Let's move this to the ComponentPreprocessor class to benefit from the
instance properties, and also to consolidate include functionality
as well.


ang-zeyu:
@marvinchin just an update on the changes here: rather large diff (hence difficult to review), but mostly just syntactic indentation / whitespace diffs. 🙂 

The change here mainly is **a.** to make `componentPreprocessor` a class (hence the diffs) **b.** "cut and paste" `includeFile` out of `Parser.js` so we can remove `Parser.js` entirely (as part of #810).

Will be dealing with the remaining methods in `Parser.js` in 2 more upcoming PRs!

PR #1312, opened 2020-08-02
Refactor parser utility functions

ang-zeyu:

• [x] Other, please explain: Follow up to #1310, last part of `parser.js` of #810 

- Move `processDynamicResources` into `convertRelativeLinks` under `ComponentParser` (slight ~3% perf improvement here)
- Cut and paste `unwrapIncludeSrc` into `Page.js`

na

**Testing instructions:**
`npm run test` passes

Refactor parser utility functions

Parser contains some remaining utility functions.

Let's move these out into more appropriate places.


packages/core/src/parsers/ComponentParser.js
```diff
@@ -618,7 +631,10 @@ class ComponentParser {
     return node;
   }
 
-  render(content, filePath) {
+  render(file, content, cwf = file) {
```
ang-zeyu, line 634: parameter order standardisation with `ComponentPreprocessor`, since there is also now the additional `cwf parameter

PR #1328, opened 2020-08-17
Add site.json properties for file exclusions

### LINKED ISSUE #715: Is there a more concise way to specify file exclusions in site.json?
Is there a way to make the following site.json setting more concise?

```json
    {"glob": "se-book-adapted/common/*.md"},
    {"glob": "book/combined/exercises.md"},
    {"src": "se-book-adapted/print.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/architecture.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/cppToJava.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/designApproaches.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/designPatterns.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/gitAndGithub.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/javaTools.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/junit.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/principles.md",
      "searchable": "no"
    },
    {"glob": "se-book-adapted/chapters/intellij.md",
      "searchable": "no"
    }
```

More generally, if the folder has 100 files and if we want to exclude 20 of them, is there a way to do that concisely?

ryoarmanda:

Resolves #715 

With the usage of globs for defining pages to render, if there are any files or directories within the glob that the user wishes to exclude, there is no concise way for them to declare it. It would be better to have a dedicated file exclusion rule for this purpose.

Added two new properties:

- `globExclude` can be used inside each entry of `pages`, whose value is an array that contains glob patterns to be excluded. The exclusions will only affect files found for that particular entry.
- `pagesExclude` is the global variant of `globExclude`, designed to be used on the root level and will affect all entries.

Users can choose which property to use for their exclusions. If both are used at the same time, it will be combined into one.

NA

**Testing instructions:**
`npm run test`

Add site.json properties for file exclusions

When using globs to define pages, there is no concise way for users to
declare files or directories to exclude within the glob.

Let's add new properties to site.json for containing file patterns to
exclude and incorporate these during site building.


docs/userGuide/siteJsonFile.md
```diff
@@ -156,6 +159,11 @@ The following properties will apply to `index.md`:
 </box>
 </span>
 
+#### **`pagesExclude`**
+**An array of file patterns to be excluded from rendering.** The exclusion pattern follows the glob syntax.
+
+This property is the global variant to the `globExclude` property and is functionally identical to it. If both properties are used simultaneously, the file patterns will be combined into one.
```
ang-zeyu, line 165: would `... combined into one.` -> `The file patterns from both properties will be combined when excluding pages` be clearer?

docs/userGuide/siteJsonFile.md
```diff
@@ -33,9 +33,11 @@ Here is a typical `site.json` file:
       }
     },
     {
-      "glob": "**/index.md"
+      "glob": "**/index.md",
+      "globExclude": ["internal/index.md"]
     }
   ],
+  "pagesExclude": ["admin/index.md"],
```
ang-zeyu, line 40: should we make the example a little more realistic? 

```
"pagesExclude": ["subsite/*.md"]
```

if possible also the above one as well

ryoarmanda: Ahh yes that would be better

docs/userGuide/siteJsonFile.md
ryoarmanda: Yes I feel that would make the sentence clearer and concise. I'll reword that part.

docs/userGuide/siteJsonFile.md
ryoarmanda: I could change the `globExclude` value to `subsite/**/index.md` and the `pagesExclude` to `subsite/*.md`, would that be okay?

ang-zeyu: hmm I used `subsite` in `pagesExclude` because you would normally only want to reuse all files in the subsite, not deploy them as part of the main site. (it should be `subsite/**.md` as well my mistake)

I was thinking something like this (barring the generic `topic1` 🙈):

```
glob: "topic1/**.md",
globExclude: ["topic1/appendix_*.md"], (not sure if this a valid glob :eyes:)
layout: "subtopic"
```

where the benefit of `globExclude` becomes clear as opposed to `pagesExclude`

ryoarmanda: Ohh I see what you mean now. I was trying to find other general names for the example and just couldn't find any other than `subsite` 😓 I think the above is quite nice, I might change it a bit to be slightly more generalized (following the "topic" illustration):

```
glob: "topics/*/*.md",
globExclude: ["topics/*/appendix/*.md"],
layout: "subtopic"
```

With `pagesExclude` still contains `subsite/*/*.md` (I think this is proper glob already, so actually you didn't do any mistakes haha).

ang-zeyu: looks good! 👍

ryoarmanda: 👌 I have pushed a commit for the change

PR #947, opened 2019-12-07
[WIP] Log website URL upon deployment

yash-chowdhary:
**What is the purpose of this pull request? (put "X" next to an item, remove the rest)**

• [x] Other, please explain: Involves a logging change

Fixes #844 

**What is the rationale for this request?**
This change helps the user easily access the website after deploying it using the MarkBind CLI.

**What changes did you make? (Give an overview)**
In order to get the URL of the deployed website, the details of the repository are required. Should the remote URL of the repository not be present in the deploy section of `site.json`, an external library (NodeGit) will fetch it. The website URL (deployment URL) is constructed using this remote URL.

**Provide some example code that this change will affect:**

```js
function getUrl() {
  return new Promise((resolve) => {
    Git.Repository.open('.git').then(((repo) => {
      repo.config().then(((config) => {
        config.getStringBuf(REMOTE_ORIGIN_URL_KEY)
          .then(((buf) => {
            resolve(buf.toString());
          }));
      }));
    }));
  });
}
```

**Proposed commit message: (wrap lines at 72 characters)**
Log website URL upon MarkBind deployment

PR #1329, opened 2020-08-18
Log website URL upon deployment

KendrickAng:

• [ ] Documentation update
• [ ] Bug fix
• [ ] New feature
• [ ] Other, please explain:

 This change helps the user easily access the website after deploying it using the MarkBind CLI.

Github supports custom domains through the use of `CNAME`. However, to do so a file called `CNAME` must be placed in the root of the remote repository where we are deploying (the file is automatically created if one sets a manual deploy url through the github repo itself). Hence when `markbind deploy` is run, the following additional steps are done:
1. read the contents of the `CNAME` file in the `origin` remote with external library `simple-git`.
2. If not, there must be no custom url, and we proceed to manually construct the URL, first from the provided `repo` from `site.json`, otherwise using the remote itself.

```js
 static getDeploymentUrl(options) {
    const HTTPS_PART = 'https://';
    const SSH_PART = 'git@github.com:';
    const GITHUB_IO_PART = 'github.io';

    // https://<name|org name>.github.io/<repo name>/
    function constructGhPagesUrl(remoteUrl) {
      if (remoteUrl.includes(HTTPS_PART)) {
        // https://github.com/<name|org>/<repo>.git (HTTPS)
        const parts = remoteUrl.split('/');
        const repoName = parts[parts.length - 1].toLowerCase();
        const name = parts[parts.length - 2].toLowerCase();
        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
      } else if (remoteUrl.includes(SSH_PART)) {
        // git@github.com:<name|org>/<repo>.git (SSH)
        const parts = remoteUrl.split('/');
        const repoName = parts[parts.length - 1].toLowerCase();
        const name = (parts[0].split(':'))[0];
        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
      }
      throw new Error(`Unknown remote url ${remoteUrl}`);
    }

    const { remote, branch, repo } = options;
    const cnamePm = gitUtil.getRemoteBranchFile('blob', remote, branch, 'CNAME');
    const remoteUrlPm = gitUtil.getRemoteUrl(remote);
    const promises = [cnamePm, remoteUrlPm];

    return Promise.all(promises)
      .then((results) => {
        const cname = results[0];
        const remoteUrl = results[1];
        if (cname) {
          return cname;
        } else if (repo) {
          return constructGhPagesUrl(repo);
        }
        return constructGhPagesUrl(remoteUrl);
      })
      .catch((err) => {
        logger.error(err);
        return null;
      });
  }
```

**Testing instructions:**
1. Create a dummy website using `markbind init`.
2. Run `markbind deploy` and check the url the website is deployed at.

Log website URL upon deployment


packages/core/test/unit/Site.test.js
```diff
@@ -27,6 +27,11 @@ jest.mock('fs');
 jest.mock('walk-sync');
 jest.mock('gh-pages');
 jest.mock('../../src/Page');
+jest.mock('simple-git', () => () => ({
```
ang-zeyu, line 30: is this mock still necessary?

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
...
+   * Gets the url where the website is deployed at.
+   */
+  static getDeploymentUrl(git, options) {
+    const HTTPS_PART = 'https://';
+    const SSH_PART = 'git@github.com:';
+    const GITHUB_IO_PART = 'github.io';
+
+    // https://<name|org name>.github.io/<repo name>/
+    function constructGhPagesUrl(remoteUrl) {
+      if (remoteUrl.includes(HTTPS_PART)) {
```
ang-zeyu, line 1329: should we use `startsWith` here instead?

Likewise for `SSH_PART` (i'm not too sure on the format requirements here though, correct me if I'm wrong!)

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
...
+        const name = parts[parts.length - 2].toLowerCase();
+        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
+      } else if (remoteUrl.includes(SSH_PART)) {
+        // git@github.com:<name|org>/<repo>.git (SSH)
+        const parts = remoteUrl.split('/');
+        const repoName = parts[parts.length - 1].toLowerCase();
+        const name = (parts[0].split(':'))[0];
+        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
+      }
+      throw new Error(`Unknown remote url ${remoteUrl}`);
```
ang-zeyu, line 1342: Since the message here isn't used, let's just return `null` directly for use in the above.
```
depUrl => (depUrl != null ? resolve(`Deployed at ${depUrl}!`) : resolve('Deployed!'))
```

Could remove the `catch(..)` below as well

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
 
           return publish(basePath, options);
         })
-        .then(resolve)
+        .then(() => {
+          const git = simpleGit({ baseDir: process.cwd() });
+          const options = {};
```
ang-zeyu, line 1308: Is it possible to move the previous `options` up to avoid duplication?

packages/core/src/utils/git.js
```diff
@@ -0,0 +1,26 @@
...
+ * Wrapper around simple-git providing helper functions to retrieve repo remote URLs.
+ */
+module.exports = {
+  /**
+   * Returns the contents of a remote file, undefined otherwise.
+   * See: https://git-scm.com/docs/git-cat-file for accepted values for each input.
+   */
+  async getRemoteBranchFile(simpleGit, type, remote, branch, fileName) {
+    const catFileTarget = `${remote}/${branch}:${fileName}`;
+    return simpleGit.catFile([type, catFileTarget])
```
ang-zeyu, line 13: Not too familiar with `CNAME` file and `cat-file`, give me some time to take a look at that!
Perhaps one of the other maintainers @acjh @marvinchin might have across it before though

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
...
+        // https://github.com/<name|org>/<repo>.git (HTTPS)
+        const parts = remoteUrl.split('/');
+        const repoName = parts[parts.length - 1].toLowerCase();
+        const name = parts[parts.length - 2].toLowerCase();
+        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
+      } else if (remoteUrl.includes(SSH_PART)) {
+        // git@github.com:<name|org>/<repo>.git (SSH)
+        const parts = remoteUrl.split('/');
+        const repoName = parts[parts.length - 1].toLowerCase();
+        const name = (parts[0].split(':'))[0];
```
ang-zeyu, line 1339: should the ending be `[1]`? would `parts[0].substring(SSH_PART.length)` work instead?

ang-zeyu (commented):
`package-lock` (from running `npm run setup`) will need to be committed too

Some comments:

packages/cli/index.js
```diff
@@ -298,9 +298,7 @@ program
     const rootFolder = path.resolve(process.cwd());
     const outputRoot = path.join(rootFolder, '_site');
     new Site(rootFolder, outputRoot, undefined, undefined, options.siteConfig).deploy(options.travis)
-      .then(() => {
-        logger.info('Deployed!');
-      })
+      .then(deployMsg => logger.info(deployMsg))
```
marvinchin, line 301: It doesn't feel very intuitive for the `Site` constructor to return a message. The message feels more like the responsibility of the CLI. I think that it might be more appropriate for it to return some data about the site created (the url in this case), which we can then use to construct the deploy message here.

packages/core/src/Site/index.js
marvinchin: I think we can use a more explicit variable name (maybe `deployOptions`?) to make it clearer what kind of `options` these are as well!

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
...
-        .then(resolve)
+        .then(() => {
+          const git = simpleGit({ baseDir: process.cwd() });
+          const options = {};
+          options.remote = defaultDeployConfig.remote;
+          options.repo = this.siteConfig.deploy.repo || defaultDeployConfig.repo;
+          options.branch = this.siteConfig.deploy.branch || defaultDeployConfig.branch;
+          return Site.getDeploymentUrl(git, options);
+        })
+        .then(depUrl => (depUrl != null ? resolve(`Deployed at ${depUrl}!`) : resolve('Deployed!')))
```
marvinchin, line 1314: Can we use `!==` instead of `!=`?

packages/core/src/Site/index.js
```diff
@@ -1300,11 +1303,67 @@ class Site {
...
+        const parts = remoteUrl.split('/');
+        const repoName = parts[parts.length - 1].toLowerCase();
+        const name = (parts[0].split(':'))[0];
+        return `https://${name}.${GITHUB_IO_PART}/${repoName}`;
+      }
+      throw new Error(`Unknown remote url ${remoteUrl}`);
+    }
+
+    const { remote, branch, repo } = options;
+    const cnamePm = gitUtil.getRemoteBranchFile(git, 'blob', remote, branch, 'CNAME');
```
marvinchin, line 1346: Maybe I'm missing some context here, but I'm not sure what `Pm` stands for. Is it for promise?

marvinchin: If it is, I'd prefer to avoid using uncommon (to me, at least! if this is a convention you are more familiar with please correct me) abbreviations to avoid confusion when future devs look at this variable.

packages/core/src/utils/git.js
marvinchin: This isn't something I'm familiar with either, unfortunately :(

marvinchin (changes requested):
Thanks for looking at this! Left a couple of suggestions 🙂

ang-zeyu (approved):
Sorry for the wait

I should really have untagged this as good first issue given the amount of additions in #947 (oh well =X)
Thanks! And well done for a first issue : )

ang-zeyu (approved):

Issue #1314, opened 2020-08-06
TopNav: dropdown menu is not scrollable


damithc:
![image](https://user-images.githubusercontent.com/1673303/89481702-de3b1100-d7ca-11ea-9a57-0dd49faa55b7.png)

If the menu is too big for the screen, there is no way to reach the menu items outside the screen.

ang-zeyu:
closing per discussion in #1332

PR #1332, opened 2020-08-21
[WIP] Make Dropdown menu responsive

wxwxwxwx9:
This addresses issue #1314 by enhancing the dropdown to become responsive to window resizing. 

Currently, the dropdown menu is not responsive to the window resizing. Therefore, when it cannot fit into the current window size, it will be cut in the middle. Refer to issue #1314 for an illustration of the issue. This may affect user experience to some extent. To make it more convenient and usable for users, we should make the dropdown responsive and scrollable upon detection that it cannot fit into the current window size. 

I added a ```max-height``` CSS property to ```dropdown-menu``` and made its value dynamic with respect to window resizing. The value is stored in a CSS global variable called ```--max-dropdown-height``` and is computed with ```navbar``` height and current window height. I have also added a small debouncing feature to improve the performance. 

Implementation, code quality, and potential edge cases. 

**Testing instructions:**
1. Create dropdown and populate it with enough content (such that the height of the dropdown menu is more than the smallest size possible for window resizing)
2. Resize the window to be smaller (in height) than the dropdown menu
3. You should find the dropdown menu to shrink in height accordingly
3. Place your cursor within the boundaries of the dropdown menu
4. You should be able to scroll the menu and reach the hidden content

Make Dropdown menu responsive

**To-do**
Create snapshot test cases


ang-zeyu:
hmm,

I think let's prioritise the cleanest way of doing things here (our own dropdown)

we'd definitely want a fancy navbar over bootstrap's default, which dropdowns are a part of. may become a little messy to rely on bootatrap vue's here.

ang-zeyu:
back to this issue and pr, I'm still not sure if encouraging long dropdowns should be a thing though.

since the use case is for really long list of items, maybe we could support navigable, nested dropdowns instead? (where hovering over one link opens up a whole new dropdown of links to the left / right)

for mobile views, each nested dropdown may be expanded / collapsed like an accordion.

damithc:
> since the use case is for really long list of items, maybe we could support navigable, nested dropdowns instead? (where hovering over one link opens up a whole new dropdown of links to the left / right)
> 
> for mobile views, each nested dropdown may be expanded / collapsed like an accordion.

Yes, that works too. In fact, nested dropdowns will be nice to have either way. 
I wonder that given a mobile device on landscape mode has a very short height, even a reasonable length drop-down might still get cut off (?)

ang-zeyu:
>Yes, that works too. In fact, nested dropdowns will be nice to have either way.
>I wonder that given a mobile device on landscape mode has a very short height, even a reasonable length drop-down might still get cut off (?)

You should still be able to scroll the *page* with the dropdown open in that case though. Unlikely you would reach the end of the page content (and thus unable to scroll further down to see the dropdown) since on mobile content is 'compressed horizontally'.

any thoughts? @wxwxwxwx9

wxwxwxwx9:
Hi Zeyu, just to summarize what we have discussed. 

**Discussion regarding migrating to bootstrap-vue for "dropdown" and "navbar"**

I think the consensus is that we should either: 1) migrate both "dropdown" and "navbar" to bootstrap-vue, or 2) stick to maintaining and extending our own components for "dropdown" and "navbar". The reason for scoping down to these 2 options is to facilitate code quality by reducing the need for transformation (e.g. writing code to transform bootstrap-vue's "dropdown" to fit with our own "navbar"). It is not a good idea to just migrate "dropdown" and not "navbar". 

For option 1, although there are benefits like reliability, proper maintenance from upstream, proper accessibility tagging, and well-implemented features in general, there can be technical challenges in extending / customizing the features for bootstrap-vue's "dropdown" and "navbar" components due to the low-level API they use. For example, we might want to improve the mobile navigability of bootstrap-vue's "navbar". If we were to migrate to bootstrap-vue for "dropdown" and "navbar", we might not be able to enjoy the flexibility we can have if we were to just maintain our own components. 

Furthermore, if we were to transform our syntax into bootstrap-vue's version (like what we did for popovers, modals, and tooltips), it would increase coupling and complexity and make it harder for people to contribute. 

For the reasons above, we agreed that it would be better if we were to go with option 2, and just implement features that we want from bootstrap-vue (e.g. dynamic positioning of the dropdown). 

**Discussion regarding excessively long dropdown and nested dropdowns as a workaround**

We think that there are little use cases for a long dropdown list (e.g. >15 items). Thus, to discourage such practices, we will let it fail softly (overflow the page), like how bootstrap-vue does it. We will, however, implement nested dropdown as a workaround for users who might have a need for a dropdown tab to contain more than 15 items. We will not restrict the number of levels that the user can have for nested dropdowns. When there are too many levels, we will let it fail softly as well. 

The dropdown and nested dropdown will not be stickied to the viewport. It will flow with the page. And in mobile view, the nested dropdown will function like an accordion. 

@damithc what do you think about the above?

damithc:
> **Discussion regarding excessively long dropdown and nested dropdowns as a workaround**
> 
> We think that there are little use cases for a long dropdown list (e.g. >15 items). Thus, to discourage such practices, we will let it fail softly (overflow the page), like how bootstrap-vue does it.
> 
> We will, however, implement nested dropdown as a workaround for users who might have a need for a dropdown tab to contain more than 15 items. We will not restrict the number of levels that the user can have for nested dropdowns. When there are too many levels, we will let it fail softly as well.
> 
> The dropdown and nested dropdown will not be stickied to the viewport. It will flow with the page. And in mobile view, the nested dropdown will function like an accordion.
> 
> @damithc what do you think about the above?

Sounds good to me. 👍

ang-zeyu:
closing this per https://github.com/MarkBind/markbind/pull/1332#issuecomment-743769904, let's outline this in an issue sometime : )

ang-zeyu:
also, thanks again for the work here! @wxwxwxwx9

Issue #1313, opened 2020-08-03
Add compile time tex math support


ang-zeyu:
**Is your request related to a problem?**

Texmath is very commonly used in educational sites. Currently, users have to include external scripts on their own to support this.
This
- increases fouc, as client side rendering texmath on top of what we have is not cheap
- poses a slight inconvenience to users, having to include the scripts and stylesheets on their own

**Describe the solution you'd like**

Use the https://www.npmjs.com/package/markdown-it-texmath plugin to provide compile time texmath support using [katex](https://katex.org/docs/api.html). (unlike mathjax, it does not support the full array of syntax, but is speedy)

tex style
```
$$display math$$

$non-display math$
```

or latex style
```
\( non-display math \)

\[ display math \]
```

not too sure which style would be better / is the dominant standard

**Possible problems**

Syntax compatibility - probably no issues here, as 
1. markdown-it-texmath's rules loads quite high up on the markdown-it rule chain
2. we don't use `$$...$$` or `$...$` or `\( ... \)` or `\[ ... \]` for anything else currently

Extra stylesheet required
1. At `22kb` minified, it's not particularly small but not big either. But the same goes with octicons, github icons, etc. We can provide options to toggle these off later on

Build performance shouldn't be an issue (when not using it)
1. the 2103 site which has no instances of texmath performed within margin of error (once even quicker) in 2 rounds of testing. it's very negligible.
2. any performance loss from actually using it is made up for by providing a better reader experience

Issue #1256, opened 2020-06-18
Process .puml files for nunjucks


damithc:
Current: nunjucks code works inside inline PlantUML diagrams but doesn't work in `.puml` file

Suggestion: Allow nunjucks code in `.puml` files as well.

ang-zeyu:
I don't think it's a good idea; It's not very nunjucks like (in that content belonging to other files is only rendered via `{% include %}`) and can cause issues with users of nunjucks not expecting the .puml to be rendered.

Have you tried
```
<puml name="...">
{% include "....puml" %}
</puml>
```
?

damithc:
> Have you tried
> 
> ```
> <puml name="...">
> {% include "....puml" %}
> </puml>
> ```
> 
> ?

Yes, that works (sans live preview, but that's not a major issue).

> I don't think it's a good idea; It's not very nunjucks like (in that content belonging to other files is only rendered via `{% include %}`) and can cause issues with users of nunjucks not expecting the .puml to be rendered.

I was trying to use variables that are defined in the variables.md, not `<include>`.
e.g.,
```puml
@startwbs
+ <b><color:red>100</color></b> marks
 + Individual Project (iP): <b><color:red>{{ marks_ip }}</color></b> marks
 + Team Project (tP): <b><color:red>{{ marks_tp }}</color></b> marks
@endwbs
```

ang-zeyu:
> I was trying to use variables that are defined in the variables.md, not <include>.

ahh missed that out, hmm, I still think there's quite a bit of feature overlap with `<include>`s though, and I feel like we shouldn't be rendering files other than `.md/mbdf/mbd`

You could still achieve it with a normal `.md/mbdf` file with an inline puml inside; this would also allow you to use the `<variable>`s in `<include>`s feature
```
<include src="xx.md">
<variable name="marks_ip">...</variable>
...
</include>

xx.md

<puml name="...">
@startwbs
+ <b><color:red>100</color></b> marks
 + Individual Project (iP): <b><color:red>{{ marks_ip }}</color></b> marks
 + Team Project (tP): <b><color:red>{{ marks_tp }}</color></b> marks
@endwbs
</puml>
```

that said, it shouldn't be too difficult to render just site variables for `.puml` files as well, what do you think?

damithc:
> You could still achieve it with a normal `.md/mbdf` file with an inline puml inside; this would also allow you to use the `<variable>`s in `<include>`s feature

[...]

PR #1335, opened 2020-08-22
Add compile time tex math support

KendrickAng:

• [ ] Documentation update
• [ ] Bug fix
• [X ] New feature
• [ ] Other, please explain:

This pull request addresses issue #1313 , to add additional compile-time tex math support to Markbind.

This PR adds two new packages to markbind-core: `katex` and `markdown-it-texmath`. `markdown-it-texmath` requires `katex` as a dependency. A new dev-dependency `file-loader` is also added, since `katex`'s css depends on font files `.woff2`.

Some of the user guide changes are shown below:
1. In the landing page
image

2. Under `Formatting Contents`:
image

```js
// In core-web/src/styles/index.css
@import "markbind.css";
@import "page-nav.css";
@import "site-nav.css";
@import "~katex/dist/katex.css";
@import "~markdown-it-texmath/css/texmath.css";

// In ComponentParser.js
  static _parseAnnotationAttributes(node) {
    if (!_.has(node.attribs, 'v-pre')) {
      node.attribs['v-pre'] = true;
    }
  }
```

The quality of documentation in the User Guide

**Testing instructions:**
1. Run `npm run build:web` to rebuild `markbind.min.css` and `markbind.min.js`.
2. In any markbind website, insert use `\( ... \)` for inline equations and `\[ ... \]` for block-level equations. E.g `\(1+1=2\)`
3. To test edge cases with `{{` which conflict with nunjucks parsing, try this: `\(e^{{\frac{1}{3}} + 1}\)`

Add compile time tex math support


packages/core/src/lib/markdown-it/index.js
```diff
@@ -14,6 +14,7 @@ markdownIt.use(require('markdown-it-mark'))
   .use(require('markdown-it-table-of-contents'))
   .use(require('markdown-it-task-lists'), {enabled: true})
   .use(require('markdown-it-linkify-images'), {imgClass: 'img-fluid'})
+  .use(require('markdown-it-texmath'), {engine: require('katex'), delimiters: 'dollars'})
```
ang-zeyu, line 17: quick comment:
let's use `brackets` instead, since we're planning to use `$$` for underline
It should also be the more conventional format I think

KendrickAng: I'll do that!

KendrickAng: @ang-zeyu done!

packages/core-web/src/styles/index.css
```diff
@@ -1,3 +1,5 @@
 @import "markbind.css";
 @import "page-nav.css";
 @import "site-nav.css";
+@import url("~katex/dist/katex.css");
```
ang-zeyu, line 4: what's the additional `url` for?

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,60 @@
+## Math Formulae
+Markbind supports typesetting math equations with [KaTeX](https://katex.org).
+
+<box type="info">
+
+Insert inline equations by enclosing them in round brackets `\( ... \)`.
+
+Insert block equations by enclosing them in square brackets `\[ ... \]`.
```
ang-zeyu, line 8: Should this be `display equations` instead? It seems to be referred in that manner quite often.

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,60 @@
+## Math Formulae
+Markbind supports typesetting math equations with [KaTeX](https://katex.org).
```
ang-zeyu, line 2: Since `katex` is the rendering engine, not the spec, let's start with the spec first, then list the engine's use

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,60 @@
...
+Insert block equations by enclosing them in square brackets `\[ ... \]`.
+
+Insert numbered block equations by enclosing the equation square brackets and the equation number in curly brackets `\[ ... \] (1)`.
+</box>
+
+<box type="important">
+
+You may leave whitespace between the brackets and your equation.
+
+Correct: `\( 1+1=2 \)`, Correct: `\(1+1=2\)`
```
ang-zeyu, line 17: is this a specific trait of `markdown-it-texmath` or part of the tex math? I think we can leave this out and assume the reader already know the spec well if its the latter

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,60 @@
...
+
+Ensure that there is a newline between an inline equation and a block equation, or formatting will not work.
```
ang-zeyu, line 19: it should be not be preceded by a markdown paragraph.
Should we remove this since its a markdown spec thing instead? Then enhance the examples further below to make the use case of inline / display equations clearer instead

e.g. this does not work as well
```
abc
\[2+2=4\]

or this 

> abc
> \[2+2=4\]
```

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,60 @@
+## Math Formulae
```
ang-zeyu, line 1: I've never used latex much before, so take these with a pinch of salt 😐

packages/core-web/src/styles/index.css
ang-zeyu: Untitled

You'll need to copy the font files to the appropriate output directory too

KendrickAng: The additional `url` was how I used to import other css files. Upon checking https://developer.mozilla.org/en-US/docs/Web/CSS/@import, it seems that `url` is redundant. I'll remove it.

docs/userGuide/syntax/mathformulae.mbdf
KendrickAng: Whitespaces being allowed seems to be default behaviour for Tex (based on short experiments only on Overleaf). I had added that in previously since using `$` as delimiters didn't allow whitespace, but with the change in delimiters I suppose it would be better to remove it.

packages/core-web/src/styles/index.css
KendrickAng: image
I've fixed the issue, added extra logic in Site to copy over the files to the appropriate directory! (imo, that was more straightforward compared to two different webpack configurations for dev and prod)

KendrickAng:
@ang-zeyu I've made the changes to the user guide and Site.js based on your comments!

docs/index.md
```diff
@@ -61,6 +61,30 @@ An example that uses GFMD synatx for task lists:
...
+
+```markdown
+Euler's Identity \(e^{i\pi}+1=0\) is a beautiful formula.
+```
+  </td>
+  <td>&nbsp;→&nbsp;</td>
+  <td><br>
+<tip-box>
+
+Euler's Identity \(e^{i\pi}+1=0\) is a beautiful formula.
```
crphang, line 80: A little curious whether an expression like `e^{{\frac{1}_{3}} + 1}` will cause nunjucks to misbehave in Markbind now. Do we need this: https://markbind.org/userGuide/tipsAndTricks.html#using-raw-endraw-to-display-content?

KendrickAng: image
It does seems to be causing nunjucks rendering issues. I'm not too familiar with the intricacies nunjucks (like how to allow for `{{` here), perhaps @ang-zeyu can advise?

ang-zeyu: we allow nunjucks syntax anywhere, even in seemingly unintuitive places like fenced code blocks (we even use this in our ug), the downside of this flexibility is that `{% raw %} {% endraw %}` will be necessary to resolve any ambiguity.

since this seems like a common use case (again correct me if im wrong, havent used latex much before), maybe you can add a warning `<box>` to the ug per @crphang's observation!

`v-pre` shouldn't be though, versus using a client side tex math solution like mathjax, since the equations are spit out at compile time

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
...
+<box type="info">
+
+Insert `inline` equations by enclosing them in round brackets `\( ... \)`.
+
+Insert `display` equations by enclosing them in square brackets `\[ ... \]`.
+
+Insert numbered `display` equations by enclosing the equation square brackets and the equation number in curly brackets `\[ ... \] (1)`.
+</box>
+
+<box type="important">
```
ang-zeyu, line 13: should we move this to the bottom? since the warning would make more sense after presenting the feature

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
+## Math Formulae
+Markbind supports typesetting TeX math equations. [KaTeX](https://katex.org) is used as a fast math renderer.
+
+<box type="info">
+
+Insert `inline` equations by enclosing them in round brackets `\( ... \)`.
```
ang-zeyu, line 6: would it better to bold the **inline/display** or link it somewhere? let's use the inline code syntax just for code

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
...
+
+Insert numbered `display` equations by enclosing the equation square brackets and the equation number in curly brackets `\[ ... \] (1)`.
+</box>
+
+<box type="important">
+
+When typing `display` equations, this does not work as well:
+
+```markdown
+abc
```
ang-zeyu, line 18: https://github.com/MarkBind/markbind/pull/1335#discussion_r480207052

sorry if it was misinterpreted, I didn't mean for you to place this exactly in the docs, it was only to illustrate that it is intentional behaviour by the [commonmark spec](https://commonmark.org/)

I think we should remove _all_ problems related to the markdown spec(s) in general, and let the author get accustomed to it.
if not, just brief mentions that universally apply, like the one we have for blank lines here https://markbind.org/userGuide/usingHtmlJavaScriptCss.html, otherwise every section in our ug would be double in size =X

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
...
+```markdown
+> abc
+> \[2+2=4\]
+```
+
+If your equation requires special Nunjucks tags like {% raw %}`{{`{% endraw %} or {% raw %}`}}`{% endraw %},use a
+[raw-endraw block together with v-pre](https://markbind.org/userGuide/tipsAndTricks.html#using-raw-endraw-to-display-content):
+
+```markdown
+{% raw %}\(e^{{\frac{1}{3}} + 1}\){% endraw %} {v-pre}
```
ang-zeyu, line 32: as mentioned `v-pre` _shouldn't_ be necessary, which is one of the benefits of doing it at compile time to. do double check though

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
...
+
+\( e^{i\pi}+1=0 \)
+\(\frac{1}{3}+\frac{2}{3}=1\)
+
+\[ e^{i\pi}+1=0 \]
+\[\frac{1}{3}+\frac{2}{3}=1\] (1)
+
+{% raw %}\(e^{{\frac{1}{3}} + 1}\){% endraw %} {v-pre}
+
+Bad examples:
```
ang-zeyu, line 92: let's remove this too,
instead make the examples directly above here resemble actual use cases as best; i.e. some _short_ text around an inline equation, then present another relavant equation that is better highlighted in display format

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
+## Math Formulae
+Markbind supports typesetting TeX math equations. [KaTeX](https://katex.org) is used as a fast math renderer.
+
+<box type="info">
```
ang-zeyu, line 4: let's present this vital piece of information about the feature in plain markup, without the box which readers might just skip over

ang-zeyu (commented):
implementation looks good, thanks! : )

just a few more docs nits:

docs/index.md
crphang: When working on nunjuck-escaped special tags with @ang-zeyu, latex's syntax was one of the consideration. Maybe this can be wrapped in nunjuck-escaped special tags? You can check out puml plugin if it helps!

ang-zeyu: > nunjuck-escaped special tags

@crphang this hasen't been implemented yet, I don't think we should either, theres several cases where its useful to have nunjucks syntax complement the "special tag", e.g. https://github.com/MarkBind/markbind/issues/1256

ang-zeyu: wrapping texmath in html may also be rather troublesome (usage wise) to be worth the cost; that said if there is a way to use only the same `\[ \(` syntax that'd be great though (likely needs to be done via patching nunjucks itself)

docs/userGuide/syntax/mathformulae.mbdf
KendrickAng: That makes sense; a user would want to know how to do things first before looking at its edge cases

docs/userGuide/syntax/mathformulae.mbdf
KendrickAng: ah I see, I'll remove this section instead and assume the reader has read the `usingHtmlJavaScriptCss.html` section

docs/userGuide/syntax/mathformulae.mbdf
```diff
@@ -0,0 +1,99 @@
...
+abc
+\[2+2=4\]
+```
+
+or this
+```markdown
+> abc
+> \[2+2=4\]
+```
+
```
KendrickAng, line 27: test comment

docs/userGuide/syntax/mathformulae.mbdf
KendrickAng: I'm not exactly sure why, but leaving out the {v-pre} causes a blank page to show:
image

Whereas leaving it in allows everything to load correctly, so I left it in for now.

KendrickAng:
@ang-zeyu, I've updated the user guide based on your comments. How does it look now?

docs/userGuide/syntax/mathformulae.mbdf
ang-zeyu: seems like katex also spits out MathML accessibility output retaining the original equation by default:

from view page source:
```
<semantics>
--
  | <mrow>
  | <msup>
  | <mi>e</mi>
  | <mrow>
  | <mfrac>
  | <mn>1</mn>
  | <mn>3</mn>
  | </mfrac>
  | <mo>+</mo>
  | <mn>1</mn>
  | </mrow>
  | </msup>
  | </mrow>
  | <annotation encoding="application/x-tex">e^{{\frac{1}{3}} + 1}</annotation>
  | </semantics>

```

let's auto add the `v-pre` to this in `componentParser`

```
<annotation encoding="application/x-tex">e^{{\frac{1}{3}} + 1}</annotation>
```

ang-zeyu:
> @ang-zeyu, I've updated the user guide based on your comments. How does it look now?

really neat now! let's auto add in the `v-pre` to `<annotation encoding="application/x-tex">` which should be only a few lines in `ComponentParser`, then remove it from the ug, and we can merge this in

KendrickAng:
Done!

ang-zeyu (approved):

PR #1336, opened 2020-08-27
Highlighting: Implement slice format for text-only highlight

### LINKED ISSUE #1311: Code highlighting: don't highlighting leading/trailing empty space
Current:
image

Problem: The block-wide highlighted box interferes with the visual structure of the code (as driven by indentations).

I wonder if we can/should omit the highlighting from the leading/trailing empty area, like so:
image

ryoarmanda:

Resolves #1311 

Line highlighting highlights the full width of the code block, which can interfere with the visual structure of the code. It would be better to highlight only the text portion of the code, ignoring indentations.

Added/updated syntax for line highlighting:
- Line text-only highlighting: `lineNum`.
- Range text-only highlighting: `lineStart-lineEnd`.
- Line whole-line highlighting: Now use `lineNumber[:]` (called unbounded line-slice)
- Range whole-line highlighting: Now use `lineStart[:]-lineEnd` or `lineStart-lineEnd[:]` (basically either of the ends is an unbounded line-slice)

Example fenced code (presenting all the line-highlighting capabilities supported):
~~~
```js {start-from=5 highlight-lines="5,6[:],7-8,9[:]-10"}
function f(x) {
  const foo = 1;
  const bar = 2;
  const baz = 3;
  const square = x * x;
  return square;
}
```
~~~

Example view:
image

I'm not sure whether the code organization is proper already. Some of the lines I modified are just following the style of that particular line(s). If you have any suggestions for tidying up, please let me know in the review!

**Testing instructions:**
`markbind serve -d` (to compile the modified `markbind.css`)

Add slice format for text-only code highlight

The highlight from the highlight-lines attribute spans the entire
width of the block, which might interfere with the visual structure
of the code.

Let's add ways for authors to highlight only the text part of the code.


packages/core/src/lib/markdown-it/index.js
```diff
@@ -86,43 +101,104 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
-      if (start && end) {
-        return currentLineNumber >= start && currentLineNumber <= end;
+    // Edit (28/8/2020): I changed the approach from using some() to a simple for-loop. It is still O(n^2).
+    //                   Reason, now with different highlighting methods (whole-line/text-only),
+    //                   checks must be done to determine what method a particular rule follows.
+    //                   As now checking has to be done at rule matching and return handling,
+    //                   it's more concise to write a for-loop so that we can perform both in one block.
+    for (let i = 0; i < highlightRules.length; i++) {
+      const rule = highlightRules[i];
+      const [a, b, c] = rule; // can be up to three items
```
ang-zeyu, line 165: since the processing model is getting heavy, let's create explicit class(es) for the intermediate processing result(s)

packages/core/src/lib/markdown-it/index.js
```diff
@@ -86,43 +101,104 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
-    // Note: The algorithm is based off markdown-it-highlight-lines (https://github.com/egoist/markdown-it-highlight-lines/blob/master/src/index.js) 
+    // check if there is a highlight rule that is applicable to the line number, and handle accordingly
+    // Note: The algorithm is based off markdown-it-highlight-lines (https://github.com/egoist/markdown-it-highlight-lines/blob/master/src/index.js)
     //       This is an O(n^2) solution wrt to the number of lines
     //       I opt to use this approach because it's simple, and it is unlikely that the number of elements in `lineNumbersAndRanges` will be large
     //       There is possible room for improvement for a more efficient algo that is O(n).
-    const inRange = lineNumbersAndRanges.some(([start, end]) => {
-      if (start && end) {
-        return currentLineNumber >= start && currentLineNumber <= end;
+    // Edit (28/8/2020): I changed the approach from using some() to a simple for-loop. It is still O(n^2).
```
ang-zeyu, line 158: let's just remove the long comment from `Note: The algorithm ...`

this area is hardly a bottleneck in performance

ang-zeyu (commented):
@ryoarmanda I saw that you tagged it ready for review but the `[wip]` tag is still there; If it's still a wip do ignore these where applicable:

- Code quality can be generally improved with some processing model(s) / class(es). (e.g. `ruleComponents`)
- Should we use `[:]` to highlight the entire line instead? @damithc @ryoarmanda 
  Reason: since we are going with stripping trailing/leading spaces by default (or not?)
  Also, when we have `[2:5]` eventually the `2` would be counted from the start of the line, including whitespaces. (but this may actually be undesirable versus counting from the first non-whitespace character, which should be easier for the author)

packages/core/src/lib/markdown-it/index.js
ryoarmanda: Ah, okay, will do

packages/core/src/lib/markdown-it/index.js
ryoarmanda: Oh yes, that would be nicer. I'll create a separate class file for the rule components and its helpers (e.g. `isLineSlice`)

ryoarmanda:
Thanks for the review @ang-zeyu 👍 

I'm still marking it WIP as I haven't updated the docs yet. Will remove it once it's done and have addressed the reviews.

> Should we use [:] to highlight the entire line instead?

Ah so basically swapping the behaviour of e.g. `2-7` and `2[:]-7` right? I think it's a nice idea but it depends on what the expected common usage would be. For me personally I like to lean more into text-only highlighting as the default, and so reserving `[:]` just for line highlighting for convenience sake. But also would like to know your input as well @damithc

ang-zeyu:
> Should we use [:] to highlight the entire line instead? @damithc @ryoarmanda
> Reason: since we are going with stripping trailing/leading spaces by default (or not?)
> Also, when we have [2:5] eventually the 2 would be counted from the start of the line, including whitespaces. (but this may actually be undesirable versus counting from the first non-whitespace character, which should be easier for the author)

any thoughts? @damithc

damithc:
I think the first thing to decide is if slice positions should include leading spaces or not. 

pros of omitting leading spaces from counting: easier to count for authors

cons: not possible to start highlighting somewhere in the middle of leading spaces  e.g., if I wanted to show how line wrapping requires 8 spaces for indentation, I may want to highlight the 8 leading spaces in a line to illustrate that.

As this is an advance feature not expected to be used frequently, perhaps we choose the more flexible option so that even rare cases become possible?

Regarding the use of [:], the advantage of using it to highlight the trimmed line (not the full line) is that the current syntax remains the same. The `[s:e]` syntax comes into play only if the user wants to do more advanced stuff with highlighting. This is assuming full line highlighting is going to remain the default.

ang-zeyu:
> I think the first thing to decide is if slice positions should include leading spaces or not.
> 
> pros of omitting leading spaces from counting: easier to count for authors
> 
> cons: not possible to start highlighting somewhere in the middle of leading spaces e.g., if I wanted to show how line wrapping requires 8 spaces for indentation, I may want to highlight the 8 leading spaces in a line to illustrate that.
> 
> As this is an advance feature not expected to be used frequently, perhaps we choose the more flexible option so that even rare cases become possible?

Agree here

> Regarding the use of [:], the advantage of using it to highlight the trimmed line (not the full line) is that the current syntax remains the same. The `[s:e]` syntax comes into play only if the user wants to do more advanced stuff with highlighting. This is assuming full line highlighting is going to remain the default.

Wasn't #1311 to change the default behaviour of full line highlighting? 😮 

In python (im assuming that's where the inspiration for this came from) omitting either `s` / `e` would slice all the way to the start / end, might be small plus to follow this too. (so if you want full line highlighting it would be `[:]`)

damithc:
> Wasn't #1311 to change the default behaviour of full line highlighting? 😮
> 
> In python (im assuming that's where the inspiration for this came from) omitting either `s` / `e` would slice all the way to the start / end, might be small plus to follow this too. (so if you want full line highlighting it would be `[:]`)

Oh, right :-p
OK, I'm sold. Let's do that.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
```diff
@@ -0,0 +1,57 @@
...
+  
+  isSlice() {
+    return Array.isArray(this.components)
+      && this.components.length === 3
+      && this.components.every(Number.isInteger);
+  }
+  
+  isUnboundedSlice() {
+    return this.isSlice()
+      && this.components[1] === -1
```
ang-zeyu, line 23: let's remove the use of magic numbers in favour of named instance properties, so we don't need to refer back to the usage to understand which is which

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+  
+  offsetLines(offset) {
+    this.rule.map(comp => comp.offsetLines(offset));
+  }
+  
+  shouldApplyHighlight(lineNumber) {
+    const compares = this.rule.map(comp => comp.compareLine(lineNumber));
+    if (this.isLineRange()) {
+      return (compares[0] <= 0) && (compares[1] >= 0);
```
ang-zeyu, line 42: could the magic numbers be removed?

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,82 @@
...
+     */
+    this.rule = rule;
+  }
+
+  isLineNumber() {
+    return this.rule.length === 1 && this.rule[0].isNumber()
+  }
+  
+  isLineSlice() {
+    return this.rule.length === 1 && this.rule[0].isSlice();
```
ang-zeyu, line 16: Since its used only in `applyHighlight` and is rather simple, perhaps we could just use a named constant instead? (ignore if this is going to be removed when dealing with the below)

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
ang-zeyu: but if this is a necessary part of the algorithm, maybe you could use some named constant declarations here, which are self-documenting

`const whatIsThis = compares[0]`

ang-zeyu (commented):
Nice work extracting out the logic, let's try to remove the use of magic numbers next too:

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
ryoarmanda: Yeah, `compareLine` is basically emulating what Java's `compareTo` does, returning either a positive integer, zero, or a negative integer depending on whether the given line number is before, at, or after the component's line number. So I guess it is inevitable to compare whether it's `<= 0` like the above to ensure that the line number is within the bounds.

I'll just name these checks as constant declarations.

docs/userGuide/syntax/code.mbdf
```diff
@@ -53,11 +53,27 @@ function add(a, b) {
 
 
 ##### Line highlighting
-To highlight lines, add the attribute `highlight-lines` with the line numbers as value, as shown below. You can specify ranges or individual line numbers.
+To highlight lines, add the attribute `highlight-lines` as shown below.
+
+You can specify highlighting in many different ways, depending on how you want it to be. There are two main variants:
+
+##### Text-only highlighting
```
ang-zeyu, line 60: could try super / regular bold if `h6` dosen't work too well

docs/userGuide/syntax/code.mbdf
```diff
@@ -53,11 +53,27 @@ function add(a, b) {
...
+If you wish to highlight a full line (including whitespaces) or ranges of it, you can leverage MarkBind's own _line-slice_ syntax. Line-slices are in the form of `lineNumber[:]`, e.g. `2[:]`.
+    
+This variant's format is very similar to the previous, but instead use line-slices rather than line numbers.
+
+For ranges, you only need to use line-slices on either ends.
 
 <include src="codeAndOutputCode.md" boilerplate >
 <variable name="code">
-```java {highlight-lines="2,4,6-8"}
+```java {highlight-lines="1,3[:],6-8,10[:]-12"}
```
ang-zeyu, line 76: to clarify, what happens if its `10-12[:]`?

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,73 @@
+const { HighlightRuleComponent } = require('./HighlightRuleComponent.js');
+
+class HighlightRule {
+  constructor(rule) {
+    /**
+     * @type Array<HighlightRuleComponent>
+     */
+    this.rule = rule;
```
ang-zeyu, line 8: let's use plural for arrays

or `ruleComponents` perhaps

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+  
+  isLineRange() {
+    return this.rule.length === 2;
+  }
+  
+  static parseRule(ruleString) {
```
ang-zeyu, line 15: hmm is there a pattern to the organization of methods?

maybe we could shift this and the `components` counterpart below right below their constructors since its also somewhat of a constructor

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
```diff
@@ -0,0 +1,53 @@
+const LINESLICE_REGEX = new RegExp('(\\d+)\\[(\\d*):(\\d*)]');
+
+class HighlightRuleComponent {
+  constructor(lineNumber, isSlice, bounds) {
+    this.lineNumber = lineNumber;
```
ang-zeyu, line 5: nice cleanup here 👍

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,73 @@
...
+    const components = ruleString.split('-').map(HighlightRuleComponent.parseRuleComponent);
+    return new HighlightRule(components);
+  }
+  
+  offsetLines(offset) {
+    this.rule.map(comp => comp.offsetLineNumber(offset));
```
ang-zeyu, line 21: `forEach`

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -0,0 +1,73 @@
...
+    
+    if (isLineSlice) {
+      // TODO: Implement slice-index based highlighting
+      return HighlightRule._highlightWholeLine(line);
+    }
+    
+    return HighlightRule._highlightTextOnly(line);
+  }
+
+  static _splitCodeAndIndentation(codeStr) {
```
ang-zeyu, line 54: let's keep related methods together; i.e. below `_highlightTextOnly`

ang-zeyu (commented):
Mostly looks good now 👍 

Just a couple more nits:

docs/userGuide/syntax/code.mbdf
ryoarmanda: It does the same thing as `10[:]-12` at the moment, highlights full-line.

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
ryoarmanda: I wrote it in the order of usage flow: parsing of the rule, post-parse processing (offset line numbers), test for application, and lastly the application of the rule itself. I feel like this way is intuitive if you want to read the whole class and understand the sequence of process.

Although, the structure check like `isLineSlice` (and `isUnboundedSlice` on the components side) does not fit too well on this scheme. I put it where it was now (on top of everything, after the constructor) if any developers want to understand what are the possible "variants" of the structure possible for this class before reading deep into the class methods.

Not sure if this is the best way, I don't have a strong preference really. What do you think?

ang-zeyu: ok, thanks for the explanation!

How about shifting `isLineRange` to the bottom at least then? Then we can keep the constructors together (parseRule and constructor) while still fitting to your scheme. Likewise for the `Component` version

ryoarmanda: Okay then :+1:

ang-zeyu (commented):
Sorry, missed it out earlier

Let's add / modify some simple tests to `testCodeBlocks.md` in the test_site as well, so we can detect unintended changes in the future

ang-zeyu (approved):

PR #1337, opened 2020-08-29
[Hotfix] Extend live preview to layout scripts and stylesheets

### LINKED ISSUE #1229: Extend live preview to layout scripts and stylesheets
Steps:
1. serve a site locally 
1. modify a css file inside _markbind/layouts

Problem: the changes are not reflected by the live preview

KendrickAng:

• [ ] Documentation update
• [X] Bug fix
• [ ] New feature
• [ ] Other, please explain:

Fixes #1229 , where live reload wasn't working with changes in the _markbind/layout folder.

Some digging around showed that live-server was checking the `_site/markbind/layouts` folder for changes (not too sure why, if anyone knows please share - for my own knowledge):
image

However, currently the layouts folder is being incorrectly copied to _site/_markbind/layouts, when it should be _site/markbind/layouts.
This seems to be because this.layoutsAssetPath is hardcoded to `markbind` (not too sure why):
```js
  addLayoutScriptsAndStyles() {
    this.asset.layoutScript = path.join(this.layoutsAssetPath, this.layout, 'scripts.js');
    this.asset.layoutStyle = path.join(this.layoutsAssetPath, this.layout, 'styles.css');
  }
```
This hotfix checks explicitly for the string `_markbind/layouts` when copying assets, and resolves it to `markbind/layouts`.

```js
const copyAssets = fileIgnore.filter(fileRelativePaths)
  .map((asset) => {
    // Hotfix: _markbind/layouts incorrectly copied to _site/_markbind/layouts (should be _site/markbind)
    const toAsset = asset.startsWith(path.join('_markbind', 'layouts'))
      || asset.startsWith('_markbind/layouts')
      ? asset.replace('_markbind', 'markbind')
      : asset;
    return fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, toAsset));
  });
```

**Testing instructions:**
1. Remove `_markbind` from the `site.json` ignores array (or anything that is to be tested)
2. Add `window.alert("Hello world") to `_markbind/layouts/default/scripts.js`
3. `Run markbind serve` and verify that the alert shows
4. Change `_markbind/layouts/default/script.js` to something else, say `window.alert("Goodbye world")
5. Verify the live reload is working

[Bugfix] Extend live preview to layout scripts and stylesheets

### DIFF
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -792,7 +792,14 @@ class Site {
     const fileIgnore = ignore().add(this.siteConfig.ignore);
     const fileRelativePaths = uniquePaths.map(filePath => path.relative(this.rootPath, filePath));
     const copyAssets = fileIgnore.filter(fileRelativePaths)
-      .map(asset => fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, asset)));
+      .map((asset) => {
+        // Hotfix: _markbind/layouts incorrectly copied to _site/_markbind/layouts (should be _site/markbind)
+        const toAsset = asset.startsWith(path.join('_markbind', 'layouts'))
+          || asset.startsWith('_markbind/layouts')
+          ? asset.replace('_markbind', 'markbind')
+          : asset;
+        return fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, toAsset));
+      });
     return Promise.all(copyAssets)
       .then(() => logger.info('Assets built'));
   }
@@ -816,9 +823,15 @@ class Site {
       // Scan and copy assets (excluding ignore files).
       this.listAssets(fileIgnore)
         .then(assets =>
-          assets.map(asset =>
-            fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, asset))),
-        )
+          assets.map((asset) => {
+            // Hotfix: _markbind/layouts wrongly copied to _site/_markbind/layouts (should be _site/markbind)
+            const toAsset = asset.startsWith(path.join('_markbind', 'layouts'))
+            || asset.startsWith('_markbind/layouts')
+              ? asset.replace('_markbind', 'markbind')
+              : asset;
+            return fs.copyAsync(path.join(this.rootPath, asset), path.join(this.outputPath, toAsset));
+          },
+          ))
         .then(copyAssets => Promise.all(copyAssets))
         .then(() => logger.info('Assets built'))
         .then(resolve)


ang-zeyu:
thanks @KendrickAng, I don't think this is the right solution though

> Some digging around showed that live-server was checking the _site/markbind/layouts folder for changes

I couldn't reproduce this, is `_markbind` is in the ignore array in the `site.json` of the user docs?

---

To answer your question, there are two types of files: source files (`.md/mbdf/html`, a page's dependencies) and assets (static files like images).
For assets a direct copy-paste is done when changes are detected by the file watcher.
In this issue the layout scripts and stylesheets simply need to be flagged as the page's dependencies, which would trigger a page rebuild and then the layout scripts / stylesheets gets copied to the correct directory

> However, currently the layouts folder is being incorrectly copied to _site/_markbind/layouts, when it should be _site/markbind/layouts.

This also occurs because of the above `_markbind`, and hence the scripts / stylesheets are being recognised as assets. So it just gets a direct copy paste from the relative source directory to the same relative output directory.

> [Bugfix] Extend live preview to layout scripts and stylesheets

The hard limit for commit message title is `72 - 8` (8 from the pr number github tags on), but the general recommended (soft) limit is 50 for the commit title to encourage succinct writing and a small commit scope

KendrickAng:
@ang-zeyu Yep! For some reason, _markbind is being ignored when building assets:
image

damithc:
I thought the ignore array in site.json is to tell markbind which files to ignore when _deploying_ the site, not when _building_ the site?

KendrickAng:
Maybe 'building assets' was a poor choice of words - I meant to say the `_markbind` directory (and its subdirectories) are currently set to be ignored when `_buildMultipleAssets` or `buildAssets` is called during live preview when the command `markbind serve` is run from the CLI. Please correct me if my understanding is wrong.

ang-zeyu:
> When attempting to reproduce, `_markbind` should be removed temporarily from the `ignore` array. Apologies for not clarifying. Could I check that you removed it before following the reproduction steps?

> I thought the ignore array in site.json is to tell markbind which files to ignore when deploying the site, not when building the site?

@KendrickAng @damithc per the UG, it's not used in the deploy process either (but rather `build` which must be run before `deploy`):
> An array of file patterns to be ignored when copying files to the generated site. By default, MarkBind will copy all the files as assets of the generated site.

The part that copies over the layouts scripts / stylesheets is a separate process; They aren't meant to be assets, but dependencies of the page.
Assets are things like images, static html pages not passed through markbind, that are copied over without any processing.

I think the confusion here is why the ignoring is exposed to the user; I'm not too sure of the context since this was a very early feature, but my take is that it provides flexibility when needed.
A general `ignore` config which should suit the majority of use cases is also provided in the template sites when you run `markbind init`, so the user dosen't have to set this up themselves, but can if they want to.

```
"ignore": [
    "_markbind/layouts/*",
    "_markbind/logs/*",
    "_site/*",
    "site.json",
    "*.md",
    "*.mbd",
    "*.mbdf",
    "*.njk",
    ".git/*"
  ],
```

damithc:
> I think the confusion here is why the ignoring is exposed to the user; I'm not too sure of the context since this was a very early feature, but my take is that it provides flexibility when needed.

From the user POV, this allows to specify which files should not be uploaded to the deployed site e.g., whether to upload the CNAME file need to be a user's decision.
Having said that, some things can be not-uploaded by default e.g., _markbind folder but such defaults might cause inconveniences too e.g., if we ignore all .md files by default, users will not be able to share an .md file via the web site.

Issue #654, opened 2019-02-03
PageNav: give a 'back to top' link


damithc:
Suggestion: in the pageNav, we can give a `back to top` link

We can even give an easy way for users to add a back-to-top link anywhere on a page.

damithc:
> Implementation wise is simple, we just have to add anchor text to the `page-nav` template. Although not sure where should we position the `back to top` anchor and whether or not the position should be static.

Perhaps a better solution to give a way to add a back-to-top to any page, not PageNav? 
1. It can be a small translucent icon that stick to the bottom of the page?
1. We can enable it via frontmatter

Chng-Zhi-Xuan:
Sounds good, but `front-matter` is starting to get cluttered. Since this is a small component, would implementing it via a variable or Vue be better? (Although we need to check there should only be 1 such component on every page).

damithc:
> Sounds good, but `front-matter` is starting to get cluttered. Since this is a small component, would implementing it via a variable or Vue be better? (Although we need to check there should only be 1 such component on every page).

Sure. As it is a page-level decoration, frontmatter seems like the right place though. In fact, that applies to TopNav too. It's more consistent that way. The page body can be more for content and not decorations.

tlylt:
Hello, I think this issue has been resolved in #1338, perhaps it was not closed automatically by that PR as the description did not use the 'fixes xxx' syntax.

jonahtanjz:
Yup this has been resolved. Thanks for checking on this @tlylt 👍

PR #1338, opened 2020-08-30
Create "Scroll-to-Top" button

wxwxwxwx9:

• [x] New feature
This PR addresses issue #654 by adding a Scroll-to-Top button on every page rendered. 

This would improve user experience by providing them a quick button to scroll to the top of the page. 

When Page adds "Content Wrapper" to pageData, I introduced another step to add a Scroll-to-Top button to the end of the pageData html. 

The "Scroll-to-Top" button can be found on the **bottom-right corner** of the window. 

I have also added some dynamic features to it: 
1) the button will only appear after user scrolls down, 
2) the button will disappear when the user has fully scrolled to the top, 
3) hover causes the button to darken from opacity 40% to 70%. 

**Browser View**:
Screen Shot 2020-08-30 at 3 55 29 PM

**Mobile View**:

Screen Shot 2020-08-30 at 3 56 44 PM

1) Code quality

2) Currently, the button will appear on both **mobile view** and **browser view**. There is currently an option for users to press the "Home" button on their keyboard to go to the top of the page. 

But I still feel that there is an incentive in making sure it appears on the **browser view** as well in case user doesn't know about the "Home" keyboard button feature. 

It also provides the user some minor convenience in that they most likely would already have their hand on the mouse, so they can just click on the "Scroll-to-Top" button, whereas it is less likely that they would have their hand on the keyboard (as compared to the mouse). 

3) The button design -- whether the current button is fine or we should opt for other designs. The current design was chosen from FontAwesome, specifically "fa-arrow-circle-up". 

**Testing instructions:**

1) ```markbind serve -d``` on an appropriate Markbind project folder (e.g. the current user guide)
2) Scroll down from the top, as the scroll comes to a halt, you should see the "Scroll-to-Top" button appear
3) After clicking on the button, your screen should scroll to the top of the page **smoothly** and the button should disappear

Note: Remember to use the ```-d``` option when you serve the project (```markbind serve -d```).  

Implement Scroll-to-Top button on every rendered page


damithc:
How does it look like?

wxwxwxwx9:
Hi Prof. Damith, I have updated the post with the UI screenshots.

damithc:
> Hi Prof. Damith, I have updated the post with the UI screenshots.

Thanks. Looks nice.

ang-zeyu (commented):
some quick comments:

> I have also added some dynamic features to it:

neat! left some suggestions about these below

> But I still feel that there is an incentive in making sure it appears on the browser view as well in case user doesn't know about the "Home" keyboard button feature.

I still think this should be a mobile-only thing

Its really very very rare for websites to have this in non-mobile view. Even when they do, out of habit, I often find myself just pressing `home` or using the scrollbar. Since the button appears on-scroll here, it can also become somewhat distracting to the reader.

For long pages (where this feature has the most value) the page would also very likely have headings already, and hence a page nav. (you can go back to the top with the first heading as well)

damithc:
> I still think this should be a mobile-only thing

Let's leave it in first and see how. I think it depends on personal habits too. I rarely use the Home button to go to the top. In some laptops, the home button is not easy to access as one has to use Fn + Up buttons in place of Home button.

ang-zeyu:
> > I still think this should be a mobile-only thing
> 
> Let's leave it in first and see how. 

ok

> I think it depends on personal habits too. I rarely use the Home button to go to the top. In some laptops, the home button is not easy to access as one has to use Fn + Up buttons in place of Home button.

that's true, there is the page nav still though, unless there's no heading near the top

packages/core-web/src/index.js
```diff
@@ -89,6 +89,31 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
+  if (document.body.scrollTop > 20 || document.documentElement.scrollTop > 20) {
+    scrollTopButton.style.display = 'block';
+  } else {
+    scrollTopButton.style.display = 'none';
+  }
+}
+
+function initDisplayScrollTopButton() {
+  let timer;
+  window.addEventListener('scroll', () => {
```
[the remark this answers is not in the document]
wxwxwxwx9: Hi Zeyu! Thanks for reviewing my code. 

The timeout is actually to delay the call to the function displayScrollTopButton(), meaning that the call to the function will only be made if user has stopped scrolling for more than 100ms. I think this would give significant improvement to the performance but correct me if I'm wrong (quite inexperienced in performance-related issues). 

I think that your idea "maybe we can introduce a timeout to hide the button after 1/2s instead? (to avoid obscuring the content)" is really good. I'll look into this!

packages/core-web/src/styles/scroll-top-button.css
```diff
@@ -0,0 +1,13 @@
...
+    position: fixed;
+    bottom: 20px;
+    right: 30px;
+    z-index: 99;
+    cursor: pointer;
+    opacity: 40%;
+}
+
+#scroll-top-button:hover {
+    opacity: 70%;
```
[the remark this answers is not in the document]
wxwxwxwx9: Alright! Do you have any recommendation for the styles? 

I'm thinking of just a "css default dark-grey color" for the :active state. What do you think?

packages/core/src/Page/index.js
```diff
@@ -1304,7 +1304,13 @@ class Page {
   static addContentWrapper(pageData) {
     const $ = cheerio.load(pageData);
     $(`#${CONTENT_WRAPPER_ID}`).removeAttr('id');
-    return `<div id="${CONTENT_WRAPPER_ID}">\n\n${$.html()}\n</div>`;
+    return `<div id="${CONTENT_WRAPPER_ID}">\n\n${this.addScrollToTopButton($.html())}\n</div>`;
+  }
+
+  static addScrollToTopButton(pageData) {
```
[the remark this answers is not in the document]
wxwxwxwx9: Yep, I agree. This would make the code a lot cleaner. Will make the revision. Thanks for the advice!

packages/core-web/src/index.js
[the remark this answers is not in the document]
ang-zeyu: > The timeout is actually to delay the call to the function displayScrollTopButton()

Ahh ok; You can refer to [interaction observers](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to set up the scroll-button-appears-when-not-at-the-top behaviour in `displayScrollTopButton` if you're really keen on optimizing the performance, it may also end up being easier /clearer to implement.

But I'm fine with either since performance after the initial render isn't really an issue for a static site : -)

packages/core-web/src/styles/scroll-top-button.css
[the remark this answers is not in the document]
wxwxwxwx9: I've added an "opacity: 100%" for this so that when users click on it, they can get a feedback that the button is clicked! The opacity will transition from 40% to 100% from the moment they click, until the moment they release their mouse left click, then it will transition back to 40%.

packages/core-web/src/index.js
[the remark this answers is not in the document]
wxwxwxwx9: Hi Zeyu, I tested out your idea of "hiding the button after 1/2s" but I found it to be difficult for users to be able to use the "scroll-to-top" button. 

Essentially, I feel like they have to "chase" for the button before it disappears within the timeframe (if the timeout is too short). However, if the timeout is too long, the button will obscure the content on the user's screen. 

I feel that it is hard to decide on an appropriate timeout value, such that it strikes a balance. Thus, I decided to experiment with "lightening" the button instead. What I have in mind is that we will lighten the button to a point where it wouldn't quite obscure the user's screen. 

so, currently there are 4 states of opacity: 15% (lighten), 40% (normal), 70% (hover), 100% (active)

Whenever the user scrolls, the button will turn to 40% normal opacity. But when user stops scrolling, the button will turn to 15% lightened opacity after 1 second. 

I tested the above out and I think it works pretty well for me, but I would love to hear your opinion!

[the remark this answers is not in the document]
ang-zeyu: > Essentially, I feel like they have to "chase" for the button before it disappears within the timeframe

hadn't thought of that, that makes sense.

packages/core-web/src/styles/scroll-top-button.css
```diff
@@ -0,0 +1,19 @@
+#scroll-top-button {
+    --scroll-top-button-opacity: 40%;
```
ang-zeyu, line 2: would removing the variable be clearer? since its only going to be used in one element

ang-zeyu: may also be causing this (netlify preview, chrome)

scrolltopbutton

packages/core-web/src/styles/scroll-top-button.css
```diff
@@ -0,0 +1,19 @@
...
+
+    display: none;
+    position: fixed;
+    bottom: 20px;
+    right: 30px;
+    z-index: 99;
+    cursor: pointer;
+    opacity: var(--scroll-top-button-opacity);
```
ang-zeyu, line 10: let's `transition:` the opacity as well!

packages/core-web/src/index.js
```diff
@@ -89,6 +89,47 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
+  document.body.scrollIntoView({ block: 'start', behavior: 'smooth' });
+};
+
+function setScrollTopButtonOpacity(scrollTopButton, opacity = '40%') {
+  scrollTopButton.style.setProperty('--scroll-top-button-opacity', opacity);
+}
+
+function displayScrollTopButton(scrollTopButton) {
+  if (document.body.scrollTop > 20 || document.documentElement.scrollTop > 20) {
+    scrollTopButton.style.display = 'block';
```
ang-zeyu, line 102: would it make sense to move `setScrollTopButtonOpacity` in here instead? since setting opacity before `display: block` has no effect

then in `timer.lightenButtonTimer` you'd just set the opacity as well

packages/core-web/src/index.js
```diff
@@ -89,6 +89,47 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
+  } else {
+    scrollTopButton.style.display = 'none';
+  }
+}
+
+function triggerScrollTopButton(timers) {
```
ang-zeyu, line 108: although the method quite literally triggers the button timers, let's use more informative method names instead, e.g. `displayOrHideScrollButton`

packages/core-web/src/styles/scroll-top-button.css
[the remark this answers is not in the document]
ang-zeyu: per the above, this is not working as well =C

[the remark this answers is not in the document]
wxwxwxwx9: This was resolved by changing percentage values to its represented decimal values!

packages/core-web/src/styles/scroll-top-button.css
wxwxwxwx9: This was resolved by changing percentage values to its represented decimal values!

packages/core-web/src/index.js
```diff
@@ -89,6 +89,55 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
+  const timers = {
+    showOrHideButtonTimer: 0,
+    lightenButtonTimer: 0,
+  };
+  window.addEventListener('scroll', () => {
+    promptScrollTopButton(timers);
+  });
+}
+
+initScrollTopButton();
```
ang-zeyu, line 139: looks good!

one last suggestion: let's move all off the stuff you added here to another file, then import and call this method here

packages/core-web/src/styles/scroll-top-button.css
```diff
@@ -0,0 +1,25 @@
...
+    opacity: 0.4;
+    transition-property: opacity;
+    transition-duration: 0.25s;
+}
+
+#scroll-top-button.lighten {
+    opacity: 0.15;
+}
+
+@media (hover: hover) {
```
ang-zeyu, line 17: sorry missed this out, let's remove this, since we aren't doing any additional styling

packages/core-web/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  'extends': [
+    '../../.eslintrc.js',
+    'plugin:vue/recommended',
```
ang-zeyu, line 4: let's remove this, and the `vue/` rules

likely the browser environment needs to be enabled here, and the node environment disabled https://eslint.org/docs/user-guide/configuring#specifying-environments

packages/core-web/src/index.js
wxwxwxwx9: Moved them to scrollTopButton.js file!

packages/core-web/src/styles/scroll-top-button.css
wxwxwxwx9: This is to resolve hover opacity styling being "stucked" on mobile view; or rather, on any devices that do not support hovering capability. Such devices will cause CSS styling for hover to "stick". 

More details in this website:
[https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/](url)

packages/core-web/.eslintrc.js
wxwxwxwx9: Updated!

ang-zeyu: `node: true` should be switched off as well (correct me if its not merged in the way I'm thinking),
let's put `env` as the first key per the order in the root `eslintrc` too

packages/core-web/webpack.dev.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
ang-zeyu, line 1: is this due to `require`?

ang-zeyu: let's use `/* eslint-env node */` here instead https://eslint.org/docs/user-guide/configuring#specifying-environments

wxwxwxwx9: Hi Zeyu! It is due to module.exports and __dirname.

packages/core-web/webpack.build.js
```diff
@@ -6,6 +6,7 @@ const TerserPlugin = require('terser-webpack-plugin');
 const VueLoaderPlugin = require('vue-loader/lib/plugin');
 const config = require('./webpack.common.js');
 
+/* eslint-disable-next-line */
```
ang-zeyu, line 9:
```suggestion
/* eslint-env node */
```

ang-zeyu (approved):

ang-zeyu:
Ok looks good, thanks =)

I fixed one last `eslint-env node` since its just one occurence, hope you don't mind

Regarding the commit message, (requoting from a comment on another pr)

>The hard limit for commit message title is `72 - 8` (8 from the pr number github tags on), but ~the~ a general recommended (soft) limit is 50 for the commit title to encourage succinct writing and a small commit scope
>
>We don't use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) spec here too, so you can remove the `[Bugfix]`
>
>a good read https://chris.beams.io/posts/git-commit/ if you have the time 

shortened it as well since just a few words needed to be trimmed

damithc:
Good work @wxwxwxwx9 Thanks for shepherding this PR through @ang-zeyu

PR #1345, opened 2020-09-05
Refactor Page dependency checking

ang-zeyu:
• [x] Bug Fix
• [x] Other, please explain: code maintenance

- extract some repeated functionalities related to page dependency checking into reusable methods
- replace the misleading source file extension check to determine whether to build a page with these reusable methods
  - fixes a bug where the page rebuilding process would be run even if the file (which ends in one of the previously used source file extensions) is an asset, not a page

Refactor Page dependency checking

Checking for dependencies of a Page instance is done repeatedly using
interfacing with the page instance's properties.

During a file system event when serving a MarkBind site, the site's
source files are rebuilt if the file has a "source file extension".
This unnecessarily runs the source files rebuilding process even when
such files are used as assets.

Let's extract these page dependency checking into reusable methods as
such, and use these methods in place of the extension check.

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -129,7 +129,7 @@ program
     const addHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (fsUtil.isSourceFile(filePath) || site.isPluginSourceFile(filePath)) {
+        if (site.isFilepathAPage(filePath) || site.isDependencyOfPage(filePath)) {
           return site.rebuildSourceFiles(filePath);
         }
         return site.buildAsset(filePath);
@@ -141,7 +141,7 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (fsUtil.isSourceFile(filePath) || site.isPluginSourceFile(filePath)) {
+        if (site.isDependencyOfPage(filePath)) {
           return site.rebuildAffectedSourceFiles(filePath);
         }
         return site.buildAsset(filePath);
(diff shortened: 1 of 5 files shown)


packages/cli/index.js
```diff
@@ -129,7 +129,7 @@ program
     const addHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (fsUtil.isSourceFile(filePath) || site.isPluginSourceFile(filePath)) {
+        if (site.isDependencyOfPage(filePath)) {
```
ang-zeyu, line 132: In reference to https://github.com/MarkBind/markbind/pull/1337#issuecomment-683319939
@KendrickAng, perhaps the confusion was here

Currently we are determining whether to run the page rebuilding process based on whether the file change has a "source file extension". but there is no such thing, because even files with source file extensions can be **assets** (e.g. a raw `.html` file you don't want processed)

This PR refactors page dependency checking into reusable methods and uses these everywhere instead, which is hopefully clearer!

KendrickAng: Hmm.. so just to clarify:
1. A source file is one that is parsed by Markbind, and also marked for live reload (in the old version of the code).
2. An asset is one that isn't parsed by Markbind

What exactly is defined a `dependency` in markbind? Is it the file `_variables.md` or files inserted using `<include>`?
Or can I simply define it as all the other files that a particular file we are parsing in markbind need, in order to be parsed fully?

ang-zeyu: > A source file is one that is parsed by Markbind, and also marked for live reload (in the old version of the code).
> An asset is one that isn't parsed by Markbind

yes, rather processed, we mostly rely on self-patched external packages for parsing

> What exactly is defined a dependency in markbind?
> Or can I simply define it as all the other files that a particular file we are parsing in markbind need, in order to be parsed fully?

anything a page _references_, recursively, which includes both examples you mentioned

Issue #1194, opened 2020-04-11
Migrate from fs-extra-promise back to fs-extra


ang-zeyu:
**Is your request related to a problem?**

`fs-extra-promise` originally provided promisified versions of `fs-extra`'s methods.
It hasn't been updated in quite some time however, and documentation on it is more difficult to find than `fs-extra`.
 `fs-extra` has also been updated to provide promisified versions quite a while ago, and is still being actively maintained, with widely available documentation.

**Describe the solution you'd like**

Migrate all code and tests back to `fs-extra`.

**Describe alternatives you've considered**
na

**Additional context**
na

PR #1351, opened 2020-09-12
Migrate from fs-extra-promise back to fs-extra

KendrickAng:

• [X] Other, please explain:
This PR maintains existing features while removing dependency fs-extra-promise in favor of fs-extra, which is more regularly updated and maintained.

Address #1194 .

As posted originally by Zeyu:
fs-extra-promise originally provided promisified versions of fs-extra's methods.
It hasn't been updated in quite some time however, and documentation on it is more difficult to find than fs-extra.
fs-extra has also been updated to provide promisified versions quite a while ago, and is still being actively maintained, with widely available documentation.

Replaces all fs-extra-promise methods with their equivalent fs-extra methods.

Here is a brief document comparing the changes between `fs-extra` and `fs-extra-promise`:
https://docs.google.com/document/d/1-HCLbtx7GslshHjaZI9bIbVW79wGvB_UPWJZThgFLYI/edit?usp=sharing

In short, most difference in functionality would arise from `fs-extra-promise` using older versions of `fs-extra`, an example being that JSON in older versions was being written to files beautified automatically; this has to be specified in the current `fs-extra`.

```js

```

There are some relatively non-trivial changes in core\src\Page\index.js line 1116 and core\src\Page\index.js line 1111, please have a look at them and verify those changes are acceptable.

**Testing instructions:**
Serve the docs folder and verify that functionality has not changed.

Migrate from fs-extra-promise back to fs-extra


packages/core/src/Site/index.js
```diff
@@ -1269,7 +1270,8 @@ class Site {
           })),
       };
 
-      fs.outputJsonAsync(siteDataPath, siteData)
+      // TODO: figure out why the only usage of fse breaks when changed to outputJson.
+      fse.outputJsonAsync(siteDataPath, siteData)
```
KendrickAng, line 1274: @ang-zeyu I'm having some trouble at this part; changing this line to `fs.outputJson(siteDataPath, siteData)` causes `test_site/_site/siteData.json` to output the correct json, but incorrectly formatted. For example"
image

On the other hand, leaving this statement as it is outputs a correctly formatted json file:
image

Do you happen to know why this is so? I suspected usage of a JSON.stringify(.., null, 2) but couldn't find any. Or maybe it's some side-effect of another library?

ang-zeyu: I'm assuming the tests fail because `outputJson` is not beautified
Try the options here https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputJson.md

for 3rd party libraries, always go back to the documentation to double check the behaviour first if there are issues =P
For this pr especially, do check if there are obscure differences between the method pairs (e.g. `copy` vs `copyAsync`)

KendrickAng: noted! I'll be sure to note down the differences, if any

packages/core/src/Site/index.js
```diff
@@ -1269,7 +1270,9 @@ class Site {
           })),
       };
 
-      fs.outputJsonAsync(siteDataPath, siteData)
+      fse.outputJsonAsync();
```
ang-zeyu, line 1273: 👀 

let's remove all of it:
- there's a few redundant mocks for fs-extra-promise still as well;
- the walk-sync mock which relies on it is also redundant
- package.json entries, then updating `package-lock`

KendrickAng: 😓

KendrickAng:
I've added the following changes:
* removed the fs-extra-promise mock (since no tests rely on fs-extra-promise anymore)
* removed fs-extra-promise from package.json, regenerated package-lock.json with `npm run setup`
* changed walk-sync to rely on fs-extra itself rather than the fs-extra-promise mock

packages/core/src/Page/index.js
```diff
@@ -1110,11 +1110,10 @@ class Page {
       const srcPath = path.resolve(plugin._pluginAbsolutePath, assetPath);
       const srcBaseName = path.basename(srcPath);
 
-      fs.existsAsync(plugin._pluginAssetOutputPath)
-        .then(exists => exists || fs.mkdirp(plugin._pluginAssetOutputPath))
+      fs.ensureDir(plugin._pluginAssetOutputPath)
```
ang-zeyu, line 1113: 👍

ang-zeyu (approved):
> Here is a brief document comparing the changes between fs-extra and fs-extra-promise:
> https://docs.google.com/document/d/1-HCLbtx7GslshHjaZI9bIbVW79wGvB_UPWJZThgFLYI/edit?usp=sharing

Thanks for the thorough comparison!

This looks good 👍

---

Regarding the walk-sync mock, did removing it (and the `jest.mock` call) not work?
Since we have a `jest.mock('fs')` call in the same file which _should_ propagate to walk-sync

But it can be investigated (if it was even necessary when added) and dealt with at a later time 

---

> Address #1194

Small nit: you can use one of these when the pr completely resolves the issue 🙂
https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #1355, opened 2020-09-15
Add dark style for fenced code blocks

### LINKED ISSUE #1219: Code fences / inline code: provide a 'dark' mode
na

- provide a dark mode for code fences / inline code using `markdown-it-attrs`
  - maybe make it the default as well, since contrast is generally better in dark mode

na

**Additional context**
na

ryoarmanda:

Resolves #1219 

Fenced code blocks are presented with a light style. It would be better to include dark style as well, and provide user to toggle between the two (with dark style being the default).

- Added minified CSS `vs2015.min.css` from Highlight.js
- Added optional `site.json` config `highlightStyle` to indicate which style the fenced code blocks should be presented. The default is `"dark"`.
- Adjusted assets preparation to include either `github` or `vs2015` depending on the value in `site.json`
- Adjusted CSS for code block background, line highlighting, and print view.

Preview of the new dark mode style:

image

N/A

**Testing instructions:**
`npm run test` (to test) or `markbind serve -d` (to view the site)

Code fences are presented on a light background, and it is the only
style available.

Let's provide a dark style for the code block presentation and provide
user configuration to specify which styles to use.


packages/cli/test/functional/test_site/expected/bugs/index.html
```diff
@@ -13,7 +13,7 @@
   <link rel="stylesheet" href="../markbind/fontawesome/css/all.min.css">
   <link rel="stylesheet" href="../markbind/glyphicons/css/bootstrap-glyphicons.min.css">
   <link rel="stylesheet" href="../markbind/css/octicons.css">
-  <link rel="stylesheet" href="../markbind/css/github.min.css">
+  <link rel="stylesheet" href="../markbind/css/vs2015.min.css">
```
ang-zeyu, line 16: hmm, does this mean no light mode? =\

To clarify, what I meant is to use the light styles (`github.min.css`) during printing when dark mode is on (default) and under normal operation if dark mode is off (provide some way to toggle off dark mode)

e.g.

in site.json, maybe add a `style` key
```
{
  style: {
    codeTheme: "light" // implicitly dark by default
  }
}
```

> the problem shifts to us needing to provide the whole 'light' styling inside the rule.


regarding your earlier comment on duplication as well, you could try the `not print` operator https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries for the dark mode styles

ryoarmanda: Oh I thought you want to set it as default (like the only one) as well from what I read on the issue 😅  I'll add the toggle on `site.json`

docs/userGuide/siteJsonFile.md
```diff
@@ -19,6 +19,7 @@ Here is a typical `site.json` file:
   "faviconPath": "myfavicon.png",
   "titlePrefix": "FooBar Dev Docs",
   "theme": "bootswatch-cerulean",
+  "highlightStyle": "light",
```
ang-zeyu, line 22: should we do 

```
style: {
  codeTheme: "..."
  theme: ... // later pr
}
```

instead? so we can move `theme` inside there later on since there's now multiple style properties

ang-zeyu: `highlightStyle` sounds a little like the highlight text style we have, https://deploy-preview-1355--markbind-master.netlify.app/userguide/formattingcontents#text-styles, should we change this to something more explicit?

packages/core/src/Page/index.js
```diff
@@ -796,6 +796,14 @@ class Page {
     }
   }
 
+  adjustHighlightStyle() {
+    const $ = cheerio.load(this.content);
```
ang-zeyu, line 800: hmm, parsing the entire document for this seems pretty expensive (its current liberal use is also our main performance bottleneck); let's try to to move it to the main html processing area in `ComponentParser`

packages/cli/test/functional/test_site_templates/test_minimal/expected/index.html
```diff
@@ -13,7 +13,7 @@
   <link rel="stylesheet" href="markbind/fontawesome/css/all.min.css">
   <link rel="stylesheet" href="markbind/glyphicons/css/bootstrap-glyphicons.min.css">
   <link rel="stylesheet" href="markbind/css/octicons.css">
-  <link rel="stylesheet" href="markbind/css/github.min.css">
+  <link rel="stylesheet" href="markbind/css/vs2015.min.css">
```
ang-zeyu, line 16: nice, swapping the entire style sheet out should be more printer friendly

packages/core-web/src/styles/markbind.css
```diff
@@ -21,14 +21,21 @@ pre.hljs > code {
...
     border-radius: 5px;
     counter-reset: line;
 }
 
+pre > code.hl-light {
+    background: ghostwhite;
+}
+
+pre > code.hl-dark {
+    background: #1e1e1e;
```
ang-zeyu, line 36: I have no comments about the vs2015 theme choice, maybe we can get @damithc's opinion though

the background here does seem a little too dark though, propose: `#333333` (lightness +8%) if its not just me =P

ang-zeyu: the line number needs and heading background to be brightened as well
https://deploy-preview-1355--markbind-master.netlify.app/userguide/formattingcontents#syntax-coloring
https://deploy-preview-1355--markbind-master.netlify.app/userguide/formattingcontents#heading

ang-zeyu (commented):
Thanks! Some issues:

docs/userGuide/siteJsonFile.md
ryoarmanda: `style` looks more apt to consolidate all the style configurations for the site (I don't know how it will be in the future, we only have two style properties so far). I'll introduce that property, and also rename `highlightStyle` to `codeTheme`.

ryoarmanda: Also, a quick question, what do you think should happen when authors write an ineligible option to `codeTheme` (e.g. `"qwerty"`)? Should MarkBind halt or let it quietly discard the option and continue to use the default?

ang-zeyu: hmm haven't thought of validating site.json before, my preference here is to `logger.warn` the user softly and fall back to the dark theme since style isn't critical

for other properties like `baseUrl` that are we could throw exceptions (in a later pr perhaps)

ang-zeyu: if you're interested, you could add validation for all applicable properties together in a later pr as well, and just leave this one be for now :-)

packages/core-web/src/styles/markbind.css
ryoarmanda: Ah the background is actually identical to the default theme in VS Code so personally I've grown accustomed to it haha, the line numbering color is also similar. But `#333333` is also not bad as an alternative. Would like to know what @damithc prefers as well

damithc: What are my choices? I don't mind going with a scheme used by another popular tool.

ryoarmanda: For heading background, do we need to match it with the code theme? The styling for heading hasn't been affected so far (it's still the same as master), but if we're to match it with the theme, could be a bit tricky as well as I'm planning to target `'pre'` nodes in `ComponentParser` to apply the theme but the heading is outside that.

ryoarmanda: @damithc The theme choices are available on Highlight.js' demo here https://highlightjs.org/static/demo/ (there are loads of it, light and dark). The dark theme implemented in this PR is `Vs 2015`, but it's just my pick as it is very similar to VS Code which I'm very familiar with. You can check how it looks like on the links @ang-zeyu posted above.

Also, if you're okay with it, are you also okay with the current background color (re: `#1e1e1e` or `#333333`)?

damithc: Let's go with VS-like theme. Re background color, hard to decide without a side-by-side comparison in-action.

ryoarmanda: @damithc Okay :+1: Here's the comparison for the background colors (left `#1e1e1e`, right `#333333`):

image

damithc: PyCharm (seems to be in the middle):
image

So, something in the middle? But frankly, I have no strong objections to either of the three.

ryoarmanda: Oh yes, JetBrains IDEs use that background color. I think I'm going to use that instead, it's a good middle ground. Thanks!

@ang-zeyu for the line numbering color, I think I'll match it with the normal text color on the code block. For heading background, I'm not sure whether I should change it. The styling's still untouched from master, and IMO the current one is still okay-ish for dark, it kind of contrasts with the code block well (?). What's your opinion on this?

ryoarmanda:
I found a complication. Because the asset provided for Highlight.js styling is universal for all code highlighting in the site, inline code is also stylized that way. This causes unexpected presentation like this:

image

In light of this, should we also present inline code in dark theme, or should we reserve the theming only for the code blocks? 

If we choose the former, it's easier to act on (only need to adjust CSS for the inline code, will likely need to introduce a little bit of padding), but I don't know if this would be desirable to you guys. If we choose the latter, then the bad news is that I'd need to scrap the current approach of swapping the CSS according to theme (as we need two separate stylings happening at once now), and probably just pasting the whole dark theme CSS (<100 lines, with some slight selector adjustment) onto our MarkBind CSS.

Any suggestion for this? @ang-zeyu @damithc

packages/core-web/src/styles/markbind.css
ang-zeyu: >  for the line numbering color, I think I'll match it with the normal text color on the code block.

👍 

> For heading background, I'm not sure whether I should change it.

looks a little odd 👀 

Untitled

ang-zeyu:
> In light of this, should we also present inline code in dark theme, or should we reserve the theming only for the code blocks?
> 
> If we choose the former, it's easier to act on (only need to adjust CSS for the inline code, will likely need to introduce a little bit of padding), but I don't know if this would be desirable to you guys. If we choose the latter, then the bad news is that I'd need to scrap the current approach of swapping the CSS according to theme (as we need two separate stylings happening at once now), and probably just pasting the whole dark theme CSS (<100 lines, with some slight selector adjustment) onto our MarkBind CSS.

The former should be more consistent, maybe to distinguish it you could use a slightly lighter background color than code blocks (like the subtle difference in the current light style)

packages/core-web/src/styles/markbind.css
```diff
@@ -59,10 +58,25 @@ pre > code.hljs[heading] {
     display: block;
 }
 
-code > span.highlighted {
+code > span.highlighted-light {
```
ang-zeyu, line 61: actually, since you already havemutually exclusive stylesheets, how about moving all code related styles into `github.min.css` / `vs2015.css` (renaming appropriately and preserving license notices)?

should be easier to implement than requiring separate classes for light / dark

damithc:
> The former should be more consistent

I agree. If possible, apply it for both types.

packages/core-web/src/styles/markbind.css
damithc: > for the line numbering color, I think I'll match it with the normal text color on the code block.

Perhaps better for the line numbers to appear less prominent than normal text? They are two different things and the line numbers are less important, right?

ryoarmanda: > Perhaps better for the line numbers to appear less prominent than normal text? They are two different things and the line numbers are less important, right?

Hmm, now that you mention it, editors do stylize the line numbers to be a bit darker than normal text. Just found out that the color that we're using regularly (`#888888`) is very similar, almost identical to what VS Code uses (`#858585`). JetBrains IDEs uses approximately `#606366`, which is even darker than the previous two.

Personally, I don't see any problem with the current color, but I can also see if you guys think it should be brightened to make the line numbers more stand out. Any opinions? @ang-zeyu

damithc: It's less about the absolute color but more about how line numbers appear relative to normal code.

ang-zeyu: Sorry, missed the ping!

> Perhaps better for the line numbers to appear less prominent than normal text? They are two different things and the line numbers are less important, right?

> Personally, I don't see any problem with the current color, but I can also see if you guys think it should be brightened to make the line numbers more stand out. Any opinions?

ok, good point. let's keep the current line color then

damithc: > but I can also see if you guys think it should be brightened to make the line numbers more stand out.

I was suggesting line numbers should stand out _less_, not more, as they are less important than the code.

ang-zeyu: > Untitled

Like this image? (current) It is already slighty darker than the code text

damithc: > Like this image? (current) It is already slighty darker than the code text

Yes, that works. I thought the same color is being used (I saw something like that earlier in the thread).

ang-zeyu: ahhh no that was my original suggestion to @ryoarmanda 

> Hmm, now that you mention it, editors do stylize the line numbers to be a bit darker than normal text.

lets go with this instead, as you guys mentioned the line numbers shouldn't receive too much attention

ryoarmanda: Okay then, we'll use `#333333` for the line numbers, thanks all :+1:

packages/core-web/src/styles/markbind.css
ryoarmanda: Good idea, I'll try to move the theme-specific styling to their respective CSS files, but some will be kept in `markbind.css` for identical things like layouting and common colors (if any) so that it can be shared by both.

I'll also keep the license but add something like `(slightly modified for MarkBind)` to preserve credits

packages/core-web/src/styles/markbind.css
ang-zeyu: > Edit: oh wait, that is the original color! It's the same as the light style 😅

yes, we went full circle 😅

ryoarmanda:
@ang-zeyu @damithc here's the updated dark style, changes are on the heading and inline code styling. Reviews are welcome 

image

damithc:
> @ang-zeyu @damithc here's the updated dark style, changes are on the heading and inline code styling. Reviews are welcome

Looks OK to me 👍

ang-zeyu:
The heading background looks great 👍 

Are there specific languages in which the inline code styling would apply only?
It dosen't seem to be showing up on the netlify preview https://deploy-preview-1355--markbind-master.netlify.app/, also tried locally

some regressions with the light style too:
Untitled

ryoarmanda:
The inline styling do appear on the Netlify 

image

I'll handle the regression in a minute

ang-zeyu:
> The inline styling do appear on the Netlify

it applies only when you specify a language;
https://deploy-preview-1355--markbind-master.netlify.app/userguide/formattingcontents#horizontal-rules

per the code here https://github.com/MarkBind/markbind/blob/master/packages/core/src/lib/markdown-it/index.js
let's move the `token.attrSet('class', `hljs inline ${lang}`)` upward to apply in both cases

ang-zeyu:
> Though it only applies to syntax-colored variant of inline code.

Let's keep it consistent! A little strange for it to go back and forth between light / dark =X

ryoarmanda:
> let's move the `token.attrSet('class', hljs inline ${lang})` upward

As the `lang` variable will be `null` if the language is not specified, I think it would be cleaner to omit it entirely in that case. So probably I would just make another `token.attrSet` in the else branch but with class just `hljs inline` only.

packages/core-web/asset/css/github.min.css
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+.hljs{display:block;overflow-x:auto;padding:.5em;color:#333;background:#f8f8ff}.hljs-comment,.hljs-quote{color:#998;font-style:italic}.hljs-keyword,.hljs-selector-tag,.hljs-subst{color:#333;font-weight:700}.hljs-literal,.hljs-number,.hljs-tag .hljs-attr,.hljs-template-variable,.hljs-variable{color:teal}.hljs-doctag,.hljs-string{color:#d14}.hljs-section,.hljs-selector-id,.hljs-title{color:#900;font-weight:700}.hljs-subst{font-weight:400}.hljs-class .hljs-title,.hljs-type{color:#458;font-weight:700}.hljs-attribute,.hljs-name,.hljs-tag{color:navy;font-weight:400}.hljs-link,.hljs-regexp{color:#009926}.hljs-bullet,.hljs-symbol{color:#990073}.hljs-built_in,.hljs-builtin-name{color:#0086b3}.hljs-meta{color:#999;font-weight:700}.hljs-deletion{background:#fdd}.hljs-addition{background:#dfd}.hljs-emphasis{font-style:italic}.hljs-strong{font-weight:700}.hljs>span.highlighted{background:#e6e6fa}.hljs.inline{background:#f8f8f8}.hljs:not(.inline){border:1px solid #c8c8c8;border:1px solid rgba(200,200,200,.3)}.code-block-heading{background:#f2f2ff;color:#8787a5}
```
ang-zeyu, line 1: pink highlight is missing:
Untitled

 let's rename these files as well https://github.com/MarkBind/markbind/pull/1355#discussion_r492646202, since they contain custom styles now too.

ang-zeyu: wondering if we should separate and format our styles as well
I believe the extra whitespace is hardly an issue over being able to figure out which are our custom additions

packages/core/src/lib/markdown-it/index.js
```diff
@@ -87,7 +87,7 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
     // counter is incremented on each span, so we need to subtract 1
     token.attrJoin('style', `counter-reset: line ${startFromZeroBased};`);
   }
-  
+
```
ang-zeyu, line 90: let's keep unrelated changes to another PR (our custom renderers should probably be moved to entirely separate linted file)

packages/core/src/lib/markdown-it/index.js
```diff
@@ -149,13 +149,15 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+  const inlineClass = `hljs inline`;
 
   if (lang && hljs.getLanguage(lang)) {
-    token.attrSet('class', `hljs inline ${lang}`);
+    token.attrSet('class', `${inlineClass} ${lang}`);
     return '<code' + slf.renderAttrs(token) + '>'
       + hljs.highlight(lang, token.content, true).value
       + '</code>';
   } else {
+    token.attrSet('class', `${inlineClass}`);
```
ang-zeyu, line 160: `inlineClass`

packages/core/src/lib/markdown-it/index.js
ryoarmanda: Are you referring to the whitespace changes here? Looks like WebStorm fixed it out automatically for some reason. Let's see if I can re-add these back

ang-zeyu: yup

packages/core-web/asset/css/github.min.css
ryoarmanda: I have to add a `no-lang` class here to detect this. Okay then.

ryoarmanda: > wondering if we should separate and format our styles as well

I agree, it would be easier to read. How about we move these CSS files into `core-web/src/styles` folder instead, alongside `markbind.css`? Reason is that we have already customized it, so it's not purely a third-party asset anymore, like the ones in the current `asset` folder.

I think if we do that, and (if I'm not wrong) webpack will build all CSS inside that folder, then we can just store the un-minified version of it, and we'll pick and choose which one to include during `copyCoreWebAsset` in `Site/index.js`.

ang-zeyu: > I have to add a `no-lang` class here to detect this. Okay then.

okay 

alternatively you could do `[class="hljs inline"]`, but I don't there's much of an issue in either case

> How about we move these CSS files into core-web/src/styles folder instead, so that it's now alongside markbind.css?

Yes that'd be really nice, your understanding is correct too. (also you get the hot reload package when modifying the styles)

The issue at the moment is that there is no webpack configuration-compilation going at site-building time, which requires some amount of groundwork to implement. (currently its only updated at every release)
I'm ok if you want to take this route, but it'll likely take a substantial amount of time to do; Leaving it as `assets` for the moment may be a more feasible approach

We also have another issue https://github.com/MarkBind/markbind/issues/903 which definitely requires this (configuring webpack compilation at site building time)

ryoarmanda: Hmm, I see. Well, let's put a pin on this and revisit in the future when the webpack configuration is able to set this up.

Pushing the latest changes now.

packages/cli/test/functional/test_site/expected/markbind/css/codeblock-light.min.css
```diff
@@ -0,0 +1 @@
+.hljs{display:block;overflow-x:auto;padding:.5em;color:#333;background:#f8f8ff}.hljs-comment,.hljs-quote{color:#998;font-style:italic}.hljs-keyword,.hljs-selector-tag,.hljs-subst{color:#333;font-weight:700}.hljs-literal,.hljs-number,.hljs-tag .hljs-attr,.hljs-template-variable,.hljs-variable{color:teal}.hljs-doctag,.hljs-string{color:#d14}.hljs-section,.hljs-selector-id,.hljs-title{color:#900;font-weight:700}.hljs-subst{font-weight:400}.hljs-class .hljs-title,.hljs-type{color:#458;font-weight:700}.hljs-attribute,.hljs-name,.hljs-tag{color:navy;font-weight:400}.hljs-link,.hljs-regexp{color:#009926}.hljs-bullet,.hljs-symbol{color:#990073}.hljs-built_in,.hljs-builtin-name{color:#0086b3}.hljs-meta{color:#999;font-weight:700}.hljs-deletion{background:#fdd}.hljs-addition{background:#dfd}.hljs-emphasis{font-style:italic}.hljs-strong{font-weight:700}.hljs>span.highlighted{background:#e6e6fa}.hljs.inline{background:#f8f8f8}.hljs.inline.no-lang{color:#e83e8c}.hljs:not(.inline){border:1px solid #c8c8c8;border:1px solid rgba(200,200,200,.3)}.code-block-heading{background:#f2f2ff;color:#8787a5}
```
ang-zeyu, line 1: ok looks good functional wise

let's clealy break off our custom styles as well to facilitate future changes

ryoarmanda: Ok, I think I'll break it off to another line (but still in minified form), and add a short comment to indicate these are MarkBind-customized styles

ang-zeyu (approved):
ok looks good, thanks for implementing this staple : )

Issue #1353, opened 2020-09-14
Deprecate <import> and <variable>


ang-zeyu:
**Is your request related to a problem?**

Opening an issue to track: https://github.com/MarkBind/markbind/pull/751#issuecomment-691595708

**Describe the solution you'd like**

- [x] allow `<include><variable>...</variable></include>` to override `{% set / import %}`
- [x] re-implement external json variables using a custom nunjucks tag
- [x] add deprecation warnings after the above for `<import>` and `<variable>`
- [x] add live preview for nunjucks sources (anytime)
- [ ] finally, remove `<import>` and `<variable>` (v3.0)

**Additional context**

See discussion here https://github.com/MarkBind/markbind/pull/751#issuecomment-691595708

ang-zeyu:
(1) and (4) have been merged in

for (2), I noticed that the current requirement of allowing variables in json to also reference other variables makes the re-implementation considerably more complex, but not impossible

I'm thinking that external variables should not be allowed to do so, since
-  the use case for sourcing external variables usually is some dataset that is _independent_ of the markbind project in the first place
- having nunjucks syntax in such standard files may not be rather user friendly - it requires the user to consider the placement of tag that sets the external variable carefully; any other variables declared before it in the page; any `{% import %}`s made, ..., which could all affect the result of the variables extracted from the file

Since its a new feature / replacement of an existing feature, we could do away with this (existing `<variable type="json">`s would still work till we deprecate that as well)

What do you think? @damithc 
Are there worthwhile cases where we want nunjucks syntax to appear in such external files? (`.json`, `.csv`, `.xlsx`)

damithc:
> What do you think? @damithc
> Are there worthwhile cases where we want nunjucks syntax to appear in such external files? (`.json`, `.csv`, `.xlsx`)

 I guess we can live without it. So, go ahead. 👍

ang-zeyu:
will be getting a new release out tonight with a few hotfixes (hopefully that's stable =P)

since that's all on the deprecation plate, I will be fast tracking the code to v3.0 from now on as well since full deprecation of this is necessary for the aims.

It'll mostly be backend changes requiring / including deprecation of some frontend features:
- Much simpler architecture
- Much improved performance ~2x
- Deprecate and change some things about plugins to facilitate the above
- Deprecate old syntatic features (`<modal title>` -> `<modal header>`)
- ...

PR #1369, opened 2020-09-27
Override nunjucks variables with include variables

ang-zeyu:

• [x] Bug fix?
• [x] Other, please explain: Fixes something that should have been, but was worked around

Part of #1353 

- Allow MarkBind variables (e.g. `<include><variable>...</variable></include>`, `variables.md`) to override `{% set / import %}`

- Patch various nunjucks methods
   - put programatically provided variables (markbind variables) somewhere special
   - small modifications to nunjucks lookup functions to lookup this special place first

- should we fork nunjucks instead? patching seems like the obvious lazy option here but has its benefits:
   - pros: 
       - easier to update when nunjucks updates
       - easier to run nunjucks' own tests
   - cons:
       - the diff seems large but the number of lines changed in the patch is really only <20.
       - extra work needed to develop cross-repo (`npm link`)
       - looks like nunjucks 4.0 is coming in the next few months; the implementation may need to be changed substantially (maybe we can delay forking till then)
       - I don't see needing to make anymore changes to nunjucks code apart from this

**Testing instructions**
all tests in nunjucks repo passes with these same changes

Override nunjucks variables with include variables


ang-zeyu:
@marvinchin any thoughts on if we should fork or leave this as a patch?

Issue #824, opened 2019-04-07
Crawler to check for broken links (Feature Suggestion)


sijie123:
* **MarkBind Version:**
MarkBind v2.1.0

While working on #823, I realise that I'm somehow constantly involved in a process of making changes (to the docs and the parser) that changes the behaviour of existing links. During this process, it's likely that some links become broken, but left unnoticed because there are simply too many links to check manually.

In addition to having a rigorous unit/system test, one avenue where I often discover bugs is by building existing known, good, versions of large MarkBind projects such as CS2103T website or even our own MarkBind docs.

I wonder if it's feasible and/or worthwhile to write a small crawler to recursively check a site for any broken links. I would imagine that this crawler would have benefits for site authors too, since they can use it to quickly check if they have mis-referenced anything.

damithc:
Good idea. I sometimes use a tool called Xenu to check for broken links. I wonder if we can find a tool that can be integrated into CI so that the check is done every time. Could be useful for the MarkBind as well as sites using MarkBind.

ang-zeyu:
Raising this to medium as its quite a staple.

This can be done within the existing code for performance though, no need for a separate tool / crawler stage.

PR #1371, opened 2020-10-03
Fix layout insertion cwf

wxwxwxwx9:
• [x] Bug fix

As I was working on issue #824, I discovered a bug where the current working file (cwf) is not set for layout insertions. This PR fixes the bug. 

Wrapped the return value of the layout insertion functions with a `div` containing the correct `cwf`. 

**Testing instructions:**
`npm run test`

Fix layout insertion cwf


packages/cli/test/functional/test_site/expected/testThumbnails.html
```diff
@@ -46,8 +46,8 @@
           <thumbnail src="http://placehold.it/100x300" circle></thumbnail>
         </p>
         <p><strong>An actual image</strong>
-          <thumbnail circle size="300" src="/test_site/images/deer.jpg"></thumbnail>
-          <pic src="/test_site/images/deer.jpg" height="300"></pic>
+          <thumbnail circle size="300" src="/test_site/_markbind/layouts/default/images/deer.jpg"></thumbnail>
+          <pic src="/test_site/_markbind/layouts/default/images/deer.jpg" height="300"></pic>
```
wxwxwxwx9, line 50: Not sure why the test updates are broken!

packages/core/src/Page/index.js
```diff
@@ -586,7 +586,7 @@ class Page {
 
     const renderedFooter = this.pageConfig.variableProcessor.renderSiteVariables(this.pageConfig.sourcePath,
                                                                                  footerContent, pageSources);
-    return `${pageData}\n${renderedFooter}`;
+    return `<div data-included-from="${this.footer}">${pageData}</div>\n${renderedFooter}`;
```
ang-zeyu, line 589: 👀

ang-zeyu: let's add a test or two too: any relative link in one of the test_site layouts

wxwxwxwx9: Ah, this was very careless of me. Thanks for spotting this!

packages/core/src/Page/index.js
```diff
@@ -567,7 +567,7 @@ class Page {
 
     const renderedHeader = this.pageConfig.variableProcessor.renderSiteVariables(this.pageConfig.sourcePath,
                                                                                  headerContent, pageSources);
-    return `${renderedHeader}\n${pageData}`;
```
ang-zeyu, line 570: the newlines are important to ensure the different segments get their markdown rendered correctly

ang-zeyu (commented):
thanks! let's add tests too (something like relative links that relies on cwf would do)

packages/core/src/Page/index.js
wxwxwxwx9: Ah, I see! Have added the newline back to the code.

wxwxwxwx9:
Hi Zeyu, I have added a test for the bug fix. Let me know if it's okay and whether there are other things that I should add! Thanks so much :-)

packages/cli/test/functional/test_site/testRelativeLink.md
```diff
@@ -0,0 +1 @@
+**Relative Link Test** This is a relative Intra-Site link (see [Tags](testTags.md))
```
ang-zeyu, line 1: let's move this to one of the layout files instead, which is the use case you fixed here

wxwxwxwx9: Updated the test! But I'm not sure if it is the most appropriate. Let me know if I should make more changes!

packages/cli/test/functional/test_site/_markbind/layouts/default/head.md
```diff
@@ -2,3 +2,5 @@
 <meta name="default-head-top">
 </head-top>
 <meta name="default-head-bottom">
+
+**Relative Link Test** This is a relative Intra-Site link (see [Tags](../../../testTags.html))
```
ang-zeyu, line 6: you want `header.md`, the similar sounding `head.md` is for inserting things into the `<head>` tag =P

ang-zeyu (approved):
_Lgtm_

Issue #1278, opened 2020-07-14
Make the copy button more subtle


damithc:
Currently, the button attracts too much attention.
![image](https://user-images.githubusercontent.com/1673303/87439333-3bf09900-c623-11ea-8fe9-801717683043.png)

Can we make it more subtle?

KendrickAng:
@ang-zeyu Is this issue still relevant? I tried to find the [cs2103/T website](https://nus-cs2103-ay2021s1.github.io/website/se-book-adapted/projectDuke/index.html#a-moreoop) to play with and noticed that the copy button no longer exists:
![image](https://user-images.githubusercontent.com/38814428/95011482-979b5300-0663-11eb-8c60-a0da9328a024.png)

ang-zeyu:
I think it (plugin) may have been toggled off there, you can refer to the ug for how to toggle it on https://markbind.org/userGuide/usingPlugins.html#plugin-codeblockcopybuttons

KendrickAng:
I was reading an article about button UI https://uxdesign.cc/ui-cheat-sheets-buttons-7329ed9d6112 and thought that a line button might be what we want here; what do you think?

![image](https://user-images.githubusercontent.com/38814428/95459651-8a090480-09a6-11eb-90e0-3b01c65d051d.png)

damithc:
> I was reading an article about button UI https://uxdesign.cc/ui-cheat-sheets-buttons-7329ed9d6112 and thought that a line button might be what we want here; what do you think?

Also, appear only on hover?

KendrickAng:
Hmm if the button appears only on hover, how would the user know the copy button is there?

ang-zeyu:
> Also, appear only on hover?

This might make the button redundant I think, the reader's first instinct would then be to just drag-and-ctrl-c the code 

> I was reading an article about button UI https://uxdesign.cc/ui-cheat-sheets-buttons-7329ed9d6112 and thought that a line button might be what we want here; what do you think?

Sounds good, or just leave the copy icon, anything to avoid drawing too much attention to the buttons

Something to keep in mind when styling it: #1355

KendrickAng:
To add on to ^ zeyu's comment, from https://javascript.info/dom-nodes:

![image](https://user-images.githubusercontent.com/38814428/95464821-fab31f80-09ac-11eb-9f06-90ac46486c33.png)

PR #1373, opened 2020-10-08
Make copy code button more subtle

KendrickAng:

More colors: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

**Current button**
image
On hover:
image

**Current alternative buttons**
image

Addresses #1278 by changing the look-and-feel of the code copy button.

Update the CSS for code block buttons, remove unused css, update the icon used for the copy button. Updated user guide to reflect the changed copy code block css.

```js

```

**Testing instructions:**
1. Serve any copy of markbind with code blocks enabled.
2. Verify that the button works and looks as intended.

Make copy code button more subtle


KendrickAng:
@damithc @ang-zeyu I've put in some buttons candidates in the description - do let me know which one would be preferred (or if you think a different design should be used)! Currently, I plan to use a simple glow effect on hover.

I've also put in a link for other alternative colors, if you'd like to suggest I use other types of colors.

damithc:
> @damithc @ang-zeyu I've put in some buttons candidates in the description - do let me know which one would be preferred (or if you think a different design should be used). Currently, I plan to use a simple glow effect on hover.

how about the last one without the border?

KendrickAng:
that sounds fine, I'll do that first and we can make changes later on if needed.

packages/core-web/src/styles/markbind.css
```diff
@@ -404,23 +404,26 @@ pre {
     position: relative;
 }
 
+/* Override user-agent stylesheets */
+#copy-btn--ua {
```
ang-zeyu, line 408: hmm what's the purpose of the separate `id` styling?

would adding it to the existing `copy-btn` class work?

packages/core-web/src/styles/markbind.css
```diff
@@ -404,23 +404,26 @@ pre {
...
-    background-color: #b4b4b9;
-    border-radius: 0.25rem;
-    color: #f8f8ff;
+    background-color: transparent;
+    color: darkgray;
     display: inline-block;
-    font-size: 75%;
-    line-height: 1;
-    padding: 0.25em 0.4em;
+    padding: 0.4rem 0.4rem;
```
ang-zeyu, line 416:
```suggestion
    padding: 0.35rem;
```

`0.4` seems a little large, may just be me though =P

leave this one up to you

packages/core-web/src/styles/markbind.css
```diff
@@ -404,23 +404,26 @@ pre {
     position: relative;
 }
 
+/* Override user-agent stylesheets */
+#copy-btn--ua {
+    border: none;
+}
+
 .copy-btn {
```
ang-zeyu, line 412: let's add `cursor: pointer` as well

ang-zeyu (commented):
Nice work styling this! Just a few nits:

packages/core-web/src/styles/markbind.css
KendrickAng: I tried that initially (adding to the existing `copy-btn` class), but noticed that it wasn't working on firefox. I thought it was happening because:

Referring to https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

1. User agent stylesheets were being applied *after* markbind's stylesheets
2. User agent stylesheets were being applied on `button` and `[type='button']`
3. For css, if I recall correctly, for the same specificity, the later styles take precendence over earlier defined styles from MDN Specificity), so user agent stylesheets were overriding markbind styles
4. So I used an id selector to override those styles

EDIT: Noticed that button.copy-btn has a higher specificity than .copy-btn alone: https://specifishity.com/;that would be a better solution

packages/core-web/src/styles/markbind.css
KendrickAng: sounds good, I'll do that

packages/core-web/src/styles/markbind.css
ang-zeyu: user agent stylesheets should always have the lowest priority https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade
I tried the `.copy-btn` class alone on firefox, it seems to work. Could it be the work of an extension?


I'm ok with `button.copy-btn` though, which should be better than a generic sounding `copy-btn` class

KendrickAng: Hmm I tried it again and `.copy-btn` did work. Maybe firefox was sending me cached pages when I tried it out myself... 

I think I'll opt to keep it as `button.copy-btn`, since it would provide more context on what `.copy-btn` should be used in

ang-zeyu: okay

ang-zeyu (approved):
Lgtm : )

PR #1374, opened 2020-10-09
Make dotted/dashed lines for triggers more visible

### LINKED ISSUE #1357: Modal/tooltip markers too fine to distinguish
image

The dotted line is so fine that it looks like a normal line. The dashes can be made bigger?

le0tan:

Closes #1357 

In `markbind.css`, `3px` below is changed from `1px` before

```
.trigger {
    border-bottom: 3px dotted currentColor;
}

.trigger-click {
    cursor: pointer;
    border-bottom: 3px dashed currentColor;
}
```

If we're following the vue-strap convention, I suppose I don't need to minify the `.css` file for this PR?

Increased trigger underline visibility

### DIFF
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -368,12 +368,12 @@ li.footnote-item:target {
 
 /* styles for triggers, popovers, tooltips */
 .trigger {
-    border-bottom: 1px dotted currentColor;
+    border-bottom: 3px dotted currentColor;
 }
 
 .trigger-click {
     cursor: pointer;
-    border-bottom: 1px dashed currentColor;
+    border-bottom: 3px dashed currentColor;
 }
 
 .trigger-click:focus {


damithc:
How's this PR going?
BTW, what caused these lines to become too fine in the first place? Something done by Browsers?

ang-zeyu:
Clean interim fix 👍 

UntitledOriginal

Seems a little big though
Perhaps we can keep the dashed style length as is (1px) since it's already quite visible, and scale down the dotted one? (e.g. 2px)

ang-zeyu:
> How's this PR going?
> BTW, what caused these lines to become too fine in the first place? Something done by Browsers?

there is no control over the spacing between the dots per css spec https://www.w3.org/TR/css-backgrounds-3/#the-border-style, its left completely to the browser

Possible workarounds I've seen:
- https://stackoverflow.com/questions/6250394/how-to-increase-space-between-dotted-border-dots - rather hacky approach, also dosen't work when the background is obscured
- `border-image` - we will need an custom made border image (am not sure of the performance implications of this), but more importantly we lose the ability to change the border color according to the `color`

damithc:
> Seems a little big though
> Perhaps we can keep the dashed style length as is (1px) since it's already quite visible, and scale down the dotted one? (e.g. 2px)

Yes, slightly smaller might still be OK.


> there is no control over the spacing between the dots per css spec https://www.w3.org/TR/css-backgrounds-3/#the-border-style, its left completely to the browser

I see. I guess we can use this workaround then. Thx for checking.

ang-zeyu:
@le0tan any thoughts on the tweaks?

tlylt:
Can probably close this PR since the root issue #1357 is resolved (the lines are reasonably visible to need a fix).

PR #1375, opened 2020-10-10
[WIP] Broken link crawler

wxwxwxwx9:
• [x] New feature

This PR resolves issue #824 by introducing a crawler that validates all intra-link during site generation. 

This will aid users in finding out whether they have any broken or invalid intra-link in their files. 

Implemented a method to check for all intra links during parsing. 

Code quality and edge cases. 

**Testing instructions:**
`npm run test`

Implement broken intra-link crawler


packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,28 +39,57 @@ const defaultTagLinkMap = {
...
+  if (link.hash) {
+    // remove the hash from fullResourcePath (to properly validate the file existence)
+    fullResourcePath = link.path;
+  }
+
+  // currently users would indicate .html extension, but they should indicate .md
+  // as we will automatically convert the .md extension to .html extension
+  if (fullResourcePath.slice(-5) === '.html') {
+    fullResourcePath = `${fullResourcePath.slice(0, fullResourcePath.length - 5)}.md`;
+  }
```
wxwxwxwx9, line 84: Currently, all our docs have .html extension instead of .md extension. Should I manually convert all of them .md extension and remove this piece of code?

ang-zeyu: Thanks for the initial implementation @wxwxwxwx9 

Two types of files can be linked: pages (`Page` model) and assets (raw unprocessed files)

For assets what you're doing here looks good so far 👍
You'll need to make it respect the [ignore config](https://markbind.org/userGuide/siteJsonFile.html#ignore) though

For pages, because the source file may not end up in `_site` (e.g. `xx.md` -> outputs `xx.html` only), you'll need to first check whether the link is referring to a `Page`, if that fails then you could fallback to the checks here for assets

there's some things in `Site` keeping track of these ^

as an example, the conversion here won't work when you have an actual raw html file (or a html file passed through our processing)

wxwxwxwx9: Hey Zeyu! I have implemented a draft version based on what you have explained to me. I think the implementation has addressed the following:
- respecting the ignore config for assets
- ensuring the Page checks are not done via FS, but done via memory (passed down as argument)

The implementation is not the most elegant / optimized version. I think it can be improved more. But I just wanted to make sure that this version is working correctly at least! 

I'm not sure if I have captured everything because I think I may have misunderstood the problem or your explanation. But maybe this draft version can help to reveal my misunderstanding.

wxwxwxwx9: Hey Zeyu! I have moved "validateIntraLink" function to after "convertMdOrMbdExtToHtmlExt" as per your recommendation. I have also made some updates based on that. 

I think everything should be working alright now. The only thing left to do is to see how I can make the code cleaner or more efficient. 

May I know your thoughts? (e.g. anything not working as it should, or ideas to make the code cleaner / optimize it)

Thanks so much for your help so far! :-)

ang-zeyu (commented):
Just some nits : )

---

I saw that you moved `addressablePageSource = ...` elsewhere as well, to confirm does that make this work with live reload?

Gulruhbet:
Do you have a code intra

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,101 @@ const defaultTagLinkMap = {
...
+      // Page source file exists
+      return;
     }
   }
+
+  // file is an asset
+
+  // checks if the asset file is an ignored file
+  const fileIgnore = ignore().add(config.ignore);
+  const tempResourcePath = resourcePath.substring(1); // remove the '/' first char in the string
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup! I'll see how I can insert the removal part to ensure there's little / no repetition.

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,101 @@ const defaultTagLinkMap = {
...
+  // since this function is called after convertMdAndMbdExtToHtml,
+  // we can assume that page source mostly has .html extension (unless user indicates no-convert)
+  const hasHtmlExt = resourcePath.slice(-5) === '.html';
+  const hasMdExt = resourcePath.slice(-3) === '.md'; // suppose user indicates no-convert
+  const hasMbdExt = resourcePath.slice(-4) === '.mbd'; // suppose user indicates no-convert
+  if (hasHtmlExt || hasMdExt || hasMbdExt) {
+    // remove ext and '/' first char
+    const resourcePathWithNoExt = fsUtil.removeExtension(resourcePath).substring(1);
+    const isPageSrc = config.addressablePagesSource.includes(resourcePathWithNoExt);
+    // if not pageSrc, it is a raw asset
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup agreed! I was intending to extract all the inline comments out once everything is settled.

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,101 @@ const defaultTagLinkMap = {
...
+      return;
     }
   }
+
+  // file is an asset
+
+  // checks if the asset file is an ignored file
+  const fileIgnore = ignore().add(config.ignore);
+  const tempResourcePath = resourcePath.substring(1); // remove the '/' first char in the string
+  if (tempResourcePath !== '' && fileIgnore.ignores(tempResourcePath)) {
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup the `!resourcePath` above actually checks for empty string. 

However, I noticed that are cases where the intra link is just a backslash, `/`. Thus, when I remove the backslash from `resourcePath` in line 115, I actually get an empty string. 

I'm thinking of just adding a new condition to the first guard clause to check if the `resourcePath` is only a backslash. I think there's nothing to check for when it's just a backslash, right? Correct me if I'm wrong.

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,101 @@ const defaultTagLinkMap = {
...
+  if (resourcePathUrl.hash) {
+    resourcePath = resourcePathUrl.path;
+  }
+
+  // check if the file is a Page source
+  // since this function is called after convertMdAndMbdExtToHtml,
+  // we can assume that page source mostly has .html extension (unless user indicates no-convert)
+  const hasHtmlExt = resourcePath.slice(-5) === '.html';
+  const hasMdExt = resourcePath.slice(-3) === '.md'; // suppose user indicates no-convert
+  const hasMbdExt = resourcePath.slice(-4) === '.mbd'; // suppose user indicates no-convert
```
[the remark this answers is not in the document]
wxwxwxwx9: > we can be sure here that they are pointing to assets, because pages always generate `.html`s

Hmm, I'm pretty confused here. If I understand you correctly, you mean that pages (regardless .mb or .mbd), always generates and outputs `.html`s to the `_site` folder, right? I am not sure how 'pages always generate .html' comes into the picture for the current context. 

My thought process is this: 
1. User has intra-link in document which can be of `.md, .mbd, or .html` extensions (however, they are mostly .html extension in our current user docs and dev docs, due to past user behaviors, as we did not have the auto conversion back then)
2. After converting the intra-links using `convertMdAndMbdExtToHtmlExt`, the intra-links can still be one of `.md, .mbd, or .html` extensions, right? We can have `.md or .mbd` extensions in the scenario where the user indicates the `no-convert` attribute in their tag. That was why I thought we have to check for those extensions here. If we just check for `.html` extension, we won't be able to validate the intra-links that have `.md or .mbd` extensions. 

Regarding your point about pages always generating and outputting `.html` into `_site`, how would that make the `.md and .mbd` extension checks redundant? Or rather, how are they related? I can't seem to draw the relation :-/

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,101 @@ const defaultTagLinkMap = {
...
     const { href } = node.attribs;
     const hasMdExtension = href.slice(-3) === '.md';
-    const hasNoConvert = lodashHas(node.attribs, 'no-convert');
-    if (hasMdExtension && !hasNoConvert) {
+    if (hasMdExtension) {
       const newHref = `${href.substring(0, href.length - 3)}.html`;
       node.attribs.href = newHref;
+      return;
+    }
+    const hasMbdExtension = href.slice(-4) === '.mbd';
```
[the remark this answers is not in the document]
wxwxwxwx9: Alright noted! I'm not too familiar with the various git merging methods, but I'll look it up! Just to clarify, what I would have to do is to merge all of current my commits into a few (e.g. 2 to 5) logically-named commits, right?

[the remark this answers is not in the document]
ang-zeyu: yup, in the case just 2.

forgot to mention: do rebase on top of the latest commit as well so we can keep things clean

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
ang-zeyu: if pages always generate `.html`, then we know that `.md` / `.mbd` links would definitely be pointing to assets already.

because by default we assume such links are pointing to pages (hence converting them `.html`).
if there is still a `.md` / `.mbd` link post conversion (indicated via `no-convert`), that means the user wants to point to a raw asset file (e.g. the common case of `readme.md`)

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
ang-zeyu: hmm good point, a `/` is still a valid link though so we don't want to skip that actually

that reminds me as well, you'll need to take care of cases where the link implicitly points to some `index.html` (e.g. `href="/userGuide/"` and  `href="/userGuide"` both points to `href="/userGuide/index.html"`) for both pages and assets

so at this point if we had a `/` link, it would be `/index.html`

*also note that checking for `href="/userGuide"` is a little more tricky (versus `href="/userGuide/"`) because you'd also have to check for `userGuide` as a raw file 😓 

you can refer to `cli/index.js` which has similar link normalization for the lazy live reload (`-o`) option

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
wxwxwxwx9: > .md / .mbd links would definitely be pointing to assets already.

Hmm, let me clarify my basic understanding. By this, you mean that `.md` and `.mbd` links would be pointing to the generated `.html` files right? That is why we convert the `.md` and `.mbd` extensions to `.html` extensions. 

> if there is still a .md / .mbd link post conversion (indicated via no-convert), that means the user wants to point to a raw asset file (e.g. the common case of readme.md)

Hmm, so you mean that for the case of `readme.md`, we will fall back to the asset fs check? I noticed that the current ignore config that we have, `*.md` files are ignored. This means that we wouldn't have to check for the existence of `readme.md` files, right? 


Also, I think the `readme.md` would fall under the addressablePages, right? So if we want to check for its existence, we can either check under the 1. Pages check step or 2. the asset fs check, right?

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
wxwxwxwx9: Alright noted!

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
wxwxwxwx9: Alright thanks for raising this up! Will take this into account

wxwxwxwx9:
> I saw that you moved addressablePageSource = ... elsewhere as well, to confirm does that make this work with live reload?

Actually, I created `addressablePageSource` as a new variable! I'm not sure if you're referring to `addressablePage` :O

I don't think I've moved the existing variables around. Let me know if I misinterpreted your message!

packages/core/src/parsers/linkProcessor.js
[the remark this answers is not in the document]
ang-zeyu: > Hmm, let me clarify my basic understanding. By this, you mean that .md and .mbd links would be pointing to the generated .html files right? That is why we convert the .md and .mbd extensions to .html extensions.

yup 👍 

> I noticed that the current ignore config that we have, *.md files are ignored. This means that we wouldn't have to check for the existence of readme.md files, right?

it may not be ignored in all cases; the solution should be generalizable to anything that the user can configure (not just our docs)

> Also, I think the readme.md would fall under the addressablePages, right?

not if it is excluded (or just not included in any glob / src)
https://markbind.org/userGuide/siteJsonFile.html#pagesexclude

packages/core/src/Site/index.js
```diff
@@ -523,6 +526,8 @@ class Site {
         : filteredPage;
     });
     this.addressablePages = Object.values(filteredPages);
+    this.addressablePagesSource = Object.values(filteredPages).map(page =>
```
ang-zeyu, line 529: > I don't think I've moved the existing variables around. Let me know if I misinterpreted your message!

hmm sorry, might've misread the earlier code

this looks good but to make it work with live reload you may need to maintain the same reference since we're only passing it to `pageConfig` once (clear and repush the array instead of `addressablePagesSource  = ...`)

wxwxwxwx9: Ah, can I check if this would apply to `addressablePages` as well? 

Because in line 528, there's also the reassignment. Will `addressablePages` still work with live reload?

ang-zeyu: yup, its not being passed anywhere else if I recall

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,114 @@ const defaultTagLinkMap = {
...
+  if (tempResourcePath && fileIgnore.ignores(tempResourcePath)) {
+    return;
+  }
+
+  // asset file is not ignored, check for the existence of asset using fs
+
+  // only dealing with absolute paths because relative paths were already converted to absolute
+  let fullResourcePath = path.join(config.rootPath, resourcePath);
+  if (config.baseUrl) {
+    fullResourcePath = fullResourcePath.replace(config.baseUrl, '');
```
ang-zeyu, line 124: hmm shouldn't this be first and foremost? (replacing `tempResourcePath = ...`)
since the paths in `addressablePagesSource` are not prepended by the baseUrl

let's do it more safely as well, e.g. ``path.posix.relative(`${config.baseUrl}/`, resourcePath)``, in case the baseUrl appears in the middle of the path

packages/core/src/Site/index.js
```diff
@@ -523,6 +526,10 @@ class Site {
         : filteredPage;
     });
     this.addressablePages = Object.values(filteredPages);
+    this.addressablePagesSource.clear();
```
ang-zeyu, line 529: this is pseudo code (there's no `clear()` function 😅)

do retest when making changes!

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,114 @@ const defaultTagLinkMap = {
...
+  }
+
+  if (fullResourcePath.endsWith('/')) {
+    fullResourcePath += 'index.html';
```
ang-zeyu, line 128: this also needs to be before the page check, since its `.html`

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,114 @@ const defaultTagLinkMap = {
...
+
+  if (!utils.fileExists(fullResourcePath)) {
+    const err = `You might have an invalid intra-link! Ignore this warning if it was intended.
+'${originalResourcePath}' found in file '${cwf}'`;
+    logger.warn(err);
+  }
+
+  const hasFileExtension = path.posix.extname(fullResourcePath);
+  // does not end with '/' and has no file extension
+  if (!hasFileExtension) {
```
ang-zeyu, line 139: since these types of links can point to either a raw file or some `index.html`, you'll also need to do the `addressablePagesSource` check here.

it should be before the asset check at line 131 as well: `if (!utils.fileExists(fullResourcePath)) {`
otherwise we may get an error logged when it is actually pointing to a page

ang-zeyu (commented):
Hmm... seems like there's quite a few edge cases here now with the implicit `html` links added  - shall we make the function testable (e.g. return `true / false`) and add a few simple unit tests? (then move the logging out of the tested function)

This should also help to communicate the intra link requirements more clearly, and prevent regressions when you're adjusting based on the comments below. E.g:
- page and asset check should be done for `.html`, paths ending with `/`, paths ending with no extension (6 cases)
- all other links should only get the asset check (1 case)

You can use the `htmlparser2.parseDOM / cheerio.parseHtml` method to get the test `node` parameters easily as well

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,114 @@ const defaultTagLinkMap = {
...
+  }
+
+  let resourcePath = originalResourcePath;
+  // remove hash portion (if any) in the resourcePath
+  const resourcePathUrl = url.parse(resourcePath);
+  if (resourcePathUrl.hash) {
+    resourcePath = resourcePathUrl.path;
+  }
+
+  const tempResourcePath = resourcePath.substring(1); // remove the '/' first char
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup, I really should. Thanks for the suggestion!

packages/core/src/parsers/linkProcessor.js
wxwxwxwx9: Hmm, that's true. My bad for missing this out. Will do as what you suggested!

packages/core/src/Site/index.js
wxwxwxwx9: Ah, right. That was a very careless mistake. Will make sure I retest every time from now on!

packages/core/src/parsers/linkProcessor.js
wxwxwxwx9: Noted!

packages/core/src/parsers/linkProcessor.js
wxwxwxwx9: Alright noted!

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,114 @@ const defaultTagLinkMap = {
...
+  // checks if the asset file is an ignored file
+  const fileIgnore = ignore().add(config.ignore);
+  if (tempResourcePath && fileIgnore.ignores(tempResourcePath)) {
+    return;
+  }
+
+  // asset file is not ignored, check for the existence of asset using fs
+
+  // only dealing with absolute paths because relative paths were already converted to absolute
+  let fullResourcePath = path.join(config.rootPath, resourcePath);
```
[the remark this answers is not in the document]
wxwxwxwx9: Noted!

wxwxwxwx9:
> Hmm... seems like there's quite a few edge cases here now with the implicit `html` links added - shall we make the function testable (e.g. return `true / false`) and add a few simple unit tests? (then move the logging out of the tested function)
> 
> This should also help to communicate the intra link requirements more clearly, and prevent regressions when you're adjusting based on the comments below. E.g:
> 
> * page and asset check should be done for `.html`, paths ending with `/`, paths ending with no extension (6 cases)
> * all other links should only get the asset check (1 case)
> 
> You can use the `htmlparser2.parseDOM / cheerio.parseHtml` method to get the test `node` parameters easily as well

Sounds like a good idea! I'm not very familiar with function unit testing in the codebase currently. And from first glance, I don't think I see anything that unit tests functions(s). Though I may have saw wrongly. Can I check if there is any unit testing material in the codebase that I can used as a reference / guideline?

Anyway, I have updated my code to fit your suggestions as much as I could! There is quite a bit of changes but I think it is more readable now. Hopefully there's some improvement from my previous push. 

> page and asset check should be done for .html, paths ending with /, paths ending with no extension (6 cases)
all other links should only get the asset check (1 case)

I think I should have dealt with all the cases above! Let me know if I'm still missing anything out :-/

ang-zeyu (commented):
Nice work! Looks really neat now 👍 

Given the way you organised things I don't think a high level overview jsdoc would be necessary as well since its pretty clear, but feel free to add it.

More importantly let's add the few unit tests as mentioned, split up the commits, and I think we can merge this in.

> Sounds like a good idea! I'm not very familiar with function unit testing in the codebase currently. And from first glance, I don't think I see anything that unit tests functions(s). Though I may have saw wrongly. Can I check if there is any unit testing material in the codebase that I can used as a reference / guideline?

yup, we use [jest](https://jestjs.io/docs/en/getting-started) for unit testing, they should be under `packages/core/test`. for stubbing the filesystem and files we use https://www.npmjs.com/package/memfs

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,120 @@
...
+  const mockLink = '<a href="https://markbind.org">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = 'Not Intralink';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test valid ".html" extension link ', () => {
+  // should be checked as file asset and page
```
ang-zeyu, line 46:
```suggestion
  // should be checked as page
```

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,120 @@
...
+  const mockLink = '<a href="/userGuide/raw.html">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_ERR_RESULT = 'Intralink with ".html" extension is a valid Page Source or File Asset';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_ERR_RESULT);
+});
+
+test('Test valid link ending with /', () => {
+  // should be checked as file asset and page
```
ang-zeyu, line 76: let's add one that checks for `/xx/index.html` as an asset too

ang-zeyu:
```suggestion
  // should be checked as page
```

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,120 @@
...
+  const mockLink = '<a href="/userGuide/">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = 'Intralink ending with "/" is a valid Page Source or File Asset';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test valid link ending with no extension', () => {
+  // should be checked as file asset and page
```
ang-zeyu, line 86:
```suggestion
  // should be checked as page
```

ang-zeyu: needs two more that checks as an asset
some `/xx/index.html` (missed this out in earlier review)
`/xx` (the raw file)

packages/core/src/parsers/linkProcessor.js
```diff
@@ -23,41 +41,141 @@ const defaultTagLinkMap = {
...
+    return 'Intralink ending with "/" is a valid Page Source or File Asset';
+  }
+
+  const hasNoFileExtension = path.posix.extname(resourcePath) === '';
+  if (hasNoFileExtension) {
+    // does not end with '/' and no file ext (e.g. /userGuide)
+    // check as file asset and implicit index.html
+    const implicitResourcePath = `${resourcePath}/index.html`;
+    const asFileAsset = resourcePath;
+    if (!isValidPageSource(implicitResourcePath, config) && !isValidFileAsset(asFileAsset, config)) {
```
ang-zeyu, line 151: will need to check `implicitResourcePath` as an asset too

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,120 @@
...
+test('Test valid file asset links (png)', () => {
+  const mockLink = '<img src="/images/logo.png">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = 'Intralink is a valid File Asset';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test valid file asset links (css)', () => {
```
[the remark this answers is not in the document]
wxwxwxwx9: Ah, they are testing the same thing actually: file assets. I just added another extra check, which is redundant in hindsight. Should I remove it?

packages/core/test/unit/LinkProcessor.test.js
wxwxwxwx9: Added the test!

packages/core/test/unit/LinkProcessor.test.js
wxwxwxwx9: Added the test!

packages/core/src/parsers/linkProcessor.js
wxwxwxwx9: Added this condition!

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,175 @@
...
+  const mockLink = '<a href="/userGuide/">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = 'Intralink ending with "/" is a valid Page Source or File Asset';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test valid link ending with /', () => {
+  // should be checked as file asset
```
ang-zeyu, line 88:
```suggestion
  // should be checked as page and file asset
```

packages/core/test/unit/LinkProcessor.test.js
```diff
@@ -0,0 +1,175 @@
...
+  const mockLink = '<a href="/userGuide">Test</a>';
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = 'Intralink with no extension is a valid Page Source or File Asset';
+
+  expect(linkProcessor.validateIntraLink(mockNode, mockCwf, mockConfig)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test valid link ending with no extension', () => {
+  // should be checked as file asset (implicit index resource path will be true)
```
ang-zeyu, line 118:
```suggestion
  // should be checked as page and file asset (implicit index resource path will be true)
```

packages/core/test/unit/LinkProcessor.test.js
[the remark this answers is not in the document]
ang-zeyu: sounds good

ang-zeyu (commented):
Just a few more minor corrections:
(could also start splitting the commits meanwhile)

Issue #561, opened 2019-01-08
Indicate currently loaded page in topNav


damithc:
Similar to #560 but for the top navigation bar

Example (from Jekyll):
![image](https://user-images.githubusercontent.com/1673303/50804235-0f6e0980-1328-11e9-962e-c78d4910e623.png)

PR #1379, opened 2020-10-23
Indicate currently loaded page in topNav

KendrickAng:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Addresses #561 

This PR allows highlighting of the top-level navbar link. An exact link match is first done between `window.location.href` and all `a.nav-link`/`a.dropdown-item` anchortags. If this cannot be done, if a user has specified a *fallback option* through the `default-highlight-on` tag in the <navbar> element, then link highlighting is done using that option.

The `default-highlight-on` tag currently supports these highlighting rules:
- `sibling-or-child` - performs a link match only if the `a` tag has same prefix as `window.location.href`
- `sibling` - `markbind.org/tests/foo.html` is a sibling of `markbind.org/tests/bar.html`
- `child` - `markbind.org/tests/test1/foo.html` is a child of `markbind.org/tests/index.html`
- `none` - performs no link matching; no link is highlighted
- `exact` - performs an exact link match

What kind of link highlight styling is preferred?

**Testing instructions:**
Build core web assets, then serve a page

indicate current page in top navigation bar

---

- [X] No blatantly unrelated changes
- [X] Pinged someone for a review!


ang-zeyu:
Initial thoughts: Since we're using vue, how about moving this into the navbar component itself? (or is there a reason for separating this?)

KendrickAng:
Hmm I didn't think of that; I don't really understand how the vue components tie into the overall application yet. Are they loaded into a `page` at some point in the build process?

Also, I realise that my current implementation is not entirely correct yet, hence I changed it back to WIP

ang-zeyu:
> Hmm I didn't think of that; I don't really understand how the vue components tie into the overall application yet. Are they loaded into a `page` at some point in the build process? Could I ask where a vue component is parsed into regular HTML?

its rendered at runtime (browser)

the code we write (.vue files) is compiled by webpack into something the browser is happy with at bundle build time though (`markbind s -d` or `npm run build:web`)

KendrickAng:
Ahh, I see, it would make sense to add the script logic to the Navbar itself

KendrickAng:
@ang-zeyu could you have a look at this?

packages/vue-components/src/Navbar.vue
```diff
@@ -77,10 +77,41 @@
...
+          }
+          // otherwise, check all children recursively
+          return Array.from(elements.children).some(node => hasMatchingUrl(node, currPage));
+        }
+        document.addEventListener('DOMContentLoaded', () => {
+          const navLinks = document.querySelectorAll('.navbar .navbar-nav .nav-link');
+          if (!navLinks) {
+            return;
+          }
+          const currPage = window.location.href;
```
ang-zeyu, line 103: the detection does not work currently for implicit `index.html` links (e.g. `markbind.org/userGuide` / `markbind.org/userGuide/`-> points to `markbind.org/userGuide/index.html`)

you'll need to normalize both the `href` and `window.location.href`s when checking

ang-zeyu: should we make the checking "generalizable" as well?

e.g. if the link is `markbind.org/userGuide/index.html`,

`userGuide/index.html`
`userGuide/formattingContents.html`
`userGuide/formattingContents.html`

would all match

Since the links in the navbar are usually only a small subset

---

On the other hand, this may make the checking ambiguous though; we'll need to define the cases clearly and see if there's any such case before proceeding

What do you think? @KendrickAng

ang-zeyu (commented):
Thanks!

I think we'll need to sort out how the link checking is done (`window.location.href` comment) before going further first

packages/vue-components/src/Navbar.vue
KendrickAng: > 
> 
> the detection does not work currently for implicit `index.html` links (e.g. `markbind.org/userGuide` / `markbind.org/userGuide/`-> points to `markbind.org/userGuide/index.html`)
> 
> you'll need to normalize both the `href` and `window.location.href`s when checking

What do you mean by "implicit index.html links"? I thought that all links in the navbar are well-defined, i.e we hardcode them in the `header.md` in layouts.

Didn't think of the normalisation part, I'll do that!

> 
> 
> should we make the checking "generalizable" as well?
> 
> e.g. if the link is `markbind.org/userGuide/index.html`,
> 
> `userGuide/index.html`
> `userGuide/formattingContents.html`
> `userGuide/anotherfolder/formattingContents.html`
> 
> would all match
> 
> Since the links in the navbar are usually only a small subset
> 
> On the other hand, this may make the checking ambiguous though; we'll need to define the cases clearly and see if there's any such case before proceeding
> 
> What do you think? @KendrickAng

In that case, the navbar would need to maintain some sort of state (luckily we use vue components, which has state)
But then this brings up another issue - what if the side nav href doesn't match the top nav href? For example link is `markbind.org/userGuide/index.html`, side nav has `userGuide/formattingContents.html` and ` anotherName/anotherfolder/formattingContents.html`.

In that case, there wouldn't be a way of knowing that those links have the same parent `markbind.org/userGuide/index.html`. Do you have an idea how to overcome this?

ang-zeyu: > What do you mean by "implicit index.html links"?

>`markbind.org/userGuide` / `markbind.org/userGuide/`-> points to `markbind.org/userGuide/index.html`

links like these

e.g. the main page of netlify preview only highlights `home` if the address is https://deploy-preview-1379--markbind-master.netlify.app/index.html (vs  https://deploy-preview-1379--markbind-master.netlify.app/ which should also highlight it)

> In that case, the navbar would need to maintain some sort of state

state wouldn't be needed since we're checking once-per-page only, just some extra logic around how you match the current address to the `href`s 

> what if the side nav href doesn't match the top nav href?

hmm "generalizable" was a bad choice of words (our site nav already has highlighting)

I meant to "relax" the search conditions - so in the above example, those 3 addresses are _browser addresses_, the one before that is the navbar link

For now I'm thinking simply:

If after link normalization (changing links like `userGuide/` into `userGuide/index.html`), the link points to some `index.html`, we could check if `window.location.href` is a _sibling_ or _child_ of this. (e.g. `userGuide/formattingContents.html` or `userGuide/anotherfolder/somefile.html`) If so we highlight the link

If after link normalization, it points to something other than some index.html, we do the exact equality check already present in this pr

KendrickAng: oh so normalization meant adding an `index.html` to links that don't end with `.html`, I thought you mean something like this: thttps://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

ang-zeyu: yup

packages/vue-components/src/Navbar.vue
```diff
@@ -77,10 +77,41 @@
...
         e && e.preventDefault()
         this.collapsed = !this.collapsed
+      },
+      highlightCurrentPage() {
+        function hasMatchingUrl(elements, currPage) {
+          if (!elements || !elements.children) {
+            return false;
+          }
+          // Only check <a> leaf nodes
+          if (elements.children.length === 0) {
```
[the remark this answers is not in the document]
KendrickAng: hmm, a query selector on a class might not work since dropdown menus in nav bars only have a class `dropdown-item` whereas ordinary navbar links have a class of `nav-link`. So if we want highlighting on dropdown links, another document.query doesn't seem to be possible, hence the class-agnostic DFS

image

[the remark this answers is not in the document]
ang-zeyu: not on the class, but the anchor links you're checking here `'a[href]'`

packages/vue-components/src/Navbar.vue
KendrickAng: Actually, I realise that this checking of sibling or child may not work well, given that there may be cases like this, where multiple nav links have a common prefix:
image
I think we could employ a stricter check (the dfs check) for an exact matching, and if that doesn't work, use the relaxed checking you suggested here. I believe that works for the current markbind docs and the example I listed here

KendrickAng:
@ang-zeyu I've changed the highlight logic to a prototype "strict to relaxed" approach. If this approach is all right, I'll proceed to optimise + clean up the code

packages/vue-components/src/Navbar.vue
ang-zeyu: Hmm we could do some sort of priority matching here -> the more "groups" (`/`) the link has, the more priority we give to it. And then we only highlight the one with the highest priority

I think this could actually be desired behaviour on the other hand though
e.g. if you have this dropdown structure
```
userGuide/
  topic1.html
```
both the main dropdown link and `topic1.html` is highlighted

Any thoughts 😮

ang-zeyu: Another alternative is to allow some additional attributes to control the highlighting behaviour:

e.g.
```
<a href="..."
  highlight-on="none / exact / child / sibling / child or sibling" // default this to "child or sibling" unless specified
  highlight-link="..." // we can even highlight non-links this way (another pr)
  highlight-class="..." // maybe even allow the user to specify this, but default it to something reasonable (another pr maybe)
  />
```

ang-zeyu: > when we match on groups (/), clicking on topic3a.html causes topic1.html's link to be incorrectly highlighted instead, since the relaxed checking checks for a common prefix (contents in this case)

> If after link normalization, it points to something other than some index.html, we do the exact equality check already present in this pr

Do you mean if the address bar is `contents/topic3a.html`, `contents/topic1.html` would also be highlighted?
My suggestion here was to do _only_ exact equality checks for such cases (non `index.html`s). (versus falling back to the prefix match at all)

But this behaviour may be rather obscure (falling back in the case of `index.html`s, not falling back at all otherwise).

https://github.com/MarkBind/markbind/pull/1379#discussion_r515591206
Hence we could always fall back, but provide an alternative  to control the mode of checking

> However, I believe that it would be quite rare where a user does something like that though (not using the exact url, like in markbind docs)

Hmm not sure I understood this, what does "not using the exact url" refer to? (wouldn't links as specified by `<a href="..." />` always point to an exact url?)

KendrickAng: I think we may be talking about different issues here... The problem was that since we could have something like `topic1.html` and `topic3a.html` having the same prefix `contents`, when we match on groups (/), clicking on `topic3a.html` causes `topic1.html`'s link to be highlighted instead, since the relaxed checking checks for a common prefix (contents in this case)

Hence I thought of doing an exact URL match first, then if that doesn't work, do a match on the highest-priority URL (with the most groups matched).

KendrickAng: >Hmm not sure I understood this, what does "not using the exact url" refer to? (wouldn't links as specified by <a href="..." /> always point to an exact url?)

I was talking about the case where the href of an anchor tag doesn't match the *end result* of `window.location.href` once it is clicked, for example:

1. Suppose we have this entry in `layouts/devGuide/header.md` <div tags="environment--dg"><li><a href="{{baseUrl}}/devGuide/index.html" class="nav-link">DEVELOPER GUIDE</a></li></div>`
1. However, `devGuide/index.md` contains this code:
><script>
> window.location.href = "devGuide.html"
></script>
1. So when the anchor tag is clicked, the final window.location.href is `devGuide/devGuide.html` instead of `devGuide/index.html` - so exact matching on the anchor tag will not work

--------------------------------
Also, I think I'm quite confused about the algorithm being described so far, so could I check if this is what you mean:
```js
const url = normalize(window.location.href)
if url ends wth `index.html`, do an exact match search
else:
    fallback to highlighting based on highlight-on attribute
    (where child or sibling means to highlight the 'highest-priority-by-number-of-groups-matched links')
```

ang-zeyu: My original suggestion here https://github.com/MarkBind/markbind/pull/1379#discussion_r511595069:

The condition should be the opposite. The `url` variable should be the link, not `window.location.href` as well:

```js
const link = normalize(`<a href>` link)
if link ends wth `index.html`
  do an exact match search.
  If that fails,
    fallback to the relaxed search
    // because xx/index.html commonly is the (sub)home page of a whole bunch of other urls
else:
  only do exact search
```

---

The alternative I mentioned here https://github.com/MarkBind/markbind/pull/1379#discussion_r515591206:

```js
const link = normalize(`<a href>` link)
do an exact match search.
If that fails, **and** the mode of checking allows for falling back
  fallback to the relaxed search
  // because xx/index.html commonly is the (sub)home page of a whole bunch of other urls
```

This is also what's currently being done in your pr, if I read your code right.
The problem you encountered can be resolved by adding some sort of `highlight-on` attribute that allows controlling the mode of checking. (priority based checking is probably a bad idea as you've found, because we can have equal priority links)


Option 1 would likely result in less work here and on the author side, but:
- it is somewhat obscure to the author in that the highlighting behaviour depends on the link
- I'm not sure if there's any use cases that is impossible to accomodate to using this

Option 2 would need some more intervention from the author and implementation here, but should be the most flexible option

ang-zeyu: > > `Hmm not sure I understood this, what does "not using the exact url" refer to? (wouldn't links as specified by <a href="..." /> always point to an exact url?)`
> 
> I was talking about the case where the href of an anchor tag doesn't match the _end result_ of `window.location.href` once it is clicked, for example:
> 
> 1. Suppose we have this entry in `layouts/devGuide/header.md`
> 
> ```
> <div tags="environment--dg">
> <li>
> <a href="{{baseUrl}}/devGuide/index.html" class="nav-link">DEVELOPER GUIDE</a>
> </li>
> </div>
> ```
> 
> 1. However, `devGuide/index.md` contains this code:
> 
> > <script>
> > window.location.href = "devGuide.html"
> > </script>
> 
> 1. So when the anchor tag is clicked, the final window.location.href is `devGuide/devGuide.html` instead of `devGuide/index.html` - so exact matching on the anchor tag will not work

So in this example, both options 1 & 2 would accomodate it without any intervention from the author
(the `<a href="{{baseUrl}}/devGuide/index.html" class="nav-link">DEVELOPER GUIDE</a>` is highlighted)

KendrickAng: Option 2 would be more future-proof; I believe it'll be better to go ahead with that

KendrickAng:
@ang-zeyu What do you think about this approach applying what we have talked about:
1. Allow the user to (optionally) specify his navbar highlighting preferences in `site.json`, which is either `none`, `exact`, `sibling-or-child`.
1. Pass the preferences (I believe it was through PageConfig) to the parser, which then applies the rules as necessary in `componentParser.render`
1. In `Navbar.vue`, extract the `data-highlight-on` attribute if exact URL matching doesn't work

Additionally, right now I'm simply hardcoding the parser to insert a `data-highlight-on` attribute to all `a.nav-link` and `a.dropdown-item` in `<navbar>`. If you find the above approach fine, I'll proceed with it

ang-zeyu:
> @ang-zeyu What do you think about this approach applying what we have talked about:
> 
> 1. Allow the user to (optionally) specify his navbar highlighting preferences in `site.json`, which is either `none`, `exact`, `sibling-or-child`.

how about navbar-level vs site-level? `<navbar default-highlight-on="..." />`. That way we can keep this feature solely on the frontend.

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+                  // check for same ancestor
+                  if (aParts[0] === bParts[0]) {
+                    li.classList.add('current');
+                    return true;
+                  }
+                  break;
+                case 'sibling':
+                  // check for exact same prefix parts in both URLs
+                  if (aParts.length === bParts.length) {
+                    for (let i = 0; i < aParts.length; i++) {
```
ang-zeyu, line 134: should this be `aParts.length - 1`?

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+              // checks by strict equality of URL
+              if (aNorm === urlNorm) {
+                li.classList.add('current');
+                return true;
+              }
+            } else {
+              const aTagUrl = new URL(aNorm);
+              const browserUrl = new URL(urlNorm);
+              const aParts = `${aTagUrl.host}${aTagUrl.pathname}`.split('/').slice(1);
+              const bParts = `${browserUrl.host}${browserUrl.pathname}`.split('/').slice(1);
```
ang-zeyu, line 122: hmm should we move the browser parts out of the loops? since they're the same for any iteration

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+                        return false;
+                      }
+                    }
+                    li.classList.add('current');
+                    return true;
+                  }
+                  break;
+                case 'child':
+                  // check for same ancestor, and <a> tag url has more parts
+                  if (aParts[0] === bParts[0] && aParts.length > bParts.length) {
```
ang-zeyu, line 145: similar to `sibling-or-child`

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+              }
+            } else {
+              const aTagUrl = new URL(aNorm);
+              const browserUrl = new URL(urlNorm);
+              const aParts = `${aTagUrl.host}${aTagUrl.pathname}`.split('/').slice(1);
+              const bParts = `${browserUrl.host}${browserUrl.pathname}`.split('/').slice(1);
+              switch (this.defaultHighlightOn) {
+                case 'sibling-or-child':
+                  // check for same ancestor
+                  if (aParts[0] === bParts[0]) {
```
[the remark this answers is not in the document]
ang-zeyu: suggestion: create some helper methods to reuse for sibling, child, sibling-or-child

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+      highlightLink(url) {
+        // Performs an exact equality match on links
+        if (this.hasLinkMatch(url, true)) {
+          return;
+        }
+        return this.hasLinkMatch(url, false);
+      },
+      hasLinkMatch(url, isStrict) {
+        const navLis = Array.from(this.$el.querySelector('.navbar-nav').children);
+        for (const li of navLis) {
```
ang-zeyu, line 104: thanks for implementing this as well! 🙂 

To clarify, the behaviour here is: if any link belonging to the dropdown / navbar item matches, we highlight the dropdown / navbar item as well.

Just a thought, is it possible to:
- apply the `current` styling on `<a>` tags
- additionally apply it to **only** dropdowns?

Reason being we may want to highlight **both** the parent (dropdown header) and child (dropdown item) links

If so maybe we can raise this (and all the other things) in an issue after this pr (unless you want to include it here as well)

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+                    li.classList.add('current');
+                    return true;
+                  }
+                  break;
+                case 'sibling':
+                  // check for exact same prefix parts in both URLs
+                  if (aParts.length === bParts.length) {
+                    for (let i = 0; i < aParts.length; i++) {
+                      if (aParts[i] !== bParts[i]) {
+                        return false;
```
ang-zeyu, line 136: will need to continue checking the rest of the `a`'s and `li`'s

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+              if (aNorm === urlNorm) {
+                li.classList.add('current');
+                return true;
+              }
+            } else {
+              const aTagUrl = new URL(aNorm);
+              const browserUrl = new URL(urlNorm);
+              const aParts = `${aTagUrl.host}${aTagUrl.pathname}`.split('/').slice(1);
+              const bParts = `${browserUrl.host}${browserUrl.pathname}`.split('/').slice(1);
+              switch (this.defaultHighlightOn) {
```
ang-zeyu, line 123: > In Navbar.vue, extract the data-highlight-on attribute if exact URL matching doesn't work

one more nit: the logic for this seems to be missing

could we shorten it to `highlight-on` as well? (for brevity on the author side)

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
+                  break;
+                case 'child':
+                  // check for same ancestor, and <a> tag url has more parts
+                  if (aParts[0] === bParts[0] && aParts.length > bParts.length) {
+                    li.classList.add('current');
+                    return true;
+                  }
+                  break;
+                case 'exact':
+                  return this.hasLinkMatch(url, true);
```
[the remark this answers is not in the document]
KendrickAng: hmm I don't think so; I suppose I'll just remove the highlight = exact option

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,89 @@
...
-      }
+      },
     },
     methods: {
       toggleCollapse (e) {
         e && e.preventDefault()
         this.collapsed = !this.collapsed
+      },
+      // add directory index, e.g http://foo/ or http://foo becomes http://foo/index.html
+      normalizeUrl(url) {
```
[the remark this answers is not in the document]
KendrickAng: https://community.netlify.com/t/my-url-paths-are-forced-into-lowercase/1659 - netlify apparently does it on purpose to handle filesystems that deal with case insensitivity, didn't know

Given that Windows/Mac doesn't care about case sensitivity when serving files, I believe it'll be correct to do the same (not care about case sensitivity) - also, for linux systems the file would (I assume) even fail to serve when we try to serve a file case-insensitive-ly, so it shouldn't matter even if link highlight is incorrectly enabled for linux

packages/vue-components/src/Navbar.vue
KendrickAng: In view of the upcoming deadline for CP3106, I would like to leave that to another PR 😅

packages/vue-components/src/Navbar.vue
KendrickAng: right, it should be early termination on finding a match, not early termination on absence of a match

packages/vue-components/src/Navbar.vue
KendrickAng: Will shorten the `highlight-on` part

@ang-zeyu I was under the impression that the user just needs to pass the `highlight-on` attribute like so: <navbar highlight-on='none'>, which should be a valid way to pass props to a vue component. After which, the prop is automatically extracted, so no explicit extraction should be needed. Is this correct?

ang-zeyu: hmm to clarify, from your comment here https://github.com/MarkBind/markbind/pull/1379#issuecomment-723534533, the `highlight-on` configuration is an addition to option 2 discussed here https://github.com/MarkBind/markbind/pull/1379#discussion_r515606407 right?

my suggestion was to move the optional default configuration from your site.json suggestion to the navbar, so we can keep this entirely on the frontend, and allow different default behaviours for different navbars.

we would still need the ability to adjust the behaviour at a _link level_ (`<a highlight-on="...">`) because a single navbar can still require different sorts of highlighting (a simple example being the ug - exact for `home`, `showcase`, `about` but sibling-or-child for `user guide`).

> shorten

the shortening here refers to your proposal for `data-highlight-on` on `<a>` tags, not `default-highlight-on` (since this is just written once-per-navbar) as well

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,103 @@
...
+          if (hParts[i] !== uParts[i]) {
+            return false;
+          }
+        }
+        return true;
+      },
+      isChild(url, href) {
+        const hParts = this.splitUrl(href);
+        const uParts = this.splitUrl(url);
+        if (hParts.length <= uParts.length) {
```
ang-zeyu, line 117: should this be reversed? (since we're checking whether the page you are at (url) is a _child_ of the link (href))

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,103 @@
...
+          return false;
+        }
+        for (let i = 0; i < uParts.length - 1; i++) {
```
ang-zeyu, line 120:
```suggestion
        for (let i = 0; i < hParts.length - 1; i++) {
```

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,103 @@
...
+            }
+          }
+        }
+        // fallback to user preference, otherwise
+        for (const li of navLis) {
+          const navLinks = Array.from(li.querySelectorAll('a.nav-link'));
+          const dropdownLinks = Array.from(li.querySelectorAll('a.dropdown-item'));
+          const allNavLinks = navLinks.concat(dropdownLinks).filter(a => a.href);
+          for (const a of allNavLinks) {
+            if (hlMode === 'sibling-or-child') {
```
ang-zeyu, line 154: in reference to https://github.com/MarkBind/markbind/pull/1379#discussion_r521406793, what needs to be checked is `<a highlight-on> attribute || this.highlightMode`

would need an additional if clause for `none` here as well

packages/vue-components/src/Navbar.vue
KendrickAng: Hmm I only had a navbar-level highlight-on configuration in mind, so there wasn't link-level highlighting

packages/vue-components/src/Navbar.vue
KendrickAng: Hmm could I ask why?

KendrickAng: Hmm ok, I think the implementation I had in mind was different; checking if href was a child of url (which was incorrect)

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,114 @@
...
+        for (const li of navLis) {
+          const navLinks = Array.from(li.querySelectorAll('a.nav-link'));
+          const dropdownLinks = Array.from(li.querySelectorAll('a.dropdown-item'));
+          const allNavLinks = navLinks.concat(dropdownLinks).filter(a => a.href);
+          for (const a of allNavLinks) {
+            const hlMode = a.getAttribute('data-highlight') || defHlMode;
+            if (hlMode === 'none') {
+              continue;
+            }
+            if (hlMode === 'exact') {
```
ang-zeyu, line 160: is this needed with the one above? (let's keep the one above if not since all exact checks should come first)

docs/userGuide/syntax/navBars.mbdf
```diff
@@ -69,13 +73,42 @@ If you wish to further customize your navbar beyond the primary, dark, and light
 <md>In `{your-site}/_markbind/headers/header.md`, you can change `<navbar type="dark/primary/light">` to `<navbar type="none" add-class="bg-warning/danger/info/primary/success">` to apply Bootstrap background styles.</md>
 </modal>
 
+**Navbars let you specify rules to customise link highlighting.**
```
ang-zeyu, line 76: Let's move this above the summary options table, since you've already mentioned `default-highlight-on` here

docs/userGuide/syntax/navBars.mbdf
```diff
@@ -69,13 +73,42 @@ If you wish to further customize your navbar beyond the primary, dark, and light
...
+- Define `default-highlight-on` in `<navbar>` to specify fallback highlight rules.
+- Define `data-highlight` in `<a>` tags with the class `nav-link` or `dropdown-item` to specify individual highlight rules.
+
+<include src="codeAndOutput.md" boilerplate >
```
ang-zeyu, line 80: since its a little hard to demonstrate this on one page, how about `<include>`ing the userGuide navbar? (just the code snippet) :-)

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,114 @@
...
+            }
+          }
+        }
+        // fallback to user preference, otherwise
+        for (const li of navLis) {
+          const navLinks = Array.from(li.querySelectorAll('a.nav-link'));
+          const dropdownLinks = Array.from(li.querySelectorAll('a.dropdown-item'));
+          const allNavLinks = navLinks.concat(dropdownLinks).filter(a => a.href);
+          for (const a of allNavLinks) {
+            const hlMode = a.getAttribute('data-highlight') || defHlMode;
```
ang-zeyu, line 156: hmm how about `highlight-on`? its only slightly shorter but would facilitate future attributes (`highlight-class`, `highlight-link`, ...)

packages/vue-components/src/Navbar.vue
```diff
@@ -71,12 +75,114 @@
...
+                return;
+              }
+            } else if (hlMode === 'child') {
+              if (this.isChild(url, a.href)) {
+                li.classList.add('current');
+                return;
+              }
+            } else {
+              // invalid option, ignore
+              return;
```
ang-zeyu, line 182:
```suggestion
              console.log(something);
```

so we can continue checking the rest of the links with valid options (if any)

ang-zeyu (commented):
couple last nits:

packages/vue-components/src/Navbar.vue
KendrickAng: I'm not too sure about this, but when I used `highlight-on` instead, the attribute was stripped away from the final result, but when I used data-highlight, it wasn't removed

ang-zeyu (approved):
Lgtm 👍 Thanks! @KendrickAng

PR #1383, opened 2020-11-07
Correct logging when there are no assets to be removed

### LINKED ISSUE #970: Adding a file is shown as 'assets removed' in the console
v2.8.0
When a file is added/removed while the author is in live-preview mode, MarkBind does a full rebuild of the previewed site, while showing a warning to the user.
image

In the warning message, `assets removed` is shown even when a file is added. Rephrase as `assets removed or added`?

Furthermore, the `assets removed` should appear above the `warn` rather than the below (because effect should follow the cause)?

ryoarmanda:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #970 

**Overview of changes:**
Added a check on whether there are no files to be removed on `_removeMultipleAssets`

- Opted to catch empty list of files at `_removeMultipleAssets` instead of its caller, `_rebuildSourceFiles`, to let the caller process without extra handling (the post-remove callback feels like it should be executed on either cases any way)
- In the issue itself, there were talks about relegating the "Assets removed" message to DEBUG level. I actually have thought of a way to make the CLI run on DEBUG level through a command option, but I realized that there are other messages on the same vein as this one (e.g. "Assets built" just above `_removeMultipleAssets`). Relegating only the removed message while leaving everything similar alone feels a bit weird (I feel like when one changes, everything should be changed as well), and I don't actually know if this should be within the scope of the PR. Should I include the change here?

**Testing instructions:**
With the test site, remove any page files. It should trigger the "Assets removed" message.

Then add back the file. Console should output other logs about the rebuild but the "Assets removed" one should not be present anymore.

Correct logging when there are no assets to be removed

---

### DIFF
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -785,8 +785,10 @@ class Site {
     const filesToRemove = fileRelativePaths.map(
       fileRelativePath => path.join(this.outputPath, fileRelativePath));
     const removeFiles = filesToRemove.map(asset => fs.remove(asset));
-    return Promise.all(removeFiles)
-      .then(() => logger.info('Assets removed'));
+    return removeFiles.length === 0
+      ? Promise.resolve('')
+      : Promise.all(removeFiles)
+        .then(() => logger.info('Assets removed'));
   }
 
   buildAssets() {
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -788,7 +788,7 @@ class Site {
     return removeFiles.length === 0
       ? Promise.resolve('')
       : Promise.all(removeFiles)
-        .then(() => logger.info('Assets removed'));
+        .then(() => logger.debug('Assets removed'));
   }
 
   buildAssets() {


ryoarmanda:
Hi @ang-zeyu can you review this? (very short I know 😅) Also any thoughts about the DEBUG level change I highlighted in the PR comment above? If it's okay to include it in this PR I will push the change.

ang-zeyu:
Nice work investigating this 🙂 

> Opted to catch empty list of files at _removeMultipleAssets instead of its caller, _rebuildSourceFiles, to let the caller process without extra handling (the post-remove callback feels like it should be executed on either cases any way)

agree, this should be the right fix

> Hi @ang-zeyu can you review this? (very short I know 😅) Also any thoughts about the DEBUG level change I highlighted in the PR comment above? If it's okay to include it in this PR I will push the change.

Hmm instead of a command option, how about just `logger.debug`? (it should show up in the log file but not in the console)
I think its definitely within scope though, no worries there

ryoarmanda:
> Hmm instead of a command option, how about just `logger.debug`? (it should show up in the log file but not in the console)

Yeah I was going for that as well. The command option thing is so that MarkBind can also run with the debug logs printed to the console, as discussed in the issue.

ang-zeyu:
> Yeah I was going for that as well. The command option thing is so that MarkBind can also run with the debug logs printed to the console, as discussed in the issue.

Sounds good, might be handy to have debug logs directly console printed for developer use. (maybe we can use the existing `--dev` option for `markbind serve` as well?)
In that case a separate PR for printing DEBUG level things to the console would be better though as you mentioned

ryoarmanda:
> Sounds good, might be handy to have debug logs directly console printed for developer use. (maybe we can use the existing `--dev` option for `markbind serve` as well?)

Yeah, that would be a good place.

> In that case a separate PR for printing DEBUG level things to the console would be better though as you mentioned

Agree, I'll just create another PR for it I guess. For now, I just set the "Assets removed" to DEBUG level.

ang-zeyu (approved):
Lgtm 👍

Issue #896, opened 2019-06-26
Better support for tables


damithc:
Current: we support GFMD syntax for tables
Suggestions for improvements:
* Ability to use more complex tables such as merged cells e.g., https://www.npmjs.com/package/markdown-it-multimd-table
* Ability to create a table based on content in a csv file
* Ability to sort columns
* Ability to specify larger amounts and different varieties of content for a cell (e.g., a panel within a cell)

ang-zeyu:
just a note: markdown-it-multimd-table's multiline feature (using `\` to continue lines) is wildly incompatible with our current process unfortunately, as we parse html before markdown. It would require a significant architectural change and breaking footnotes to accomodate it.
It is _toggleable_ though, so if the other features have no conflict we could still look to replace gfmd table with it

damithc:
I think asciidoc has a more powerful tables syntax. Of course unlikely we can reuse their parser but we can get some inspiration for the syntax.

Would using our own tag (like `<puml>`) to surround the table content allow us to use an entirely different syntax and parse it at a later stage? Not sure about the technical side but I had a feeling that @crphang was aiming for that kind of flexibility to use different syntax schemes for different components.

ang-zeyu:
unfortunately not as well, as you'd end up still having to parse the html inside the tag. (since tables can contain html)
The ability to treat custom tags like script/style tags is more apt for an proprietary syntax, like puml.

damithc:
> unfortunately not as well, as you'd end up still having to parse the html inside the tag. (since tables can contain html)
> The ability to treat custom tags like script/style tags is more apt for an proprietary syntax, like puml.

Yes, that's true.

ang-zeyu:
possible with processing order change in #1403

yiwen101:
Will work on adding a table component that has a "src" attribute pointing to relative path to then csv file.
Will try make the component to extensible.

Tim-Siu:
I will look into enabling asciidoc syntax for tables. It seems possible with the current pipeline.

Tim-Siu:
I am currently working on a baseline implementation that puts ASciidoc tables in a special tag like `<table-pro>`. It will then be processed by a plugin that works similarly to the `PlantUML` support. Users will have to download `Asciidoctor` first and then the plugin will use `Asciidoctor` to generate HTML from the Asciidoc syntax.

Tim-Siu:
As discussed with Prof. Damith, the two features of priority would be:
1. Making table sortable.
[...]

PR #1403, opened 2020-12-05
V3 baseline

### LINKED ISSUE #539: Allow reusing of files in page layouts
**Version: v1.15.2**

**Problem**

Currently page layouts would need to duplicate files if the exact same files are used as layouts are treated as stand-alone folders. 

**Potential solution**

> Introduce a new file e.g., settings.md into layouts that can contain this kind of settings?

To extend your idea, currently the layout files for footer and other items are separated. It would be neater if all the different files (footer, siteNav, css) are all stored in their respective folders and have a simple `settings.md` file to "store" the file names required. Basically abstracting layouts to just a front matter file.

E.G, content within `settings.md`:
```
css: colorTheme1.css, blackTheme2.css
footer: simpleFooter.md
pageNav: default
pageNavHeader: "This is a list of stuff"
siteNav: mySiteNav.md
script: myScript1.js, myScript2.js
```

Then all the different file types are just stored in their respective folders within `_/markbind`

Original comment by @Chng-Zhi-Xuan in https://github.com/MarkBind/markbind/pull/465#issuecomment-451625452_

### LINKED ISSUE #810: Refactor the Parser/Site/Page trinity into more comprehensible components

### LINKED ISSUE #972: Footnotes inside panels don't work
Example (see the footnote inside the _tutorials_ panel) https://nus-cs2113-ay1920s2.github.io/website/admin/weeklySchedule.html

image

### LINKED ISSUE #973: Support search bar in siteNav
v2.9.0
Current: the search bar is in the topNav
Problem: user needs to scroll to top before searching
Suggestion: provide a way to insert a search bar into the siteNav (which has a sticky behavior compared to topNav)

### LINKED ISSUE #1188: Add more hooks for plugins

### LINKED ISSUE #1240: layouts: don't include files that are missing
v2.14.1

Current: Even if a layout is missing a standard file e.g., styles.css, a link to the file is inserted into every page, resulting in the browser reporting 'missing file'. See screenshot (results of a tool checking for broken links)

image

Suggestion: don't insert a link to such files if the file is missing.

### LINKED ISSUE #1308: Clarify the requirements of layouts
Currently, layouts (header, footer, page (expressive) layouts, navigation) cannot:
- [ ] use include features (`<panel src="...">` & `<include>`) (with the exception of expressive layouts)
- [ ] use `<variable>` / `<import>` correctly (variables are not erased, imports do not work with live reload)
- [ ] use plugin elements correctly (as `preRender` is not executed for layouts)
- ❔ 

Is this by design? 

If not, a very large refactor may be due to fix the processing order

ang-zeyu:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [x] Others, please explain:

The architecture involves too much redundant processing; And the order of these processing stages is far too coupled.

This PR revolves around following this architecture: nunjucks -> markdown -> html, and also overhauls the layouts system in favour of one-file expressive layouts per #1308 at the same time (only because it makes the rewrite simpler =P)

Some plugin interface tweaks inline / required by the architecture simplification are also done

Resolves #1308 
Resolves #1240 
Resolves #539 
Resolves #1188 
Resolves #973 
Resolves #810 
Fixes #972 

Related:
#896 

**Overview of changes:**
- oop-fied codebase (new manager classes: `LayoutManager`, `ExternalManager`, `PluginManager` new object classes `Layout`, `External`, `Plugin`)
- simpler architecture that always follows: nunjucks -> markdown -> html
   - `nodePreprocessor` and `nodeProcessor` is combined into the html stage
- plugin system tweaks
- major performance improvements ~3x
- updated layouts docs

Todos:
- all done

**Testing instructions:**
manually tested layouts system

Rebase - commits need to be merged as one but are quite distinct

---

- [ ] No blatantly unrelated changes 🤔 bigger than I'd like but its far too coupled to deliver separately


ang-zeyu:
@damithc @acjh  any thoughts on the layouts changes?

https://deploy-preview-1403--markbind-master.netlify.app/userguide/tweakingthepagestructure

So far I've noticed:
- additional work for fixed headers is now needed - the `fixed-header-padding` class must be added by the user to the appropriate elements
- a heavy shift of layout related styles to userland - e.g. `#content-wrapper`, `#site-nav`, .... This now becomes the user's responsibility to maintain. (https://github.com/MarkBind/markbind/pull/1403/files#diff-309b964023ca899c9505205f36d3f4d5b36a6487e5c9b2e242204ee06bbc9ce9)

Its something we can reduce the work needed in template sites though, and may be warranted by the flexibility provided. (many interesting layouts will be made possible by this)

Not sure if there are more alternatives to alleviate this though

damithc:
> * additional work for fixed headers is now needed - the `fixed-header-padding` class must be added by the user to the appropriate elements

In every page? Or just in the layout pages?


> * a heavy shift of layout related styles to userland - e.g. `#content-wrapper`, `#site-nav`, .... This now becomes the user's responsibility to maintain. (https://github.com/MarkBind/markbind/pull/1403/files#diff-309b964023ca899c9505205f36d3f4d5b36a6487e5c9b2e242204ee06bbc9ce9)

I noticed `<span id="title" class="d-none">{{ title }}</span>` being added to many pages. Is it quick of your own site or something authors have to do in every page in every site?

docs/userGuide/usingComponents.md
```diff
@@ -1,9 +1,10 @@
-<variable name="title" id="title">Using Components</variable>
```
ang-zeyu, line 1: > In every page? Or just in the layout pages?

Just layouts : )

Still something extra to take note of for the author though

> > * a heavy shift of layout related styles to userland - e.g. `#content-wrapper`, `#site-nav`, .... This now becomes the user's responsibility to maintain. (https://github.com/MarkBind/markbind/pull/1403/files#diff-309b964023ca899c9505205f36d3f4d5b36a6487e5c9b2e242204ee06bbc9ce9)
> 
> I noticed `<span id="title" class="d-none">{{ title }}</span>` being added to many pages. Is it quick of your own site or something authors have to do in every page in every site?

This one would be belong to the page `<variable>` / `<import>` removal pr (to be split), there's no way to set the id for a nunjucks tag so had to create another `<span>`. The title `id` is used for the userguide macro that creates the left / right navigation buttons at the bottom

damithc: Got it.

ang-zeyu:
Any thoughts on the docs / feature set? @damithc 

https://github.com/MarkBind/markbind/pull/1403/commits/44d81670e7fba7edbccb4c7c81858d3e15351616

I'll give the code a few more looks here and there this week, then maybe merge it in this weekend otherwise.

damithc:
> Any thoughts on the docs / feature set? @damithc

Looks fine for me :+1: 
Any idea how much change this will require in CS2103 website?

ang-zeyu:
> Looks fine for me 👍
> Any idea how much change this will require in CS2103 website?

Hmm from what I can see there aren't a lot of layouts, so most of the work should be from converting page `<variable>` / `<import>`s to nunjucks `{% set / import %}`. Would depend on how many of those you have =X

For layouts you can re-run `markbind init` and copy over the default layout as a baseline, then 'simply `<include>`' them, and ctrl-R the layout names

Also the old attribute names which have been deprecated for a while now (`title`, `text`, ...) may be removed in a later PR (or we can postpone maybe, I don't think its affecting maintenance too much)

ang-zeyu:
Merging this in a little earlier in case anyone starts looking at the backend.

\*fingers crossed\*, but we can find (hopefully) and fix any issues before release

PR #1404, opened 2020-12-05
Fix inline code in dark mode when printing

### LINKED ISSUE #1401: Code blocks: inconsistent light/dark mode use when printing
v2.17.0

current: inline code remain in dark mode while regular code blocks change to light mode
suggested: both should use the light mode for printing

image

raysonkoh:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #1401 

**Overview of changes:**
Updated the css selector in `markbind.css`, under the media print query for code blocks, to additionally target inline code.

* The accuracy of the CSS selector used to target inline code - `code.hljs.inline`
* Color of inline-code might be too similar to the background color in `<box>`

image

**Testing instructions:** Ran `npm run test`, no errors.

**Proposed commit message: (wrap lines at 72 characters)**: Fix inline code in dark mode when printing

---


ang-zeyu:
Thanks! 👍 

> The accuracy of the CSS selector used to target inline code - code.hljs.inline

 looks good, I think this is as specific as it can get

> Color of inline-code might be too similar to the background color in `<box>`

Untitled

Rather, it may look too much like normal text. Should we add a border similar to code blocks?

damithc:
> Rather, it may look too much like normal text. Should we add a border similar to code blocks?

If this refers to inline code, a border may be too visually jarring. How about a slightly shaded bg?
Example (from GitHub):
>image

BTW, whichever it is, should be the same when using the light theme for the code, right?

ang-zeyu:
> BTW, whichever it is, should be the same when using the light theme for the code, right?

Not necessarily; I think its ok to provide only a simpler, at-least printable dark theme. 
Duplicating the light theme requires duplicating the styles which can be harder to maintain, and a more bloated bundle.

also when printing textbooks I really doubt many would print with colour anyway =P

raysonkoh:
I tried adding the same border as the code blocks, but with a smaller border-width size. Would this be okay?

image

damithc:
> Not necessarily; I think its ok to provide only a simpler, at-least printable dark theme.
> Duplicating the light theme requires duplicating the styles which can be harder to maintain, and a more bloated bundle.

I see. I was under the impression that we are switching to the light theme for printing but if that's not the case, a simple style fix is fine.

> also when printing textbooks I really doubt many would print with colour anyway =P

Ink saving applies to B&W too :-)

damithc:
> I tried adding the same border as the code blocks, but with a smaller border-width size. Would this be okay?
> 
> image
> Edit: lighter border-color

Looks OK to me. @ang-zeyu what do you think?

ang-zeyu:
Lgtm 👍 thanks! @raysonkoh

ang-zeyu (approved):

Issue #1201, opened 2020-04-16
Improve minimized design for panels


ang-zeyu:
**Is your request related to a problem?**

![minimalpanelminimized](https://user-images.githubusercontent.com/3306138/79480627-ef57ea80-8040-11ea-9adb-bc7162bf07aa.png)

**Describe the solution you'd like**

Starting to think that simply removing the border would be better. ( so you could have it sit in a paragraph seamlessly )

Some extra signifiers could be placed beside it so the user knows it can be clicked as well. ( e.g. mouse icon )

**Describe alternatives you've considered**
na

**Additional context**
na

damithc:
Yes, it would be good to have a different visualization for collapsed panels.
Note that collapsed panels we have are not a common thing. Ideally, it should appear different from other things. Right now they appear like buttons, which is not ideal either. They shouldn't appear like normal links either.

ang-zeyu:
🤔 perhaps we could borrow the design of tipbox icons here in place of the 'mouse icon' for all the designs. If there's no suitable ones, we could find more from `fas/octicons`
So each button would be borderless, and just paired with an icon of some type, maybe with some color / opacity as well.

Something along these lines that gives it a unique feel

damithc:
On the other hand, collapsed panels are used a lot in CS2103 website to embed content taken from elsewhere in the site. e.g.,
![image](https://user-images.githubusercontent.com/1673303/79490439-71e7a680-804f-11ea-952d-ad33de399300.png)

For those cases, the panel should not be too seamless so that embedded content can be easily spotted and not mistaken for native (i.e., not embedded) content.

KendrickAng:
Hi, @ang-zeyu @damithc, I'd like to try taking this as my first issue, would that be fine?

ang-zeyu:
Yes, sure!

Note that the frontend workflow we have been following up till now is manually copying the bundle (`components.min.js`) to the output directory of the site everytime we change source files, which is quite cumbersome. (though I think we all became pretty efficient at it =X)

I'll try to get #1306 (preferably with a review) in and add live / hot reload by this weekend (something like `markbind s -d`) if you prefer to wait, otherwise you could get started using the above workflow, or peek around some back end issues!

KendrickAng:
In that case, I'll try working on another issue first! But could I ask how copying `components.min.js` simulates hot reloading? Not too familiar with this aspect of things. Thank you!

ang-zeyu:
[...]

PR #1422, opened 2020-12-13
Branch improve minimal panel design

hhdqirui:

Provide a suggestion for #1201

**Overview of changes:**

Add an icon to the minimal panel to indicate that the minimal panel is clickable so that user would be aware of that.
Adjust the margin of the minimal panel so that the minimal panels do not clutter together.

image

Add icon and adjust margin for minimal panel.

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -4,2986 +4,670 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"jest": {
-			"version": "25.5.4",
-			"resolved": "https://registry.npmjs.org/jest/-/jest-25.5.4.tgz",
-			"integrity": "sha512-hHFJROBTqZahnO+X+PMtT6G2/ztqAZJveGqz//FnWWHurizkD05PQGzRZOhF3XP6z7SJmL+5tCfW8qV06JypwQ==",
+		"@babel/code-frame": {
+			"version": "7.10.4",
+			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
+			"integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
 			"dev": true,
 			"requires": {
-				"@jest/core": "^25.5.4",
-				"import-local": "^3.0.2",
-				"jest-cli": "^25.5.4"
+				"@babel/highlight": "^7.10.4"
+			}
+		},
+		"@babel/core": {
+			"version": "7.12.10",
+			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.10.tgz",
+			"integrity": "sha512-eTAlQKq65zHfkHZV0sIVODCPGVgoo1HdBlbSLi9CqOzuZanMv2ihzY+4paiKr1mH+XmYESMAmJ/dpZ68eN6d8w==",
+			"dev": true,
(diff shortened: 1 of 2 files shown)


packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -1,9 +1,18 @@
 <template>
   <span ref="cardContainer" :class="['card-container', addClass]">
     <div v-show="localMinimized" class="morph">
-      <button :class="['morph-display-wrapper', 'btn', btnType, 'card-title']" @click="open()">
+      <button
+        :class="['morph-display-wrapper', 'btn', btnType, 'card-title']"
+        @click="open()"
+        id="minimised-panel"
```
ang-zeyu, line 7: hmm would the existing `morph-display-wrapper` class work instead? since we have multiple of these with the same `id`s

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -1,9 +1,18 @@
...
+      >
         <slot name="_alt">
           <slot name="header"></slot>
+          <span
+            data-v-5681bde8=""
```
ang-zeyu, line 12: 👀 what's this for?

packages/cli/package-lock.json
```diff
@@ -4,4680 +4,6767 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"jest": {
-			"version": "25.5.4",
-			"resolved": "https://registry.npmjs.org/jest/-/jest-25.5.4.tgz",
-			"integrity": "sha512-hHFJROBTqZahnO+X+PMtT6G2/ztqAZJveGqz//FnWWHurizkD05PQGzRZOhF3XP6z7SJmL+5tCfW8qV06JypwQ==",
+		"@babel/code-frame": {
```
ang-zeyu, line 7: you can discard the one-time diff from your first `npm link` here

ang-zeyu (commented):
Thanks for contributing!

Wondering if we should provide some way to toggle the icon off / customise the icon as well, or just leave it to the existing `alt` attribute. @damithc @hhdqirui

packages/cli/package-lock.json
hhdqirui: Done.

packages/vue-components/src/panels/NestedPanel.vue
hhdqirui: Forgot to remove this.

packages/vue-components/src/panels/NestedPanel.vue
hhdqirui: Yes, it works and it is fixed.

damithc:
Sorry @hhdqirui for not stepping in earlier. I'm not sure if adding an icon is a good idea. Furthermore, it increases the height without much benefit. If at all, the icon should be in the same line. Still, given this is supposed to be the 'minimal' form, I'm wary of them taking more space than necessary. 
@ang-zeyu Shall we put the PR and the issue on hold as we are not entirely sure what we want?

ang-zeyu:
ok, rather difficult design issue =C

requirements:
- minimal (providing greater edit flexibility)
- dosen't feel like a button
- but user must know it can be clicked
- ❔

any ideas? @hhdqirui @damithc

damithc:
> * dosen't feel like a button

Looking like a button is not half bad, actually. For one thing, the reader will know it's something to click. Plus, given the context, the reader will know it's not really a conventional button. While we can add an arrow icon, I'm not sure having such will add value in cases like this:
image

ong6:
Closing this issue since it seems to be resolved by @kaixin-hc in PR #1694.

PR #1425, opened 2020-12-14
Implement peek feature for panels

### LINKED ISSUE #905: Add a 'peek' mode for panels
Current: panels have expanded, collapsed, and minimized modes.
Suggestion: we can also support a 'peek' (or 'preview') where just a bit of the content is shown.

e.g., 
image

or
image

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

This PR resolves #905. 

**Overview of changes:**
Introduced various variables and directives to implement the 'peek' feature for panels.

I'm quite unsure of the best approach to implement this feature. For my approach, I decided to make localExpanded to always have a true value whenever the panel is on show-preview mode. Thus, the content of the panel will always be displayed (effectively overriding the v-show). I then use css to achieve the effect of a 'peek' mode. To be specific, I used css transition, max-height, and overflow hidden to toggle between collapsed ('peek' mode) and expanded panel. 

I also tried to make the behaviour of the 'peek' mode panel as separate from the original panel's toggling behaviour as much as possible (as can be seen from the conditions used for the click behaviours, e.g. "@click.stop="shouldShowPreview ? closePreviewCard() : close()", this is to prevent conflicts between the original panel behaviours from the peek mode panel behaviours). 

However, I do face various issues in terms of nested panels. As my css transition relies heavily on max-height, we always need to know the expanded size of a panel before we can apply the transition smoothly (the default requirement for css transition for max-height). In nested panels, we have many unexpanded panels. Suppose there are multiple nested panels, it is difficult to set an **accurate** max-height for the panel we are concerned with. For example, if we look at the first layer panel, it can adopt different heights (depending on how many nested panels we expand). Thus, it is difficult to set an accurate max-height for css transition to achieve a smooth transition (the max-height has to be as accurate as possible for the transition to not be 'jerky'). 

At the moment, nested panels can be quite buggy in terms of transitions. But the 'peek' functionality works fine. 

Beyond the nested panels issue, I think generally it works alright. However, more tests have to be conducted to be sure. 

It would be great if there are some suggestions on how I can improve the implementation (can be in terms of code quality or even a better way of approaching the problem), or resolve the issue regarding nested panels. 

For now, I will continue to see how I can resolve the issue and think of a better way to implement this feature :-)

@ang-zeyu would you have any advice for me? 

**Testing instructions:**
`npm run test`

Using the existing user guide, add the 'show-preview' attribute to any panels and serve the program, you should be able to see the preview mode that looks something like below. 

Screen Shot 2020-12-15 at 12 38 38 AM

Implement peek feature for panels

To find out what content a panel contains, a user has to expand it. 
This may be inconvenient when the user just wants to have an idea 
of whether the panel contains relevant content. 

Let's implement a feature to let users have a glimpse of the content, 
which would help them decide whether it is worth to expand the panel. 

---


packages/vue-components/src/panels/PanelBase.js
```diff
@@ -53,6 +53,22 @@ export default {
       type: Boolean,
       default: false,
     },
+    showPreview: {
+      type: Boolean,
+      default: false,
+    },
+    previewCardExpanded: {
```
ang-zeyu, line 60: hmm are these props meant to be configurable?
if not, let's use `data()` per vue's recommendations.

ang-zeyu (commented):
Thanks! @wxwxwxwx9 

The design looks great 👍 
Just an overall review on the approach first:

> I also tried to make the behaviour of the 'peek' mode panel as separate from the original panel's toggling behaviour as much as possible

Regarding this, let's not shy away from making changes to the existing code. Max-height transitions generally all have the same logic surrounding `scrollHeight` and `maxHeight` - while this prevents regressions to the existing behaviour in the short term it may be quite troublesome to maintain two nearly identical sets of transition code in the long run.

> However, I do face various issues in terms of nested panels. As my css transition relies heavily on max-height, we always need to know the expanded size of a panel before we can apply the transition smoothly (the default requirement for css transition for max-height). In nested panels, we have many unexpanded panels. Suppose there are multiple nested panels, it is difficult to set an accurate max-height for the panel we are concerned with. For example, if we look at the first layer panel, it can adopt different heights (depending on how many nested panels we expand). Thus, it is difficult to set an accurate max-height for css transition to achieve a smooth transition (the max-height has to be as accurate as possible for the transition to not be 'jerky').

I'm assuming you're referring to panels with `src` attribute.

The only alternative I can see is setting up [provide / inject](https://v3.vuejs.org/guide/component-provide-inject.html) between panel components, then communicating child heights to parent heights when the child loads, then adding the height, ...
But then you also have to consider other components which may load content dynamically (which may be *anything* under the sun).

Let's just let the user absorb the jerky cost of dynamically loading content here.

damithc:
@wxwxwxwx9 Good progress. Note that some amount of panel content should be visible already, or else it's no different from a collapsed panel.

wxwxwxwx9:
@damithc currently I'm applying a fixed preview height. So if the user has multiple new lines before the actual content, the preview might not be able to display the content. 

Do you think the preview height should be dynamic (to ensure that part of the actual content is displayed)? It might challenging to implement but I can look into it.

wxwxwxwx9:
> Thanks! @wxwxwxwx9
> 
> The design looks great 👍
> Just an overall review on the approach first:
> 
> > I also tried to make the behaviour of the 'peek' mode panel as separate from the original panel's toggling behaviour as much as possible
> 
> Regarding this, let's not shy away from making changes to the existing code. Max-height transitions generally all have the same logic surrounding `scrollHeight` and `maxHeight` - while this prevents regressions to the existing behaviour in the short term it may be quite troublesome to maintain two nearly identical sets of transition code in the long run.

Hmm, I'm thinking that it would not be possible to have a single set of transition code for both preview panels and normal panels. I might be wrong but let me explain my thought process. Currently, the normal panel is using v-show and the localExpanded boolean variable to aid the transition. The v-show directive either shows 0% of the panel content or 100% of the panel content. And the v-show directive is the crucial part that enables the transition feature for the normal panels. 

However, for my preview panel, I have to show, say, 10% of the content as the preview. But this is not achievable with v-show (correct me if I'm wrong). Thus, what I did was to set localExpanded to true by default, and at all times for preview panels. This is to ensure that v-show is displaying 100% of the content. Then, I make use of css to help me toggle between 10% of the content shown and 100% of the content shown. 

The reason for this is because v-show can't show 10% of the content (it is either 0% or 100%), thus I have to workaround this limitation of v-show and make v-show display 100% of the content at all times, and work on top of it. This also means that I can no longer use the normal transition feature that is coupled with v-show. Therefore, I need to create a separate set of transition code to handle my preview panel's transition. 

I'm not sure if there's a better way to implement this. What do you think? :O @ang-zeyu

damithc:
> @damithc currently I'm applying a fixed preview height. So if the user has multiple new lines before the actual content, the preview might not be able to display the content.

Fixed height is fine. just need to ensure the default fixed height is enough for some content to be visible in all typical cases. It's too bad if the author puts empty space at the top of the panel.
We can also consider letting the user override the height, but that can be a separate PR.

damithc:
> @wxwxwxwx9 Good progress. Note that some amount of panel content should be visible already, or else it's no different from a collapsed panel.

When I wrote the above comment, I mistakenly thought the nested panel is the one in the peek mode. I now realize that the outer parent panel is in the peek mode and some content are visible already.


> Screen Shot 2020-12-15 at 12 38 38 AM

ang-zeyu:
> I'm not sure if there's a better way to implement this. What do you think? :O @ang-zeyu

hmm good point, could we instead discard the use of `<transition>` + `v-show`? Since it seems to be preventing the implementation here.

My pov is that this feature is basically the `expanded` option, just starting out at a different `maxHeight`. It should be feasible to change the existing implementation hence and keep all the transitions.

Discarding `<transition>` + `v-show` (we may replace this with our own `style=display:none` + etc.) is ok if needed to provide more flexibility here

wxwxwxwx9:
> > @damithc currently I'm applying a fixed preview height. So if the user has multiple new lines before the actual content, the preview might not be able to display the content.
> 
> Fixed height is fine. just need to ensure the default fixed height is enough for some content to be visible in all typical cases. It's too bad if the author puts empty space at the top of the panel.

Got it! 

> We can also consider letting the user override the height, but that can be a separate PR.

That's a good idea. Will work on it in my next PR. 

> > I'm not sure if there's a better way to implement this. What do you think? :O @ang-zeyu
> 
> hmm good point, could we instead discard the use of `<transition>` + `v-show`? Since it seems to be preventing the implementation here.
> 
> My pov is that this feature is basically the `expanded` option, just starting out at a different `maxHeight`. It should be feasible to change the existing implementation hence and keep all the transitions.
> 
> Discarding `<transition>` + `v-show` (we may replace this with our own `style=display:none` + etc.) is ok if needed to provide more flexibility here

Yep, sounds like a good approach to achieve a single transition for our panels. Will work on this implementation 👍

wxwxwxwx9:
> There's also the concern where if the user doesn't have enough content, then the preview might not be very meaningful. For example, if we have a preview of height 100px, but user only placed a content of 50px in the panel, the preview essentially would display all of the content. And when the panel is expanded, we would go from height 100px to 50px. This is a bit jarring and weird from a user experience perspective, in my opinion. How can we handle this?
> 
> Should we let it behave as it is and treat it as a soft failure? Or should we, by default, disallow preview mode if their content is below a certain threshold in terms of height? For example, having only 50px of content when the preview is a fixed 100px in height.
> 
> I have attached some screenshots below to illustrate this problem.
> 
> Before expansion:
> Screen Shot 2020-12-16 at 1 12 17 PM
> 
> After expansion:
> Screen Shot 2020-12-16 at 1 12 48 PM

@damithc @ang-zeyu do you have any opinion on this? :O

damithc:
> There's also the concern where if the user doesn't have enough content, then the preview might not be very meaningful. For example, if we have a preview of height 100px, but user only placed a content of 50px in the panel, the preview essentially would display all of the content. And when the panel is expanded, we would go from height 100px to 50px. This is a bit jarring and weird from a user experience perspective, in my opinion. How can we handle this?
> 
> Should we let it behave as it is and treat it as a soft failure? Or should we, by default, disallow preview mode if their content is below a certain threshold in terms of height? For example, having only 50px of content when the preview is a fixed 100px in height.

Don't worry about this case. The author shouldn't use the peek mode for panels without enough content. It's a judgement call the author has to make.

wxwxwxwx9:
@ang-zeyu Hi Zeyu! This is my first implementation of the panel preview feature after the rewrite of the panel transition code. After testing it with the different scenarios and variations (preload, src, expanded, minimized, isMinimal), I think it works pretty well. 

I have improved the implementation and used position absolute on the preview chevron to ensure that it doesn't take up space in the DOM. 

Do let me know if there is anything I missed out or if any of my implementations can be improved on. :-)

Also, I have discussed with Prof. Damith about the minimal panel with preview mode. For reference, the minimal panel with preview mode looks something like below:

Screen Recording 2021-01-06 at 8.23.54 PM.mov

Although preview mode on minimal feature doesn't look the best (and not that bad), we agreed that we will roll out the preview feature for it as well and let the user decide on the his/her own use case. What do you think?

wxwxwxwx9:
Also another minor issue: for the collapsedPanelHeight, there currently is some semblance of magic number and I was thinking whether we should create another variable in data for it. 

```
    initPanelCollapsedHeight() {
      this.collapsedPanelHeight = this.showPreview ? 125 : 0;
    },
```

wxwxwxwx9:
@ang-zeyu My `package-lock.json` currently has a lot of changes after running `npm run setup`. Should this actually be happening? :O From what I can find out, it seems like it is expected behaviour but the diff seems to be too large and I'm a little concerned. Do you know anything about this?

ang-zeyu:
> @ang-zeyu My `package-lock.json` currently has a lot of changes after running `npm run setup`. Should this actually be happening? :O From what I can find out, it seems like it is expected behaviour (correct me if I'm wrong) but the diff is quite huge and I'm a little concerned. Do you know anything about this?

hmm there shouldn't be any changes; try resetting the package-lock, deleting *all* node_modules folders and re-running npm run setup

ang-zeyu:
> Also another minor issue: for the collapsedPanelHeight, there currently is some semblance of magic number and I was thinking whether we should create another variable in data for it.
> 
> ```
>     initPanelCollapsedHeight() {
>       this.collapsedPanelHeight = this.showPreview ? 125 : 0;
>     },
> ```
> 
> Do you have any opinions about this?

`data()` should be used for reactive state, so we can reason about what changes in a component easily.

that aside, numbers are magic numbers when the intent is not clear. In this case its purpose is sufficiently clear in my opinion.
Moving this to a named constant increases the verbosity of the code unnecessarily, also since it is only used in _one place_

How about a [computed property](https://vuejs.org/v2/guide/computed.html) though?

wxwxwxwx9:
> > @ang-zeyu My `package-lock.json` currently has a lot of changes after running `npm run setup`. Should this actually be happening? :O From what I can find out, it seems like it is expected behaviour (correct me if I'm wrong) but the diff is quite huge and I'm a little concerned. Do you know anything about this?
> 
> hmm there shouldn't be any changes; try resetting the package-lock, deleting _all_ node_modules folders and re-running npm run setup

There was some issues with my npm and after I reinstalled and updated it, and followed your instructions here, package-lock is now back to normal. Thanks so much for the advice!

> > Also another minor issue: for the collapsedPanelHeight, there currently is some semblance of magic number and I was thinking whether we should create another variable in data for it.
> > ```
> >     initPanelCollapsedHeight() {
> >       this.collapsedPanelHeight = this.showPreview ? 125 : 0;
> >     },
> > ```
> > 
> > 
> > Do you have any opinions about this?
> 
> `data()` should be used for reactive state, so we can reason about what changes in a component easily.
> 
> that aside, numbers are magic numbers when the intent is not clear. In this case its purpose is sufficiently clear in my opinion.
> Moving this to a named constant increases the verbosity of the code unnecessarily, also since it is only used in _one place_
> 
> How about a [computed property](https://vuejs.org/v2/guide/computed.html) though?

I see, thanks for sharing your insight! I agree with them. I think that maybe we can leave this out for now :-)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -175,6 +182,10 @@ export default {
         // Only preload, do not expand the panel.
         return;
       }
+      if (this.showPreview && !this.localExpanded) {
```
ang-zeyu, line 185: let's tighten the above branch as well with `localExpanded`, since `wasRetrieverLoaded` was modified

could move to self-explanatory named constants here as well in place of the comments and multiple branches

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -145,6 +159,46 @@ export default {
...
+        text-align: center;
+        bottom: 10px;
+    }
+
+    .preview-read-more:hover {
+        cursor: pointer;
+        opacity: 0.4;
+    }
+
+    .preview-read-more-fade-enter-active, .preview-read-more-fade-leave-active {
```
ang-zeyu, line 194: could move `transition: opacity 0.5s;` into the main `preview-read-more` class

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -145,6 +159,46 @@ export default {
...
+        transition: opacity 0.5s;
+    }
+
+    .preview-read-more-fade-enter, .preview-read-more-fade-leave-to {
```
ang-zeyu, line 198: stylelint error here

ang-zeyu (commented):
nice, some suggestions:

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -53,6 +53,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    showPreview: {
```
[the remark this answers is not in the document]
wxwxwxwx9: Doesn't that mean the user will have to indicate "peek" instead of "show-preview" to use the preview feature?

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: So sorry, I don't quite understand the first part, "tighten the above branch as well with localExpanded, since wasRetrieverLoaded was modified". Can you elaborate more on this? :O

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -232,8 +246,10 @@ export default {
       this.localExpanded = false;
     }
 
-    this.wasRetrieverLoaded = this.localExpanded;
+    this.wasRetrieverLoaded = this.localExpanded || this.showPreview;
     this.localMinimized = this.minimizedBool;
+
+    this.initPanelCollapsedHeight();
```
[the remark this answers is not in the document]
wxwxwxwx9: Alright! 👍

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
ang-zeyu: yup, suggesting a user side change here (less to type)

packages/vue-components/src/panels/PanelBase.js
ang-zeyu: this branch `if (this.preloadBool && !this.wasRetrieverLoaded) {`*

you may want to double check with a boolean expression simplifier though

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
wxwxwxwx9: Hmm, I'm thinking that it may be a little confusing for the user. If we were to go ahead with this, then I think perhaps we should package/name the feature as "peek" instead of "preview" (that means in the user guide we will strictly use "peek" to refer to this feature). 

However, I personally think "preview" sounds better but this is very subjective so I am open to both :-)

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: I see! Thanks for spotting this :-) I will make the changes!

packages/vue-components/src/panels/NestedPanel.vue
wxwxwxwx9: Fixed!

packages/vue-components/src/panels/NestedPanel.vue
wxwxwxwx9: Ah, thanks for the suggestion. It makes the code more concise. Will make the changes.

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
ang-zeyu: > we should package/name the feature as "peek" instead of "preview" 

yup.

or just `preview` would be good too (vs `show-preview`)

I'm open to either as well, maybe we can get a user opinion here? 🙂
@damithc

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Hmm, after considering your comments and reviewing my code, I realised that all the conditions above are unnecessary and I can just use `(!this.localExpanded)`, since the purpose of the function is to update maxHeight after loading src. 

If the panel is not expanded then, we can simply return the function , regardless of whether it is preload or preview. 

Thanks for the suggestion to improve the code quality anyway 👍

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
wxwxwxwx9: I'll shorten `show-preview` to `preview` in the meantime, while waiting for Prof's opinion.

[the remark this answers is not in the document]
damithc: I'm OK with either `preview` or `peek`. I thought `peek` is a better fit because it feels like _showing only a bit_ but I could be wrong. Also, the word _preview_ might be used in other contexts, e.g., _live preview_ of pages via the `markbind serve` command.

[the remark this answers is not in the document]
wxwxwxwx9: Hmm, considering that the word "preview" is used in other contexts, now I am leaning towards "peek". 

I'll change it to "peek" then :-)

wxwxwxwx9:
@ang-zeyu Hi Zeyu, this PR is ready for your review :-)

docs/userGuide/syntax/panels.mbdf
```diff
@@ -40,12 +40,31 @@
...
 <variable name="highlightStyle">html</variable>
 <variable name="code">
-<panel header="This header will only show when the Panel is minimized" expand-headerless>
+<panel header="This header will only show when the Panel is collapsed" expand-headerless>
   Lorem ipsum ...
 </panel>
 </variable>
 </include>
 
+**With the `peek` attribute, you may showcase part of your content without expanding the panel.**
```
ang-zeyu, line 49: let's update the options table below too

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -93,14 +97,19 @@ export default {
     shouldShowHeader() {
       return (!this.localExpanded) || (!this.expandHeaderless);
     },
+    shouldShowPeek() {
+      return this.peek && !this.localExpanded;
+    },
+    getCollapsedPanelHeight() {
```
ang-zeyu, line 103: let's name it `collapsedPanelHeight` per the examples in the vue docs instead

you may also have setters for computed properties https://vuejs.org/v2/guide/computed.html#Computed-Setter

ang-zeyu (approved):
Lgtm 👍

PR #1430, opened 2020-12-21
Rewrite panel transition code

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

This PR is meant to rewrite some of the panel transition code and lay the groundwork to implement a cleaner version of the panel preview feature (related to #1425). 

**Overview of changes:**
Removed the usage of v-show for the panel refs for both nestedPanel and minimalPanel. Implemented new transition code with javascript and CSS (max-height). 

I tried to ensure that there is no regression to the existing functionalities for my rewrite. There were some code that I have removed, like the event @src-loaded and setMaxHeight from the retriever, as I did not see the need for it for my implementation. However, I'm unsure if I am right in doing that or whether it would cause regressions in ways that I couldn't foresee. Do let me know if I am breaking any existing features by doing that :O

Also, perhaps my code quality can be improved (e.g. the way I implement certain functionalities). So, I think that may be one area to look at. 

I have written some comments in my code to explain the rationale for some of my implementations. 

**Testing instructions:**
`markbind serve` and test the panels on the user guide section. 

Rewrite panel transition for preview feature

It is not easy to implement the panel preview feature with the existing
transition code in our panels due to the use of v-show.

Let's rewrite some of the transition code and implement our own set
of transition code to allow for more flexibility. This will allow a cleaner
implementation of the panel preview feature. 

---

- [ ] No blatantly unrelated changes

### DIFF
--- a/packages/vue-components/src/Retriever.vue
+++ b/packages/vue-components/src/Retriever.vue
@@ -62,14 +62,12 @@ export default {
             template: `<div>\n${result}\n</div>`,
           });
           new TempComponent().$mount(this.$el);
-          this.$emit('src-loaded');
         })
         .fail((error) => {
           // eslint-disable-next-line no-console
           console.error(error.responseText);
           this.$el.innerHTML
               = `<strong>Error</strong>: Failed to retrieve content from source: <em>${this.src}</em>`;
-          this.$emit('src-loaded');
         });
     },
   },
--- a/packages/vue-components/src/Retriever.vue
+++ b/packages/vue-components/src/Retriever.vue
@@ -62,12 +62,14 @@ export default {
             template: `<div>\n${result}\n</div>`,
           });
           new TempComponent().$mount(this.$el);
+          this.$emit('src-loaded');
         })
         .fail((error) => {
           // eslint-disable-next-line no-console
           console.error(error.responseText);
           this.$el.innerHTML
               = `<strong>Error</strong>: Failed to retrieve content from source: <em>${this.src}</em>`;
+          this.$emit('src-loaded');
         });
     },
   },
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -1,7 +1,7 @@
 <template>
   <span ref="cardContainer" :class="['card-container', addClass]">
     <div v-show="localMinimized" class="morph">
-      <button class="morph-display-wrapper btn card-title morph-title" @click="open()">
+      <button class="morph-display-wrapper btn card-title morph-title" @click="minimalOpen()">
         <slot name="_alt">
           <slot name="header"></slot>
         </slot>
@@ -11,7 +11,7 @@
       <div
         :class="['header-wrapper',
                  { 'header-wrapper-bottom': isHeaderAtBottom, 'header-toggle': isExpandableCard }]"
-        @click.prevent.stop="isExpandableCard && toggle()"
+        @click.prevent.stop="isExpandableCard && minimalToggle()"
       >
         <transition name="header-fade">
           <span
@@ -33,7 +33,7 @@
               v-show="!noCloseBool"
               class="minimal-button"
               type="button"
-              @click.stop="close()"
+              @click.stop="minimalClose()"
             >
               <span class="glyphicon glyphicon-remove minimal-close-button" aria-hidden="true"></span>
             </button>
@@ -55,32 +55,20 @@
           </slot>
         </div>
       </div>
-      <transition
-        v-if="preloadBool || wasRetrieverLoaded"
-        @before-enter="beforeExpandMinimal"
-        @enter="duringExpand"
-        @after-enter="afterExpand"
-        @before-leave="beforeCollapse"
-        @leave="duringCollapse"
-        @after-leave="afterCollapseMinimal"
+      <div
+        ref="panel"
+        class="card-collapse"
       >
-        <div
-          v-show="localExpanded"
-          ref="panel"
-          class="card-collapse"
-        >
-          <div class="card-body">
-            <slot></slot>
-            <retriever
-              v-if="hasSrc"
-              ref="retriever"
-              :src="src"
-              :fragment="fragment"
-              @src-loaded="setMaxHeight"
-            />
-          </div>
+        <div class="card-body">
+          <slot></slot>
+          <retriever
+            v-if="hasSrc"
+            ref="retriever"
+            :src="src"
+            :fragment="fragment"
+          />
         </div>
-      </transition>
+      </div>
     </div>
   </span>
 </template>
@@ -110,15 +98,31 @@ export default {
     },
   },
   methods: {
-    beforeExpandMinimal(el) {
-      this.isHeaderAtBottom = true;
-      this.beforeExpand(el);
+    minimalToggle() {
+      if (this.localExpanded) {
+        // this is a collapse, set isHeaderAtBottom to true only at the end of transition
+        // to achieve the correct collapse transition effect of minimal panel
+        const onCollapseDone = () => {
+          this.isHeaderAtBottom = !this.isHeaderAtBottom;
+          this.$refs.panel.removeEventListener('transitionend', onCollapseDone);
+        };
+        this.$refs.panel.addEventListener('transitionend', onCollapseDone);
+      } else {
+        // this is an expansion, immediately set isHeaderAtBottom to true
+        this.isHeaderAtBottom = !this.isHeaderAtBottom;
+      }
+      this.toggle();
     },
-    afterCollapseMinimal() {
+    minimalOpen() {
+      this.open();
+      this.isHeaderAtBottom = true;
+    },
+    minimalClose() {
+      this.close();
       this.isHeaderAtBottom = false;
     },
   },
-  created() {
+  mounted() {
     this.isHeaderAtBottom = this.localExpanded;
   },
 };
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -69,7 +69,7 @@
             ref="retriever"
             :src="src"
             :fragment="fragment"
-            @src-loaded="setMaxHeight"
+            @src-loaded="retrieverUpdateMaxHeight"
           />
         </div>
       </div>
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -57,10 +57,10 @@
       </div>
       <div
         ref="panel"
-        v-if="wasRetrieverLoaded || preloadBool"
         class="card-collapse"
       >
         <div
+          v-if="wasRetrieverLoaded || preloadBool"
           class="card-body"
         >
           <slot></slot>
@@ -103,19 +103,6 @@ export default {
   },
   methods: {
     minimalToggle() {
-      // if (!this.wasRetrieverLoaded) {
-      //   console.log(this.$refs.panel.style.maxHeight);
-      //   this.wasRetrieverLoaded = true;
-      //   this.localExpanded = !this.localExpanded;
-      //   this.isHeaderAtBottom = !this.isHeaderAtBottom;
-      //   this.$nextTick(() => {
-      //     this.$refs.panel.style.maxHeight = `${0}px`;
-      //     this.$nextTick(() => {
-      //       this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
-      //     });
-      //   });
-      //   return;
-      // }
       if (this.localExpanded) {
         // this is a collapse, set isHeaderAtBottom to true only at the end of transition
         // to achieve the correct collapse transition effect of minimal panel
@@ -131,20 +118,6 @@ export default {
       this.toggle();
     },
     minimalOpen() {
-      // if (!this.wasRetrieverLoaded) {
-      //   console.log(this.$refs.panel.style.maxHeight);
-      //   this.wasRetrieverLoaded = true;
-      //   this.localExpanded = true;
-      //   this.localMinimized = false;
-      //   this.isHeaderAtBottom = true;
-      //   this.$nextTick(() => {
-      //     this.$refs.panel.style.maxHeight = `${0}px`;
-      //     this.$nextTick(() => {
-      //       this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
-      //     });
-      //   });
-      //   return;
-      // }
       this.open();
       this.isHeaderAtBottom = true;
     },
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -89,9 +89,9 @@ export default {
   data() {
     return {
       /*
-      'Copy' of localExpanded that is updated on certain animation events.
-      It is minimal-panel specific due to its design (the header can shift to the bottom).
-      Its purpose is to show the header text only once the collapse animation has finished.
+        'Copy' of localExpanded that is updated on certain animation events.
+        It is minimal-panel specific due to its design (the header can shift to the bottom).
+        Its purpose is to show the header text only once the collapse animation has finished.
        */
       isHeaderAtBottom: false,
     };
@@ -104,15 +104,18 @@ export default {
   methods: {
     minimalToggle() {
       if (this.localExpanded) {
-        // this is a collapse, set isHeaderAtBottom to true only at the end of transition
-        // to achieve the correct collapse transition effect of minimal panel
+        /*
+          This is a collapse.
+          Set isHeaderAtBottom to true only at the end of transition.
+          So that we can achieve the correct collapse transition effect of minimal panel.
+        */
         const onCollapseDone = () => {
           this.isHeaderAtBottom = false;
           this.$refs.panel.removeEventListener('transitionend', onCollapseDone);
         };
         this.$refs.panel.addEventListener('transitionend', onCollapseDone);
       } else {
-        // this is an expansion, immediately set isHeaderAtBottom to true
+        // This is an expansion. Set isHeaderAtBottom to true *immediately*.
         this.isHeaderAtBottom = true;
       }
       this.toggle();
@@ -126,7 +129,7 @@ export default {
       this.isHeaderAtBottom = false;
     },
   },
-  mounted() {
+  created() {
     this.isHeaderAtBottom = this.localExpanded;
   },
 };
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -56,6 +56,7 @@
         </div>
       </div>
       <div
+        v-if="wasRetrieverLoaded || preloadBool"
         ref="panel"
         class="card-collapse"
       >
@@ -66,6 +67,7 @@
             ref="retriever"
             :src="src"
             :fragment="fragment"
+            @src-loaded="setMaxHeight"
           />
         </div>
       </div>
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -56,11 +56,13 @@
         </div>
       </div>
       <div
-        v-if="wasRetrieverLoaded || preloadBool"
         ref="panel"
+        v-if="wasRetrieverLoaded || preloadBool"
         class="card-collapse"
       >
-        <div class="card-body">
+        <div
+          class="card-body"
+        >
           <slot></slot>
           <retriever
             v-if="hasSrc"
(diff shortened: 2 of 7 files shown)


packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,85 @@ export default {
...
     },
     openPopup() {
       window.open(this.popupUrl);
     },
-    setMaxHeight() {
-      // Don't play the transition for this case as the loading should feel 'instant'
-      if (this.expandedBool) {
-        this.$refs.panel.style.maxHeight = 'none';
+    setExpandedPanelHeight() {
+      this.expandedPanelHeight = this.$refs.panel.scrollHeight;
```
ang-zeyu, line 152: hmm would this ever not be `this.$refs.panel.scrollHeight`? (also taking into account the panel preview pr)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,85 @@ export default {
...
+      if (this.localExpanded) {
+        // COLLAPSE
+        // panel maxHeight is 'none' at this moment (as event listener set it to 'none' after expansion)
+        // thus, we need to reset the maxHeight to its current height (for collapse transition to work)
+        this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
+        // we need to reset / 'memorize' the expandedPanelHeight everytime we collapse
+        // this is because we may have expanded nested panels and the parent panel height is now different
+        this.expandedPanelHeight = this.$refs.panel.scrollHeight;
+        // need to wait for DOM update, after resetting panel maxHeight
+        setTimeout(() => {
```
ang-zeyu, line 117: same comment with `nextTick` here

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,85 @@ export default {
...
+          // transition to collapsed panel
+          this.$refs.panel.style.maxHeight = `${this.collapsedPanelHeight}px`;
+        }, 0); // nextTick doesn't work, not sure why :O
       } else {
-        this.localExpanded = !this.localExpanded;
+        // EXPAND
+        this.$refs.panel.style.maxHeight = `${this.expandedPanelHeight}px`;
```
ang-zeyu, line 123: you should be able to reference `this.$refs.panel.scrollHeight` here directly as well

ang-zeyu (commented):
Nice work with the rewrite! @wxwxwxwx9 

Left some observations that may require some reworking internally; also, `expanded` + `src` attribute currently leaves the panel 'stuck'

Untitled

Since there are quite a few combinations of options which affect the loading, you may want to try creating a whole bunch of `<panel>`s mixed with / without these options which may affect the functionality:
- vanilla
- `preload`
- `expanded`
- `src`
- `type=minimal`
- `minimized`

packages/vue-components/src/Retriever.vue
```diff
@@ -62,14 +62,12 @@ export default {
             template: `<div>\n${result}\n</div>`,
           });
           new TempComponent().$mount(this.$el);
-          this.$emit('src-loaded');
```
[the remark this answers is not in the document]
wxwxwxwx9: Have added it back into the code!

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,85 @@ export default {
...
-      el.style.maxHeight = `${el.scrollHeight}px`;
-    },
-    afterExpand(el) {
-      el.style.maxHeight = 'none';
-    },
-    beforeCollapse(el) {
-      el.style.maxHeight = `${el.scrollHeight}px`;
-    },
-    duringCollapse(el) {
-      if (this.$el.getBoundingClientRect().top < 0) {
```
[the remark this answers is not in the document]
wxwxwxwx9: Have adapted it into the code, thanks for the heads-up!

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,85 @@ export default {
...
-      this.$nextTick(() => {
-        this.localExpanded = true;
-      });
       this.localMinimized = false;
+      // need to wait for DOM update, after setting minimized to false
+      setTimeout(() => {
+        // transition to expanded
+        this.localExpanded = true;
+        this.$refs.panel.style.maxHeight = `${this.expandedPanelHeight}px`;
+      }, 0); // nextTick doesn't work, not sure why :O
```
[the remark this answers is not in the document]
wxwxwxwx9: Hmm, I don't think that's actually the issue :O I discovered some weird behaviours about it, not very sure what is the actual issue. But I have documented my observations in the new commit. Perhaps you would know something about it?

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Thanks for bringing this up! I have removed `expandedPanelHeight` from my commit, since it is redundant, as you have pointed out.

wxwxwxwx9:
Hi Zeyu, thank you so much for your review and suggestions. I think with this new commit, most (if not all) of the functionalities should be working properly. I have tested with all the options you listed. There doesn't seem to be any regressions (fingers crossed) based on the exploratory testing that I've done. 

There is, however, one "bug" in my code that I have resolved but don't quite seem to understand the root cause. I was thinking maybe you would know something about it. I have documented my observations about that particular section thoroughly. It's regarding the nextTick issue :-/

I have ensured that `preload` by using *v-if* for the card-body. I have also re-added the setMaxHeight for src/retriever, with some tweaks for the current usage, which I missed out in my last commit. 

Hopefully there won't be anymore major issues like the src/retriever issue you pointed out!

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
ang-zeyu: ahh ok, thanks for investigating!

this dosen't work because you are changing `this.$refs` directly (directly interfacing with the dom) instead of using vue's data interactivity (changing `localMinimized` in the previous comment / version).

> // some weird behaviours / observations here
>        // -- this.$nextTick would work if I either:
>      // 1. put a "console.log(this.$refs.panel.scrollHeight)" before it
>     // 2. OR place "if (this.$el.getBoundingClientRect().top < 0) { ... }" within the callback of nextTick
>   // (like how it is currently done below)
> // However, it would also work if i just use setTimeout instead of nextTick

Just a wild guess: these work due to referencing `scrollHeight` / `getBoundingClientRect` (e.g. `const x = this.$refs.panel.scrollHeight` would also work), which likely forces the browser to finish painting the previous `maxHeight` update first, before executing the next ``this.$refs.panel.style.maxHeight = `${this.collapsedPanelHeight}px`;``

In this case `setTimeout` is ok, though perhaps we should be using `requestAnimationFrame` instead ([3rd example here](https://css-tricks.com/using-css-transitions-auto-dimensions/)), not including the other uses of `nextTick` in (`if (!this.wasRetrieverLoaded) {` branch and in `open()`)

you may want to take a look at this comparison article between `setTimeout/Interval` and `requestAnimationFrame` as well: http://www.javascriptkit.com/javatutors/requestanimationframe.shtml

ang-zeyu (commented):
Nice work again.

One minor bug left, and couple of nits:

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -180,15 +223,26 @@ export default {
...
-    // Ensure this expr ordering is maintained
-    this.localExpanded = notExpandableNoExpand || this.expandedBool;
-    // If it is expanded, load the retriever immediately.
-    this.wasRetrieverLoaded = this.localExpanded;
-    this.localMinimized = this.minimizedBool;
+    // at the beginning
+    // expand everything to retrieve the appropriate heights
+    // ensure everything is rendered
+    // the variables will be reassigned to what the user indicated afterwards
+    this.localMinimized = false;
```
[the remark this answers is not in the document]
wxwxwxwx9: Ah, good point! Can I double check what are the things you think should be shifted back to created()? 

Other than `notExpandableNoExpand` and `localExpanded`, were you also thinking that the below should be shifted to created() as well? 
`     
 this.wasRetrieverLoaded = this.localExpanded;

      if (this.minimizedBool) {
        if (this.localExpanded) {
          // user indicated both minimized and expanded option
          // resolution: leave panel expanded and minimize it
          // so when user opens the minimized panel, it is automatically in expanded state (no transition)
          this.localMinimized = true;
          return;
        }
        // if panel is minimized, simply close the panel
        this.close();
      }
`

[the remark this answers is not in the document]
ang-zeyu: >Other than notExpandableNoExpand and localExpanded, were you also thinking that the below should be shifted to created() as well?

yup, save for `close()`

essentially anything that does not require the component to have mounted (anything not referencing `$refs` in this case ) and can be shifted correctly

packages/vue-components/src/panels/MinimalPanel.vue
```diff
@@ -110,15 +102,31 @@ export default {
...
+        // this is a collapse, set isHeaderAtBottom to true only at the end of transition
+        // to achieve the correct collapse transition effect of minimal panel
+        const onCollapseDone = () => {
+          this.isHeaderAtBottom = !this.isHeaderAtBottom;
+          this.$refs.panel.removeEventListener('transitionend', onCollapseDone);
+        };
+        this.$refs.panel.addEventListener('transitionend', onCollapseDone);
+      } else {
+        // this is an expansion, immediately set isHeaderAtBottom to true
+        this.isHeaderAtBottom = !this.isHeaderAtBottom;
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup agreed, that would be a better practice.

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,119 @@ export default {
...
-        as we need to know the scrollHeight of the content for the animation to work.
-        */
+      if (!this.wasRetrieverLoaded) {
         this.wasRetrieverLoaded = true;
         this.$nextTick(() => {
-          this.localExpanded = true;
+          this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
         });
-      } else {
         this.localExpanded = !this.localExpanded;
```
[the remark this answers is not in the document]
wxwxwxwx9: Ah, right, we should. The logic is almost the same. Thanks for spotting this!

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -99,76 +99,119 @@ export default {
...
-      set our own transition end handlers here for the initial loading of the content.
-      */
-      const onExpandDone = () => {
-        this.$refs.panel.style.maxHeight = 'none';
-        this.$refs.panel.removeEventListener('transitionend', onExpandDone);
-      };
-
-      this.$refs.panel.addEventListener('transitionend', onExpandDone);
+      // for expansion transition to 'continue' after src is loaded
       this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
```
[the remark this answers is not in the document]
wxwxwxwx9: Thanks for pointing this out! It's the first time I'm dealing with such performance issue haha. 

So what I did to workaround this is to introduce a new boolean flag `isRetrieverLoadDone` to help me determine whether I should set the maxHeight to 'none' for the first expansion transition. If src is not done loading yet, then we would not set it to 'none' and we would let Retriever's event emitter to help us invoke the setMaxHeight function to help us set the maxHeight to 'none' instead. 

I have also made some changes to the namings of functions for more clarity (but it's entirely based on my own opinion, so maybe you can see if the namings I came up with makes sense :O).

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
wxwxwxwx9: Alright! I have made the shiftings that I thought were appropriate :-)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
-          scrollTop: window.scrollY + this.$el.getBoundingClientRect().top - 3,
-        }, 500, 'swing');
+    initPanel() {
+      this.$refs.panel.addEventListener('transitionend', () => {
+        /*
+          If the panel's content grows, its height should accomodate it accordingly (e.g. nested panels).
+          We will always set the panel's maxHeight to 'none' after expansion.
+          This is to prevent maxHeight (which is used for transition) from restricting panel's growth.
+
+          At the end of first expansion transition, if src has not finished loading, then we should not
```
ang-zeyu, line 193: Let's explain the high level rationale instead, and leave the reader to figure out the details from the code easily :-)

```
However, if the panel has a 'src' attribute and it has not finished loading,
then this is delegated later to the retriever src-loaded event handler,
allowing a second maxHeight transition once it is loaded.
```

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
     },
-    duringCollapse(el) {
-      if (this.$el.getBoundingClientRect().top < 0) {
-        jQuery('html').animate({
-          scrollTop: window.scrollY + this.$el.getBoundingClientRect().top - 3,
-        }, 500, 'swing');
+    initPanel() {
+      this.$refs.panel.addEventListener('transitionend', () => {
+        /*
+          If the panel's content grows, its height should accomodate it accordingly (e.g. nested panels).
```
ang-zeyu, line 189:
```suggestion
          If the panel's content grows, its max-height should accomodate it accordingly (e.g. nested panels).
```

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
+      this.collapsedPanelHeight = 0;
     },
-    duringCollapse(el) {
-      if (this.$el.getBoundingClientRect().top < 0) {
-        jQuery('html').animate({
-          scrollTop: window.scrollY + this.$el.getBoundingClientRect().top - 3,
-        }, 500, 'swing');
+    initPanel() {
+      this.$refs.panel.addEventListener('transitionend', () => {
+        /*
```
ang-zeyu, line 188: some suggestions to shorten this comment:

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
+          If the panel's content grows, its height should accomodate it accordingly (e.g. nested panels).
+          We will always set the panel's maxHeight to 'none' after expansion.
```
ang-zeyu, line 190: let's remove this line and the one immediately below

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
+          Once that 'second' transition is finished, the flag isRetrieverLoadDone would have already
+          been set to true and the end of the 'second' transition will trigger this event listener
+          again to set maxHeight to none.
+        */
+        if (this.localExpanded && this.isRetrieverLoadDone) {
+          this.$refs.panel.style.maxHeight = 'none';
+        }
+      });
+
+      this.wasRetrieverLoaded = this.localExpanded;
```
ang-zeyu, line 211: could move this to `created()` too

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -180,15 +233,30 @@ export default {
...
     }
+
     // Edge case where user might want non-expandable card that isn't expanded by default
     const notExpandableNoExpand = !this.expandableBool && this.expanded !== 'false';
 
     // Set local data to computed prop value
 
     // Ensure this expr ordering is maintained
     this.localExpanded = notExpandableNoExpand || this.expandedBool;
-    // If it is expanded, load the retriever immediately.
```
ang-zeyu, line 190: could preserve these in favour of `if (this.minimizedBool && this.localExpanded) { ... }` and `this.wasRetrieverLoaded = this.localExpanded;` in `mounted()`

ang-zeyu (commented):
one more edge case and some nits:

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
+          again to set maxHeight to none.
+        */
+        if (this.localExpanded && this.isRetrieverLoadDone) {
+          this.$refs.panel.style.maxHeight = 'none';
+        }
+      });
+
+      this.wasRetrieverLoaded = this.localExpanded;
+
+      if (this.minimizedBool && !this.localExpanded) {
```
ang-zeyu, line 213: this branch may be redundant as well with the fix to https://github.com/MarkBind/markbind/pull/1430/files#r551016203

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
+          set the flag isRetrieverLoadDone to true and update the maxHeight again.
+
+          This will trigger a second expansion transition. And when this happens, we need the original
+          maxHeight (from first expansion transition) to ensure the transition can 'continue' smoothly.
+
+          Once that 'second' transition is finished, the flag isRetrieverLoadDone would have already
+          been set to true and the end of the 'second' transition will trigger this event listener
+          again to set maxHeight to none.
+        */
+        if (this.localExpanded && this.isRetrieverLoadDone) {
```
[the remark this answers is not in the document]
wxwxwxwx9: Thanks for spotting this! I should have set the maxHeight to none for any panels that were indicated to be expanded. 
However, I was thinking that this event bubbling for transitionend perhaps would be beneficial for our use case? :O 

Since it seems like for any inner nested panel that we expand, the outer panels' maxHeight should always be set to none regardless. It seems like this "feature" would provide sort of a safeguard. Although I agree that this might not be the "right" way to do things, since it may bring in other unexpected behaviours. 

Is this something that we can perhaps keep because it seems to be useful for our use case? Or is it a poor practice that might cause more problems down the road?

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Sorry, for the preserving, can I check which one are u referring to? :O

wxwxwxwx9: Can I check if you mean: 
1. remove `if (this.minimizedBool && this.localExpanded) { ... }` and `this.wasRetrieverLoaded = this.localExpanded;` from mounted()
2. keep `this.wasRetrieverLoaded` in created()

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
ang-zeyu: additional safeguards isn't necessarily bad; It would really depend on the use case / performance considerations / etc. Though over here my preference would be to make it work right from the start, since there's only a finite number of cases to verify

More importantly `transitionend` events are very often emitted. (note not only nested panels - e.g. any element / component that has a fade in transition on load could interfere with the transition)

packages/vue-components/src/panels/PanelBase.js
ang-zeyu: both, also we could keep the line below `this.wasRetrieverLoaded = this.localExpanded` which is:
`this.localMinimized = this.minimizedBool`

note you have
```
if (this.minimizedBool && this.localExpanded) {
  this.localMinimized = true
```
in `created()`

and then 

```
if (this.minimizedBool && !this.localExpanded) {
  // If panel is minimized, simply close the panel.
  this.close();
  return;
}
```
in `mounted()` (where `close()` also sets `localMinimized = true`)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -95,80 +95,133 @@ export default {
...
-    duringExpand(el) {
-      jQuery('html').stop();
-      el.style.maxHeight = `${el.scrollHeight}px`;
-    },
-    afterExpand(el) {
-      el.style.maxHeight = 'none';
-    },
-    beforeCollapse(el) {
-      el.style.maxHeight = `${el.scrollHeight}px`;
+    initCollapsedPanelHeight() {
```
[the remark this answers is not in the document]
wxwxwxwx9: Agreed! Will remove it.

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Thanks for the suggestion! I agree that it was rather lengthy :-)

packages/vue-components/src/panels/PanelBase.js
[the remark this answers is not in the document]
wxwxwxwx9: Understood! Thank you for providing your insights for the design decision!

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Hmm yeah you're right. I was still holding onto some old assumptions that this needs to be here. But with the new revisions according to your review, this can be shifted to created().

packages/vue-components/src/panels/PanelBase.js
wxwxwxwx9: Ah, I think I see what you mean! Those two branches are actually redundant. 

I will make the changes accordingly. Do let me know if I happen to misinterpret your explanation.

packages/vue-components/src/panels/MinimalPanel.vue
```diff
@@ -110,11 +102,30 @@ export default {
...
   methods: {
-    beforeExpandMinimal(el) {
+    minimalToggle() {
+      if (this.localExpanded) {
+        /*
+          This is a collapse.
+          Set isHeaderAtBottom to true only at the end of transition.
+          So that we can achieve the correct collapse transition effect of minimal panel.
+        */
+        const onCollapseDone = () => {
```
ang-zeyu, line 112: needs an `event.target` guard here too

wxwxwxwx9: Added! :-)

ang-zeyu (approved):
Lgtm 👍 

Let's `updatetest`

PR #1445, opened 2021-01-14
Make navigation menus accessible on smaller screens

### LINKED ISSUE #591: siteNav: Move the content to the navbar when the siteNav is hidden

1. Create a website with a navbar and a siteNav.
2. Serve the website on a very small screen (mobile screen size), so that the siteNav gets hidden as part of the responsive design.

The content of the siteNav should be relocated to the navbar, so that the reader can still access the content of the siteNav.

Readers of the website cannot use the siteNav at all.

This problem is inspired by https://github.com/MarkBind/markbind/pull/518#pullrequestreview-192981860.

### LINKED ISSUE #980: Redesign the navbar

Both nav directly accessible:

image

image

Both nav inaccessible:

image

image

image

View 2:

image

na

**Additional context**

na

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #980, fixes #591. 

Currently, siteNav and pageNav are hidden when screen size is small (< 993px). This PR adds an additional navigation bar below topNav to toggle the siteNav and pageNav menu when screen is small (similar to Docusaurus design). 

This is done by adding additional HTML code directly in `index.js` of `Page`, similar to how scroll to top button is implemented. Additional Javascript is also added to help with the toggling of the buttons. To check if a page contains siteNav or pageNav, a `<script>` tag is added to `siteNavProcessor.js` and `index.js` of `Page` to call the custom js function to let it know if they are present respectively on the page.

Most of the functions are binded to `window` and the function calls are done using `<script>`. Is this method of implementation recommended? Is there a better way to call these functions when building the siteNav or pageNav?

**Testing instructions:**
1. Create a site with siteNav and pageNav
2. Decrease screen size to less than 993px
3. Navigation menu should appear below topNav  

Implement mobile site and page navigation menus

---

### DIFF
--- a/docs/userGuide/syntax/navBars.mbdf
+++ b/docs/userGuide/syntax/navBars.mbdf
@@ -158,3 +158,26 @@ Name | Description
 </navbar>
 
 </span>
+
+****Page and site navigation menus****  
+Both [site navigation]({{ baseUrl }}/userGuide/usingComponents.html#site-navigation-menus) and [page navigation]({{ baseUrl }}/userGuide/usingComponents.html#page-navigation-menus) menus will be hidden on smaller screens.
+To make these accessible on smaller screens, you can use the `<site-nav-button />` and `<page-nav-button />` components in the `lower-navbar` slot. By default, if the `lower-navbar` slot is not specified, both the site and 
+page navigation buttons will automatically be added if they exist.
+
+```html{.no-line-numbers}
+<navbar>
+  <!-- Any normal navbar items -->
+  <a slot="brand" href="/" title="Home" class="navbar-brand">MarkBind</a>
+  <li><a href="/userGuide/usingComponents.html#navbars" class="nav-link">Highlighted Link</a></li>
+  <!-- Use slot to wrap the buttons in the lower navbar -->
+  <div slot="lower-navbar" class="nav-menu-container">
+    <site-nav-button />
+    <page-nav-button />
+  </div>
+</navbar>
+```
+
+Component | Description
+--- | ---
+`page-nav-button` | Pulls any element with an identifier, `id=page-nav` into the menu. If no such element exists, it pulls any [page navigation menu]({{ baseUrl }}/userGuide/usingComponents.html#page-navigation-menus) used in the layout.
+`site-nav-button` | Pulls any element with an identifier, `id=site-nav` into the menu. If no such element exists, it pulls the first [site navigation menu]({{ baseUrl }}/userGuide/usingComponents.html#site-navigation-menus) used in the layout.
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -50,6 +50,40 @@ function detectAndApplyFixedHeaderStyles() {
         margin-top: calc(-${headerHeight}px - ${bufferHeight}rem);
         height: calc(${headerHeight}px + ${bufferHeight}rem);
       }`);
+  insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
+
+  const addResizeHeaderListener = () => {
+    const resizeObserver = new ResizeObserver(() => {
+      const newHeaderHeight = headerSelector.height();
+      const sheets = document.styleSheets;
+      for (let i = 0; i < sheets.length; i += 1) {
+        const rules = sheets[i].cssRules;
+        // eslint-disable-next-line lodash/prefer-get
+        if (rules && rules[0] && rules[0].selectorText) {
+          switch (rules[0].selectorText) {
+          case '.fixed-header-padding':
+            sheets[i].deleteRule(0);
+            sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important }`);
+            break;
+          case 'span.anchor':
+            rules[0].style.top = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
+            break;
+          case 'span.card-container::before':
+            rules[0].style.marginTop = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
+            rules[0].style.height = `calc(${newHeaderHeight}px + ${bufferHeight}rem)`;
+            break;
+          case '.nav-menu-open':
+            rules[0].style.maxHeight = `calc(100% - ${newHeaderHeight}px + 50px)`;
+            break;
+          default:
+            break;
+          }
+        }
+      }
+    });
+    resizeObserver.observe(headerSelector[0]);
+  };
+  addResizeHeaderListener();
 }
 
 function updateSearchData(vm) {
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -120,7 +120,7 @@ code.hljs.inline {
 header[fixed] {
     position: fixed;
     width: 100%;
-    z-index: 1000;
+    z-index: 1001;
 }
 
 /* #app is treated as the main container */
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -1,31 +1,39 @@
 <template>
-  <nav ref="navbar" :class="['navbar', 'navbar-expand-md', themeOptions, addClass, fixedOptions]">
-    <div class="container-fluid">
-      <div class="navbar-brand">
-        <slot name="brand"></slot>
-      </div>
-      <button
-        v-if="!slots.collapse"
-        class="navbar-toggler"
-        type="button"
-        aria-expanded="false"
-        aria-label="Toggle navigation"
-        @click="toggleCollapse"
-      >
-        <span class="navbar-toggler-icon"></span>
-        <slot name="collapse"></slot>
-      </button>
+  <div>
+    <nav ref="navbar" :class="['navbar', 'navbar-expand-md', themeOptions, addClass, fixedOptions]">
+      <div class="container-fluid">
+        <div class="navbar-brand">
+          <slot name="brand"></slot>
+        </div>
+        <button
+          v-if="!slots.collapse"
+          class="navbar-toggler"
+          type="button"
+          aria-expanded="false"
+          aria-label="Toggle navigation"
+          @click="toggleCollapse"
+        >
+          <span class="navbar-toggler-icon"></span>
+          <slot name="collapse"></slot>
+        </button>
 
-      <div :class="['navbar-collapse',{collapse:collapsed}]">
-        <ul class="navbar-nav mr-auto mt-2 mt-lg-0">
-          <slot></slot>
-        </ul>
-        <ul v-if="slots.right" class="navbar-nav navbar-right">
-          <slot name="right"></slot>
-        </ul>
+        <div :class="['navbar-collapse',{collapse:collapsed}]">
+          <ul class="navbar-nav mr-auto mt-2 mt-lg-0">
+            <slot></slot>
+          </ul>
+          <ul v-if="slots.right" class="navbar-nav navbar-right">
+            <slot name="right"></slot>
+          </ul>
+        </div>
       </div>
+    </nav>
+    <div ref="lowerNavbar" v-show="isLowerNavbarShowing" class="lower-navbar-container">
+      <slot name="lower-navbar">
+          <site-nav-button />
+          <page-nav-button />
+      </slot>
     </div>
-  </nav>
+  </div>
 </template>
 
 <script>
@@ -52,11 +60,17 @@ export default {
       default: 'sibling-or-child',
     },
   },
+  provide() {
+    return {
+      toggleLowerNavbar: this.toggleLowerNavbar,
+    };
+  },
   data() {
     return {
       id: 'bs-example-navbar-collapse-1',
       collapsed: true,
       styles: {},
+      isLowerNavbarShowing: false,
     };
   },
   computed: {
@@ -198,6 +212,13 @@ export default {
         }
       }
     },
+    toggleLowerNavbar() {
+      if (this.$refs.lowerNavbar.childElementCount > 0) {
+        this.isLowerNavbarShowing = true;
+      } else {
+        this.isLowerNavbarShowing = false;
+      }
+    },
   },
   created() {
     this._navbar = true;
@@ -228,10 +249,14 @@ export default {
 
     // highlight current nav link
     this.highlightLink(window.location.href);
+
+    this.toggleLowerNavbar();
+    $(window).on('resize', this.toggleLowerNavbar);
   },
   beforeDestroy() {
     $('.dropdown', this.$el).off('click').offBlur();
     if (this.slots.collapse) $('[data-toggle="collapse"]', this.$el).off('click');
+    $(window).off('resize', this.toggleLowerNavbar);
   },
 };
 </script>
@@ -254,5 +279,13 @@ export default {
   >>> .dropdown-current {
     color: #fff !important;
     background: #007bff;
+  }
+
+  .lower-navbar-container {
+    background-color: #fff;
+    border-bottom: 1px solid #c1c1c1;
+    height: 50px;
+    width: 100%;
+    position: relative;
   }
 </style>
--- a/packages/vue-components/src/Overlay.vue
+++ b/packages/vue-components/src/Overlay.vue
@@ -1,0 +1,109 @@
+<template>
+  <div>
+    <span
+      :class="[{ 'nav-menu-close-icon': show }]"
+      @click="toggleNavMenu"
+    >
+      <slot name="navMenuIcon"></slot>
+    </span>
+    <div ref="navMenuContainer" :class="['nav-menu', { 'nav-menu-open': show }]">
+      <retriever
+        v-if="hasIdentifier"
+        :src="src"
+        :fragment="fragment"
+        @src-loaded="navMenuLoaded"
+      />
+      <div v-else ref="navigationMenu"></div>
+    </div>
+  </div>
+</template>
+
+<script>
+import retriever from './Retriever.vue';
+import $ from './utils/NodeList';
+import { publish, subscribe } from './utils/pubsub';
+
+export default {
+  components: {
+    retriever,
+  },
+  props: {
+    type: {
+      type: String,
+      default: null,
+    },
+    src: {
+      type: String,
+      default: null,
+    },
+    fragment: {
+      type: String,
+      default: null,
+    },
+    hasIdentifier: {
+      type: Boolean,
+      default: false,
+    },
+    getNavMenuContent: {
+      type: Function,
+      default: () => {},
+    },
+  },
+  inject: ['toggleLowerNavbar'],
+  data() {
+    return {
+      show: false,
+    };
+  },
+  methods: {
+    toggleNavMenu() {
+      if (!this.show) { publish('closeOverlay'); }
+      this.show = !this.show;
+    },
+    navMenuLoaded() {
+      this.toggleLowerNavbar();
+      $(this.$refs.navMenuContainer).find('a').on('click', () => {
+        this.toggleNavMenu();
+      });
+    },
+  },
+  mounted() {
+    const navMenu = this.$refs.navigationMenu;
+    const buildNav = (navMenuItems) => {
+      if (!navMenuItems) { return; }
+      for (let i = 0; i < navMenuItems.childNodes.length; i += 1) {
+        navMenu.appendChild(navMenuItems.childNodes[i].cloneNode(true));
+      }
+      this.navMenuLoaded();
+    };
+
+    if (!this.hasIdentifier) {
+      buildNav(this.getNavMenuContent());
+    }
+
+    subscribe('closeOverlay', () => { this.show = false; });
+  },
+};
+</script>
+
+<style scoped>
+    .nav-menu {
+        display: none;
+        padding: 50px 10px 20px;
+        position: fixed;
+        overflow-y: auto;
+        z-index: -1;
+    }
+
+    .nav-menu-close-icon > span::before {
+        content: "\e014" !important;
+    }
+
+    .nav-menu-open {
+        display: block !important;
+        width: 100% !important;
+        background: #fff;
+        height: 100%;
+        clear: both;
+    }
+</style>
--- a/packages/vue-components/src/PageNavButton.vue
+++ b/packages/vue-components/src/PageNavButton.vue
@@ -1,0 +1,77 @@
+<template>
+  <overlay
+    v-if="showPageNav"
+    type="pageNav"
+    fragment="page-nav"
+    :src="src"
+    :has-identifier="hasIdentifier"
+    :get-nav-menu-content="getPageNavContent"
+  >
+    <span slot="navMenuIcon" :class="['glyphicon', 'toggle-page-nav-button']"></span>
+  </overlay>
+</template>
+
+<script>
+import $ from './utils/NodeList';
+
+export default {
+  data() {
+    return {
+      hasPageNav: false,
+      hasIdentifier: false,
+      show: false,
+      src: '',
+    };
+  },
+  computed: {
+    showPageNav() {
+      return this.show && this.hasPageNav;
+    },
+  },
+  methods: {
+    getPageNavContent() {
+      const wrapper = document.createElement('div');
+      const pageNavTitle = document.getElementsByClassName('page-nav-title')[0];
+      const pageNavLinks = document.getElementById('mb-page-nav');
+      if (pageNavTitle) {
+        wrapper.appendChild(pageNavTitle.cloneNode(true));
+      }
+      if (pageNavLinks) {
+        wrapper.appendChild(pageNavLinks.cloneNode(true));
+      }
+      return wrapper;
+    },
+    togglePageNavButton() {
+      if (window.innerWidth < 1300) {
+        this.show = true;
+      } else {
+        this.show = false;
+      }
+    },
+  },
+  mounted() {
+    this.src = window.location.pathname;
+    this.hasIdentifier = document.getElementById('page-nav') !== null;
+    this.hasPageNav = document.getElementById('mb-page-nav') !== null;
+    this.togglePageNavButton();
+    $(window).on('resize', this.togglePageNavButton);
+  },
+  beforeDestroy() {
+    $(window).off('resize', this.togglePageNavButton);
+  },
+};
+</script>
+
+<style scoped>
+    .toggle-page-nav-button {
+        cursor: pointer;
+        right: 0;
+        padding: 15px;
+        position: absolute;
+    }
+
+    .toggle-page-nav-button::before {
+        content: "\e235";
+        font-size: 20px;
+    }
+</style>
--- a/packages/vue-components/src/SiteNavButton.vue
+++ b/packages/vue-components/src/SiteNavButton.vue
@@ -1,0 +1,68 @@
+<template>
+  <overlay
+    v-if="showSiteNav"
+    type="siteNav"
+    fragment="site-nav"
+    :src="src"
+    :has-identifier="hasIdentifier"
+    :get-nav-menu-content="getSiteNavContent"
+  >
+    <span slot="navMenuIcon" :class="['glyphicon', 'toggle-site-nav-button']"></span>
+  </overlay>
+</template>
+
+<script>
+import $ from './utils/NodeList';
+
+export default {
+  data() {
+    return {
+      hasIdentifier: false,
+      hasSiteNav: false,
+      show: false,
+      src: '',
+    };
+  },
+  computed: {
+    showSiteNav() {
+      return this.show && this.hasSiteNav;
+    },
+  },
+  methods: {
+    getSiteNavContent() {
+      return document.getElementsByClassName('site-nav-root')[0];
+    },
+    toggleSiteNavButton() {
+      if (window.innerWidth < 992) {
+        this.show = true;
+      } else {
+        this.show = false;
+      }
+    },
+  },
+  mounted() {
+    this.src = window.location.pathname;
+    this.hasIdentifier = document.getElementById('site-nav') !== null;
+    this.hasSiteNav = document.getElementsByClassName('site-nav-root').length !== 0;
+    this.toggleSiteNavButton();
+    $(window).on('resize', this.toggleSiteNavButton);
+  },
+  beforeDestroy() {
+    $(window).off('resize', this.toggleSiteNavButton);
+  },
+};
+</script>
+
+<style scoped>
+    .toggle-site-nav-button {
+        cursor: pointer;
+        left: 0;
+        padding: 15px;
+        position: absolute;
+    }
+
+    .toggle-site-nav-button::before {
+        content: "\e236";
+        font-size: 20px;
+    }
+</style>
--- a/docs/userGuide/syntax/pageNavigationMenus.mbdf
+++ b/docs/userGuide/syntax/pageNavigationMenus.mbdf
@@ -17,6 +17,8 @@
 
 3. **Position the page navigation menu** within your layout using the `{% raw %}{{ pageNav }}{% endraw %}` variable.
 
+4. **(Optional) To make pageNav accessible on smaller screens, you can use the `<page-nav-button />` component in the [navbar]({{baseUrl}}/userGuide/usingComponents.html#navbars).**
+
 <div id="short" class="indented">
 
 {{ icon_example }}
--- a/docs/userGuide/syntax/siteNavigationMenus.mbdf
+++ b/docs/userGuide/syntax/siteNavigationMenus.mbdf
@@ -7,6 +7,7 @@
 Steps to add a siteNav:
 1. Format your siteNav as an unordered Markdown list
 2. Include it under a `<site-nav>` element.
+3. (Optional) To make siteNav accessible on smaller screens, you can use the `<site-nav-button />` component in the [navbar]({{baseUrl}}/userGuide/usingComponents.html#navbars).
 
 <div id="short">
 
--- a/packages/vue-components/src/__tests__/Navbar.spec.js
+++ b/packages/vue-components/src/__tests__/Navbar.spec.js
@@ -1,0 +1,87 @@
+import { mount } from '@vue/test-utils';
+import Navbar from '../Navbar.vue';
+import PageNavButton from '../PageNavButton.vue';
+import SiteNavButton from '../SiteNavButton.vue';
+
+const DEFAULT_STUBS = {
+  'page-nav-button': PageNavButton,
+  'site-nav-button': SiteNavButton,
+};
+
+const NAVBAR_CONTENT = `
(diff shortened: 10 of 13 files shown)


packages/core/src/Page/index.js
```diff
@@ -347,7 +347,7 @@ class Page {
     const headingStack = [];
     Object.keys(this.navigableHeadings).forEach((key) => {
       const currentHeadingLevel = this.navigableHeadings[key].level;
-      const currentHeadingHTML = `<a class="nav-link py-1" href="#${key}">`
+      const currentHeadingHTML = `<a onclick="closePageNav();" class="nav-link py-1" href="#${key}">`
```
ang-zeyu, line 350: we may catch the bubbled event instead (also checking `e.target` and other parameters to make it as restrictive as possible) to hide the overlay. This way it stays decoupled

ang-zeyu (commented):
Thanks! @jonahtanjz 

Regarding the usage (from author pov), I think we could try to encapsulate the implementation in the already existing `Navbar.vue`, and create some reusable components. Something along the lines of this that dosen't make this a fixed, mandatory placement:

```
// author writing a navbar
<navbar>
  ...
  <div slot="lower-navbar"> // a new slot for the "secondary navbar" you have here
    <page-nav-button> // new component the user may use to link this up
    <site-nav-button> // similarly
  </div>
</navbar>
```

that way, we don't restrict the author to a single layout (secondary navbar), they may even omit the secondary navbar entirely, or place the buttons in the "main (top most) navbar" if they wish.

we could also provide customisation options for `page-nav-button` / etc. in the future.

---

With the comments below:

> Most of the functions are binded to window and the function calls are done using <script>. Is this method of implementation recommended? Is there a better way to call these functions when building the siteNav or pageNav?

So I guess, the answer to this lies in the (author) usage design. Changing it should result in a more sealed off implementation as well.


There are probably better ways of doing this besides the two suggestions here, but this is what I can propose right now 🤔🤔🤔

Would love to hear your suggestions @jonahtanjz @damithc @acjh

jonahtanjz:
Thanks @ang-zeyu for the suggestions and review.

Yup I agree that making the `site-nav` and `page-nav` buttons into reusable components would allow the implementation to be more customizable and cleaner. As mentioned, doing this will require users to add in additional code to show these buttons on mobile but this should be fine if the user intent to customize it. Is it also possible if we default it to the secondary bar if it is not specified? This way, current Markbind sites will not require additional code to make these buttons appear on mobile. Not sure which is better 🤔.  

I can probably work on making the `<site-nav-button>` and `<page-nav-button>` components first if it fine to go in this direction.

ang-zeyu:
Any thoughts? @damithc 

Option 1:

\+ prettier (overlay would extend from navbar downward only)
\- restrictive, overlay may only contain a pagenav (no titles, etc.), or one (the first) `<site-nav>`

```
// author writing a navbar
<navbar>
  ...
  <div slot="lower-navbar"> // a new slot for the "secondary navbar" you have here
    <page-nav-button> // new component the user may use to link this up
    <site-nav-button> // similarly
  </div>
</navbar>
```

Option 2:

\+ generic, easy to implement
\+ overlay may contain different content / designs
\- more verbose for the author,
\- less pretty (overlay would cover the entire page)

```
// author writing a navbar
<navbar>
  ...
  <div slot="lower-navbar"> // a new slot for the "secondary navbar" you have here
    <overlay icon="...">
       <site-nav>
          ...
       </site-nav>
    </overlay>
    <overlay icon="...">
        {{ pageNav }} 
    </overlay>
  </div>
</navbar>
```

---

@jonahtanjz In either case, feel free to make the [`{{ pageNav }}`](https://markbind-master.netlify.app/userguide/tweakingthepagestructure#constructing-a-page-navigation-menu) variable an element instead as well without deprecating anything, as its unreleased.

damithc:
> Any thoughts? @damithc

I need some info to understand this one.
Is this going to affect how authors create navbars? Because the feature being added doesn't sound like something the author should specifically know about i.e., it should work behind the scene.
I guess it's fine to tweak the syntax but the new syntax should not reveal anything about smaller screens?

ang-zeyu:
> Is it also possible if we default it to the secondary bar if it is not specified?

Sorry missed this out earlier. Does this mean `page / site-nav-button` would not need to be placed under a slot?

> I need some info to understand this one.
> Is this going to affect how authors create navbars? Because the feature being added doesn't sound like something the author should specifically know about i.e., it should work behind the scene.
> I guess it's fine to tweak the syntax but the new syntax should not reveal anything about smaller screens?

yup. 

the main issue here is that shifting to expressive layouts means there's no longer a mandatory "site-nav section". (can't know which section of the layout file that is supposed to be) (versus before, all files in `_markbind/navigation` could be identified)

We certainly could identify `<site-nav>`, or `{{ pageNav }}`, but then the overlay would and can only contain these. (e.g. no "User Guide" title, images, etc. ). There's also the issue of being able to use multiple `<site-nav>`s, and how that would work with the mobile overlay. (which / all of the `<site-nav>`s should be pulled into the overlay automatically)

My opinion here is that the placement, styling of the button, etc. should not be fixed either (e.g. the author should be able to omit it entirely, or place it somewhere else, change the icon, ...). Suggestions hence:
- option 1 - dosen't really deal with the main issue, but provides basic customisability (`<site/page-nav-button options... />`)
- option 2 - deals with the main issue, but is rather verbose (`<overlay options...><include src="#site-nav" /></overlay>`)

We could also stick the current PR, which means nothing to do for the author at all - but it would still be a little restrictive (e.g. no "User Guide" title, images, etc. ).

(Option 3) Or minimally, we can require the author to specify some sort of identifier in any element in the layout file https://markbind-master.netlify.app/userguide/tweakingthepagestructure#layouts (e.g. `id="mobile-site-nav"`) that tells our code what to pull into the overlay. (but then no customisability)

Not sure if there are better compromises

damithc:
Related question: does this require the author to specify two site-navs, one for normal use and one for mobile? If yes, that feels like extra work an author shouldn't have to do.

jonahtanjz:
> Sorry missed this out earlier. Does this mean page / site-nav-button would not need to be placed under a slot?
Was thinking more along the lines of having a slot with fallback content in the `navbar` or something similar so that current Markbind sites will not need to do anything to be able to have the buttons on mobile.

ang-zeyu:
> Was thinking more along the lines of having a slot with fallback content in the `navbar` or something similar so that current Markbind sites will not need to do anything to be able to have the buttons on mobile.

Ahh, that sounds nice 👍 

> Related question: does this require the author to specify two site-navs, one for normal use and one for mobile? If yes, that feels like extra work an author shouldn't have to do.

Would require some extra things inside the existing `<navbar>` (not a separate one)

With @jonahtanjz 's proposal above, it is also optional (can be omitted, but also customised if needed)

damithc:
> Would require some extra things inside the existing `<navbar>` (not a separate one)
> 
> With @jonahtanjz 's proposal above, it is also optional (can be omitted, but also customised if needed)

Sounds good. 👍

ang-zeyu:
The main issue still exists though (identifying what to pull into the overlay).

Suggestions:

1. Assume the user is only using <= 1 sitenav and <= 1 pagenav per page, then pull these in automatically if they exist.
To facilitate more expressive layouts, we can fallback to @jonahtanjz's proposal and something like option 2 (`<overlay>...any MarkBind content... can be included as well...</overlay>`), or something similar

2. Or, require the author to pull specify some identifier in the layout file of what to pull in (e.g. `id="mobile-site-nav"`), also combining with @jonahtanjz 's proposal but something like option 1 for customisability (`<site/page-nav-button />`)

3. ❔

jonahtanjz:
> 1. Assume the user is only using <= 1 sitenav and <= 1 pagenav per page, then pull these in automatically if they exist.
>   To facilitate more expressive layouts, we can fallback to @jonahtanjz's proposal and something like option 2 (`<overlay>...any MarkBind content... can be included as well...</overlay>`), or something similar
> 1. Or, require the author to pull specify some identifier in the layout file of what to pull in (e.g. `id="mobile-site-nav"`), also combining with @jonahtanjz 's proposal but something like option 1 for customisability (`<site/page-nav-button />`)

@ang-zeyu I would say option 2 is better as it is more clear and straightforward. This will allow users to specify which `site-nav` to use to increase flexibility but this will also mean current markbind sites will need to include this new identifier for it to work. We can also pull the default identifier, `id=site-nav` and `id=page-nav`, as a workaround for users that did not change them? Not sure how many current markbind sites have changed them. That's my thought on this.

ang-zeyu:
> @ang-zeyu I would say option 2 is better as it is more clear and straightforward. This will allow users to specify which `site-nav` to use to increase flexibility but this will also mean current markbind sites will need to include this new identifier for it to work.

👍 let's go with that then

Could also still add in the default behaviour of assuming <= 1 page/sitenav (then pulling those in at a lower priority), so the author dosen't have to do anything. (if the page dosen't contain `id=site-nav` nor `id=page-nav`)

> We can also pull the default identifier, `id=site-nav` and `id=page-nav`, as a workaround for users that did not change them? Not sure how many current markbind sites have changed them. That's my thought on this.

Since its a new feature, don't worry about backward compatibility. Our template sites (`markbind init`) have these identifiers set up already, which should make things slightly easier as well.

So, in this order:
- pull in `id=site-nav` / `id=page-nav` if they exist in the page (we can add on in another PR to change this identifier even if need be)
- pull in the first `<site-nav>` / the only `{{ pageNav }}` if it exists
 
Sounds right? @damithc @jonahtanjz

damithc:
> So, in this order:
> 
> * pull in `id=site-nav` / `id=page-nav` if they exist in the page (we can add on in another PR to change this identifier even if need be)
> * pull in the first `<site-nav>` / the only `{{ pageNav }}` if it exists
> 
> Sounds right? @damithc @jonahtanjz

Yes, no objections from me. 👍

jonahtanjz:
@ang-zeyu The PR is ready for review. 

I've created separate vue components for `<site-nav-button />` and `<page-nav-button />`. For now, they can be used under a slot in the `navbar` so that when it is opened, the menus extend nicely below the navbar. Both components will pull in identifier first `id=site/page-nav` then fallback to the first or only site/page navigation respectively. Likewise, I've also included fallback content in the `lower-navbar` slot so that if the `<site/page-nav-button />` components are not specified, they will still show up by default. Currently, both components are `navbar` specific as some of the logics are handled in the `navbar` (hiding and showing buttons on different screen sizes, as well as adjust the fixed header padding when the secondary navbar appears/disappears). Perhaps in the future if we want to increase customizability, we can extract it out of the `navbar` and have a custom wrapper component just for the buttons 🤔?

Also added tests and documentations :)

packages/vue-components/src/utils/utils.js
```diff
@@ -159,3 +161,40 @@ export function VueFixer (vue) {
   vue.mixins.unshift(mixin)
   return vue
 }
+
+export function resizeHeader(isNavMenuShowing) {
```
ang-zeyu, line 165: hmm not sure if we should be putting this within `navbar.vue`. (note a header isn't limited to just a navbar)

how about adding the window listener in `index.js` where the fixed header padding is originally setup?
It should keep these concerns closer.

Also, could override the selectors' properties (by cascading) instead querying every element

packages/vue-components/src/PageNavButton.vue
```diff
@@ -0,0 +1,98 @@
...
+    };
+  },
+  inject: ['initPageNav'],
+  methods: {
+    togglePageNav() {
+      this.show = !this.show;
+      this.toggleSiteNavButton();
+    },
+    toggleSiteNavButton() {
+      const siteNavButton = document.getElementById('toggle-site-nav-button');
```
ang-zeyu, line 29: would `:class` work? Similar to what you did for `nav-menu-open` above. (also for the page nav variant)

packages/vue-components/src/Navbar.vue
```diff
@@ -198,6 +217,33 @@ export default {
...
+      this.toggleNavMenu();
+    },
+    showNavMenu() {
+      this.isNavMenuShowing = true;
+    },
+    hideNavMenu() {
+      this.isNavMenuShowing = false;
+    },
+    toggleNavMenu() {
+      if ((this.hasPageNav && window.innerWidth < 1300)
```
ang-zeyu, line 235: how about (to avoid coupling this slot showing with the buttons):
- let the button components define their own `window.innerWidth` limits, and destroy themselves upon exceeding this limit  (or show themselves)
- let `navbar` do a generic check for whether there is any html within the div you added above (then this slot can also be used for non-button content)

packages/vue-components/src/Navbar.vue
```diff
@@ -1,36 +1,46 @@
...
-        class="navbar-toggler"
-        type="button"
-        aria-expanded="false"
-        aria-label="Toggle navigation"
-        @click="toggleCollapse"
-      >
-        <span class="navbar-toggler-icon"></span>
-        <slot name="collapse"></slot>
-      </button>
+  <div ref="navbarContainer">
```
ang-zeyu, line 2: what's this `ref` for?

packages/vue-components/src/PageNavButton.vue
```diff
@@ -0,0 +1,98 @@
...
+      this.initPageNav();
+      this.hasPageNav = true;
+      const { pageNavMenu } = this.$refs;
+      for (let i = 0; i < pageNavItems.childNodes.length; i += 1) {
+        pageNavMenu.appendChild(pageNavItems.childNodes[i].cloneNode(true));
+      }
+    };
+
+    const $el = $(this.$refs.pageNavMenu);
+    const pageNav = document.getElementById('page-nav');
```
ang-zeyu, line 50: cloning the html directly would disable reactivity / other things if the `site-nav` / `page-nav` has reactive elements that have already been processed in some way (e.g. vue components)

you can refer to `retriever.vue` (used in `<panel src="...">`) for mounting a clone of the original html source (as received by the browser).
However, it makes a webrequest (although it would likely hit the cache), so if there's a 'native' method to retrieve the original page source, all the more better. If not, directly reusing `<retriever>` should be fine too.

packages/core-web/src/styles/nav-menu-bar.css
```diff
@@ -0,0 +1,39 @@
+.hide-nav-button {
```
ang-zeyu, line 1: how about: create a _internal_ base `<overlay>` component, with these styles

use it in `site/page-nav-button` so we can scope these styles still, also to avoid repeating similar logic / template in both

ang-zeyu (commented):
Thanks for the rework! : )

> @ang-zeyu The PR is ready for review.
> 
> I've created separate vue components for `<site-nav-button />` and `<page-nav-button />`. For now, they can be used under a slot in the `navbar` so that when it is opened, the menus extend nicely below the navbar. Both components will pull in identifier first `id=site/page-nav` then fallback to the first or only site/page navigation respectively. Likewise, I've also included fallback content in the `lower-navbar` slot so that if the `<site/page-nav-button />` components are not specified, they will still show up by default. 

Couple of user issues:
- not scrollable (tried on mobile device + chrome resized)
- the page nav button shows up when there's no page nav

> Currently, both components are `navbar` specific as some of the logics are handled in the `navbar` (hiding and showing buttons on different screen sizes, as well as adjust the fixed header padding when the secondary navbar appears/disappears). Perhaps in the future if we want to increase customizability, we can extract it out of the `navbar` and have a custom wrapper component just for the buttons 🤔?
> 
> Also added tests and documentations :)

This sounds good, also not just for customisability - but separation of concerns

Left a few suggestions below on how to tackle the separation here (customisability can come later)

packages/vue-components/src/utils/utils.js
jonahtanjz: > Also, could override the selectors' properties (by cascading) instead querying every element

Sorry I don't really get what you meant by this 😅. Do you mean inserting new CSS rules to target these classes?

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Hmm I'm not sure if `:class` would work for this because this is querying `page-nav-button` from `site-nav` when it is opened. They should be considered sibling elements so not sure how to pass the data to each other unless we pass it to the parent and then back to the child. Will just using `getElementById` be better because the buttons `id` are fixed and will not be changed by the user 🤔?

packages/vue-components/src/utils/utils.js
ang-zeyu: yup

jonahtanjz: Thanks for the clarification :)

> hmm not sure if we should be putting this within navbar.vue. (note a header isn't limited to just a navbar)
>
> how about adding the window listener in index.js where the fixed header padding is originally setup?
> It should keep these concerns closer.

 One more thing, the reason why I placed this in `navbar` is because a header padding resize is only necessary if the `lower-navbar` slot is present. In this case, is it ok to let the listener stay in `navbar`?

packages/vue-components/src/PageNavButton.vue
ang-zeyu: ahh, my bad. I didn't notice that this was for the other component at all. 🙈 


> They should be considered sibling elements so not sure how to pass the data to each other unless we pass it to the parent and then back to the child.

I agree this is a little messy (and again reintroduces coupling with the navbar).

my concern here is with trying to limit dom effects that can be limited to vue to be limited to vue, to minimise side effects outside of vue. (as opposed to interfacing with `<ul>` / `<li>` in dropdowns - no choice there)

but introducing global state management for this seems overkill as well.. hmm.......

packages/vue-components/src/utils/utils.js
ang-zeyu: > One more thing, the reason why I placed this in navbar is because a header padding resize is only necessary if the lower-navbar slot is present. In this case, is it ok to let the listener stay in navbar?

not necessarily, e.g.

`<header><box dismissible>...</box></header>` -> close box -> `fixed-header-padding` is not changed (everything appears pushed downward)

It's part of a larger generic issue (not listed anywhere) that can be dealt with separately, or in this PR since its a prerequisite (preferably a separate commit)

packages/vue-components/src/PageNavButton.vue
ang-zeyu: how about just a event subscribe / consume implementation?

```
// pagenavbutton / sitenavbutton
subscribe('closeoverlays', () => this.show = false)
publish('closeoverlays')

// simple event implementation
window.subscribers = [ ... ]
......... or just this https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events
```

this way the dom manipulation stays within the subscriber component (page / site button); won't need to tie the implementation to the classes and ids as well

packages/vue-components/src/utils/utils.js
jonahtanjz: Ahh thanks, wasn't aware of this issue. Added a listener in `index.js` to resize the header if it changes. Should be fixed now :)

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Thanks for the suggestion 😄

packages/vue-components/src/Navbar.vue
jonahtanjz: Shifted this to their individual component. `navbar` should be free of `page/site-nav-button` components other than the fallback content in the `slot`.

packages/vue-components/src/Navbar.vue
jonahtanjz: Oops it's not in used anymore 😬. Has been removed.

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Alright I've decided to use `<retriever>` to get the contents of the identifier `id=page/site-nav`. However, if these identifiers are not present, will have to look for their respective classes still so it will fallback to using `appendChild`. Not sure if that's fine?

packages/core-web/src/styles/nav-menu-bar.css
jonahtanjz: Created an `overlay` component for both `<page/site-nav-button>`. Abstracted all the similar template and logic into this component.

jonahtanjz:
@ang-zeyu Updated the PR accordingly. Would appreciate if you can look through it :)

packages/vue-components/src/Overlay.vue
```diff
@@ -0,0 +1,185 @@
...
+      break;
+    default:
+      break;
+    }
+  },
+};
+</script>
+
+<style scoped>
+    #toggle-site-nav-button {
```
ang-zeyu, line 138: let's rename these into classes since the ids are no longer necessary

packages/vue-components/src/PageNavButton.vue
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      hasPageNav: false,
+      hasIdentifier: false,
+      show: false,
+      src: '',
+    };
+  },
+  computed: {
+    showPageNav() {
+      return this.show && this.hasPageNav;
```
ang-zeyu, line 25: should it have `|| this.hasIdentifier` as well?

packages/vue-components/src/SiteNavButton.vue
```diff
@@ -0,0 +1,45 @@
...
+      if (window.innerWidth < 992) {
+        this.show = true;
+      } else {
+        this.show = false;
+      }
+    },
+  },
+  computed: {
+    showSiteNav() {
+      return this.show && this.hasSiteNav;
```
ang-zeyu, line 34: same here

packages/vue-components/src/PageNavButton.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <overlay
+    v-if="showPageNav"
+    :type="'pageNav'"
```
ang-zeyu, line 4: could do
```suggestion
    type="pageNav"
```

also for site nav, and `:fragment`

packages/core-web/src/index.js
```diff
@@ -50,6 +50,41 @@ function detectAndApplyFixedHeaderStyles() {
         margin-top: calc(-${headerHeight}px - ${bufferHeight}rem);
         height: calc(${headerHeight}px + ${bufferHeight}rem);
       }`);
+  insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
+}
+
+function addResizeHeaderListener() {
```
ang-zeyu, line 56: shouhld we put this under `detectAndApplyFixedHeaderStyles` instead? (it has a guard for no fixed header as well)

packages/core-web/src/index.js
```diff
@@ -50,6 +50,41 @@ function detectAndApplyFixedHeaderStyles() {
...
+        case '.nav-menu-open':
+          rules[0].style.maxHeight = `calc(100% - ${headerHeight}px + 50px)`;
+          break;
+        default:
+          break;
+        }
+      }
+    }
+  });
+  resizeObserver.observe(document.querySelector('header'));
```
ang-zeyu, line 87: would `headerSelector[0]` work?

packages/core-web/src/index.js
```diff
@@ -50,6 +50,41 @@ function detectAndApplyFixedHeaderStyles() {
...
+
+function addResizeHeaderListener() {
+  const headerSelector = jQuery('header[fixed]');
+  const bufferHeight = 1;
+  const resizeObserver = new ResizeObserver(() => {
+    const headerHeight = headerSelector.height();
+    const sheets = document.styleSheets;
+    for (let i = 0; i < sheets.length; i += 1) {
+      const rules = sheets[i].cssRules;
+      // eslint-disable-next-line
```
ang-zeyu, line 64:
```suggestion
      // eslint-disable-next-line lodash/prefer-get
```

a little more specific

packages/vue-components/src/Overlay.vue
```diff
@@ -0,0 +1,185 @@
...
+    };
+    const buildSiteNav = () => {
+      const siteNav = document.getElementsByClassName('site-nav-root')[0];
+      buildNav(siteNav);
+    };
+
+    switch (this.type) {
+    case 'siteNav':
+      if (!this.hasIdentifier) { buildSiteNav(); }
+      subscribe('closeSiteNav', () => { this.show = false; });
```
ang-zeyu, line 124: would a generic event work? (e.g. `closeOtherOverlays`)

so no need to tie `overlay` to site / page nav specifically

packages/vue-components/src/Overlay.vue
```diff
@@ -0,0 +1,185 @@
...
+  },
+  mounted() {
+    const navMenu = this.$refs.navigationMenu;
+    const buildNav = (navMenuItems) => {
+      for (let i = 0; i < navMenuItems.childNodes.length; i += 1) {
+        navMenu.appendChild(navMenuItems.childNodes[i].cloneNode(true));
+      }
+      this.navMenuLoaded();
+    };
+    const buildPageNav = () => {
```
ang-zeyu, line 104: > Alright I've decided to use <retriever> to get the contents of the identifier id=page/site-nav. However, if these identifiers are not present, will have to look for their respective classes still so it will fallback to using appendChild. Not sure if that's fine?

sounds good, these don't have anything reactive

should we shift these upward though? (into site / page-nav-button) (as a callback accepting `navMenu` maybe)
so the base component only handles generic logic, the specific stuff stays up there

packages/vue-components/src/Overlay.vue
```diff
@@ -0,0 +1,185 @@
+<template>
+  <div>
+    <span
```
ang-zeyu, line 3: suggestion: pull the icon upward using scoped slots, so the component specific styles can be moved up there

not sure if it would end up more verbose though 😐

packages/vue-components/src/Overlay.vue
```diff
@@ -0,0 +1,185 @@
...
+    .nav-menu-open {
+        display: block !important;
+        width: 100% !important;
+        background: #fff;
+        height: 100%;
+        clear: both;
+    }
+
+    .site-nav-menu,
+    .page-nav-menu {
```
ang-zeyu, line 161: could we combine these?

docs/userGuide/syntax/navBars.mbdf
```diff
@@ -158,3 +158,26 @@ Name | Description
 </navbar>
 
 </span>
+
+****Page and site navigation menus****  
+Both [site navigation]({{ baseUrl }}/userGuide/usingComponents.html#site-navigation-menus) and [page navigation]({{ baseUrl }}/userGuide/usingComponents.html#page-navigation-menus) menus will be hidden on smaller screens.
+To make these accessible on smaller screens, you can use the `<site-nav-button />` and `<page-nav-button />` components in the `lower-navbar` slot. By default, if the `lower-navbar` slot is not specified, both the site and 
+page navigation buttons will automatically be added if they exist.
+
+```html
```
ang-zeyu, line 167: let's disable the line numbers here

docs/userGuide/syntax/navBars.mbdf
```diff
@@ -158,3 +158,26 @@ Name | Description
...
+  <div slot="lower-navbar" class="nav-menu-container">
+    <site-nav-button />
+    <page-nav-button />
+  </div>
+</navbar>
+```
+
+Component | Description
+--- | ---
+`page-nav-button` | Will pull everything wrapped with an identifier, `id=page-nav`, in the layout file into the menu. Alternatively, if the identifier is not present, it will pull the page navigation menu in automatically if it exist.
```
ang-zeyu, line 182: let's shorten this 

```suggestion
`page-nav-button` | Pulls any element with an identifier, `id=page-nav` into the menu. If no such element exists, it pulls any [page navigation menu](insert link here) used in the layout.
```

or anything else :-)

packages/core-web/src/styles/markbind.css
```diff
@@ -120,7 +120,7 @@ code.hljs.inline {
 header[fixed] {
     position: fixed;
     width: 100%;
-    z-index: 1000;
```
ang-zeyu, line 123: One more minor styling issue: (navbar example in ug)

Untitled

ang-zeyu (commented):
Thanks, just a couple more nits:

packages/vue-components/src/Overlay.vue
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Thanks! Updated :)

packages/core-web/src/index.js
jonahtanjz: Thanks! Shifted this into `detectAndApplyFixedHeaderStyles`.

packages/core-web/src/index.js
jonahtanjz: Thanks! Updated :)

packages/core-web/src/index.js
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/Overlay.vue
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/Overlay.vue
jonahtanjz: Yup sounds good. Shifted this up to `site / page-nav-button`

packages/vue-components/src/Overlay.vue
jonahtanjz: Shifted this up. Doesn't look too verbose :)

packages/vue-components/src/Overlay.vue
jonahtanjz: Thanks! Updated :)

docs/userGuide/syntax/navBars.mbdf
jonahtanjz: Alright updated :)

docs/userGuide/syntax/navBars.mbdf
jonahtanjz: Thanks! Updated :)

packages/core-web/src/styles/markbind.css
jonahtanjz: Thanks! Fixed :)

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Hmm actually this was implemented before but this will cause the page/site nav button to show up even if they are not present on the page. This happens when there is an identifier, for example, `id=page-nav` but the page does not have any headers, which will display an empty page nav menu when opened.

jonahtanjz:
@ang-zeyu Thanks for the review! The PR is updated accordingly except for the part regarding the `|| this.hasIdentifier` in `page/site-nav-button` component :)

packages/vue-components/src/Navbar.vue
```diff
@@ -228,10 +249,14 @@ export default {
...
     // highlight current nav link
     this.highlightLink(window.location.href);
+
+    this.toggleLowerNavbar();
+    $(window).on('resize', this.toggleLowerNavbar);
   },
   beforeDestroy() {
     $('.dropdown', this.$el).off('click').offBlur();
     if (this.slots.collapse) $('[data-toggle="collapse"]', this.$el).off('click');
+    $(window).off();
```
ang-zeyu, line 259: let's make it specific, so other listeners don't get removed

packages/vue-components/src/PageNavButton.vue
ang-zeyu: ahh I see. Thanks for clarifying!

packages/vue-components/src/Navbar.vue
```diff
@@ -52,11 +60,17 @@ export default {
       default: 'sibling-or-child',
     },
   },
+  provide() {
+    return {
+      toggleLowerNavbar: this.toggleLowerNavbar,
```
ang-zeyu, line 65: how about just a prop? since they're in the same compilation scope

jonahtanjz: Hmm just to clarify, if this is passed to `<page/site-nav-button>` component through a prop, since `lower-navbar` is a slot, if a user defines new content in the layout file for this slot without specifying the prop, for example,
```
<div slot="lower-navbar">
  <site-nav-button />
</div>
``` 
then it won't be passed down to the component in this case?

packages/vue-components/src/PageNavButton.vue
```diff
@@ -0,0 +1,78 @@
...
+    return {
+      hasPageNav: false,
+      hasIdentifier: false,
+      show: false,
+      src: '',
+    };
+  },
+  provide() {
+    return {
+      getNavMenuContent: this.getPageNavContent,
```
ang-zeyu, line 27: sorry, commented on the wrong line 🙊

packages/vue-components/src/Navbar.vue
ang-zeyu: https://github.com/MarkBind/markbind/pull/1445#discussion_r571418385

packages/vue-components/src/Navbar.vue
jonahtanjz: Alright thanks. Updated :)

packages/vue-components/src/PageNavButton.vue
jonahtanjz: Thanks for the quick clarification 👍. Updated :)

jonahtanjz:
@ang-zeyu Thanks for the review 👍. PR is updated accordingly :)

ang-zeyu (approved):
Lgtm! 👍 

Let's organise the commits if that's ok with you (just separate `addResizeHeaderListener ` will do, with suitable commit titles / messages)

> Implement support for site and page navigation menus on smaller screens

Note the hard limit for the merge / squash commit title is 72 - 8 characters (brackets and pr number github appends)

While not necessary, try to keep < 50 characters too (succinct titles go a long way in making the history easy to [read](https://chris.beams.io/posts/git-commit/))

E.g. Implement mobile site and page navigation menus
would do for the merge commit title

jonahtanjz:
Thanks @ang-zeyu 👍

Have separated the commit for `addResizeHeaderListener` here 784c69cf1432582371fbd249d0e53fa807ff99c9 and updated the merge commit title.

ang-zeyu:
> Thanks @ang-zeyu 👍
> 
> Have separated the commit for `addResizeHeaderListener` here [784c69c](https://github.com/MarkBind/markbind/commit/784c69cf1432582371fbd249d0e53fa807ff99c9) and updated the merge commit title.

@jonahtanjz Thanks, sorry for the over simplification, you'll need to shift and squash the other ones into one commit too (I'm going to do a merge commit here)

jonahtanjz:
Oh right sorry about that 😅. Updated the commits :)

PR #1448, opened 2021-01-16
Add omitFrontmatter attribute to include

### LINKED ISSUE #720: include: give a way to omit frontmatter
Suggestion: give a convenient way to include a file but omit the frontmatter section
e.g., `<include src="foo.md" omit-frontmatter />`

Reason: Some files can be used as independent pages as well as embedded in other pages.

ryoarmanda:

Resolves #720

**Overview of changes:**
- Added `omitFrontmatter` attribute to `<include>` so that the frontmatter data of the included page will not be present to override the "parent" page's frontmatter.
- Added new property `processingOptions` to `Context` to contain options that are relevant for processing, such as discarding front matter data.
- Added two tests to ensure include processing routine preserves/discards the frontmatter data depending on whether the `omitFrontmatter` attribute is present.

**Testing instructions:**

Add omitFrontmatter attribute to include

Including a file combines the front matter of the file with the main
page. When authors wish to include the full content of the file but not
the front matter, they have to make their whole content a segment as a
work around, making the content less concise.

Let's add a new attribute to indicate whether the front matter data of
the included file should be omitted.

---


packages/core/src/html/includePanelProcessor.js
```diff
@@ -228,6 +230,12 @@ function processInclude(node, context, pageSources, variableProcessor, renderMd,
     }
   }
 
+  if (shouldOmitFrontmatter) {
+    const $ = cheerio.load(actualContent);
```
ang-zeyu, line 234: let's modify the [`Context` class](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/Context.js) which contains the parent -> child -> child -> ... "context" instead, and the `frontmatter` processing function

`cheerio.load` is a rather expensive operation (parses the entire page again)

ryoarmanda: Ah I see. Okay, will try to do that.

ryoarmanda: I have updated the approach. Just checking, is it okay if I put the `omitFrontmatter` flag in the `variables` property of the context? I'm afraid on whether it will pollute site/page variables inside.

ang-zeyu: how about adding a new property to the `Context` class? should avoid the concern you mentioned here

ryoarmanda: Alright. I will add another property `processingOptions` (or something along those lines) that will have the `omitFrontmatter` flag (and other processing options if something to be added in the future)

ryoarmanda:
@ang-zeyu , I was trying to check a test that I have written for this and I noticed that when I run `npm run test` on the `core` package, it only reports 4 test suites/files, Jest didn't account for `includePanelProcessorTest.js`, and I might be wrong but I feel like Jest didn't run any tests in that file. Is that intended? I'm not sure how to check my tests now.

ang-zeyu:
> @ang-zeyu , I was trying to check a test that I have written for this and I noticed that when I run `npm run test` on the `core` package, it only reports 4 test suites/files, Jest didn't account for `includePanelProcessorTest.js`, and I might be wrong but I feel like Jest didn't run any tests in that file. Is that intended? I'm not sure how to check my tests now.

Dosen't sound right 

> when I run `npm run test` on the `core` package

is this run on just the core package? or at the root? (as vue-components package has 4 test suites)
I'm seeing 5 tests running on your github action logs as well

if not, try running the `jest includePanelProcessorTest.js` manually and see what happens

> I'm not sure how to check my tests now.

You can create a quick failing test to make sure (e.g. `expect(true).toBe(false)`)

ryoarmanda:
Tried testing within the `core` package, still only 4 test suites:

image

Perhaps because naming of the test file? It is `includePanelProcessorTest.js`. I tried renaming it to `includePanelProcessor.test.js` and Jest detects it, though a lot of failing tests in that file. 😅

ang-zeyu:
> Tried testing within the `core` package, still only 4 test suites:
> 
> image
> 
> Checked the github action logs as well, only 4 too
> 
> Perhaps because naming of the test file? It is `includePanelProcessorTest.js`. I tried renaming it to `includePanelProcessor.test.js` and Jest detects it, though most of the existing tests in that file failed. 😅
> 
> (Edit: Looks like the tests are outdated, the way the includes are structured is different. I might need to rewrite the expected html of the tests...)

Yikes, nice catch, I renamed it at one point

I'll fix it in a separate PR shortly, unless you've already started

ryoarmanda:
> If you're fixing it, the diffs are expected and are a result of the combining markdown processing and <include /> stages. (we used to have nunjucks -> html (just to process <include />s) -> markdown -> html)

I see. I'm on it, will create another PR soon.

ryoarmanda:
@ang-zeyu I am now all done with this PR, can you help take a look on the code and the tests?

ang-zeyu (approved):

PR #1455, opened 2021-01-21
Support sub-menu for dropdown

### LINKED ISSUE #212: Drop-down menus in navbar: support sub-menu items
v1.7.2

current: drop down menus in the top nav bar supports only one level
e.g.,
image

suggested: support multiple levels

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #212.

Sub-menus can be added by nesting dropdowns so that users do not have to remember a new syntax. A new component `<submenu>` is created to handle the logic. `<dropdown>` component is edited such that if a `<dropdown>` is nested, it will be converted to a `<submenu>`. Sub-menus can also contain sub-menus using the same syntax for multiple levels. Currently, sub-menus will appear to the right of a dropdown on desktop version and becomes an accordion on mobile.

Implementation.

**Testing instructions:**
Sub-menus can be added by nesting `<dropdown>`.

For example:
```
<dropdown>
    <li>...</li> // normal dropdown item
    <li>...</li>
    <dropdown> // will be converted to submenu
        <li>...</li> // submenu item
        <li>...</li>
        <dropdown> // 2nd-level submenu
            <li>...</li> // 2nd-level submenu item
            <li>...</li>
        </dropdown>
    </dropdown>
</dropdown>

```
A dropdown with 2 levels of submenu will be created.

Add sub-menu support for dropdown

---

### DIFF
--- a/docs/userGuide/syntax/dropdowns.mbdf
+++ b/docs/userGuide/syntax/dropdowns.mbdf
@@ -56,6 +56,23 @@
 </variable>
 </include>
 
+**You can also use create multi-level submenu by nesting Dropdowns.**
+
+<include src="codeAndOutput.md" boilerplate >
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+<!-- Can simply nest the dropdown syntax -->
+<dropdown header="*Multi-Level Dropdown*" type="primary">
+  <li><a href="#dropdown" class="dropdown-item">Item</a></li>
+  <li><a href="#dropdown" class="dropdown-item">Another item</a></li>
+  <dropdown header="*Submenu*">
+    <li><a href="#dropdown" class="dropdown-item">Item</a></li>
+    <li><a href="#dropdown" class="dropdown-item">Another item</a></li>
+  </dropdown>
+</dropdown>
+</variable>
+</include>
+
 ****Options****
 
 Name | Type | Default | Description
--- a/docs/userGuide/syntax/dropdowns.mbdf
+++ b/docs/userGuide/syntax/dropdowns.mbdf
@@ -61,7 +61,7 @@
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">html</variable>
 <variable name="code">
-<!-- Can simply nest the dropdown syntax -->
+<!-- Nest the dropdown syntax to create dropdown submenus -->
 <dropdown header="*Multi-Level Dropdown*" type="primary">
   <li><a href="#dropdown" class="dropdown-item">Item</a></li>
   <li><a href="#dropdown" class="dropdown-item">Another item</a></li>
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -14,18 +14,11 @@
     </slot>
   </li>
   <submenu v-else-if="isSubmenu" ref="submenu">
-    <slot slot="button" name=button>
-      <a class="submenu-toggle" role="button" :class="{disabled: disabled}">
-        <slot name="_header">
-          <slot name="header"></slot>
-        </slot>
-      </a>
-    </slot>
-    <slot slot="dropdown-submenu" name="dropdown-menu" :class="menuClasses">
-      <ul class="dropdown-menu" :class="menuClasses">
-        <slot></slot>
-      </ul>
-    </slot>
+    <slot slot="button" name=button></slot>
+    <slot slot="_header" name="_header"></slot>
+    <slot slot="header" name="header"></slot>
+    <slot slot="dropdown-submenu" name="dropdown-menu" :class="menuClasses"></slot>
+    <slot></slot>
   </submenu>
   <div v-else ref="dropdown" :class="classes">
     <slot name="before"></slot>
@@ -121,20 +114,12 @@ export default {
       }
     },
     hideDropdownMenu() {
-      this.hideSubmenu();
       this.show = false;
       $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', false));
     },
     showDropdownMenu() {
       this.show = true;
       $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', true));
-    },
-    hideSubmenu() {
-      $(this.$refs.dropdown).findChildren('ul').each(ul => {
-        $(ul).findChildren('li.dropdown-submenu').each(sm => {
-          $(sm).findChildren('ul').each(ul => ul.classList.toggle('show', false));
-        })
-      });
     }
   },
   created() {
@@ -189,44 +174,5 @@ export default {
 
 .dropdown-toggle {
   cursor: pointer;
-}
-
-.submenu-toggle {
-  display: inline-block;
-  width: 100%;
-  padding: .25rem .75rem .25rem 1.5rem;
-}
-
-@media (min-width: 768px) {
-  .submenu-toggle:after {
-    display: inline-block;
-    width: 0;
-    height: 0;
-    vertical-align: .255em;
-    content: "";
-    border-top: .3em solid transparent;
-    border-right: 0;
-    border-bottom: .3em solid transparent;
-    border-left: .3em solid;
-    float: right;
-    margin-top: .5em;
-  }
-}
-
-@media (max-width: 767px) {
-  .submenu-toggle:after {
-    display: inline-block;
-    width: 0;
-    height: 0;
-    margin-left: .255em;
-    vertical-align: .255em;
-    content: "";
-    border-top: .3em solid;
-    border-right: .3em solid transparent;
-    border-bottom: 0;
-    border-left: .3em solid transparent;
-    float: right;
-    margin-top: .5em;
-  }
 }
 </style>
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -17,7 +17,7 @@
     <slot slot="button" name=button></slot>
     <slot slot="_header" name="_header"></slot>
     <slot slot="header" name="header"></slot>
-    <slot slot="dropdown-submenu" name="dropdown-menu" :class="menuClasses"></slot>
+    <slot slot="dropdown-submenu" name="dropdown-menu"></slot>
     <slot></slot>
   </submenu>
   <div v-else ref="dropdown" :class="classes">
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -156,7 +156,10 @@ export default {
       }
       return false
     })
-    $el.findChildren('ul').on('click', 'li>.dropdown-item', e => { this.hideDropdownMenu() })
+    $el.findChildren('ul').on('click', 'li>a', e => { 
+      if (e.target.classList.contains('submenu-toggle')) { return }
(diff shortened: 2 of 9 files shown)


jonahtanjz:
@ang-zeyu this is ready for review and do let me know on how this can be improved on. If the implementation is fine, will proceed to add in tests and documentation as well.

ang-zeyu (commented):
Just a review on the overall approach / issues:

1. I was thinking of using the https://v3.vuejs.org/guide/component-provide-inject.html#working-with-reactivity. 

e.g.

you could do
```
<dropdown>
  <div class="some-wrapper">
    <dropdown></dropdown> // would still work when nested
  </div>
</dropdown>
```

the approach here is more consistent with vue-strap's, but as you might have observed its a rather low level approach that ties itself heavily to the dom. (note vue-strap didn't have this api at the time)

But I would be ok with accepting this as quite a bit of work has been done already. If you're up for it, let's change this over to provide/inject (could even make a separate PR to shift the original implementation to it first if it makes things easier)

2. Positioning of submenus needs to be dynamic (we can assume right position first then flip it over to the left otherwise). (note the submenu may overflow over the screen currently) You could explore the [popperjs](https://popper.js.org/) library to do this (bootstrap-vue uses this internally for its popovers,dropdowns...)

3. Really small nit: Make the submenu show on-hover as well? 🤔

ang-zeyu:
One more note since you're tackling #980 as well: you may want to finish shifting the mobile navbar design to it first, as dropdowns contained within navbars look / behave quite differently - less work to do here

jonahtanjz:
@ang-zeyu I've made some changes accordingly.
> I was thinking of using the https://v3.vuejs.org/guide/component-provide-inject.html#working-with-reactivity.

I didn't use provide/inject in the end as when I was trying to implement this, I managed to remove all the unnecessary communication between the `dropdown` and `submenu` components. Not sure where else to use it as the rest of the search of the virtual DOM seems necessary, for example listening for clicks on `li>a` as these are user defined.

> Positioning of submenus needs to be dynamic (we can assume right position first then flip it over to the left otherwise). (note the submenu may overflow over the screen currently) You could explore the popperjs library to do this (bootstrap-vue uses this internally for its popovers,dropdowns...)

I tried out popperjs but the auto positioning didn't work very well/a little buggy. I switched to using a modified bootstrap javascript which similarly solves this issue. Hopefully that's fine ;)

I've also added tests and updated the docs.

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,177 @@
...
+import $ from './utils/NodeList';
+import positionSubmenu from './utils/submenu';
+
+export default {
+  props: {
+    show: {
+      type: Boolean,
+      default: false,
+    },
+    'class': null,
```
ang-zeyu, line 33: 👀

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,177 @@
...
+  },
+  data() {
+    return {
+      dropright: true,
+      dropleft: false,
+    };
+  },
+  computed: {
+    classes() {
+      return [
```
ang-zeyu, line 47: let's place this directly in the template only for consistency with other components'

docs/userGuide/syntax/dropdowns.mbdf
```diff
@@ -56,6 +56,23 @@
 </variable>
 </include>
 
+**You can also use create multi-level submenu by nesting Dropdowns.**
+
+<include src="codeAndOutput.md" boilerplate >
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+<!-- Can simply nest the dropdown syntax -->
```
ang-zeyu, line 64:
```suggestion
<!-- Nest the dropdown syntax to create dropdown submenus -->
```

packages/vue-components/src/Navbar.vue
```diff
@@ -126,9 +126,19 @@ export default {
...
-    addClassIfDropdown(dropdownLinks, a) {
+    addClassIfDropdown(dropdownLinks, a, li) {
       if (dropdownLinks.includes(a)) {
         a.classList.add('dropdown-current');
+        this.addClassIfSubmenu(a, li);
+      }
+    },
+    addClassIfSubmenu(a, li) {
+      let el = a.parentElement;
+      while (el != li) {
```
ang-zeyu, line 137: `!==`

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,177 @@
...
+    showBool() {
+      return toBoolean(this.show);
+    },
+    slots() {
+      return this.$slots.default;
+    },
+  },
+  methods: {
+    hideSubmenu() {
+      this.show = false;
```
ang-zeyu, line 64: let's use `data()` for properties that could change

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,177 @@
...
+
+<style scoped>
+.dropdown-submenu {
+  color: #212529 !important;
+  padding: 0 !important;
+  position: relative;
+}
+
+@media (max-width: 767px) {
+  .dropdown-submenu > ul {
```
ang-zeyu, line 132: hmm can this be moved to the same media query below?

ang-zeyu (commented):
> I didn't use provide/inject in the end as when I was trying to implement this, I managed to remove all the unnecessary communication between the `dropdown` and `submenu` components.

Not so much for removing unnecessary communication but just allowing parent / child communication across different [compilation scopes](https://vuejs.org/v2/guide/components-slots.html#Compilation-Scope). 

Nice work isolating the implementation as much as possible - only have one not-so-beneficial suggestion where this can be used now. (see comment below)

You could take a look at the following for examples how it could be used more beneficially:
- [bv components](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/dropdown/dropdown-item.js)
- our own `<quiz>` / `<question>` / `<q-option>` components

> Not sure where else to use it as the rest of the search of the virtual DOM seems necessary, for example listening for clicks on `li>a` as these are user defined.

yes this is still necessary unfortunately, due to our use of standard dom elements.

> I tried out popperjs but the auto positioning didn't work very well/a little buggy. I switched to using a modified bootstrap javascript which similarly solves this issue. Hopefully that's fine ;)

Seems ok too 👍 

---


Some strange behaviours:

dropdown-sub
 (main / submenu closing on hover)

Untitled
(this is on mobile view)

Some suggestions:

jonahtanjz:
@ang-zeyu Thanks for the suggestions and clarification 😄. I've updated it accordingly.

> Some strange behaviours:
> (main / submenu closing on hover)

For the main menu, it is closing because when the inner children of the submenu toggle is clicked/hovered, it triggers an `blur` event. I've fixed this by preventing the event from bubbling up. 

As for the submenu closing on hover, it is because when the user hovers over the submenu toggle, it will mirror a click. That is the reason why the second hover will close the submenu. One workaround is to only allow submenus to be opened by hovering over and not by clicking. This is the current implementation for now.

packages/vue-components/src/Submenu.vue
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/Submenu.vue
jonahtanjz: Updated :)

packages/vue-components/src/Navbar.vue
jonahtanjz: Thanks! Updated :)

docs/userGuide/syntax/dropdowns.mbdf
jonahtanjz: Updated :)

packages/vue-components/src/Submenu.vue
jonahtanjz: Alright. Updated :)

packages/vue-components/src/Submenu.vue
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/Dropdown.vue
```diff
@@ -72,6 +79,7 @@ export default {
       return toBoolean(this.disabled);
     },
     isLi () { return this.$parent._navbar || this.$parent.menu || this.$parent._tabset },
+    isSubmenu() { return this.$parent._dropdown || this.$parent._submenu },
```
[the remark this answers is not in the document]
jonahtanjz: Thanks for the clarification. Have added this in :)

packages/vue-components/src/Dropdown.vue
```diff
@@ -13,6 +13,13 @@
       </ul>
     </slot>
   </li>
+  <submenu v-else-if="isSubmenu" ref="submenu">
+    <slot slot="button" name=button></slot>
```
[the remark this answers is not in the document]
jonahtanjz: Thanks! Updated :)

packages/vue-components/src/Dropdown.vue
```diff
@@ -112,7 +120,7 @@ export default {
     showDropdownMenu() {
       this.show = true;
       $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', true));
-    },
+    }
```
ang-zeyu, line 123: let's correct these in another PR; we could enable linting for `dropdown.vue` as well. (the unlinted ones are files that haven't changed much from vue-strap)

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,183 @@
...
+      show: false,
+      dropright: true,
+      dropleft: false,
+    };
+  },
+  computed: {
+    disabledBool() {
+      return toBoolean(this.disabled);
+    },
+    showBool() {
```
ang-zeyu, line 53: this (could replace with just `show`) and `slots()` dosen't seem used

packages/vue-components/src/Submenu.vue
```diff
@@ -0,0 +1,183 @@
...
+    const $el = $(this.$refs.submenu);
+    $el.offBlur();
+    $el.findChildren('a,button').off();
+    $el.findChildren('ul').off();
+  },
+};
+</script>
+
+<style scoped>
+.dropdown-submenu {
```
ang-zeyu, line 126: let's indent these to fit the stylelint rules too (they've never been validated in our scripts for `.vue` files but should be)

ang-zeyu (commented):
Couple more nits:

packages/vue-components/src/Dropdown.vue
jonahtanjz: Alright sure will open another PR for this 👍

packages/vue-components/src/Submenu.vue
jonahtanjz: Thanks!. Replaced all `showBool` with `show`

packages/vue-components/src/Submenu.vue
jonahtanjz: Updated :)

jonahtanjz:
@ang-zeyu PR is updated accordingly :)

ang-zeyu (approved):
Lgtm 👍

PR #1463, opened 2021-01-28
Fix minor bugs in submenu

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes minor bugs present in the `submenu` for the PR #1455. 
1. Removed extra word in UG for `submenu` section under `dropdown`. 
2. Currently, `submenu` will call the method `showSubmenu` again even if it is opened. For some edge cases, this may cause the `submenu` to re-position and cause it to overflow. This can be fixed by preventing `submenu` from calling `showSubmenu` again if it is already opened. 

NIL

**Testing instructions:**
`npm run test`

Fix minor bugs in submenu

---

### DIFF
--- a/packages/vue-components/src/Submenu.vue
+++ b/packages/vue-components/src/Submenu.vue
@@ -100,7 +100,7 @@ export default {
     $el.findChildren('a,button').on('mouseover', (e) => {
       e.preventDefault();
       if (window.innerWidth > 767) {
-        if (this.show || this.disabledBool) { return false; }
+        if (this.disabledBool) { return false; }
         e.currentTarget.click();
         this.showSubmenu();
       }
--- a/packages/vue-components/src/Submenu.vue
+++ b/packages/vue-components/src/Submenu.vue
@@ -100,7 +100,7 @@ export default {
     $el.findChildren('a,button').on('mouseover', (e) => {
       e.preventDefault();
       if (window.innerWidth > 767) {
-        if (this.disabledBool) { return false; }
+        if (this.show || this.disabledBool) { return false; }
         e.currentTarget.click();
         this.showSubmenu();
       }
--- a/docs/userGuide/syntax/dropdowns.mbdf
+++ b/docs/userGuide/syntax/dropdowns.mbdf
@@ -56,7 +56,7 @@
 </variable>
 </include>
 
-**You can also create multi-level submenu by nesting Dropdowns.**
+**You can also create multi-level submenus by nesting Dropdowns.**
 
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">html</variable>
--- a/docs/userGuide/syntax/dropdowns.mbdf
+++ b/docs/userGuide/syntax/dropdowns.mbdf
@@ -56,7 +56,7 @@
 </variable>
 </include>
 
-**You can also use create multi-level submenu by nesting Dropdowns.**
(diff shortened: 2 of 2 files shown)


jonahtanjz:
@ang-zeyu Sorry missed out on a few bugs in the previous PR 😅. Would appreciate if you are able to look through and review it :)

packages/vue-components/src/Submenu.vue
```diff
@@ -100,7 +100,7 @@ export default {
     $el.findChildren('a,button').on('mouseover', (e) => {
       e.preventDefault();
       if (window.innerWidth > 767) {
-        if (this.disabledBool) { return false; }
+        if (this.show || this.disabledBool) { return false; }
```
ang-zeyu, line 103: > Currently, submenu will call the method showSubmenu again even if it is opened. For some edge cases, this may cause the submenu to re-position and cause it to overflow. This can be fixed by preventing submenu from calling showSubmenu again if it is already opened.

Hmm how is this reproduced? Can't seem to find an edge case

If anything, shouldn't repositioning it 'fix' the position? (e.g. open submenu -> not overflown -> scroll down such that it is -> overflown -> hover again -> not overflown)

jonahtanjz: ezgif-7-63eaab259163

The `isRightAlign` method checks whether the current position of the submenu is overflowing on the right side, if it is, then it switches to the left side. However, if it is already on the left side, then it is not overflowing when the check happens hence moving the submenu back to the right side. 

> If anything, shouldn't repositioning it 'fix' the position? (e.g. open submenu -> not overflown -> scroll down such that it is -> overflown -> hover again -> not overflown)

I guess one workaround to solve both of these is to not call the `isRightAlign` method if the submenu is already opened.

jonahtanjz: After looking around I think better way to do it is to keep it on the left if it is opened and already on the left. Else if it is on the right, will check for overflow and flip to left if necessary.

ang-zeyu: Thanks for the gif!

> After looking around I think a better way to do it is to keep it on the left if it is opened and already on the left. Else if it is on the right, will check for overflow and flip to left if necessary.

How does the fix look like? If its too complex, could leave it as is. Repositioning on-hover should really never happen anyway =P (and we don't have reposition on-resize, etc.)

jonahtanjz: The fix I had in mind is to insert this additional rule, on line 64, in the `showSubmenu` method to check if the submenu is already on the left side. If it is, then it will continue to be on the left side.
```
if (!this.dropleft && positionSubmenu.isRightAlign(ul)) {
  this.alignMenuRight();
} else {
  this.alignMenuLeft();
}
```
I'm also fine with leaving it as it is because this scenario is rare and the fix does not fully solve the issue (just preventing a re-positioning if it is on the left side only).

jonahtanjz: Alright sounds good 👍 

Have also updated this PR :)

ang-zeyu: meant to leave the `this.show` in the current pr (prevent repositioning) =X

jonahtanjz: Oops 😬 Added back.

ang-zeyu (approved):

PR #1464, opened 2021-01-31
Update deprecated Vue slot syntax

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

This PR addresses part of #1078 in removing the deprecated Vue slot syntax.

**Overview of changes:**
Update all the internal Vue slot syntax to v-slot (e.g. `slot="example"` to `v-slot:example`)

Since the changes are coupled with the processing of the nodes, we have to be careful in ensuring that existing features are not affected. While I have ensured that all the existing test cases are passing, I'm not sure if this change will break some other existing features. Would be great if I can get some advice here :-/

There is, however, one area that I identified to have regressed but was not able to pinpoint the actual cause of it from my changes. 

Given the following, the item within the slot currently are not appearing as the panel header. I'm still looking into this issue. 
```
<panel  type="info" alt="hi">
  <p slot="header" class="card-title">
    <i><strong>
      <span style="color:#FF0000;">R</span>
      <span style="color:#FF7F00;">A</span>
      <span style="color:#FFFF00;">I</span>
      <span style="color:#00FF00;">N</span>
      <span style="color:#0000FF;">B</span>
      <span style="color:#4B0082;">O</span>
      <span style="color:#9400D3;">W</span>
    </strong></i>
  </p>
</panel>
```

**Testing instructions:**
`npm run test`

Update deprecated Vue slot syntax. 

Our internal Vue slot syntax are deprecated according to official 
Vue documentation. To ensure our codebase is up-to-date, let's 
convert all the deprecated syntax to the latest Vue slot shorthand 
syntax that was introduced in version 2.6.0. 

We will still support the old slot syntax for backward compability, 
by processing and converting the old slot syntax to the new 
shorthand syntax. 

Externally, users will be able to use the latest shorthand syntax for
slots as well, as our processing targets the latest shorthand syntax. 

---

### DIFF
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -11,7 +11,7 @@
 Most component attributes allow a richer form of formatting using slots, denoted by an attribute<strong>^\[S\]^</strong> superscript in the respective components' tables.
 In other cases, when the option is of type "Slot", only the slot option is available.
 
-You can define such a slot within the component by adding a `slot="attribute name"` attribute to any element within the slot.
+You can define such a slot within the component by adding a `slot="attribute_name"` attribute to any element within the slot.
 
 {{ icon_example }}
 
@@ -36,6 +36,21 @@ You can define such a slot within the component by adding a `slot="attribute nam
 <variable name="highlightStyle">html</variable>
 </include>
 </div>
+
+<box type="tip">
+
+You may define a slot by using the shorthand notation `#attribute_name` as well! 
+
+```html
+  <panel>
+    <div #header>
+      Look at the slot shorthand notation directly above. 
+    </div>
+    Now you know how to use the shorthand notation!
+  </panel>
+```
+</box>
+
   
 <modal header="Richer formatting of attributes using slots" id="on-slots" large>
 <include src="advanced.md#slots" />
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -11,14 +11,14 @@
 Most component attributes allow a richer form of formatting using slots, denoted by an attribute<strong>^\[S\]^</strong> superscript in the respective components' tables.
 In other cases, when the option is of type "Slot", only the slot option is available.
 
-You can define such a slot within the component by adding a `slot="attribute_name"` attribute to any element within the slot.
+You can define such a slot within the component by adding a `#attribute_name` attribute to any element within the slot.
 
 {{ icon_example }}
 
 <include src="codeAndOutput.md" boilerplate>
 <variable name="code">
 <panel expanded>
-  <p slot="header" class="card-title">
+  <p #header class="card-title">
     <i><strong>
       <span style="color:#FF0000;">R</span>
       <span style="color:#FF7F00;">A</span>
@@ -37,16 +37,16 @@ You can define such a slot within the component by adding a `slot="attribute_nam
 </include>
 </div>
 
-<box type="tip">
+<box type="info">
 
-You may define a slot by using the shorthand notation `#attribute_name` as well! 
+You may define a slot by using `slot="attribute_name"` as well. 
+However, we **++do not recommend++** using this syntax as it is deprecated and will be removed in the future. 
 
 ```html
   <panel>
-    <div #header>
-      Look at the slot shorthand notation directly above. 
+    <div slot="header">
+      Look at the slot syntax directly above. 
     </div>
-    Now you know how to use the shorthand notation!
   </panel>
 ```
 </box>
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -11,7 +11,7 @@
 Most component attributes allow a richer form of formatting using slots, denoted by an attribute<strong>^\[S\]^</strong> superscript in the respective components' tables.
 In other cases, when the option is of type "Slot", only the slot option is available.
 
-You can define such a slot within the component by adding a `#attribute_name` attribute to any element within the slot.
+You can define such a slot within the component by adding a `#slot_name` attribute to any element within the slot.
 
 {{ icon_example }}
 
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -39,7 +39,7 @@ You can define such a slot within the component by adding a `#attribute_name` at
 
 <box type="info">
 
-You may define a slot by using `slot="attribute_name"` as well. 
+You may define a slot by using `slot="slot_name"` as well. 
 However, we **++do not recommend++** using this syntax as it is deprecated and will be removed in the future. 
 
 ```html
--- a/packages/cli/test/functional/test_site_algolia_plugin/expected/index.html
+++ b/packages/cli/test/functional/test_site_algolia_plugin/expected/index.html
@@ -64,8 +64,8 @@
     <p><strong>Question hint and answer should have algolia-no-index class</strong></p>
     <question>
       Question should not have `algolia-no-index` class
-      <div slot="hint" class="algolia-no-index">Hint should have `algolia-no-index` class</div>
-      <div slot="answer" class="algolia-no-index">Answer should have `algolia-no-index` class</div>
+      <div slot="hint">Hint should have `algolia-no-index` class</div>
+      <div slot="answer">Answer should have `algolia-no-index` class</div>
     </question>
     <p><strong>Tabs except first tab should have algolia-no-index class</strong></p>
     <tabs>
--- a/packages/cli/test/functional/test_site_algolia_plugin/expected/index.html
+++ b/packages/cli/test/functional/test_site_algolia_plugin/expected/index.html
@@ -30,32 +30,33 @@
     <p>
     <p><strong>Test Algolia plugin adds algolia-no-index classes</strong></p>
     <p><strong>Dropdowns should have algolia-no-index class</strong></p>
-    <dropdown class="temp-dropdown algolia-no-index"><template slot="_header">Dropdown</template>
+    <dropdown class="temp-dropdown algolia-no-index"><template #_header>Dropdown</template>
       <li><a class="dropdown-item" href="/">One</a></li>
       <li><a class="dropdown-item" href="/">Two</a></li>
     </dropdown>
     <div class="temp-dropdown-placeholder"></div>
     <p><strong>Modal content should have algolia-no-index class</strong></p>
-    <b-modal id="modal:trigger_id" hide-footer size modal-class="mb-zoom" ref="modal:trigger_id" class="algolia-no-index"><template slot="modal-title">Modal</template>
+    <b-modal id="modal:trigger_id" hide-footer size modal-class="mb-zoom" ref="modal:trigger_id" class="algolia-no-index"><template #modal-title>Modal</template>
       Content should have `algolia-no-index` class
     </b-modal>
     <trigger for="modal:trigger_id">Trigger should not have `algolia-no-index` class</trigger>
     <p><strong>Panels that are not expanded should have algolia-no-index class</strong></p>
-    <panel class="algolia-no-index"><template slot="header">
+    <panel class="algolia-no-index"><template #header>
         <p>Panel</p>
       </template>
       Content
     </panel>
-    <panel expanded><template slot="header">
+    <panel expanded><template #header>
         <p>Panel</p>
       </template>
       Content
     </panel>
     <p><strong>Popover content should have algolia-no-index class</strong></p>
     <span effect="fade" placement="top" data-mb-component-type="popover" v-b-popover.hover.top.html="popoverInnerGetters" class="trigger"><span data-mb-slot-name="header" class="algolia-no-index">Title</span>
-      <div data-mb-slot-name="content" class="algolia-no-index">Content should have `algolia-no-index` class</div>
-      <button class="btn btn-secondary">Trigger should not have `algolia-no-index` class</button>
-    </span>
+      <span data-mb-slot-name="content" class="algolia-no-index">
+        <div>Content should have `algolia-no-index` class</div>
+      </span>
+      <button class="btn btn-secondary">Trigger should not have `algolia-no-index` class</button></span>
     <span effect="fade" placement="top" data-mb-component-type="popover" v-b-popover.hover.top.html="popoverInnerGetters" class="trigger"><span data-mb-slot-name="header" class="algolia-no-index">Title</span><span data-mb-slot-name="content" class="algolia-no-index">Content should have <code class="hljs inline no-lang" v-pre>algolia-no-index</code> class</span>
       <button class="btn btn-secondary">Trigger should not have `algolia-no-index` class</button></span>
     <p><strong>Tooltip content should have algolia-no-index class</strong></p>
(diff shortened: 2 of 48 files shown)


packages/core/src/plugins/algolia.js
```diff
@@ -24,9 +24,9 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    'popover div[slot=content]',
-    'question div[slot=hint]',
-    'question div[slot=answer]',
+    'popover div[v-slot\\:content]',
```
wxwxwxwx9, line 27: @ang-zeyu can I check if these actually should be updated?

packages/core/test/unit/html/NodeProcessor.test.js
```diff
@@ -20,6 +20,8 @@ const processAndVerifyTemplate = (template, expectedTemplate, postProcess = fals
     const nodeProcessor = getNewDefaultNodeProcessor();
 
     if (postProcess) {
+      // need to process node first (convert slot to v-slot) before doing post-processing
+      dom.forEach(node => nodeProcessor.processNode(node, new Context(path.resolve(''))));
```
wxwxwxwx9, line 24: Post-processing involves adding of panel id so we have to process the node first before doing post-processing.

ang-zeyu (commented):
just some initial comments:

---

>There is, however, one area that I identified to have regressed but was not able to pinpoint the actual cause of it from my changes.
>
>Given the following, the item within the slot currently are not appearing as the panel header. I'm still looking into this issue.

Likely this:

(vue docs)
> Note that v-slot can only be added to a <template> (with one exception), unlike the deprecated slot attribute.

packages/core/src/Page/index.js
```diff
@@ -194,15 +194,15 @@ class Page {
       }
 
       // Check if heading / panel is under a panel's header slots, which is handled specially below.
-      const slotParents = $(elem).parentsUntil(context).filter('[slot="header"]').not(elem);
+      const slotParents = $(elem).parentsUntil(context).filter('[v-slot\\:header]').not(elem);
```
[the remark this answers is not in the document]
wxwxwxwx9: Noted! Made the changes :-)

wxwxwxwx9:
@ang-zeyu Thanks so much for the review!

I have made the changes accordingly and tested as much as I could on my side. I think it should work fine. I've verified that the test cases failing in the CI currently are expected due to the changes I've made. Let me know if there happens to be any edge case that I missed out. 

I'm thinking of adding more test cases to our test suite for NodeProcessor to ensure that both deprecated syntax and the new shorthand syntax would work. Should I go ahead with this? 

If so, I'm not sure if I should create test cases for every component since it seems a little overkill. I'm thinking of just creating for a few components (not sure which ones I should pick though, any of them should do in my opinion). What do you think? 

> > There is, however, one area that I identified to have regressed but was not able to pinpoint the actual cause of it from my changes.
> > Given the following, the item within the slot currently are not appearing as the panel header. I'm still looking into this issue.
> 
> Likely this:
> 
> (vue docs)
> 
> > Note that v-slot can only be added to a  (with one exception), unlike the deprecated slot attribute

Thanks for highlighting this to me. Managed to resolve it :-) what I did was to convert all tag names to "template" if there is a shorthand vue slot syntax in the node.

packages/core/src/html/NodeProcessor.js
```diff
@@ -88,17 +89,70 @@ class NodeProcessor {
...
+                Both syntax's slot names match what we are looking for.
+                We will use the the shorthand syntax and delete the deprecated syntax.
+              */
+              delete child.attribs.slot;
+            }
+          }
+          if (isTargetVslotShorthand) {
+            /*
+              User may use tag names like "div" or "p" but we have to change it to "template"
+              for vslot shorthand syntax to work, according to Vue2.6 syntax documenatation.
```
ang-zeyu, line 136: you'll need to "shift it one level deeper" in the dom tree instead (then wrap with `<template #xx>`) as the `<p slot="..."` syntax dosen't erase the `p` element.

packages/core/src/html/NodeProcessor.js
```diff
@@ -123,12 +177,17 @@ class NodeProcessor {
...
     node.children.forEach((child) => {
-      const slot = child.attribs && child.attribs.slot;
-      if (slot) {
-        // Turns <div slot="content">... into <div data-mb-slot-name=content>...
-        child.attribs['data-mb-slot-name'] = slot;
-        delete child.attribs.slot;
+      // Turns <div #content>... into <div data-mb-slot-name=content>...
+      const vslotShorthand = NodeProcessor.getVslotShorthand(child);
+      if (vslotShorthand) {
+        const slotName = vslotShorthand.substring(1, vslotShorthand.length); // remove # from vslot shorthand
```
ang-zeyu, line 188: this seems repeated quite often. how about moving it into `getVslotShorthand`?

packages/core/src/html/NodeProcessor.js
```diff
@@ -88,17 +89,70 @@ class NodeProcessor {
...
+          if (isTargetVslotShorthand) {
+            /*
+              User may use tag names like "div" or "p" but we have to change it to "template"
+              for vslot shorthand syntax to work, according to Vue2.6 syntax documenatation.
+            */
+            child.name = 'template';
+            return true;
+          }
+        }
+        if (hasSlot) {
```
ang-zeyu, line 142: how about:

- at every element (can maintain a set) (`processNode`) that requires slot processing
   - check if its children has `slot="..."`, transform to new syntax if so (e.g. a `transformOldSlotSyntax(node)` method)
   - shift its slot children deeper one level if `node.name !== 'template'` (previous comment) (similarly something like `shiftSlotDeeper(node)`?)
  
that way:
- you'll only need to maintain one flow (your new code targetting `#slotName`)
- the old tests using `slot="xx"` also follow this flow, so we only need to include a simple test for old slot syntax transformation

ang-zeyu (commented):
Mostly looks good! 🙂

Just a couple suggestions (one to remove the old slot processing code entirely) and one issue:

packages/core/src/html/NodeProcessor.js
wxwxwxwx9: I've made the changes but I'm not sure if it's exactly what you're suggesting :O let me know if I have misinterpreted something (especially regarding the set and not having duplicate function calls)!

packages/core/src/html/NodeProcessor.js
wxwxwxwx9: Hmm yup that's possible!

packages/core/src/html/NodeProcessor.js
```diff
@@ -204,8 +223,14 @@ class NodeProcessor {
...
    */
   static _assignPanelId(node) {
-    const slotChildren = node.children && node.children.filter(child => _.has(child.attribs, 'slot'));
-    const headerSlot = slotChildren.find(child => child.attribs.slot === 'header');
+    const slotChildren = node.children
+      && node.children.filter(child => NodeProcessor.getVslotShorthandName(child) !== '');
+
+    const headerSlot = slotChildren.find((child) => {
+      const vslotShorthandName = NodeProcessor.getVslotShorthandName(child);
+      // vslot is guaranteed to exist since we have filtered
```
ang-zeyu, line 231:
```suggestion
      
```
(all good if it dosen't exist too)

docs/userGuide/components/advanced.md
```diff
@@ -36,6 +36,21 @@ You can define such a slot within the component by adding a `slot="attribute nam
 <variable name="highlightStyle">html</variable>
 </include>
 </div>
+
+<box type="tip">
+
+You may define a slot by using the shorthand notation `#attribute_name` as well!
```
ang-zeyu, line 42: let's encourage using this shiny new shorthand (list it upfront), then mention the deprecated syntax would still work too

ang-zeyu (commented):
just a couple of nits:

packages/core/src/plugins/algolia.js
```diff
@@ -24,9 +24,9 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    'popover div[slot=content]',
-    'question div[slot=hint]',
-    'question div[slot=answer]',
+    'popover div["#content"]',
```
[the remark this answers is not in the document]
wxwxwxwx9: Thanks for the spot! Have rectified this :-)

docs/userGuide/components/advanced.md
wxwxwxwx9: Have made the changes! :-)

docs/userGuide/components/advanced.md
```diff
@@ -36,6 +36,21 @@ You can define such a slot within the component by adding a `slot="attribute nam
 <variable name="highlightStyle">html</variable>
 </include>
 </div>
+
+<box type="info">
+
+You may define a slot by using `slot="attribute_name"` as well.
```
ang-zeyu, line 42:
```suggestion
You may define a slot by using `slot="slot_name"` as well. 
```

docs/userGuide/components/advanced.md
```diff
@@ -11,14 +11,14 @@
 Most component attributes allow a richer form of formatting using slots, denoted by an attribute<strong>^\[S\]^</strong> superscript in the respective components' tables.
 In other cases, when the option is of type "Slot", only the slot option is available.
 
-You can define such a slot within the component by adding a `slot="attribute name"` attribute to any element within the slot.
+You can define such a slot within the component by adding a `#attribute_name` attribute to any element within the slot.
```
ang-zeyu, line 14:
```suggestion
You can define such a slot within the component by adding a `#slot_name` attribute to any element within the slot.
```
let's update the old one to make a little more sense

ang-zeyu (commented):
Lgtm 👍

Let's update the tests too

ang-zeyu:
> Update deprecated Vue slot syntax

Let's add a body too, describing the two types of changes here: external - shorthand for user, internal - .vue files

packages/cli/test/functional/test_site_algolia_plugin/expected/index.html
```diff
@@ -81,16 +81,16 @@
...
     </popover>
     <popover effect="fade" header="Title" content="Content as attribute does not require `algolia-no-index` class" placement="top">
       <button class="btn btn-secondary">Trigger should not have `algolia-no-index` class</button></popover>
     <p><strong>Question hint and answer should have algolia-no-index class</strong></p>
     <question>
       Question should not have `algolia-no-index` class
-      <div slot="hint" class="algolia-no-index">Hint should have `algolia-no-index` class</div>
-      <div slot="answer" class="algolia-no-index">Answer should have `algolia-no-index` class</div>
+      <div slot="hint">Hint should have `algolia-no-index` class</div>
+      <div slot="answer">Answer should have `algolia-no-index` class</div>
```
wxwxwxwx9, line 93: Hmm, zeyu, I think this shouldn't be happening. The deprecated slot syntax should be transformed into the shorthand syntax and the node should be shifted one layer deeper. 

I'm currently looking into this to find out why it happens. I discovered that if I remove the <dropdown> from above, the deprecated slot syntax would be transformed properly. 

But I'm still investigating this :O

wxwxwxwx9: For this problem, I realised that it's `insertTemporaryStyles` that is causing the weird behaviour (any element that has slot after dropdown is not transformed properly). We can see many nodes of `<div class="temp-dropdown temp-dropdown-placeholder"></div>` being created before the `<dropdown>`. Can I check why is `insertTemporaryStyles` needed? 

Also, I just realised another problem for adding the `algolia-no-index` class. Currently, what we have for the selectors are 
`
    'popover div[\\#content]',
    'question div[\\#hint]',
    'question div[\\#answer]',
`

This would work for the following code:
`
<popover>
   <div #content>test</div>
</popover>
`

However, our processing actually shifts the node one layer deeper. So we would actually have the following instead:
`
<popover>
   <template #content>
      <div>test</div>
   </template>
</popover>
`

This means that our selectors would not work, and `algolia-no-index` class won't be added successfully. To overcome this, I think we may need to have the following instead:
`
    'popover template[\\#content]',
    'question template[\\#hint]',
    'question template[\\#answer]',
`

But I just wish to clarify if having `algolia-no-index` class on template html element would break anything? :O

wxwxwxwx9: > For this problem, I realised that it's `insertTemporaryStyles` that is causing the weird behaviour (any element that has slot after dropdown is not transformed properly). We can see many nodes of `<div class="temp-dropdown temp-dropdown-placeholder"></div>` being created before the `<dropdown>`.

I think we may be having a bug in the expected html file where we have many duplicated `<div class="temp-dropdown temp-dropdown-placeholder"></div>` generated above of `<dropdown>` to combat FOUC. 

From my debugging, I discovered that the problem stems from the function `insertTemporaryStyles` where we add the dropdown placeholders before the dropdown node itself:
```
    $dropdown.before(placeholder);
    $dropdown.prev().addClass(attributes.class).addClass(TEMP_DROPDOWN_PLACEHOLDER_CLASS);
```

I think this somehow messes up the processing of the nodes. All the nodes (e.g. popover, modal) that follows the dropdown would not get processed. I tested it out by checking if any nodes that are of the name "popover", "modal", "panel" are processed after the dropdown but it doesn't seem so. Instead, we are processing the "dropdown" node multiple times, in place of the rest of the nodes (e.g. "popover", "modal", "panel") for the test. 

I think it can be fixed by adding the placeholder after instead. Here's the modified code:
```
    $dropdown.after(placeholder);
    $dropdown.next().addClass(attributes.class).addClass(TEMP_DROPDOWN_PLACEHOLDER_CLASS);
```

I think the above should fix the problem and we can see that in the expected html file, the duplicated temporary dropdown placeholder are gone and we are left with only one of it (which should be the expected behaviour, correct me if I'm wrong).

ang-zeyu: > Otherwise, we can actually use a more specific css selector to place the algolia-no-index class on the div instead.

Not exactly sure how algolia does its indexing. To be safe, let's go with this instead since `template`s are erased by vue after compilation.

> I think we may be having a bug in the expected html file where we have many duplicated <div class="temp-dropdown temp-dropdown-placeholder"></div> generated above of <dropdown> to combat FOUC.

Thanks for investigating this! : )

> I think this somehow messes up the processing of the nodes. All the nodes (e.g. popover, modal) that follows the dropdown would not get processed.

To confirm -- `.after()` would result in a 'normal / fixed' processing order?

If so let's do that then, preferably in a separate PR

wxwxwxwx9: > To confirm -- `.after()` would result in a 'normal / fixed' processing order? (as a 'sibling' operation seems to be causing this)
> 
> If so let's do that then, preferably in a separate PR

Can I clarify about "normal / fixed" processing order? :O 

Suppose we have: 
```
<modal></modal>
<popover></popover>
<panel></panel>
```

You mean if it would be processed in the sequential order, right? Like `modal` -> `popover` -> `panel`?

If so, yup, .after() would give us the correct sequential processing order.

ang-zeyu: e.g.

```
// suppose these are the last 3 elements at the same level
<dropdown>
<div class="temp-dropdown-placeholder></div> // gets inserted here with after()
<panel> // this actually gets processed
```

wxwxwxwx9: Yup, that's the behaviour I'm seeing currently :-)

ang-zeyu (commented):
a few more minor suggestions I missed earlier:

packages/core/src/html/NodeProcessor.js
```diff
@@ -88,17 +89,35 @@ class NodeProcessor {
...
    * @param attribute Attribute name to process
    * @param isInline Whether to process the attribute with only inline markdown-it rules
    * @param slotName Name attribute of the <slot> element to insert, which defaults to the attribute name
    */
   _processAttributeWithoutOverride(node, attribute, isInline, slotName = attribute) {
     const hasAttributeSlot = node.children
-      && node.children.some(child => _.has(child.attribs, 'slot') && child.attribs.slot === slotName);
+      && node.children.some((child) => {
+        const vslotShorthandName = NodeProcessor.getVslotShorthandName(child);
+        return vslotShorthandName && vslotShorthandName === slotName;
```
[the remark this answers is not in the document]
wxwxwxwx9: Yup, that's a good idea!

packages/core/src/html/NodeProcessor.js
```diff
@@ -124,11 +143,11 @@ class NodeProcessor {
...
-        child.attribs['data-mb-slot-name'] = slot;
-        delete child.attribs.slot;
+      // Turns <div #content>... into <div data-mb-slot-name=content>...
+      const vslotShorthandName = NodeProcessor.getVslotShorthandName(child);
+      if (vslotShorthandName) {
+        child.attribs['data-mb-slot-name'] = vslotShorthandName;
+        delete child.attribs[`#${vslotShorthandName}`];
       }
       // similarly, need to transform templates to avoid Vue parsing
       if (child.name === 'template') {
```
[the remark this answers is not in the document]
wxwxwxwx9: Ahh right, thanks for the catch!

ang-zeyu (approved):
Lgtm 👍 Thanks! @wxwxwxwx9

PR #1468, opened 2021-02-07
Fix algolia selectors for no-indexing

wxwxwxwx9:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

This PR fixes the erroneous algolia selectors after the PR fix from #1467. 

**Overview of changes:**
Updated the algolia selectors to correctly select elements to add `algolia-no-index`.

If there would be other selectors that need to be fixed in algolia. 

**Testing instructions:**
`npm run test`

Fix algolia selectors for no-indexing. 

The algolia selectors are not selecting the correct elements to insert
the `algolia-no-index` class to exclude the element from indexing. 

Let's fix the algolia selectors by updating them to select the correct
elements.

---

### DIFF
--- a/packages/core/src/plugins/algolia.js
+++ b/packages/core/src/plugins/algolia.js
@@ -24,8 +24,7 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    // to target popover and tooltips
-    '[data-mb-component-type] [data-mb-slot-name="content"], [data-mb-slot-name="_content"]',
+    '[data-mb-component-type] :not(span)[data-mb-slot-name]',
     'question div[slot=hint]',
     'question div[slot=answer]',
     'tab:nth-of-type(n+2)',
--- a/packages/core/src/plugins/algolia.js
+++ b/packages/core/src/plugins/algolia.js
@@ -24,7 +24,8 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    '[data-mb-component-type] :not(span)[data-mb-slot-name]',
+    // to target both popover and tooltip
+    '[data-mb-component-type] [data-mb-slot-name]',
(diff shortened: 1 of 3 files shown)


wxwxwxwx9:
Hi @ang-zeyu, can I get your help to review this PR? :-)

packages/core/src/plugins/algolia.js
```diff
@@ -24,11 +24,11 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    'popover div[slot=content]',
+    'span[data-mb-component-type="popover"] div[data-mb-slot-name="content"]',
```
ang-zeyu, line 27: let's generalise this so tooltips get the fix too, also non `<div>` elements

packages/core/src/plugins/algolia.js
```diff
@@ -24,11 +24,12 @@ function addNoIndexClasses(content) {
     'dropdown',
     'b-modal',
     'panel:not([expanded])',
-    'popover div[slot=content]',
+    // to target popover and tooltips
+    '[data-mb-component-type] [data-mb-slot-name="content"], [data-mb-slot-name="_content"]',
```
ang-zeyu, line 28: how about:

```suggestion
    '[data-mb-component-type] [data-mb-slot-name]',
```
so the popover `header`s don't get indexed too

wxwxwxwx9: Thanks for the review! Made the changes as per our discussion :-)

packages/core/src/plugins/algolia.js
wxwxwxwx9: Thanks for the review! Made the changes as per our discussion :-)

ang-zeyu (approved):
Nice work. Thanks for the catches here @wxwxwxwx9

PR #1469, opened 2021-02-07
Restyle minimal panel header to full opacity

### LINKED ISSUE #1273: Minimal panels: don't dim the heading
Current: 
image

Suggestion: Do not dim the heading, as the user can always use `%%` to dim the heading if so desired.

ryoarmanda:

Resolves #1273 

**Overview of changes:**
- Removed the 50% opacity styling of the panel header during collapsed and minimized states.
- As the heading is now not dimmed, there is no need for the dim-to-full hover transition, so I also removed it.

- Originally I thought to keep the existing dim-to-full hover transition for the case when the user explicitly wants the header to be dim, but as discussed in the issue comments, there may be extra work needed in achieving that, such as the user has to specify dim header through a new panel attribute rather than the usual `%%` for text, which may be a bit unnatural. So, I decided to scrap it. I have no preference though so if you think we should keep it do let me know. @ang-zeyu 

**Testing instructions:**
Test with `-d` flag.

Restyle minimal panel header to full opacity

---

### DIFF
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -159,7 +159,6 @@ export default {
 
   .morph-title {
     padding: 0 0.3em;
-    color: black;
     border-color: black;
     background-color: transparent;
     vertical-align: top;
@@ -181,10 +180,6 @@ export default {
     flex-direction: column;
     border: 0;
     margin: 0;
-  }
-
-  .card-title {
-    color: black;
   }
 
   .header-fade-enter {
--- a/packages/vue-components/src/panels/MinimalPanel.vue
+++ b/packages/vue-components/src/panels/MinimalPanel.vue
@@ -17,7 +17,7 @@
           <span
             v-show="!isHeaderAtBottom"
             ref="headerWrapper"
-            :class="['card-title', 'card-title-transparent', { 'ellipses': !hasHeaderBool }]"
+            :class="['card-title', { 'ellipses': !hasHeaderBool }]"
           >
             <span class="card-title-inline"><slot name="header"></slot></span>
             <span
@@ -159,14 +159,13 @@ export default {
 
   .morph-title {
     padding: 0 0.3em;
-    color: rgba(0, 0, 0, 0.5);
+    color: black;
     border-color: rgba(0, 0, 0, 0.5);
     background-color: transparent;
     vertical-align: top;
   }
 
   .morph-title:hover, .morph-title:active, .morph-title:focus {
-    color: black;
     border-color: black;
     background-color: rgba(244, 244, 244, 0.3);
   }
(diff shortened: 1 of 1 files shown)


wxwxwxwx9:
Hi @ryoarmanda I think the the non-minimized version looks fine. My only concern is when the minimal panel is in its minimized form. 

Screen Shot 2021-02-09 at 3 11 25 PM

In the minimized form, the hover behaviour of the minimal panel seems to have quite the opposite behaviour as compared to the rest of the other panel types. Specifically, the other panels have the "inverting" of the colors feature (where the text becomes light and the background of the panel becomes dark. I think maybe we can replicate this behaviour somehow for minimal panel as well? 

My two reasons for suggesting this: 1) makes it more aligned with the rest of the panels in terms of interactivity, 2) seems to give the user a better sense of UX affordance for the minimal panel (currently there isn't much contrast between the hover state and non-hover state). 

However, I think this is a highly subjective issue so we may want to explore other options as well. But what do you think? :-)

ryoarmanda:
> In the minimized form, the hover behaviour of the minimal panel seems to have quite the opposite behaviour as compared to the rest of the other panel types. Specifically, the other panels have the "inverting" of the colors feature (where the text becomes light and the background of the panel becomes dark). I think maybe we can replicate this behaviour somehow for minimal panel as well?

Ahh yes, I think so too, we can have the invert color transition as well for consistency with the rest of the minimized panels. Will do 👍

ryoarmanda:
I have added the invert color styling for the minimized version of minimal panels. Would like your review on the color schemes :smile: thanks! @wxwxwxwx9 @ang-zeyu

packages/vue-components/src/panels/MinimalPanel.vue
```diff
@@ -183,13 +183,8 @@ export default {
...
 
-  .card-title-transparent {
-    opacity: 0.5;
-    transition: opacity 0.5s;
-  }
-
-  .card:hover .card-title-transparent {
-    opacity: 1;
+  .card-title {
+    color: black;
```
ang-zeyu, line 187: Thanks! 🙂 
Just one nit:

hmm is this still needed? (also for `.morph-title` above)

ryoarmanda: Oh, I explicitly put black color just as a safety if there's conflicting styles from any stylings from the panel's parent or something. Actually, I don't really know how likely that is to happen though, what do you think? If it's not likely, I can remove it.

ang-zeyu: ahh I see, I was thinking of letting these inherit the color (default behaviour) actually since it might be more convenient / expected for the author.

We should be good here since you've ensured the critical one is `color: white` under `morph-title:hover`. (since the contrast is essential against a dark background here)

Sound good?

ryoarmanda: Yeah, looking at it, it made more sense to inherit as a default. Okay then, will remove the explicit black color. 👍

ang-zeyu (approved):
Lgtm! 👍

PR #1473, opened 2021-02-09
Implement Markdown rules for bigger / smaller text

### LINKED ISSUE #28: Support an easy syntax for bigger/smaller text size
Currently we use block quote formatting to show text in bigger size. This is an abuse of syntax. Instead, we should tweak the block quote formatting to actually look like quotes as we do have an actual use case for showing quotes in our eContent.

On the other hand, we could support an easy syntax to indicate whether to make text bigger/smaller relative to the default size. e.g. `+++` `---`

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #28.

**Overview of changes:**
I reused our internal existing plugin, *markdown-it-dimmed*, as a dynamic markdown-it plugin that allows us to define double-delimiter inline rules for the following syntaxes: ++ (larger text), -- (smaller text), $$ (underline). 

++ used to be the syntax for underline but as discussed in the issue thread #28, we will use it to represent larger text syntax instead, and use $$ to represent underline. 

Whether the implementation approach can be improved. 

**Testing instructions:**
Create a new `md` file in your Markbind and test out the following syntaxes: `++larger++`, `--small--`, `$$underline$$`.

You should be able to mix-and-match the different syntaxes and achieve a combined result (e.g. `%%++larger_dimmed++%%` should give you a text that becomes larger and is dimmed). 

Implement Markdown rules for bigger / smaller text.

There is no easy way for users to increase or decrease their text size.

Let's deprecate `++` for underline and use `$$` to represent it instead.
In turn, we will use `++` to represent larger text and `--` to represent
smaller text.

---


wxwxwxwx9:
We probably should remove **markdown-it-ins** from our packages since we don't need it anymore.

damithc:
Wait, don't we already use `++` for underlines?

wxwxwxwx9:
Hmm @damithc, what I understand from the issue thread is that moving forward we will be using `$$` for underlines, and use `++` for larger texts instead :O 

Have I misunderstood the requirements?

docs/userGuide/syntax/textStyles.mbdf
```diff
@@ -26,7 +26,7 @@ Syntax added by MarkBind:
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">markdown</variable>
 <variable name="code">
-****Super Bold****, ++Underline++, ==Highlight==, %%Dim%%, Super^script^, Sub~script~
+****Super Bold****, $$Underline$$, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~
```
ryoarmanda, line 29: Users that are already used to use `$$` for underlining might not catch that the new syntax is `++`. Should we note in the docs that there's a change from this version onwards?

packages/core/src/lib/markdown-it/index.js
```diff
@@ -7,9 +7,16 @@ const slugify = require('@sindresorhus/slugify');
...
+
 // markdown-it plugins
+
+markdownIt.use(createDoubleDelimiterInlineRule('%%', 'dimmed', 'emphasis'))
+  .use(createDoubleDelimiterInlineRule('$$', 'underline', 'dimmed'))
+  .use(createDoubleDelimiterInlineRule('++', 'large', 'underline'))
+  .use(createDoubleDelimiterInlineRule('--', 'small', 'large'));
+
 markdownIt.use(require('markdown-it-mark'))
-  .use(require('markdown-it-ins'))
```
ryoarmanda, line 12: I'm not really well versed on `markdown-it-ins` and its supposed usage in our project, was it related to the double-delimiter plugin? If not, I think it's better to address this in a code maintenance PR, to keep the changes here about the new syntax.

If we end up really not use the plugin anymore, we can also remove it from our dependencies in `package.json` as well.

docs/userGuide/syntax/textStyles.mbdf
wxwxwxwx9: Hmm good point. On second thought, it might not be good to change the syntax for underlining straightaway. However, it seems it won't be easy to introduce some form of backward compatibility for this new feature. 

Do you think it would be better to introduce some form of intermediary change (e.g. support both `++` and `$$` as underline and use another syntax for larger text for in the meantime, before we fully deprecate `++` for underlining and use it for larger text)?

@damithc would you have any advice for this?

packages/core/src/lib/markdown-it/index.js
wxwxwxwx9: `markdown-it-ins` is actually to enable the `++` syntax for underlining. 

Since in this PR we are using `++` for larger text and `$$` for underlining instead, this plugin is no longer needed. And I think it's better to remove it since it may interfere with the parsing for the `++` syntax. 

Thus, I'm not sure if we should address it here or in a code maintenance PR :-/ 

Yup, I think it should be removed from `package.json` as well, but would need to seek advice from @ang-zeyu for it.

docs/userGuide/syntax/textStyles.mbdf
damithc: > @damithc would you have any advice for this?

As I'm the only user as of now, just let me know when the change happens so that I can migrate the code when I upgrade the version.

packages/core-web/src/styles/markbind.css
```diff
@@ -142,6 +142,18 @@ footer {
     color: #777;
 }
 
+.large {
+    font-size: 125%;
+}
+
+.small {
+    font-size: 75%;
+}
```
ryoarmanda, line 151: Just a minor opinion here, I think it would be somewhat satisfying (not sure if it is "better" though, feel free to discuss) to have 80% percentage for small. That way, `++--text--++` will have exactly the same size as plain `text` as the percentages would be reciprocal 😄 

The percentage difference from the normal size against larger/smaller size would be unequal, but 5% difference from 75% could also be deemed small enough to be negligible. Again, this is pretty much subjective, so it is up to you (and others if anyone wants to weigh in) =)

packages/core/src/lib/markdown-it/index.js
ryoarmanda: Oh I see, I thought it was used for other purposes. I feel it can be kept here. :+1:

ang-zeyu: let's remove it (plugin), since we aren't using it anywhere else

docs/userGuide/syntax/textStyles.mbdf
ang-zeyu: I'll be able to pull _just_ the `$$...$$` syntax into another branch separately and get a new release out with just that once this is done; but a PR against the branch would be welcome as well 🙂 https://github.com/MarkBind/markbind/tree/v2.7.x-hotfixes

packages/core-web/src/styles/markbind.css
ang-zeyu: agree here, its also consistent with bootstrap's reboot css sizing for the html `<small>` tag

packages/core/src/lib/markdown-it/markdown-it-dimmed.js
```diff
@@ -1,125 +0,0 @@
-'use strict';
```
ang-zeyu, line 1: one minor nit: let's find out where this was from and add in the license notice, if there's one!

packages/core-web/src/styles/markbind.css
wxwxwxwx9: Thanks for the suggestion! @ryoarmanda :-)

packages/core/src/lib/markdown-it/markdown-it-dimmed.js
wxwxwxwx9: I think the code was adapted using the below as reference:
https://github.com/markdown-it/markdown-it/blob/master/lib/rules_inline/strikethrough.js

The license of the repository can be found here:
https://github.com/markdown-it/markdown-it/blob/master/LICENSE

In fact, we have already written similar licensing notice before in `markdown-it-footnotes.js` so I'm going to copy the template over :-)

docs/userGuide/syntax/textStyles.mbdf
wxwxwxwx9: > I'll be able to pull _just_ the `$$...$$` syntax into another branch separately and get a new release out with just that once this is done; but a PR against the branch would be welcome as well 🙂 

Have made the intermediary PR here :-)

https://github.com/MarkBind/markbind/pull/1476

ang-zeyu (approved):
Lgtm 👍 

let's update all the other existing usages of `++...++` in our docs and template sites too

wxwxwxwx9:
I have updated all the existing usages of `++...++` :-)

ang-zeyu (approved):

Issue #1406, opened 2020-12-05
Vue compilation at build-time and SSR


ang-zeyu:
**Is your request related to a problem?**

MarkBind has rather observable fouc. Some temporary styles are in place to combat this, but does not solve the root of the issue:
- Vue compilation for the page happens in-browser. This very costly especially for a static site generator where a single page can get very, very large
- No SSR in place

**Describe the solution you'd like**

1. At the very least, we should compile the vue app (i.e. the page) at build time. This should eliminate FOUC for 80% of use cases.
2. Explore SSR with our setup

**Additional context**

#889 (FOUC)

PR #1474, opened 2021-02-10
Toggle fixed header on scroll

### LINKED ISSUE #980: Redesign the navbar

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #980.

As discussed, fixed header might take up too much space especially on devices with smaller screen, leaving less space for content. A workaround is to hide the fixed header when scrolling down to give more space to content. Header can still be easily accessed by scrolling up.

Should the header hide on both mobile and desktop version when scrolling down? Since this issue affect mainly mobile users, we can continue to fixed the header for desktop version so the header does not need to be hidden there. The current implementation works on both versions.

**Testing instructions:**
`npm run test`

Implement fixed header toggling on scroll

---

### DIFF
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -52,7 +52,7 @@ function detectAndApplyFixedHeaderStyles() {
       }`);
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
 
-  const resizeHeader = () => {
+  const adjustHeaderClasses = () => {
     const newHeaderHeight = headerSelector.height();
     const sheets = document.styleSheets;
     for (let i = 0; i < sheets.length; i += 1) {
@@ -83,23 +83,12 @@ function detectAndApplyFixedHeaderStyles() {
 
   const toggleHeaderOverflow = () => {
     const headerMaxHeight = headerSelector.css('max-height');
+    // reset overflow when header shows again to allow content
+    // in the header such as search dropdown etc. to overflow
     if (headerMaxHeight === '100%') {
       headerSelector.css('overflow', '');
-      resizeHeader();
+      adjustHeaderClasses();
     }
-  };
-
-  const addResizeHeaderListener = () => {
-    const resizeObserver = new ResizeObserver(() => {
-      const headerMaxHeight = headerSelector.css('max-height');
-      // set overflow to hidden if max-height is not 100%
-      if (headerMaxHeight !== '100%') {
-        headerSelector.css('overflow', 'hidden');
-        return;
-      }
-      resizeHeader();
-    });
-    resizeObserver.observe(headerSelector[0]);
   };
 
   let lastOffset = 0;
@@ -118,7 +107,16 @@ function detectAndApplyFixedHeaderStyles() {
     lastOffset = currentOffset;
   };
 
-  addResizeHeaderListener();
+  const resizeObserver = new ResizeObserver(() => {
+    const headerMaxHeight = headerSelector.css('max-height');
+    // hide header overflow when user scrolls to support transition effect
+    if (headerMaxHeight !== '100%') {
+      headerSelector.css('overflow', 'hidden');
+      return;
+    }
+    adjustHeaderClasses();
+  });
+  resizeObserver.observe(headerSelector[0]);
   headerSelector[0].addEventListener('transitionend', toggleHeaderOverflow);
   window.addEventListener('scroll', toggleHeaderOnScroll);
 }
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -52,34 +52,54 @@ function detectAndApplyFixedHeaderStyles() {
       }`);
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
 
-  const addResizeHeaderListener = () => {
-    const resizeObserver = new ResizeObserver(() => {
-      const newHeaderHeight = headerSelector.height();
-      const sheets = document.styleSheets;
-      for (let i = 0; i < sheets.length; i += 1) {
-        const rules = sheets[i].cssRules;
-        // eslint-disable-next-line lodash/prefer-get
-        if (rules && rules[0] && rules[0].selectorText) {
-          switch (rules[0].selectorText) {
-          case '.fixed-header-padding':
-            sheets[i].deleteRule(0);
-            sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important }`);
-            break;
-          case 'span.anchor':
-            rules[0].style.top = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
-            break;
-          case 'span.card-container::before':
-            rules[0].style.marginTop = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
-            rules[0].style.height = `calc(${newHeaderHeight}px + ${bufferHeight}rem)`;
-            break;
-          case '.nav-menu-open':
-            rules[0].style.maxHeight = `calc(100% - ${newHeaderHeight}px + 50px)`;
-            break;
-          default:
-            break;
-          }
+  const resizeHeader = () => {
+    const newHeaderHeight = headerSelector.height();
+    const sheets = document.styleSheets;
+    for (let i = 0; i < sheets.length; i += 1) {
+      const rules = sheets[i].cssRules;
+      // eslint-disable-next-line lodash/prefer-get
+      if (rules && rules[0] && rules[0].selectorText) {
+        switch (rules[0].selectorText) {
+        case '.fixed-header-padding':
+          sheets[i].deleteRule(0);
+          sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important }`);
+          break;
+        case 'span.anchor':
+          rules[0].style.top = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
+          break;
+        case 'span.card-container::before':
+          rules[0].style.marginTop = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
+          rules[0].style.height = `calc(${newHeaderHeight}px + ${bufferHeight}rem)`;
+          break;
+        case '.nav-menu-open':
+          rules[0].style.maxHeight = `calc(100% - ${newHeaderHeight}px + 50px)`;
+          break;
+        default:
+          break;
         }
       }
+    }
+  };
+
+  const toggleHeaderOverflow = () => {
+    const headerMaxHeight = headerSelector.css('max-height');
+    if (headerMaxHeight === '100%') {
+      headerSelector.css('overflow', '');
+      resizeHeader();
+    }
+  };
+
+  const addResizeHeaderListener = () => {
+    const resizeObserver = new ResizeObserver(() => {
+      const headerMaxHeight = headerSelector.css('max-height');
+      // set overflow to hidden if max-height is not 100%
+      if (headerMaxHeight !== '100%') {
+        headerSelector.css('overflow', 'hidden');
+        // re-check after 0.6s to account for transition time
+        setTimeout(toggleHeaderOverflow, 600);
+        return;
+      }
+      resizeHeader();
     });
     resizeObserver.observe(headerSelector[0]);
   };
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -86,9 +86,12 @@ function detectAndApplyFixedHeaderStyles() {
 
   let lastOffset = 0;
   const toggleHeaderOnScroll = () => {
+    // prevent toggling of header on desktop site
     if (window.innerWidth > 767) { return; }
     const currentOffset = window.pageYOffset;
-    if ((window.innerHeight + currentOffset) >= document.body.offsetHeight) { return; }
+    const isEndOfPage = (window.innerHeight + currentOffset) >= document.body.offsetHeight;
+    // to prevent page from auto scrolling when header is toggled at the end of page
+    if (isEndOfPage) { return; }
     if (currentOffset > lastOffset) {
       headerSelector.addClass('hide-header');
     } else {
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -83,7 +83,20 @@ function detectAndApplyFixedHeaderStyles() {
     });
     resizeObserver.observe(headerSelector[0]);
   };
+
+  let lastOffset = 0;
+  const toggleHeaderOnScroll = () => {
+    const currentOffset = window.pageYOffset;
+    if (currentOffset > lastOffset) {
+      headerSelector.addClass('hide-header');
+    } else {
+      headerSelector.removeClass('hide-header');
+    }
+    lastOffset = currentOffset;
+  };
+
   addResizeHeaderListener();
+  setTimeout(() => window.addEventListener('scroll', toggleHeaderOnScroll), 500);
 }
 
 function updateSearchData(vm) {
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -95,8 +95,6 @@ function detectAndApplyFixedHeaderStyles() {
       // set overflow to hidden if max-height is not 100%
       if (headerMaxHeight !== '100%') {
         headerSelector.css('overflow', 'hidden');
-        // re-check after 0.6s to account for transition time
-        setTimeout(toggleHeaderOverflow, 600);
         return;
       }
       resizeHeader();
@@ -121,6 +119,7 @@ function detectAndApplyFixedHeaderStyles() {
   };
 
   addResizeHeaderListener();
+  headerSelector[0].addEventListener('transitionend', toggleHeaderOverflow);
   window.addEventListener('scroll', toggleHeaderOnScroll);
 }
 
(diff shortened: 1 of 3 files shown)


packages/vue-components/src/Overlay.vue
```diff
@@ -57,7 +57,12 @@ export default {
...
   methods: {
     toggleNavMenu() {
-      if (!this.show) { publish('closeOverlay'); }
+      if (!this.show) {
+        publish('closeOverlay');
+        document.body.style.overflow = 'hidden';
+      } else {
+        document.body.style.removeProperty('overflow');
+      }
       this.show = !this.show;
```
jonahtanjz, line 66: This extra block is needed so that when the `overlay` is opened, it prevents the header from hiding as a result of scroll chaining scrolling the body.

packages/vue-components/src/Overlay.vue
```diff
@@ -57,7 +57,12 @@ export default {
   },
   methods: {
     toggleNavMenu() {
-      if (!this.show) { publish('closeOverlay'); }
+      if (!this.show) {
+        publish('closeOverlay');
+        document.body.style.overflow = 'hidden';
+      } else {
+        document.body.style.removeProperty('overflow');
+      }
```
jonahtanjz, line 65: This extra block is needed so that when the `overlay` is opened, it prevents the header from hiding as a result of scroll chaining scrolling the body.

raysonkoh:
Would it be possible to have a transition effect from no header <-> show header? Without it, I feel that the transition might be a little abrupt. What do you think @jonahtanjz ?

Screen-Recording-2021-02-10-at-11 59 55-PM

raysonkoh:
> Should the header hide on both mobile and desktop version when scrolling down? Since this issue affect mainly mobile users, we can continue to fixed the header for desktop version so the header does not need to be hidden there. The current implementation works on both versions.

I feel that we can continue to fix the header for desktop version since they already have enough screen real estate and the header wouldn't occupy much of the screen.

packages/vue-components/src/Overlay.vue
wxwxwxwx9: Can I just clarify what you mean by this? 

Screen Shot 2021-02-11 at 12 01 31 AM

Is it something like the screenshot above where the navbar is hidden partially when you scroll down?

wxwxwxwx9: > Would it be possible to have a transition effect from no header <-> show header? Without it, I feel that the transition might be a little abrupt. What do you think @jonahtanjz ?
> 
> Screen-Recording-2021-02-10-at-11 59 55-PM

I agree with @raysonkoh on this. Actually, I was thinking if the overlay effect that I mentioned above would be a suitable "transition" effect? I don't see many transition effect options that are suitable for this, to be honest :O

wxwxwxwx9:
@jonahtanjz There's currently a weird behavior when I overscroll the page. 

https://user-images.githubusercontent.com/37920115/107542428-223c2200-6c03-11eb-8e07-55ff354ed8ca.mov

packages/vue-components/src/Overlay.vue
wxwxwxwx9: I've dabbled around the transitions and here's something I came up with. It's not the best solution but I was hoping that it may help you come up with better ideas :-)

https://user-images.githubusercontent.com/37920115/107546913-d8096f80-6c07-11eb-878f-5897fc650336.mov

packages/core-web/src/index.js
```diff
@@ -83,7 +83,20 @@ function detectAndApplyFixedHeaderStyles() {
...
+    if (currentOffset > lastOffset) {
+      headerSelector.addClass('hide-header');
+    } else {
+      headerSelector.removeClass('hide-header');
+    }
+    lastOffset = currentOffset;
+  };
+
   addResizeHeaderListener();
+  setTimeout(() => window.addEventListener('scroll', toggleHeaderOnScroll), 500);
```
wxwxwxwx9, line 99: Can I check why do we need this setTimeout here? :O

packages/vue-components/src/Overlay.vue
jonahtanjz: Thanks for the suggestions @wxwxwxwx9 

> Can I just clarify what you mean by this?

This is when either the site or page nav is opened and the user scroll to the end, if the user continues scrolling, it will scroll the main body as well. Can read more about it here: https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior. You can also try it out on the current user guide's site nav: https://markbind-master.netlify.app/userguide/gettingstarted.

> I agree with @raysonkoh on this. Actually, I was thinking if the overlay effect that I mentioned above would be a suitable "transition" effect? I don't see many transition effect options that are suitable for this, to be honest :O

Yup agree with you and @raysonkoh on the transition and I was planning to implement it but was kind of stuck on this as the standard css transitions doesn't really work well here. Will have to find alternatives.

> I've dabbled around the transitions and here's something I came up with using max-height css transition. It's not the best solution but I was hoping that it may help with the discussion :-)

This actually looks good. Thanks for looking into this 👍. I'm ok with using this. @raysonkoh what do you think about this?

jonahtanjz:
> Would it be possible to have a transition effect from no header <-> show header? Without it, I feel that the transition might be a little abrupt. What do you think @jonahtanjz ?

Thanks for the suggestion @raysonkoh 👍. Agree with you on the transition. @wxwxwxwx9 came up with a possible transition effect on the navbar discussed above. What do you think about that?

packages/core-web/src/index.js
jonahtanjz: This is for rare situations where the size of the navbar can change when the page is refreshed. For example, when a `<box dismissable>` is placed in the header. This will cause the navbar to change size while rendering resulting in an unwanted scroll event to fire before the page is fully loaded. Although not the best way to counter this, a delay here would prevent the unwanted scroll event to be picked up. I believe the timeout would not be necessary anymore after #1406 is implemented.

jonahtanjz: https://user-images.githubusercontent.com/47470981/107607087-2a33aa80-6c73-11eb-881a-33ff299bc27a.mp4

jonahtanjz:
> @jonahtanjz There's currently a weird behavior when I overscroll the page.

Thanks for the catch @wxwxwxwx9 👍

packages/vue-components/src/Overlay.vue
raysonkoh: Yup I think it looks good. We can go with this. 👍

wxwxwxwx9: > This is when either the site or page nav is opened and the user scrolls to the end, if the user continues scrolling, it will scroll the main body as well. Can read more about it here: https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior. You can also try it out on the current user guide's site nav: https://markbind-master.netlify.app/userguide/gettingstarted.

Ah, I see! Understood. I tried out another solution: place `overscroll-behaviour-y: none` under `.nav-menu`. It's working on my side. Can you try if it works for you? I think it's a more succinct solution but I'm not sure if it would interfere with the features you have implemented so far (e.g. siteNav, navBar, and this PR). What do you think? :-)

packages/core-web/src/index.js
wxwxwxwx9: Thanks for the explanation! I'm not sure if `setTimeout` would be appropriate in the case where the user has slow loading speed (e.g. page takes longer than the stipulated 500ms to load). I think the solution here won't work in this scenario. However, I'm also unsure if there is a better way to solve this issue :-/ 

I tried looking through the list of event hooks that we can possibly use to detect when the page is fully-rendered but I don't think I found any.

https://developer.mozilla.org/en-US/docs/Web/Events

I also tried a few other hacky ways to get a solution to work but no luck so far. 

But a surefire way to solve this problem at some expense can be `window.history.scrollRestoration = 'manual';`, where we always start off at the top whenever a page is reloaded, instead of where the user was at previously. I'm not sure if this is the way we want to go forward but I think it's a potential solution.

packages/vue-components/src/Overlay.vue
jonahtanjz: Yup agree that using this css would be much more succinct solution as I initially wanted to use this as well. However, for some reason, it does not work very well when the nav menus are not scrollable. One example is https://markbind-master.netlify.app/userguide/tweakingthepagestructure, able to scroll the body of page nav when scrolling the ride side of the overlay. This is more obvious on larger screen sizes.

https://user-images.githubusercontent.com/47470981/107728574-6b869180-6d29-11eb-89b1-a4d9990b1b20.mp4

packages/core-web/src/index.js
jonahtanjz: Thanks for investigating this @wxwxwxwx9 👍. Since this issue is only apparent on desktop version of the navbar, if we follow @raysonkoh suggestion of only hiding the header on mobile, then the unwanted scroll event will not be captured. I've tested the mobile version and it works fine as the box is contained within the dropdown so a reload will not trigger a scroll.

jonahtanjz:
@raysonkoh and @wxwxwxwx9 I've updated the PR to include transitions and addressed some of the bugs/issues.

packages/vue-components/src/Overlay.vue
wxwxwxwx9: Ahh, I see! Yup, it's not working on my side as well. I think it's because when nav menus are not scrollable (not enough content), overscroll behaviour does not kick in since there's nothing to scroll. 

In that case, I think your solution of making the body hidden when overlay is opened makes sense, since we don't want them to be able to scroll the body content when they have overlay opened (since the focus should be on overlay). 

For easy reference in the future as to why we add this block, I think we should add some comments to explain the rationale. Maybe something like `document.body.style.overflow = 'hidden'; // to prevent scrolling of the body when overlay is overscrolled`.

packages/core-web/src/index.js
wxwxwxwx9: Alright! That would work as well. :-)

packages/core-web/src/index.js
```diff
@@ -83,7 +83,22 @@ function detectAndApplyFixedHeaderStyles() {
     });
     resizeObserver.observe(headerSelector[0]);
   };
+
+  let lastOffset = 0;
+  const toggleHeaderOnScroll = () => {
+    if (window.innerWidth > 767) { return; }
+    const currentOffset = window.pageYOffset;
+    if ((window.innerHeight + currentOffset) >= document.body.offsetHeight) { return; }
```
wxwxwxwx9, line 91: Hmm I think we can add some comment here as to why this is needed. I wasn't quite sure what this was for until I tested it out. And maybe we can assign it to a variable as well, maybe something like `isEndOfPage`.

raysonkoh: Yup I agree with @wxwxwxwx9 . 

Other than that everything LGTM! Thanks for making the changes @jonahtanjz !

jonahtanjz: Thanks for the suggestion @wxwxwxwx9 and @raysonkoh. Have added the comments :)

packages/vue-components/src/Overlay.vue
ang-zeyu: how about just `@scroll.stop` on the root element of the overlay to prevent the scroll event from bubbling up?

packages/core-web/src/styles/markbind.css
```diff
@@ -118,11 +118,19 @@ code.hljs.inline {
 /* Header */
 
 header[fixed] {
+    max-height: 100%;
+    overflow: hidden;
```
ang-zeyu, line 122: will need to be dynamically adjusted (`visible` once it reaches max-height 100%, `hidden` otherwise) 

searchbar:
Untitled

packages/core-web/src/index.js
```diff
@@ -83,7 +83,25 @@ function detectAndApplyFixedHeaderStyles() {
...
+    if (isEndOfPage) { return; }
+    if (currentOffset > lastOffset) {
+      headerSelector.addClass('hide-header');
+    } else {
+      headerSelector.removeClass('hide-header');
+    }
+    lastOffset = currentOffset;
+  };
+
   addResizeHeaderListener();
```
ang-zeyu, line 103: could we prevent this from running when caused by the changes in `max-height`?

so we don't observe the secondary movements of the content when hiding / showing the navbar

navbar-hiding

ang-zeyu (commented):
Thanks! @jonahtanjz

This mostly looks good. 🙂 
Couple of suggestions / issues:

packages/core-web/src/styles/markbind.css
jonahtanjz: I've managed to fixed this by moving the `overflow: hidden` to a media query for mobile. Seems to be working for both mobile and desktop sites. Is this approach ok instead of looking at the max-height?

jonahtanjz:
> how about just @scroll.stop on the root element of the overlay to prevent the scroll event from bubbling up?

I tried adding this but doesn't seem to be working. I also tried using `e.stopPropagation()` but it's not working as well 😕

packages/core-web/src/styles/markbind.css
ang-zeyu: hmm I guessing it may not work well if the user uses a popover, etc. in the header 

also there's the upper topnav of items in #980 and how that might work with dropdowns (might overflow similarly)

ang-zeyu:
> > how about just @scroll.stop on the root element of the overlay to prevent the scroll event from bubbling up?
> 
> I tried adding this but doesn't seem to be working. I also tried using `e.stopPropagation()` but it's not working as well 😕

oops, looks like I misunderstood the issue here (scrolling, not the event is bubbling up), let's leave this as is then!

packages/core-web/src/styles/markbind.css
jonahtanjz: Alright got it 👍. Updated `overflow` to be dynamic such that it will only be visible if `max-height` is 100%.

packages/core-web/src/index.js
jonahtanjz: Alright updated :)

packages/core-web/src/index.js
```diff
@@ -52,38 +52,76 @@ function detectAndApplyFixedHeaderStyles() {
...
-          default:
-            break;
-          }
-        }
+      const headerMaxHeight = headerSelector.css('max-height');
+      // set overflow to hidden if max-height is not 100%
+      if (headerMaxHeight !== '100%') {
+        headerSelector.css('overflow', 'hidden');
+        // re-check after 0.6s to account for transition time
+        setTimeout(toggleHeaderOverflow, 600);
```
ang-zeyu, line 99: Generally not a good idea to use `setTimeout` unless a delay / timer is explicitly needed, as the timing of events (here being the transition finishing and resize handler) is not guaranteed. (especially across browsers / devices)

You could take a look at the https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event (and run) here instead to detect whether it is transitioning / finished transitioning

jonahtanjz: Thanks for the suggestion @ang-zeyu :) Was not aware of this transition event. This makes it much better. Have removed the `setTimeout` and replaced it with `transitionend` listener.

packages/core-web/src/index.js
```diff
@@ -52,38 +52,75 @@ function detectAndApplyFixedHeaderStyles() {
       }`);
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
 
+  const resizeHeader = () => {
```
ang-zeyu, line 55:
```suggestion
  const adjustHeaderClasses = () => {
```

since the header itself isn't resized =P

packages/core-web/src/index.js
```diff
@@ -52,38 +52,75 @@ function detectAndApplyFixedHeaderStyles() {
...
+
+  const toggleHeaderOverflow = () => {
+    const headerMaxHeight = headerSelector.css('max-height');
+    if (headerMaxHeight === '100%') {
+      headerSelector.css('overflow', '');
+      resizeHeader();
+    }
+  };
+
   const addResizeHeaderListener = () => {
```
ang-zeyu, line 92: should we remove the wrapping method? seems a little verbose now that you've extracted a huge chunk out

packages/core-web/src/index.js
```diff
@@ -52,38 +52,75 @@ function detectAndApplyFixedHeaderStyles() {
...
-            break;
-          case '.nav-menu-open':
-            rules[0].style.maxHeight = `calc(100% - ${newHeaderHeight}px + 50px)`;
-            break;
-          default:
-            break;
-          }
-        }
+      const headerMaxHeight = headerSelector.css('max-height');
+      // set overflow to hidden if max-height is not 100%
```
ang-zeyu, line 95: let's explain the why here instead

packages/core-web/src/index.js
```diff
@@ -52,38 +52,75 @@ function detectAndApplyFixedHeaderStyles() {
...
+        default:
+          break;
+        }
+      }
+    }
+  };
+
+  const toggleHeaderOverflow = () => {
+    const headerMaxHeight = headerSelector.css('max-height');
+    if (headerMaxHeight === '100%') {
```
ang-zeyu, line 86: a small why (we need to reset overflow when the header shows again) comment here would help too

ang-zeyu (commented):
Thanks! Last few nits:

jonahtanjz:
@ang-zeyu Updated the PR :)

ang-zeyu (approved):
Lgtm 👍

PR #1478, opened 2021-02-12
Code blocks: Support partial text highlighting

### LINKED ISSUE #1381: Fenced code blocks: support highlighting words
Current: we support highlighting specific lines in a code block

Suggestion: extend that feature to support highlighting specific parts of a line

ryoarmanda:

Resolves #1381 

**Overview of changes:**
Partial text highlighting is available with the line-slice syntax format `lineNumber[start:end]`. You can see this in action in the deployed UG [here](https://deploy-preview-1478--markbind-master.netlify.app/userguide/fullsyntaxreference#code).

The processing is done in NodeProcessor in order to efficiently leverage the parsed HTML to traverse the code block node. For each line, the highlighting range is carried over by `hl-start` and `hl-end` attributes, specified from the `markdown-it` patch, which will be extracted in NodeProcessor.

In short, the traversal strategy is as follows:
- If the node is a "text" node, determine whether some/all/none of the text should be highlighted. As this node only exists as a child of another node, inform the node's parent on the decision via return value.
- If the node is a "tag" node, recursively call the traversal to each child, and collate the highlight data from the return value as above.
  - If all children want the node to highlight, we can just add the `highlighted` class in the node for conciseness
  - If not, handle the highlight for each child accordingly, which can include transforming "text" nodes into "tag" nodes so highlighting can be applied to them.

With the addition of partial text highlighting, better support for range highlighting is achievable. Users can specify whether to
start/end the range with partial text highlight. Range highlight processing is modified to support this.

- Any comments for improvements on the general approach is appreciated. Particularly, the DOM manipulation specifics (such as setting the references) might need a bit more set of eyes, I don't know if I missed anything.
- About range highlighting: In the past we have agreed that the default (single-line) highlight style should be full-text, but we haven't really discussed on the ranged version. Should the default highlight for the ones in-the-middle of the range be full-text or full-line? Currently it's full-text (and at the moment, full-line can only be invoked when empty line-slices are present), and I'm open for suggestions.
- Edit: About indices convention: should it be 0-based or 1-based? As the slicing is akin to python's slice, it's natural to start with 0 as users may be accustomed to do it in that way. However, I also feel that it's more natural to start counting the characters in the line from 1.

**Testing instructions:**

Code blocks: Support partial text highlighting

The code blocks line highlighting functionality only supports full-text
and full-line highlights. Authors who wish to only highlight certain
words or key variables are not able to do so with the supported syntax.

Let's add a new highlight syntax to provide authors with the ability to
partially highlight text in a line.

---


packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
```diff
@@ -46,6 +46,28 @@ class HighlightRuleComponent {
...
+   * @returns {[number, number]} The actual bounds computed
+   */
+  computeLineBounds(line) {
+    const [lineStart, lineEnd] = [0, line.length - 1];
+    if (!this.isSlice || this.isUnboundedSlice()) {
+      return [lineStart, lineEnd];
+    }
+
+    const [boundStart, boundEnd] = this.bounds;
+    const start = lineStart <= boundStart && boundStart <= lineEnd ? boundStart : lineStart;
```
raysonkoh, line 67: Should we put a parenthesis around the two conditions for readability?

packages/core/src/html/NodeProcessor.js
```diff
@@ -176,6 +176,142 @@ class NodeProcessor {
...
+    const shouldHighlight = node.children.map((child) => {
+      const [traversed, data] = this._traverseLinePart(
+        child, hlStart - curr, hlEnd - curr,
+      );
+      curr += traversed;
+
+      return data;
+    });
+
+    if (shouldHighlight.every(v => v === true)) {
```
raysonkoh, line 265: Would it be better to have `.every(v => v)` if `shouldHighlight` is an array of booleans?

packages/core/src/html/NodeProcessor.js
```diff
@@ -176,6 +176,142 @@ class NodeProcessor {
...
+      return [true, curr];
+    }
+
+    // If node level highlighting is not possible, highlight the individual children as needed
+    // For tag nodes, it is easy, just add the highlight class
+    // For text nodes, it is trickier, as we have to wrap the text inside a <span> first
+    // Essentially, we have to change the text node to become a tag node
+
+    node.children.forEach((child, idx) => {
+      if (shouldHighlight[idx] === false) {
```
raysonkoh, line 277: Would it be better to have `!shouldHighlight[idx]` if `shouldHighlight` is an array of booleans?

packages/core/src/html/NodeProcessor.js
```diff
@@ -176,6 +176,142 @@ class NodeProcessor {
...
+
+      if (child.type === 'tag') {
+        child.attribs.class = child.attribs.class ? `${child.attribs.class} highlighted` : 'highlighted';
+        return;
+      }
+
+      const text = child.data;
+      let newElement;
+
+      if (shouldHighlight[idx] === true) {
```
raysonkoh, line 289: This might be similar to the issue above.

raysonkoh (changes requested):
Nice work at getting this to work @ryoarmanda !

For the proposed syntax `lineNumber[start:end]`, would it be better to have `start` and `end` refer to the *index of a word in a line*? I think this would reduce the work on the part of the user as they would not need to manually count the character indices. 

What are your thoughts about this?

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
ryoarmanda: Sure, no problem

packages/core/src/html/NodeProcessor.js
ryoarmanda: Not all items in `shouldHighlight` array is boolean. It is just an array to collate the return value as described in the beginning of the function, which may be `true`, `false`, or an array of two numbers. I explicitly uses `=== true` and `=== false` to not include the array of two numbers (which is truthy in itself).

In hindsight maybe I should call this just `highlightData`

ryoarmanda:
> For the proposed syntax `lineNumber[start:end]`, would it be better to have `start` and `end` refer to the _index of a word in a line_? I think this would reduce the work on the part of the user as they would not need to manually count the character indices.

Hmm I was working under the spec on the original discussion [here](https://github.com/MarkBind/markbind/issues/1311#issuecomment-667683495). My two cents is, if the unit of highlighting is words rather than characters, we would have to do some extra work in determining which of these characters can be grouped up as words. Moreover, with how code is generally written, sometimes users may want to highlight beyond words, such as words continued with special characters, etc, and then the distinction of words becomes blurry.

damithc:
> For the proposed syntax `lineNumber[start:end]`, would it be better to have `start` and `end` refer to the _index of a word in a line_? I think this would reduce the work on the part of the user as they would not need to manually count the character indices.

Having to count characters is not ideal, but I guess we can anticipate users will do that in two passes i.e., uses a rough count in the first try and tweak the numbers based on the result.
We can consider supporting an additional format where the user specifies the exact text to highlight e.g., `12['void main']` but that can be a separate PR, if at all?

ryoarmanda:
> We can consider supporting an additional format where the user specifies the exact text to highlight e.g., `12['void main']` but that can be a separate PR, if at all?

This can be a good addition as the easier version of the line-slice syntax, provided the section to be highlighted is not that long. The line-slice syntax can then be a more advanced-user variant where users can express highlights in a more concise way. In fact, we can build up this easier syntax on top of the line-slice one.

But to (user-side) writing and (dev-side) parsing the rule can be quite tricky, there might be quotes present in the text to be highlighted, in which the user may have to manually escape this in order for the whole `highlight-lines` to keep properly parsed, and MarkBind needs to match and unescape this as well.

May need to put some proper thought into this, I might address it in a separate PR for now.

damithc:
> But to (user-side) writing and (dev-side) parsing the rule can be quite tricky, there might be quotes present in the text to be highlighted, in which the user may have to manually escape this in order for the whole `highlight-lines` to keep properly parsed, and MarkBind needs to match and unescape this as well.

True. But given it is an additional convenience feature only, we can specify under what conditions the simpler syntax should be used. In all other cases, the user should use the slice syntax instead. Hence, we can avoid dealing with those complex corner cases.

ryoarmanda:
> True. But given it is an additional convenience feature only, we can specify under what conditions the simpler syntax should be used. In all other cases, the user should use the slice syntax instead. Hence, we can avoid dealing with those complex corner cases.

Just now I found out a way to support the convenience syntax on top of the line-slice processing (behind the scenes, the former rule will be converted to the latter, before handing over to NodeProcessor). I might push it here after docs and tests updates.

But yeah, looks like we have a hard restriction, for some reason we can't have the same type of quotes that specifies the whole `highlight-lines` attribute in whatever inside the `[]`s, even with backslash-escaping. I'm not sure where the main issue lies here, maybe with `markdown-it-attrs` and/or the patch of it.

So, something with `highlight-lines="4['It\"s designed']"` can't be done, but `highlight-lines="4['It\'s designed']'` can (and interestingly: `highlight-lines="4['It's designed']'` can also be parsed properly even if there are no backslash-escapes)

ang-zeyu:
> So, something with `highlight-lines="4['It\"s designed']"` can't be done, as is `highlight-lines="4['It"s designed']"`.
> But, `highlight-lines="4['It\'s designed']"` can (and interestingly unexpected: `highlight-lines="4['It's designed']"` is parsed properly even if there are no backslash-escapes)

How about sticking to positioning? we could introduce something like `4[1::3]` (`::` for words instead of `:`)

> Just now I found out a way to support the convenience syntax on top of the line-slice processing (behind the scenes, the former rule will be converted to the latter, before handing over to NodeProcessor). I might push it here after docs and tests updates.

but you can try removing the patch now as well as its no longer needed with #1403 

it was introduced as we previously had multiple nunjucks passes to make `{% raw %}` work, and one markdown pass in between, causing issues with `{% raw %}` and markdown-it-attrs https://github.com/MarkBind/markbind/pull/1220

ryoarmanda:
> How about sticking to positioning? we could introduce something like `4[1::3]` (`::` for words instead of `:`)

Perhaps we can explore this syntax as well. I might define a word as just a sequence of non-whitespace characters in order to include special characters (e.g. `List<Item>` is considered one word). Otherwise, discerning words can be tricky.

However, specific portions of words text might not be able to be highlighted which I guess is to be expected as word-level highlight is inherently less fine than character-level.

For example, if a user wants to highlight only the `Item` part of `private List<Item> items`, they can only do so with the character-level highlight.

With this syntax, do you think we should keep the previously proposed one?

ang-zeyu:
> With this syntax, do you think we should keep the previously proposed one?

I'm open to either, specifying the words is more flexible, but also more repititive and verbose.
Specifying positions seems more consistent with character highlight and less verbose, but may be more restrictive as you mentioned, then users would have to fallback to character highlight.

On a personal stance I would go with positions however, for consistency with the whole line-slice syntax. Also, since repitition is against the trend here =P.

Any thoughts? @damithc @raysonkoh 

> So, something with highlight-lines="4['It\"s designed']" can't be done, as is highlight-lines="4['It"s designed']".
> But, highlight-lines="4['It\'s designed']" can (and interestingly unexpected: highlight-lines="4['It's designed']" is parsed properly even if there are no backslash-escapes)

Does `&quot;` work?

backslash escapes `\"` are only a [commonmark spec](https://spec.commonmark.org/0.29/#backslash-escapes), I suppose markdown-it-attrs is free to disregard this =X

damithc:
> On a personal stance I would go with positions however, for consistency with the whole line-slice syntax. Also, since repitition is against the trend here =P.
> 
> Any thoughts? @damithc @raysonkoh

@ryoarmanda what are the choices being considered here? I haven't been following the previous discussion.

ryoarmanda:
> What are the choices being considered here? I haven't been following the previous discussion.

To sum up the discussion, we have two proposed syntax for word-highlighting.

One is the line-part syntax `lineNumber[part]` (e.g. `4['void main']`)
- Pros:
  - Straightforward usage on the user-side (if they want to highlight `void main`, just write `void main` within the bracket)
  - Retains the fine granularity of character-highlighting (if they wish, user can highlight `Item` from `List<Item>` just fine)
- Limitations:
  - Can become quite repetitive and verbose as users have to repeat the content to be highlighted
  - Somehow we can't use the type of quote that is exactly the same as the one used for specifying the `highlight-lines` value, even with escaping (still under investigation)

Another is the word variant of line-slice syntax `lineNumber[start::end]` (note the double `:`s, e.g. `4[1:3]`)
- Pros:
  - Similar to the character variant of line-slice syntax, so user does not need to remember an entirely different syntax.
  - More concise
- Limitations:
  - Highlight is word-level, so granularity is ultimately less fine than character highlight (using the previous example, highlighting `Item` from `List<Item>` cannot be achieved with this syntax, and have to use character highlight as fallback)

With their own pros and limitations, which of the two do you prefer to be supported? Or should we just support both at once?

damithc:
> To sum up the discussion, we have two proposed syntax for word-highlighting.

Why not support both but let `lineNumber[start::end]` specify character positions instead of word positions?
Or `lineNumber[start:end]` for character positions and `lineNumber[start::end]` for word positions?

IIRC, we used `[ : ]` syntax earlier with the plan to extend it for more fine grain control. So, shouldn't we continue with the single colon syntax? 

Sorry if these were discussed before.

raysonkoh:
I think we can support `lineNumber[start:end]` for character positions and `lineNumber[start::end]` for word positions for a start.

> Somehow we can't use the type of quote that is exactly the same as the one used for specifying the highlight-lines value, even with escaping (still under investigation). So, this limits the syntax content to be those who don't have quotes.

For `lineNumber[part]`, this would be more of a convenient syntax (i.e. anything that can be achieved by `lineNumber[part]` can be achieved by `lineNumber[start:end]`, but user types less)  but as you said, more investigation is needed for some edge case. I think we can raise a separate PR for that.

raysonkoh:
Some additional concerns about the expected behavior of `lineNumber[part]`:
1. If there are duplicates `part`s in the line, do we highlight all `part`s or only the first occurrence?
2. If `part` is a substring of another word in the same line, would it get highlighted too?

damithc:
> Some additional concerns about the expected behavior of `lineNumber[part]`:
> 
> 1. If there are duplicates `part`s in the line, do we highlight all `part`s or only the first occurrence?
> 2. If `part` is a substring of another word in the same line, would it get highlighted too?

My vote:
1. all
2. yes

ang-zeyu:
> IIRC, we used [ : ] syntax earlier with the plan to extend it for more fine grain control. So, shouldn't we continue with the single colon syntax?

yup

> Or lineNumber[start:end] for character positions and lineNumber[start::end] for word positions?

I think we should stick with this (versus single colon for words). Its consistent with the existing `1[:]` (meaning 'first' to 'last' character) whole line syntax.

> Why not support both but let lineNumber[start::end] specify character positions instead of word positions?


I'm open to supporting both as well (although it is a niche feature)

@ryoarmanda 

If possible, this PR could be split into 4 commits (or prs) as well:
- character positions
- word positions (or can handle in another PR too)
- `part` syntax
- docs

I'll be doing a couple things shortly as well to make the reviewing easier:
- some reorganization of the markdown-it directory (the files here untouched)
- then enabling eslint appropriately as mentioned previously (hopefully not too many conflicts, do eslint the modifications before merging in to reduce the conflicts)

ang-zeyu:
> then enabling eslint appropriately as mentioned previously (hopefully not too many conflicts, do eslint the modifications before merging in to reduce the conflicts)

All done. Sorry, this should really have been done beforehand. 😓

ryoarmanda:
> If possible, this PR could be split into 4 commits (or prs) as well:
> 
> * character positions
> * word positions (if you're up for it) (or can handle in another PR too - preferably here for 'chronological' order)
> * `part` syntax
> * docs

Currently in this PR there are already character positions and `part` syntax (although the part syntax does not yet support highlighting multiple occurrences as proposed by @damithc , will try to figure out more). I will develop the word positions syntax  here too.

I am not sure if this is too much introduction of new features in one PR, should I break away the `part` syntax to another PR, or would it be okay to keep it here as well?

ang-zeyu:
> I am not sure if this is too much introduction of new features in one PR, should I break away the `part` syntax to another PR, or would it be fine to keep it here as well?

Both sounds fine. It all ties under one logical change (partial text highlighting), but is a little large as you mentioned.

Let's split up the commits if sticking to the PR though. You can also use [fixups](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) (or just rebasing in your git ui) to address reviews later in new, separate commits, then squash them with the relevant commit once done)

packages/core/src/html/NodeProcessor.js
```diff
@@ -176,6 +176,142 @@ class NodeProcessor {
     cheerio(node).remove();
   }
 
+  /*
```
ang-zeyu, line 179: Let's move these to a separate file for a start

packages/core/src/utils/index.js
```diff
@@ -9,6 +9,14 @@ const {
   markdownFileExts,
 } = require('../constants');
 
+const htmlUnescapedMapping = {
```
ang-zeyu, line 12: how about something like https://www.npmjs.com/package/he?

might become a little tedious to maintain the list

ryoarmanda: This mapping is taken from `markdown-it` library, just reversed. I did this as in our codebase we use `markdown-it/utils` function `escapeHtml` for a lot of purposes, but that package didn't expose unescaping methods for HTML, so I took it upon myself to add a complementary function. Looks like it's going to be only that 5 entries though, as the fact that this entry only aims to reverse the `escapeHtml` function and that function only escapes those 5 entries.

Though if you feel we can add a new package to make everything concise I'm also okay with it.

ang-zeyu: ahh, thanks for clarifying!

In that case let's stick with this then, we might not want to unescape things markdown-it did not escape as well.

On the same lines, should we put this somewhere in the lib/markdown-it folder? Also a comment to document why / where the mappings came from may help 🙂

ryoarmanda: :+1: I agree. This is more related to operations involving markdown-it as well so it makes sense to put it in the lib/markdown-it folder. Though I can't actually expose the new function through the `markdownIt` object in `lib/markdown-it/index.js`, and I can't disturb the exports there unless I replace all imports in the project, so I will just make it available for manual import through `lib/markdown-it/utils`.

It's the best middle ground I have, but I'm afraid it will somewhat be confusing as now we have two different modules for utilities on markdown-it, one from the library itself usually referred to as `md.utils`, the other is the new `utils` module. I might put an explanation on the top of our `utils` module that this is a separate extension of the one from the library.

docs/userGuide/syntax/code.mbdf
```diff
@@ -58,42 +58,100 @@ function add(a, b) {
...
+You can add the `highlight-lines` attribute to add highlighting to your code block. Refer to the example below for
+all possible ways of highlighting a code block.
 
-**Text-only highlighting**
+The value of `highlight-lines` is composed of *highlight rules*, separated by commas.
+These rules dictate where and how MarkBind should highlight your code block.
 
-To highlight only the text portion of the line, you can just use the line numbers as is.
+You can specify the rules in many different ways, depending on how you want it to be. There are three main variants:
+full text, substring, bounded (character-wise or word-wise), or full line highlighting.
```
ang-zeyu, line 69: seems a little lengthy; Could we compress all this into a table? 🤔

ang-zeyu (commented):
Nice work! 🙂

Just 1 / 2 edges cases and some minor suggestions for the code:

packages/cli/test/functional/test_site/testCodeBlocks.md
```diff
@@ -56,8 +56,8 @@ Content in a fenced code block
 20
 ```
 
-**`highlight-lines` attr with line-slice syntax of empty indices should highlight leading/trailing spaces
-```xml {highlight-lines="2[:],4[:]-5[:]"}
+**`highlight-lines` attr with empty (any variant) line-slice syntax should highlight leading/trailing spaces**
```
ang-zeyu, line 59: let's add in the edge cases (if they are) earlier once fixed as well

packages/core/src/lib/markdown-it/index.js
```diff
@@ -58,7 +58,38 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+  const highlightLinesInput = getAttributeAndDelete(token, 'highlight-lines');
+  let highlightRules = [];
+  if (highlightLinesInput) {
+    const highlightLines = highlightLinesInput.split(',');
+    highlightRules = highlightLines.map(HighlightRule.parseRule).filter(rule => rule);
+  }
+  highlightRules.forEach((rule) => {
+    // Note: authors provide line numbers based on the 'start-from' attribute if it exists,
+    //       so we need to shift line numbers back down to start at 0
+    rule.offsetLines(-startFromZeroBased);
```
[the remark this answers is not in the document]
ryoarmanda: Oh yeah, I hadn't thought of this approach. I can take it one step further so that the result of `parseRuleComponent` is always a properly defined rule component wrt the code block (i.e. already accounting for line offset, figuring out the actual bounds wrt the line, converting to word slices wrt the line, etc), or null if it's not valid.

So, no need to have intermediate properties such as `isWordSlice`, `linePart`, anymore.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
```diff
@@ -46,6 +86,88 @@ class HighlightRuleComponent {
...
+    this.linePart = '';
+  }
+
+  convertWordSliceToCharSlice(content) {
+    if (!this.isSlice || !this.isWordSlice) {
+      return;
+    }
+
+    const [contentStart, contentEnd] = [0, content.length];
+    const words = content.split(' ');
```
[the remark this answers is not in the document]
ryoarmanda: Yup sure. I think to account for consecutive spaces I'll have to do `.split(/\s+/)` so as to not have empty strings in the split result. But that means the additional count for whitespace in the next few line should not be always 1. I'll think on how to rework this part.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js
```diff
@@ -46,6 +86,88 @@ class HighlightRuleComponent {
...
+    let start = contentRemaining.indexOf(this.linePart);
+    const bounds = [];
+
+    if (start === -1) {
+      bounds.push([0, 0]);
+    }
+
+    let curr = 0;
+    while (start !== -1) {
+      const end = start + this.linePart.length;
```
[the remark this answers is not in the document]
ryoarmanda: Whoops, nice catch. I'll add another check on whether the line part is an empty string

packages/core/src/lib/markdown-it/index.js
```diff
@@ -58,7 +58,38 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+
+  if (startFromOneBased > 1) {
+    // counter is incremented on each span, so we need to subtract 1
+    token.attrJoin('style', `counter-reset: line ${startFromZeroBased};`);
+  }
+
+  const highlightLinesInput = getAttributeAndDelete(token, 'highlight-lines');
+  let highlightRules = [];
+  if (highlightLinesInput) {
+    const highlightLines = highlightLinesInput.split(',');
```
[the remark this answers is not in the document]
ryoarmanda: Yeah, this will also eat up the commas inside the brackets. I'll add a regex for a more robust split so that it can ignore the inner commas.

packages/core/src/html/codeblockProcessor.js
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Applies pending highlighting to the code block.
+ * This looks into each line for highlighting data, and if found,
+ * traverses over the line and applies the highlight.
+ * @param node Root of the code block element, which is the 'pre' node
+ * @return
+ */
+function highlightCodeBlock(node) {
+  const codeNode = node.children.find(c => c.name === 'code');
```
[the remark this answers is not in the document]
ryoarmanda: Wouldn't `codeNode` be more proper? This line is looking for the node that represents the `<code></code>` element. `preEl` seems like talking about the `<pre></pre>` node, or did I misunderstood the naming 😅

packages/core/src/html/codeblockProcessor.js
```diff
@@ -0,0 +1,143 @@
...
+        `<span>${pre}<span class="highlighted">${highlighted}</span>${post}</span>`,
+      );
+    }
+
+    delete newElement.root;
+    node.children[idx] = newElement;
+  });
+
+  // Set the references accordingly
+  node.children.forEach((child, idx) => {
```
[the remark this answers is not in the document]
ryoarmanda: I think this certainly is more concise if the text is undisturbed (like the case when `highlightData[idx] === true` i.e. highlight spans the whole text), but so far I wasn't able to work a similar solution for the `else` case (highlight only spans partially). The problem is that the case needs to break up the text, creating multiple elements/nodes in the process before being wrapped up by a span. I tried out something like `cheerio(child).wrap('<span></span>'); cheerio(child).html(...)` and it doesn't work :/

docs/userGuide/syntax/code.mbdf
ryoarmanda: Yeah I agree :/ Felt like a lot to explain from the various types, syntax formats, and its limitations. But not sure how far I can condense it to a table, looks like some parts may need to just have a brief sentence or two without making the table feel like a wall of text. I'll try to do it soon.

packages/core/src/html/codeblockProcessor.js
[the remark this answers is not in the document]
ryoarmanda: Update: I found out a way to make it work with cheerio, will implement this approach instead of manually assigning the references.

ryoarmanda:
Hi @ang-zeyu, I have reworked the processing flow for the highlight rules, now the major work is done during parsing the component rule (including offsetting line numbers, computing actual bounds, etc) so it is ensured that after parsing, the rule can be applied immediately to the code block without further processing. Can you help review whether the flow is streamlined enough? Thanks!

packages/cli/test/functional/test_site/testCodeBlocks.md
```diff
@@ -66,6 +66,56 @@ Content in a fenced code block
...
+```xml {highlight-lines="1[0::1],2[3::4],3[0::2],4[2::4],5[1::3]"}
+<foo>
+  <bar type="name"> goo </bar>
+  <baz type="name"> goo </baz>
+  <qux type="name"> goo </qux>
+  <quux type="name"> goo </quux>
+</foo>
+```
+
+**`highlight-lines` attr with partial word-variant line-slice syntax should defaults highlight to start/end of line**
```
ang-zeyu, line 109:
```suggestion
**`highlight-lines` attr with partial word-variant line-slice syntax should default highlight to start/end of line**
```

docs/userGuide/syntax/code.mbdf
```diff
@@ -58,42 +58,63 @@ function add(a, b) {
...
-
-For ranges, you only need to use line-slices on either ends.
+Type | Format | Example
+-----|--------|--------
+**Ranged full text highlight**<br>Highlights only the text portion of the lines within the range | `lineStart-lineEnd` | `2-4`
+**Ranged full line highlight**<br>Highlights the entirety of the lines within the range | `lineStart[:]-lineEnd` or `lineStart-lineEnd[:]` | `1[:]-5`,`10-12[:]`
+**Ranged character-bounded highlight**<br>Highlights the text portion of the lines within the range, but starts/ends at an arbitrary character | `lineStart[start:]-lineEnd` or `lineStart-lineEnd[:end]` | `3[2:]-7`, `4-9[:17]`
+**Ranged word-bounded highlight**<br>Highlights the text portion of the lines within the range, but starts/ends at an arbitrary word | `lineStart[start::]-lineEnd` or `lineStart-lineEnd[::end]` | `16[1::]-20`,`22-24[::3]`
 
 <include src="codeAndOutputCode.md" boilerplate >
```
ang-zeyu, line 88: let's shift this up before `the value of highlight-lines...` so the user has a brief idea what the usage is like first :-)

docs/userGuide/syntax/code.mbdf
```diff
@@ -58,42 +58,63 @@ function add(a, b) {
...
+for range highlighting consists of two single line highlight rules as listed above joined by a dash (`-`).
 
-If you wish to highlight a full line (including whitespaces) or ranges of it, you can leverage MarkBind's own _line-slice_ syntax. Line-slices are in the form of `lineNumber[:]`, e.g. `2[:]`.
-    
-This variant's format is very similar to the previous, but instead use line-slices rather than line numbers.
-
-For ranges, you only need to use line-slices on either ends.
+Type | Format | Example
+-----|--------|--------
+**Ranged full text highlight**<br>Highlights only the text portion of the lines within the range | `lineStart-lineEnd` | `2-4`
```
ang-zeyu, line 83:
```suggestion
**Ranged full text highlight**<br>Highlights from the first non-whitespace character to the last non-whitespace character | `lineStart-lineEnd` | `2-4`
```

interpretation of `text portion` is subjective (e.g. in-between whitespaces should / should not be highlighted) =P

docs/userGuide/syntax/code.mbdf
```diff
@@ -58,42 +58,63 @@ function add(a, b) {
...
+**Ranged full line highlight**<br>Highlights the entirety of the lines within the range | `lineStart[:]-lineEnd` or `lineStart-lineEnd[:]` | `1[:]-5`,`10-12[:]`
+**Ranged character-bounded highlight**<br>Highlights the text portion of the lines within the range, but starts/ends at an arbitrary character | `lineStart[start:]-lineEnd` or `lineStart-lineEnd[:end]` | `3[2:]-7`, `4-9[:17]`
```
ang-zeyu, line 85: `Like ranged full text highlight, but...`
the word one too

ang-zeyu (commented):
Really excellent work, especially on the test cases and docs.

Just a few more nits:

docs/userGuide/syntax/code.mbdf
ang-zeyu: Looks really neat now 👍 thanks!

packages/core/src/html/codeblockProcessor.js
```diff
@@ -0,0 +1,136 @@
...
+  /*
+   * The remaining possibility is that node is a tag node.
+   * It has at least one child (to contain the text content).
+   * It may have more children, such as inner tag nodes.
+   */
+
+  let curr = 0;
+  const highlightData = node.children.map((child) => {
+    const data = traverseLinePart(child, hlStart - curr, hlEnd - curr);
+    curr += data.numCharsTraversed;
```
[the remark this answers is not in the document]
ryoarmanda: True, but replacing `curr` with `resData.numCharsTraversed` might end up being more verbose (especially at the `data` constant declaration there). But I'll try to remove `curr` and rework the `hlStart - curr` and `hlEnd - curr` to be computed before the recursive call.

docs/userGuide/syntax/code.mbdf
ryoarmanda: Ah yes, will do

ryoarmanda:
The docs has been updated to address the comments, can have a look again and see if there is anything I missed out. Thanks for all the reviews so far @ang-zeyu 🙇

docs/userGuide/syntax/code.mbdf
ang-zeyu: might have missed one ^ @ryoarmanda 
disregard if it looks stranger

ryoarmanda: I did this already @ang-zeyu, I just reworded `the value of highlight-lines...` to the ones in line `85-86`

ang-zeyu: > this

was suggesting the reverse - moving the example up (the `<include>` tag), so the user gets a brief overview of the entire syntax usage first before the details

ryoarmanda: Ahh okay, will rebase the fix shortly

ryoarmanda: Done, the change is reverted in the force push :)

ang-zeyu (approved):
Lgtm 👍

damithc:
@ryoarmanda try to avoid 'all possible usage in one example' technique. It might work for test cases but not a good option in user documentation. From user POV, the example is too complicated and unrealistic. Can you do a minor PR to split the example into separate small examples?

damithc:
BTW, kudos for getting this difficult feature to work @ryoarmanda 💯

PR #1462, opened 2021-01-28
Enable linting for dropdown.vue

jonahtanjz:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [ ] Feature addition or enhancement
- [x] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
This is a follow-up PR to #1455. Linting for `dropdown.vue` is enabled and code is refactored to be consistent with the lint rules. 

**Anything you'd like to highlight / discuss:**
Previously, there is a `class` which is set to `null` in the props. Not sure what is the function of it as it seems to work fine after removal. If it is needed, will proceed to add it back. 

**Testing instructions:**
`npm run test`

**Proposed commit message: (wrap lines at 72 characters)**
Enable linting for dropdown.vue

Currently, stylelint is only enabled for css files. Vue components may
also contain <style> tags with valid css but these are not linted. 

Let's enable stylelint css for vue files to provide linting to them as well.

---

**Checklist:** :ballot_box_with_check:

- [x] Updated the documentation for feature additions and enhancements
- [x] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No blatantly unrelated changes
- [x] Pinged someone for a review!

PR #1482, opened 2021-02-18
Eslint and stylelint all .vue files

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
A follow-up PR to #1462. Enable eslint and stylelint for all `.vue` files in `vue-components`.

1. There are a few `console.log`s used which eslint will throw a warning. Should I disable the checks for these `console.log`s or remove them?
2. Use of `v-html` in some `.vue` files is also throwing warnings during the check. Should this be change from `<span v-html="html"></span>`  to `<span>{{ html }}</span>` instead to follow eslint rules? 

**Testing instructions:**
`npm run test`

Eslint and stylelint all .vue files

Currently, most .vue files are ignored during the linting checks.

Since these files are frequently modified, we should validate 
them in our ci tests to ensure they follow eslint rules. Furthermore, 
.vue files can contain raw css hence appropriate to also stylelint them.

Let's eslint and stylelint all .vue files so that they can be validated 
as well. 

---


packages/vue-components/src/Tab.vue
```diff
@@ -13,81 +16,82 @@
...
-      this.$parent.tabs.push(this)
+      this.$parent.tabs.push(this);
     }
   },
-  beforeDestroy () {
-    if (this._tabset.active === this.index) { this._tabset.active = 0 }
+  beforeDestroy() {
+    if (this._tabset.active === this.index) { this._tabset.active = 0; }
     if (this._ingroup) {
-      var index = parent.tabs.indexOf(this);
```
ang-zeyu, line 75: 👀  should this have been referencing `$parent`?

jonahtanjz: Yup seems like it should have been `$parent` as the function `parent.tabs.indexOf` does not exist.

ang-zeyu (approved):
Lgtm!

> There are a few console.logs used which eslint will throw a warning. Should I disable the checks for these console.logs or remove them?
> Use of v-html in some .vue files is also throwing warnings during the check. Should this be change from <span v-html="html"></span> to <span>{{ html }}</span> instead to follow eslint rules?

Thanks for the catch, I think we can resolve the second issue separately since it requires a substantial functional change.

For the console.logs, might be better to remove them as we shouldn't be relying on it to inform users of anything. Where appropriate, could replace them with comments for anyone taking a look at these files in the future.

jonahtanjz:
@ang-zeyu Updated the PR to remove the consoles or replaced them with comments.

ang-zeyu (approved):

PR #1483, opened 2021-02-19
Remove shorthand slot syntax from documentation

### LINKED ISSUE #1078: Allow vue named slots shorthand and start deprecating old syntax

wxwxwxwx9:

- [x] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

This PR is a follow-up to #1464.

**Overview of changes:**
~~Changed all the v-slot old syntax to the latest shorthand syntax.~~

We have moved this PR to a new direction as per the discussion below. We will not support the `#example` shorthand syntax notation for slot (due to CommonMark spec), but we will explore the possibility of a new shorthand in a new issue/PR. Instead, we will remove the shorthand slot syntax from our documentation for this PR. 

~~Whether there is anything that should not be changed. Also, I updated the Vue slot syntax for `PageNavButton` and `SiteNavButton`, should it remain the same?~~

We have discussed issues with this PR below. 

**Testing instructions:**
`npm run test`

Remove shorthand slot syntax from documentation. 

Using `#` in attribute name is illegal according to the CommonMark
specification (v0.29). To comply with the specification, we should not
support the intended shorthand slot syntax.

Let's remove the shorthand slot syntax from our user guide.

---

### DIFF
--- a/docs/userGuide/components/advanced.md
+++ b/docs/userGuide/components/advanced.md
@@ -11,14 +11,14 @@
 Most component attributes allow a richer form of formatting using slots, denoted by an attribute<strong>^\[S\]^</strong> superscript in the respective components' tables.
 In other cases, when the option is of type "Slot", only the slot option is available.
 
-You can define such a slot within the component by adding a `#slot_name` attribute to any element within the slot.
+You can define such a slot within the component by adding a `slot="slot_name"` attribute to any element within the slot.
 
 {{ icon_example }}
 
 <include src="codeAndOutput.md" boilerplate>
 <variable name="code">
 <panel expanded>
-  <p #header class="card-title">
+  <p slot="header" class="card-title">
     <i><strong>
       <span style="color:#FF0000;">R</span>
       <span style="color:#FF7F00;">A</span>
@@ -36,21 +36,6 @@ You can define such a slot within the component by adding a `#slot_name` attribu
 <variable name="highlightStyle">html</variable>
 </include>
 </div>
-
-<box type="info">
-
-You may define a slot by using `slot="slot_name"` as well. 
-However, we **$$do not recommend$$** using this syntax as it is deprecated and will be removed in the future. 
-
-```html
-  <panel>
-    <div slot="header">
-      Look at the slot syntax directly above. 
-    </div>
-  </panel>
-```
-</box>
-
   
 <modal header="Richer formatting of attributes using slots" id="on-slots" large>
 <include src="advanced.md#slots" />


ryoarmanda:
Nice work :) I took a slight check on the changes and looks like it's pretty straightforward. Most of the pages you updated reflect well on the deployed site, though I am not sure on this section though. Refer: https://deploy-preview-1483--markbind-master.netlify.app/devguide/writingplugins

image

Looks like the popover content failed to be rendered properly here. Maybe there's a problem with that particular component/section?

docs/index.md
```diff
@@ -93,7 +93,7 @@ The example paragraph below has the following dynamic elements: a tooltip, a pop
 In <tooltip content="Computer Science">CS</tooltip>, a binary tree is a <trigger for="pop:index-tree">tree data structure</trigger> in which each node has at most two children, which are referred to as the _left child_ and the _right child_. <trigger trigger="click" for="modal:index-primitive">Primitive data types</trigger> on the other hand ...
 
 <popover id="pop:index-tree" header="An example tree data structure" placement="top">
-<div slot="content">
+<div #content>
```
jonahtanjz, line 96: Seems like this popover is not rendered as well.

image

docs/userGuide/components/advanced.md
```diff
@@ -65,11 +65,11 @@ However, we **$$do not recommend$$** using this syntax as it is deprecated and w
 <trigger for="modal:tip-example" trigger="click">Click here to show Modal.</trigger>
 
 <modal id="modal:tip-example">
-  <span slot="header" class="modal-title text-center">
+  <span #header class="modal-title text-center">
     <span style="font-size:20pt"><span style="color:red;">BIG</span> header</span>
   </span>
     Modal allows you to style both header and footer individually, with style classes and inline styles.
-  <span slot="footer" class="text-center">
+  <span #footer class="text-center">
```
jonahtanjz, line 72: Similar issue here

image

docs/userGuide/syntax/questions.mbdf
```diff
@@ -229,7 +229,7 @@ If no keywords are provided, the answer will always be marked as correct when pl
 Keywords are validated by simply looking for the keyword as a pattern in the user's answer!
 This works well for some
 <popover header="When does validation work?">cases
-  <span slot="content">
+  <span #content>
```
jonahtanjz, line 232: Here as well

image

jonahtanjz (commented):
Mostly looks good 👍 Just to add on to @ryoarmanda, seems like there are a few sections which are not rendering after changing from `slot` to `#`.

wxwxwxwx9:
Thanks for the spot guys! I should have checked more thoroughly. It came across as a little surprise to me that this happened as it worked for our existing test cases. After investigating the issue, it seems that I have missed out a crucial aspect of the `#attribute` shorthand feature that would break it. I discovered that the library we are using, `markdown-it`, is not fully-compliant with the [CommonMark spec](https://spec.commonmark.org/0.29). Thus, most of the time `#attribute` will be parsed successfully in our context, and sometimes it fails as per the CommonMark spec. 

Specifically, the syntax that we implemented is illegal under the CommonMark spec. 
Screen Shot 2021-02-22 at 9 22 06 AM

However, since `markdown-it` stated in their repo that they only *follow* the CommonMark spec, I should have looked into it more before assuming that they are fully compliant with the spec. My assumption was set in stone even more when it worked for the test cases we had in the codebase. I have really learnt a lesson here :-)

After discussing with @ang-zeyu about the possible resolutions / workaround for this problem, we have came up with the following:
- Patch the `markdown-it` library to make `#attribute` a legal attribute name.
  - Pros: The reason why we think this may be something we want is because it is a 1-to-1 match with [Vue's shorthand slot syntax](https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand). 
  - Cons: We will be breaking the CommonMark spec but in view of the benefit above, it may be worth the cost (up for discussion). Furthermore, we may be setting a bad precedent. 
- Use another shorthand (e.g. _attribute, :attribute) for the slot syntax.
  - Pros: Resolves the illegal attribute name issue for `#attribute` while complying with the CommonMark spec.
  - Cons: Have to decide which shorthand syntax to adopt (which may be a non-issue actually, but depends). 
- Stick with the original `slot="example"` syntax on the frontend for the users. 
  - Pros: Straightforward and no changes required on the user end. We have already patched the syntax to the new shorthand syntax when processing thus we don't have to worry about using deprecated syntax anymore (which is our main objective in the first place). 
  - Cons: Not so sure what cons this approach has currently (other than making this PR redundant). Let me know if you think if any!

@damithc would you have any advice regarding this? @jonahtanjz @ryoarmanda feel free to chip in as well if you have any thoughts about this!

damithc:
> @damithc would you have any advice regarding this?

No strong opinion. I guess we want to avoid patching third party libraries unless absolutely necessary. If the benefit of a new syntax is low, we can stick with the current syntax.

jonahtanjz:
>@jonahtanjz @ryoarmanda feel free to chip in as well if you have any thoughts about this!

I feel that option 2 is a good alternative as we want to allow users to use the slots shorthand for friendlier syntax as mentioned in #1078, as well as removing the use of deprecated Vue syntax on the user side (Not sure if this is a good enough benefit since it doesn't really matter, just ensuring that the syntax used on the user side conforms to Vue's deprecated syntax). This way we can avoid patching the third party library as Prof has mentioned and also abide by the CommonMark spec.

Option 3 seems ok as well since all `slot="exmaple"` will be processed to the new syntax and no changes on user side.

wxwxwxwx9:
Thanks for the input @damithc @jonahtanjz! I think we will shift the direction of this PR towards option 3. This means leaving the documentation slot syntax the same and reverting the changes in the user guide for shorthand syntax from #1464. 

We also have to update the Vue slot syntax for PageNavButton and SiteNavButton to the shorthand slot syntax as it's our internal component. I think I will change this in a new PR. 

@ang-zeyu does the above sound alright? If so, is it alright if I close this PR and create 2 new PRs to work on the 2 items above? :-)

ang-zeyu:
Sure, please go ahead. @wxwxwxwx9 

You could also reuse this PR to tackle the first item (remove the documentation for the `#shorthandsyntax`) as well, since there's a bit of discussion here

wxwxwxwx9:
@ang-zeyu I have made the changes and the relevant PRs. Regarding the possibility of a new shorthand syntax for slot, I think we can continue the discussion in the original issue #1078. Thanks so much for your advice and input!

ang-zeyu (approved):
> @ang-zeyu I have made the changes and the relevant PRs. Regarding the possibility of a new shorthand syntax for slot, I think we can continue the discussion in the original issue #1078. Thanks so much for your advice and input!

👍

PR #1493, opened 2021-02-26
Update intralink auto-conversion to convert links with URL fragment

### LINKED ISSUE #1490: Auto link extension conversion not working when linking to specific section in .md file
MacOS 11.1, Firefox v85.0.2

**MarkBind Version:**
v2.17.0

In `tipsAndTricks.md`, I tried to link to `deployingTheSite#using-ci-platforms` as such:

`[default CI configuration for deployment](deployingTheSite.md#using-ci-platforms)`

Link should be auto-converted to `deployingTheSite.html#using-ci-platforms`, as specified in https://markbind.org/userGuide/formattingContents.html#links

After clicking on the link:
image

wxwxwxwx9:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #1490.

**Overview of changes:**
Updated the implementation of `convertMbAndMbdExtToHtmlExt` to also convert links which have URL fragment (e.g. `index.html#section`). Have also added some tests to ensure that the conversions are working properly. 

Whether the implementation can be improved. 

**Testing instructions:**
`npm run test`

Update intralink auto-conversion to convert links with URL fragment. 

The intralink auto-conversion function does not account for links which
have URL fragment. 

Let's update it to convert links with URL fragment as well. 

---

### DIFF
--- a/packages/core/src/html/linkProcessor.js
+++ b/packages/core/src/html/linkProcessor.js
@@ -90,24 +90,27 @@ function convertMdAndMbdExtToHtmlExt(node) {
 
     const { href } = node.attribs;
 
+    if (utils.isUrl(href)) {
+      // Not intralink
+      return;
+    }
+
     const hrefUrl = url.parse(href);
 
     // get the first instance of URL fragment (first encounter of hash)
     const urlFragment = hrefUrl.hash === null ? '' : hrefUrl.hash;
-    const urlFragmentIdx = urlFragment === '' ? href.length : href.indexOf(urlFragment);
+    const urlPathName = hrefUrl.pathname === null ? '' : hrefUrl.pathname;
+    const ext = path.posix.extname(urlPathName);
 
-    let ext = '';
-    ext = href.substring(urlFragmentIdx - 3, urlFragmentIdx) === '.md' ? '.md' : ext;
-    ext = href.substring(urlFragmentIdx - 4, urlFragmentIdx) === '.mbd' ? '.mbd' : ext;
-    if (!ext) {
+    const isExtMdOrMbd = ext === '.md' || ext === '.mbd';
+    if (!isExtMdOrMbd) {
       // extension is neither .md nor .mbd, we do not need to process the link
       return;
     }
 
-    const pathWithoutExtAndFragmentLen = href.length - ext.length - urlFragment.length;
-    const pathWithoutExtAndFragment = href.substring(0, pathWithoutExtAndFragmentLen);
+    const urlPathNameWithoutExt = urlPathName.substring(0, urlPathName.length - ext.length);
 
-    const newHref = `${pathWithoutExtAndFragment}.html${urlFragment}`;
+    const newHref = `${urlPathNameWithoutExt}.html${urlFragment}`;
     node.attribs.href = newHref;
   }
 }
--- a/packages/core/src/html/linkProcessor.js
+++ b/packages/core/src/html/linkProcessor.js
@@ -98,9 +98,9 @@ function convertMdAndMbdExtToHtmlExt(node) {
     const hrefUrl = url.parse(href);
 
     // get the first instance of URL fragment (first encounter of hash)
-    const urlFragment = hrefUrl.hash === null ? '' : hrefUrl.hash;
-    const urlPathName = hrefUrl.pathname === null ? '' : hrefUrl.pathname;
-    const ext = path.posix.extname(urlPathName);
+    const fragment = hrefUrl.hash === null ? '' : hrefUrl.hash;
+    const pathName = hrefUrl.pathname === null ? '' : hrefUrl.pathname;
+    const ext = path.posix.extname(pathName);
 
     const isExtMdOrMbd = ext === '.md' || ext === '.mbd';
     if (!isExtMdOrMbd) {
@@ -108,9 +108,9 @@ function convertMdAndMbdExtToHtmlExt(node) {
       return;
     }
 
-    const urlPathNameWithoutExt = urlPathName.substring(0, urlPathName.length - ext.length);
+    const pathNameWithoutExt = pathName.substring(0, pathName.length - ext.length);
 
-    const newHref = `${urlPathNameWithoutExt}.html${urlFragment}`;
+    const newHref = `${pathNameWithoutExt}.html${fragment}`;
     node.attribs.href = newHref;
   }
 }
--- a/packages/core/src/html/linkProcessor.js
+++ b/packages/core/src/html/linkProcessor.js
@@ -90,30 +90,24 @@ function convertMdAndMbdExtToHtmlExt(node) {
 
     const { href } = node.attribs;
 
-    const hasMdExtension = href.includes('.md');
-    const hasMbdExtension = href.includes('.mbd');
+    const hrefUrl = url.parse(href);
 
-    const isExclusiveOr = (!hasMdExtension && hasMbdExtension) || (hasMdExtension && !hasMbdExtension);
-    if (!isExclusiveOr) {
-      // 1) has both '.md' and '.mbd' extension OR 2) has neither of the extensions
+    // get the first instance of URL fragment (first encounter of hash)
+    const urlFragment = hrefUrl.hash === null ? '' : hrefUrl.hash;
+    const urlFragmentIdx = urlFragment === '' ? href.length : href.indexOf(urlFragment);
+
+    let ext = '';
+    ext = href.substring(urlFragmentIdx - 3, urlFragmentIdx) === '.md' ? '.md' : ext;
+    ext = href.substring(urlFragmentIdx - 4, urlFragmentIdx) === '.mbd' ? '.mbd' : ext;
+    if (!ext) {
+      // extension is neither .md nor .mbd, we do not need to process the link
       return;
     }
 
-    const ext = hasMdExtension ? '.md' : '.mbd';
+    const pathWithoutExtAndFragmentLen = href.length - ext.length - urlFragment.length;
+    const pathWithoutExtAndFragment = href.substring(0, pathWithoutExtAndFragmentLen);
 
-    const extIdx = href.indexOf(ext);
-    const pathWithoutExtension = href.substring(0, extIdx);
-
-    const hasUrlFragment = href.includes(`${ext}#`);
-    if (hasUrlFragment) {
-      const fragmentIdx = extIdx + ext.length;
-      const fragment = href.substring(fragmentIdx, href.length);
-      const newHref = `${pathWithoutExtension}.html${fragment}`;
-      node.attribs.href = newHref;
-      return;
-    }
-
-    const newHref = `${pathWithoutExtension}.html`;
+    const newHref = `${pathWithoutExtAndFragment}.html${urlFragment}`;
     node.attribs.href = newHref;
   }
 }
--- a/packages/core/src/html/linkProcessor.js
+++ b/packages/core/src/html/linkProcessor.js
@@ -95,10 +95,10 @@ function convertMdAndMbdExtToHtmlExt(node) {
       const mdIdx = href.indexOf('.md');
       const pathWithoutExtension = href.substring(0, mdIdx);
 
-      const hasURIFragment = href.includes('.md#');
-      if (hasURIFragment) {
-        const URIFragment = href.substring(mdIdx + 3, href.length);
-        const newHref = `${pathWithoutExtension}.html${URIFragment}`;
+      const hasURLFragment = href.includes('.md#');
+      if (hasURLFragment) {
+        const fragment = href.substring(mdIdx + 3, href.length);
+        const newHref = `${pathWithoutExtension}.html${fragment}`;
         node.attribs.href = newHref;
         return;
       }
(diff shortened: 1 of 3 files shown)


packages/core/src/html/linkProcessor.js
```diff
@@ -87,16 +87,41 @@ function convertMdAndMbdExtToHtmlExt(node) {
     if (hasNoConvert) {
       return;
     }
+
     const { href } = node.attribs;
-    const hasMdExtension = href.slice(-3) === '.md';
+
+    const hasMdExtension = href.includes('.md');
     if (hasMdExtension) {
```
raysonkoh, line 94: Would having a guard clause for `hasMdExtension` and `hasURLFragment` be better so as to reduce nesting?

packages/core/src/html/linkProcessor.js
```diff
@@ -87,16 +87,41 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
+      }
+
+      const newHref = `${pathWithoutExtension}.html`;
       node.attribs.href = newHref;
       return;
     }
-    const hasMbdExtension = href.slice(-4) === '.mbd';
+
+    const hasMbdExtension = href.includes('.mbd');
     if (hasMbdExtension) {
```
raysonkoh, line 112: Similar comment as above - having guard clauses for `hasMbdExtension` and `hasURLFragment`

raysonkoh (commented):
Thanks for the fast response @wxwxwxwx9! Good work on adding the tests. I have checked and it is now auto-converting links as expected now!

While testing the `no-convert` property, I tried to follow the syntax as shown in the userGuide: `[description](link) {.no-convert}`, but it is not stopping the conversion from .md to .html. It only works when it is as such: `[description](link){no-convert}` . Could you help to verify this?

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Yup, that's a good suggestion! I think would make the different cases clearer as well. Will work on it 👍

wxwxwxwx9:
> While testing the `no-convert` property, I tried to follow the syntax as shown in the userGuide: `[description](link) {.no-convert}`, but it is not stopping the conversion from .md to .html. It only works when it is as such: `[description](link){no-convert}` . Could you help to verify this?

Yup, there's currently an error in the user guide! I've made PR #1494 to rectify it already :-)

ryoarmanda (commented):
The processing for `.md` and `.mbd` extension looks identical as far as I can see, the only difference is in making the `fragment` variable which I guess is related to the length of the extension.

I think we can make it more concise by figuring out first which extension is applicable to us, and then do the processing with respect to that extension.

wxwxwxwx9:
@ryoarmanda @raysonkoh thanks so much for the suggestion! I've made the changes accordingly and the code is more succinct now. I tried my best to make the logics straightforward but let me know if the logic is confusing or if there is anything I can improve on!

packages/core/src/html/linkProcessor.js
```diff
@@ -87,18 +87,34 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
       return;
     }
+
     const { href } = node.attribs;
-    const hasMdExtension = href.slice(-3) === '.md';
-    if (hasMdExtension) {
-      const newHref = `${href.substring(0, href.length - 3)}.html`;
-      node.attribs.href = newHref;
+
+    const hasMdExtension = href.includes('.md');
```
ang-zeyu, line 93: let's enhance the check since we're supporting hashes now: e.g. you may have a false positive `x.md.jpg`

packages/core/src/html/linkProcessor.js
```diff
@@ -87,18 +87,34 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
+    const hasMbdExtension = href.includes('.mbd');
+
+    const isExclusiveOr = (!hasMdExtension && hasMbdExtension) || (hasMdExtension && !hasMbdExtension);
+    if (!isExclusiveOr) {
+      // 1) has both '.md' and '.mbd' extension OR 2) has neither of the extensions
```
ang-zeyu, line 98: what's case 1 for? 👀

packages/core/src/html/linkProcessor.js
```diff
@@ -87,18 +87,34 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
-    const hasMbdExtension = href.slice(-4) === '.mbd';
-    if (hasMbdExtension) {
-      const newHref = `${href.substring(0, href.length - 4)}.html`;
+
+    const ext = hasMdExtension ? '.md' : '.mbd';
+
+    const extIdx = href.indexOf(ext);
+    const pathWithoutExtension = href.substring(0, extIdx);
+
+    const hasUrlFragment = href.includes(`${ext}#`);
```
ang-zeyu, line 107: could use https://nodejs.org/api/url.html to simplify things here

packages/core/test/unit/html/linkProcessor.test.js
```diff
@@ -35,6 +35,83 @@ const mockConfig = {
   ],
 };
 
+test('Test converting .md to .html', () => {
```
ang-zeyu, line 38: Since the logic is the same, could use jest's `test.each` here

packages/core/test/unit/html/linkProcessor.test.js
```diff
@@ -35,6 +35,83 @@ const mockConfig = {
...
+  const mockNode = cheerio.parseHTML(mockLink)[0];
+
+  const EXPECTED_RESULT = '<a href="/index">Test</a>';
+
+  linkProcessor.convertMdAndMbdExtToHtmlExt(mockNode);
+
+  expect(cheerio.html(mockNode)).toEqual(EXPECTED_RESULT);
+});
+
+test('Test no conversion when both .md and .mbd extensions are present', () => {
```
ang-zeyu, line 104: if I read this right, this can be generalised into a test for `.md/mbd` appearing in the middle

ang-zeyu (commented):
Thanks! Couple of suggestions:

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Would this work better? 

```
    const hasMdExtension = href.slice(-3) === '.md' || href.includes('.md#');
    const hasMbdExtension = href.slice(-4) === '.mbd' || href.includes('.mbd#');
```

The above should handle the case that you suggested, `x.md.jpg`. 

Typically, only one of the two conditions will be true (XOR). 

The edge case is if the link has both `md` and `mbd` extension, which can fall into the following cases:
1) .md#.mbd
2) .mbd#.md
3) both .md# and .mbd# appears somewhere in the link

I think cases 1 and 2 are still manageable but should we handle case 3?

packages/core/src/html/linkProcessor.js
wxwxwxwx9: I was thinking that if both cases were present, then we would deem it as "invalid". I thought that it would be quite rare to have such a case and I don't think I would expect it to happen. What do you think? :O

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Thanks for the tip! Will use the API :-)

packages/core/src/html/linkProcessor.js
ang-zeyu: hmm would just the old check with https://github.com/MarkBind/markbind/pull/1493#discussion_r588378063 work for all 3 cases?

packages/core/src/html/linkProcessor.js
ang-zeyu: ^ https://github.com/MarkBind/markbind/pull/1493#discussion_r588983611

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Yup, using hash function in the URL library would work. Thanks for the advice!

packages/core/test/unit/html/linkProcessor.test.js
wxwxwxwx9: I've updated the test cases to use `test.each`. However, it comes at a cost of slightly less readable test cases. I tried to alleviate it by adding a descriptive test comment above each test case array. What do you think about it? @ang-zeyu

packages/core/test/unit/html/linkProcessor.test.js
wxwxwxwx9: Hmm not sure if I fully understand this part, but I've added new test cases which I think may cover this. For example, `// Test conversion when both .md and .mbd are present, and .mbd is extension, with URL fragment`. Let me know if I'm still missing out on this!

packages/core/test/unit/html/linkProcessor.test.js
```diff
@@ -35,6 +35,36 @@ const mockConfig = {
...
+  ['.mbd', '', '#test-1', '', '<a href="/index.html#test-1">Test</a>'],
+  // Test conversion when both .md and .mbd are present, and .mbd is extension, with URL fragment
+  ['.md', '.mbd', '#test-1', '', '<a href="/index.md.html#test-1">Test</a>'],
+  // Test conversion when both .md and .mbd are present, and .md is extension, with URL fragment
+  ['.mbd', '.md', '#test-1', '', '<a href="/index.mbd.html#test-1">Test</a>'],
+  // Test no conversion when no-convert attribute is present
+  ['.md', '', '#test-1', 'no-convert', '<a href="/index.md#test-1" no-convert>Test</a>'],
+  // Test no conversion when neither .md nor .mbd extensions are present
+  ['', '', '', '', '<a href="/index">Test</a>'],
+])('Test link auto-conversion for (%s, %s, %s, %s)', (ext1, ext2, fragment, attr, expected) => {
```
wxwxwxwx9, line 59: It's a little hard to come up with a proper test name that covers all the scenarios above. Do you think this is alright? Or should we aim to describe it more? @ang-zeyu

packages/core/src/html/linkProcessor.js
```diff
@@ -87,18 +87,28 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
-      node.attribs.href = newHref;
+
+    const hrefUrl = url.parse(href);
+
+    // get the first instance of URL fragment (first encounter of hash)
+    const urlFragment = hrefUrl.hash === null ? '' : hrefUrl.hash;
+    const urlFragmentIdx = urlFragment === '' ? href.length : href.indexOf(urlFragment);
+
+    let ext = '';
+    ext = href.substring(urlFragmentIdx - 3, urlFragmentIdx) === '.md' ? '.md' : ext;
```
ang-zeyu, line 100: `href.substring(urlFragmentIdx - 3, urlFragmentIdx)` is available under `hrefUrl.pathname`

packages/core/src/html/linkProcessor.js
```diff
@@ -87,18 +87,28 @@ function convertMdAndMbdExtToHtmlExt(node) {
...
+      // extension is neither .md nor .mbd, we do not need to process the link
       return;
     }
-    const hasMbdExtension = href.slice(-4) === '.mbd';
-    if (hasMbdExtension) {
-      const newHref = `${href.substring(0, href.length - 4)}.html`;
-      node.attribs.href = newHref;
-    }
+
+    const pathWithoutExtAndFragmentLen = href.length - ext.length - urlFragment.length;
```
ang-zeyu, line 107: can be simplified slightly with `hrefUrl.pathname` too

packages/core/test/unit/html/linkProcessor.test.js
ang-zeyu: this looks ok, especially since you've documented the test cases above.

Rather, it might be a little roundabout with the many parameters.

Let's compress the inputs into just one parameter (the input html) so the reader dosen't have to stitch them together, especially since all of them are rather short. Its also self-documenting (comments probably not needed in that case).

wxwxwxwx9: Yep, that sounds a lot better! I initially separated them into different parameters as I thought it would make it more "modifiable" but in hindsight, it really is quite a roundabout (and quite unnecessary). 

But I think there's some value in keeping the comments, to make it more explicit and easier for the reader to quickly find out the semantics of the test case. What do you think? @ang-zeyu

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Hmm actually I'm trying to get the extension here (not the pathname). In the updated implementation I used `path.posix.extname` on `hrefUrl.pathname` instead, thanks for your suggestion! :-)

packages/core/src/html/linkProcessor.js
wxwxwxwx9: Yup, that's more succinct. Thanks for the suggestion!

wxwxwxwx9:
@ang-zeyu I noticed that the other test cases in `linkProcessor.test.js` could also be converted `test.each` for brevity. I was thinking maybe we can leave it as a "good first issue" or should we rectify it in another PR? :-)

packages/core/test/unit/html/linkProcessor.test.js
ang-zeyu: sure 👍

ang-zeyu (approved):
Lgtm 👍

ang-zeyu:
@wxwxwxwx9 I've shortened it but take note the commit title needs to be under 64 as a hard limit (8 characters from pr number github appends).

PR #1494, opened 2021-02-26
Fix no-conversion syntax for intralink in user guide

wxwxwxwx9:

- [x] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

There's currently a minor error in the user guide for no-conversion syntax. 

**Overview of changes:**
Updated the error in the user guide for no-conversion syntax. 

NIL

**Testing instructions:**
`npm run test`

Fix no-conversion syntax for intralink in user guide.

---

### DIFF
--- a/docs/userGuide/syntax/links.mbdf
+++ b/docs/userGuide/syntax/links.mbdf
@@ -44,7 +44,7 @@ You may link to markdown files using its original extension (**.md**) as it will
 If you wish to disable the auto-conversion, you may use the `no-convert` attribute in your link. 
 
 {{ icon_example }}
-`Click [here](index.md) {.no-convert}`
+`Click [here](index.md){no-convert}`
 
 </box>
 


ryoarmanda (commented):
Nice catch! :+1: Been wondering why I can't make the auto-conversion work haha

docs/userGuide/syntax/links.mbdf
```diff
@@ -44,7 +44,7 @@ You may link to markdown files using its original extension (**.md**) as it will
 If you wish to disable the auto-conversion, you may use the `no-convert` attribute in your link. 
 
 {{ icon_example }}
-`Click [here](index.md) {.no-convert}`
+`Click [here](index.md){no-convert}`
```
jonahtanjz, line 47: Just a suggestion, should we leave a space before the '{' to ensure that it is consistent with the user guide on [Classes, Attributes & Identifiers](https://markbind-master.netlify.app/userguide/formattingcontents#classes-attributes-and-amp-identifiers)?

wxwxwxwx9: Hi @jonahtanjz! The library that we are using, `markdown-it-attrs` actually parses inline curly braces a little differently. There must not be a space in between in this example, otherwise it wouldn't work. The space I believe is only for block-level elements. 

But to be honest, I think our user guide probably isn't that clear on this aspect. For example, the phrase `Some other types of markdown have different placements of the curly group {...}.` from the user guide doesn't quite stand out but it's an important piece of information. 

I was thinking maybe we can relook at this portion of the user guide sometime in the future to improve the clarity. But what do you think about this section of the user guide? :-)

jonahtanjz: > Hi @jonahtanjz! The library that we are using, markdown-it-attrs actually parses inline curly braces a little differently. There must not be a space in between in this example, otherwise it wouldn't work. The space I believe is only for block-level elements.

Ahh alright thanks for clarifying 👍

> But what do you think about this section of the user guide? :-)

Yup agree with you that this portion is slightly vague. Perhaps can draw more attention towards this line and also include a few examples of the different placements of `{...}` so that users can have a clearer idea of this.

ang-zeyu (approved):

PR #1498, opened 2021-03-02
Show all tabs when printing

### LINKED ISSUE #1491: Support a better format for tabs
Tabs are supposed to contain alternative contents. Readers can click on each tab to see its content.
However, when a page is printed (or converted to PDF), the reader can see only the content of the active tab.

Suggestion: Show all tabs when printing.

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1491. 

1. Show all the content and display the header of the tabs when printing. 
2. Tabs in `tab-group` will be grouped together. 
3. The tab navigation is hidden when printing to prevent repetition as the header is already added to the top of each section.

1. Should disabled tabs be printed as well? The current implementation will print out all tabs regardless of whether they are disabled or not. 
2. Is the current format of the `tab-group` fine? Do let me know if this can be improved on.

**Testing instructions:**
1. Create a few `tabs` and `tab-group`
```
<tabs>
  <tab header="First tab">
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla ullamcorper ultrices lobortis.
  </tab>
  <tab header="Second tab">
    Some content of the second tab...
  </tab>
  <tab header="Third tab">
    Additional content from third tab...
  </tab>
  <tab-group header="Example Tab Group">
    <tab header="Tab Group 1">
      Some content in a tab group 1...
    </tab>
    <tab header="Tab Group 2">
      Some content in a tab group 2...
    </tab>
  </tab-group>
</tabs>
```
2. Print the page
3. All tabs should be shown in the preview

Support printing of all tabs

Tabs are supposed to contain alternative contents. 
Readers can click on each tab to see its content. However, 
when a page is printed or converted to PDF, the reader can 
only see the content of the active tab.

Let's display all the tabs and its content when it is printed.

---

### DIFF
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -3,13 +3,13 @@
     <div
       v-show="show"
       role="tabpanel"
-      class="tab-pane active"
+      class="tab-pane active printable-tab-pane"
       :class="{hide:!show}"
     >
-      <slot></slot>
-      <div ref="header" class="d-none">
+      <div ref="header" class="d-none printable-tab-header">
         <slot name="_header"></slot>
       </div>
+      <slot></slot>
       <hr />
     </div>
   </transition>
@@ -93,5 +93,18 @@ export default {
 
     .fade-leave-active {
         transition: opacity 0s;
+    }
+
+    @media print {
+        .printable-tab-header {
+            display: block !important;
+            margin: 10px 0;
+            font-weight: 600;
+            font-size: 1.5rem;
+        }
+
+        .printable-tab-pane {
+            display: block !important;
+        }
     }
 </style>
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -3,13 +3,18 @@
     <div
       v-show="show"
       role="tabpanel"
-      class="tab-pane active printable-tab-pane"
+      class="tab-pane active printable-tab-pane d-print-block"
       :class="{hide:!show}"
     >
-      <div ref="header" class="d-none printable-tab-header">
-        <slot name="_header"></slot>
+      <div class="nav-tabs printable-tab-header d-none d-print-flex">
+        <div class="nav-link active">
+          <span><slot name="_header"></slot></span>
+        </div>
       </div>
       <slot></slot>
+      <div ref="header" class="d-none">
+        <slot name="_header"></slot>
+      </div>
       <hr />
     </div>
   </transition>
@@ -97,14 +102,15 @@ export default {
 
     @media print {
         .printable-tab-header {
-            display: block !important;
-            margin: 10px 0;
-            font-weight: 600;
-            font-size: 1.3rem;
+            margin-bottom: 15px;
+            border-bottom: 1px solid #dee2e6;
+        }
+
+        .printable-tab-header > div {
+            margin-bottom: -2px;
         }
 
         .printable-tab-pane {
-            display: block !important;
             padding: 10px;
         }
     }
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -100,11 +100,12 @@ export default {
             display: block !important;
             margin: 10px 0;
             font-weight: 600;
-            font-size: 1.5rem;
+            font-size: 1.3rem;
         }
 
         .printable-tab-pane {
             display: block !important;
+            padding: 10px;
         }
     }
 </style>
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -7,14 +7,11 @@
       :class="{hide:!show}"
     >
       <div class="nav-tabs printable-tab-header d-none d-print-flex">
-        <div class="nav-link active">
-          <span><slot name="_header"></slot></span>
+        <div ref="header" class="nav-link active">
+          <slot name="_header"></slot>
         </div>
       </div>
       <slot></slot>
-      <div ref="header" class="d-none">
-        <slot name="_header"></slot>
-      </div>
       <hr />
     </div>
   </transition>
--- a/packages/vue-components/src/TabGroup.vue
+++ b/packages/vue-components/src/TabGroup.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="printable-tab-group">
-    <div ref="header" class="d-none printable-tab-group-header">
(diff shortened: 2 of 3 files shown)


packages/vue-components/src/Tab.vue
```diff
@@ -94,4 +94,18 @@ export default {
     .fade-leave-active {
         transition: opacity 0s;
     }
+
+    @media print {
+        .printable-tab-header {
+            display: block !important;
```
wxwxwxwx9, line 100: I'm thinking maybe we can use the `d-print-block` class, since we are using bootstrap. And it may look a little better than having to use `!important` for css. However, I don't think it's going to impact much if we choose either, because we still need the print media query and classes here to set the styling for the pdf to look presentable. Just a personal suggestion :-)

packages/vue-components/src/Tab.vue
```diff
@@ -94,4 +94,18 @@ export default {
...
+            margin: 10px 0;
+            font-weight: 600;
+            font-size: 1.3rem;
+        }
+
+        .printable-tab-pane {
+            display: block !important;
```
wxwxwxwx9, line 107: Same as the above!

packages/vue-components/src/TabGroup.vue
```diff
@@ -71,4 +71,19 @@ export default {
...
+    @media print {
+        .printable-tab-group {
+            border: 1px solid #dee2e6;
+            border-radius: 5px;
+            padding: 10px;
+            margin: 10px 0;
+        }
+
+        .printable-tab-group-header {
+            display: block !important;
```
wxwxwxwx9, line 84: Same as the above!

packages/vue-components/src/Tabset.vue
```diff
@@ -2,7 +2,7 @@
   <div :class="[addClass]">
     <!-- Nav tabs -->
     <ul
-      class="nav nav-tabs"
+      class="nav nav-tabs no-print"
```
wxwxwxwx9, line 5: We can actually use `d-print-none` here, so we don't need the print media query :-)

wxwxwxwx9:
Hi @jonahtanjz thanks for your work on this PR! 

> Should disabled tabs be printed as well? The current implementation will print out all tabs regardless of whether they are disabled or not.

Actually I don't have a strong opinion of whether it should be printed, because I'm not sure why a tab would be disabled in the first place (can't really think of a use-case). So I think for brevity sake, I'm alright with them being printed. Since if the user wants to show the tab to the viewers in the first place (although disabled), it probably wants the viewer to be aware of the tab's existence so I think it makes sense for the disabled tab to appear in the print. 

> Is the current format of the tab-group fine? Do let me know if this can be improved on.

I think it looks decent! I think the font bolding for each tab's header and the underlining of the tab group header makes sense. I didn't have any trouble discerning between tab groups and tabs, and their respective content. Seems quite clear to me. 

Overall everything looks good to me, just have some suggestions on how to make things a little more succinct using [bootstrap's utilities](https://getbootstrap.com/docs/4.0/utilities/display/#display-in-print).

The implementation seems to be working well. Good job! :-)

damithc:
@jonahtanjz can provide some screenshots for the feature in action?

wxwxwxwx9:
@damithc posting on behalf of @jonahtanjz since I still have the page open. 

Web view
Screen Shot 2021-03-04 at 6 18 16 PM

Print view
Screen Shot 2021-03-04 at 6 18 40 PM

damithc:
Thanks for the screenshots @wxwxwxwx9 
Is it possible to style the content to show each as a single tab? So that the reader can immediately recognize these as tabs being spread out.

jonahtanjz:
> Overall everything looks good to me, just have some suggestions on how to make things a little more succinct using bootstrap's display utilities.

Thanks @wxwxwxwx9 for the review and posting the screenshots 👍 Yup, agree that using bootstrap's display utilities will make things more succinct. Will make the changes accordingly :)

>Is it possible to style the content to show each as a single tab? So that the reader can immediately recognize these as tabs being spread out.

I think it's possible to do it for normal tabs but I'm not very sure if this format will work well with tab groups. Maybe for tab groups we can keep the current formatting with the box but each of the tabs will be display as a single tab inside the box. What do you think about this @damithc?

damithc:
> I think it's possible to do it for normal tabs but I'm not very sure if this format will work well with tab groups. Maybe for tab groups we can keep the current formatting with the box but each of the tabs will be displayed as a single tab inside the box. What do you think about this @damithc?

Need to see how it looks. The goal is to visually indicate to the reader that the content came from separate tabs, as there was a reason for them to be in tabs in the first place (most likely because they were alternatives) and that additional information should be preserved if possible.

jonahtanjz:
> Need to see how it looks.

I've come up with something like this. Using the same example from the screenshot provided above, each of the `tab` will be displayed as a single tab inside a box so that reader can recognize that all these are part of the same set of `tabs`. Tab groups will further have another box with the header underlined.

image

damithc:
> I've come up with something like this. Using the same example from the screenshot provided above, each of the `tab` will be displayed as a single tab inside a box so that reader can recognize that all these are part of the same set of `tabs`. Tab groups will further have another box with the header underlined.

Looks good @jonahtanjz I don't think the tab group font should be bigger than the top level tab names, to preserve the hierarchy.

jonahtanjz:
> I don't think the tab group font should be bigger than the top level tab names, to preserve the hierarchy.

Alright noted. I have changed the font size to match with the rest of the headers :)

packages/vue-components/src/Tab.vue
```diff
@@ -3,9 +3,14 @@
     <div
       v-show="show"
       role="tabpanel"
-      class="tab-pane active"
+      class="tab-pane active printable-tab-pane d-print-block"
       :class="{hide:!show}"
     >
+      <div class="nav-tabs printable-tab-header d-none d-print-flex">
```
ang-zeyu, line 9: Thanks!

Just one question: could we merge this with the existing `header` below? similar to what you did for `TabGroup`

jonahtanjz: Yup it's possible. Thanks for the suggestion and I've made the changes accordingly :)

ang-zeyu (approved):
Lgtm 👍

PR #1501, opened 2021-03-04
Fix escaped html characters in search result

### LINKED ISSUE #1486: Special characters not rendered correctly in search results
This is from the current UG
image

wxwxwxwx9:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #1486.

**Overview of changes:**
Used `v-html` to render unescaped html characters (in place of the original mustache expression). 

NIL

**Testing instructions:**
`npm run test`

Serve the user guide and search for "include" keyword. 

`&lt;` and `&gt;` that originally appears should now be replaced by `<` and `>` respectively, since the characters are no longer escaped. 

Fix escaped html characters in search result.

Certain html characters are automatically escaped for
security reasons. However, it is not needed in our context
as the html is supplied by the user.  

Let's prevent html characters from getting escaped so as
to render the html correctly. 

---

### DIFF
--- a/packages/vue-components/src/SearchbarPageItem.vue
+++ b/packages/vue-components/src/SearchbarPageItem.vue
@@ -1,7 +1,7 @@
 <template>
   <div v-if="item.heading" class="heading">
     <div class="heading-text">
-      {{ item.heading.text }}
+      <span v-html="item.heading.text"></span>
     </div>
     <div class="heading-text-items">
       <small v-html="highlight(item.heading.text, value)"></small>


jonahtanjz:
Thanks @wxwxwxwx9 for fixing this :) I've tested it and it seems to be working well 👍 

On a separate note, since in our context usage of `v-html` doesn't cause any security threat, I was thinking that we should disable the eslint rule for `no-v-html` so that we can resolve all the eslint warnings during the ci checks but that can be a separate discussion/issue. What do you think about this?

wxwxwxwx9:
@jonahtanjz thanks for the review! 

> On a separate note, since in our context usage of v-html doesn't cause any security threat, I was thinking that we should disable the eslint rule for no-v-html so that we can resolve all the eslint warnings during the ci checks but that can be a separate discussion/issue. What do you think about this?

Yup! I agree with you. We can do that in another PR :-)

ang-zeyu (approved):
Lgtm 👍 

> > On a separate note, since in our context usage of v-html doesn't cause any security threat, I was thinking that we should disable the eslint rule for no-v-html so that we can resolve all the eslint warnings during the ci checks but that can be a separate discussion/issue. What do you think about this?
> 
> Yup! I agree with you. We can do that in another PR :-)

It's a good rule to have in case anything slips by, I think, even if it means spewing these out constantly.

We could deal with these carefully on a case by case basis though. (`eslint-disable-next-line`)

Issue #1472, opened 2021-02-09
Last dropdown header not processed


jonahtanjz:
**Tell us about your environment**

* **MarkBind Version:** 2.17.0

**What did you do? Please include the actual source code causing the issue.**
Created three dropdowns wrapped in a `div`.

```html
<div>
  <!-- First dropdown -->
  <dropdown header="*Dropdown 1*" type="primary">
    <li><a href="#dropdown" class="dropdown-item">Action</a></li>
  </dropdown>
  <!-- Second dropdown -->
  <dropdown header="Dropdown 2" type="primary" menu-align-right>
    <li><a href="#dropdown" class="dropdown-item">Something else here</a></li>
  </dropdown>
  <!-- Third dropdown -->
  <dropdown header="Dropdown 3" type="primary" menu-align-right>
    <li><a href="#dropdown" class="dropdown-item">Something else here</a></li>
  </dropdown>
</div>
```

I noticed that this issue only occurs when the dropdowns are in a wrapper like `div` or `box`. This issue can be seen on the user guide https://markbind-master.netlify.app/userguide/usingcomponents#dropdowns, where the `Right aligned list` header is missing.

**What did you expect to happen?**
Three dropdowns with their respective header shown
![image](https://user-images.githubusercontent.com/47470981/107372580-8639ea00-6b20-11eb-90e9-e479c1d6bf03.png)

**What actually happened? Please include the actual, raw output.**
However, the last dropdown header is missing. 
![image](https://user-images.githubusercontent.com/47470981/107372354-496df300-6b20-11eb-957f-f630f3e3676c.png)

Upon further inspection, it seems the header attribute for the last dropdown is not processed.
![image](https://user-images.githubusercontent.com/47470981/107372495-70c4c000-6b20-11eb-962a-5ecb8f45e9c5.png)

ang-zeyu:
Thanks for the catch! @jonahtanjz 

Could this be related to #1467? @wxwxwxwx9
The version in markbind.org seems fine

wxwxwxwx9:
@jonahtanjz thanks for highlighting this :-) 

[...]

PR #1512, opened 2021-03-10
Compile Vue Page at build time to alleviate FOUC

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

Addresses the first part of #1406 by compiling our page (Vue App) during build time and thus eliminating FOUC for majority of the use-cases. 

**Overview of changes:**

**The approach has largely changed. Refer to the discussion in the reviews below.**

~~**<ins> Explanation of Approach </ins>**~~
~~During the generation of the page and after the relevant processings of the page are finished, I introduced a new step of compiling the page into a Vue Application using `vue-template-compiler`. The compilation returns me with an object that contains the relevant information about the rendered Vue app. Only 2 attributes are important: `render` and `staticRenderFns`. These are the render functions provided by Vue that will be used to generate the view in the browser.~~

~~After getting these 2 things, I map them to the route of the page in an object. This object keeps a mapping of a page's route (e.g. `/index.html`) to its respective render function, so that we can pass in this render function to Vue during mounting to render the correct page view according to the current page route.~~

~~This object will then be output into a javascript file, `pageVueRenderFns.min.js`, that will be sent along with the rest of the assets to the browser so that our MarkBind plugin (core-web) can access the object, retrieve the appropriate render function according to the current page route, and pass the render function into the Vue instance during mounting.~~

~~Technically, with the above, we no longer need to insert the page content into our `page.njk` template. This is because we have already compiled the content into the render function. However, it is still necessary to insert the page content into the Nunjucks template because we have to keep track of the DOM structure for snapshot testing.~~

~~This also means that we will have to send the uncompiled page along with `index.html` to the browser. However, this is undesirable as we do not want the uncompiled version of the page to be rendered (this will cause FOUC and defeat the purpose of pre-compiling in the first place). Thus, I introduced another script, `initAppNodeForPageVueRender.min.js`, and placed it before the uncompiled page to delete it before it can be rendered. The script will also create a fresh `#app` node for Vue to mount, using the render function we have from pre-compiling.~~

~~**<ins> Removal of the interim fix for FOUC </ins>**~~
~~Due to the resolution of the FOUC problem, we no longer need the interim fix provided by `tempStyleProcessor`. Thus, I have deleted all the relevant code and the file itself.~~

~~By removing the interim fix, this PR will also address the dropdown issue from #1472.~~

**<ins> Adding `type="application/javascript"` attribute to script tags </ins>**
By default, Vue compilation ignores and discards any script nodes during its parsing and compilation. It will provide the following error/warning message. 

`Error: Templates should only be responsible for mapping the state to the UI. Avoid placing tags with side-effects in your templates, such as <script>`

I noticed that we have some use-cases for script tags in our source files currently. For example, in the `index.md` of our UserGuide, we will redirect the user with the following script tag:

```
<script>
  window.location.href = "gettingStarted.html"
</script>
```

If we do not add the attribute `type="application/javascript"` to the script tag, the above will be discarded during compilation. Thus, I introduced a new processing step for `script` nodes to add the attribute. I don't think this will affect any existing features or cause any regression. At least I have not seen any so far. But let me know if this is something that may be undesirable :-)

**<ins> Changing `<style>` tags name to workaround Vue compilation discard </ins>**
Vue compilation discards any `<style>` tags, similar to `<script>`. But `<style>` tags does not have a straightforward solution like `<script>` (read above). We have to change the name of style tag into a placeholder name for Vue compilation so that Vue doesn't discard the element. We will then change the name back to style tag after compilation.

In general:
- Whether there is any problem with my approach or anything that was mentioned above
- Ways to improve the implementation
- Any potential bugs that may be introduced by the changes that I should look into
- Whether it will affect any of the existing features or implementations

~~Also, I have currently commented out the code where I minified `pageVueRenderFns.min.js` and `initAppNodeForPageVueRender.min.js` using `UglifyJs`, as I did not notice a significant reduction in file size (from 4.3MB to 4MB only for `pageVueRenderFns.min.js`). Just wanted to check if it's worth using the library considering that there's not much reduction?~~

**Testing instructions:**
`npm run test`

~~Serve a sizable MarkBind project (e.g. our documentation) and you should see that the FOUC doesn't occur anymore for the navbar.~~

Compile Vue Page at build time to alleviate FOUC.

Compiling each Vue page in the browser may be expensive as it is 
possible for a page to get very large in size. This creates the problem
of flash-of-unstyled-content (FOUC) for the user which is unsightly
and may affect user experience. Although there are temporary styles
in place to tackle the problem, the FOUC problem largely remains. 

Let's compile every Vue page during build time to alleviate FOUC. 

---


jonahtanjz:
@wxwxwxwx9 Good work on this PR 👍 

There seems to be an issue with rendering some of the pages such as [User Guide](https://deploy-preview-1512--markbind-master.netlify.app/userguide/) and [Developer Guide](https://deploy-preview-1512--markbind-master.netlify.app/devguide/). It is displaying a blank page.

After looking around, it seems the [`pageVueRenderFns`](https://github.com/MarkBind/markbind/blob/f4130525bbfce91a583efb0c865d6e5d4bd4c215/packages/core-web/src/index.js#L180) in `core-web` is not loading the updated version. It only has `/` and `/index.html` from what I can see after visiting different pages.

Another part that I think might have an issue is the [`pagePath`](https://github.com/MarkBind/markbind/blob/f4130525bbfce91a583efb0c865d6e5d4bd4c215/packages/core/src/Page/index.js#L507) in the `compileVuePageAndMapRenderFn` function. The relative path on Windows will use backslash (eg. userGuide\index.html) whereas [`window.location.pathname`](https://github.com/MarkBind/markbind/blob/f4130525bbfce91a583efb0c865d6e5d4bd4c215/packages/core-web/src/index.js#L177) will use forward slash. This will result in a different path which will cause `pageVueRenderFns[pageRoute]` to return `undefined`.

wxwxwxwx9:
Thanks for the review and update! @jonahtanjz 

From what you described, it seems there are indeed some OS compatibility issues. It's currently working alright on my MacOS. 

I will look into the issue :-)

packages/cli/test/functional/test_site/expected/markbind/js/pageVueRenderFns.min.js
```diff
@@ -0,0 +1 @@
..._v(\"qux\")]),_v(\" \"),_c('span',{pre:true,attrs:{\"class\":\"hljs-attr\"}},[_v(\"type\")]),_v(\"=\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-string\"}},[_v(\"\\\"name\\\"\")]),_v(\">\")]),_c('span',{pre:true,attrs:{\"class\":\"highlighted\"}},[_v(\"go,o\")]),_c('span',{pre:true,attrs:{\"class\":\"hljs-tag highlighted\"}},[_v(\"</\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-name\"}},[_v(\"qux\")]),_v(\">\")]),_v(\"\\n\")]),_c('span',[_v(\"  \"),_c('span',{pre:true,attrs:{\"class\":\"hljs-tag\"}},[_v(\"<\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-name\"}},[_v(\"quux\")]),_v(\" \"),_c('span',{pre:true,attrs:{\"class\":\"hljs-attr\"}},[_v(\"type\")]),_v(\"=\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-string\"}},[_v(\"\\\"name\\\"\")]),_v(\">\")]),_c('span',{pre:true,attrs:{\"class\":\"highlighted\"}},[_v(\"go'o\")]),_c('span',{pre:true,attrs:{\"class\":\"hljs-tag\"}},[_v(\"</\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-name\"}},[_v(\"quux\")]),_v(\">\")]),_v(\"\\n\")]),_c('span',[_c('span',{pre:true,attrs:{\"class\":\"hljs-tag\"}},[_v(\"</\"),_c('span',{pre:true,attrs:{\"class\":\"hljs-name\"}},[_v(\"foo\")]),_v(\">\")]),_v(\"\\n\")])])]),_c('p',[_c('strong',[_c('cod
```
ang-zeyu, line 1: Thanks! Look forward to the finished product. @wxwxwxwx9 

quick comment: let's split this up - one file per page (or just inline it into the html) to make this scalable for something like the 2103 site.

wxwxwxwx9: Thanks for the advice! I have split it accordingly. :-)

But I'm not sure if the naming convention that I came up with is the best. Let me know what you think about it!

packages/cli/test/functional/test_site_special_tags/expected/markbind/js/initAppNode.page-vue-render.js
```diff
@@ -0,0 +1,7 @@
+
+      var body = document.querySelector('body');
```
ang-zeyu, line 2: 👀 not sure if this part is a good way to reduce fouc; it does remove the flash completely, but at the cost of a longer time to first-content paint.

this should be fixed with ssr as well

packages/core/src/Page/index.js
```diff
@@ -468,8 +470,15 @@ class Page {
       ...layoutManager.getLayoutPageNjkAssets(this.layout),
     };
 
+    const pageVueRenderFnsScriptFileName = this.getPageVueRenderFnsScriptFileName();
```
ang-zeyu, line 473: how about just the file's basename? (+ some prefix maybe) (then output it right beside the page, instead of everything under `assets`)

also since you can't have two files with the same name in any directory - so no need to ensure no conflicts, which is still possible here (`x/y/abc.js` -> `x-y-abc.js`, `x-y-abc.js` -> `x-y-abc.js`).

packages/core/src/html/NodeProcessor.js
```diff
@@ -823,8 +837,6 @@ class NodeProcessor {
 
     this.postProcessNode(node);
 
-    insertTemporaryStyles(node);
```
ang-zeyu, line 826: hmm not sure if we should be removing this here right away.
compilation alone isn't enough to completely remove fouc

ang-zeyu (commented):
Mostly looks good 👍

A few things I think we should be defering to the SSR pr, and some nits:

packages/cli/test/functional/test_site_special_tags/expected/markbind/js/initAppNode.page-vue-render.js
wxwxwxwx9: Hmm, should we implement SSR in this PR as well? If not, do you think this implementation (at the cost of longer first-content paint time) is alright until we can implement the SSR? Personally I think it's better to have a longer time to first-content paint than FOUC. How should we approach this? :O

packages/core/src/Page/index.js
wxwxwxwx9: Ah, yup, thanks for the spot! The solution you suggested is a lot more fool-proof. Will look into it!

packages/cli/test/functional/test_site_special_tags/expected/markbind/js/initAppNode.page-vue-render.js
ang-zeyu: > If not, do you think this implementation (at the cost of longer first-content paint time) is alright until we can implement the SSR?

That's subjective, but I think I would personally prefer having some content show up first instead of a blank white screen with a spinning wheel so I know my browser hasn't hanged. =P That said, first-content-paint (not tti) is one of the very common [metrics](https://web.dev/first-contentful-paint/).

But more importantly, no hurry to fix FOUC completely. Let's fix this in proper when implementing SSR, otherwise it would be removed then anyway =P

packages/core/src/Page/index.js
ang-zeyu: the `_site` output folder should be polluted, which the user shouldn't be concerned with, not the user's source file directories

wxwxwxwx9: Ah right my bad! I misinterpreted the comment and was confused for a bit.

packages/cli/test/functional/test_site_special_tags/expected/markbind/js/initAppNode.page-vue-render.js
wxwxwxwx9: Understood! I'll remove the script here and look to work on the SSR :-)

wxwxwxwx9:
@ang-zeyu I have made the changes as per your suggestions :-) let me know if there's anything else I should work on!

packages/core-web/src/index.js
```diff
@@ -134,11 +134,15 @@ function removeTemporaryStyles() {
   jQuery('.temp-dropdown-placeholder').remove();
 }
 
-function executeAfterCreatedRoutines() {
-  removeTemporaryStyles();
+// Changes 'style-bypass-vue-compilation' tags back into 'style' tags
+function restoreStyleTags() {
+  jQuery('style-bypass-vue-compilation')
+    .replaceWith(`<style>${jQuery('style-bypass-vue-compilation').html()}</style>`);
```
ang-zeyu, line 140: hmm this may not work if the `<style>` tag contains conflicting html syntax

packages/core-web/src/index.js
```diff
@@ -161,9 +165,12 @@ function setup() {
   // eslint-disable-next-line no-unused-vars
   const vm = new Vue({
     el: '#app',
-    created() {
-      executeAfterCreatedRoutines();
+    render(createElement) {
+      // eslint-disable-next-line no-undef
```
ang-zeyu, line 169: how about `/* global ... */`? https://eslint.org/docs/rules/no-undef

wxwxwxwx9: Thanks for the tip! I've updated it accordingly :-)

packages/core-web/src/index.js
wxwxwxwx9: As per our discussion, the problem here is that `<style>` tag has a special parsing treatment, akin to `<script>` tag. Thus, if I simply change `<style>` to `<style-bypass-vue-compilation>` the block of code within the tag will be parsed as a normal html and there's a chance there may be invalid html syntax. This will affect the parsing of the page when I compile it with `vue-template-compiler`. 

To workaround it, I've piggybacked on the idea we used for `<script>` tags (`type="application/javascript"`) so that we can get the Vue compiler to recognize it as a script and ignore it during compilation (without discarding it). To make sure that it does not get executed as a script, I've added `src` and `defer` attributes to delay the execution of the script (as it is not meant to be an actual script, it was simply transformed into script tag to bypass Vue compilation), until our MarkBind plugin can restore it back into a proper `<style>` tag. 

I believe the above workaround should resolve the problem you mentioned. I've played around with it and so far it seems to be working fine (browser treatment of this script tag seems to be alright). 

Another alternative would be to transform the `<style>` tag to `<textarea>` tag. It will give us the same behaviour as above but the html syntax contained within the tag will be escaped instead.

packages/core/src/Page/index.js
```diff
@@ -482,6 +486,46 @@ class Page {
...
 
+  /**
+   * Compiles page into Vue Application to get the page render function and places
+   * it into a script so that the browser can retrieve the page render function to
+   * render the page.
+   *
+   * This is to avoid the overhead of compiling the page into Vue application
+   * on the client's browser (alleviates FOUC).
+   *
+   * By pre-compiling the page, we can get the Vue render function for the page
```
ang-zeyu, line 497: is this a leftover from the earlier one-huge-file? seems like a repeat of the first two sentences.

packages/core/src/html/NodeProcessor.js
```diff
@@ -667,9 +667,44 @@ class NodeProcessor {
...
+   *
+   * We work around this problem by piggybacking on the approach we used for <script> tags to be ignored
+   * and not discarded; by changing <style> tags to the placeholder tag as shown above.
+   * We will change it back to <style> tag after Vue compilation and after the element is created by Vue.
+   */
+  static processStyleTag(node) {
+    node.name = 'script';
+    /*
+      This node is not intended to be ran as a script. Thus, we have to defer the execution of the script
+      until the node is restored as a style node (which is handled by MarkBind's script plugin).
```
ang-zeyu, line 696:
```suggestion
      until the node is restored as a style node (which is handled by MarkBind's script).
```
plugin? 👀

packages/core/src/html/NodeProcessor.js
```diff
@@ -667,9 +667,44 @@ class NodeProcessor {
   }
 
   /*
-   * API
+   * Adds attribute "type=application/javascript" to <script> tag.
```
ang-zeyu, line 670: Let's move these related functionalities to a separate file

ang-zeyu (commented):
Thanks for the workaround 👍 

Last few nits:

packages/core/src/html/NodeProcessor.js
wxwxwxwx9: I have shifted the two functions to another file but I'm not sure what would be an appropriate name for the new file. I've named it literally as `scriptAndStyleTagProcessor.js` for now which can be quite 'limiting' but I think it should work for now 😅

packages/cli/test/functional/test_site_special_tags/expected/index.html
```diff
@@ -31,19 +32,19 @@ <h1 id="functional-test-for-htmlparser2-and-markdown-it-patches-for-special-tags
...
       x += 1;
       alert(x);
     </script>
     <hr>
     <p>There should be no text between this and the next <code class="hljs inline no-lang" v-pre>&lt;hr&gt;</code> tag in the browser, since it is a <code class="hljs inline no-lang" v-pre>&lt;style&gt;</code> tag.</p>
-    <style>
-      .bigger-text {
-        font-size: 10em;
+    <script src defer type="application/javascript" style-bypass-vue-compilation>
+      .bigger - text {
```
ang-zeyu, line 44: 👀 

sorry, just noticed this as well. Could you investigate it? @wxwxwxwx9

wxwxwxwx9: Hmm, I noticed this as well but I figured it shouldn't be a cause of concern, since we are not running the script and the conversion back to <style> tag doesn't seem to have any problem. Do you think it may cause a problem?

ang-zeyu: how about just js-beautify's `content_unformatted` option? though this means blanket ignoring all script tags which should be ok

wxwxwxwx9: That should work fine as well, but do you think we should put a notice for it in `https://markbind.org/userGuide/siteJsonFile.html#disablehtmlbeautify` accordingly? That `<script>` tags are not beautified by default.

ang-zeyu: let's keep it in mind for now.

beautifying might not even make sense after ssr with the "gibberish" output (current purpose is to allow the author to figure out markdown nuances / etc. causing unexpected outputs)

ang-zeyu (approved):
Lgtm :+1:

Issue #48, opened 2017-05-01
Support live preview for new entries to sites.json


damithc:
Adding the following to sites.json required a restart of the server to see the page `http://127.0.0.1:8080/site-demo-cs2103/book/common/References.html`.

```json
    {
      "src": "book/common/References.md",
      "title": "CS2103/T Textbook - References"
    }
```

damithc:
@alyip98 can you check if this issue is still applicable?

ang-zeyu:
Very much still applicable

PR #1514, opened 2021-03-13
Implement live preview for pages in site.json

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Addresses part of #48.

This PR intends to support live preview for the `pages` property in `site.json`. When `site.json` is edited, `reloadSiteConfig` function will read in the new values and check for changes in `pages`. If there is added/removed pages, all pages will be rebuilt. This implementation is similar to setting the glob to track all `.md` and `.mbd` files in `pages` (`"glob": ["**/index.md", "**/*.+(md|mbd)"]`). If new pages are added/removed, all pages will be rebuilt in the [`addHandler`](https://github.com/MarkBind/markbind/blob/master/packages/cli/index.js#L144).

If the attributes of a page are edited, then only that specific page/pages dependent on this will be rebuilt. For example, if the `"title"` of a page at `contents/index.md` is changed, then only `contents/index.html` will be regenerated. The rest of the pages will not be rebuilt.

Current implementation will rebuild all pages when there is a change in `pages`. ~Not sure if I should only 'watch' for added/removed pages and then only update the relevant pages that will be affected.~ 

EDIT: After looking around at the other handlers ([`addHandler`](https://github.com/MarkBind/markbind/blob/master/packages/cli/index.js#L140) and [`removeHandler`](https://github.com/MarkBind/markbind/blob/master/packages/cli/index.js#L164)), it seems all pages should be rebuilt if the added/removed file is a page/dependency of page.

**Testing instructions:**
1. Serve a MarkBind site using `markbind serve`
2. Open and edit the `pages` property in `site.json`
3. Live preview should be updated to reflect he latest changes

Implement live preview for pages in site.json

When pages are added or removed from site.json, the respective 
generated pages are not built or removed from the live preview. 
Similarly, edited pages attributes are not reflected as well. A server 
reboot is necessary to display the new changes.

Let's support live preview for pages in site.json so that the user 
does not have to restart the server each time the pages property 
is edited.

---

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -18,6 +18,7 @@ const {
   INDEX_MARKDOWN_FILE,
   INDEX_MARKBIND_FILE,
   LAZY_LOADING_SITE_FILE_NAME,
+  SITE_CONFIG_NAME,
 } = require('@markbind/core/src/Site/constants');
 
 const cliUtil = require('./src/util/cliUtil');
@@ -152,8 +153,8 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (filePath.replace(/^.*[\\\/]/, '') === "site.json") {
-          return site.reloadSiteConfig();
+        if (filePath.replace(/^.*[\\\/]/, '') === SITE_CONFIG_NAME) {
+          return site.reloadSiteConfig(filePath);
         }
         if (site.isDependencyOfPage(filePath)) {
           return site.rebuildAffectedSourceFiles(filePath);
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -152,6 +152,9 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
+        if (filePath.replace(/^.*[\\\/]/, '') === "site.json") {
+          return site.reloadSiteConfig();
+        }
         if (site.isDependencyOfPage(filePath)) {
           return site.rebuildAffectedSourceFiles(filePath);
         }
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -153,7 +153,7 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (filePath.replace(/^.*[\\\/]/, '') === SITE_CONFIG_NAME) {
+        if (filePath.replace(/^.*[\\/]/, '') === SITE_CONFIG_NAME) {
           return site.reloadSiteConfig(filePath);
         }
         if (site.isDependencyOfPage(filePath)) {
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -154,7 +154,7 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
         if (filePath.replace(/^.*[\\/]/, '') === SITE_CONFIG_NAME) {
-          return site.reloadSiteConfig(filePath);
+          return site.reloadSiteConfig();
         }
         if (site.isDependencyOfPage(filePath)) {
           return site.rebuildAffectedSourceFiles(filePath);
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -153,7 +153,7 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
-        if (filePath.replace(/^.*[\\/]/, '') === SITE_CONFIG_NAME) {
+        if (path.basename(filePath) === SITE_CONFIG_NAME) {
           return site.reloadSiteConfig();
         }
         if (site.isDependencyOfPage(filePath)) {
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -823,7 +823,7 @@ class Site {
     const oldAddressablePages = this.addressablePages.slice();
     const oldPagesSrc = oldAddressablePages.map(page => page.src);
     await this.readSiteConfig();
-    this.updateAddressablePages();
+    this.collectAddressablePages();
     // Get pages with edited attributes but with the same src
     const editedPages = _.differenceWith(this.addressablePages, oldAddressablePages, (newPage, oldPage) => {
       if (!_.isEqual(newPage, oldPage)) {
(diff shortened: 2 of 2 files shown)


packages/cli/index.js
```diff
@@ -152,6 +153,9 @@ program
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       Promise.resolve('').then(() => {
+        if (filePath.replace(/^.*[\\/]/, '') === SITE_CONFIG_NAME) {
```
ryoarmanda, line 156: Did you intend to get the filename from the changed file's path? Maybe can do it with `path.basename()`

jonahtanjz: Yup this is intended to get the filename and `path.basename` will be a better approach. Thanks for the catch! :)

packages/core/src/Site/index.js
```diff
@@ -816,6 +819,59 @@ class Site {
     }
   }
 
+  async reloadSiteConfig() {
+    const oldAddressablePages = this.addressablePages.slice();
+    const oldPagesSrc = oldAddressablePages.map(page => page.src);
+    await this.readSiteConfig();
+    this.updateAddressablePages();
```
ang-zeyu, line 826:
```suggestion
    this.collectAddressablePages();
```
would this do?

packages/core/src/Site/index.js
```diff
@@ -816,6 +819,59 @@ class Site {
...
+  }
+
+  /**
+   * Creates new pages and replaces the original pages with the updated version
+   */
+  updatePages(pagesToUpdate) {
+    pagesToUpdate.forEach((pageToUpdate) => {
+      this.pages.forEach((page, index) => {
+        if (page.pageConfig.src === pageToUpdate.src) {
+          const newPage = this.createPage({
```
ang-zeyu, line 859: how about something like a `mapAddressablePageToPage` function to reduce duplication with `mapAddressablePagesToPages` (or if could generalise the latter even better)

packages/core/src/Site/index.js
```diff
@@ -816,6 +819,59 @@ class Site {
...
+
+    // Comparator for the _differenceWith comparison below
+    const isNewPage = (newPage, oldPage) => _.isEqual(newPage, oldPage) || newPage.src === oldPage.src;
+
+    const addedPages = _.differenceWith(this.addressablePages, oldAddressablePages, isNewPage);
+    const removedPages = _.differenceWith(oldAddressablePages, this.addressablePages, isNewPage)
+      .map(filePath => Site.setExtension(filePath.src, '.html'));
+    if (!_.isEmpty(addedPages) || !_.isEmpty(removedPages)) {
+      await this.removeAsset(removedPages);
+      await this.rebuildSourceFiles();
```
ang-zeyu, line 843: let's use the non delayed version `this._rebuildSourceFiles();` below too for `rebuildAffectedSourceFiles`

packages/core/src/Site/index.js
```diff
@@ -816,6 +819,59 @@ class Site {
...
+      await this.writeSiteData();
+    } else {
+      this.updatePages(editedPages);
+      const siteConfigDirectory = path.dirname(path.join(this.rootPath, this.siteConfigPath));
+      this.rebuildAffectedSourceFiles(editedPages.map(page => path.join(siteConfigDirectory, page.src)));
```
ang-zeyu, line 848: hmm... would it be correct to instead use `regenerateAffectedPages` (in `_rebuildAffectedSourceFiles`) directly? (or something even lower level)

The `filePaths` parameter of `rebuildAffectedSourceFiles` indicates changed source files, but over here its the page configuration that changed, not the file. So the extra fluff (e.g. layout building) in `_rebuildAffectedSourceFiles` shouldn't be needed, and even with `regenerateAffectedPages` there's still quite a bit of fluff (e.g. change in a page's properties wouldn't affect other pages even if said page's source file is a dependency of those other pages).

Only exactly the pages which's properties have changed in `site.json` should be rebuilt. (and even then if using `markbind s -o` only the page being viewed should be rebuilt)

If this needs a substantial addition here that's not worth it, we can do it after enhancing the rebuild process for page addition / removal (above comment) + #1513 though, then streamlining the overall flow, which should make this easier; also since this is a feature addition.

ang-zeyu (commented):
Thanks @jonahtanjz 

Some nits:

---
> EDIT: After looking around at the other handlers (addHandler and removeHandler), it seems all pages should be rebuilt if the added/removed file is a page/dependency of page.

This has been existing behaviour for a long while already but can probably be made into an issue 🙂

packages/core/src/Site/index.js
jonahtanjz: > hmm... would it be correct to instead use regenerateAffectedPages (in _rebuildAffectedSourceFiles) directly? (or something even lower level)

> The filePaths parameter of rebuildAffectedSourceFiles indicates changed source files, but over here its the page configuration that changed, not the file. So the extra fluff (e.g. layout building) in _rebuildAffectedSourceFiles shouldn't be needed, and even with regenerateAffectedPages there's still quite a bit of fluff (e.g. change in a page's properties wouldn't affect other pages even if said page's source file is a dependency of those other pages).

Yup agree that changing page properties wouldn't affect other pages and they should not be rebuilt. I will look into this and perhaps open another PR to work on it as it seems like it will take some time and will require quite a bit of change. For now I've switched to using `regenerateAffectedPages` as suggested.

packages/core/src/Site/index.js
jonahtanjz: Alright I've updated!

packages/core/src/Site/index.js
jonahtanjz: Hmm I tried to combine them but it seems like their logic is quite different. As a workaround, I've extracted out the `createPage` part into a separate function so that both `mapAddressablePagesToPages` and this can call the new function which resolves the repetition issue. Is this ok?

packages/core/src/Site/index.js
jonahtanjz: Alright updated :)

packages/core/src/Site/index.js
```diff
@@ -816,6 +819,51 @@ class Site {
     }
   }
 
+  async reloadSiteConfig() {
+    const oldAddressablePages = this.addressablePages.slice();
+    const oldPagesSrc = oldAddressablePages.map(page => page.src);
+    await this.readSiteConfig();
+    this.collectAddressablePages();
+    // Get pages with edited attributes but with the same src
```
ang-zeyu, line 827: could this be scoped to the `else` branch?

ang-zeyu (commented):
Last 2 nits, generally looks good to go 👍 

Please hold on while I merge #1513 in first; There shouldn't be any conflicts but would be good to test this on top of that.

packages/core/src/Site/index.js
jonahtanjz: Yup this would be better since `editedPages` is only used inside the `else` branch.

packages/core/src/Site/index.js
```diff
@@ -938,7 +995,6 @@ class Site {
 
     this._setTimestampVariable();
     this.mapAddressablePagesToPages(addressablePages, faviconUrl);
-
```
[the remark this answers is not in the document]
jonahtanjz: Oops must have accidentally removed a line when browsing 😬

jonahtanjz:
@ang-zeyu I have updated the PR accordingly.

> Please hold on while I merge #1513 in first; There shouldn't be any conflicts but would be good to test this on top of that.

Alright sounds good :)

ang-zeyu (approved):
Lgtm 👍

Issue #1495, opened 2021-02-28
Some more refactoring ideas


ang-zeyu:
**Is your request related to a problem?**

- [ ] Combine page nav and search heading indexing into one inside `NodeProcessor`, possibly managed in-memory by some `Indexer` abstraction of sorts (d.moderate)
   - both contains very similar duplicate logic (e.g. don't index unexpanded panel's content, but index its header)
   - the latter is performance hungry, requiring re-reading and parsing html files of `External`s
- [ ] Split up `NodeProcessor` further into separate files (d.easy)
- [ ] Separate site conversion and live reload logic from Site.js (d.easy)
- [ ] PageConfig has a lot of duplicate properties with `SiteConfig` - pass down and use site config where appropriate (d.easy)

raysonkoh:
Just to clarify for the second point regarding splitting up `NodeProcessor` into separate files, I am thinking of separating the methods that are specific to the various components such as `_processQOption`, `_processQuiz`, `_processTooltip` to a separate file (maybe `componentProcessor`?). @ang-zeyu Is this what you meant? 😅

ang-zeyu:
> Just to clarify for the second point regarding splitting up `NodeProcessor` into separate files, I am thinking of separating the methods that are specific to the various components such as `_processQOption`, `_processQuiz`, `_processTooltip` to a separate file (maybe `componentProcessor`?). @ang-zeyu Is this what you meant? 😅

@raysonkoh sorry for the overly brief description. I was thinking it might be more cohesive to group things by functionality here, so you know right away what a file contains only. (quicker to grasp instead of going through each component file and figuring what goes on for each component)

E.g. iirc, some main functions left are:
- rendering markdown-in-attributes into slots
- combining / shifting footnotes from included files into one giant chunk at the end
- heading id assignment
- bootstrap vue transformations for modals, popovers, tooltips
- deprecation warnings (**might be able to resolve this completely in https://github.com/MarkBind/markbind/issues/1504 first though)
- ...

Organising by components might be ok too though, also some components / elements only have 1 functionality - if it turns out better let's go for that.

tlylt:
Additional refactoring ideas putting here for discussion:
- Consolidate frontmatter handling to its own class/file
- Consolidate asset handling to its own class/file

tlylt:
For Pt 3, 
>  Separate site conversion (d.easy) and live reload logic (d.moderate) from Site.js

[...]

Issue #1504, opened 2021-03-04
Fully deprecate some component attributes


wxwxwxwx9:
While working on some issues related to dropdown, I realised that we are still processing the `text` attribute for dropdown. However, the `text` attribute is currently stated to be deprecated in the user guide. For code maintenance purposes, should we fully remove any processing code related to `text` attribute?

ang-zeyu:
@damithc should we deprecate these attributes for v3 as well?

If its too much changes could postpone https://github.com/MarkBind/markbind/pull/975

> dropdowns - "text" -> "header"
> boxes - "heading" -> "header"
> popovers - "title" -> "header"
> modals - "title" -> "header"

damithc:
> @damithc should we deprecate these attributes for v3 as well?

Sure. Does that mean there will be a warning message or those will be ignored altogether?

ang-zeyu:
Ignored altogether

damithc:
The current version gives a warning already, correct? In that case should be fine as I've dealt with those warnings already.

ang-zeyu:
> The current version gives a warning already, correct? In that case should be fine as I've dealt with those warnings already.

yup

ang-zeyu:
For context:  V3 has deprecated these attributes on the changelog front but the implementation remains.

We should remove the implementation completely for V4

PR #1515, opened 2021-03-16
NodeProcessor refactor: split into separate files

raysonkoh:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

#1495 2nd task.

**Overview of changes:**
Extracted some methods of `NodeProcessor` into the following files:
- BootstrapVueProcessor: Responsible for bootstrap vue transformations for modals, popovers, tooltips
- ComponentProcessor: Responsible for processing nodes of various Markbind components such as panels, dropdowns, tabs, etc
- FootnoteProcessor
- headerProcessor: header-related helper functions
- MarkdownProcessor: Extracted out `_renderMd` and `_renderMdInline` and `_processAttributeWithoutOverride` + bookkeeping of `docId`
- vueSlotSyntaxProcessor: functions related to "transforming Vue old slot syntax to the new syntax"
- warnings: contains functions that does logs warnings

- Whether there are any other mtds in NodeProcessor that should be extracted out into a separate file.

**Testing instructions:**
`npm run test`

```
NodeProcessor refactor: split into separate files
```

---


wxwxwxwx9:
Good job on this PR! @raysonkoh Just some preliminary comments. 
- I think you forgot to remove `_transformSlottedComponent` from `NodeProcessor`. 
- How about shifting the functions related to "transforming Vue old slot syntax to the new syntax" to another file as well? Looking at `getVslotShorthandName`, `transformOldSlotSyntax`, and `shiftSlotNodeDeeper`. Maybe the file can be called `vueSlotSyntaxProcessor`?
- There seems to be quite tight coupling (cyclic dependency) due to the code below between `NodeProcessor` and `BootstrapVueProcessor` and `ComponentProcessor` respectively. 
```
    this.bootstrapVueProcessor = new BootstrapVueProcessor(this);
    this.componentProcessor = new ComponentProcessor(this);
```

Seems like the main reason is because of the need for `_processAttributeWithoutOverride` which is contained in `NodeProcessor`. I'm thinking if we should abstract this function out from `NodeProcessor` into a separate file? But I'm not sure if it would fit into any of the existing files or if we would have to create a new file for it (which will only contain this function). 

Also, the deprecation-warning functions are also contributing to the coupling issue. Should we shift them to its own file as well? 

Let me know what you think :-)

raysonkoh:
Thanks for the review @wxwxwxwx9!
> I think you forgot to remove _transformSlottedComponent from NodeProcessor.

Good catch!

> How about shifting the functions related to "transforming Vue old slot syntax to the new syntax" to another file as well? Looking at getVslotShorthandName, transformOldSlotSyntax, and shiftSlotNodeDeeper. Maybe the file can be called vueSlotSyntaxProcessor?

Yes it seems like these functions can be refactored out since they are stateless. I will try to do that.

> Seems like the main reason is because of the need for `_processAttributeWithoutOverride` which is contained in `NodeProcessor`. I'm thinking if we should abstract this function out from `NodeProcessor` into a separate file? But I'm not sure if it would fit into any of the existing files or if we would have to create a new file for it (which will only contain this function).

Yes I was trying to decouple them, I think it is possible to do this if I update `renderMd` to take in the `docId`, so the one that is responsible for the bookkeeping of `docId` would be `NodeProcessor`. In that case, `MarkdownRenderer` need not be a class since it wouldn't be storing `docId` anymore and I can update `_processAttributeWithoutOverride` to take in the node as well as the `docId`.

> Also, the deprecation-warning functions are also contributing to the coupling issue. Should we shift them to its own file as well?

Good idea! I didn't shift them to a separate file because I initially thought that they were just temporary functions (until the next major version upgrade ?) but now it seems cleaner if they are all in a separate file since they don't really modify / use the state of `NodeProcessor`

wxwxwxwx9:
Thanks for the effort @raysonkoh! Everything looks cleaner and more easy to understand. `NodeProcessor` looks really clean :-) Just a little more subjective organization suggestions to organize the codebase more (let me know what you think). 

It seems like `docIdManager` is really only used for the functions contained in `markdownProcessor` (e.g. `renderMd`, `renderMdInline`, `processAttributeWithoutOverride`). 

Thus, I was thinking that instead of passing down `docIdManager` to the various processing components (e.g. `BootstrapVueProcessor`, `ComponentProcessor`), and then passing it down to the `markdownProcessor` functions that are used in the aforementioned processing components, how about we make `markdownProcessor` into a class and instantiate it in `NodeProcessor`? Much like how it is currently done for the other processing components like `BootstrapVueProcessor`. Then we can just pass in `docIdManager` when instantiating `markdownProcessor` just one time. And in turn, we will pass in `markdownProcessor` to the other processing components that need it (e.g. `BootstrapVueProcessor`, `ComponentProcessor`). 

Along the same line, I was thinking we can do this for the `warning` logger and `vueSlotSyntaxProcessor` as well -- instantiating classes for them and passing them down to the processing components that need them. I think one benefit this will provide is that when we look at the constructor of `NodeProcessor`, we can get a good overview of how the different processing components interact with one another (gives us a good idea of the inter-dependencies). This will help readers to understand the codebase with more ease in my opinion. For example, it would be immediately clear that `BootstrapVueProcessor` does some form of `md` processing. 

Just to illustrate my point, the below code currently shows the functions imported by `BootstrapVueProcessor`

```
const { processAttributeWithoutOverride } = require('./markdownProcessor');
const { getVslotShorthandName } = require('./vueSlotSyntaxProcessor');
const { warnDeprecatedAttributes, warnDeprecatedSlotNames } = require('./warnings');

/**
 * Class responsible for bootstrap vue transformations for modals, popovers, tooltips
 */
class BootstrapVueProcessor {
  constructor(docIdManager) {
    this.docIdManager = docIdManager;
  }
  ....
}
```

The modified code may look something like below, where `markdownProcessor` has already taken in the instance of `docIdManager`. 

```
/**
 * Class responsible for bootstrap vue transformations for modals, popovers, tooltips
 */
class BootstrapVueProcessor {
  constructor(markdownProcessor, vueSlotSyntaxProcessor, warningLogger) {
    this.docIdManager = docIdManager;
  }
  ....
}
```

Although I'm not sure if it would be a little weird to use `vueSlotSyntaxProcessor` as a class. Let me know what you think about this! :-)

packages/core/src/html/warnings.js
```diff
@@ -0,0 +1,79 @@
...
+        });
+    }
+  });
+}
+
+/*
+ * Body
+ */
+
+function preprocessBody(node) {
```
wxwxwxwx9, line 69: I'm thinking we can rename this into something more appropriate. Maybe something like `warnBodyTag`?

raysonkoh:
Thanks for the review @wxwxwxwx9! 😄

> It seems like `docIdManager` is really only used for the functions contained in `markdownProcessor` (e.g. `renderMd`, `renderMdInline`, `processAttributeWithoutOverride`).
> 
> Thus, I was thinking that instead of passing down `docIdManager` to the various processing components (e.g. `BootstrapVueProcessor`, `ComponentProcessor`), and then passing it down to the `markdownProcessor` functions that are used in the aforementioned processing components, how about we make `markdownProcessor` into a class and instantiate it in `NodeProcessor`?
> 
> Much like how it is currently done for the other processing components like `BootstrapVueProcessor`. Then we can just pass in `docIdManager` when instantiating `markdownProcessor` just one time. And in turn, we will pass in `markdownProcessor` to the other processing components that need it (e.g. `BootstrapVueProcessor`, `ComponentProcessor`).

Yes, I think this may work. I can integrate `docId` into a wrapper class for rendering markdown. Thanks for the suggestion!

> Along the same line, I was thinking we can do this for the `warning` logger and `vueSlotSyntaxProcessor` as well -- instantiating classes for them and passing them down to the processing components that need them.

Correct me if I'm wrong but I think the `warning` logger does not need to be passed down to the processing components as it does not carry any state. In that case, I think it may be better to just import it from the `utils` folder instead of passing the logger around to other components.

For `vueSlotSyntaxProcessor` the methods that are currently in it does not contain any state too. I can make it into a class too, but the all of its method will have to be made static. Also, making it into a class may seem out of place with respect to the other files that only contain functions like `codeBlockProcessor` and `includePanelProcessor`.

packages/core/src/html/warnings.js
raysonkoh: Yes this sounds more in-line with its functionality.

wxwxwxwx9:
> Correct me if I'm wrong but I think the warning logger does not need to be passed down to the processing components as it does not carry any state. In that case, I think it may be better to just import it from the utils folder instead of passing the logger around to other components.

> For vueSlotSyntaxProcessor the methods that are currently in it does not contain any state too. I can make it into a class too, but the all of its method will have to be made static. Also, making it into a class may seem out of place with respect to the other files that only contain functions like codeBlockProcessor and includePanelProcessor.

@raysonkoh Yup, those are my doubts as well. It's a little awkward to make them into classes, so I was unsure if it's a good idea after all. Considering the points you brought up above, I think let's discard that idea. Thanks for your input! :-)

Anyway, the latest changes look good to me! 👍

packages/core/src/html/ComponentProcessor.js
```diff
@@ -0,0 +1,162 @@
...
+    this.markdownProcessor.processAttributeWithoutOverride(node, 'header', false);
+  }
+
+  /**
+   * Assigns an id to the root element of a panel component using the heading specified in the
+   * panel's header slot or attribute (if any), with the header slot having priority if present.
+   * This is to ensure anchors still work when panels are in their minimized form.
+   * @param node The root panel element
+   */
+  static assignPanelId(node) {
```
ang-zeyu, line 32: how about `headerProcessor.js`? seems closely related

packages/core/src/html/MarkdownProcessor.js
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+ * Processes the markdown attribute of the provided element, inserting the corresponding <slot> child
+ * if there is no pre-existing slot child with the name of the attribute present.
+ * @param node Element to process
+ * @param attribute Attribute name to process
+ * @param isInline Whether to process the attribute with only inline markdown-it rules
+ * @param slotName Name attribute of the <slot> element to insert, which defaults to the attribute name
+ */
+  processAttributeWithoutOverride(node, attribute, isInline, slotName = attribute) {
```
ang-zeyu, line 36: 👀 

`ComponentProcessor.js`? since its related to rendering markdown-in-attributes

packages/core/src/html/ComponentProcessor.js
```diff
@@ -0,0 +1,162 @@
...
+const { findHeaderElement } = require('./headerProcessor');
+const { getVslotShorthandName } = require('./vueSlotSyntaxProcessor');
+const { warnConflictingAttributes, warnDeprecatedAttributes } = require('./warnings');
+
+const _ = {};
+_.has = require('lodash/has');
+
+const logger = require('../utils/logger');
+
+class ComponentProcessor {
```
ang-zeyu, line 12: Hmm should we split this up further? Namely
- conflicting attribute logging
- deprecated attribute logging (can be fully removed in a prior pr also though, if easier https://github.com/MarkBind/markbind/issues/1504)
- the katex one can be further split with e.g. the one for `<pre>` (and any others) into something like `vPreAdder.js` (or just in the switch-case since its pretty short)

So this file / class would tackle just rendering markdown-in-attributes

packages/core/src/html/BootstrapVueProcessor.js
```diff
@@ -0,0 +1,167 @@
...
+    if (_.has(node.attribs, originalAttribute)) {
+      node.attribs[newAttribute] = node.attribs[originalAttribute];
+      delete node.attribs[originalAttribute];
+    }
+  }
+
+  processPopover(node) {
+    warnDeprecatedAttributes(node, { title: 'header' });
+
+    this.markdownProcessor.processAttributeWithoutOverride(node, 'content', true);
```
ang-zeyu, line 85: could move these to `ComponentProcessor` so this file actually only handles the bootstrap vue transformations; similarly for the warnings

ang-zeyu (commented):
Thanks! looks really clean now. Couple of suggestions:

packages/core/src/html/ComponentProcessor.js
ang-zeyu: to further add, my pov is that `NodeProcessor` is still and should be the one doing the 'organizing by elements', so it may make more sense to tie up / order various functionalities required by those elements there anyway. (instead of pushing it one level deeper down to somewhere like `BootstrapVueProcessor`)

packages/core/src/html/ComponentProcessor.js
raysonkoh: Yup, makes sense to be in `headerProcessor.js`

packages/core/src/html/MarkdownProcessor.js
raysonkoh: I have moved it to `ComponentProcessor.js`, the only modifications for this to work is to add `processTrigger`, `processPopover` and `processModalAttributes` in `ComponentProcessor` since they are calling `processAttributeWithoutOverride`. I think this is better also to separate the different functionalities (rendering markdown-in-attributes and doing the bootstrap-vue transformations).

packages/core/src/html/ComponentProcessor.js
raysonkoh: > conflicting attribute logging
deprecated attribute logging (can be fully removed in a prior pr also though, if easier #1504)

I have created a mapping of `node.name` to functions that does the warnings for conflicting attributes and deprecated attributes in `warnConflictingAtributesMap` and `warnDeprecatedAtributesMap` respectively in `warning.js`. And the various warnings are called before the switch-case statements of `NodeProcessor#processNode`. I thought that this is quite simple and easy for someone to easily add / remove warnings for various components via the mapping. What are your thoughts on this?

> the katex one can be further split with e.g. the one for `<pre>` (and any others) into something like vPreAdder.js (or just in the switch-case since its pretty short)

Since the logic is pretty short, I have added it in the switch-case.

> to further add, my pov is that NodeProcessor is still and should be the one doing the 'organizing by elements', so it may make more sense to tie up / order various functionalities required by those elements there anyway. (instead of pushing it one level deeper down to somewhere like BootstrapVueProcessor)

I may need some clarifications on this. Based on what I understand, you are saying `NodeProcessor` should have some utility functions which are used by all elements? 

From `BootstrapVueProcessor`, I have moved `_renameSlot` and `_renameAttributes` to `NodeProcessor` since it is quite general and not really specific to the functionality of `BootstrapVueProcessor`. However, I feel that it be may better for certain functions such as `_addTriggerClass` and `_transformSlottedComponents` to be in `BootstrapVueProcessor` as they are helper functions for performing the bootstrap-vue transformations.  Do let me know if I am misunderstanding this.

raysonkoh:
Based on the latest changes, I have modified the functionality of the following files:

- `ComponentProcessor.js`: responsible for rendering of markdown-in-attributes only
- `bootstrapVueProcessor.js`: responsible for the bootstrap-vue transformation logic for popovers, tooltip and modals only
- `warnings.js`: responsible for logging the warnings for both conflicting and deprecated attributes for the various components only

packages/core/src/html/BootstrapVueProcessor.js
raysonkoh: Okay, I have moved the warnings to `warning.js` and the rendering of markdown-in-attributes to `ComponentProcessor`

packages/core/src/html/ComponentProcessor.js
ang-zeyu: > What are your thoughts on this?

Looks good, even cleaner now. 👍 

> I may need some clarifications on this. Based on what I understand, you are saying NodeProcessor should have some utility functions which are used by all elements?

Sorry, no action needed here. Just a remark on what I think `NodeProcessor` should be handling (consolidating and ordering functionalities from the separate files for each element)

> `ComponentProcessor`

Should we rename the class as such? Maybe `MdAttributeRenderer`? 😮

packages/core/src/html/NodeProcessor.js
```diff
@@ -597,141 +140,85 @@ class NodeProcessor {
...
         break;
       case 'modal':
-        this._processModalAttributes(node);
+        this.componentProcessor.processModalAttributes(node);
+        transformBootstrapVueModalAttributes(node,
+                                             (originalName, newName) =>
+                                               NodeProcessor._renameSlot(node, originalName, newName),
+                                             (originalAttribute, newAttribute) =>
+                                             // eslint-disable-next-line max-len
+                                               NodeProcessor._renameAttribute(node, originalAttribute, newAttribute));
```
ang-zeyu, line 198: could we move `_renameSlot` +`_renameAttribute` into `bootstrapvueprocessor`? since its only used there

packages/core/src/html/headerProcessor.js
```diff
@@ -0,0 +1,87 @@
...
+          + 'Please report this to the MarkBind developers. Thank you!');
+    }
+
+    node.attribs.id = header.attribs.id;
+  }
+}
+
+module.exports = {
+  setHeadingId,
+  findHeaderElement,
```
ang-zeyu, line 85: unused export

ang-zeyu (commented):
couple more nits:

packages/core/src/html/NodeProcessor.js
```diff
@@ -597,141 +140,85 @@ class NodeProcessor {
...
+        this.componentProcessor.processDropdownAttributes(node);
         break;
       case 'thumbnail':
-        this._processThumbnailAttributes(node);
+        this.componentProcessor.processThumbnailAttributes(node);
         break;
       case 'annotation':
-        NodeProcessor._processAnnotationAttributes(node);
+        // Annotations are added automatically by KaTeX when rendering math formulae.
+        if (!_.has(node.attribs, 'v-pre')) { node.attribs['v-pre'] = true; }
```
ang-zeyu, line 215: let's fallthrough with the one for `<code>`

packages/core/src/html/ComponentProcessor.js
raysonkoh: > Should we rename the class as such? Maybe `MdAttributeRenderer`? 😮

Yes I have no issues with this name ^_^

packages/core/src/html/NodeProcessor.js
raysonkoh: Okay I have moved the functions over

packages/core/src/html/headerProcessor.js
raysonkoh: Thanks for the catch!

packages/core/src/html/NodeProcessor.js
raysonkoh: Okay I have made the modifications for this

ang-zeyu (approved):
Lgtm 👍

PR #1521, opened 2021-03-27
Fix highlightjs warnings

### LINKED ISSUE #1518: Fix deprecation warnings for highlight.js v10.7
While serving a markbind site using `markbind serve`, I noticed in the outputs there are quite a lot of warnings relating to deprecation warning specifically for the `highlight(lang, code, ...args)` method of highlight.js. Ref:  highlightjs/highlight.js#2277

image

This might be quite annoying for the users, as showing more than one instance of the deprecation warning may not be useful and the user shouldn't be seeing such deprecation warnings as it is more relevant for the devs. Should we refactor the `highlight` method in our code to follow the new standard?

raysonkoh:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

Resolves #1518 

**Overview of changes:**

Context:
The current method of doing line-by-line highlighting is via the use of the deprecated `highlight` method which uses `(language, code, ignore_illegals, continuation)` as arguments. In the latest version, the new arguments are `(code, { language, ignoreIllegals})`, `continuation` is no longer being supported. In fact, in their docs, it was stated that it was not recommended to use `continuation` for line-by-line highlighting. Ref: https://highlightjs.readthedocs.io/en/latest/api.html#highlight-languagename-code-ignore-illegals-continuation

Changes made:
Instead of doing line-by-line highlighting, the entire block is highlighted first. Afterwards, the result will be split according t o `'\n'`, and each line will be transformed into `open tags + line + close tags`. The token inside the tags will be maintain by a stack `tokenStack`. More explanation can be found in the code comments.
This was inspired by the approach taken in  hexojs/hexo-util#22

Please let me know if there are any edge case that I might have missed 😅 

**Testing instructions:**
`npm run test`

```
Update highlightjs method call syntax

With highlightjs v10.7.0, the highlight API syntax has been 
modified. Specifically, the usage of continuation as an 
extra argument, which Markbind uses for line-by-line 
highlighting, has been deprecated.

Highlighting the entire code block first and adding 
the neccessary opening and closing tags for each 
line would enable line-by-line highlighting and also 
allow us to use the latest highlightjs API syntax.

Let's,
* update the highlight method to use the latest 
version's syntax
* update the logic for line-by-line highlighting to 
do highlighting of the entire code block first and do 
post-processing to add the neccessary opening 
and closing tags per line.
```

---


packages/core/src/lib/markdown-it/index.js
```diff
@@ -75,31 +75,77 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
   if (highlightLinesInput) {
     const highlightLines = highlightLinesInput.split(HIGHLIGHT_LINES_DELIMITER_REGEX);
     highlightRules = highlightLines
-      .map(ruleStr => HighlightRule.parseRule(ruleStr, -startFromZeroBased, lines))
+      .map(ruleStr => HighlightRule.parseRule(ruleStr, -startFromZeroBased, str.split('\n')))
```
ryoarmanda, line 78: Nit/micro-optimization: maybe store the `str.split('\n')` results before the `map` so that it does not repeat the split computation every time the callback function is applied.

packages/core/src/lib/markdown-it/index.js
```diff
@@ -75,31 +75,77 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
-      // state stores the current parse state of hljs, so that we can pass it on line by line
-      let state = null;
+      lines = hljs.highlight(str, { language: lang, ignoreIllegals: true }).value;
+      const tokenStack = [];
+
+      // remove <span class="hljs-section"> ... </span>
+      // as the whole section would be wrapped by html tags at the end
+      if (lines.includes('hljs-section')) {
+        lines = lines.slice(27);
+        lines = lines.slice(0, lines.length - 7);
```
ryoarmanda, line 129: Magic numbers here. For easier understanding, better to put the lengths of `<span class="hljs-section">` and its closing counterpart somewhere, not as a direct number, but derived from the length of the respective strings.

packages/core/src/lib/markdown-it/index.js
```diff
@@ -75,31 +75,77 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+
       lines = lines.map((line) => {
-        const highlightedLine = hljs.highlight(lang, line, true, state);
-        state = highlightedLine.top;
-        return highlightedLine.value;
+        const prepend = tokenStack.map(tok => `<span class="${tok}">`).join('');
+        const re = /(<span class="(.*?)">|<\/span>)/g; // match all (<span class="xyz"> and </span>)
+        let match = re.exec(line);
+        while (match !== null) {
+          if (match[0] === '</span>') {
```
ryoarmanda, line 138: Nit/readability: prefer to use array destructuring and save the value as a constant

packages/core/src/lib/markdown-it/index.js
```diff
@@ -75,31 +75,77 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+            // pop from stack
+            tokenStack.shift();
+          } else {
+            // push to stack
+            tokenStack.unshift(match[2]);
```
ryoarmanda, line 143: Also here, prefer to save value as constant

ryoarmanda (changes requested):
Thanks for researching and updating our highlight.js usage to the new API!. It's good that MarkBind can be on top of dependency updates.

Generally I feel the approach is quite solid, I understood your approach and workaround now that we can't supply the previous highlight state anymore. Looks like this approach is the one that's put forth by one of the library's maintainers [here](https://stackoverflow.com/questions/64280814/how-can-i-correctly-highlight-a-line-by-line-code-using-highlight-js-react).

I've left a couple of first-round nits and reviews here, will review further afterwards. Do also add new testcases that ensures that we parse multi-line tokens properly (like the one examplified in the StackOverflow answer above).

packages/core/src/lib/markdown-it/index.js
ryoarmanda: Additional: As your first capturing group is essentially your whole regex, I feel it's kind of redundant as your full-match results is equal to the first capturing group results.

packages/core/src/lib/markdown-it/index.js
```diff
@@ -75,31 +75,77 @@ markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
...
+       "-----",
+       "</span>",
+      ]
+
+      tokenStack maintains the visited tokens. For eg, for <span class="xyz">, the token would be "xyz".
+      If it encounters a </span>, it pops from the stack.
+      Else, it pushes the token to the stack.
+
+      At the end of each line, it will be prepended by the necessary
+      opening tags and appended by the neccessary closing tags.
```
ryoarmanda, line 113: The comment becomes a bit too long. Maybe can trim out this part and refer to the GitHub issue you're basing the approach on (or even the StackOverflow link in the discussion of that issue). Imo the comment should just give an example, the result after calling `highlight`, and the result after you post-process it.

(Aside, this could be a good design notes writeup)

raysonkoh:
Thanks for the review @ryoarmanda 🚀 ! I agree with the points that you have raised and have revised the PR accordingly.

I have removed the logic for slicing `<span class="hljs-section">` as I realised that it was an unintended behavior with the current method of line-by-line highlighting. Also I have added some tricky testcases in `testCodeBlocks.md`.

ryoarmanda (approved):
Good work on the tests! Checked the expected site and it looks like everything is handled properly. LGTM :+1:

PR #1522, opened 2021-03-28
Lazy serve: fix untracked tab bug on backwards navigation

ryoarmanda:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Bug caught in the implementation of PR #1513, where navigating backwards with the back button causes the multiple-tab development structure to not work properly, as in that case the page immediately undergoes socket establishment phase, skipping request-response phase altogether as previously assumed.

Reproduction is outlined in here https://github.com/MarkBind/markbind/pull/1513#issuecomment-803596210

**Overview of changes:**
- Moved active tab entry addition to socket establishment phase.
- `isLiveReloadRequest` is now unused and thus removed.
- Now `client` is only empty if it is cleared before the live-reload. The check on whether an entry is active or in reloading can now be tested with that property instead of `isReloading`. Thus, `isReloading` is removed.
- Also, there is no need to have `prevClient` to differentiate whether an entry is undergoing reload or still waiting for socket establishment. `prevClient` is also removed.
- Added extra guard on fs events so that opened pages list sync only happens on lazy serve mode.

**Testing instructions:**
Use `markbind serve -o`

To test the fix (using the `docs` folder as the source files, can try it in default template or any other with adaptation):
1. Navigate to another page in an already-open tab (e.g. from the `index.html` page automatically open on start)
2. Navigate back to the previous page with the back button in the browser
3. First, verify there are no errors thrown in the logs
4. Edit the source file of that page and save it
5. Verify that the updated entries of the opened pages list is reflected accurately in the logs.

Also try out the updated test cases outlined here https://github.com/MarkBind/markbind/pull/1522#issuecomment-808908369

Lazy serve: fix untracked tab bug on backwards navigation

---

- [x]  Pinged someone for a review!

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -140,11 +140,9 @@ program
 
     const addHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
-      if (onePagePath) {
-        logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
-        site.changeCurrentOpenedPages(normalizedActiveUrls);
-      }
+      logger.info('Synchronizing opened pages list before reload');
+      const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
+      site.changeCurrentOpenedPages(normalizedActiveUrls);
       Promise.resolve('').then(() => {
         if (site.isFilepathAPage(filePath) || site.isDependencyOfPage(filePath)) {
           return site.rebuildSourceFiles(filePath);
@@ -157,11 +155,9 @@ program
 
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
-      if (onePagePath) {
-        logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
-        site.changeCurrentOpenedPages(normalizedActiveUrls);
-      }
+      logger.info('Synchronizing opened pages list before reload');
+      const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
+      site.changeCurrentOpenedPages(normalizedActiveUrls);
       Promise.resolve('').then(() => {
         if (path.basename(filePath) === SITE_CONFIG_NAME) {
           return site.reloadSiteConfig();
@@ -177,11 +173,9 @@ program
 
     const removeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file deletion: ${filePath}`);
-      if (onePagePath) {
-        logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
-        site.changeCurrentOpenedPages(normalizedActiveUrls);
-      }
+      logger.info('Synchronizing opened pages list before reload');
+      const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
+      site.changeCurrentOpenedPages(normalizedActiveUrls);
       Promise.resolve('').then(() => {
         if (site.isFilepathAPage(filePath) || site.isDependencyOfPage(filePath)) {
           return site.rebuildSourceFiles(filePath);
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -142,10 +142,7 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       if (onePagePath) {
         logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
-          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
-          return fsUtil.removeExtension(completeUrl);
-        });
+        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
         site.changeCurrentOpenedPages(normalizedActiveUrls);
       }
       Promise.resolve('').then(() => {
@@ -162,10 +159,7 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       if (onePagePath) {
         logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
-          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
-          return fsUtil.removeExtension(completeUrl);
-        });
+        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
         site.changeCurrentOpenedPages(normalizedActiveUrls);
       }
       Promise.resolve('').then(() => {
@@ -185,10 +179,7 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file deletion: ${filePath}`);
       if (onePagePath) {
         logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
-          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
-          return fsUtil.removeExtension(completeUrl);
-        });
+        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
         site.changeCurrentOpenedPages(normalizedActiveUrls);
       }
       Promise.resolve('').then(() => {
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -142,7 +142,10 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file add: ${filePath}`);
       if (onePagePath) {
         logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
+        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
+          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
+          return fsUtil.removeExtension(completeUrl);
+        });
         site.changeCurrentOpenedPages(normalizedActiveUrls);
       }
       Promise.resolve('').then(() => {
@@ -159,7 +162,10 @@ program
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
       if (onePagePath) {
         logger.info('Synchronizing opened pages list before reload');
-        const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
+        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
+          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
+          return fsUtil.removeExtension(completeUrl);
+        });
         site.changeCurrentOpenedPages(normalizedActiveUrls);
       }
       Promise.resolve('').then(() => {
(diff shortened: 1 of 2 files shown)


ryoarmanda:
### Test cases

Important: MarkBind's opened pages list is only updated whenever there are additions/changes/removals in source files. So the effects of some cases will not be apparent in the logs, but internally, the changes in the active tabs are caught and handled appropriately by `live-server`.

No| Cases | Expected behaviour (MarkBind)
--|-------|-------------------
1 | Visits a page on a new tab | Opened pages list is only updated when the page is new/rebuilt.
2 | Visits a page that actually just a redirect to another page (e.g. our docs' `userGuide/index`) | Opened pages list is only updated when the page is new/rebuilt, but if so, it can update up to two times.
3 | Visits a page when the same page is also opened on another tab | Opened pages list is not affected.
4 | Reload a tab | Opened pages list is not affected.
5 | Close a tab | Opened pages list is not affected.
6 | Visits a page on an existing tab | Opened pages list is not affected.
7 | Edit a page | Opened pages list is synchronized with `live-server`'s active tabs before issuing rebuild.
8 | Navigate tab backwards/forwards | Opened pages list is not affected.

wxwxwxwx9:
Hi @ryoarmanda, thanks for working on this rather complex feature! It took me awhile to go through the original PR properly so that I have a better context of this problem. But I would just like to clarify my understanding of the whole issue. So here's what I understand of the entire implementation and what this fix is targeting (correct me if I'm wrong). 

Initially, we went with the approach of ordering the opened pages on our own (with a cap of 5 pages). However, since we were able to track the active tabs on the browser by patching the `live-server` code a little, we went with this approach instead. The benefit that arises from this is that we can track fully the workspace of the user and it can always provide us with an accurate list of the tabs opened on the browser. Thus, we don't have to keep a limit on the number of pages that are tracked. The responsibility of tracking of the list of opened tabs is solely on the `live-server`. What MarkBind does is simply to retrieve this list of opened tabs whenever necessary (e.g. addition / deletion / modification of source files). 

In the previous implementation, we actually implemented the addition of the tab into the list in the `request-response` phase (in the `file` method) as the idea was to ensure that the list of opened tabs is the "most-updated" possible (even real-time maybe), since if we place it in the `socket-initialisation` phase, the list won't be the "most-updated" possible, as socket initializations only happen after the `request-response` phase. 

However, we did not expect an edge case scenario where the `request-response` phase is skipped entirely and dives straight into the `socket initialization` phase (e.g. backwards navigation). In the case of backwards navigation, we will initialize the socket straightaway (which assumes the current tab entry is available), however the tab would not be existent since the past tab has been removed from the list of opened tabs already, and the new tab was not created in the `request-response` phase (in the `file` method specifically). Thus, this causes the `Cannot set property of undefined` error that we are seeing currently, since the tab entry doesn't exist. 

Furthermore, regarding the concern of keeping the list of opened tabs as the "most-updated" possible (which is why we implemented the addition of tab into the list in the `request-response` phase, in the `file` method), it is actually not necessary to keep it "most-updated" possible, since MarkBind only retrieves the list of opened tabs from `live-server` only upon fs-events. We don't have to be concerned about the real-time updates of the list of opened tabs due to browser-related activities (except for visiting a page which hasn't been built before). We should only be concerned about getting the most updated list of opened tabs when we edit the source files and thus triggering fs-events. Therefore, conceptually, there shouldn't be a concern of where we implement the addition of tab into the list (either `request-response` phase or `socket-initialization` phase). But due to the edge case bug we are facing now, we definitely have to shift it to `socket-initialization` phase. 

Can I check if I have understood the context correctly so far? :-)

ryoarmanda:
Thanks for going through the history of the implementation @wxwxwxwx9 ! What you have summed up is so far accurate.

wxwxwxwx9:
@ryoarmanda Thanks for confirming my understanding :-) I think this PR is rather straightforward, I have tested the various cases and things mostly seems to be working fine. However, I would just like to clarify the expected behaviors for the test cases:

> ### Test cases
> Important: MarkBind's opened pages list is only updated whenever the filesystem watch events fired. So the effects of some cases will not be apparent in the logs in real-time, but internally, the changes in the active tabs are caught and handled appropriately by `live-server`.
> 
> No	Cases	Expected behaviour (MarkBind)
> 1	Visits a page on a new tab	Opened pages list is only updated when the page is new/rebuilt.
Isn't it the case that the opened pages list in MarkBind will only update during fs events (when source files are added/removed/modified)? Or could it be that you are referring to the copy of opened pages list in the `live-server`? If that's the case, then the opened pages list (a.k.a activeTabs in `live-server`) is updated 1-to-1 against the opened tabs in browser, right? I'm not really sure about the part "only updated when the page is new/rebuilt". 
> 2	Visits a page that actually just a redirect to another page (e.g. our docs' `userGuide/index`)	Opened pages list is only updated when the page is new/rebuilt, but if so, it can update up to two times.
Same question as test case 1.
> 3	Visits a page when the same page is also opened on another tab	Opened pages list is not affected until fs events are detected.
From what I understand, the opened pages list on `live-server` will have two tab entries pointing to the same URL (but different client websockets). And the opened pages list on MarkBind will have only 1 entry (the most-recent entry, which is just the duplicate tab that comes earlier), but this is after we have executed the `uniq` function on the list retrieved from `live-server`. So technically, the opened pages list on MarkBind will not be affected by this user action right?
> 4	Reload a tab	Opened pages list is not affected until fs events are detected.
For the live-reload initiated by `live-server`, I think both the opened pages list on `live-server` and MarkBind would not be affected (both before and after fs events) right? Since the patch implementation we have currently retains the tab entry. However, if it is a reload initiated by the user, then the page will shift up to the top of the pages list on `live-server`, and this will only reflect on MarkBind's copy of opened pages list upon fs events. 
> 5	Close a tab	Opened pages list is not affected until fs events are detected.
> 6	Visits a page on an existing tab	Opened pages list is not affected until fs events are detected.
> 7	Edit a page	Opened pages list is synchronized with `live-server`'s active tabs before issuing rebuild.
> 8	Navigate tab backwards/forwards	Opened pages list is not affected until fs events are detected.

Another minor issue that I found is that empty path (that is `index.html` by default) actually is not reactive to changes to `index.md`. This may be confusing for the user when he/she clicks on the url port provided by the console (e.g. `http://127.0.0.1:8080`) and discovers that the page is not updating when he makes changes to `index.md`. The log displayed  for the change will display an empty page as well. Do u think there's any workaround for this? 
Screen Shot 2021-03-31 at 9 01 09 PM

ryoarmanda:
> Isn't it the case that the opened pages list in MarkBind will only update during fs events (when source files are added/removed/modified)? Or could it be that you are referring to the copy of opened pages list in the `live-server`? If that's the case, then the opened pages list (a.k.a activeTabs in `live-server`) is updated 1-to-1 against the opened tabs in browser, right? I'm not really sure about the part "only updated when the page is new/rebuilt".

Oh yeah, this is supposed to be only updated during fs events 😅 I was basing this from the testcase table from the old PR at https://github.com/MarkBind/markbind/pull/1513#issuecomment-801943121. It's updated now.

> From what I understand, the opened pages list on `live-server` will have two tab entries pointing to the same URL (but different client websockets). And the opened pages list on MarkBind will have only 1 entry (the most-recent entry, which is just the duplicate tab that comes earlier), but this is after we have executed the `uniq` function on the list retrieved from `live-server`. So technically, the opened pages list on MarkBind will not be affected by this user action right?

Correct. I might need to add this clarifying point on the expected behaviour. Thanks :)

ryoarmanda:
> Another minor issue that I found is that empty path (that is `index.html` by default) actually is not reactive to changes to `index.md`. This may be confusing for the user when he/she clicks on the url port provided by the console (e.g. `http://127.0.0.1:8080`) and discovers that the page is not updating when he makes changes to `index.md`. The log displayed for the change will display an empty page as well. Do u think there's any workaround for this?
> Screen Shot 2021-03-31 at 9 01 09 PM

Nice catch. I developed the implementation on our docs as the testing grounds, which always have "complete" urls, so I didn't catch this sooner.

On MarkBind side, "incomplete" urls are appended `index` before being recorded as the current page viewed. I might be able to apply the same procedure before calling `changeCurrentOpenedPages`.

wxwxwxwx9:
@ryoarmanda thanks for the changes! I tried out the changes and it seems to be working good. The documentation of the test cases in the PR seems pretty comprehensive as well. I can't seem to think of other edge cases that we may have missed out so I think what we have covered so far should be sufficient at least. 

Perhaps just one last minor nit -- in the `Current opened pages, from most-to-least recent:` log, I'm thinking of adding a comment to tell user that duplicated pages would not be displayed here. Thus even if the user has 5 same pages opened, the log will only display 1 entry. It's a minor thing but I think it will help the user. 

@ang-zeyu I'm okay with approving this PR, but I'm thinking of getting thumbs up from you first. Do you think there's anything I missed out? :-)

packages/cli/src/lib/live-server/index.js
```diff
@@ -382,11 +365,22 @@ LiveServer.start = function(options) {
...
     const reqUrl = path.dirname(request.url);
-    const tab = LiveServer.activeTabs.find(tab => tab.url === reqUrl && !tab.client);
-    tab.client = ws;
-    tab.isReloading = false;
+    // Guard clause for MarkBind's _include_ files, no need to recognize as an active tab
+    if (reqUrl.endsWith('._include_.html')) {
+      return;
+    }
+
+    // If present an entry with empty client, reuse existing entry to maintain order from pre-reload
```
ang-zeyu, line 375:
```suggestion
    // If an entry with empty client is present, reuse existing entry to maintain order from pre-reload 
```

packages/cli/src/lib/live-server/index.js
```diff
@@ -382,11 +365,22 @@ LiveServer.start = function(options) {
...
+
+    // If present an entry with empty client, reuse existing entry to maintain order from pre-reload 
+    const existingTab = LiveServer.activeTabs.find(tab => tab.url === reqUrl && !tab.client);
+    if (existingTab) {
+      existingTab.client = ws;
+      return;
+    }
+
+    // Insert new entry to the active tabs list
+    LiveServer.activeTabs.unshift({ url: reqUrl, client: ws});
```
ang-zeyu, line 383:
```suggestion
    LiveServer.activeTabs.unshift({ url: reqUrl, client: ws });
```

packages/cli/src/lib/live-server/index.js
```diff
@@ -382,11 +365,22 @@ LiveServer.start = function(options) {
...
     };
 
-    // CHANGED: Added line to record tab client socket on creation
+    // CHANGED: Enhanced client websocket addition process to record the client as an active tab entry
     const reqUrl = path.dirname(request.url);
-    const tab = LiveServer.activeTabs.find(tab => tab.url === reqUrl && !tab.client);
-    tab.client = ws;
-    tab.isReloading = false;
+    // Guard clause for MarkBind's _include_ files, no need to recognize as an active tab
+    if (reqUrl.endsWith('._include_.html')) {
```
ang-zeyu, line 371: shouldn't be needed with the socket approach as well since its not a page

packages/cli/index.js
```diff
@@ -155,9 +160,14 @@ program
 
     const changeHandler = (filePath) => {
       logger.info(`[${new Date().toLocaleTimeString()}] Reload for file change: ${filePath}`);
-      logger.info('Synchronizing opened pages list before reload');
-      const normalizedActiveUrls = liveServer.getActiveUrls().map(url => fsUtil.removeExtension(url));
-      site.changeCurrentOpenedPages(normalizedActiveUrls);
+      if (onePagePath) {
+        logger.info('Synchronizing opened pages list before reload');
+        const normalizedActiveUrls = liveServer.getActiveUrls().map((url) => {
+          const completeUrl = path.extname(url) === '' ? path.join(url, 'index') : url;
```
ang-zeyu, line 166: let's separate these unrelated ones to another PR with another appropriate commit message though

ang-zeyu (commented):
Mostly lgtm 👍 

Just a few nits:

---

> Perhaps just one last minor nit -- in the Current opened pages, from most-to-least recent: log, I'm thinking of adding a comment to tell user that duplicated pages would not be displayed here. Thus even if the user has 5 same pages opened, the log will only display 1 entry. It's a minor thing but I think it will help the user.

Hmm I think that's ok, since the user would just need to figure it once.
Since the message is repeatedly logged might be good to keep it minimal as well.

packages/cli/index.js
ryoarmanda: I think that's better. I will do a revert commit and put this as a new minor PR.

packages/cli/src/lib/live-server/index.js
ryoarmanda: Brain fart there haha. Will fix.

packages/cli/src/lib/live-server/index.js
ryoarmanda: Ah I see, alright.

ang-zeyu (approved):
Looks good 👍, thanks!

PR #1529, opened 2021-03-30
Implement live preview for site.json

### LINKED ISSUE #48: Support live preview for new entries to sites.json

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #48.

A follow up to #1514, which supported live preview for the pages attribute. This PR will support live preview for all the other attributes in `site.json`. A server restart is no longer required for these changes to be reflected when serving the site through `markbind serve`. Since most of the attributes in `site.json` (except `baseUrl`, `pages` and `deploy`) will affect majority of the pages, hence all pages will be regenerated when any of these attributes (except those listed above) is modified. 

For the `style` attribute, if a `bootswatch` theme is specified/added, its `bootstrap.min.css` theme file will overwrite the existing file and all pages are rebuilt. Similarly, if a `bootswatch` theme is removed, the default `bootstrap.min.css`, which is located at `core-web/asset/css` will overwrite the existing file.

Would appreciate if you can let me know if I missed any attributes or if any of the attributes (other than those listed above) that does not require all pages to be rebuilt when it is modified.

**Testing instructions:**
1. Serve a MarkBind site using `markbind serve`
2. Open and edit the any of the attributes in site.json
3. Live preview should be updated to reflect the latest changes

Implement live preview for site.json

When attributes other than pages in site.json are modified, a server 
restart is required to reflect the changes on the live preview. This can 
become a hassle to the user as they will need to frequently restart 
the server when trying out different site configuration.

Let's support live preview for all the attributes in site.json so that the 
latest changes can be reflected without restarting the server.

---

### DIFF
--- a/docs/userGuide/glossary.md
+++ b/docs/userGuide/glossary.md
@@ -10,6 +10,8 @@
 **_Live preview_** is:
 - Regeneration of affected content upon any change to <tooltip content="`.md`, `.mbd`, `.mbdf`, `.njk` files ... anything your content depends on!">source files</tooltip>, then reloading the updated site in the Browser.
 
+- Regeneration will also occur upon any modification to attributes in `site.json` with the exception of [`baseUrl`](siteJsonFile.md#baseurl).
+
 - Copying <tooltip content="files that don't affect page generation (eg. images), but are used in the site">assets</tooltip> to the site output folder.
 
 Use [the `serve` command](cliCommands.html#serve-command) to launch a live preview.
--- a/docs/userGuide/glossary.md
+++ b/docs/userGuide/glossary.md
@@ -14,7 +14,6 @@
 
 Use [the `serve` command](cliCommands.html#serve-command) to launch a live preview.
 
-
 </span>
 
 <br>
--- a/docs/userGuide/glossary.md
+++ b/docs/userGuide/glossary.md
@@ -10,9 +10,12 @@
 **_Live preview_** is:
 - Regeneration of affected content upon any change to <tooltip content="`.md`, `.mbd`, `.mbdf`, `.njk` files ... anything your content depends on!">source files</tooltip>, then reloading the updated site in the Browser.
 
+- Regeneration will also occur upon any modification to attributes in `site.json` with the exception of [`baseUrl`](siteJsonFile.md#baseurl-no-live-preview).
+
 - Copying <tooltip content="files that don't affect page generation (eg. images), but are used in the site">assets</tooltip> to the site output folder.
 
 Use [the `serve` command](cliCommands.html#serve-command) to launch a live preview.
+
 
 </span>
 
--- a/docs/userGuide/glossary.md
+++ b/docs/userGuide/glossary.md
@@ -10,8 +10,6 @@
 **_Live preview_** is:
 - Regeneration of affected content upon any change to <tooltip content="`.md`, `.mbd`, `.mbdf`, `.njk` files ... anything your content depends on!">source files</tooltip>, then reloading the updated site in the Browser.
 
-- Regeneration will also occur upon any modification to attributes in `site.json` with the exception of [`baseUrl`](siteJsonFile.md#baseurl-no-live-preview).
-
 - Copying <tooltip content="files that don't affect page generation (eg. images), but are used in the site">assets</tooltip> to the site output folder.
 
 Use [the `serve` command](cliCommands.html#serve-command) to launch a live preview.
--- a/docs/userGuide/siteJsonFile.md
+++ b/docs/userGuide/siteJsonFile.md
@@ -81,12 +81,10 @@ Here is a typical `site.json` file:
 
 <include src="deployingTheSite.md#warning-about-baseUrl" />
 
-<span id="baseurl-no-live-preview">
 <box type="warning">
 
-Note: `baseUrl` does not support live preview as there is no use case for changing it in during `markbind serve`.
+Note: `baseUrl` does not support [live preview](glossary.md#live-preview) as there is no use case for changing it in during `markbind serve`.
 </box>
-</span>
 
 #### **`faviconPath`**
 
--- a/docs/userGuide/siteJsonFile.md
+++ b/docs/userGuide/siteJsonFile.md
@@ -81,6 +81,10 @@ Here is a typical `site.json` file:
 
 <include src="deployingTheSite.md#warning-about-baseUrl" />
 
+<box type="warning">
+
+Note: `baseUrl` does not support live reload as it should not change in `markbind serve` mode.
+</box>
 
 #### **`faviconPath`**
 
--- a/docs/userGuide/siteJsonFile.md
+++ b/docs/userGuide/siteJsonFile.md
@@ -81,10 +81,12 @@ Here is a typical `site.json` file:
 
 <include src="deployingTheSite.md#warning-about-baseUrl" />
 
+<span id="baseurl-no-live-preview">
 <box type="warning">
 
-Note: `baseUrl` does not support live reload as it should not change in `markbind serve` mode.
+Note: `baseUrl` does not support live preview as there is no use case for changing it in during `markbind serve`.
 </box>
+</span>
 
 #### **`faviconPath`**
 
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -1276,15 +1276,18 @@ class Site {
 
   /**
    * Copies bootstrap theme to the assets folder if a valid theme is specified
-   * @param {Boolean} toCopyDefault bootstrap theme to the assets folder
+   * @param {Boolean} isRebuild only true if it is a rebuild
    */
-  copyBootstrapTheme(toCopyDefault) {
+  copyBootstrapTheme(isRebuild) {
     const { theme } = this.siteConfig;
-    if (!toCopyDefault && (!theme || !_.has(SUPPORTED_THEMES_PATHS, theme))) {
+
+    // If is it the initial build using the default theme or if the theme specified
+    // is not valid, then do nothing.
+    if ((!isRebuild && !theme) || (theme && !_.has(SUPPORTED_THEMES_PATHS, theme))) {
       return _.noop;
     }
 
-    const themeSrcPath = toCopyDefault && !theme
+    const themeSrcPath = !theme
       ? require.resolve('@markbind/core-web/asset/css/bootstrap.min.css')
       : SUPPORTED_THEMES_PATHS[theme];
     const themeDestPath = path.join(this.siteAssetsDestPath, 'css', 'bootstrap.min.css');
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -1281,7 +1281,7 @@ class Site {
   copyBootstrapTheme(isRebuild) {
     const { theme } = this.siteConfig;
 
-    // If is it the initial build using the default theme or if the theme specified
+    // If it is the initial build using the default theme or if the theme specified
     // is not valid, then do nothing.
     if ((!isRebuild && !theme) || (theme && !_.has(SUPPORTED_THEMES_PATHS, theme))) {
       return _.noop;
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -856,16 +856,25 @@ class Site {
     const oldPagesSrc = oldAddressablePages.map(page => page.src);
     await this.readSiteConfig();
     await this.handleIgnoreReload(oldSiteConfig);
-    await this.handlePageReload(oldAddressablePages, oldPagesSrc,
-                                oldSiteConfig, this.isGlobalConfigModified(oldSiteConfig));
+    await this.handlePageReload(oldAddressablePages, oldPagesSrc, oldSiteConfig);
     await this.handleStyleReload(oldSiteConfig.style);
   }
 
   /**
-   * Checks if any of the global attributes of site.json is modified
+   * Handles the rebuilding of modified pages
    */
-  isGlobalConfigModified(oldSiteConfig) {
-    return !_.isEqual(oldSiteConfig.faviconPath, this.siteConfig.faviconPath)
+  async handlePageReload(oldAddressablePages, oldPagesSrc, oldSiteConfig) {
+    this.collectAddressablePages();
+
+    // Comparator for the _differenceWith comparison below
+    const isNewPage = (newPage, oldPage) => _.isEqual(newPage, oldPage) || newPage.src === oldPage.src;
+
+    const addedPages = _.differenceWith(this.addressablePages, oldAddressablePages, isNewPage);
+    const removedPages = _.differenceWith(oldAddressablePages, this.addressablePages, isNewPage)
+      .map(filePath => Site.setExtension(filePath.src, '.html'));
+
+    // Checks if any attributes of site.json requiring a global rebuild are modified
+    const isGlobalConfigModified = () => !_.isEqual(oldSiteConfig.faviconPath, this.siteConfig.faviconPath)
         || !_.isEqual(oldSiteConfig.titlePrefix, this.siteConfig.titlePrefix)
         || !_.isEqual(oldSiteConfig.style, this.siteConfig.style)
         || !_.isEqual(oldSiteConfig.externalScripts, this.siteConfig.externalScripts)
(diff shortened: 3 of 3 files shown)


packages/core/src/Site/index.js
```diff
@@ -889,6 +919,35 @@ class Site {
     });
   }
 
+  /**
+   * Handles the reloading of ignore attributes
+   */
+  async handleIgnoreReload(oldIgnore) {
+    const assetsToRemoved = _.difference(this.siteConfig.ignore, oldIgnore);
+
+    if (!_.isEqual(oldIgnore, this.siteConfig.ignore)) {
```
raysonkoh, line 928: Hmm, is this check neccessary? Would there be a case where `assetsToRemoved` is truthy but `oldIgnore == this.siteConfig.ignore`?

packages/core/src/Site/index.js
```diff
@@ -889,6 +919,35 @@ class Site {
...
+
+  /**
+   * Handles the reloading of the style attribute if it has been modified
+   */
+  async handleStyleReload(oldStyle) {
+    if (!_.isEqual(oldStyle.bootstrapTheme, this.siteConfig.style.bootstrapTheme)) {
+      if (this.siteConfig.style.bootstrapTheme) {
+        await this.copyBootswatchTheme();
+      } else {
+        const defaultBootstrapTheme = require.resolve('@markbind/core-web/asset/css/bootstrap.min.css');
```
raysonkoh, line 942: It seems like this block of logic is similar to `Site#copyBootswatchTheme()`. Would it be better to modify `copyBootswatchTheme()` to also account for copying the default bootstrap theme and then using calling that function in here instead?

raysonkoh (commented):
Neat code! Would it be possible to add some tests for this?

packages/core/src/Site/index.js
jonahtanjz: I believe this check is necessary. For example, if one or more entries are removed from `ignore`, then `assetsToRemove` will be empty but `oldIgnore !== this.siteConfig.ignore` so the assets will be rebuilt. Another scenario is when `oldIgnore == this.siteConfig.ignore`, then in this case, if they are equal then no assets will need to be built or removed.

packages/core/src/Site/index.js
jonahtanjz: Sounds good. I have updated the method to copy the default theme if applicable during live reload of site.json.

packages/core/src/Site/index.js
raysonkoh: Ah okay got it. We can leave this in then 👍

packages/core/src/Site/index.js
```diff
@@ -1282,15 +1275,18 @@ class Site {
...
   /**
-   * Copies bootswatch theme to the assets folder if a valid theme is specified
+   * Copies bootstrap theme to the assets folder if a valid theme is specified
+   * @param {Boolean} toCopyDefault bootstrap theme to the assets folder
    */
-  copyBootswatchTheme() {
+  copyBootstrapTheme(toCopyDefault) {
     const { theme } = this.siteConfig;
-    if (!theme || !_.has(SUPPORTED_THEMES_PATHS, theme)) {
+    if (!toCopyDefault && (!theme || !_.has(SUPPORTED_THEMES_PATHS, theme))) {
```
raysonkoh, line 1283: Minor nit: would it be possible to remove the need for the `toCopyDefault`? It seems like if `toCopyDefault` is True and `theme` is defined, then it would copy the theme's file instead of the default file. 

What do you think of the following flow:
If `theme` is undefined, then we will assume that the caller wants to copy the default bootstrap theme.
Else check if `theme` is a supported theme and copy over the relevant theme file.

jonahtanjz: > What do you think of the following flow:
> If theme is undefined, then we will assume that the caller wants to copy the default bootstrap theme.
> Else check if theme is a supported theme and copy over the relevant theme file.

Hmm I think this will make a redundant copy of the default theme if no theme is specified in the initial build. Maybe I can rename `toCopyDefault` to `isRebuild` so that it is more concise. What do you think?

raysonkoh: > Hmm I think this will make a redundant copy of the default theme if no theme is specified in the initial build. 

Hmm, I see. In that case, should we log a warning / throw exception for the case `toCopyDefault = True && theme`?

> Maybe I can rename toCopyDefault to isRebuild so that it is more concise. What do you think?

I am fine with either name.

jonahtanjz: Sorry I think I wasn't too clear about this. I think I should rename `toCopyDefault` as it is a little misleading. This parameter should only be True if it is a rebuild (site.json was modified).

What do you think about this check:
```
if ((!isRebuild && !theme) || (theme && !_.has(SUPPORTED_THEMES_PATHS, theme))) {
      return _.noop;
}
```
It will be a `noop` during the first build if theme is not specified/theme is invalid. Else, if it is subsequent rebuild/theme is specified and valid, then, it will copy over the relevant theme's file (including default theme if `!theme`).

raysonkoh: Yes that looks good! Can you also add a comment next to this check to indicate its purpose for future reference?

jonahtanjz: Alright I've updated this and also added the comment :)

packages/core/src/Site/index.js
```diff
@@ -844,9 +844,39 @@ class Site {
...
+  /**
+   * Checks if any of the global attributes of site.json is modified
+   */
+  isGlobalConfigModified(oldSiteConfig) {
+    return !_.isEqual(oldSiteConfig.faviconPath, this.siteConfig.faviconPath)
+        || !_.isEqual(oldSiteConfig.titlePrefix, this.siteConfig.titlePrefix)
+        || !_.isEqual(oldSiteConfig.style, this.siteConfig.style)
+        || !_.isEqual(oldSiteConfig.externalScripts, this.siteConfig.externalScripts)
+        || !_.isEqual(oldSiteConfig.globalOverride, this.siteConfig.globalOverride)
+        || !_.isEqual(oldSiteConfig.ignore, this.siteConfig.ignore)
```
ang-zeyu, line 865: 👀 what's this for?

ang-zeyu (commented):
Thanks! @jonahtanjz 

> Since most of the attributes in site.json (except baseUrl, pages and deploy) will affect majority of the pages, hence all pages will be regenerated when any of these attributes (except those listed above) is modified.

Hmm shouldn't `baseUrl` be included? (`{{ baseUrl }}` variable changes).
Although, there might not be much value in supporting this in `serve` mode (since `baseUrl` would really only change for deployment), we can document this exception if so. 

As there are a few distinct groups / individual properties in this PR with separate logic some commit structure would be great here, e.g.:
- Support ignore...
- Support bootstrap theme...
- Support plugins... this needs some special handling as well (see `PluginManager`)
- Support "miscellaneous" page properties (all else under your `isGlobalConfigModified` function)
- ... another other appropriate ones

Separate PRs for each of these would be welcome instead as well

---

(optional) You can also consider implementing the fourth part of #1495 first as well to ensure all pages have the same updated copy of `SiteConfig`'s properties

packages/core/src/Site/index.js
```diff
@@ -889,6 +919,28 @@ class Site {
     });
   }
 
+  /**
+   * Handles the reloading of ignore attributes
+   */
+  async handleIgnoreReload(oldIgnore) {
+    const assetsToRemove = _.difference(this.siteConfig.ignore, oldIgnore);
```
ang-zeyu, line 926: :eyes: should this be the other way round?

packages/core/src/Site/index.js
jonahtanjz: This is to check if any of these configs have changed and if they have changed, would require a rebuild of all pages.

packages/core/src/Site/index.js
jonahtanjz: Hmm I think this is right as new items that are added to ignore will have to be removed?

jonahtanjz:
> Hmm shouldn't baseUrl be included? ({{ baseUrl }} variable changes).
> Although, there might not be much value in supporting this in serve mode (since baseUrl would really only change for deployment), we can document this exception if so.

Since {{ baseUrl }} is a variable change then I supposed it would make sense to be included so that the behaviour would be consistent.

> Plugins would need some special handling as well like themes and ignore. (see PluginManager)

For this portion were you referring to running `collectBaseUrl()` method? Since this is the only part where a new `PluginManager` is created and then subsequently assigned to all the new pages that will be rebuilt.

packages/core/src/Site/index.js
ang-zeyu: 👍 my bad, was looking from perspective of `siteConfig."include"`

packages/core/src/Site/index.js
ang-zeyu: (`ignore` in particular) Shouldn't be necessary I think; `ignore` is only used down the line for link validation warnings

ang-zeyu (commented):
> Since {{ baseUrl }} is a variable change then I supposed it would make sense to be included so that the behaviour would be consistent.

@jonahtanjz its common and ok to forgo supporting something if the use case is practically zero, especially since `baseUrl` needs extra special care.

I might be missing a use case though. Could get a user opinion from @damithc here. (would live-reload for `baseUrl` parameter in site config be useful?)

> For this portion were you referring to running collectBaseUrl() method? Since this is the only part where a new PluginManager is created and then subsequently assigned to all the new pages that will be rebuilt.

Was thinking a more targeted approach that dosen't rerun the baseUrl collection. (can do a simple refactor / extraction of method here)
On that note (good catch), `External/LayoutManager`'s copies in `collectBaseUrl()` should be refreshed as well.

damithc:
> I might be missing a use case though. Could get a user opinion from @damithc here. (would live-reload for `baseUrl` parameter in site config be useful?)

I think it is OK not support live preview support for changes to the baseUrl as it is a very rare case. Just need to specify it clearly.

packages/core/src/Site/index.js
jonahtanjz: Ahh ok. I was under the impression that if a page is dependent on an asset that is later added to `ignore`, then that entire page will have to be rebuilt. Might have misunderstood the functionality of `ignore` here 😅

jonahtanjz:
@ang-zeyu @raysonkoh I've updated the PR to include the exception of not supporting live reload for `baseUrl` in the UG and refresh the `managers` when necessary. Let me know if I've missed out anything else :)

packages/core/src/Site/index.js
```diff
@@ -889,6 +927,42 @@ class Site {
...
+
+  /**
+   * Chceks the relevant site configurations for changes and rebuild managers if necessary
+   */
+  rebuildManagers(oldSiteConfig) {
+    if (!_.isEqual(oldSiteConfig.ignore, this.siteConfig.ignore)
+      || !_.isEqual(oldSiteConfig.plugins, this.siteConfig.plugins)
+      || !_.isEqual(oldSiteConfig.pluginsContext, this.siteConfig.pluginsContext)
+      || !_.isEqual(oldSiteConfig.disableHtmlBeautify, this.siteConfig.disableHtmlBeautify)
+      || !_.isEqual(oldSiteConfig.intrasiteLinkValidation, this.siteConfig.intrasiteLinkValidation)) {
```
ang-zeyu, line 951: missing `timeZone / locale / style` as well - should / could we just remove the checks here as well?
I feel there's not much value in further narrowing it down since `buildManagers` isn't very costly

packages/core/src/Site/index.js
```diff
@@ -844,9 +851,39 @@ class Site {
...
     const oldPagesSrc = oldAddressablePages.map(page => page.src);
     await this.readSiteConfig();
+    await this.handleIgnoreReload(oldSiteConfig);
+    await this.handlePageReload(oldAddressablePages, oldPagesSrc,
+                                oldSiteConfig, this.isGlobalConfigModified(oldSiteConfig));
+    await this.handleStyleReload(oldSiteConfig.style);
+  }
+
+  /**
+   * Checks if any of the global attributes of site.json is modified
```
ang-zeyu, line 865:
```suggestion
   * Checks if any attributes of site.json requiring a global rebuild are modified
```
`global attributes` might be a little vague

packages/core/src/Site/index.js
```diff
@@ -844,9 +851,39 @@ class Site {
   }
 
   async reloadSiteConfig() {
+    const oldSiteConfig = this.siteConfig;
     const oldAddressablePages = this.addressablePages.slice();
     const oldPagesSrc = oldAddressablePages.map(page => page.src);
     await this.readSiteConfig();
+    await this.handleIgnoreReload(oldSiteConfig);
+    await this.handlePageReload(oldAddressablePages, oldPagesSrc,
+                                oldSiteConfig, this.isGlobalConfigModified(oldSiteConfig));
```
ang-zeyu, line 860: hmm should we `isGlobalConfigModified` inside `handlePageReload`? since its not used anywhere else and concerns page rebuilding only

packages/core/src/Site/index.js
jonahtanjz: > I feel there's not much value in further narrowing it down since buildManagers isn't very costly

Alright since there's not much value for the checks as it is not costly, then `buildManagers` will be called for every full rebuild and whenever `ignore` is modified. I've updated the PR accordingly.

jonahtanjz:
@ang-zeyu Thanks for the catch and suggestions. I've updated the PR to remove the further checks for `buildManagers` and also moved the `isGlobalConfigModified` method into `handlePageReload`.

docs/userGuide/siteJsonFile.md
```diff
@@ -81,6 +81,10 @@ Here is a typical `site.json` file:
 
 <include src="deployingTheSite.md#warning-about-baseUrl" />
 
+<box type="warning">
+
+Note: `baseUrl` does not support live reload as it should not change in `markbind serve` mode.
```
ang-zeyu, line 86: Let's update the definition here as well with this https://deploy-preview-1529--markbind-master.netlify.app/userguide/glossary

Minor wording nit:
```suggestion
Note: `baseUrl` does not support live preview as there is no use case for changing it in during `markbind serve`.
```

Otherwise, Lgtm! 👍 

Let's wait and see if @raysonkoh has anything to add

packages/core/src/Site/index.js
```diff
@@ -1223,15 +1279,21 @@ class Site {
...
-   * Copies bootswatch theme to the assets folder if a valid theme is specified
+   * Copies bootstrap theme to the assets folder if a valid theme is specified
+   * @param {Boolean} isRebuild only true if it is a rebuild
    */
-  copyBootswatchTheme() {
+  copyBootstrapTheme(isRebuild) {
     const { theme } = this.siteConfig;
-    if (!theme || !_.has(SUPPORTED_THEMES_PATHS, theme)) {
+
+    // If it is the initial build using the default theme or if the theme specified
```
ang-zeyu, line 1288: Let's stick to `/* ... */` for multiline comments

packages/core/src/Site/index.js
```diff
@@ -856,8 +873,24 @@ class Site {
...
+        || !_.isEqual(oldSiteConfig.enableSearch, this.siteConfig.enableSearch)
+        || !_.isEqual(oldSiteConfig.disableHtmlBeautify, this.siteConfig.disableHtmlBeautify)
+        || !_.isEqual(oldSiteConfig.timeZone, this.siteConfig.timeZone)
+        || !_.isEqual(oldSiteConfig.locale, this.siteConfig.locale)
+        || !_.isEqual(oldSiteConfig.intrasiteLinkValidation, this.siteConfig.intrasiteLinkValidation);
+
+    if (isGlobalConfigModified() || !_.isEmpty(addedPages) || !_.isEmpty(removedPages)) {
       await this.removeAsset(removedPages);
+      this.buildManagers();
       await this._rebuildSourceFiles();
```
ang-zeyu, line 894: Additional remark (no action needed): 

>> EDIT: After looking around at the other handlers (addHandler and removeHandler), it seems all pages should be rebuilt if the added/removed file is a page/dependency of page.
>
>This has been existing behaviour for a long while already but can probably be made into an issue 🙂

Per #1514's comment, the number of dependent pages generated / regenerated can be improved later as well.

packages/core/src/Site/index.js
ang-zeyu: sorry, actually just meant changing the call site. But the current changes are ok as well : )

packages/core/src/Site/index.js
jonahtanjz: Alright. I'll open an issue for this and the previous discussion.

jonahtanjz:
@ang-zeyu @raysonkoh I've updated the PR accordingly :)

docs/userGuide/glossary.md
```diff
@@ -10,8 +10,6 @@
 **_Live preview_** is:
 - Regeneration of affected content upon any change to <tooltip content="`.md`, `.mbd`, `.mbdf`, `.njk` files ... anything your content depends on!">source files</tooltip>, then reloading the updated site in the Browser.
 
-- Regeneration will also occur upon any modification to attributes in `site.json` with the exception of [`baseUrl`](siteJsonFile.md#baseurl-no-live-preview).
```
ang-zeyu: 👀

jonahtanjz: I've re-added this portion

ang-zeyu (approved):

raysonkoh:
LGTM 👍

raysonkoh (approved):

PR #1538, opened 2021-04-04
Fix delay to return delayed function result for use

### LINKED ISSUE #1537: Async delay function does not return result
Minor bug found during development. If one tries to `delay` (from `packages/core/src/utils/delay.js`) an async function that is expected to return value (e.g. `Promise<boolean>`), the delayed function does not return the value for further use.

As currently our applications of delay are on async functions that does not return anything (e.g. `Promise<void>`), nor do we do anything to the return results (no values returned in the first place), this has gone unnoticed as it doesn't break anything in the codebase.

Example debug result during my development (`site.backgroundBuildNotViewedFiles` is a `delay`-ed function that should return `Promise<boolean>`):

image

Reproduction:
1. Create a function that returns a value
2. Apply `delay` to it
3. Use the delayed function
4. Then observe the return of the function

ryoarmanda:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

Resolves #1537.

**Overview of changes:**

The part of the `delay` function that caused the issue is specifically this:

https://github.com/MarkBind/markbind/blob/5c826ce0c447e86137534a64be3c048acc5754c1/packages/core/src/utils/delay.js#L25-L34

Notice the use of `finally` to start calling the delayed function. `finally` is a bit finicky, through some experiments, turns out the return of `finally` cannot be used in the next `then`/after `await` (in fact, `finally` does not disrupt the return value of the promise, ref: [here](https://dev.to/annarankin/finally-in-promises--trycatch-2c44)) so the `return funcPromise` is effectively unable to be used.

Changing `finally` to `then` fixes the issue, and I believe still preserves the behaviour of the previous implementation.

Reproduction and fixed demos is available here: https://codesandbox.io/s/cool-sid-25hmb. Check the console tab on the bottom right part of the page and refresh the "browser" on the right panel to test it over and over. Can play with `flowOriginal` and `flowFixed` for the original and fixed implementation, the argument is the number of calls of the delayed `doWork` function (can try `1` first to understand the behaviours, I put `2` to test consistency of the approach).

Notice the condition of the image in the issue is reproduced when using `flowOriginal` and all the `doWork` calls get resolved.

**Testing instructions:**

Fix delay to return delayed function result for use

---

### DIFF
--- a/packages/core/src/utils/delay.js
+++ b/packages/core/src/utils/delay.js
@@ -24,7 +24,7 @@ module.exports = function delay(func, wait) {
 
     if (waitingPromise === null) {
       waitingPromise = Promise.all([Promise.delay(wait), runningPromise])
-        .finally(() => {
+        .then(() => {
           runningPromise = waitingPromise || Promise.resolve();
           waitingPromise = null;
           const funcPromise = func.apply(context, [pendingArgs]);


packages/core/src/utils/delay.js
```diff
@@ -24,7 +24,7 @@ module.exports = function delay(func, wait) {
 
     if (waitingPromise === null) {
       waitingPromise = Promise.all([Promise.delay(wait), runningPromise])
```
jonahtanjz, line 26: One question: Will this promise ever be rejected? If it does will this cause a change in behaviour (`.then` will not run in this case but `.finally` will always run regardless)?

jonahtanjz (commented):
Nice work and it looks good 👍 Just one question below.

packages/core/src/utils/delay.js
ryoarmanda: I don't think there is a case for that promise to be rejected. My observation is that `waitingPromise` is initially a `Promise.all` of a `Promise.delay` and a `Promise.resolve`, and based on the callback, subsequent `waitingPromise` will be either:

- Similar to the original (`Promise.all` of `Promise.delay` and `Promise.resolve`; or
- A `Promise.all` of `Promise.delay` and the *previous* `waitingPromise`

As the behaviour of `Promise.all` depends on the behaviour of the Promises inside, `waitingPromise` essentially just boils down to the behaviour of `Promise.delay` and `Promise.resolve`. Both have no chance of rejection: `Promise.delay` without second argument returns undefined on resolve, the same with `Promise.resolve`. So in the end, `waitingPromise` will never be rejected.

jonahtanjz: Alright :) Thanks for the clarification and detailed explanation 👍

jonahtanjz (approved):
LGTM :)

PR #1446, opened 2021-01-16
Refactor Site.js to use async/await

jonahtanjz:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [ ] Feature addition or enhancement
- [x] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Related #809.
Refactor Site.js promise chaining to async await.

**Anything you'd like to highlight / discuss:**
Code style for functions that were declared and executed inside `.then` previously.

Example: 
```
.then((assets) => {
  ...
})
```

is refactored to:
```
return await (async (assets) => {
  ...
})(assets)
```
Is the current implementation acceptable or should I extract these functions to improve readability?

**Testing instructions:**
NIL

**Proposed commit message: (wrap lines at 72 characters)**
Refactor Site.js to use async await

---

**Checklist:** :ballot_box_with_check:

- [x] Updated the documentation for feature additions and enhancements
- [x] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No blatantly unrelated changes
- [x] Pinged someone for a review!

PR #1541, opened 2021-04-06
Fix missing return termination

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Follow up to PR #1446

Missed out on a [return termination](https://github.com/MarkBind/markbind/pull/1446#discussion_r607076078) in the `rebuildRequiredPages` method.

`NIL`

**Testing instructions:**
`NIL`

Fix missing return termination in rebuildRequiredPages

---

### DIFF
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -795,7 +795,8 @@ class Site {
       this.mapAddressablePagesToPages(this.addressablePages || [], this.getFavIconUrl());
 
       await this.rebuildPageBeingViewed(this.currentPageViewed);
-      return await this.lazyBuildAllPagesNotViewed();
+      await this.lazyBuildAllPagesNotViewed();
+      return;
     }
 
     logger.warn('Rebuilding all pages...');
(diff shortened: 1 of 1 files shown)


wxwxwxwx9:
@jonahtanjz LGTM! But just to clarify why this fix is required. 

In this async function, if we don't have the return statement, then what is implicitly returned is a promise that resolves to undefined. If we add this `return`, would it also implicitly return a promise that resolves to undefined?

jonahtanjz:
@wxwxwxwx9 Thanks for the review!

>But just to clarify why this fix is required.

The return termination is required so that if `this.onePagePath` is true, it should lazy build all the other pages and should not run `this.buildSourceFiles()`. Therefore it should terminate after `this.lazyBuildAllPagesNotViewed()`.

>In this async function, if we don't have the return statement, then what is implicitly returned is a promise that resolves to undefined. If we add this return, would it also implicitly return a promise that resolves to undefined?

IIRC an async function will return a promise that will resolve to the return value of that function. In this case, `return;` will return `undefined` and hence the async function should also return a promise that resolves to `undefined`.

wxwxwxwx9:
> The return termination is required so that if this.onePagePath is true, it should lazy build all the other pages and should not run this.buildSourceFiles(). Therefore it should terminate after this.lazyBuildAllPagesNotViewed().

Ah right, missed this part. Initially I was thinking the fix had something to do with promise resolution. Thanks for clarifying!

wxwxwxwx9 (approved):
LGTM!

PR #1548, opened 2021-04-17
Fix custom components markdown parsing

### LINKED ISSUE #958: Unexpected behaviour of components when immediately prepended by GFMD

wxwxwxwx9:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #958. 

This issue is also related to #1534. We were prompted to re-visit #958 as it needs to be fixed before we can proceed to implement #1534. 

As @ang-zeyu mentioned, we have never defined how our components should behave as html tags (block / inline (default)) regarding markdown. 

Thus, we decided to adapt VuePress implementation to fit our own needs. VuePress treats every custom component (or unknown tags) as block-level element. This applies mostly in our use-case as well, except for `<panel minimized>` component (panels which are minimized), which is treated as inline element. 

**Overview of changes:**
Adapted VuePress `markdown-it` [patch](https://github.com/vuepress/vuepress-next/tree/main/packages/%40vuepress/markdown/src/plugins/customComponentPlugins) for custom components. 

The patch used by VuePress replaces the `html_block` rule in `markdown-it`. However, on our side, we are also replacing `html_block` rule with `special-escape-tags` already. 

Thus, instead of modifying / merging `special-escape-tags` rule and VuePress's `custom-component` rule, I decided to just add VuePress's custom component patch rule under `special-escape-tags` rule. 

Should I look into combining the rules together somehow? Seeing that there is a bit of code duplication here and time-wise it is a little more expensive as `markdown-it` has to run through an additional rule (which can be combined into one).

But I think I am more comfortable with not combining them for separation of concern reasons. It would be easier to read the code with them being separate in my opinion. 

**Testing instructions:**
`npm run test`

Try out the examples as raised by @ang-zeyu in #958. The bugs should be fixed. 

However, when you use the minimized option for panel, then the "bug" (not actually a bug since it is expected behaviour) should still be present, as minimized panel is treated as an inline element instead of block-level element.

Fix custom components markdown parsing.

We have never defined the behavior of our custom components as html tags
(whether block-level or inline) in markdown parsing. This causes some 
undesirable edge cases during markdown parsing, as raised up in #958. 

Let's adopt the way VuePress define the behaviour for custom components
and adapt it to fit our needs. 

---


packages/core/src/lib/markdown-it/plugins/custom-component/customComponentPlugin.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Replacing the default htmlBlock rule to allow using custom components in markdown
+ */
+const customComponentPlugin = (md) => {
+  /*
+   * Note that html_block is to be replaced by markdown-it-escape-special-tags.
+   * Thus, we have to push a new rule after the original html_block rule instead.
+   */
+  md.block.ruler.after('html_block', 'custom_component_html_block', customComponentHtmlBlockRule, {
```
ang-zeyu, line 12: Let's combine the two patches, since both relate to the idea of custom components and edit the same markdown rule.

(could move this over to the `markdown-it/patches/` directory instead as well since it modifies markdown-it default behaviour)

packages/core/src/lib/markdown-it/plugins/custom-component/inlineTags.js
```diff
@@ -0,0 +1,78 @@
...
+ * treat them as "block" tags
+ *
+ * So we have to distinguish between "inline" and "unknown" tags ourselves
+ *
+ * The inline tags list comes from MDN
+ *
+ * @see https://spec.commonmark.org/0.29/#raw-html
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements
+ */
+const inlineTags = [
```
ang-zeyu, line 19: Would it be possible to whitelist `pic / thumbnail` as well? https://github.com/MarkBind/markbind/pull/1534#pullrequestreview-637316649
So its consistent with `` tags (e.g. author may add `float` style to float the image to the right of paragraph)

The components would likely have to be adapted slightly in the ssr pr then in terms of just inline tags.

packages/core/src/lib/markdown-it/plugins/custom-component/htmlBlockRule.js
```diff
@@ -0,0 +1,114 @@
...
+ * by using the regex below. 
+ */
+const minimalPanelStringRegex = '(panel.*minimized.*)';
+
+// Forked and modified from 'markdown-it/lib/rules_block/html_block.js'
+
+// An array of opening and corresponding closing sequences for html tags,
+// last argument defines whether it can terminate a paragraph or not
+const HTML_SEQUENCES = [
+  [/^<(script|pre|style)(?=(\s|>|$))/i, /<\/(script|pre|style)>/i, true],
```
ang-zeyu, line 18: per above, can be combined

ang-zeyu (commented):
Brief look:

packages/core/src/lib/markdown-it/plugins/custom-component/customComponentPlugin.js
wxwxwxwx9: @ang-zeyu Hmm now that you mention it, I agree that they do relate to the idea of custom components. I've made the changes accordingly. Let me know what you think of it :-)

packages/vue-components/src/Thumbnail.vue
```diff
@@ -1,20 +1,22 @@
...
-      :src="src"
-      class="thumb-image"
-      :width="computedWidth"
-      :height="computedHeight"
-      :alt="alt"
-      @load.once="computeImgSize"
-    />
-    <slot></slot>
-  </div>
+  <span>
```
ang-zeyu, line 2: > Hmm I was thinking that making the component changes would be appropriate in this PR as well. Do you think we can keep the changes in this PR? Regardless, I've made the changes to check if they are what you were talking about. :-)

Hmm I'm thinking the SSR pr might be better here as it allows us to actually test whether this is SSR-compliant.

E.g. over here there's a `span` within a `div`, which is likely to cause issues.

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
```diff
@@ -0,0 +1,146 @@
...
+  const minimizedPanelStringRegex = '(panel.*minimized.*)';
+
+  // Forked and modified from 'markdown-it/lib/rules_block/html_block.js'
+
+  // An array of opening and corresponding closing sequences for html tags,
+  // last argument defines whether it can terminate a paragraph or not
+  const HTML_SEQUENCES = [
+    // MODIFIED (MarkBind): ignore special tags
+    [ startingSpecialTagRegex, endingSpecialTagRegex, true ],
+    [/^<(script|pre|style)(?=(\s|>|$))/i, /<\/(script|pre|style)>/i, true],
```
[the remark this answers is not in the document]
wxwxwxwx9: My bad! 😓  I thought u were referring to merging with `escape-special-tags` in general. I have fixed it now! Instead of removing that line you highlighted above (which is adapted from `VuePress`), I opted to remove the `script`, `pre`, and `style` from the special tags regex instead as it seemed more appropriate. Let me know if I should do the opposite :-)

packages/vue-components/src/Thumbnail.vue
wxwxwxwx9: Hmm I see where you're coming from. In that case, yeap I agree as well!

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
```diff
@@ -11,7 +11,7 @@ function initCustomComponentHtmlBlockRule(tagsToIgnore) {
    * Custom patch for the api to escape content in certain special tags.
    */
   const specialTagsRegex = Array.from(tagsToIgnore)
-  .concat(['include|script|pre|style|variable|site-nav|markdown|md'])
```
ang-zeyu: 👀 should we remove the later one instead? less regex runs

wxwxwxwx9: Yep sure! I have shifted the `script`, `pre`, and `style` to the back of the array here and made a comment that they are for Vue, just for documentation :-)

packages/cli/test/functional/test_site/expected/index.html
```diff
@@ -608,9 +600,7 @@ <h2 id="panel-with-src-from-another-markbind-site-header-2"><span id="panel-with
             </template></panel>
         </div>
         <p><strong>Modal with panel inside</strong></p>
-        <p>
-          <trigger for="modal-with-panel">trigger</trigger>
-        </p>
+        <trigger for="modal-with-panel">trigger</trigger>
```
ang-zeyu, line 603: missed this previously: let's add in popover / trigger / tooltip to the inline tags list as well

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
```diff
@@ -0,0 +1,147 @@
...
+  const startingSpecialTagRegex = new RegExp(
+  `^<(${specialTagsRegex})${selfClosingNegativeLookahead}(?=(\\s|>|$))`, 'i');
+  const endingSpecialTagRegex = new RegExp(`<\\/(${specialTagsRegex})>`, 'i');
+
+  /*
+  * In MarkBind, we consider minimized panel components to be inline elements.
+  * Thus, we have to ensure that markdown-it parses minimized panel as inline element
+  * by using the regex below. 
+  */
+  const minimizedPanelStringRegex = '(panel.*minimized.*)';
```
ang-zeyu, line 39:
```suggestion
  const minimizedPanelStringRegex = '(panel${attribute}*minimized)';
```
would this be better instead?
- `${attribute}*` preceding `minimized` to ensure its not part of an attribute value
- removal of second `.*` handled since it should be handled by vuepress' regex

let's add a few positive / negative tests for this as well.

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
```diff
@@ -0,0 +1,147 @@
+const blockNames = require('markdown-it/lib/common/html_blocks');
+const { HTML_OPEN_CLOSE_TAG_RE } = require('./htmlRe');
+const inlineTags = require('./inlineTags');
+const vueReservedTags = require('./vueReservedTags');
+
```
ang-zeyu, line 5: let's add the modified source (vuepress) to these files as well so we other devs can refer back easily

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
```diff
@@ -0,0 +1,147 @@
...
+  const HTML_SEQUENCES = [
+    // MODIFIED (MarkBind): ignore special tags
+    [ startingSpecialTagRegex, endingSpecialTagRegex, true ],
+    [/^<!--/, /-->/, true],
+    [/^<\?/, /\?>/, true],
+    [/^<![A-Z]/, />/, true],
+    [/^<!\[CDATA\[/, /\]\]>/, true],
+    // MODIFIED HERE: Treat vue reserved tags as block tags
+    [
+      new RegExp('^</?(' + vueReservedTags.join('|') + ')(?=(\\s|/?>|$))', 'i'),
```
ang-zeyu, line 54: hmm should we keep this? (since we aren't exactly documenting / supporting general vue development)

wxwxwxwx9: Hmm good point. I thought of keeping it as similar to VuePress modified source file as possible but let's make it more adapted to our needs and remove unnecessary stuff.

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
wxwxwxwx9: Should I add the source link to every file? Or is it fine if we just add it to the "main" file, which is `customComponentPlugin`?

ang-zeyu: sorry, missed the one in `index.js`, that should be fine.

The link (https://github.com/vuepress/vuepress-next/tree/main/packages/%40vuepress/markdown/src/plugins/customComponentPlugins) is broken though.

packages/core/src/lib/markdown-it/patches/custom-component/index.js
```diff
@@ -0,0 +1,11 @@
+/*
+ * https://github.com/vuepress/vuepress-next/tree/main/packages/%40vuepress/markdown/src/plugins/customComponentPlugins
+ *
+ * The files in this folder, custom-component, are adapted from the above link to fit MarkBind's needs. 
+ *
+ * It helps us to define Vue custom components (or unknown components) as block-level elements during markdown-it parsing.
```
ang-zeyu, line 6:
```suggestion
 * It helps us to define Vue custom components (or unknown components) as block or inline level elements during markdown-it parsing.
```

could use a mention of the "special tags" patch as well (`<script/style>`)

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
wxwxwxwx9: > The link (https://github.com/vuepress/vuepress-next/tree/main/packages/%40vuepress/markdown/src/plugins/customComponentPlugins) is broken though.

Ah, there's an additional 's' at the end. Thanks for the heads up!

packages/core/src/lib/markdown-it/patches/custom-component/htmlBlockRule.js
wxwxwxwx9: Thanks for the suggestion :-) I went with the former `*minimized.*` because I figured the chances of this edge case is rather slim. However, the regex that you suggested works really well and seems to cover the edge case sufficiently, if not fully. I've added a few test cases using the example that you provided in #958. Let me know if I missed out other edge cases!

packages/cli/test/functional/test_site/expected/requirements/NonFunctionalRequirements._include_.html
```diff
@@ -3,9 +3,7 @@
     <p>In contrast to <strong>Functional Requirements</strong> which specify what the system should do,
       <strong>Non-Functional Requirements</strong> specify the constraints under which system is developed and operated.</p>
   </div>
-  <p>
-    <seg id="preview">Requirements gathering, requirements elicitation, requirements analysis,
-      requirements capture are some of the terms commonly <strong>and</strong> interchangeably used to represent the activity
-      of understanding what a software product should do.</seg>
-  </p>
+  <seg id="preview">Requirements gathering, requirements elicitation, requirements analysis,
```
ang-zeyu, line 6: should we change this to an inline tag (e.g. `<span>`?) to preserve the `<strong>` tag?

packages/core/test/unit/markdown-it-custom-component.test.js
```diff
@@ -0,0 +1,96 @@
...
+    ];
+
+    return source.join('\n');
+  };
+
+  const tag1 = '<panel minimized>'; // should be rendered as inline element
+  const tag2 = '<panel alt="hi" header="hi" minimized>'; // should be rendered as inline element
+  const tag3 = '<panel alt="hi" minimized header="hi">'; // should be rendered as inline element
+  const tag4 = '<panel minimized alt="hi" header="hi">'; // should be rendered as inline element
+  const tag5 = '<panel alt="hi" header="hi">'; // should be rendered as block element
```
ang-zeyu, line 25: would this one work? - `<panel header="test for minimized as part of attribute value">`


I was actually just expecting functional site tests here since they're easier to create/maintain and directly test e2e behaviour 😅, but this is great as well.

But _if you're up for it_, let's move it over somewhere to `functional/test_site`.

packages/cli/test/functional/test_site/expected/requirements/NonFunctionalRequirements._include_.html
wxwxwxwx9: Hmm yep, that would be good. On another note, should we update `<seg>` to something else in the SSR PR since we don't "allow" unknown components?

packages/core/test/unit/markdown-it-custom-component.test.js
wxwxwxwx9: > would this one work? - `<panel header="test for minimized as part of attribute value">`

Yep! The behaviour is as expected. 

> I was actually just expecting functional site tests here since they're easier to create/maintain and directly test e2e behaviour 😅, but this is great as well.
> 
> But _if you're up for it_, let's move it over somewhere to `functional/test_site`.

No problem! Have moved the tests over to `functional/test_site` :-)

packages/cli/test/functional/test_site/expected/index.html
```diff
@@ -314,11 +313,9 @@ <h1 id="heading-with-hidden-keyword"><span id="heading-with-hidden-keyword" clas
           <p><strong>Normal include</strong></p>
           <div>
             <h3 id="establishing-requirements"><span id="establishing-requirements" class="anchor"></span>Establishing Requirements<a class="fa fa-anchor" href="#establishing-requirements" onclick="event.stopPropagation()"></a></h3>
-            <p>
-              <seg id="preview">Requirements gathering, requirements elicitation, requirements analysis,
-                requirements capture are some of the terms commonly <strong>and</strong> interchangeably used to represent the activity
-                of understanding what a software product should do.</seg>
-            </p>
+            <seg id="preview">Requirements gathering, requirements elicitation, requirements analysis,
```
ang-zeyu, line 316: >Hmm yep, that would be good. On another note, should we update <seg> to something else in the SSR PR since we don't "allow" unknown components?

sounds good, over here might be better in fact since you've changed the behaviour of unknown tags

packages/cli/test/functional/test_site/expected/index.html
```diff
@@ -337,7 +334,7 @@ <h4 id="focus-groups"><span id="focus-groups" class="anchor"></span>Focus groups
           </div>
           <p><strong>Include segment</strong></p>
           <div>Requirements gathering, requirements elicitation, requirements analysis,
-            requirements capture are some of the terms commonly <strong>and</strong> interchangeably used to represent the activity
+            requirements capture are some of the terms commonly **and** interchangeably used to represent the activity
```
ang-zeyu, line 337: ^

packages/cli/test/functional/test_site/expected/index.html
```diff
@@ -458,53 +455,48 @@ <h2 id="feature-list"><span id="feature-list" class="anchor"></span>Feature list
...
           <panel src="/test_site/requirements/SpecifyingRequirements._include_.html#preview" type="minimal" fragment="preview"><template #header>
               <p><strong>same test with panels</strong></p>
             </template></panel>
-          </p>
-          <p><strong>Include a file in a sub-folder that uses baseUrl</strong>
+          <p><strong>Include a file in a sub-folder that uses baseUrl</strong></p>
           <div>
             <div>Requirements gathering, requirements elicitation, requirements analysis,
-              requirements capture are some of the terms commonly <strong>and</strong> interchangeably used to represent the activity
+              requirements capture are some of the terms commonly **and** interchangeably used to represent the activity
```
ang-zeyu, line 469: ^ and below

packages/cli/test/functional/test_site/expected/testIncludeMultipleModals.html
```diff
@@ -186,14 +186,14 @@ <h3 id="testing-site-nav"><span id="testing-site-nav" class="anchor"></span>Test
         <div>
           <b-modal id="modal:bugRepro" hide-footer size="lg" modal-class="mb-zoom" ref="modal:bugRepro"><template #modal-title>Establishing Requirements</template>
             <div>Requirements gathering, requirements elicitation, requirements analysis,
-              requirements capture are some of the terms commonly <strong>and</strong> interchangeably used to represent the activity
+              requirements capture are some of the terms commonly **and** interchangeably used to represent the activity
```
ang-zeyu, line 189: ^ and below

packages/cli/test/functional/test_site/testPanelMarkdownParsing.md
```diff
@@ -0,0 +1,55 @@
...
+<panel minimized>
+line 1 which behaves as per normal ( empty line still counts )
+
+lines 3 onward are 'placed' outside of the component as it is a minimized panel (inline element)
+...
+...
+</panel>
+
+markdown pre text **immediately** prepending a component, without an empty line after
+<panel alt="test" header="test minimized placed extreme left" minimized>
```
ang-zeyu, line 31:
```suggestion
<panel alt="test" header="test minimized placed extreme right" minimized>
```

and the one below

ang-zeyu (commented):
couple more:

ang-zeyu (commented):
Lgtm! 👍

ang-zeyu (approved):

ang-zeyu:
Remark: this PR introduces a lot of breaking changes in the form of properly defined block/inline behaviour. (should be a step in the right direction irregardless of SSR or not)

Issue #1543, opened 2021-04-11
Enhance regeneration of pages during live preview


jonahtanjz:
**Is your request related to a problem?**

A follow up to discussions in #1514 and #1529.

When a page is added or removed, all pages are rebuilt if the added/removed file is a page/dependency of page. Similarly, a change in page's properties will cause dependent pages to be rebuilt although this wouldn't affect other pages. 

For context: 
https://github.com/MarkBind/markbind/pull/1514#pullrequestreview-614438432
https://github.com/MarkBind/markbind/pull/1514#discussion_r596153844
https://github.com/MarkBind/markbind/pull/1529#discussion_r608436801

**Describe the solution you'd like**

1. Only the pages that are dependent on the added/removed pages should be rebuilt.
2. Only the pages which's properties have changed in `site.json` should be rebuilt. (and even then if using `markbind s -o` only the page being viewed should be rebuilt)
3. When the attributes in the site configuration (`site.json`) are modified, only the affected pages should be rebuilt.

PR #1550, opened 2021-04-22
Lazy serve: Rebuild current opened pages on add/remove page

ryoarmanda:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Follow up on #1513 (ref: https://github.com/MarkBind/markbind/pull/1513#issuecomment-818603397).

After multi-tab development is established, the add/remove page flow has yet to be updated to use the current opened pages. Currently it is still rebuilding only the current viewed page.

**Overview of changes:**
- Modified `rebuildPageBeingViewed` to adopt the page generation tasks flow
- Changed the call of `rebuildPageBeingViewed` on `rebuildSourceFiles` to use `Site.currentOpenedPages`

Another problem arises, as on `rebuildSourceFiles`, the line after is a call to `lazyBuildAllPagesNotViewed`, which tests the pages against `Site.currentPageViewed`, so `toRebuild` will be refilled with almost every page, which will trigger another rebuild once live-reload starts. So, I also:

- Modified `lazyBuildAllPagesNotViewed` to use a supplied `viewedPages` and test the pages against that instead
- Modified any call to the above function appropriately, e.g. on `lazyBuildSourceFiles` to only use the current viewed page, on `rebuildSourceFiles` to use the current opened pages.

- This PR only aims to adapt the add/remove flow to use the current opened pages, so the behaviour of add/remove flow is still kept as is. That means, the issue stated in #1543 still exists for now.

**Testing instructions:**

With the default template from `markbind init`:
- `markbind serve -o`
- Aside from the index page, open another page on a new tab
- Add a new page (easy way is to duplicate one of the files in `contents`)
- Observe in the logs that both the pages opened are rebuilt

Lazy-serve: Rebuild opened pages on add/remove page

With the addition of the multi-tab development flow, single-page serve
can keep track of all opened pages instead of a single currently
viewed page. This flow is already adopted in page edit events,
but it has yet to be adopted in page addition or removal.

Let's adopt the multi-tab development mechanism on the page addition' +
and removal flow.

---

### DIFF
--- a/packages/core/src/Site/index.js
+++ b/packages/core/src/Site/index.js
@@ -747,26 +747,23 @@ class Site {
      to trigger multiple page builds before the first one has finished building,
      hence we need to take this into account.
      */
-    const regeneratePagesBeingViewed = uniqueUrls.map(async (normalizedUrl) => {
+
+    const pagesToRebuild = this.pages.filter(page =>
+      uniqueUrls.some(pageUrl => FsUtil.removeExtension(page.pageConfig.sourcePath) === pageUrl));
+    const pageGenerationTask = {
+      mode: 'async',
+      pages: pagesToRebuild,
+    };
+
+    try {
       this._setTimestampVariable();
-      const pageToRebuild = this.pages.find(page =>
-        FsUtil.removeExtension(page.pageConfig.sourcePath) === normalizedUrl);
+      await this.runPageGenerationTasks([pageGenerationTask]);
+      await this.writeSiteData();
+      Site.calculateBuildTimeForRebuildPageBeingViewed(startTime);
+    } catch (err) {
+      await Site.rejectHandler(err, [this.tempPath, this.outputPath]);
+    }
 
-      if (!pageToRebuild) {
-        return;
-      }
-
-      this.toRebuild.delete(normalizedUrl);
-      try {
-        await pageToRebuild.generate(this.externalManager);
-        await this.writeSiteData();
-        Site.calculateBuildTimeForRebuildPageBeingViewed(startTime);
-      } catch (error) {
-        await Site.rejectHandler(error, [this.tempPath, this.outputPath]);
-      }
-    });
-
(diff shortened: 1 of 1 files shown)


packages/core/src/Site/index.js
```diff
@@ -762,26 +763,23 @@ class Site {
      to trigger multiple page builds before the first one has finished building,
```
ang-zeyu, line 763: let's move this comment to the original call site of the function since the use case is expanded

packages/core/src/Site/index.js
```diff
@@ -853,8 +851,8 @@ class Site {
     if (this.onePagePath) {
       this.mapAddressablePagesToPages(this.addressablePages || [], this.getFavIconUrl());
 
-      await this.rebuildPageBeingViewed(this.currentPageViewed);
-      await this.lazyBuildAllPagesNotViewed();
+      await this.rebuildPageBeingViewed(this.currentOpenedPages);
```
ang-zeyu, line 854: should we do:
```suggestion
      await this._rebuildPagesBeingViewed(this.currentOpenedPages);
```
?
- plural for the expanded use case
- `_` to use the non-delayed version (possible here?)

ang-zeyu (commented):
Thanks for the fix! Tested and works 👍

Just some nits:

packages/core/src/Site/index.js
ryoarmanda: Agreed, I'll refactor the method (and related helpers) to use plural name, and use the non-delayed version on `rebuildSourceFiles` (as calls to it is already delayed)

packages/core/src/Site/index.js
ryoarmanda: Right, moving this to the call site at `changeCurrentPage`

ang-zeyu (approved):
Lgtm! 👍

PR #1556, opened 2021-04-30
Use portals for mobile site and page nav

ang-zeyu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

4th item in https://github.com/MarkBind/markbind/pull/1534#issuecomment-827530954

**Overview of changes:**

Change the retrieval to use portals https://portal-vue.linusb.org/guide/getting-started.html
- Setup portal sources at build time (for `<site-nav> / {{pageNav}} / id="site/page-nav"`)
   - These are transformed into an `OverlaySource` (new) vue component
- Change `Overlay` retrieval to use portals instead, for `<site-nav> / {{pageNav}} / id="site/page-nav"`

**Testing instructions:**
Manual testing needed

Use portals for mobile site and page nav

With Vue SSR implemented, site and page nav mobile overlays are unable
to retrieve the original page source describing the respective
navigation components.
This results in being unable to mount an interactive copy of said
components.

Let's change the mobile site and page nav overlays to use portals as
the content source instead.
Portal sources (site / page nav) are identified and created during
build time.

---


ang-zeyu:
@wxwxwxwx9 @jonahtanjz your reviews would be great here! Especially on possible alternative approaches.

Per https://github.com/MarkBind/markbind/pull/1534#issuecomment-827530954, the old retrieval method won't work with ssr as the page is ssr-rendered (template / app information is lost).

The overall approach is changed to use portals, powered by the portal-vue library.

The main upside would be no network requests => html parsing, etc.
The main downside would be more backend coupling.

Some alternatives considered: 
- Duplicate the html node during build time, then use `retriever.vue` as before to retrieve that.
   -  should be straightforward to implement, but still results in some form of backend coupling. Also, increased html size.
- Storing the site/page navs in separate files, then likewise using `retriever.vue` to retrieve
   - backend coupling, file bloat.

Not too sure if there are methods to do this without backend coupling 🤔

packages/vue-components/src/PageNavButton.vue
```diff
@@ -19,30 +16,16 @@ import $ from './utils/NodeList';
...
-      hasIdentifier: false,
+      portalName: undefined,
       show: false,
-      src: '',
     };
   },
   computed: {
     showPageNav() {
-      return this.show && this.hasPageNav;
+      return this.show && this.portalName;
```
jonahtanjz, line 25: I believe we need a separate variable to check for page nav as the identifier `page-nav` can exist without a page nav. 
For exmaple,
```
  <nav id="page-nav" class="fixed-header-padding">
    <div class="nav-component slim-scroll">
      {{ pageNav }}
    </div>
  </nav>
```
The `nav` element will always be present even if `{{pageNav}}` is empty.

This is causing the `page-nav-button` to show up on [pages](https://deploy-preview-1556--markbind-master.netlify.app/userguide/gettingstarted) without page nav.

image

packages/vue-components/src/PageNavButton.vue
```diff
@@ -19,30 +16,16 @@ import $ from './utils/NodeList';
...
-      const pageNavTitle = document.getElementsByClassName('page-nav-title')[0];
-      const pageNavLinks = document.getElementById('mb-page-nav');
-      if (pageNavTitle) {
-        wrapper.appendChild(pageNavTitle.cloneNode(true));
-      }
-      if (pageNavLinks) {
-        wrapper.appendChild(pageNavLinks.cloneNode(true));
-      }
-      return wrapper;
-    },
```
jonahtanjz, line 45: I believe this method is still needed so that we can still pull in the page nav if the `page-nav` identifier is not present? As discussed [here](https://github.com/MarkBind/markbind/pull/1445#issuecomment-764419350), we should pull in the identifier first if they exist and if not, we should find the page nav and then pull that in.

Alternatively, we can also change the requirement such that users will have to explicitly specify the page nav with the identifier if that makes it better?

packages/vue-components/src/PageNavButton.vue
```diff
@@ -52,9 +35,12 @@ export default {
     },
   },
   mounted() {
-    this.src = window.location.pathname;
-    this.hasIdentifier = document.getElementById('page-nav') !== null;
-    this.hasPageNav = document.getElementById('mb-page-nav') !== null;
+    if (document.getElementById('page-nav') !== null) {
+      this.portalName = 'page-nav';
+    } else if (document.getElementById('mb-page-nav') !== null) {
```
jonahtanjz, line 40: As mentioned above, `page-nav` if present will always exist. As `mb-page-nav` will only be present if a page nav is present, we can use this to check if the `page-nav-button` should be shown.

packages/vue-components/src/SiteNavButton.vue
```diff
@@ -19,21 +16,16 @@ import $ from './utils/NodeList';
...
-      hasSiteNav: false,
+      portalName: undefined,
       show: false,
-      src: '',
     };
   },
   computed: {
     showSiteNav() {
-      return this.show && this.hasSiteNav;
+      return this.show && this.portalName;
```
jonahtanjz, line 25: Similar to page nav

packages/vue-components/src/SiteNavButton.vue
```diff
@@ -19,21 +16,16 @@ import $ from './utils/NodeList';
...
     },
   },
   methods: {
-    getSiteNavContent() {
-      return document.getElementsByClassName('site-nav-root')[0];
-    },
```
jonahtanjz, line 36: Similar to page nav (Should pull the first site nav if the identifier is not present).

packages/vue-components/src/SiteNavButton.vue
```diff
@@ -43,9 +35,12 @@ export default {
     },
   },
   mounted() {
-    this.src = window.location.pathname;
-    this.hasIdentifier = document.getElementById('site-nav') !== null;
-    this.hasSiteNav = document.getElementsByClassName('site-nav-root').length !== 0;
+    if (document.getElementById('site-nav') !== null) {
+      this.portalName = 'site-nav';
+    } else if (document.getElementsByClassName('site-nav-root').length !== 0) {
```
jonahtanjz, line 40: Similar to page nav

packages/vue-components/src/Overlay.vue
```diff
@@ -65,30 +49,20 @@ export default {
...
-    const buildNav = (navMenuItems) => {
-      if (!navMenuItems) { return; }
-      for (let i = 0; i < navMenuItems.childNodes.length; i += 1) {
-        navMenu.appendChild(navMenuItems.childNodes[i].cloneNode(true));
-      }
-      this.navMenuLoaded();
-    };
-
-    if (!this.hasIdentifier) {
-      buildNav(this.getNavMenuContent());
```
jonahtanjz, line 91: Similar to the comment for page nav, we will need this to manually pull in the first site nav or the page nav if their identifiers are not present.

jonahtanjz (commented):
Thanks @ang-zeyu!

Just a brief review. I believe using portals is a good approach. I do not have any objection with that. I've noticed some regressions with the current implementation regarding the site and page navigation.

ang-zeyu:
Thanks for the speedy review! @jonahtanjz 

ahh, my bad, I understand why that part of the code was structured as such now 🤣; I also missed the root issue with https://github.com/MarkBind/markbind/pull/1445#discussion_r571365872 you brought up previously.

I went with the spec described in the [docs](https://markbind-master.netlify.app/userguide/usingcomponents#navbars) and the [original post](https://github.com/MarkBind/markbind/pull/1445#issuecomment-764419350) in implementing it this time.

i.e. using MarkBind's auto generated `{{ pageNav }}` isn't necessary, only `<some-element id="page-nav">`. (to allow more customisation)

---

> This is causing the page-nav-button to show up on pages without page nav.

One option is to create new layouts in our docs without the empty `id="page-nav"` element, following the current spec. This seems like the most "correct" option but is a lot of work for the author (need to maintain and configure two layout sets)

Another is to add this to the spec. i.e. `id="page-nav"` must also contain `{{ pageNav }}`. This is convenient for the author, but might be a little restrictive.

> Alternatively, we can also change the requirement such that users will have to explicitly specify the page nav with the identifier if that makes it better?

This is also an option, and can simplify the implementation quite a bit. The downside is the user would then have to be aware of the additional need for `id="page-nav"`. Also, still need to maintain and configure two layout sets.

Likewise for the site-nav.

What do you think? @jonahtanjz

packages/vue-components/src/PageNavButton.vue
ang-zeyu: > I believe this method is still needed so that we can still pull in the page nav if the page-nav identifier is not present? As discussed here, we should pull in the identifier first if they exist and if not, we should find the page nav and then pull that in.

I standardised this to a portal implementation as well

ang-zeyu:
Case-by-case illustration:

Option 1: (my interpretation)

1. `id="page-nav"` contains `{{ pageNav }}` - **mobile page nav shown**
2. `id="page-nav"` is present but does not contain `{{ pageNav }}` - **shown**
3. `id="page-nav"` is absent, but `{{ pageNav }}` is present - **shown**

More work in that need to configure two layout sets.

Option 2: (your interpretation)

1. `id="page-nav"` contains `{{ pageNav }}` - **shown**
2. `id="page-nav"` is present but does not contain `{{ pageNav }}` - **not shown**
3. `id="page-nav"` is absent, but `{{ pageNav }}` is present - **shown**

More restrictive in that user must use MarkBind's auto generated `{{ pageNav }}`. Less work as only 1 layout is needed.

Option 3: (new suggestion above)

1. `id="page-nav"` contains `{{ pageNav }}` - **shown**
2. `id="page-nav"` is present but does not contain `{{ pageNav }}` - **shown**
3. `id="page-nav"` is absent, but `{{ pageNav }}` is present - **not shown**

Need to configure two layout sets. Simpler, implementation wise. Slightly less convenient (author must be aware of `id="page-nav"` restriction).

damithc:
@ang-zeyu is there any change from the user's POV? Is there a corresponding update to the user guide?

ang-zeyu:
The current PR goes by the user guide's description (option 1), but will need a restructure of the layouts (e.g. below).
i.e. separate layouts / configs for pages with or without pagenav is required.

e.g. (option 1)
```
// layoutWithPageNav.md
<include src="base.md">
<variable name="doAddPageNav>true</variable>
</include>

// layoutWithoutPageNav.md
<include src="base.md" />

//base.md
...
{% if doAddPageNav %}
<div id="site/page-nav">
... the mobile page / site-nav button shows up if this id is present ... regardless of what's contained inside
</div>
{% endif %}
...
```

The netlify preview [here](https://deploy-preview-1556--markbind-master.netlify.app/userguide/gettingstarted) shows what happens if one layout is stuck to -- mobile page nav button is still present even with an empty `{{ pageNav }}`.

The previous implementation in #1445 abides by option 2 (I missed while reviewing the earlier PR); That will need a ug update of the feature description.

It is more convenient however; Case in point the entire user guide is currently using only one layout file (https://raw.githubusercontent.com/MarkBind/markbind/master/docs/_markbind/layouts/userGuide.md). This is possible because empty `{{ pageNav }}` renders into nothing, and the implementation detects that as "page nav is not present".

The downside is MarkBind's site/auto generated page-nav (`<site-nav> / {{ pageNav }}`) must be present.

ang-zeyu:
One more alternative: Check if the `id="page/site-nav"` contains any **text**, if so, the mobile page/site-nav is shown.

This should bring the best of both options 1 & 2.
The obvious downside is that `id="page/site-nav"` must contain some text (which should be true 99% of the time), and this definition might be a little technical for the user

damithc:
Not sure if I understand the picture fully. I'll leave you guys to consider the author's POV as well and choose the best option.

jonahtanjz:
@ang-zeyu Option 1 seems to allow users to have more control and flexibility but also more things for them to take note of. This may be a little inconvenient on the user side?

> One more alternative: Check if the id="page/site-nav" contains any text, if so, the mobile page/site-nav is shown.
>
> This should bring the best of both options 1 & 2.
> The obvious downside is that id="page/site-nav" must contain some text (which should be true 99% of the time), and this definition might be a little technical for the user

This alternative sounds good as it reduces the need for the user to do extra work. I think the benefits of abstracting this process from the user exceed the downside (since it only occurs in rare situation and if we are clear in the ug, should be sufficient).

ang-zeyu:
Agreed!

I wonder if there's an even more intuitive solution than looking for text though. 🤔

Documenting `for your mobile site / page navigation menu to show up, it must contain some text` may seem very bewildering to the user.

It also dosen't take care of such cases nicely. In this case option 2 might be even better:
```
<nav id="page-nav">
  <h2>My custom page nav title!</h2>
  <img src="..." alt="accompanying image" />
  {{ pageNav }}
</nav>
```

Should we look `<a />` tags instead?

jonahtanjz:
> Should we look `<a />` tags instead?

I think this is the most intuitive approach as a site nav and page nav should contain anchor tags. Having a mobile site nav or page nav that contains no links will not add much value to the users and will most likely not make a big difference if it is hidden from them.

> So the documentation becomes:
> `If your site or page nav dosen't contain any anchor tags, MarkBind will automatically hide them for you.`
> Might still be a little odd though.

How about <code>If your site or page nav dosen't contain any <del>anchor tags</del> links, MarkBind will automatically hide them for you.</code>?
This will probably make it easier for the users to understand and also let them know that it works with Markdown's link syntax as well (as these will be converted to `<a />`).

ang-zeyu:
👍, let's go with this then.

> How about `If your site or page nav dosen't contain any ~anchor tags~ links, MarkBind will automatically hide them for you.`?
> This will probably make it easier for the users to understand and also let them know that it works with Markdown's link syntax as well (as these will be converted to `<a />`).

oops, meant `links` 😅

I'll put this change up in a separate PR however so its clear this PR / commit is just for changing the background implemention.
It should be just be some extra checks surrounding these:

```
if (document.getElementById('site-nav') !== null && siteNavIdHasAnchorMagic) {
      this.portalName = 'site-nav';
} else if (document.getElementsByClassName('site-nav-root').length !== 0 && notNeeded) {
...

// similar for the page-nav branch
```

I'll add back in the `{{ pageNav }} / <site-nav>` checks in a while as well. (should be at the same location - `siteNavIdHasAnchorMagic` -> `siteNavIdHasSiteNavComponentMagic`)

packages/vue-components/src/SiteNavButton.vue
ang-zeyu: also now using the portal implementation (more important here as `<site-nav>` can contain interactive vue components)

packages/vue-components/src/PageNavButton.vue
ang-zeyu: reverted temporarily

packages/vue-components/src/PageNavButton.vue
ang-zeyu: reverted temporarily

packages/vue-components/src/SiteNavButton.vue
ang-zeyu: also reverted temporarily

packages/vue-components/src/Overlay.vue
```diff
@@ -65,30 +49,20 @@ export default {
         publish('closeOverlay');
         // to prevent scrolling of the body when overlay is overscrolled
         document.body.style.overflow = 'hidden';
+        this.show = true;
+
+        this.$nextTick(() => {
+          $(this.$refs.navMenuContainer).find('a').on('click', this.toggleNavMenu);
+        });
```
jonahtanjz, line 56: I've also noticed this strange behaviour where after clicking on a page nav's link, the overlay does not close.

For example on the [Using Components](https://deploy-preview-1556--markbind-master.netlify.app/userguide/usingcomponents) page:

[](https://user-images.githubusercontent.com/47470981/116770225-6e656880-aa74-11eb-9d3c-dd3cda9bb8dc.mp4)

Notice how after clicking on some of the links, the url changes but the overlay does not close (only after 2-3 clicks then it closes). I'm not sure if this method is causing this issue.

jonahtanjz (commented):
Just found a few more edge cases which may be causing some issues.

ang-zeyu:
Thanks for the catches!

I've fixed both issues -- the first was a problem with where `overlay-source` for the auto page nav should be added (fixed by shifting), the second was due to `toggleNavMenu` firing multiple times from the event handler (over-attached) (fixed by using the bubbled event instead).

---

I found another undiscussed point with the previous implementation around the forwarding of styles / attributes however.
Based on https://markbind-master.netlify.app/userguide/formattingcontents#line-breaks, the current behaviour is:
- Tag names / classes / styles / attributes / ids are not forwarded for `id="site-nav/page-nav"`, resulting in slightly different styling.
- the auto generated page nav + site nav components are, however.


Adding 55cf54d _alone_ which forwards said things makes the site nav go poof, as there is a media query for `display: none` (`<998....px`).
Adding reasonable overrides (619d775) fixes this to some extent.


I wonder if we should
- Leave the current behaviour be, then update the ug with a warning to reflect this behaviour - since it would break some descendant selectors / the selected element the author wants to carry forward.
- Or adopt 55cf54d & 619d775 in another PR, then update the ug with this behaviour. The author would then be expected to add extra styles with the `.override` class if he/she wants to override our reasonable defaults.

This can really go both ways.

e.g. **With forwarding**, our user guide adds a thin grey bar on the right as expected, but might not be suitable for the mobile site nav. The author then has to override this by compositing `#site-nav.override { ... }`



**Without**, it breaks descendent selectors and the actual element's selectors.
I would think its more difficult to add separate styles for the mobile version as well because the root element becomes "generic" (just a `<div>` without tag, id, etc. information)

Wdyt? @jonahtanjz

jonahtanjz:
> I wonder if we should
>
> * Leave the current behaviour be, then update the ug with a warning to reflect this behaviour - since it would break some descendant selectors / the selected element the author wants to carry forward.
> * Or adopt 55cf54d & 619d775 in another PR, then update the ug with this behaviour. The author would then be expected to add extra styles with the .override class if he/she wants to override our reasonable defaults.

I think going with forwarding would make more sense as this will preserve as much of the original navigations as possible. We can add more default overrides if needed (so normal users would need to do as little work as possible) and the more advance users can customise the styling etc.

I feel if we go without forwarding it will make it too restrictive and also break some of the selectors as mentioned. That's just my thought on this.

> e.g. With forwarding, our user guide adds a thin grey bar on the right as expected, but might not be suitable for the mobile site nav. The author then has to override this by compositing #site-nav.override { ... }

We can also remove this by default.

packages/vue-components/src/Overlay.vue
```diff
@@ -60,35 +47,21 @@ export default {
     };
   },
   methods: {
-    toggleNavMenu() {
-      if (!this.show) {
+    toggleNavMenu(ev) {
+      if (ev.target.tagName.toLowerCase() === 'a' || this.show) {
```
jonahtanjz, line 51:
```suggestion
      if (ev.target.tagName.toLowerCase() === 'a' && this.show) {
```
I think this should be `&&` so that the site nav title does not close the overlay when it is clicked?

[](https://user-images.githubusercontent.com/47470981/116780910-05044a80-aab2-11eb-9079-9c000c31d89a.mp4)

ang-zeyu: Fixed! Still `||`, but added some extra checks, `&&` would disable the closing button.

ang-zeyu:
>I feel if we go without forwarding it will make it too restrictive and also break some of the selectors as mentioned. That's just my thought on this.

👍, that's my preference as well. I've reverted those two commits for moving to another pr.

The auto generated site/pagenav without the wrapper `id=site/page-nav` will temporarily look a little strange pending this forwarding fix.

docs/_markbind/layouts/userGuide.md
```diff
@@ -2,6 +2,7 @@
 
 <div id="flex-body">
   <nav id="site-nav" class="fixed-header-padding">
+<panel header="To test interactivity">Expanding and closing should work</panel>
```
jonahtanjz, line 5: Should this be removed?

jonahtanjz (commented):
Just one more question. Other than that LGTM 👍

docs/_markbind/layouts/userGuide.md
ang-zeyu: almost forgot 😅

jonahtanjz (approved):
LGTM 👍

PR #1565, opened 2021-05-07
Hoist `<script>` and `<style>` tags to after `<body>` tag

wxwxwxwx9:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

This addresses the seventh item in [#1534 (comment)](https://github.com/MarkBind/markbind/pull/1534#issuecomment-827530954).

**Overview of changes:**
- Removed unescaping of special characters after SSR
- Combined processing of script and style tags
- Hoist script and style tags found during the processing of page / layout to the bottom of the page

If there's a better way to implement this and if I missed out anything.

**Testing instructions:**
`npm run test`

Create a new MarkBind project and try adding `<style>` and/or `<script>` tags in either the layout page or a normal page. 

You should find that regardless of layout page or a normal page, all `<style>` and `<script>` will be hoisted to the bottom of the rendered HTML file (when you view the source tab in developer console). 

Hoist `<script>` and `<style>` tags to after `<body>` tag.

Our setup compiles whatever is in the content section of the page into
Vue application. We also do server-side rendering on the content. Within
the content section, we allow users to write `<script>` and/or `<style>` 
tags. However, during compilation, Vue will ignore and discard these tags.
Furthermore, during server-side rendering, special characters in 
`<script>` tags will be escaped. While we have measures to tackle the 
problems above, we should look to implement a better workaround.

Let's remove the temporary measures and prevent the compilation and
rendering of `<script>` and/or `<style>`, by hoisting them out of the 
content section and to after `<body>` tag.

---


packages/cli/test/functional/test_site/expected/bugs/index.html
```diff
@@ -216,6 +216,9 @@ <h1 id="heading-in-footer-should-not-be-indexed"><span id="heading-in-footer-sho
     </div>
   </div>
 </body>
+
+<script src="/test_site/headFiles/customScriptTop.js"></script>
```
ang-zeyu, line 220: this one's a feature https://markbind-master.netlify.app/userguide/tweakingthepagestructure#inserting-content-into-the-head

packages/cli/test/functional/test_site_convert/expected/Home.html
```diff
@@ -73,6 +73,8 @@
     </footer>
   </div>
 </body>
+
```
ang-zeyu, line 76: You could use https://mozilla.github.io/nunjucks/templating.html#whitespace-control to remove the blank line diffs + make the html a little tighter.

packages/core/src/html/scriptAndStyleTagProcessor.js
```diff
@@ -1,44 +1,23 @@
...
-    The defer attribute requires the src attribute to work. Thus, the src attribute is used as a dummy.
-  */
-  node.attribs.src = '';
-  node.attribs.defer = '';
-  node.attribs.type = 'application/javascript'; // to bypass Vue compilation as a script tag
-  node.attribs['style-bypass-vue-compilation'] = ''; // to allow specific query selection of this element
+function processScriptAndStyleTag(node, userScriptsAndStyles) {
+  const idx = node.parent.children.indexOf(node);
+  if (idx !== -1) {
+    node.parent.children.splice(idx, 1);
```
ang-zeyu, line 15: just to double check -- does this cause issues with sibling traversal?

```
<script>...</script>
<panel>this actually gets processed</panel>
```

packages/core/src/html/NodeProcessor.js
```diff
@@ -190,10 +191,9 @@ class NodeProcessor {
         this.footnoteProcessor.processMbTempFootnotes(node);
         break;
       case 'script':
-        processScriptTag(node);
-        break;
       case 'style':
-        processStyleTag(node);
+        processScriptAndStyleTag(node, this.docId === 'layout'
+          ? this.config.layoutUserScriptsAndStyles : this.config.pageUserScriptsAndStyles);
```
ang-zeyu, line 196: how about passing a bucket `[]` from `Layout/Page`? So this `NodeProcessor` dosen't have to be aware of how it's being used

packages/core/src/Page/PageVueServerRenderer.js
```diff
@@ -76,18 +69,7 @@ async function renderVuePage(compiledVuePage) {
...
-   * https://github.com/vuejs/vue/blob/dev/packages/vue-server-renderer/basic.js
-   *
-   * vue-server-renderer automatically escapes various characters during rendering for security reasons.
-   * We un-escape these characters as there shouldn't be any security concerns in our use-case.
-   *
-   * We only un-escape & and " for now so that the script redirect for userGuide/devGuide will work.
-   * Un-escaping < and/or > will lead to hydration issues so we do not do it for now.
-   */
-  renderedContent = renderedContent.replace(/&(quot|amp);/g, match => vueRenderEscapeChars[match]);
+  const renderedContent = await renderToString(VueAppPage);
```
ang-zeyu, line 72: minor nit: could inline this return value since its rather straightforward

ang-zeyu (commented):
Thanks! @wxwxwxwx9

packages/core/src/html/NodeProcessor.js
ang-zeyu: (or maintain the bucket here and pull it out in `Layout/Page`)

packages/cli/test/functional/test_site/expected/bugs/index.html
wxwxwxwx9: Thanks for the spot!

packages/core/src/html/NodeProcessor.js
wxwxwxwx9: Thanks for the suggestion! This way is much cleaner, not sure why it didn't occur to me earlier 😕

packages/core/src/html/scriptAndStyleTagProcessor.js
wxwxwxwx9: Thanks for bringing up this point! As discussed offline, it should be fine as our tests are currently passing.

wxwxwxwx9:
Hi @ang-zeyu! I've made the changes according to your review. Let me know if anything still looks off :-)

packages/core/src/html/scriptAndStyleTagProcessor.js
```diff
@@ -1,44 +1,30 @@
...
-  */
-  node.attribs.src = '';
-  node.attribs.defer = '';
-  node.attribs.type = 'application/javascript'; // to bypass Vue compilation as a script tag
-  node.attribs['style-bypass-vue-compilation'] = ''; // to allow specific query selection of this element
+function processScriptAndStyleTag(node, userScriptsAndStyles) {
+  // Do not process script tags that are meant to be inserted in head/bottom of HTML
+  const isHeadOrBottomScript = node.parent.name === 'head-top'
+    || node.parent.name === 'head-bottom' || node.parent.name === 'script-bottom';
+  if (isHeadOrBottomScript) {
```
ang-zeyu, line 16: might need to exclude `External`s from this as well since they are loaded on-demand. (should be no issue as they don't need to be ssr-rendered / pre vue-compiled)

ang-zeyu (commented):
To double check -- do the plugins using _"inline"_ `<script>` tags still work? (disqus / copy code block / algolia)

packages/core/src/html/scriptAndStyleTagProcessor.js
wxwxwxwx9: > To double check -- do the plugins using _"inline"_ `<script>` tags still work? (disqus / copy code block / algolia)

Yup, I've tested those plugins and they seem to be working fine for me. 

However, I did notice one problem with Disqus plugin regarding SSR. Having `<disqus/>` will cause hydration issues because it is not a registered component for Vue. And as discussed before, any unrecognized HTML tags will cause hydration issues -- I will look into how I can fix this and make another PR to rectify it :-)

> might need to exclude Externals from this as well since they are loaded on-demand. (should be no issue as they don't need to be ssr-rendered / pre vue-compiled)

I've made use of `docId` to ensure that we will not process `External` scripts but I'm not sure if it's the best approach. I'm thinking if I can make a new parameter for the instantiation of `NodeProcessor`. This parameter will indicate where the `NodeProcessor` was instantiated (e.g. `Page`, `Layout`, or `External`) and we can use it to determine whether we should process the `<script>` tags. However, using `docId` would suffice actually. What do you think?

Also, just to make sure I'm understanding correctly, things would still work fine even if we process the `External` scripts, right? But just to ensure things are "done properly", we would not process the `External` scripts because they will not affect SSR (so no point processing them).

ang-zeyu: > And as discussed before, any unrecognized HTML tags will cause hydration issues -- I will look into how I can fix this and make another PR to rectify it :-)

👍 

> This parameter will indicate where the NodeProcessor was instantiated (e.g. Page, Layout, or External) and we can use it to determine whether we should process the <script> tags. However, using docId would suffice actually. What do you think?

Since you're already passing the bucket `[]`, how about passing `undefined` in place of it and checking for that?

> Also, just to make sure I'm understanding correctly, things would still work fine even if we process the External scripts, right?

It would break for "externals" as the script / style tags simply disappear (collected but not used / hoisted)

e.g. a simple
```
<script>alert('hello world')</script>
```
would simply disappear from the external generated as `..._include_.html`.

ang-zeyu (approved):
Lgtm!

PR #1569, opened 2021-05-30
Resolve SSR Hydration Issue for Disqus plugin

wxwxwxwx9:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

The hydration issue that Disqus plugin has is discovered in #1565. This PR will address the discovered hydration issue for Disqus plugin. 

**Overview of changes:**
Processed `<disqus>` HTML element to be added with `v-pre` in `NodeProcessor`.

- Whether my approach will cause issues that I did not consider
- Whether there is potentially other unknown HTML elements that can cause hydration issues.

**Testing instructions:**
Use Disqus plugin and insert `<disqus/>` HTML element in the page. Afterwards, serve the page by `markbind serve -d`. There should not be any warnings of hydration issue in the developer console.

Resolve SSR Hydration Issue for Disqus plugin.

`<disqus>` HTML tag is an unknown element to Vue as it is neither
a registered Vue component nor an official HTML element. This causes 
hydration issue in our SSR setup whenever the Disqus plugin is used.

Let's add `v-pre` tag to `<disqus>` HTML tag so that Vue will ignore
the tag when compiling/rendering, thus resolving hydration issue.

---

### DIFF
--- a/packages/core/src/html/NodeProcessor.js
+++ b/packages/core/src/html/NodeProcessor.js
@@ -198,7 +198,6 @@ class NodeProcessor {
       case 'style':
         processScriptAndStyleTag(node, this.userScriptsAndStyles);
         break;
-      case 'disqus':
       case 'code':
       case 'annotation': // Annotations are added automatically by KaTeX when rendering math formulae.
       case 'eq': // markdown-it-texmath html tag
--- a/packages/core/src/html/NodeProcessor.js
+++ b/packages/core/src/html/NodeProcessor.js
@@ -198,6 +198,7 @@ class NodeProcessor {
       case 'style':
         processScriptAndStyleTag(node, this.userScriptsAndStyles);
         break;
+      case 'disqus':
       case 'code':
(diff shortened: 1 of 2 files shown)


packages/core/src/html/NodeProcessor.js
```diff
@@ -194,6 +194,7 @@ class NodeProcessor {
       case 'style':
         processScriptAndStyleTag(node, this.userScriptsAndStyles);
         break;
+      case 'disqus':
```
ang-zeyu, line 197: Thanks! How about just renaming the tag? So this fix lives strictly in the plugin

wxwxwxwx9: @ang-zeyu Thanks for the suggestion! We should really avoid unnecessary coupling. I've made the changes to avoid the coupling. 

Regarding "renaming the tag", I'm not sure if we are thinking of the same thing. Let me know if you actually had another idea in mind!

ang-zeyu (approved):
Lgtm!

PR #1572, opened 2021-06-05
Page nav as element

ang-zeyu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

Standardise this syntax before release

**Overview of changes:**
- Adopt the same hacky uuid approach to `{{ pageNav }}` replacement for `<page-nav />`
- `<page-nav />` is replaced into and enforced as a _unique_ uuid
- This uuid is post replaced after layout processing with the page's page nav.

na

**Testing instructions:**
Automated tests only have trivial whitespace diffs, a result of collapsing `{{ pageNav }}`'s whitespace.

Change {{ pageNav }} variable to <page-nav />

Using a nunjucks variable for page navs with the new layout system is
inconsistent with the use of <site-nav />.

Let's standardise these syntaxes.

---

- [ ] No blatantly unrelated changes

### DIFF
--- a/packages/core/src/Layout/Layout.js
+++ b/packages/core/src/Layout/Layout.js
@@ -6,7 +6,6 @@ const { NodeProcessor } = require('../html/NodeProcessor');
 const logger = require('../utils/logger');
 
 const LAYOUT_PAGE_BODY_VARIABLE = 'content';
-const LAYOUT_PAGE_NAV_VARIABLE = 'pageNav';
 
 class Layout {
   constructor(sourceFilePath, config) {
@@ -16,8 +15,7 @@ class Layout {
     this.includedFiles = new Set([this.sourceFilePath]);
     this.layoutProcessed = '';
     this.layoutPageBodyVariable = '';
-    this.layoutPageNavVariable = '';
-    this.hasPageNav = false;
+    this.layoutPageNavUuid = '';
     this.headTop = [];
     this.headBottom = [];
     this.scriptBottom = [];
@@ -33,7 +31,6 @@ class Layout {
   async generate() {
     let triesLeft = 10;
     let numBodyVars;
-    let numPageNavVars;
     let pageSources;
     let nodeProcessor;
 
@@ -44,11 +41,7 @@ class Layout {
       };
       pageSources = new PageSources();
       this.layoutPageBodyVariable = `{{${uuidv4()}-${uuidv4()}}}`;
-      this.layoutPageNavVariable = `{{${uuidv4()}-${uuidv4()}}}`;
-      const layoutVariables = {
-        [LAYOUT_PAGE_BODY_VARIABLE]: this.layoutPageBodyVariable,
-        [LAYOUT_PAGE_NAV_VARIABLE]: this.layoutPageNavVariable,
-      };
+      const layoutVariables = { [LAYOUT_PAGE_BODY_VARIABLE]: this.layoutPageBodyVariable };
 
       const { variableProcessor, pluginManager } = this.config;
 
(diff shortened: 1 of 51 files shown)


packages/core/src/Layout/Layout.js
```diff
@@ -59,26 +52,23 @@ class Layout {
...
       const pageBodyVarRegex = new RegExp(this.layoutPageBodyVariable, 'g');
-      const pageNavVarRegex = new RegExp(this.layoutPageNavVariable, 'g');
       const bodyVarMatch = this.layoutProcessed.match(pageBodyVarRegex);
-      const pageNavVarMatch = this.layoutProcessed.match(pageNavVarRegex);
       numBodyVars = bodyVarMatch ? bodyVarMatch.length : 0;
-      numPageNavVars = pageNavVarMatch ? pageNavVarMatch.length : 0;
 
       triesLeft -= 1;
-    } while (triesLeft > 0 && (numBodyVars > 1 || numPageNavVars > 1));
+    } while (triesLeft > 0 && numBodyVars > 1);
```
wxwxwxwx9, line 64: Just to confirm my understanding of this "hacky" implementation, with the new changes, we won't actually need this "retries" loop anymore, right? 

Considering that we only have 1 variable left -- `LAYOUT_PAGE_BODY_VARIABLE`, so there won't ever be false-positives where `uuidv4` generates the same uuid for 2 different variables (from my inference, the 10 retries is to handle this scenario of false positives). 

This is also to say that any `bodyVarMatch` with count >= 2 will always be duplicates. 

However, we will leave the existing implementation as it is for now, in case there will be new variables needed in the future.

packages/core/src/html/siteAndPageNavProcessor.js
```diff
@@ -13,6 +14,53 @@ const {
...
+    this.uuidTextNode = undefined;
+  }
+
+  getUuid() {
+    return (this.uuidTextNode && this.uuidTextNode.data) || '';
+  }
+
+  renderPageNav(node) {
+    [this.uuidTextNode] = cheerio.parseHTML(uuidv4());
+    cheerio(node).replaceWith(this.uuidTextNode);
```
wxwxwxwx9, line 38: Can I check what does this line of code actually does? :O

I think the purpose is to completely replace the node with the `uuid` right?

wxwxwxwx9 (commented):
Thanks for the implementation @ang-zeyu! Just to make sure I understand the changes in your PR correctly, I'll try to explain what you did in my own words. If there's anything off, do correct me :-)

The main idea is rather straightforward -- we simply "extract" the existing implementation of Nunjucks variable for `PageNav` and put them into a class of its own, and process them as nodes (`<page-nav/>`) instead. 

We could have done the same thing for `PageBody` variable but it's better to leave it as it is, because it seems proper to have the user syntax for content as `{{ content }}`, and not much so for `PageNav`.

Overall I think the implementation should be fine. Just wanted to ask some questions to confirm my understanding.

ang-zeyu:
> We could have done the same thing for PageBody variable but it's better to leave it as it is, because it seems proper to have the user syntax for content as {{ content }}, and not much so for PageNav.

Thanks for taking a look!

Yup, along the lines that `<page-nav>` / `<site-nav>` (both navigation elements) should be standardised syntax wise before release.

packages/core/src/Layout/Layout.js
ang-zeyu: @wxwxwxwx9 Its still possible if the user content contains the generated uuid, but really shouldn't happen in a lifetime.
The loop guards against that nevertheless.

Its a rather hacky approach, but this flow allows processing pages and layouts independently for performance. (also, architecture / code quality wise)

ang-zeyu: > Its a rather hacky approach

something we should fix, but haven't been able to think of a good alternative for a while 🤔

packages/core/src/html/siteAndPageNavProcessor.js
ang-zeyu: yep, the `<page-nav />` wrapper itself is also replaced here with just the uuid text

packages/core/src/Layout/Layout.js
wxwxwxwx9: > > Its a rather hacky approach
> 
> something we should fix, but haven't been able to think of a good alternative for a while 🤔

Hmm yeah this would require some thinking.

wxwxwxwx9 (approved):
LGTM! Just need to touch up the commit message :-)

ang-zeyu:
> LGTM! Just need to touch up the commit message :-)

Nice catch, didn't escape the `<html>` inside 😅

I've fixed it up in the merged commit, thanks for the review!

PR #1570, opened 2021-05-30
Update Developer Guide on SSR

wxwxwxwx9:
**What is the purpose of this pull request?**

- [x] Documentation update
- [ ] Bug fix
- [ ] Feature addition or enhancement
- [ ] Code maintenance
- [ ] Others, please explain:

With the SSR implementation in #1534, we should also update the developer guide to let contributors know what they should be aware of (regarding SSR) when making contributions to MarkBind. 

**Overview of changes:**
Added a new section for SSR under `Design` in the developer guide.

**Anything you'd like to highlight / discuss:**
- Whether the content is enough/appropriate
- Whether the layout of the information is alright
- Suggestions to improve the readability
- Concept accuracy
- The wordings/phrasings

**Testing instructions:**
NIL

**Proposed commit message: (wrap lines at 72 characters)**
Update Developer Guide on SSR.

There are a few important information that contributors should
take note of regarding SSR (especially hydration issues).

Let's update the developer guide to include those information.

---

**Checklist:** :ballot_box_with_check:

- [x] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No blatantly unrelated changes
- [x] Pinged someone for a review!

PR #1573, opened 2021-06-05
Update Developer Guide on SSR based on feedback

wxwxwxwx9:

- [x] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
This is a follow-up to PR #1570 where I update the SSR portion in developer guide based on @ang-zeyu [feedback](https://github.com/MarkBind/markbind/pull/1570#pullrequestreview-676746328).

Whether there's anymore to add.

**Testing instructions:**
NA

Update Developer Guide on SSR based on feedback. 

---


docs/devGuide/design/serverSideRendering.md
```diff
@@ -33,7 +33,15 @@ Here is a short list of questions to check your understanding of Vue:
 - Are there any differences between compiling Vue on client-side versus server-side?
 - What is the difference between compiling and rendering?
 
-### MarkBind's Packages
+<box type="info" header="##### Useful Resources to understand Vue better">
```
ang-zeyu, line 36: shall we standardise to use the seamless boxes? the one below as well

docs/devGuide/design/serverSideRendering.md
```diff
@@ -70,11 +93,15 @@ Supposedly, hydration issues typically occurs due to minor differences between c
 Conceptually, to prevent hydration issue, what we should always strive to achieve is a "universal application". 
 
 It is not difficult to achieve a "universal application" per-se because we merely have to ensure two things:
-1) the state data are the same between client-side and server-side.
+1) the initial states are the same between client-side and server-side.
```
ang-zeyu, line 96:
```suggestion
1) the initial state is the same between client-side and server-side.
```

docs/devGuide/design/serverSideRendering.md
```diff
@@ -70,11 +93,15 @@ Supposedly, hydration issues typically occurs due to minor differences between c
...
 
-However, beyond achieving a "universal application", there are also some more specific rules that we should adhere to, so that we do not run into hydration issue: 
-- Do not violate HTML spec as much as possible (e.g. having block-level elements within `<p>` tag).
+Beyond achieving a "universal application", note that the HTML specifications should also be adhered to. 
+
+Some common mistakes are as such: 
+- Having block-level elements within `<p>` tag
 - Having unknown HTML elements within our Vue application during compilation/rendering (though this can be easily resolved by adding `v-pre` to the unknown element, so that Vue will ignore that element during compilation). 
 
+If you are unsure what elements are allowed within other elements, or what constitutes invalid HTML in general, a good resource to reference would be the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span).
```
ang-zeyu, line 105: `<box type="success" seamless>` might be a good fit here

docs/devGuide/design/serverSideRendering.md
```diff
@@ -53,10 +61,25 @@ The main motivation that we had for introducing SSR into MarkBind is to enhance
...
 
 This is known as "Hydration Issue" and it is one of the main challenges you will face with SSR in MarkBind. 
 
+<box type="warning" header="##### Development vs. Production Build">
+
+There are some significant differences in how Vue **reports and handles hydration issues** between development and production builds.
+
+!!**Development Build**!!
+
+In development build, when Vue reports hydration issue, it will also provide information about the **exact mismatching node(s)** that is/are causing the hydration issue and inform the user that full CSR will be executed. Typically, it is very difficult to resolve hydration issues without the detailed error logs provided by the development build.
```
ang-zeyu, line 74: The box seems a little large -- should we remove this paragraph (also the 'there are some...' paragraph)? It seems explained by the section itself already, but I might have missed something.

ang-zeyu (commented):
thanks for fixing this up!

wxwxwxwx9:
@ang-zeyu thanks for the feedback! Have made the changes accordingly :-)

docs/devGuide/design/serverSideRendering.md
ang-zeyu: @wxwxwxwx9 I think missed this one

wxwxwxwx9: @ang-zeyu Oops, yup I missed that part. Thanks for the spot. Removed it!

ang-zeyu: @wxwxwxwx9

Hmm I actually meant to remove the entire `Development build` paragraph at first actually so this box only highlights what happens in production. Its a minor nit with keeping the alert box small.

and avoiding repeating what happens in development vue builds which was illustrated in your earlier paragraph accurately.
```
During the hydration process, Vue essentially compares your SSR HTML markup against the virtual DOM generated by the render function on the client-side. If any difference is found, meaning that the application that we have on the client-side (the virtual DOM tree) differs from the SSR HTML mark-up that we send to the client, Vue will reject the SSR HTML output, bail Client-side Hydration, and execute full CSR.
```

>Development Build
>In development build, when Vue reports hydration issue, it will also provide information about the exact mismatching node(s) that is/are causing the hydration issue and inform the user that full CSR will be executed.

> In production build, when Vue reports hydration issue, it will only log a generic "Failed to execute 'appendChild' on 'Node': This node type does not support this method" error message and execute full CSR silently. We have even observed cases where the production build does not warn about existing hydration issues (silent failures).
>
> The above reasons are why MarkBind uses the development build by default when you serve it in development mode (-d option).

The comparison may add confusion as -- note the `Failed to execute 'appendChild' on 'Node': This node type does not support this method` message is non standard, it'll vary depending on the error resulting from hydration itself. e.g. Is it a missing node? A node with an attribute that shouldn't be there? An extra node? -- many possible causes. Such errors may also not appear on page start (e.g. surfacing only when you click on a button that does some dom manipulation).

We could simply state that the production build does not perform the hydration check at all. Any such errors resulting from hydration prompts a full CSR render again. And then highlight that its not wise to test hydration errors with production bulids.

wxwxwxwx9: > Hmm I actually meant to remove the entire `Development build` paragraph at first actually so this box only highlights what happens in production. Its a minor nit with keeping the alert box small.

Got it! Misinterpreted it in the beginning 😓 

> The comparison may add confusion as -- note the `Failed to execute 'appendChild' on 'Node': This node type does not support this method` message is non standard, it'll vary depending on the error resulting from hydration itself. e.g. Is it a missing node? A node with an attribute that shouldn't be there? An extra node? -- many possible causes. Such errors may also not appear on page start (e.g. surfacing only when you click on a button that does some dom manipulation).
> 
> We could simply state that the production build does not perform the hydration check at all. Any such errors resulting from hydration prompts a full CSR render again. And then highlight that its not wise to test hydration errors with production bulids.

Thanks for the suggestion! I've made the changes already :-)

PR #1594, opened 2021-06-19
Handle baseUrl in live-preview reload check

### LINKED ISSUE #1583: Live preview shows spinner instead of new content
v3.0.1

When I change the content of a page, the live preview triggers a refresh but the spinner doesn't goes away. The new content appears only after I do a manual refresh.
The same applies when I click on a link in the previewed page. The browser goes to the new page, but get stuck at the spinner. It shows the content only after a manual refresh of the page.

I was using `markbind serve -o` but I don't think the flag matters. I think it depends on the site size though. Didn't notice it in smaller sites but happens in big sites such as the se-book.

ryoarmanda:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #1583 

The problem lies on a mismatch between the HTTP-request routine and the filesystem-watch routine in the `live-server` patch. During HTTP-request, the URLs - which is used to identify WebSocket clients - actually include the `baseUrl`, while during the filesystem-watch, the filepaths of changed files do not. Thus, reloads are unable to be triggered due to lack of `baseUrl`.

The solution is to bring the `baseUrl` information to the patch. Fortunately, we have this information in our `mount` property of the server configuration, which is used to route paths prefixed by `baseUrl` to serve files from the `_site` directory.
 
For convenience and simplicity for MarkBind use, we now identify WebSocket clients by its *normalized* url, that is the url without the `baseUrl`. That way, when MarkBind calls `LiveServer.getActiveUrls()`, it will receive urls without needing further processing of removing the base url.

**Overview of changes:**
- `live-server` patch
  - Added `baseUrl` property to the server object
  - Added `baseUrl` initialization during `mount` rule application
  - Modified socket-establishment routine to normalize the request url and record it as the identifier for the active tab

Review of the general approach is appreciated.

**Testing instructions:**
Test these on the default site (`markbind init`), with defined and empty `baseUrl` property in `site.json`:

1. Visit a page: it should be reloading promptly after page is built
2. Edit the currently viewed page: recently viewed pages list should be properly reflected and changes should be promptly displayed

Handle baseUrl in live-preview reload check

In live-preview, checks are implemented to only reload pages that
correspond to the changed file. However, the comparison used for the
check has not taken into account the base url of the site, which
causes these checks to always fail and thus pages are not automatically
reloaded.

Let's add base url handling to the check, so that sites with specified
can properly benefit from the live-preview feature.

---

### DIFF
--- a/packages/cli/src/lib/live-server/index.js
+++ b/packages/cli/src/lib/live-server/index.js
@@ -38,12 +38,14 @@ const pathToLiveServerDir = path.dirname(require.resolve('live-server'));
 // CHANGED: correctly resolve to the live-server directory
 var INJECTED_CODE = fs.readFileSync(path.join(pathToLiveServerDir, "injected.html"), "utf8");
 
-// CHANGED: added active tabs property
 var LiveServer = {
   server: null,
   watcher: null,
   logLevel: 2,
-  activeTabs: []
+
+  // CHANGED: added properties relevant to MarkBind live-preview
+  activeTabs: [],
+  baseUrl: '',
 };
 
 function escape(html){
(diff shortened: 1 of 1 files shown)


damithc:
Thanks for the quick fix @ryoarmanda 👍

PR #1595, opened 2021-06-20
Fix unresponsive fixed header padding

### LINKED ISSUE #1590: Collapsed page-nav covers content
V2 behavior:
image

V3 behavior:
image

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1590.

The fixed header padding is dynamically adjusted by accessing the `cssRules` property of individual stylesheet. But due to [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) security, the `cssRules` is not accessible and its rules cannot be modified if the stylesheet is imported from a different domain (external stylesheet etc.), causing the function to throw an error. This is resolved by wrapping the function in a try/catch block so that stylesheets that are imported from another domain will be skipped.

`NIL`

**Testing instructions:**
1. Import stylesheet from another domain. Eg. `<link rel="stylesheet" href="https://somedomain.com/main.css">`
2. Adjust the window size till page/site nav appears
3. Fixed header padding should be dynamically adjusted

Fix unresponsive fixed header padding

When external stylesheets from different domains are used,
the cssRules of these stylesheets cannot be accessed or
modified due to CORS policy. This causes the method that
adjusts the fixed header padding to throw an error and
terminate prematurely.

Let's wrap this method in a try/catch block so that external
stylesheets from other domains will be skipped if cssRules
is not accessible.

---

### DIFF
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -61,27 +61,31 @@ function detectAndApplyFixedHeaderStyles() {
     const newHeaderHeight = headerSelector.height();
     const sheets = document.styleSheets;
     for (let i = 0; i < sheets.length; i += 1) {
-      const rules = sheets[i].cssRules;
-      // eslint-disable-next-line lodash/prefer-get
-      if (rules && rules[0] && rules[0].selectorText) {
-        switch (rules[0].selectorText) {
-        case '.fixed-header-padding':
-          sheets[i].deleteRule(0);
-          sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important }`);
-          break;
-        case 'span.anchor':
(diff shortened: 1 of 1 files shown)


ang-zeyu (approved):
wonder if we should switch over to bootstrap cdn as a precaution against cors issues 🤣

jonahtanjz:
> wonder if we should switch over to bootstrap cdn as a precaution against cors issues

@ang-zeyu That actually sounds good. Perhaps we can just switch to using cdn on our documentation site while keeping the rest so we can still test for cors issue 🤔

PR #1603, opened 2021-06-27
Replace panel ::before element with scroll-margin-top

### LINKED ISSUE #1601: User Guide: Unable to open details panel on mobile
Safari, iOS.

IMG_2596</img>

On the Nav Bar section at https://markbind.org/userGuide/readerFacingFeatures.html, I am unable to expand the details panel (circled in red). I am able to open other details panels on the same page without any issues.

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1601.

By using `::before` to create spacing before a panel, this blocks certain elements before the panel which will hinder interactive elements (buttons not clickable etc.). The spacing is needed so that panels do not get blocked by the `fixed-header` when using anchor navigation.

Replaced `::before` with `scroll-margin-top` which similarly creates spacing before the panels when using anchor navigation and also do not block other elements.

If there is a better workaround.

**Testing instructions:**
1. Create panel with header
```.html
<panel header="# Test">
</panel>
```
2. Click on panel link in `page-nav`
3. Panel should appear below the `fixed-header`
4. Can refer to #1601 issue, visit nav bar section in UG, details panel should be clickable now

Fix blocking ::before pseudo element in panel

Using ::before in panels may block interactive elements before
the panels. The spacing created using ::before is needed so that
panels do not get blocked by the fixed header when using anchor
navigation.

Let's replace ::before with scroll-margin-top which similarly creates
spacing before the panels when using anchor navigation and also do
not block other elements.

---

### DIFF
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -49,12 +49,7 @@ function detectAndApplyFixedHeaderStyles() {
     top: calc(-${headerHeight}px - ${bufferHeight}rem)
     }`,
   );
-  insertCss(`.card-container::before {
-        display: block;
-        content: '';
-        margin-top: calc(-${headerHeight}px - ${bufferHeight}rem);
-        height: calc(${headerHeight}px + ${bufferHeight}rem);
-      }`);
+  insertCss(`.card-container { scroll-margin-top: calc(${headerHeight}px + ${bufferHeight}rem); }`);
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
 
   const adjustHeaderClasses = () => {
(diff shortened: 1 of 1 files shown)


ang-zeyu (approved):

ang-zeyu:
Thanks @jonahtanjz! Didn't know this property existed (was going to suggest `width: 1px` or avoid assigning the ids to the `<div>` version of panels).

We can probably explore this as an alternative to the dummy `<span>`s inserted for headings with `ids` as well (or even universally applying this to everything under a `fixed-header-padding` container with an `id`).

ang-zeyu:
edit: might have jumped the gun here a little after getting excited with this property 😅 --

wonder if we should revert this for now and keep in view, seems like support for it on recent ios releases isn't too good yet (
https://caniuse.com/?search=scroll-margin-top) Scroll to fragment which is what we're depending on here especially is also bugged apparently, even with its in house `scroll-snap-margin-top` property.

jonahtanjz:
> wonder if we should revert this for now and keep in view, seems like support for it on recent ios releases isn't too good yet (
https://caniuse.com/?search=scroll-margin-top) Scroll to fragment which is what we're depending on here especially is also bugged apparently, even with its in house scroll-snap-margin-top property.

Hmm alright didn't know it was bugged on ios. Will raise another PR to temporary fix this using one of the suggestions. Yup we can probably wait till it is fully supported and replace the existing approach to `fixed-header-padding`.

PR #1609, opened 2021-06-28
Implement span.anchor in panel to replace id

### LINKED ISSUE #1601: User Guide: Unable to open details panel on mobile

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [x] Others, please explain:

**Overview of changes:**
Follow-up to https://github.com/MarkBind/markbind/pull/1603#issuecomment-869170969. The support of `scroll-margin-top` on ios is still limited thus require a different approach to handling anchor navigation to panels.

Replace existing panel id with span.anchor (similar to how normal headers navigation works). This prevents the panels from getting blocked by the `fixed-header` bar as well as resolves #1601.

If there is a better workaround.

Apparantly changing the width of `::before` does not fix #1601 as the entire panel block occupies the vertical space if ::before as well.
image

The container which holds `::before` still has the width of the panel
image

**Testing instructions:**
1. Create panel with header
```.html
<panel header="# Test">
</panel>
```
2. Click on panel link in `page-nav`
3. Panel should appear below the `fixed-header`
4. Can refer to #1601 issue, visit nav bar section in UG, details panel should be clickable now

Implement span.anchor in panel to replace id

The support of scroll-margin-top on ios is still limited hence
would require a different approach to handling anchor navigation
for panels while keeping elements before panels from getting
blocked.

Let's introduce span.anchor into panel, similar to how the headers
navigation works, to support anchor navigation as well as to prevent
previous elements from getting blocked.

---


packages/vue-components/src/panels/MinimalPanel.vue
```diff
@@ -93,6 +103,7 @@ import retriever from '../Retriever.vue';
 import panelBase from './PanelBase';
 
 export default {
+  inheritAttrs: false,
```
ang-zeyu, line 106: hmm what's this for?

jonahtanjz: [Vue passes non-prop attributes such as `id` to the root node of a component](https://v3.vuejs.org/guide/component-attrs.html#non-prop-attributes). In this case, we do not want the main panel container, `div` or `span`, to have the `id` as this interferes with anchor navigation.
The `inheriAttrs: false` is needed to prevent the root node of `minimalPanel` and `nestedPanel` from inheriting the `id` attribute. 

https://v3.vuejs.org/guide/component-attrs.html#disabling-attribute-inheritance

ang-zeyu: Thanks for clarifying! Hmm how about changing the property name under `assignPanelId` (then using it as a prop)?

So there's no risk of side effects from disabling attribute inheritance entirely.

jonahtanjz: Yup that would work as well. Will update it.

jonahtanjz:
@ang-zeyu Updated the PR to pass in `panelId` as prop.

ang-zeyu:
Lgtm 👍, just this portion that would have to be edited also since we aren't using `::before` anymore.


 (handles the auto scroll back to top)
```
// To enable behaviour of auto window scrolling during panel collapse
if (this.$el.getBoundingClientRect().top < 0) {
  jQuery('html').animate({
    scrollTop: window.scrollY + this.$el.getBoundingClientRect().top - 3,
  }, 500, 'swing');
}
```

jonahtanjz:
@ang-zeyu Thanks for the catch. Have updated the PR :)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -141,8 +141,9 @@ export default {
         requestAnimationFrame(() => {
           // To enable behaviour of auto window scrolling during panel collapse
           if (this.$el.getBoundingClientRect().top < 0) {
+            const headerHeight = jQuery('header[fixed]').height() || 0;
```
ang-zeyu, line 144: a really minor nit for keeping everything under the panel implementation -- would this work?
```
const el = $refs.anchor || this.$el;
el .getBoundingClientRect......
```

jonahtanjz: Hmm added `ref="anchor"` to the `span`s and doesn't seem to be working... The panel header gets hidden under the navbar still.

ang-zeyu: try adding `display: inline-block` to the scoped anchor class here

jonahtanjz: Slightly better but half of it is still under the navbar... When lower navbar is present, it gets hidden again 😅  

image

ang-zeyu: Looks like the sibling after span.anchor is interfering 🤣...  let's leave it for now and revisit later

ang-zeyu (approved):
Lgtm 👍

PR #1612, opened 2021-06-28
Support opt-out of printing certain tabs

### LINKED ISSUE #1605: Make 'print all tabs' optional
In #1498, we made _all_ tabs appear in the print view. In retrospect (my bad), this did not consider the case where we actually _don't_ want some tabs to be printed (e.g., if they contained optional content). 

I think we need to provide a way to opt-in or opt-out of showing a specific tab in the print view.

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1605.

Add `no-print` prop to `<tabs>`, `<tab-group>` and `<tab>` which will exclude the entire tab, tab group or single tab respectively when printing.

`NIL`

**Testing instructions:**
1. Create tab with `no-print` attribute
```.html
<tabs>
  <tab header="First tab">
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla ullamcorper ultrices lobortis.
  </tab>
  <tab header="Second tab">
    Some content of the second tab...
  </tab>
  <tab header="Third tab" no-print>
    Additional content from third tab...
  </tab>
  <tab-group header="Example Tab Group" no-print>
    <tab header="Tab Group 1">
      Some content in a tab group 1...
    </tab>
    <tab header="Tab Group 2">
      Some content in a tab group 2...
    </tab>
  </tab-group>
</tabs>
```
2. Print the page
3. `Third tab` and `Example Tab Group` should be excluded from the print preview

Support opt-out of printing certain tabs

Users may want to omit some tabs when printing.

Let's remove usages of !important in tabs so that
users are able to specify which tab components
to omit when printing using bootstrap's display
property.

---

### DIFF
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -1,11 +1,10 @@
 <template>
   <div
-    v-show="show"
     role="tabpanel"
-    class="tab-pane active printable-tab-pane"
-    :class="[{hide:!show}, printableClass]"
+    class="tab-pane active printable-tab-pane tab-print-block"
+    :class="{hide:!show}"
   >
-    <div class="nav-tabs printable-tab-header d-none d-print-flex">
+    <div class="nav-tabs printable-tab-header tab-print-flex">
       <div ref="header" class="nav-link active">
         <slot name="header"></slot>
       </div>
@@ -28,10 +27,6 @@ export default {
       type: Boolean,
       default: false,
     },
-    noPrint: {
-      type: Boolean,
-      default: false,
-    },
   },
   computed: {
     headerRendered() {
@@ -48,9 +43,6 @@ export default {
     },
     disabledBool() {
       return toBoolean(this.disabled);
-    },
-    printableClass() {
-      return this.noPrint ? 'd-print-none' : 'd-print-block';
     },
   },
   created() {
@@ -91,6 +83,14 @@ export default {
         margin: 0;
     }
 
+    .tab-pane.hide {
+        display: none;
+    }
+
+    .printable-tab-header {
+        display: none;
+    }
+
     @media print {
         .printable-tab-header {
             margin-bottom: 15px;
@@ -103,6 +103,14 @@ export default {
 
         .printable-tab-pane {
             padding: 10px;
+        }
+
+        .tab-pane.printable-tab-pane.tab-print-block {
+            display: block;
+        }
+
+        .printable-tab-header.tab-print-flex {
+            display: flex;
         }
     }
 </style>
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -2,8 +2,8 @@
   <div
     v-show="show"
     role="tabpanel"
-    class="tab-pane active printable-tab-pane d-print-block"
-    :class="{hide:!show}"
+    class="tab-pane active printable-tab-pane"
+    :class="[{hide:!show}, printableClass]"
   >
     <div class="nav-tabs printable-tab-header d-none d-print-flex">
       <div ref="header" class="nav-link active">
@@ -28,6 +28,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    noPrint: {
+      type: Boolean,
+      default: false,
+    },
   },
   computed: {
     headerRendered() {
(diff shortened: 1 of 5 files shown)


raysonkoh (approved):
LGTM! Have tested with some small examples and verified that it is working as intended.

Thanks @jonahtanjz!

damithc:
@MarkBind/active-devs let's do a release after this PR is merged.

jonahtanjz:
There's some changes to the implementation as discussed here https://github.com/MarkBind/markbind/issues/1605#issuecomment-873098643 @raysonkoh, this PR is not yet updated.

Prof @damithc, I will most likely make the changes and update the PR by tomorrow night!

damithc:
> Prof @damithc, I will most likely make the changes and update the PR by tomorrow night

Got it. Thanks @jonahtanjz

jonahtanjz:
@ang-zeyu Noticed that we can't really remove the current usages of `!important` as discussed https://github.com/MarkBind/markbind/issues/1605#issuecomment-873629602. Should we go with the implementation in this PR?

jonahtanjz:
@ang-zeyu Thanks for the suggestion 👍 

Have changed `v-show` and `d-print-block/flex` to our own custom classes. Needed to add them individually to `tab-group` and `tab` so that we can increase the specificity from there. `no-print` is added to the global css file with `!important` so that it can be reused by other components as well once we have changed out all the other usages of `!important`.

packages/vue-components/src/TabGroup.vue
```diff
@@ -84,5 +88,9 @@ export default {
             margin-bottom: 10px;
             text-decoration: underline;
         }
+
+        .printable-tab-group-header.tab-group-print-block {
```
ang-zeyu, line 92: > @ang-zeyu Thanks for the suggestion 👍
> 
> Have changed `v-show` and `d-print-block/flex` to our own custom classes. Needed to add them individually to `tab-group` and `tab` so that we can increase the specificity from there. `no-print` is added to the global css file with `!important` so that it can be reused by other components as well once we have changed out all the other usages of `!important`.

hmm... not sure I follow, what's the addition `tab-group-print-block` for? (the variants in `tab.vue` as well)

packages/vue-components/src/Tab.vue
```diff
@@ -84,6 +83,14 @@ export default {
         margin: 0;
     }
 
+    .tab-pane.hide {
```
ang-zeyu, line 86: would `tab-pane-hide` work instead? To avoid adding empty classes below to increase the specificity

jonahtanjz: Seems like there is a css rule in bootstrap, `.tab-content > .active` that is more specific than `.tab-pane-hide`. We can also override that by using `.tab-content > .tab-pane-hide` in our scoped css but that would mean having to add `.tab-content > .printable-tab-pane` in the print rule as well. Should we go with this as this doesn't seem to resolve the empty classes issue?

packages/vue-components/src/TabGroup.vue
jonahtanjz: Ahh seems like css that appear the latest will take effect anyway if same level of specificity. Initially added the `tab-group-print-block` so that it can have higher specificity over the `dispaly: none` rules to ensure that it is always overriding it.

packages/vue-components/src/Tab.vue
ang-zeyu: How about using `<style scoped>`? It adds `data-******` attributes under the hood and may be sufficient to override `.tab-content > .active`'s specificity.

packages/vue-components/src/TabGroup.vue
ang-zeyu: > Ahh seems like css that appear the latest will take effect anyway if same level of specificity.

yes, one of the main reasons its named *cascading* style sheets. Let's remove it then!

packages/vue-components/src/Tab.vue
jonahtanjz: Thanks this works 👍 I've updated the PR accordingly.

ang-zeyu (approved):
Lgtm 👍. Thanks for accomodating the various change requests here!

PR #1630, opened 2021-07-06
Fix warning for boolean props

### LINKED ISSUE #1628: Vue warning for boolean variables
Error message in the build console:
>---> <NestedPanel> at vue-components/src/panels/NestedPanel.vue
       <Panel> at vue-components/src/Panel.vue
         <Root>
[Vue warn]: Invalid prop: type check failed for prop "expandable". Expected Boolean, got String with value "true".
found in
---> <NestedPanel> at vue-components/src/panels/NestedPanel.vue
       <Panel> at vue-components/src/Panel.vue
         <Root>

To reproduce, 

1. clone https://github.com/nus-cs2103-AY2021S2/website
1. clone the submodule, as explained in the README
1. checkout the branch `cs2103`
1. run `markbind build`

I guess similar to #1610 

Remember to test against the cs2103 website in case there are similar errors but not exactly same as the one given above.

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1628.

Add `String` type to Boolean props with default value of `true`. This is because HTML attribute value is passed as `String`.
In this case, `false` will be rendered as `String`:
```.html
<panel header="..." expandable="false">
  ...
</panel>
```
In this case, `minimized` will be passed as `Boolean`:
```.html
<panel header="..." minimized>
  ...
</panel>
```

Should we add `String` type to all Boolean props that are from HTML attributes? Currently I only see the need for props with default value of `true` as the only way to set it to `false` as a HTML attribute is to pass a `"false"` value. 
However, if a user input this,
```.html
<panel header="..." minimized="true">
  ...
</panel>
```
minimized will be rendered as `String`. Not sure if the above example is the correct usage of HTML in general. Should we leave this to the user such that if they want an attribute to be true, just include the attribute there without any value? The above example of passing a `true` value will trigger warning unless we include `String` type for all `Boolean` props coming from HTML.

**Testing instructions:**
Follow steps in #1628.

Fix warning for Boolean props

---

### DIFF
--- a/packages/vue-components/src/TabGroup.vue
+++ b/packages/vue-components/src/TabGroup.vue
@@ -13,7 +13,7 @@ import { coerce } from './utils/utils';
 export default {
   props: {
     disabled: {
-      type: Boolean,
+      type: [Boolean, String],
       default: false,
     },
     header: {
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -63,7 +63,7 @@ export default {
   },
   props: {
     disabled: {
-      type: Boolean,
+      type: [Boolean, String],
       default: false,
     },
     type: {
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -62,7 +62,7 @@ export default {
       default: '',
     },
     fixed: {
-      type: Boolean,
+      type: [Boolean, String],
       default: false,
     },
     defaultHighlightOn: {
--- a/packages/vue-components/src/Retriever.vue
+++ b/packages/vue-components/src/Retriever.vue
@@ -20,7 +20,7 @@ export default {
       default: null,
     },
     delay: {
-      type: Boolean,
+      type: [Boolean, String],
       default: false,
     },
     hasFetched: {
--- a/packages/vue-components/src/Submenu.vue
+++ b/packages/vue-components/src/Submenu.vue
@@ -33,7 +33,7 @@ export default {
       default: '',
     },
     disabled: {
-      type: Boolean,
+      type: [Boolean, String],
       default: false,
     },
   },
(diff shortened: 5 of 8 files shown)


ang-zeyu:
> Should we leave this to the user such that if they want an attribute to be true, just include the attribute there without any value?

Agree with this. Going forward we should (have been for while already though) stick with this for new component props. It should be simpler for both the user and implementation wise.

For the previous existing attributes using this behaviour, it is a breaking change, so we can slate that for v4.
In the meantime I think the docs need some updating on `true / false` versus _"just put the property name there"_ though =X

> Should we add String type to all Boolean props that are from HTML attributes? 

It should rightly be added for all props using the `toBoolean` function (a relic from vue v1->v2 migration), but I think its ok to fix-this-as-we-go to minimise possible side effects, the warning from vue is not critical in comparison. But if you're up for investigating each usage let's do that 😀

jonahtanjz:
Thanks for the review @ang-zeyu

> In the meantime I think the docs need some updating on true / false versus "just put the property name there" though

Yup agree. Will open another issue and PR for this :)

>It should rightly be added for all props using the toBoolean function (a relic from vue v1->v2 migration), but I think its ok to fix-this-as-we-go to minimise possible side effects, the warning from vue is not critical in comparison. But if you're up for investigating each usage let's do that

Ok got it! Let's go with this 👍

jonahtanjz:
@ang-zeyu I have updated the PR to include `String` to all relevant props using `toBoolean`.

packages/vue-components/src/TabGroup.vue
```diff
@@ -13,7 +13,7 @@ import { coerce } from './utils/utils';
 export default {
   props: {
     disabled: {
-      type: Boolean,
+      type: [Boolean, String],
```
ang-zeyu, line 16: 👍

ang-zeyu (commented):
Lgtm!

ang-zeyu (approved):

Issue #771, opened 2019-03-14
Split up test_site into multiple test sites


Xenonym:
**Tell us about your environment**

* **MarkBind Version:** 1.20.0

**Rationale:** Our current `test_site` contains 10 pages, and is getting more unwieldy as we add new features to MarkBind that require functional testing. Since we now have the ability to use multiple test sites (#588), we should split up these pages into multiple test sites, as well as consider if the main test page can be split up as well. This will reduce the size of the diffs when updating functional tests and make it easier to verify that the updates are correct.

Related: https://github.com/MarkBind/markbind/issues/761#issuecomment-472721916

---
**Alternate proposal:** Should we consider migrating to [Jest snapshot testing](https://jestjs.io/docs/en/snapshot-testing)? It does seem like we have accidentally rebuilt this functionality with our test sites 😅

yamgent:
> **Alternate proposal:** Should we consider migrating to [Jest snapshot testing](https://jestjs.io/docs/en/snapshot-testing)? It does seem like we have accidentally rebuilt this functionality with our test sites 😅

Sure, using frameworks is definitely better than relying on our own rolled-out stuff. Maybe modify the OP to discuss about Jest snapshot instead? Or open a new issue.

tlylt:
With the number of test sites that we have now, it might be good to migrate the invocation of these test cases to jest to
- standardize with the rest of our unit cases
- improve the time taken by exploring running the tests in parallel

Needs to look into the existing setup in https://github.com/MarkBind/markbind/blob/da0cd19097dd6454f71122ba0aa3a5a2a8c685c2/packages/cli/test/functional/test.js

Warning: I recall having some weird issues when I toyed with this idea some time ago.

PR #1645, opened 2021-08-08
Fix running functional tests in packages/cli

### LINKED ISSUE #1641: Functional tests failed to run properly when executing `npm run test` in `packages/cli`
Windows 10

 **MarkBind Version:**
3.0.6

According to the developer guide > test section:
>If you only want to run tests for one of the packages (packages/*), simply switch into the appropriate directory and use npm run test as well!

So I ran `npm run test` in the root folder, everything seems fine and tests are passing. However, when I ran `npm run test` in `packages/cli`, I got stuck.

Expect test cases to run successfully. In particular, expect the functional tests in `cli/test/functional` to run successfully. This is correctly shown when running in the root directory:
image

However, when running in `packages/cli`, the unit tests are ok, but the functional ones are not ok:
image
image

it gets stuck at running test_site tests and does not continue to the next test. i.e. 'test_site_algolia_plugin' and 'test_site_special_tags'.

Not sure if this is an issue with outdated documentation or a bug as I did not make any changes to the files but running the test cases in root directory or packages/cli have different outcome.

tlylt:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #1641 
**Overview of changes:**
#### ✍️ Action taken  
Setting the environment variable `TEST_MODE=true` before running a comparison between expected and actual HTML generated.

#### 📘 Background & Elaboration
Initially running `npm run test` within the `cli` package results in inconsistent behavior as compared to running `npm run test` in the root directory. After some investigation and hints provided by @wxwxwxwx9 [here](https://github.com/MarkBind/markbind/issues/1641#issuecomment-894279137), I discovered that indeed tests under `packages/cli` behaves differently because the test_site HTML generated is different. To be exact, the HTML generated is minified. This caused a problem with the `jsdiff.diffChars` function used in functional tests, making it both slow running and reaching a conclusion that the two files being compared are different.

By setting the environment variable, the following function in `packages\core\src\Page\index.js` will output beautified HTML as required, thus passing the test cases.
```javascript
  async outputPageHtml(content, pageNav) {
    const renderedTemplate = this.pageConfig.template.render(
      this.prepareTemplateData(content, !!pageNav)); // page.njk

    const outputTemplateHTML = process.env.TEST_MODE
      ? htmlBeautify(renderedTemplate, this.pageConfig.pluginManager.htmlBeautifyOptions)
      : renderedTemplate;

    await fs.outputFile(this.pageConfig.resultPath, outputTemplateHTML);
  }
```

This issue is indeed a regression from a past PR. Though the fix is simple, it served as a great first issue for me as I had to spend some hours digging around the packages, which ended up helping me understand the repo better. So thanks @wxwxwxwx9!
To continue the discussion in #1641, I have the following points that perhaps would like to get others to comment on:
- I played around with [`diff-match-patch`](https://github.com/google/diff-match-patch) in place of `jsdiff` and it seems to be more performant to do the text comparison. There are a few issues around performance in jsdiff's repo about it vs diff-match-patch, kpdecker/jsdiff#225 and kpdecker/jsdiff#163. Perhaps we could explore this alternative in the future.
- On the same note, perhaps we could explore switching to jest's snapshot testing instead.
- Similarly, I am wondering if we should trigger the functional tests with jest as well, instead of calling the `node test.js` command. With jest and some refactoring, we could have better test feedback for functional tests. This could also potentially be a good starting point where we organize the existing functional tests code better.

**Testing instructions:**
Run `npm run test` in the root directory and in `packages/cli`, both should run successfully. 

Fix running functional tests in packages/cli

Functional tests are failing when invoked under packages/cli.

Let's set `TEST_MODE = true`  within `test.js` and `updatetest.js` to
output beautified HTML for comparison and make the tests pass again.

Use of `cross-env TEST_MODE=true` in `package.json` has been removed
in favor of inline configuration.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "lint": "eslint . --ext .js,.vue && npm run csslint",
     "lintfix": "eslint . --ext .js,.vue --fix && npm run csslintfix",
     "setup": "npm ci && lerna clean --yes && lerna bootstrap --hoist",
-    "test": "npm run lint && cross-env TEST_MODE=true lerna run test --stream",
-    "updatetest": "cross-env TEST_MODE=true lerna run updatetest --stream"
+    "test": "npm run lint && lerna run test --stream",
+    "updatetest": "lerna run updatetest --stream"
   },
   "devDependencies": {
     "eslint": "^7.25.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/MarkBind/markbind.git"
   },
   "scripts": {
-    "test": "jest --colors && cd test/functional && node test.js",
+    "test": "jest --colors && cd test/functional && cross-env TEST_MODE=true node test.js",
     "updatetest": "cd test/functional && node updatetest.js"
   },
   "dependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/MarkBind/markbind.git"
   },
   "scripts": {
-    "test": "jest --colors && cd test/functional && cross-env TEST_MODE=true node test.js",
+    "test": "jest --colors && cd test/functional && node test.js",
     "updatetest": "cd test/functional && node updatetest.js"
   },
   "dependencies": {
--- a/packages/cli/test/functional/test.js
+++ b/packages/cli/test/functional/test.js
@@ -18,6 +18,7 @@ function printFailedMessage(err, siteName) {
   console.log(`Test result: ${siteName} FAILED`);
 }
 
+process.env.TEST_MODE = true;
 process.env.FORCE_COLOR = '3';
 
 const execOptions = {
--- a/packages/cli/test/functional/updatetest.js
+++ b/packages/cli/test/functional/updatetest.js
@@ -17,6 +17,7 @@ function printFailedMessage(err, siteName) {
   console.log(`Failed to update: ${siteName}`);
 }
 
+process.env.TEST_MODE = true;
 process.env.FORCE_COLOR = '3';
 
 const execOptions = {


wxwxwxwx9:
@tlylt Thanks for working on this! Yup, it's because of our SSR that we need to set `cross-env TEST_MODE=true`. We currently produce different HTML output for production and test environment. I think you would have already read the rationale for that in the codebase comments as well :-) 

Just a small issue, we would need to add `cross-env TEST_MODE=true` for the `updatetest` script command as well. Otherwise, I think everything LGTM. 

> On the same note, perhaps we could explore switching to jest's snapshot testing instead.
Similarly, I am wondering if we should trigger the functional tests with jest as well, instead of calling the node test.js command. With jest and some refactoring, we could have better test feedback for functional tests. This could also potentially be a good starting point where we organize the existing functional tests code better. 

Hmm I agree that we can look into using Jest for our functional tests as well. Will make the testing in our codebase more standardized. Just require some rewrite of the test. I personally don't mind going in this direction. 

I've checked the past PRs to check if there was a reason for us to not use Jest for our functional tests in CLI but it seems like there isn't any. However, just to be sure, @ang-zeyu can I check if you know anything about this? 

If it's not feasible to switch to Jest for some reason (be it implementation-wise or other reasons), I think it's fine to keep the functional tests as they are currently as well. But we should look into `diff-match-path` to improve the performance in this case. 

@MarkBind/active-devs Feel free to chime in as well :-)

tlylt:
Hi @wxwxwxwx9 , thanks for the reminder, I should have tested `npm run updatetest` as well 😂 

By the way, I am wondering if we should set `TEST_MODE = true` in `package.json` or should we do it within `test.js` and `updatetest.js` instead.
So for example in `test.js`, adding `process.env.TEST_MODE = true;` alongside setting of `FORCE_COLOR`:
image

This might potentially make it more resistant to missing it out in future updates.

I will update the PR after your kind advice on this. Thanks!

wxwxwxwx9:
@tlylt Yup, I'm good with that as well! You can go ahead with it :-)

ang-zeyu:
> I've checked the past PRs to check if there was a reason for us to not use Jest for our functional tests in CLI but it seems like there isn't any. However, just to be sure, @ang-zeyu can I check if you know anything about this?

nope, its an open issue https://github.com/MarkBind/markbind/issues/771, definitely worth exploring.
we also used to have 2 sets of test scripts (`.sh` and `.bat`) doing the same thing as the current js script https://github.com/MarkBind/markbind/issues/829, likely back then (5 + years ago) the js test ecosystem wasn't as well developed and well-known as it is now and javascript / jest wasn't really considered as such.

wxwxwxwx9 (approved):
LGTM!

wxwxwxwx9:
> > I've checked the past PRs to check if there was a reason for us to not use Jest for our functional tests in CLI but it seems like there isn't any. However, just to be sure, @ang-zeyu can I check if you know anything about this?
> 
> nope, its an open issue #771, definitely worth exploring.
> we also used to have 2 sets of test scripts (`.sh` and `.bat`) doing the same thing as the current js script #829, likely back then (5 + years ago) the js test ecosystem wasn't as well developed and well-known as it is now and javascript / jest wasn't really considered as such.

Ahh I see, I should have checked the issues as well. Got it! 

@tlylt If you're interested in taking this up :-)

tlylt:
> > > I've checked the past PRs to check if there was a reason for us to not use Jest for our functional tests in CLI but it seems like there isn't any. However, just to be sure, @ang-zeyu can I check if you know anything about this?
> > 
> > 
> > nope, its an open issue #771, definitely worth exploring.
> > we also used to have 2 sets of test scripts (`.sh` and `.bat`) doing the same thing as the current js script #829, likely back then (5 + years ago) the js test ecosystem wasn't as well developed and well-known as it is now and javascript / jest wasn't really considered as such.
> 
> @ang-zeyu Ahh I see, I should have checked the issues as well. Got it! Thanks for the explanation 👍
> 
> @tlylt If you're interested in taking this up :-)

Okay will put it in my todo list :)
I saw that issue on jest snapshot but that one led to the implementation of vue component snapshot testing... so I wasn't sure if it's still relevant to test_site anymore 😂

PR #1647, opened 2021-08-09
Box component theming update

### LINKED ISSUE #1031: <box>: support normal bootstrap styles
Current: box components has their own set of types (e.g., definition, right, success, ...) some of which overlap with bootstrap styles (e.g., primary, secondary, success, ...). Should we eliminate the overlap (by renaming) but support both categories? e.g.,
* custom types: use the darker shades (similar to current), come with preset icons, some existing types need to be renamed `success` -> `?`
* standard types (primary, info, secondary, etc.): use the regular/bright bootstrap colors used for panels, no icons by default

### LINKED ISSUE #1581: Box components: support applying built-in color schemes
v3.0.1

Box components follow a different color scheme based on the type of box. Example:
image

In some cases, the default color scheme might not go well with the overall style of the page. In the example above, blue/black/grey would fit better with the page.

Perhaps we can let the user override? e.g., `<box type="tip dark" ...`

Yes, we do support fine-grained color changes for boxes but it's easier for the user if s/he can take advantage of predefined styles such as `primary, success, warning etc.` more easily.

### LINKED ISSUE #1642: Indented boxes get cut off
When indented, the right edge of boxes seem to get cut-off

image

Example https://nus-cs2103-ay2122s1.github.io/website/admin/ip-w4.html

ong6:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
This PR directly addresses adding a built-in colour scheme to the Box Components as @damithc suggested in #1581. With this, users can now choose to add colours along with an optional bootstrap theme.

With this PR, I have added the following styles to be used in the `theme` prop: 
['primary', 'secondary', 'success', 'danger', 'warning', 'info', 'light', 'dark', 'muted', 'white']

A few examples of the new styles added are as follows:

```
<box type="warning"  theme="success" seamless>
    test
</box>

<box type="info" theme="warning">
    test
</box>

<box type="definition" theme="dark" light>
    test
</box>
```

image

Some refactoring was done for the `switch, case` functions in TipBox to reduce code duplication and the box cut-off issue in #1642 is also fixed.

The `theme` is completely optional and will work if nothing is passed in. `theme` will also override any styles the original type may have had. It might be more user friendly to instead name the theme by colours instead of by the bootstrap themes.
`primary`, `secondary` etc... 
to
`blue`, `red` etc.. instead

**Testing instructions:**
No test cases have been made for this component as of now.

Added colour customization to the box component

---

- [X] No blatantly unrelated changes

Fixes #1581
Fixes #1031 
Fixes #1642


ang-zeyu:
> Oof, did not read that you had settled on the theme part. Shall I redo this to work with the theme instead?

hmm after some thought I think that might still be better, but only because we should also shift the `red / black / etc.` names to the bootstrap ones --- so the user dosen't have to remember two sets of color names, and because `red / black / etc.` are predefined css color names (there's an existing `background-color` prop for that).

If we use bootstrap names, then the separate `theme` key is also necessary (the "conflict" mentioned [here](https://github.com/MarkBind/markbind/issues/1581#issuecomment-864533209))

ong6:
Ok, then I'll work on adding a theme prop then! Thanks for the input @ang-zeyu

packages/vue-components/src/utils/utils.js
```diff
@@ -159,3 +159,41 @@ export function VueFixer (vue) {
...
+  case 'important':
+  case 'wrong':
+    primaryStyle = 'danger';
+    break;
+  default:
+    primaryStyle = 'default';
+    break;
+  }
+
+  return {baseStyle: primaryStyle, colorStyle: colorStyle, iconStyle: defaultStyle};
```
jonahtanjz, line 198: How about returning either `primaryStyle` or `colorStyle` instead of both since only one of them will be used. This can reduce the use of
```.js
if (colorStyle) {
  return `...-${colorStyle}`;
}
return `...-${baseStyle}`;
```
in some of the methods.

packages/vue-components/src/TipBox.vue
```diff
@@ -251,61 +257,29 @@ export default {
...
-        return 'alert-primary';
-      case 'success':
-      case 'tip':
-        return 'alert-success';
-      case 'important':
-      case 'wrong':
-        return 'alert-danger';
-      default:
-        return 'alert-default';
+      const { baseStyle, colorStyle } = classifyBootstrapStyle(this.type, this.theme);
```
jonahtanjz, line 260: Would placing this in a `computed` be better since there are a few usages in other methods? This can avoid repeated calls to `classifyBootstrapStyle`.

packages/vue-components/src/TipBox.vue
```diff
@@ -251,61 +257,29 @@ export default {
...
+      if (colorStyle) {
+        return `border-${colorStyle}`;
       }
+      return `border-${baseStyle}`;
     },
     getFontAwesomeIconStyle() {
-      switch (this.type) {
+      const { iconStyle } = classifyBootstrapStyle(this.type, this.theme);
+      switch (iconStyle) {
```
jonahtanjz, line 282: What do you think about extracting this to `classifyBootstrapStyle` as well?

packages/vue-components/src/TipBox.vue
```diff
@@ -330,7 +304,7 @@ export default {
 
 <style scoped>
     .box-container {
-        width: 100%;
+        /* width: 100%; */
```
jonahtanjz, line 307: Seems like this fixes the issue. Can remove it.

jonahtanjz (commented):
Good job and thanks for working on this @ong6 👍

Generally looks good and just a couple of suggestions.

packages/vue-components/src/utils/utils.js
ong6: Ok this does make more sense, will make this change

packages/vue-components/src/TipBox.vue
ong6: Ok this does make more sense, will make this change

packages/vue-components/src/TipBox.vue
ong6: OK will implement this then!

packages/vue-components/src/TipBox.vue
ong6: Ok this sounds like a good improvement as well, still quite new to Vue so this tip was really helpful!

packages/vue-components/src/TipBox.vue
```diff
@@ -251,86 +262,24 @@ export default {
...
-      case 'success':
-      case 'tip':
-        return 'alert-success';
-      case 'important':
-      case 'wrong':
-        return 'alert-danger';
-      default:
-        return 'alert-default';
-      }
+      console.log(this.bootStrapStyle);
```
jonahtanjz, line 265: Might want to remove this console.log

packages/vue-components/src/utils/utils.js
```diff
@@ -159,3 +159,74 @@ export function VueFixer (vue) {
...
+    = ['primary', 'secondary', 'success', 'danger', 'warning', 'info', 'light', 'dark', 'muted', 'white'];
+  
+  const typeStyle = defaultStyles.includes(type) ? type : '';
+  const themeStyle = colorStyles.includes(theme) ? theme : '';
+
+  let mainStyle;
+  let iconStyle;
+
+  if (themeStyle) {
+    mainStyle = themeStyle
```
jonahtanjz, line 178: Missing semi-colon :)

jonahtanjz (commented):
A few more minor changes @ong6. The rest looks good :)

jonahtanjz (approved):
LGTM 👍 Let's wait a while to see if @ang-zeyu has anything to add :)

packages/vue-components/src/utils/utils.js
```diff
@@ -159,3 +159,74 @@ export function VueFixer (vue) {
   vue.mixins.unshift(mixin)
   return vue
 }
+
+// Used in the TipBox component to classify the different styles used by bootstrap from the user input.
+// @params (the user input type of the box)
+export function classifyBootstrapStyle(type, theme) {
```
ang-zeyu, line 165: nice work! @ong6 @jonahtanjz 

Looks like a really clean consolidation 👍

packages/vue-components/src/utils/utils.js
```diff
@@ -159,3 +159,74 @@ export function VueFixer (vue) {
...
+  const defaultStyles
+    = ['warning', 'info', 'definition', 'success', 'danger', 'tip', 'important', 'wrong'];
+  const colorStyles
+    = ['primary', 'secondary', 'success', 'danger', 'warning', 'info', 'light', 'dark', 'muted', 'white'];
```
ang-zeyu, line 169: iirc `muted / white` are only available for bootstrap text colors, [alerts](https://getbootstrap.com/docs/4.0/components/alerts/) don't have this scheme

packages/vue-components/src/TipBox.vue
```diff
@@ -251,86 +262,23 @@ export default {
...
-      }
+      return this.bootStrapStyle.icon;
     },
   },
 };
 </script>
 
 <style scoped>
     .box-container {
-        width: 100%;
```
ang-zeyu, line 333: Ok with including this here. Try to keep future PRs as atomic as possible though, it really helps those reading the commit history to trace down bugs / etc.

docs/userGuide/syntax/boxes.mbdf
```diff
@@ -45,6 +45,47 @@ Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor i
...
 
+**The built in types can be colored.**
+
+<include src="codeAndOutput.md" boilerplate >
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+<box>
+    default
+</box>
+<box type="info" theme="primary">
```
ang-zeyu, line 56: Let's add the `theme` prop to the summary table down below too

packages/vue-components/src/utils/utils.js
ong6: Ok will remove the options for muted and white then!

docs/userGuide/syntax/boxes.mbdf
ong6: Got it!

packages/vue-components/src/TipBox.vue
ong6: Ok thanks for the feedback!

docs/userGuide/syntax/boxes.mbdf
```diff
@@ -45,6 +45,47 @@ Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor i
...
+<box type="info" theme="info">
+    info
+</box>
+<box type="info" theme="light">
+    light
+</box>
+<box type="info" theme="dark">
+    dark
+</box>
+<box type="info" theme="muted">
```
ang-zeyu, line 80: missed the ones here

ang-zeyu:
Lgtm! 👍

PR #1650, opened 2021-08-13
Update Commander.js version

### LINKED ISSUE #1648: Upgrade `commander` version to improve code quality

Main trigger for raising this issue is the following comment found in `package/cli/index.js`:
```js
// We want to customize the help message to print MarkBind's header,
// but commander.js does not provide an API directly for doing so.
// Hence we override commander's outputHelp() completely.
program.defaultOutputHelp = program.outputHelp;
program.outputHelp = function (cb) {
  printHeader();
  this.defaultOutputHelp(cb);
};
```

After some exploration of the existing commander.js [documentation](https://www.npmjs.com/package/commander#automated-help), new functionalities have been added in recent versions (v7 or up) that allowed for things like customizing help messages, such as via `.addHelpText()` function. 

Perhaps upgrade commandjs from the current "^3.0.2" to latest version available.
In addition, I think some refactoring work might be good to improve code quality for cli code. Some issues that be can addressed:
- configure help message with `.addHelpText()` or `.configureOutput()`
- print Markbind header via pre-action hooks instead of duplicating it in every command
- reduce long methods in "serve" command
- other misc refactorings

Pros:
- make use of the newly added features in commander.js instead of hacky fixes
- could take this opportunity to refactor the code for better readability

Cons:
- v3 to v8, might need to handle breaking changes with care

**Additional context**

I would like to work on this if @MarkBind/active-devs are agreeable to proceed (Or if you have other comments, please do let me know)

@wxwxwxwx9 Sorry for tagging, but errr nope I am not able to ping anyone with `@MarkBind/active-devs`, as seen above that it is greyed out 😂

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

Fixes #1648 
**Overview of changes:**
- updated commander.js to version "^8.1.0"
- replaced patching of help command to customize help text via `.addHelpText()`
- removed individual logging of MarkBind header as `.addHelpText()` triggers it automatically
- replaced custom help text for unknown command with `.showHelpAfterError()`
- other minor refactoring

I decided against separating the commands into individual files as that might not be that beneficial at the moment. However, I do think that for commands such as `markbind serve`, the code within is harder to understand due to the length and various different options that it is handling. After some tries, I think some part of the code from `package/core` (Site, Template etc) might need to be adjusted to handle different cases better. Due to my limited familiarity with MarkBind at the moment, I have decided to not work on it for now (as well as keeping the changes separate).

**Testing instructions:**
cli commands are covered by existing functional test cases.

Update Commander.js version

Some part of the existing cli code was written to patch behaviors
previously unsupported by `Commander.js`. The setting of custom 
header and modification to help text is now possible with newer 
version of `Commander.js`.

Let's update the version to "^8.1.0" to make use of the added functions.

This reduces monkey patching and improves code quality.

---

- [x] No unrelated changes

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -24,10 +24,6 @@ const liveServer = require('./src/lib/live-server');
 const cliUtil = require('./src/util/cliUtil');
 const logger = require('./src/util/logger');
 
-const {
-  ACCEPTED_COMMANDS,
-  ACCEPTED_COMMANDS_ALIAS,
-} = require('./src/constants');
 const CLI_VERSION = require('./package.json').version;
 
 process.title = 'MarkBind';
@@ -45,14 +41,14 @@ function handleError(error) {
   process.exitCode = 1;
 }
 
-// We want to customize the help message to print MarkBind's header,
-// but commander.js does not provide an API directly for doing so.
-// Hence we override commander's outputHelp() completely.
-program.defaultOutputHelp = program.outputHelp;
-program.outputHelp = function (cb) {
-  printHeader();
-  this.defaultOutputHelp(cb);
-};
+program.addHelpText('before', printHeader());
+
+program.showHelpAfterError('(run "markbind --help" to list commands)');
+
+program
+  .hook('preAction', () => {
+    printHeader();
+  });
 
 program
   .allowUnknownOption()
@@ -70,8 +66,6 @@ program
   .description('init a markbind website project')
   .action((root, options) => {
     const rootFolder = path.resolve(root || process.cwd());
-    const outputRoot = path.join(rootFolder, '_site');
-    printHeader();
     if (options.convert) {
       if (fs.existsSync(path.resolve(rootFolder, 'site.json'))) {
         logger.error('Cannot convert an existing MarkBind website!');
@@ -85,6 +79,7 @@ program
       .then(() => {
         if (options.convert) {
           logger.info('Converting to MarkBind website.');
+          const outputRoot = path.join(rootFolder, '_site');
           new Site(rootFolder, outputRoot).convert()
             .then(() => {
               logger.info('Conversion success.');
@@ -213,8 +208,6 @@ program
       mount: [],
     };
 
-    printHeader();
-
     site
       .readSiteConfig()
       .then(async (config) => {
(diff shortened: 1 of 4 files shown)


packages/cli/index.js
```diff
@@ -24,10 +24,6 @@ const liveServer = require('./src/lib/live-server');
 const cliUtil = require('./src/util/cliUtil');
 const logger = require('./src/util/logger');
 
-const {
-  ACCEPTED_COMMANDS,
-  ACCEPTED_COMMANDS_ALIAS,
-} = require('./src/constants');
 const CLI_VERSION = require('./package.json').version;
 
 process.title = 'MarkBind';
```
tlylt, line 29: I am not sure what's the purpose of line 29 and also line 30-32. I think I am able to remove them without causing any visual differences. But just to be safe, I will leave them in. If anyone who's reviewing knows what those are for, please let me know 👀

jonahtanjz: Looks like this is to make it easier to identify/manage/log the process in the process manager. I might be wrong but I think we should keep it as it is.

packages/cli/index.js
```diff
@@ -24,10 +24,6 @@ const liveServer = require('./src/lib/live-server');
 const cliUtil = require('./src/util/cliUtil');
 const logger = require('./src/util/logger');
 
-const {
-  ACCEPTED_COMMANDS,
-  ACCEPTED_COMMANDS_ALIAS,
```
jonahtanjz, line 29: Since `ACCEPTED_COMMANDS` and `ACCEPTED_COMMANDS_ALIAS` constants were only used here, what do you think of removing the `contants.js` file since it is no longer in use?

jonahtanjz (commented):
Nice work @tlylt 👍 Thanks for suggesting and working on this issue :)

packages/cli/index.js
jonahtanjz: Looks like this is to make it easier to identify/manage/log the process in the process manager. I might be wrong but I think we should keep it as it is.

jonahtanjz:
>  However, I do think that for commands such as markbind serve, the code within is harder to understand due to the length and various different options that it is handling. After some tries, I think some part of the code from package/core (Site, Template etc) might need to be adjusted to handle different cases better. Due to my limited familiarity with MarkBind at the moment, I have decided to not work on it for now (as well as keeping the changes separate).

Agree on this. The `serve` function can use a refactoring. You can always open an issue on this and come back to it once you are more comfortable and familiar with MarkBind :)

packages/cli/index.js
tlylt: Hi @jonahtanjz , I think `constants.js` has already been removed in this PR:)

jonahtanjz: Oops 😬 Sorry missed it 😅

jonahtanjz (approved):
LGTM 👍

PR #1653, opened 2021-08-17
Update linebreak-style settings

### LINKED ISSUE #1640: Line ending issues when developing in Windows environment
Windows 10 Home

**MarkBind Version:**
3.0.6

Hello,

I am trying to set up markbind locally following the contributing guide. As I am using a Windows machine and according to #777, line endings are set as "CRLF" when I cloned it locally. I am using VSCode btw.

After installing dependencies, I expect to run the test cases via `npm run test`. 

However, I am greeted with this lint [error](https://eslint.org/docs/rules/linebreak-style)  all over the place. 
>Expected linebreaks to be 'LF' but found 'CRLF'.eslintlinebreak-style

I assume this is not an uncommon issue but I could not find out what's the recommended practice in the repo's past issues (I found #777 and #1213 though). I suppose according to #777, all line endings will be checked in as LF, and it should be converted back to CRLF if a developer checkouts out to a Windows machine. This is because of `* text=auto`. If that is the case, how does one bypass the Eslint error locally in a Windows machine?

My temporary fix is to add `"linebreak-style": 0` within `.eslinrc.js`... I would love to hear from the devs if I missed out on something or am I doing something wrong here?

Thanks!

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

Fixes #1640 
**Overview of changes:**
- update `.gitattributes` to enforce `LF`. Referencing [eslint](https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems) and similar setting in [docusaurus](https://github.com/facebook/docusaurus/blob/master/.gitattributes#L3)

**Testing instructions:**

Update linebreak-style settings

Code editors complain about linebreak-style issues due to the
line ending conversion across OS. 

Let's update `.gitattributes` to enforce `LF`.

---

- [x] No unrelated changes


.eslintrc.test.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  env: {
```
jonahtanjz, line 2: Really minor nit, perhaps can consider using double quotes here, `"env"`, so that it is consistent with the `eslintrc.js` file. Similar comment for the rest of the file :)

package.json
```diff
@@ -6,9 +6,11 @@
     "csslint": "stylelint **/*.css **/*.vue",
     "csslintfix": "stylelint **/*.css **/*.vue --fix",
     "lint": "eslint . --ext .js,.vue && npm run csslint",
-    "lintfix": "eslint . --ext .js,.vue --fix && npm run csslintfix",
+    "lintlinebreakstyle": "eslint . --ext .js,.vue -c .eslintrc.test.js",
+    "lintlinebreakstylefix": "eslint . --ext .js,.vue -c .eslintrc.test.js --fix",
+    "lintfix": "eslint . --ext .js,.vue --fix && npm run lintlinebreakstylefix && npm run csslintfix",
     "setup": "npm ci && lerna clean --yes && lerna bootstrap --hoist",
-    "test": "npm run lint && lerna run test --stream",
+    "test": "npm run lint && npm run lintlinebreakstyle && lerna run test --stream",
```
ang-zeyu, line 13: hmm would this work?

https://eslint.org/docs/developer-guide/shareable-configs

so no extra commands, and maybe we can rename `lintlinebreakstyle` to something more semantic / shorter like `linttest` 😄

tlylt: Not sure what exactly are you referring to about shareable configs but I removed the extra commands. Will it be better?

I rethink what I wanted to achieve with this PR:
1. It's annoying to have a full page of squiggly lines (Eslint warning) when a file is in CRLF
2. Still want a way to check the line ending issue before committing the files

So the following changes are done:
- update .gitattributes to further ensure line ending for .vue and .js files stay in LF (tries to eliminate the root cause)
- turn off linebreak-style check in main eslintrc file (tries to solve 1)
- specify linebreak-style check in eslintrc.test.js as there is no way to specify a single rule in cli (tries to solve 2)
- update npm script to ensure when doing linting and testing, linebreak-style is checked (tries to solve 2)

So the intended dev flow is as follows:
- no line ending issues: nothing affected
- has a few files somehow saved into CRLF in development: continue to develop until finish the PR, then run lint/test, then realizing line ending issues for a bunch of files, then run `lintfix` once, then commit and PR.

ang-zeyu: @tlylt this part in particular https://eslint.org/docs/developer-guide/shareable-configs#using-a-shareable-config, you can extend and override configurations in eslint, so there's no need to run `eslint` twice here.

```
eslint . --ext .js,.vue && eslint . --ext .js,.vue -c .eslintrc.test.js
```

ang-zeyu: > update .gitattributes to further ensure line ending for .vue and .js files stay in LF (tries to eliminate the root cause)

This shouldn't be needed with your other changes here and the existing `* text = auto` either, you're likely getting squiggly red lines in your ide not because of gitattributes but the root `.eslintrc`.

tlylt: Yea that is added to hopefully keep js and vue files to LF instead of automatically changing to CRLF whenever saving or checking out via git.

tlylt: > @tlylt this part in particular https://eslint.org/docs/developer-guide/shareable-configs#using-a-shareable-config, you can extend and override configurations in eslint, so there's no need to run `eslint` twice here.
> 
> ```
> eslint . --ext .js,.vue && eslint . --ext .js,.vue -c .eslintrc.test.js
> ```
Oh I get what you mean now. Yea so that was the first thing I tried, by making `.eslintrc.test.js` extend from `.eslintrc.js`. 
The issue is: 
if I specify this file to be used in `npm run lint`, when eslint is ran in packages, the sub eslintrc files won't take over.

ang-zeyu: Could we do this for all text files actually? per the docusaurus setting you refernced https://github.com/facebook/docusaurus/blob/main/.gitattributes

So no need to add `.ts` / etc. in the future

> Oh I get what you mean now. Yea so that was the first thing I tried, by making .eslintrc.test.js extend from .eslintrc.js.

Thanks for trying, this didn't occur to me earlier.

Actually, with `eol = lf` (which presumably prevents crlf conversion on check *out*), are the eslint changes still necessary? I gave your pr a few tries without the eslint changes and all seems good (no squiggly red lines) even without those.

The existing eslint settings would then still check for line endings as is.

ang-zeyu: @tlylt small ping in case you missed my earlier message

tlylt: Hi @ang-zeyu, thanks for the reminder! I wanted to verify what we discussed one more time but have been busy lately... will update this PR by next week:)

tlylt: Hi @ang-zeyu, I tested the branch a little bit more, it seems to be that if you have converted a file from CRLF to LF locally and staged the change (The line-ending change can be staged, but after u staged it, the change is no longer visible in git, perhaps due to the .gitattribute settings), then the file will remain as LF in all local branches. In that case, the Eslint warning will cease to exist.

I am not sure if adding `* text eol=lf` will be beneficial anymore as it appears to have no effect that I intended it to have. i.e. turn all CRLF or LF back into LF when writing out into the working directory.

Should I close this PR and the relevant issue, or should I adjust the PR to just add `* text eol=lf` nonetheless?

ang-zeyu: > I tested the branch a little bit more, it seems to be that if you have converted a file from CRLF to LF locally and staged the change

this is expected but might not be an ideal solution since you lose out on the functionality of staging / working areas.

> I am not sure if adding * text eol=lf will be beneficial anymore as it appears to have no effect that I intended it to have. i.e. turn all CRLF or LF back into LF when writing out into the working directory.

Strange, its not working for me either as well anymore (files are always LF). Might be an ide update / bad testing method the last time

How about adding just `* text=auto eol=lf` in nonetheless and see how things go for a while?

If it still occurs, we can add in the `.eslintrc` changes here.

> Oh I get what you mean now. Yea so that was the first thing I tried, by making .eslintrc.test.js extend from .eslintrc.js.
>The issue is:
>if I specify this file to be used in npm run lint, when eslint is ran in packages, the sub eslintrc files won't take over. I think at least I was experiencing "env" issues as sub eslintrc specifies different "env" from the main eslintrc.

PS: just realised there's [this](https://eslint.org/docs/user-guide/command-line-interface#--rule), it might solve this issue

tlylt: Hi @ang-zeyu , I have updated the PR as per your comment. 
> PS: just realised there's this, it might solve this issue

Not sure why I didn't use this inline rule back then, but thanks! This might be a better approach than having separate eslintrc files.

ang-zeyu (approved):

package.json
ang-zeyu: Thanks! Lgtm

Issue #1667, opened 2021-09-12
deploy command declares success prematurely


damithc:
It seems `markbind deploy` command shows the `Deployed at ...` message before the actual deployment is completed?
The reason for my suspicion is that the console stays busy for a quite while after that message is displayed.

![image](https://user-images.githubusercontent.com/1673303/132977756-5435969e-c3fa-49b7-b1f1-40a4ada7b8f3.png)

ong6:
I've experienced this issue as well. Might be something to do with the async functions in index.js of cli. I will spend some time trying to fix this issue and report back.

ong6:
@damithc Hi, please help me close this issue as I forgot to tag this issue as 'To close' in my PR. Sorry for the inconvenience!

PR #1679, opened 2021-10-07
Fix premature declaration of link in deploy command

ong6:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
It Fixes the problem highlighted by Prof. Damith at #1667 as well as #1680

Added an  await function to one of the methods in the deploy function() to prevent the command `markbind deploy` from declaring prematurely. Also made a clearer output message which will improve UX.

Example of new `markbind deploy` (note the console stops when the message is printed):
deploy

Nil

**Testing instructions:**
Nil

Fixed output for `markbind deploy` to remove the console lag.

---

- [x] No unrelated changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1930,7 +1930,6 @@
 				"jest-resolve": "^26.6.2",
 				"jest-util": "^26.6.2",
 				"jest-worker": "^26.6.2",
-				"node-notifier": "^8.0.0",
 				"slash": "^3.0.0",
 				"source-map": "^0.6.0",
 				"string-length": "^4.0.1",
@@ -2549,9 +2548,6 @@
 			"resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
 			"integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
 			"dev": true,
-			"dependencies": {
-				"graceful-fs": "^4.1.6"
-			},
 			"optionalDependencies": {
 				"graceful-fs": "^4.1.6"
 			}
@@ -2636,9 +2632,6 @@
 			"resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
 			"integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
 			"dev": true,
-			"dependencies": {
-				"graceful-fs": "^4.1.6"
-			},
 			"optionalDependencies": {
 				"graceful-fs": "^4.1.6"
 			}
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1930,6 +1930,7 @@
 				"jest-resolve": "^26.6.2",
 				"jest-util": "^26.6.2",
 				"jest-worker": "^26.6.2",
+				"node-notifier": "^8.0.0",
```
ang-zeyu, line 1933: Lgtm 👍, except the package-lock changes here.

Does removing all `node_modules` and rerunning `npm run setup` cut it?

packages/cli/index.js
```diff
@@ -330,7 +330,9 @@ program
     const rootFolder = path.resolve(process.cwd());
     const outputRoot = path.join(rootFolder, '_site');
     new Site(rootFolder, outputRoot, undefined, undefined, options.siteConfig).deploy(options.ci)
-      .then(depUrl => (depUrl !== null ? logger.info(`Deployed at ${depUrl}!`) : logger.info('Deployed!')))
+      .then(depUrl => (depUrl !== null ? logger.info(
+        `The website will be available for viewing in a few moments at: ${depUrl}`)
```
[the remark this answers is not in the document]
ong6: Issue is that once the command runs, the website cannot be viewed immediately, hence I phrased it as such.

package-lock.json
ong6: Is there any way for me to uncommit the package-lock changes? Removing the node modules and rerunning npm run setup does not fix this.

ang-zeyu: > Removing the node modules and running npm run setup does not fix this.

Sorry, this should follow the `git revert` above. Post back with your environment (npm version, OS, ...) if you're still getting these strange changes after reverting 😬, might be an issue with the dev setup.

ong6: Can I ask if changing the package-lock file here even matter? Since everyone should be running `npm run setup` on the first pull, Im not sure if this makes a difference. 

But I think that the package-lock file is updated cause I pulled it from the newest merge that someone else did.

ang-zeyu: @ong6 you are mostly right that it dosen't matter from a environment setup pov.
However, It's an unrelated change to this PR there are no dependency edits here.
If it occured for a reason (and it should be found), we should be fixing it in a separate, focused commit and pr for clarity (where do we more specific testing as well -- e.g. on different OS'es)

ang-zeyu (approved):
Lgtm 🙂

Issue #781, opened 2019-03-18
Write unit tests for plugins


jamos-tay:
Follow up from #714 

Plugins, such as `anchors` and `filterTags` can be unit tested, which will help some of the edge cases that are harder to test with functional tests.

PR #1684, opened 2021-10-31
Include unit tests for default plugins

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
This PR partially addresses #781. I have included unit tests for default plugins as a means to complement existing functional tests.

In the unit test for `markbind-plugin-plantuml`, I decided to mock `fs` and `child_process` modules such that `generateDiagram` will not actually launch `plantuml.jar` to convert plantuml code into an image file. This is because that process seems to be covered in functional tests as plantuml diagrams will be generated in the `expected/diagrams` folder. I did not unit test the `generateDiagram` function because it is not an exported function. I am not sure if that is a good reason but I am thinking since it is somewhat private and inaccessible to other modules, perhaps I should not test it.

**Testing instructions:**
Run `npm run test` to verify that all tests are still passing. To only run the 3 added tests, cd into `packages/core/test/unit/plugins/default` and run 
`npx jest anchor.test.js`, or
`npx jest plantuml.test.js`, or
`npx jest shorthandSyntax.test.js`

Include unit tests for default plugins

There are severe plugins that have specific functionalities 
but are not unit tested. 

Some of the edge cases might be harder to test with functional tests.

Let's add unit tests for default plugins to verify their functionalities. 
For example, `processNode` should correctly modify the 
`node` passing in according to a plugin's purpose.

---


packages/core/test/unit/plugins/default/shorthandSyntax.test.js
```diff
@@ -0,0 +1,39 @@
...
+    + '<span>Heading</span></panel>', true)[0].children;
+  const copy = { ...spanNode };
+  shorthandSyntaxPlugin.processNode({}, spanNode);
+  expect(spanNode).toEqual(copy);
+},
+);
+
+test('processNode should not convert improper syntax nodes', () => {
+  const [divSpanNode] = cheerio
+    .parseHTML('<div><span heading>Heading</span></div>', true)[0].children;
```
wxwxwxwx9, line 28: I think it would be good to add short inline comments to explain how are these syntaxes improper, so readers don't have to refer to the function to infer what the proper syntax should be. It also serves as also a form of verification of what improper syntax we are actually testing.

packages/core/test/unit/plugins/default/shorthandSyntax.test.js
```diff
@@ -0,0 +1,39 @@
...
+  expect(spanNode.type).toEqual('tag');
+  expect(spanNode.name).toEqual('span');
+  expect(spanNode.attribs.class).toEqual('card-title');
+  expect(spanNode.attribs.slot).toEqual('header');
+  expect(spanNode.children[0].data).toEqual('Heading');
+  expect(spanNode.children[0].type).toEqual('text');
+},
+);
+
+test('processNode should not convert span node without heading attribute', () => {
```
wxwxwxwx9, line 17: I think maybe this can be included in the "improper syntax nodes" as well, but it's fine if we leave it here also. I leave it up to you :-)

wxwxwxwx9 (changes requested):
Hi @tlylt! Good job on this PR :-) I've looked through your changes and left some comments for discussion/changes. 

The tests for PlantUML looks good to me, I think the control flows for the function is well-tested. Don't think there is anything we missed out. 

For the other 2 tests, I have written some minor comments. 

Thanks for your contribution!

packages/core/test/unit/plugins/default/shorthandSyntax.test.js
tlylt: Hi, thanks for the comment. I have adjusted the test cases for the `shorthandSyntax` plugin and included in a comment at the top to show the desired outcome. Hope that's clearer.

packages/core/test/unit/plugins/default/shorthandSyntax.test.js
tlylt: I decided to split the improper syntax test case into two so that it's clearer as to why it's improper. So I will keep this one separate as well. Thank you for the suggestion.

wxwxwxwx9 (approved):

Issue #1560, opened 2021-04-30
Improve the fixed header approach to remove "header padding fouc"


ang-zeyu:
**Is your request related to a problem?**

Browser refresh:
![fixed-header-approach](https://user-images.githubusercontent.com/3306138/116706492-46d4b880-aa00-11eb-8a30-04436061c440.gif)

**Describe the solution you'd like**

Experiment with a dynamic positioning approach, e.g. https://www.w3schools.com/howto/howto_js_sticky_header.asp.

**Describe alternatives you've considered**
- Add a "best estimate" of the initial header height -- issue: some pages have headers, some don't. Also only an estimate, some fouc would still remain.

**Additional context**

#1534 eliminates fouc from vue rendering but fouc from the style detection of headers at runtime remains.

ang-zeyu:
Some existing solutions / compromises:
- `position: fixed` + we know the height of the header beforehand (e.g. see https://vuepress.vuejs.org/guide/): unfortunately not an option in our case (by far the most common one around)
- `position: sticky` + accept the con of shorter elements being hidden behind the header as you scroll down (see https://webpack.js.org/concepts/output/'s left sidebar just for illustration)
- `position: sticky` + dynamically hide the header like what we're doing for mobile (e.g. https://docusaurus.io/docs)

In general it seems like there is no "easy / ideal way out". The core of the problem is that style detection of the header height is required, something that isn't too common.

ang-zeyu:
I ended up sidetracking here and there and drafted 3 solutions below on top of the OP's, all of which achieve the aim of "no more fouc" but each has pros and cons.

1. (OP's) https://github.com/MarkBind/markbind/pull/1929
   - simplest option (implementation wise) that activates `position: fixed` only when the reader scrolls past the header. our `fixed-header-padding` class's height is also only "populated" once this is done.
   - pros: simple implementation
   - cons: not sticky (no element can be above the header) (e.g. announcement at the top of the page but should not follow the header as you scroll downward)
2. (~70% complete, edge cases with window resizing remain) https://github.com/MarkBind/markbind/pull/1930
   - this is the most complex option. It implements a sticky header using `position: fixed` + scrolling detection.
   - pros: `position: sticky` positioning rules / nuances do not have to be followed (e.g. sticky element must be placed as a direct child of "scroll container"), increasing author convenience. Vs opt 3, parts of the sidebars will not be hidden too.
[...]

PR #1687, opened 2021-11-09
[DO NOT MERGE] Move some script elements (demo)

riyadh-h:
These `script` elements found in `Page.njk` have been moved down to be just above the `body` closing tag:

```
<script src="{{ asset.vue }}"></script>
<script src="{{ asset.markBindJs }}"></script>
<script src="{{ asset.pageVueRenderJs }}"></script>
<script src="{{ asset.jQuery }}"></script>
```

There are two commits: one for where the script elements are down below, and one for where they're up above. You can check-out between these commits to see the changes without manually modifying the code.

**Testing instructions:**

0. (optional) Create a new MarkBind project.
1. Check-out the commit that you want to test.
2. Serve the MarkBind project.
* Note: the following steps assume the use of Edge/Chrome; the same (or similar) instructions might be applicable to other browsers.
3. Go to the MarkBind serving address (if it isn't open already).
4. Right-click anywhere on the page and select _Inspect_.
5. Click on the _Network_ tab, then turn on _Disabling Cache_ and set the throttling option to _Slow 3G_.
6. Refresh the page; once the page is fully-loaded, a timestamp with the label _Load_ will appear on the bottom of _Network_ tab.
    * If the scripts are at the bottom during this step, then the [header padding FOUC](https://github.com/MarkBind/markbind/issues/1560) should take more time to happen.
7. Repeat Step 6 if more tests are required for the current commit.
8. Once finished, stop the MarkBind serving process.
9. Repeat Steps 1-8 for the other commit if necessary.

### DIFF
--- a/packages/core/src/Page/page.njk
+++ b/packages/core/src/Page/page.njk
@@ -20,10 +20,6 @@
     <link rel="stylesheet" href="{{ asset.highlight }}">
     {%- if not dev -%}<link rel="stylesheet" href="{{ asset.markBindCss }}">{%- endif -%}
     <script src="{{ asset.polyfillJs }}"></script>
-    <script src="{{ asset.vue }}"></script>
-    <script src="{{ asset.markBindJs }}"></script>
-    <script src="{{ asset.pageVueRenderJs }}"></script>
-    <script src="{{ asset.jQuery }}"></script>
     {%- if asset.pluginLinks -%}
     {%- for link in asset.pluginLinks -%}
         {{ link }}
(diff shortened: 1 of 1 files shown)


ang-zeyu:
Thanks for the instructions @riyadh-h 

Some delay to header padding FOUC is to be expected (relative to the first content shown), since we're doing dynamic style detection here, and the scripts are now downloaded and run later. I think its an ok tradeoff, especially considering the underlying issue with the headers can be fixed separately.

As for the other forms of FOUC, our stylesheets are kicking in later as they are injected by [this webpack plugin](https://webpack.js.org/loaders/style-loader/) during development (`-d` cli option for hot reload of styles). Other builds (e.g. netlify) shouldn't have this issue as the css is extracted and [bundled](https://webpack.js.org/plugins/mini-css-extract-plugin/).

If you are up for it, the latter plugin now supports hot module replacement as well, I think we could experiment with that to fix this for `-d`. But otherwise we can fix it separately later on too (not high priority since its a developer issue); We can focus on whether everything still functions correctly for now (if anything in the `usingComponents` page breaks in particular).

ang-zeyu:
Closing this as its been stale for a while, but this is something we'd very much like to have done still.

PR #1690, opened 2021-12-07
Change top nav from accordion to two rows on mobile

### LINKED ISSUE #980: Redesign the navbar

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Resolves #980 

Remove current accordion for top navigation on mobile and shift it to create a second row sliding menu. `slot="right"` item will now be visible and positioned on the right side instead of being in an accordion.

Before:
image

After:
image

NIL

**Testing instructions:**
Preview markbind site with updated top nav design: https://deploy-preview-1690--markbind-master.netlify.app/index.html

Change top nav from accordion to two rows on mobile

The default slot and items in the right slot are currently
hidden in an accordion on mobile.

Let's shift the default slot items to a new row so that they are
more accessible and right slot items can now be placed on the
right side of the top nav.

---

- [x] No unrelated changes

### DIFF
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -132,7 +132,7 @@ export default {
     },
     hideDropdownMenu() {
       this.show = false;
-      $(this.$refs.dropdown).findChildren('ul').each(ul => {
+      $(this.$refs.dropdown).findChildren('ul').each((ul) => {
         ul.classList.toggle('show', false);
 
         if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
@@ -142,7 +142,7 @@ export default {
     },
     showDropdownMenu() {
       this.show = true;
-      $(this.$refs.dropdown).findChildren('ul').each(ul => {
+      $(this.$refs.dropdown).findChildren('ul').each((ul) => {
         ul.classList.toggle('show', true);
 
         // check if the dropdown is part of the sliding menu on mobile
@@ -195,7 +195,7 @@ export default {
             overflow-y: auto;
             overscroll-behavior: contain;
         }
-        
+
         .navbar-default .dropdown-menu-right {
             right: auto;
         }
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -56,6 +56,7 @@
 import Submenu from './Submenu.vue';
 import { toBoolean } from './utils/utils';
 import $ from './utils/NodeList';
+import preventOverflowOnMobile from './utils/dropdown';
 
 export default {
   components: {
@@ -131,11 +132,24 @@ export default {
     },
     hideDropdownMenu() {
       this.show = false;
-      $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', false));
+      $(this.$refs.dropdown).findChildren('ul').each(ul => {
+        ul.classList.toggle('show', false);
+
+        if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
+          ul.style.removeProperty('left');
+        }
+      });
     },
     showDropdownMenu() {
       this.show = true;
-      $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', true));
+      $(this.$refs.dropdown).findChildren('ul').each(ul => {
+        ul.classList.toggle('show', true);
+
+        // check if the dropdown is part of the sliding menu on mobile
+        if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
+          preventOverflowOnMobile(ul);
+        }
+      });
     },
   },
   mounted() {
@@ -169,6 +183,24 @@ export default {
 </script>
 
 <style scoped>
+    @media (max-width: 767px) {
+        .navbar-default .dropdown {
+            position: static;
+        }
+
+        .navbar-default .dropdown-menu {
+            position: absolute;
+            max-width: 100%;
+            max-height: 75vh;
+            overflow-y: auto;
+            overscroll-behavior: contain;
+        }
+        
+        .navbar-default .dropdown-menu-right {
+            right: auto;
+        }
+    }
+
     .secret {
         position: absolute;
         clip: rect(0 0 0 0);
--- a/packages/vue-components/src/Dropdown.vue
+++ b/packages/vue-components/src/Dropdown.vue
@@ -85,6 +85,9 @@ export default {
     hasParentDropdown: {
       default: undefined,
     },
+    isParentNavbar: {
+      default: false,
+    },
   },
   data() {
     return {
@@ -135,7 +138,7 @@ export default {
       $(this.$refs.dropdown).findChildren('ul').each((ul) => {
         ul.classList.toggle('show', false);
 
-        if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
+        if (window.innerWidth < 768 && this.isParentNavbar) {
           ul.style.removeProperty('left');
         }
       });
@@ -146,7 +149,7 @@ export default {
         ul.classList.toggle('show', true);
 
         // check if the dropdown is part of the sliding menu on mobile
-        if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
+        if (window.innerWidth < 768 && this.isParentNavbar) {
           preventOverflowOnMobile(ul);
         }
       });
@@ -219,6 +222,8 @@ export default {
 
     .dropdown-toggle {
         cursor: pointer;
+        display: block;
+        width: max-content;
     }
 
     .navbar .dropdown-toggle {
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -5,7 +5,7 @@
         <div class="navbar-brand">
           <slot name="brand"></slot>
         </div>
-        <div class="navbar-top">
+        <div ref="navbarDefault" class="navbar-default">
           <ul class="navbar-nav mr-auto mt-2 mt-lg-0">
             <slot></slot>
           </ul>
@@ -245,10 +245,30 @@ export default {
 
     this.toggleLowerNavbar();
     $(window).on('resize', this.toggleLowerNavbar);
+
+    // scroll default navbar horizontally when mousewheel is scrolled
+    const navbarDefault = this.$el.querySelector('.navbar-default');
+    $(navbarDefault).on('wheel', (e) => {
+      const isDropdown = nodes => {
+        for (let i = 0; i < nodes.length; i += 1) {
+          if (nodes[i].classList && nodes[i].classList.contains('dropdown-menu')) {
+            return true;
+          }
+        }
+        return false;
+      }
+
+      // prevent horizontal scrolling if the scroll is on dropdown menu
+      if (window.innerWidth < 768 && !isDropdown(e.path)) {
+        e.preventDefault();
+        navbarDefault.scrollLeft += e.deltaY;
+      }
+    })
   },
   beforeDestroy() {
     $('.dropdown', this.$el).off('click').offBlur();
     $(window).off('resize', this.toggleLowerNavbar);
+    $(this.$el.querySelector('.navbar-default')).off('wheel');
   },
 };
 </script>
@@ -272,7 +292,7 @@ export default {
             max-width: 40vw;
         }
 
-        .navbar-top {
+        .navbar-default {
             display: block;
             margin-top: 0.3125rem;
             width: 100%;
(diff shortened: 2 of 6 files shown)


ang-zeyu:
Nice!

Reviewing the design first; I think it looks good overall, just 2 parts that need to be fixed:
- the horizontal scrollbar dosen't seem to be functional on desktop chrome with mouse scroll wheels (works fine for touchscreens)
-  Could we preserve the look of the dropdown? (absolutely-positioned with respect to the "Dropdown" text so it dosen't add to the height of the scrollbar) (including the searchbar, which uses some common css classes iirc)
   Untitled

jonahtanjz:
Hi @ang-zeyu thanks for the review. I've made the changes accordingly.

> the horizontal scrollbar dosen't seem to be functional on desktop chrome with mouse scroll wheels (works fine for touchscreens)

I believe to scroll horizontally with the mouse wheel can hold down `SHIFT` and scroll but this doesn't really seem intuitive 😅 I've bind the mouse wheel to the horizontal scroll on the mobile navbar and this should resolve the issue.

> Could we preserve the look of the dropdown? (absolutely-positioned with respect to the "Dropdown" text so it dosen't add to the height of the scrollbar) (including the searchbar, which uses some common css classes iirc)

Yup agree that it will look better. I've updated the design so that it is now absolute positioned and also added scrolling if the dropdown overflows the bottom of the screen.

image

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            padding-right: 0;
+            padding-bottom: 0;
+        }
+
+        div.navbar-brand {
+            padding-left: 1rem;
+        }
+
+        .navbar-right {
+            position: absolute;
```
ang-zeyu, line 289: try the flex [`order`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items) property; with the right combination you should be able to remove the reliance on `absolute` positioning (not too good in this case as its sizing should be taken into account in the document flow, e.g. if `navbar-brand` is empty this overflows the container)

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            top: 14px;
+            right: 16px;
+            max-width: 40vw;
```
ang-zeyu, line 292: `.navbar-brand` might need this as well.

could we use `50%` too? ( so each part takes up at most half then wraps vertically if needed )

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+        }
+
+        .navbar-default ul {
+            flex-direction: row;
+            margin-top: 0 !important;
+            width: 100%;
+        }
+
+        .navbar-default > ul > li,
+        .navbar-default > ul > div {
```
ang-zeyu, line 318: how about just `.navbar-default > ul > *`? since there's no restriction on what can be placed in there 😃

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            padding: 0.3125rem 0.625rem;
+            flex: 1;
+            background: rgba(0, 0, 0, 0.3);
```
ang-zeyu, line 321: could be moved to `.navbar-default`, so items can set their own backgrounds

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            text-align: center;
```
ang-zeyu, line 322: shouldn't be needed, I think.

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+        }
+
+        .navbar-default a,
+        >>> .dropdown-toggle {
+            margin: 0 auto;
```
ang-zeyu, line 327: this seems to push the last few item(s) after the dropdown to the right (if there's extra space)

tried removing it and looks good; what's it for?

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            width: max-content;
```
ang-zeyu, line 328: (just for `>>> .dropdown-toggle`) Could we apply this fix generally to desktop view in dropdown.vue as well? seems to occur there too.

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+        }
+
+        >>> .dropdown-toggle {
+            display: block;
```
ang-zeyu, line 332: per above

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+            padding: 0.5rem 0;
```
ang-zeyu, line 333: how about
```
display: flex;
align-items: center;
```
(to avoid tying it to `0.5rem`)

ang-zeyu (commented):
> I believe to scroll horizontally with the mouse wheel can hold down SHIFT and scroll but this doesn't really seem intuitive 😅 I've bind the mouse wheel to the horizontal scroll on the mobile navbar so this should resolve the issue.

forgot about that 😅, but yes, agree it isn't as intuitive



I only glossed over the positioning / overflow implementation which seems quite complex, it seems to do the job quite well though!

Mostly nits on the styling implementation otherwise:

packages/vue-components/src/Dropdown.vue
```diff
@@ -131,11 +132,24 @@ export default {
     },
     hideDropdownMenu() {
       this.show = false;
-      $(this.$refs.dropdown).findChildren('ul').each(ul => ul.classList.toggle('show', false));
+      $(this.$refs.dropdown).findChildren('ul').each((ul) => {
+        ul.classList.toggle('show', false);
+
+        if (window.innerWidth < 768 && this.$refs.dropdown.closest('div.navbar-default') !== null) {
```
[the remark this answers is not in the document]
jonahtanjz: Yup this seems alot better. Thanks!

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +239,98 @@ export default {
...
+        e.preventDefault();
+        navbarDefault.scrollLeft += e.deltaY;
+      }
+    });
   },
   beforeDestroy() {
     $('.dropdown', this.$el).off('click').offBlur();
-    if (this.slots.collapse) $('[data-toggle="collapse"]', this.$el).off('click');
     $(window).off('resize', this.toggleLowerNavbar);
+    $(this.$el.querySelector('.navbar-default')).off('wheel');
```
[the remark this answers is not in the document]
jonahtanjz: Yup using the ref works. I did this part before creating the ref and forgot to update it. Thanks for the catch

packages/vue-components/src/Navbar.vue
```diff
@@ -305,6 +361,11 @@ export default {
     }
 
     /* Navbar link highlight for current page */
+    .navbar .navbar-nav >>> .current:not(.dropdown),
+    .navbar .navbar-nav >>> .dropdown.current {
+        background: none;
```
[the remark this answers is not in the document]
jonahtanjz: This is so that when a link in the dropdown is highlighted, the dropdown itself will be highlighted as well.

image

packages/vue-components/src/Navbar.vue
jonahtanjz: Hmmm I was thinking that if a link on the navbar is highlighted, its background should also change to the main colour of the navbar. If we move this to `.navbar-default`, then I don't think we can do the background highlighting for the individual link. Any suggestion?

image

packages/vue-components/src/Navbar.vue
jonahtanjz: Oh this is for center aligning the text in the default navbar in some screen sizes when the width of each link is slightly larger.

With `margin: 0 auto`:
image

Without `margin: 0 auto`:
image

jonahtanjz: Ahh ok got it. I've flushed all items to the left and it looks alot better.

packages/vue-components/src/Navbar.vue
jonahtanjz: > The alternative I had in mind was to move the said color to .navbar-default, then forward the navbar main colour to .current specifically. This means forwarding .navbar-light/dark/primary's colours to the .current element manually.

I've now shifted the background overlay to the default navbar instead of the individual link. I've also forwarded the navbar colour to a css variable for `.current` to use.

> One slight issue with 0.3 specifically though - you'll likely need to try out / set a different value for .navbar-light, 0.3 seems a little dark for it.

Yup agree with this. I've decreased the overlay for primary and dark to `0.2` and `0.1` for light. Should be better now.

Primary (`0.2`):
image

Dark (`0.2`):
image

Light (`0.1`):
image

packages/vue-components/src/Navbar.vue
```diff
@@ -254,34 +240,97 @@ export default {
...
+
+        div.navbar-brand {
+            padding-left: 1rem;
+            max-width: 50%;
+            order: 1;
+        }
+
+        .navbar-right {
+            order: 1;
+            max-width: 50%;
```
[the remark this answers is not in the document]
jonahtanjz: > also need to take note 50% does not account for the margin-right set by navbar-brand in bootstrap

Thanks for the catch! I've factored the margin into the `max-width` now and also set `white-space: normal` so that items within `navbar-brand` can wrap to next line when it exceeds `50%`.

> on second thought let's keep in mind and fix separately; found a few more issues that seems to be out of scope here.
But if not we can always keep in mind to to fix soon in the next PR or so, and just focus on the mobile version here.

Yup sure. I can open an issue after this PR to keep track of it :)

packages/vue-components/src/Navbar.vue
```diff
@@ -318,8 +332,10 @@ export default {
 
         .navbar-default > ul > * {
             padding: 0.3125rem 0.625rem;
-            flex: 1;
-            background: rgba(0, 0, 0, 0.3);
```
ang-zeyu: 👀

> ok, I see the point now, I think this works best in that case.

I actually liked the old approach with setting backgroud colour on the individual `ul>*`'s  https://github.com/MarkBind/markbind/pull/1690#discussion_r771130744. (just need a separate `.navbarlight .navbar-default > ul > * { background: ... }` of sorts)

I think dynamic style detection (`--mobile-link-highlight-color`) is something we should avoid when possible as it may introduce more noticeable / distracting visual changes after the page has been presented already. (try seeing the netlify on a slow enough mobile device, you'll get to see the background colour change)

packages/vue-components/src/Navbar.vue
```diff
@@ -283,17 +288,22 @@ export default {
 
         div.navbar-brand {
             padding-left: 1rem;
-            max-width: 50%;
+            max-width: calc(50% - 1rem);
```
ang-zeyu, line 291: should we define our own `.navbar-brand` (`.navbar-left`?) class of sorts for use here to avoid relying on `1rem` specifically?

I did a very brief look over the original definition and it seems like it only sets a few paddings / margins / font sizes, nothing too fancy we can't do ourselves 😃

ang-zeyu:
one more minor suggestion: should we scroll the scrollbar horizontally toward the "current" element if its hidden beyond the scrollbar?

ang-zeyu:
> > Could we preserve the look of the dropdown? (absolutely-positioned with respect to the "Dropdown" text so it dosen't add to the height of the scrollbar) (including the searchbar, which uses some common css classes iirc)
> 
> Yup agree that it will look better. I've updated the design so that it is now absolute positioned and also added scrolling if the dropdown overflows the bottom of the screen.

Missed the search dropdown

packages/vue-components/src/Navbar.vue
jonahtanjz: Alright I've changed back to the previous implementation. Initially when I flushed all nav items to the left https://github.com/MarkBind/markbind/pull/1690#discussion_r771835529, it left some spaces at the end on some screen sizes so I decided to apply the colour to the entire background instead.

image

I tried playing around with the flex property and decided to go with `flex-grow: 1` so that the nav items fill the bar completely and also ensure that the spacing between each item is "equal". I think this allows for setting the background of each individual nav link as well as addresses the spacing issue.

image

jonahtanjz:
> one more minor suggestion: should we scroll the scrollbar horizontally toward the "current" element if its hidden beyond the scrollbar?
> 
> It might be somewhat distracting ux wise if it runs too slowly though, you could also try hoisting the scrolling logic into the `<template>` as an inline `<script>` tag so it runs much sooner before the vue app initialisation, if needed.

I tried inserting the scrolling script right after the highlight link function and it seems fine on the preview deployment. The scroll should happen right after the link highlighting with minimal delay.

> Missed the search dropdown

Thanks for the catch! I've also increased the min-width of the search bar on mobile so that it has sufficient space to display the content. Earlier some contents of the search were too compressed.

ang-zeyu (approved):
Lgtm! 👍

PR #1694, opened 2021-12-18
Enhance appearance of minimised panels

kaixin-hc:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Referring to PR #1422 and the associated comments, I wanted to try putting them on the same line as well as using a different icon / sizing / margin. Provides a slightly different suggestion for #1201 .

**Overview of changes:**
**Before**
before

**hhdqirui suggested change**
image

**Suggested**
suggested

* Add caret icon (as most minimised panels tend to use a caret icon to show that it is a clickable button

* Add top margin (This is to prevent ugly cluttering. as expanded cards also have a top margin but not a bottom margin, top margin was chosen. _I do wonder what the reason is for having a top margin and not a bottom margin though_! )

* Could be expanded by allowing the user to turn off the icon the same way they can turn off the buttons for the panels
* I think this may not pass tests - I found a diff in `markbind/fontawesome/css/all.min.css`. Is it possible just to run a command to update specific test files or only the ones which have changed...? **ETA: oh, it passed! Huh. I wonder why npm run test stopped then 🤔 **

(Possibly to break up into other PRs?)
* I think for code maintenance, we could probably remove some of the styles in NestedPanel.vue . If I remember correctly, we only need to specify the styles to change in the responsive breakpoints, so is it really needed to specify the same style twice?

* https://developer.mozilla.org/en-US/docs/Web/CSS/float suggests that it might be unneeded to specify display: inline-block if we are also specifying float, as it would be overridden and ignored.

**Testing instructions:**

Enhance appearance of minimised panels
* Add caret icon (as most minimised panels tend to use a caret icon to
show that it is a clickable button

* Add top margin (as expanded cards also have a top margin but not a
bottom margin, top margin was chosen. This is to prevent ugly cluttering)

---

- [X] No unrelated changes


damithc:
@kaixin-hc if this is a visual change, please post before-and-after screenshots too

kaixin-hc:
> @kaixin-hc if this is a visual change, please post before-and-after screenshots too

Okay! I've edited the first message so that it is more visible.

jonahtanjz (commented):
Thanks @kaixin-hc for working on this :) I think putting them on the same line works as it maintains the minimalistic form and at the same time inform the user that the button can be expanded 👍 I am fine with the current design. Prof @damithc @ang-zeyu any thoughts on this?

Just a quick review and left some comments.

I think on mobile if the title of the panel is long, the icon and the text are on a different lines which increases the height of the button. 

image

I think it's better to force the icon to be on the same line and then wrap the title if it overflows. 

Something like this:
image

damithc:
>I think putting them on the same line works as it maintains the minimalistic form and at the same time inform the user that the button can be expanded 👍 I am fine with the current design. Prof @damithc @ang-zeyu any thoughts on this?

I'm OK with this design 👍

ang-zeyu:
ok with the design as well! A caret icon seems like a good balance.

> Could be expanded by allowing the user to turn off the icon the same way they can turn off the buttons for the panels

Agreed, let's add this in, something like `no-minimized-switch` (can be less verbose)

Another (subjective) suggestion:

round the buttons into pills to make it a little sleeker, so it blends in better with content .
   panelspillrounded

damithc:
> round the buttons into pills to make it a little sleeker, so it blends in better with content .
> panelspillrounded

Yup, this blends in better with the content. The downside is that a minimized panel 'changes shape' when expanded as panels are actually square-shaped, causing a bit of a disconnect between the minimal and the other forms. So, I'm on the fence about this one.

ang-zeyu:
> Yup, this blends in better with the content. The downside is that a minimized panel 'changes shape' when expanded as panels are actually square-shaped, causing a bit of a disconnect between the minimal and the other forms. So, I'm on the fence about this one.

Preview:

panelexpandrounded

It does seem a little disconnected, but I'm personally leaning more toward rounding the buttons for a sleeker look with the rest of the page.

I think its something we can eventually "smooth out" to some degree by adding some transitions between the minimised and expanded form as well. (or we could round the existing panels a tad bit more)

kaixin-hc:
Thanks for the review and comments everyone. The preview makes it really easy to visualise. I do think that animations will help to smooth it out, and while I don't know how to apply animations yet it might be nice even if they are not rounded. I'll look into it in the new year!

I'm not sure about fully rounding the corners though, I think it might depend on if users tend to put many minimised expandable panels near each other. I found this article https://uxdesign.cc/make-sense-of-rounded-corners-on-buttons-dfc8e13ea7f7 which made a case for fully rounded buttons for primary content when you have space to spare, to direct users attention to those buttons. They suggested to avoid fully rounded buttons when many are used next to each other as it may not be obvious which to click.

Since I assume the minimised form of the expandable panels would be used more when there are many expandable panels next to each other, I wonder if the eye might become more drawn to the round buttons.

ang-zeyu:
Thanks for sharing the article; That was a good read.

>  I think it might depend on if users tend to put many minimised expandable panels near each other

I agree fully rounded buttons wouldn't be too suitable here, it might be confused for multiple selection type interfaces.

Not sure if this is a common use case as well @damithc 


We could also experiment with slightly milder settings and see which works best (`20px` -> `10px`), I would also be fine if we delayed this decision to some later time completely =P

> I do think that animations will help to smooth it out, and while I don't know how to apply animations yet it might be nice even if they are not rounded. I'll look into it in the new year!

👍

can be done separately too, and might even be a little out of scope here as well

---

Some more screenshots for comparison:

semi rounded `10px` (fully rounded is `20px`)
seamless3

fully rounded (original [here](https://nus-cs2103-ay2122s2.github.io/website/admin/ip-w2.html)):

ok

dosentflow

damithc:
> I agree fully rounded buttons wouldn't be too suitable here, it might be confused for multiple selection type interfaces.
> 
> Not sure if this is a common use case as well @damithc

In the CS2103 website (and it's variants), we do have such cases. So, I'm leaning slightly away from fully-rounded corners in this case.

ang-zeyu:
@kaixin-hc 

> Agreed, let's add this in, something like no-minimized-switch (can be less verbose)

just in case you missed this ^

As for the corner rounding, you could try a couple of settings and post back here; Also ok if it turns out the best setting is the current / we leave it be for now.

kaixin-hc:
I tried a couple of settings for corner rounding; it looked similar to what you posted previously. However, I still feel looking at it that the original settings are best for visual consistency with the rest of MarkBind as it matches the rounding of other components. The inconsistency makes it look a little jarring to me (and again, a bit like tags!)

corners inconsistency

The current appearance is as follows:
panels
minimised no caret

docs/userGuide/syntax/panels.mbdf
```diff
@@ -104,11 +104,14 @@
 </include>
 
 
-**Show/Hide buttons using `no-switch` or `no-close`.**
+**Show/Hide buttons using `no-switch`, `no-close`, or `no-mimized-switch`.**
```
jonahtanjz, line 107:
```suggestion
**Show/Hide buttons using `no-switch`, `no-close`, or `no-minimized-switch`.**
```

Typo here :)

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -236,10 +248,19 @@ export default {
...
     .caret-wrapper {
         float: left;
-        display: inline-block;
         width: 32px;
     }
 
+    .minimal-caret-wrapper {
+        display: inline-block;
+        font-size: 13px;
+        margin-right: 2px;
```
jonahtanjz, line 257: Might want to slightly increase the margin between the minimized switch and the card title on mobile? Right now it looks a little too cluttered.

image

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -226,8 +236,10 @@ export default {
     .card-title {
         display: inline-block;
         font-size: 1em;
+        line-height: 1.2em;
```
jonahtanjz, line 239: Just an opinion: 

I think the card titles look slightly more centered vertically without setting the `line-height` here.

With `line-height`:
image


Without `line-height`:
image

packages/vue-components/src/panels/NestedPanel.vue
```diff
@@ -236,10 +248,19 @@ export default {
...
 
+    .minimal-caret-wrapper {
+        display: inline-block;
+        font-size: 13px;
+        margin-right: 2px;
+    }
+
+    .glyphicon {
+        display: inline-block;
+    }
```
jonahtanjz, line 262: I believe the provided css from bootstrap-glyphicons already set the display to inline-block so should be able to remove this.

jonahtanjz (commented):
Nice work @kaixin-hc 👍 A couple of suggestions on the implementation.

> I tried a couple of settings for corner rounding; it looked similar to what you posted previously. However, I still feel looking at it that the original settings are best for visual consistency with the rest of MarkBind as it matches the rounding of other components. The inconsistency makes it look a little jarring to me (and again, a bit like tags!)

We can leave the corner rounding as it is for now. We can come back to this again or as a separate issue, round the corners of the other components as well if it makes the entire site "sleeker".

packages/vue-components/src/panels/NestedPanel.vue
kaixin-hc: I see what you mean, but it helps for panels with long names.

With line-height:
With line-height picture

Without line-height:
Without line-height picture

kaixin-hc:
Thanks for the review @jonahtanjz, sorry for the delay in implementing changes. I've done as you suggested for all except line height (screenshot explanation in the comments)

packages/vue-components/src/panels/NestedPanel.vue
jonahtanjz: Ahh alright then can keep it as it is for now :)

jonahtanjz (approved):
Thanks for making the changes @kaixin-hc!

LGTM 👍

Issue #1616, opened 2021-06-29
[cs2103 website] Some top-nav menu items are not highlighted when visited


damithc:
Similar to #1615 , not urgent/critical but help me troubleshoot this problem too. Site/code location and steps to set up locally are similar too.

When the user visits the page in this menu item, the top-nav menu item is highlighted in blue (as shown, and as expected) **after the target page is loaded** (I'm not referring to the hover-over effects).
![image](https://user-images.githubusercontent.com/1673303/123755443-7adc3500-d8ee-11eb-8eb0-8a282e63a7ee.png)

But visiting this menu item doesn't have the same behavior i.e., the item does not get highlighted in blue after the target page is loaded.
![image](https://user-images.githubusercontent.com/1673303/123755986-0ce43d80-d8ef-11eb-8a81-4133347c23ba.png)

None of these menu items get highlighted either.
![image](https://user-images.githubusercontent.com/1673303/123755968-0786f300-d8ef-11eb-9b11-8d9c96c8ac72.png)

PR #1700, opened 2021-12-28
Fix navbar highlighting

ang-zeyu:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

** Likely Unrelated to #1616 

~Should be rebased in without squash~

**Overview of changes:**

Found these while working on #1698 
- Fix `hl-mode='child'` check -- missing a `- 1`
- ~Fix normalizeUrl appending extensionless urls not ending with `/` with a `.html`~
   - ~It should be `xx/yy` -> `xx/yy/index.html` instead~

**Testing instructions:**

~per the commits~

Fix navbar `isChild` highlighting check

---

- [x] Added tests for bug fixes or features - unit tests added for the second commit. e2e tests are needed for the first, which we currently don't have
- [x] No unrelated changes

### DIFF
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -36,6 +36,15 @@ import normalizeUrl from './utils/urls';
 import SiteNavButton from './SiteNavButton.vue';
 import PageNavButton from './PageNavButton.vue';
 
+function isEqualExceptLast(hParts, uParts) {
+  for (let i = 0; i < hParts.length - 1; i += 1) {
+    if (hParts[i] !== uParts[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
 export default {
   components: {
     SiteNavButton,
@@ -112,12 +121,7 @@ export default {
       if (hParts.length !== uParts.length) {
         return false;
       }
-      for (let i = 0; i < hParts.length - 1; i += 1) {
-        if (hParts[i] !== uParts[i]) {
-          return false;
-        }
-      }
-      return true;
+      return isEqualExceptLast(hParts, uParts);
     },
     isChild(url, href) {
       const hParts = this.splitUrl(href);
@@ -125,12 +129,7 @@ export default {
       if (uParts.length <= hParts.length) {
         return false;
       }
-      for (let i = 0; i < hParts.length - 1; i += 1) {
-        if (hParts[i] !== uParts[i]) {
-          return false;
-        }
-      }
-      return true;
+      return isEqualExceptLast(hParts, uParts);
     },
     isExact(url, href) {
       return normalizeUrl(url) === normalizeUrl(href);
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -125,7 +125,7 @@ export default {
       if (uParts.length <= hParts.length) {
         return false;
       }
-      for (let i = 0; i < hParts.length; i += 1) {
+      for (let i = 0; i < hParts.length - 1; i += 1) {
         if (hParts[i] !== uParts[i]) {
           return false;
         }
--- a/packages/vue-components/src/Navbar.vue
+++ b/packages/vue-components/src/Navbar.vue
@@ -36,15 +36,6 @@ import normalizeUrl from './utils/urls';
 import SiteNavButton from './SiteNavButton.vue';
 import PageNavButton from './PageNavButton.vue';
 
-function isEqualExceptLast(hParts, uParts) {
-  for (let i = 0; i < hParts.length - 1; i += 1) {
-    if (hParts[i] !== uParts[i]) {
-      return false;
-    }
-  }
-  return true;
-}
-
 export default {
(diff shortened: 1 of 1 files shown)


packages/vue-components/src/utils/urls.js
```diff
@@ -1,10 +1,9 @@
...
-  } else if (!url.endsWith('/')) {
-    return `${url}.html`.toLowerCase();
+  const urlLower = url.toLowerCase();
+  if (urlLower.endsWith('.html')) {
+    return urlLower;
+  } else if (urlLower.endsWith('/')) {
+    return `${urlLower}index.html`;
   }
-  return url.toLowerCase();
+  return `${urlLower}/index.html`;
```
jonahtanjz, line 8: Can I check why extensionless urls need to be appended with `/index.html`? I believe this breaks the `isExact` method because on our netlify deployment, the `.html` is omitted from the url on some of the pages.

For example this page, https://deploy-preview-1700--markbind-master.netlify.app/userguide/addingpages, the site nav corresponding url is `/userGuide/addingPages.html` and after normalizing, the href will remain as it is but the url will be appended to `/userguide/addingpages/index.html`.

Capture

ang-zeyu: Thanks for highlighting this, completely missed it.

Was thinking solely of:
`https://markbind.org/userGuide` -> `https://markbind.org/userGuide/index.html`

But its also certainly possible for it to be interpreted as `https://markbind.org/userGuide.html`. (this is not the case for `markbind s`'s local server however (404))

I think I'll discard this commit / change for now, and just stick with the first one in this PR since the logic for this case is quite a bit more complex than initially thought.

We may need to consider such `href`s (without extension and without ending `/`) twice, interpreted once as `userGuide.html` and another time as `userGuide/index.html` in some way. The order of consideration of these links should also be thought through as well.

If there's no solution, the status quo might be better too, since `https://markbind.org/userGuide` -> `https://markbind.org/userGuide.html` is more *probably* likely than `https://markbind.org/userGuide` -> `https://markbind.org/userGuide/index.html`.

packages/vue-components/src/Navbar.vue
```diff
@@ -125,7 +125,7 @@ export default {
       if (uParts.length <= hParts.length) {
         return false;
       }
-      for (let i = 0; i < hParts.length; i += 1) {
+      for (let i = 0; i < hParts.length - 1; i += 1) {
```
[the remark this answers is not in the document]
ang-zeyu: The sibling one is already functioning correctly if I'm not mistaken;

E.g. (since after `normalizeUrl` all urls end with `.html`)
`xx/yy/foo.html`
`xx/yy/bar.html`
we should check both `xx` and `yy`

In contrast the child check isn't without this change (try #1698's new subpages for an example):
E.g.
`xx/yy/foo.html`
`xx/yy/zz/bar.html`
we should check both `xx` and `yy`, but not `foo.html`

I added a new commit 5b93fd3 as well to remove the small code duplication between these 2 functions since the checks are identical.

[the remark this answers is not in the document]
jonahtanjz: Ahh ok I was thinking along the lines of having `/index.html` appended to all normalized urls earlier.

But I believe there is still a case of having:
`xx/yy/foo/index.html`
`xx/yy/bar/index.html`
we would need to stop at `-2` for this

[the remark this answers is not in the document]
ang-zeyu: > we would need to stop at -2 for this

👀 
Is this for `isSibling`?

Shouldn't it be up till `-1`, since in this case the two links are not siblings because `foo` !== `bar`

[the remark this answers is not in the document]
jonahtanjz: Oh right XD Ignore the previous comment 😅😅

packages/vue-components/src/Navbar.vue
```diff
@@ -36,6 +36,15 @@ import normalizeUrl from './utils/urls';
 import SiteNavButton from './SiteNavButton.vue';
 import PageNavButton from './PageNavButton.vue';
 
+function isEqualExceptLast(hParts, uParts) {
```
jonahtanjz, line 39: Just a minor suggestion: would it be better to place this in `methods`?

ang-zeyu: 👍 shifted back in for consistency

jonahtanjz (commented):
Thanks!

LGTM 👍

jonahtanjz (approved):

packages/vue-components/src/utils/urls.js
jonahtanjz: Yup we can further explore on this. 

> If there's no solution, the status quo might be better too, since https://markbind.org/userGuide -> https://markbind.org/userGuide.html is more probably likely than https://markbind.org/userGuide -> https://markbind.org/userGuide/index.html.

I'm fine with this as well

PR #1724, opened 2022-01-18
Escape the raw variable interpolation syntax in PageNav from Vue

### LINKED ISSUE #1720: Page Navigation Menus not working with titles that contain `{{ content }}`
Windows
* **MarkBind Version:**
v.3.1.0
While working on PR #1719, I decided to turn on the page nav menus for the tips and tricks page and it resulted in the following screenshot:
image

Snippet of the `tipsAndTricks.md` page below:
```md
{% set title = "Tips & Tricks" %}
<span id="title" class="d-none">{{ title }}</span>

<frontmatter>
  title: "User Guide: {{ title | safe }}"
  layout: userGuide.md
  pageNav: 5
</frontmatter>

# {{ title | safe }}

<span id="escapingCharacters">

##### :fas-lightbulb: Escaping Characters

For Markdown syntax: To display a literal character that are normally used for Markdown formatting, add a backslash (`\`) in front of the character.

<variable name="code">
\* item 1

* item 1 
</variable>
</span>

<small>More info: [https://www.markdownguide.org/basic-syntax#escaping-characters](https://www.markdownguide.org/basic-syntax#escaping-characters)</small>

</span>

{% raw %}

##### :fas-lightbulb: Using {% raw %}{% endraw %} to display `{{ content }}`


To display the raw variable interpolation syntax using `{% raw %}{% endraw %}`, you would also need to add
the `v-pre` attribute using markdown-it-attrs or as a html attribute.

```

I expected the page side nav to show the formatted text of ":fas-lightbulb: Using {% raw %}{% endraw %} to display `{{ content }}`",

The last part "`{{ content }}`" disappeared. On the console, it is being rendered as `&lrm;` :
image
In the console, there is an error message:
image

The bug example can be viewed at this deploy preview [page](https://deploy-preview-1719--markbind-master.netlify.app/userguide/tipsandtricks)

tlylt:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1720 
Expected result after this PR is merged:
image

Two things:
- The problem in the related issue arose because even though the code element has been properly handled by adding `v-pre` as an attribute, the PageNav HTML is constructed by taking only the text content in `core/src/Page/index.js/_collectNavigableHeadings`:
```js
		// other parts omitted
        this.navigableHeadings[$(elem).attr('id')] = {
          text: $(elem).text(),
          level: elem.name.replace('h', ''),
        };
```
An alternative would be to replace `$(elem).text()` with `$(elem).html()`, which will allow for rendering of markdown in PageNav.
image
While this works fine, it seems like it will cause an issue for headings that have include fragments. One example is:
image
vs originally:
image
I think it's worth discussing whether we want to improve the display of PageNav. Perhaps in the meantime can go with the existing approach to fix the bug.
- In terms of the code changes in this PR, I think there are a few other places where we can introduce the `v-pre` attribute. I am not sure if the change that I introduced will impact performance since it uses Regex. My guess is that it should be fine as it only runs for page nav headings which are typically short.

**Testing instructions:**
Test case added under `test_site`:
asdsd

Fix PageNav raw variable interpolation syntax display

The raw variable interpolation syntax can be displayed
by wrapping it  with {% raw %} {% endraw %}. Even though
the {{ content }} is showing up correctly in the main page,
the generated PageNav text retains {{}} and causes vue
to incorrectly hide it due to rendering error.

Let's add a v-pre attribute whenever there is {{}} in
the navigation text.

This will allow content wrapped {{}} to display in PageNav.

---

- [x] No unrelated changes


jonahtanjz (approved):
Nice work @tlylt. Implementation looks good 👍 LGTM :)

> An alternative would be to replace $(elem).text() with $(elem).html(), which will allow for rendering of markdown in PageNav. 
>
> While this works fine, it seems like it will cause an issue for headings that have include fragments.

I believe using `.text()` instead of `.html()` will help to keep the pageNav as consistent as possible. Some possible issues that I can think of when using `.html()` are when the headers contain hyperlinks or triggers.

For instance,
```.html
# [Some headers](https://google.com)
```
will probably pull the anchor tag to the pageNav as well and clicking on it will redirect to the hyperlink instead of the header's position.

Similarly,
```.html
# Some <trigger for="modal:loremipsum">trigger</trigger>
```
will most likely also pull the trigger and cause the pageNav link to open up a modal as well.

This can become really complicated when users decide to introduce their own classes and styling to the headers.

If the user wants certain parts of the header to have a larger font size, they could do something like this
```.html
# Test <span class="big-font-size-class">Big Word</span> 
```
the resulting pageNav will most likely end up with inconsistent font sizes.

We will have to consider the cost of processing all these possible cases, the benefits of having an improved pageNav and whether it's worth to implement and maintain it.

> I think it's worth discussing whether we want to improve the display of PageNav. Perhaps in the meantime can go with the existing approach to fix the bug.

Personally I feel that a simple pageNav will suffice as the goal of the pageNav is to help users navigate to the main content so I'm leaning towards the current implementation. I am open to discussion on this. You can open an issue for further discussion so others can chip in as well :)

> I am not sure if the change that I introduced will impact performance since it uses Regex. My guess is that it should be fine as it only runs for page nav headings which are typically short.

Shouldn't have any significant performance issue as each page nav link should be quite short

packages/core/src/Page/index.js
```diff
@@ -354,10 +354,15 @@ class Page {
   generatePageNavHeadingHtml() {
     let headingHTML = '';
     const headingStack = [];
+    const interpolationRegex = /{{.*}}/;
     Object.keys(this.navigableHeadings).forEach((key) => {
       const currentHeadingLevel = this.navigableHeadings[key].level;
-      const currentHeadingHTML = `<a class="nav-link py-1" href="#${key}">`
-        + `${this.navigableHeadings[key].text}&#x200E;</a>\n`;
+      // Add v-pre to prevent text interpolation for {{ }} wrapped in {% (end)raw %}
```
ong6, line 360: Is this a typo on the `(end)raw` or is this intended?

ong6 (approved):
LGTM!

packages/core/src/Page/index.js
tlylt: I used it to mean `{% raw %} {% endraw %}` 😂

tlylt:
> I believe using `.text()` instead of `.html()` will help to keep the pageNav as consistent as possible. Some possible issues that I can think of when using `.html()` are when the headers contain hyperlinks or triggers.
> 
> For instance,
> 
> ```
> # [Some headers](https://google.com)
> ```
> 
> will probably pull the anchor tag to the pageNav as well and clicking on it will redirect to the hyperlink instead of the header's position.
> 
> Similarly,
> 
> ```
> # Some <trigger for="modal:loremipsum">trigger</trigger>
> ```
> 
> will most likely also pull the trigger and cause the pageNav link to open up a modal as well.
> 
> This can become really complicated when users decide to introduce their own classes and styling to the headers.
> 
> If the user wants certain parts of the header to have a larger font size, they could do something like this
> 
> ```
> # Test <span class="big-font-size-class">Big Word</span> 
> ```
> 
> the resulting pageNav will most likely end up with inconsistent font sizes.
> 
> We will have to consider the cost of processing all these possible cases, the benefits of having an improved pageNav and whether it's worth to implement and maintain it.
> 
> > I think it's worth discussing whether we want to improve the display of PageNav. Perhaps in the meantime can go with the existing approach to fix the bug.
> 
> Personally I feel that a simple pageNav will suffice as the goal of the pageNav is to help users navigate to the main content so I'm leaning towards the current implementation. I am open to discussion on this. You can open an issue for further discussion so others can chip in as well :)

I see. I think the few cases that you mentioned are indeed problematic. The reason why I thought of improving the pageNav display was mainly to be somewhat consistent with the main navigation on the left that supports icons and formatting. I agree that we can keep it simple until there's a strong need/use case for making pageNav more complex than necessary.

ang-zeyu:
Sorry for the late input!
For performance and simplicity, would it be possible to add `v-pre` on the entire pageNav *always*? (since the pagenav contains only text)

tlylt:
> Sorry for the late input! For performance and simplicity, would it be possible to add `v-pre` on the entire pageNav _always_? (since the pagenav contains only text)

Ah ok, I think another reason why I only added v-pre when necessary was so that there are no updates to expected files for existing functional tests.

will making a new pr on this be ok @ang-zeyu ?

ang-zeyu:
Go ahead! @tlylt 👍 

Don't worry about expected test file updates, that's what pr reviews are for. Implementation simplicity should be the priority here.

PR #1730, opened 2022-01-20
Update algolia plugin

### LINKED ISSUE #1729: Update Algolia plugin
I'm having trouble getting the Algolia plugin to work with the latest search indexes provided by them (note that they migrated to a new infra recently).

Steps: 
1. Go to https://codesandbox.io/s/currying-leaf-qwo9k?file=/src/index.js  (this is a sandbox provided by Algolia devs to us)
2. Change 2021 to 2022 as indicated below
    image
3. Use the right-hand side to confirm the search is working

Now, I have trouble using the same config parameters with our algolia plugin. This is how my config looks like:
```json
    "algolia": {
      "apiKey": "a3fef7e2d9dea07c0fa2d8ed521057aa",
      "appId": "3DU4XTBX4T",
      "indexName": "nus-tic-cs",
      "algoliaOptions": { "hitsPerPage": 10 },
      "facetFilters": ["site:nus-tic2002-2022.github.io"],
      "debug": false
    }
```
Perhaps some updates needed to our Algolia plugin?

jonahtanjz:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [x] Others, please explain: Update Algolia plugin from v2 to v3

Fixes #1729 

**Overview of changes:**
Update Algolia from v2 to v3. Algolia now requires `appId` to be provided in order to use its APIs. V3 also requires that the search box be a `div` and no longer supports `input`.

User facing changes include adding new options and removing unsupported options:

**Added:** 
`appId` - compulsory field
`searchParameters` - optional

**Removed:**
`debug`
`algoliaOptions`

V3 has changed its design drastically from V2. The search bar has a new design and clicking on the search bar will open a modal instead. The searches will be done on this modal and the results will be displayed on the modal instead of the dropdown that is used by Markbind's search. Since the new search bar and modal are auto generated by Algolia's package, it will be complicated to try to make the design similar to the current Markbind's search. Is it fine to stick to their native design? 

New search bar: 
image

New search modal instead of dropdown:
image

**Testing instructions:**
Add the following to `site.json` and serve the site.
```json
"plugins": [
    "algolia"
  ],
  "pluginsContext": {
    "algolia": {
      "apiKey": "599cec31baffa4868cae4e79f180729b",
      "appId": "R2IYF7ETH7",
      "indexName": "docsearch"
    }
  }
```

Update Algolia plugin

Algolia docsearch v2 is now a legacy version. Algolia also
requires new parameters for their APIs.

Let's update the version of Algolia docsearch to v3 and change
the plugin options accordingly.

---

- [ ] No unrelated changes

### DIFF
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -50,6 +50,7 @@ function detectAndApplyFixedHeaderStyles() {
     }`,
   );
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
+  insertCss(`.DocSearch-Modal {margin-top: ${headerHeight}px }`);
 
   const adjustHeaderClasses = () => {
     const newHeaderHeight = headerSelector.height();
@@ -69,6 +70,9 @@ function detectAndApplyFixedHeaderStyles() {
             break;
           case '.nav-menu-open':
             rules[0].style.maxHeight = `calc(100% - ${newHeaderHeight}px + 50px)`;
+            break;
+          case '.DocSearch-Modal':
+            rules[0].style.marginTop = `${newHeaderHeight}px`;
             break;
           default:
             break;
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -42,15 +42,15 @@ function detectAndApplyFixedHeaderStyles() {
 
   const headerHeight = headerSelector.height();
   const bufferHeight = 1;
-  insertCss(`.fixed-header-padding { padding-top: ${headerHeight}px !important }`);
+  insertCss(`.fixed-header-padding { padding-top: ${headerHeight}px !important; }`);
   insertCss(
     `span.anchor {
     position: relative;
-    top: calc(-${headerHeight}px - ${bufferHeight}rem)
+    top: calc(-${headerHeight}px - ${bufferHeight}rem);
     }`,
   );
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
-  insertCss(`.DocSearch-Modal {margin-top: ${headerHeight}px }`);
+  insertCss(`.DocSearch-Modal { margin-top: ${headerHeight}px; }`);
 
   const adjustHeaderClasses = () => {
     const newHeaderHeight = headerSelector.height();
@@ -63,7 +63,7 @@ function detectAndApplyFixedHeaderStyles() {
           switch (rules[0].selectorText) {
           case '.fixed-header-padding':
             sheets[i].deleteRule(0);
-            sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important }`);
+            sheets[i].insertRule(`.fixed-header-padding { padding-top: ${newHeaderHeight}px !important; }`);
             break;
           case 'span.anchor':
             rules[0].style.top = `calc(-${newHeaderHeight}px - ${bufferHeight}rem)`;
--- a/packages/core-web/src/index.js
+++ b/packages/core-web/src/index.js
@@ -55,22 +55,9 @@ function detectAndApplyFixedHeaderStyles() {
     const newHeaderHeight = headerSelector.height();
     const sheets = document.styleSheets;
 
-    const handleAlgoliaPluginCss = (sheet) => {
-      sheet.cssRules[0].style.marginTop = `${newHeaderHeight}px`;
-      sheet.deleteRule(1);
-      sheet.insertRule(`
-        @media (max-width: 750px) {
-          .DocSearch-Dropdown {
-            height: calc(100% - ${newHeaderHeight}px - 66px);
-          }
-        }`, 1,
-      );
-    };
-
     for (let i = 0; i < sheets.length; i += 1) {
       try {
         const rules = sheets[i].cssRules;
-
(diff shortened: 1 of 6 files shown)


damithc:
Thanks for tackling this @jonahtanjz 

> V3 has changed its design drastically from V2. The search bar has a new design and clicking on the search bar will open a modal instead. The searches will be done on this modal and the results will be displayed on the modal instead of the dropdown that is used by Markbind's search. Since the new search bar and modal are auto generated by Algolia's package, it will be complicated to try to make the design similar to the current Markbind's search. Is it fine to stick to their native design?

I'm OK with it, if they haven't provided an easy way to use the dropdown for search results.

ryoarmanda (approved):
LGTM! 🎉

packages/core-web/src/index.js
```diff
@@ -42,34 +42,61 @@ function detectAndApplyFixedHeaderStyles() {
...
   );
   insertCss(`.nav-menu-open { max-height: calc(100% - ${headerHeight}px); }`);
+  insertCss(`.DocSearch-Modal { margin-top: ${headerHeight}px; }`);
+  insertCss(
+    `@media (max-width: 750px) {
+      .DocSearch-Dropdown {
+        height: calc(100% - ${headerHeight}px - 66px);
+      }
+    }`,
+  );
```
ang-zeyu, line 60: Just a quick look, would it be better to keep these strictly to the plugin itself? Since docsearch isn't really core functionality, but a plugin.

Some questions release wise:
- whether its possible to deliver a more minimal fix by adding the `apiKey` you mentioned onto `v3.1.1` or so (also so no immediate action needed on author end to migrate v2 -> v3)
- while we merge this bigger change into the main branch for `v4.0` (since v2 -> v3 docsearch is a breaking change)

damithc:
Perhaps we can do a minor release after this is merged? I would like to start using this in the module website ASAP.

packages/core-web/src/index.js
jonahtanjz: > Just a quick look, would it be better to keep these strictly to the plugin itself? Since docsearch isn't really core functionality, but a plugin.

I've shifted the plugin css code to within the plugin file. The rest of the plugin code related to `adjustHeaderClasses` is left there to minimise duplication as there are a couple of conditions that will need to adjust the plugin's css.

> * whether its possible to deliver a more minimal fix by adding the apiKey you mentioned onto v3.1.1 or so (also so no immediate action needed on author end to migrate v2 -> v3)
>
> * while we merge this bigger change into the main branch for v4.0 (since v2 -> v3 docsearch is a breaking change)

Yup agree with this. We should do the v2 -> v3 docsearch in our v4.0 and release the minimal fix in v3.1.1. I'll create a separate branch called `v3.1.x-hotfixes` (similar to the v2.7.x-hotfixes branch) and push the update there. Is this workflow fine?

ang-zeyu: > Yup agree with this. We should do the v2 -> v3 docsearch in our v4.0 and release the minimal fix in v3.1.1. I'll create a separate branch called v3.1.x-hotfixes (similar to the v2.7.x-hotfixes branch) and push the update there. Is this workflow fine?

👍 

Alternatively there haven't been any breaking changes since the last release so far, so we could also merge the more minimal fix directly on the main branch. (I think either is fine)

ang-zeyu: > I've shifted the plugin css code to within the plugin file. The rest of the plugin code related to adjustHeaderClasses is left there to minimise duplication as there are a couple of conditions that will need to adjust the plugin's css.


> V3 has changed its design drastically from V2. The search bar has a new design and clicking on the search bar will open a modal instead. The searches will be done on this modal and the results will be displayed on the modal instead of the dropdown that is used by Markbind's search. Since the new search bar and modal are auto generated by Algolia's package, it will be complicated to try to make the design similar to the current Markbind's search. Is it fine to stick to their native design?


Sorry, I completely missed the purpose of these changes earlier.

Are the dynamic styles added to tackle the above mentioned issue?

I'm favouring sticking with algolia's default design as much as possible if so, imo there isn't much benefit from pushing the dropdown / ui downward. Coupling our dynamic style detections over to an external element might break as well if algolia's styles / classes / dom structure changes.

If its an issue with correctness (e.g. the modal going out of the screen) I'll need a little more time to review this with a v3 algolia setup

jonahtanjz: Mainly because the modal gets blocked by the navbar if the header is fixed.

image

I could try to increase the `z-index` of the modal to be higher. That may help to resolve this issue without pushing it down below the fixed header.

jonahtanjz: Tested with z-index and it works as well. Will change to this implementation as it's much simpler.

image

packages/core-web/src/index.js
```diff
@@ -42,18 +42,19 @@ function detectAndApplyFixedHeaderStyles() {
 
   const headerHeight = headerSelector.height();
   const bufferHeight = 1;
-  insertCss(`.fixed-header-padding { padding-top: ${headerHeight}px !important }`);
+  insertCss(`.fixed-header-padding { padding-top: ${headerHeight}px !important; }`);
```
ang-zeyu, line 45: Nice work @jonahtanjz, but do try to avoid the unnecessary changes here!

PR #1734, opened 2022-01-25
Add global line number option & hide line numbers by default for code blocks

### LINKED ISSUE #1671: Make .no-line-numbers the default
My observation is that line numbers are really not useful in 90% of cases and add unnecessary distraction to the content.

Suggestions:
- make `.no-line-numbers` the default
- introduce `line-numbers` option
- show line-numbers when using `markbind serve` for development, allow forcefully hiding it with `.no-line-numbers` nevertheless
- introduce a `style.line-numbers` global config key of sorts to ease the migration / suit user preferences

tlylt:

- [x] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #1671

**Overview of changes:**
- Make line numbers hidden by default
- Enable a global line number setting via `codeLineNumbers` in `site.json`
  - individual code blocks can still specify `line-numbers ` or `no-line-numbers` for custom behavior 
- Update test cases
- Update docs to mention the above adjustment
  - also removed unnecessary `{.no-line-numbers}`

Edit:
Based on the suggestions, code blocks that have `{start-from=x}` (x is the line number) will force line numbers to show even if the global setting is to hide it.

After discussion, line numbers seem to take up too much space and add additional noise when code blocks display code outputs or other not-so-suitable situations. If enabled by default, this may be inconvenient for authors to hide line numbers one by one.
Thus, 
- code line numbers are now hidden by default
- individual code blocks that do not have `line-numbers` or `no-line-numbers` class will inherit from the global setting (if not set, default is to hide all line numbers)
- if a line number class is set, the code block will behave accordingly

In terms of code, I processed this logic in `core/src/html/NodeProcessor.js` 's `processNode`, it should also be possible to process it in `postProcessNode`, I am not sure if there is any preference or requirement on this. At the moment the current way to append `line-numbers`/`no-line-numbers` seems functional.

**Testing instructions:**
- Functional test case `testCodeBlock.md` adjusted to test the behavior
- In user guide, the part about code block syntax is edited [here](https://deploy-preview-1734--markbind-master.netlify.app/userguide/formattingcontents#line-numbering), and
- the part about `site.json` is edited [here](https://deploy-preview-1734--markbind-master.netlify.app/userguide/sitejsonfile#style)

Adjust line numbers setting for code blocks

Line numbers are provided by default and there is no way
to override it globally.

Let's include a `codeLineNumbers` option in `site.json` to
enable a global configuration of line numbers. Line numbers
are also now hidden by default.

This is more convenient for authors as line numbers tend to
be hidden manually in practice. This is also consistent with other
markdown renderers.

---

- [x] No unrelated changes


packages/core/src/html/codeblockProcessor.js
```diff
@@ -129,6 +129,28 @@ function highlightCodeBlock(node) {
   });
 }
 
+/**
+ * Adjust the class attribute of code blocks according o the global line numbers setting.
```
ong6, line 133: Typo on `according to`

ong6 (commented):
Perhaps you could add the `"codeLineNumbers": true` to the site.json so that people referencing that can easily see this feature?

Also I noticed that in the docs start-from=6 is not working? 
image

Since its working on the main website I assume something you did might have caused start-from to stop being parsed correctly. For example, I think since the default is now to not show the line numbers, start-from also does not show them hence this issue. Perhaps adding a check for start from inside codeblockProcessor could fix this issue.

packages/core/src/html/codeblockProcessor.js
tlylt: Great catch! Fixed.

tlylt:
> Perhaps you could add the `"codeLineNumbers": true` to the site.json so that people referencing that can easily see this feature?

Added.

> Also I noticed that in the docs start-from=6 is not working? image
> 
> Since its working on the main website I assume something you did might have caused start-from to stop being parsed correctly. For example, I think since the default is now to not show the line numbers, start-from also does not show them hence this issue. Perhaps adding a check for start from inside codeblockProcessor could fix this issue.

Yup thanks for the thorough check, I indeed forgot about this case. At the moment I am adding `.line-numbers` to the affected code blocks (start-from and highlight-lines) in the user guide so that the line numbers do show up. 

For `start-from`, it definitely has to go with line-numbers for it to make sense, but I am not sure whether I should explicitly allow it to work when the default is to not show line numbers. This is because the "right" way is to override line-number settings individually -> `{.line-numbers start-from=6}`. The alternative is to allow it to work without specifying `.line-numbers` when the default is to hide line numbers.

Similar logic for highlight-lines, except that it might still be ok without displaying line numbers. In the user guide, I added {.line-numbers} so that it's easier to see the effect of the highlighting, but it is totally fine for authors to want to hide the line numbers as it does not affect it's functionality.

Would like a second opinion on this as I am ok with both.
1. `start-from` will force code block to show line-numbers (and what about highlight-lines?)
2. `start-from` requires explicit `.line-numbers` when the global config is to hide line numbers by default.

ong6:
> Would like a second opinion on this as I am ok with both.
> 
> 1. `start-from` will force code block to show line-numbers (and what about highlight-lines?)
> 2. `start-from` requires explicit `.line-numbers` when the global config is to hide line numbers by default.

I was thinking about this and think that the first option is better for current users of markbind. This is my reasoning: 
1. Existing users would have used start-from without the .line-number being set.
2. Ease of use, users won't have to remember to add a .line-number to every singe code block.

jonahtanjz (commented):
Nice work @tlylt 👍 

Just a suggestion on the implementation.

What do you think of creating 2 separate css files for the different line number settings and import them respectively instead of processing each of the code block? 

For example, if the `style.codeLineNumbers` is true in `site.json`, 
* The default code block css rules will add line numbers.
* If there is `.no-line-number` class, then line numbers will not show.
(The above is the same as how the css is implemeneted now)

If the `style.codeLineNumbers` is false or not specified in `site.json`, 
* The default code block css rules will not add line numbers.
* If there is `line-number` class, then line numbers will be added.
(The opposite of how it is implemented now)

These 2 css logics can be in 2 different css files (maybe `show-default-line-number.css` and `no-default-line-number.css`). Then in the `Site/index.js`, it will choose the respective css to import based on what is specified in the `site.json` (Can refer to the implementation of `style.codeTheme` [here](https://github.com/MarkBind/markbind/blob/5c254de8454d2341cad8ae3e03fe9bb23e58ba71/packages/core/src/Site/index.js#L97) and [here](https://github.com/MarkBind/markbind/blob/5c254de8454d2341cad8ae3e03fe9bb23e58ba71/packages/core/src/Site/index.js#L305).

One reason for doing this is so that the implementation is simpler and faster. Not sure if this would create too many assets? Would like to hear your opinion on this.

> Would like a second opinion on this as I am ok with both.
>
> 1. start-from will force code block to show line-numbers (and what about highlight-lines?)
> 1. start-from requires explicit .line-numbers when the global config is to hide line numbers by default.

I agree with @ong6 on this. Option 1 would be much better as it does not make sense to have `start-from` without line numbers. So having `start-from` would naturally imply that there should be line numbers. Also existing users will not need to do anything for their current  `start-from` to work.

tlylt:
> What do you think of creating 2 separate css files for the different line number settings and import them respectively instead of processing each of the code block?
> 
> For example, if the style.codeLineNumbers is true in site.json,
> 
> * The default code block css rules will add line numbers.
> * If there is .no-line-number class, then line numbers will not show.
>   (The above is the same as how the css is implemeneted now)
> 
> If the style.codeLineNumbers is false or not specified in site.json,
> 
> * The default code block css rules will not add line numbers.
> * If there is line-number class, then line numbers will be added.
>   (The opposite of how it is implemented now)
> 
> These 2 css logics can be in 2 different css files (maybe show-default-line-number.css and no-default-line-number.css). Then in the Site/index.js, it will choose the respective css to import based on what is specified in the site.json (Can refer to the implementation of style.codeTheme [here](https://github.com/MarkBind/markbind/blob/5c254de8454d2341cad8ae3e03fe9bb23e58ba71/packages/core/src/Site/index.js#L97) and [here](https://github.com/MarkBind/markbind/blob/5c254de8454d2341cad8ae3e03fe9bb23e58ba71/packages/core/src/Site/index.js#L305).
> 
> One reason for doing this is so that the implementation is simpler and faster. Not sure if this would create too many assets? Would like to hear your opinion on this.

Hmm, I didn't consider this approach initially as I was not aware of the code-theme implementation. I think implementation wise it does feel like having separate stylesheets might be faster as we don't have to check in the token processing step and add the style class(though this rough gauge of performance degradation is probably inaccurate and unwarranted, without the use of proper tools). I guess one minor con of doing it in separate stylesheets is that, unlike themes, code line numbers are probably only going to be two and hence may not enjoy the potential extensibility from the proposed approach. 

By the way, a small question on the CSS files: I see that the code theme stylesheets you mentioned are minimized, by any chance do you know what's the tool used for that? 

> > Would like a second opinion on this as I am ok with both.
> > 
> > 1. start-from will force code block to show line-numbers (and what about highlight-lines?)
> > 2. start-from requires explicit .line-numbers when the global config is to hide line numbers by default.
> 
> I agree with @ong6 on this. Option 1 would be much better as it does not make sense to have start-from without line numbers. So having start-from would naturally imply that there should be line numbers. Also existing users will not need to do anything for their current start-from to work.

Sure I will make the relevant changes for this part.

jonahtanjz:
> Hmm, I didn't consider this approach initially as I was not aware of the code-theme implementation. I think implementation wise it does feel like having separate stylesheets might be faster as we don't have to check in the token processing step and add the style class(though this rough gauge of performance degradation is probably inaccurate and unwarranted, without the use of proper tools). I guess one minor con of doing it in separate stylesheets is that, unlike themes, code line numbers are probably only going to be two and hence may not enjoy the potential extensibility from the proposed approach.

Alright in that case we can stick to the current implementation for now 👍 

> By the way, a small question on the CSS files: I see that the code theme stylesheets you mentioned are minimized, by any chance do you know what's the tool used for that?

I'm not sure which tool was used to minify those CSS files but I believe there are online tools available which you can upload/paste in the CSS code and it will generate a minimized version.

packages/core/src/Site/index.js
```diff
@@ -280,6 +280,9 @@ class Site {
     const sourcePath = path.join(this.rootPath, config.pageSrc);
     const resultPath = path.join(this.outputPath, fsUtil.setExtension(config.pageSrc, '.html'));
     const codeTheme = this.siteConfig.style.codeTheme || 'dark';
+    // hide line numbers by default
+    const codeLineNumbers = this.siteConfig.style.codeLineNumbers !== undefined
```
ang-zeyu, line 284: Could we centralise the initialisation of default values in `SiteConfig` directly?

packages/core/src/html/codeblockProcessor.js
```diff
@@ -129,6 +129,28 @@ function highlightCodeBlock(node) {
...
+ * Adjust the class attribute of code blocks according to the global line numbers setting.
+ * This looks into the class of each code block and if no `line-numbers` or `no-line-numbers` found,
+ * append the respective class.
+ * @param node the code block element, which is the 'code' node
+ * @param showCodeLineNumbers true if line numbers should be shown, false otherwise
+ */
+function setCodeLineNumbers(node, showCodeLineNumbers) {
+  const existingClass = node.attribs.class;
+  if (existingClass
+    && (existingClass.includes('line-numbers') || existingClass.includes('no-line-numbers'))) {
```
ang-zeyu, line 142: The first check here also tests for the second since it is a string, unlike `classList` in the browser.

Perhaps we could make this a little more robust with a regex in that sense too. (avoid something like `foo-line-numbers`)

packages/core/src/html/NodeProcessor.js
```diff
@@ -60,6 +61,8 @@ class NodeProcessor {
     this.footnoteProcessor = new FootnoteProcessor();
     this.mdAttributeRenderer = new MdAttributeRenderer(this.markdownProcessor);
     this.pageNavProcessor = new PageNavProcessor();
+
+    this.codeLineNumbers = codeLineNumbers;
```
ang-zeyu, line 65: Could we put this into `config` as well?

packages/core/src/Page/index.js
```diff
@@ -463,11 +463,12 @@ class Page {
...
 
-    const { variableProcessor, layoutManager, pluginManager } = this.pageConfig;
-
+    const {
+      variableProcessor, layoutManager, pluginManager, codeLineNumbers,
+    } = this.pageConfig;
     const pageSources = new PageSources();
     const nodeProcessor = new NodeProcessor(fileConfig, pageSources, variableProcessor,
-                                            pluginManager, this.pageUserScriptsAndStyles);
+                                            pluginManager, this.pageUserScriptsAndStyles, codeLineNumbers);
```
ang-zeyu, line 471: You'll need to update the ones for `Layout` / `External` too.

packages/core-web/src/styles/markbind.css
```diff
@@ -184,13 +184,13 @@ li.footnote-item:target {
     display: block;
 }
 
-.hljs:not(.inline):not(.no-line-numbers) > span {
+.hljs:not(.inline).line-numbers:not(.no-line-numbers) > span {
```
ang-zeyu, line 187: The `:not(.no-line-numbers)` selector can likely be omitted with your changes further below

packages/core/src/html/codeblockProcessor.js
```diff
@@ -129,6 +129,28 @@ function highlightCodeBlock(node) {
...
+    return;
+  }
+
+  if (showCodeLineNumbers) {
+    node.attribs.class = `line-numbers ${existingClass}`;
+  } else {
+    node.attribs.class = `no-line-numbers ${existingClass}`;
```
ang-zeyu, line 149: On the same lines as `:not(.no-line-numbers)`, this may not be needed

ang-zeyu (commented):
> One reason for doing this is so that the implementation is simpler and faster. Not sure if this would create too many assets? Would like to hear your opinion on this.

I'm favouring @tlylt's existing implementation here as well.

Conditionally importing assets is quite a bit of work: https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L1411
It also takes the css building out of our webpack bundling process which can be difficult down the line to debug / standardise.


We currently conditionally import different assets, but also only *external* ones (bootswatch / code dark themes).

One reason is that we'd have to include these (rather huge) assets into the webpack bundling process otherwise which increases build times & possibly tooling / configuration. (and we'd have a matrix of bundles then which isn't too build time friendly 🙁)

tlylt:
Hi reviewers,
Thank you for the kind reviews and suggestions. I have adjusted the code and hopefully fixed all the pending issues accordingly.
The new changes made are:
- show line numbers if "start-from" is set
- use regex to match for 'line-numbers' or 'no-line-numbers'
- update test cases
- update CSS selector
- minor update to the docs

docs/userGuide/syntax/code.mbdf
```diff
@@ -32,7 +32,38 @@ To enable syntax coloring, specify a language next to the backticks before the f
 </div>
 
 ##### Line numbering
-Line numbers are provided by default. To hide line numbers, add the class `no-line-numbers` to the code block as below
+Line numbers are <tooltip content="Line numbers were provided by default in version v3.1.1 and below. To preserve the exact line numbers behavior of sites generated in previous versions, simply set the codeLineNumbers option in site.json to true">hidden by default</tooltip>. To configure the line numbers setting for the entire site (for example, set it to show by default),
```
ang-zeyu, line 35:
```suggestion
Line numbers are <tooltip content="Line numbers were provided by default in version v3.1.1 and below. To preserve the exact line numbers behavior of sites generated in previous versions, simply set the codeLineNumbers option in site.json to true">hidden by default</tooltip>. To enable line numbers for the entire set by default,
```

docs/userGuide/syntax/code.mbdf
```diff
@@ -32,7 +32,38 @@ To enable syntax coloring, specify a language next to the backticks before the f
...
+  "titlePrefix": "FooBar Dev Docs",
+  "style": {
+    "bootstrapTheme": "bootswatch-cerulean",
+    "codeTheme": "light",
+    "codeLineNumbers": true // optional, false if omitted
+  }
+  // code below omitted for brevity
+}
+```
+refer to the [siteJsonFile](../siteJsonFile.html) for more details.
```
ang-zeyu, line 50:
```suggestion
refer to the [siteJsonFile](../siteJsonFile.html) for more details.
```
```suggestion

```
(hyperlink ```site.json` file`` above instead)

docs/userGuide/syntax/code.mbdf
```diff
@@ -32,7 +32,38 @@ To enable syntax coloring, specify a language next to the backticks before the f
...
+
+For each code block, the following can be done to override the site wide line numbers behavior:
+- To show line numbers, add the class `line-numbers` to the code block.
+- To hide line numbers, add the class `no-line-numbers` to the code block.
+
+The following shows you how to show/hide line numbers irrespective of the site wide setting:
```
ang-zeyu, line 56:
```suggestion
The following shows you how to show/hide line numbers irrespective of the site wide setting:
```
```suggestion
For each code block, you may also use the `line-numbers` or `no-line-numbers` classes to override the site-wide setting as such:
```

docs/userGuide/syntax/code.mbdf
```diff
@@ -32,7 +32,38 @@ To enable syntax coloring, specify a language next to the backticks before the f
...
 
 ##### Line numbering
-Line numbers are provided by default. To hide line numbers, add the class `no-line-numbers` to the code block as below
+Line numbers are <tooltip content="Line numbers were provided by default in version v3.1.1 and below. To preserve the exact line numbers behavior of sites generated in previous versions, simply set the codeLineNumbers option in site.json to true">hidden by default</tooltip>. To configure the line numbers setting for the entire site (for example, set it to show by default),
+add `"codeLineNumbers": true` to the `site.json` file:
+```json {highlight-lines="8[:]"}
+{
+  "baseUrl": "/myproduct",
+  "faviconPath": "myfavicon.png",
+  "titlePrefix": "FooBar Dev Docs",
```
ang-zeyu, line 41:
```suggestion
  "titlePrefix": "FooBar Dev Docs",
```
```suggestion
// ...
```

ang-zeyu (commented):
last few nits, otherwise lgtm! 👍

packages/core/src/html/codeblockProcessor.js
```diff
@@ -129,6 +129,32 @@ function highlightCodeBlock(node) {
...
 
+/**
+ * Adjust the class attribute of code blocks according to the global line numbers setting.
+ * Append the 'line-numbers' class if the global setting is true.
+ * @param node the code block element, which is the 'code' node
+ * @param showCodeLineNumbers true if line numbers should be shown, false otherwise
+ */
+function setCodeLineNumbers(node, showCodeLineNumbers) {
+  const existingClass = node.attribs.class || '';
+  const lineNumbersRegex = /^line-numbers\s|\sline-numbers\s|\sline-numbers$|^line-numbers$/;
```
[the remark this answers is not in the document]
tlylt: Thanks for the suggestion! I thought of this initially but tried without the () around `^|\s`, causing it to not work properly... 😓

docs/userGuide/syntax/code.mbdf
ang-zeyu: Meant for this to replace `52-54`, seems like a slightly awkward repitition

docs/userGuide/syntax/code.mbdf
ang-zeyu: Actually meant to collapse the above 2 lines on baseurl / favicon, to keep the snippet slightly smaller

I'll make a point of avoiding github's suggestions if its multiline 😅

packages/core/src/html/codeblockProcessor.js
```diff
@@ -129,6 +129,26 @@ function highlightCodeBlock(node) {
...
+/**
+ * Adjust the class attribute of code blocks according to the global line numbers setting.
+ * Append the 'line-numbers' class if the global setting is true.
+ * @param node the code block element, which is the 'code' node
+ * @param showCodeLineNumbers true if line numbers should be shown, false otherwise
+ */
+function setCodeLineNumbers(node, showCodeLineNumbers) {
+  const existingClass = node.attribs.class || '';
+  const styleClassRegex = /(^|\s)(no-)?line-numbers($|\s)/;
+  const hasStyleClass = existingClass.match(styleClassRegex);
```
ang-zeyu, line 141: one more tiny nit: you could also use the [`test`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#description) method instead of `match`, its slightly faster / a little more semantic for js

(same for the one markdown-it/index.js)

docs/userGuide/syntax/code.mbdf
tlylt: Oh ok, I think I misunderstood because multiline suggestions are indeed a bit confusing and I wasn't sure what to edit.

docs/userGuide/syntax/code.mbdf
```diff
@@ -32,7 +32,31 @@ To enable syntax coloring, specify a language next to the backticks before the f
 </div>
 
 ##### Line numbering
-Line numbers are provided by default. To hide line numbers, add the class `no-line-numbers` to the code block as below
+Line numbers are <tooltip content="Line numbers were provided by default in version v3.1.1 and below. To preserve the exact line numbers behavior of sites generated in previous versions, simply set the codeLineNumbers option in site.json to true">hidden by default</tooltip>. To enable line numbers for the entire set by default,
```
ang-zeyu, line 35:
```suggestion
Line numbers are <tooltip content="Line numbers were provided by default in version v3.1.1 and below. To preserve the exact line numbers behavior of sites generated in previous versions, simply set the codeLineNumbers option in site.json to true">hidden by default</tooltip>. To enable line numbers for the entire site by default,
```

ang-zeyu (approved):

Issue #1418, opened 2020-12-12
Improve broken link reporting


damithc:
MarkBind reports broken intra-site links during the build process (e.g., `markbind serve` command). Following improvements are suggested (can do in separate PRs):

- [x] Report each broken link only once (priority.Medium) (d.moderate)
- [x] Support validating `#hashes` for `<a>` tags _efficiently_ (priority.Medium) (d.moderate / high)
- [x] Provide a way to disable broken link reporting (priority.High)
- [x] Provide a way to suppress the warning for specific pages or links (priority.Low)
- [x] Restore broken link reporting for pics (disabled earlier due to #1416) (priority.Low)

ang-zeyu:
`<pic>` link checks are only disabled for v2.7.x (separate branch), and remains in the current working branch without issues.

it's certainly fixable in v2.7.x, but would require a moderate amount of effort

damithc:
@wxwxwxwx9 Any thoughts on this item?
>- [ ] Provide a way to disable broken link reporting (priority.High)

wxwxwxwx9:
> @wxwxwxwx9 Any thoughts on this item?
> 
> > * [ ]  Provide a way to disable broken link reporting (priority.High)

@damithc I think this is quite do-able. To implement this, I'm thinking of adding a new field in "site.json". Maybe something called "validateIntralink" as a key with a boolean as the value. 

As the "site.json" is read and transformed into "SiteConfig", we can pass down "validateIntralink" to "linkProcessor" to indicate whether we should even validate any link(s) in the first place. I think this should be quite straightforward. 

@ang-zeyu is there anything else I should take note of?

jovyntls:
Hello! For this part:

> Report each broken link only once (priority.Medium) (d.moderate)

Could I check what counts as a duplicated broken link? i.e. If the same (broken) intralink is on two different pages, should it be reported twice or only once? Should we report all pages where a broken link has been found?

Implementation wise, I was thinking of maintaining a set of links that have been validated, and only validate links if it is not already in the set.

damithc:
> Could I check what counts as a duplicated broken link? i.e. If the same (broken) intralink is on two different pages, should it be reported twice or only once? Should we report all pages where a broken link has been found?

Currently, if a certain markdown file has a broken link but that file is being included in many different places, is the problem reported once or for each of those inclusions?

Technically, the number of broken-links reported should match the number of fixes required, but not sure if we can achieve that.

[...]

PR #1735, opened 2022-01-25
Fix duplicate logging

jovyntls:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes the first checkbox of #1418.
Intra-site link validation is only done once for each link in a source file, such that the number of links reported matches the number of fixes required. An optional parameter was added to `process` functions to disable intralink validation.

**Testing instructions:**
Add a broken link to a page, and try including this page in other page (e.g. through `panel` or `includes`). 

e.g. For the `markbind init` site, 

topic2.md
```html
[This should appear once only]({{baseUrl}}/should_appear_once.png)

---

[This should appear twice]({{baseUrl}}/should_appear_twice.png)

[This should appear twice]({{baseUrl}}/should_appear_twice.png)

<panel header="topic3a" src="{{baseUrl}}/contents/topic3a.md" />
```

topic3a.md
```html
# 3a

<include src="{{baseUrl}}/contents/topic3b.md#firstitem" />

<panel src="{{baseUrl}}/contents/topic3b.md" />

<include src="{{baseUrl}}/contents/topic3b.md#hello" />
```

topic3b.md
```html
# 3b

<div id="firstitem">

First invalid link in 3b
[1 - should appear once in 3b](invalid3b_1_should_appear_once).md

</div>

Second invalid link in 3b
[2 - should appear once in 3b](invalid3b_2_should_appear_once.md)

<div id="hello">

Third invalid link in 3b
[3 - should appear once in 3b](invalid3b_3_should_appear_once.md)

</div>
```

```
Fix duplicate intrasite link reporting
```

---

- [x] No unrelated changes

### DIFF
--- a/packages/core/src/html/NodeProcessor.js
+++ b/packages/core/src/html/NodeProcessor.js
@@ -1,6 +1,7 @@
 const path = require('path');
 
 const cheerio = require('cheerio');
+const lodashHas = require('lodash/has');
 const htmlparser = require('htmlparser2'); require('../patches/htmlparser2');
 const fm = require('fastmatter');
 const Promise = require('bluebird');
@@ -251,9 +252,9 @@ class NodeProcessor {
     }
   }
 
-  _process(node, context) {
+  _process(node, context, shouldValidateIntraLinks = this.config.intrasiteLinkValidation.enabled) {
     if (_.isArray(node)) {
-      return node.map(el => this._process(el, context));
+      return node.map(el => this._process(el, context, shouldValidateIntraLinks));
     }
     if (NodeProcessor._isText(node)) {
       return node;
@@ -264,7 +265,9 @@ class NodeProcessor {
     if (linkProcessor.hasTagLink(node)) {
       linkProcessor.convertRelativeLinks(node, context.cwf, this.config.rootPath, this.config.baseUrl);
       linkProcessor.convertMdAndMbdExtToHtmlExt(node);
-      if (this.config.intrasiteLinkValidation.enabled) {
+      const isSourceFile = context.callStack.length === 0
+        && (!lodashHas(node.attribs, 'src') || context.cwf.includes(node.attribs.src));
+      if (shouldValidateIntraLinks && isSourceFile) {
         linkProcessor.validateIntraLink(node, context.cwf, this.config);
       }
       linkProcessor.collectSource(node, this.config.rootPath, this.config.baseUrl, this.pageSources);
@@ -291,7 +294,7 @@ class NodeProcessor {
 
     if (node.children) {
       node.children.forEach((child) => {
-        this._process(child, context);
+        this._process(child, context, shouldValidateIntraLinks);
       });
     }
 
@@ -314,7 +317,8 @@ class NodeProcessor {
     return node;
   }
 
-  process(file, content, cwf = file, extraVariables = {}) {
+  process(file, content, cwf = file, extraVariables = {},
+          shouldValidateIntraLinks = this.config.intrasiteLinkValidation) {
     const context = new Context(cwf, [], extraVariables, {});
 
     return new Promise((resolve, reject) => {
@@ -326,7 +330,7 @@ class NodeProcessor {
         const mainHtmlNodes = dom.map((d) => {
           let processed;
           try {
-            processed = this._process(d, context);
+            processed = this._process(d, context, shouldValidateIntraLinks);
           } catch (err) {
             err.message += `\nError while rendering '${file}'`;
             logger.error(err);
--- a/packages/core/src/html/NodeProcessor.js
+++ b/packages/core/src/html/NodeProcessor.js
@@ -1,7 +1,6 @@
 const path = require('path');
 
 const cheerio = require('cheerio');
-const lodashHas = require('lodash/has');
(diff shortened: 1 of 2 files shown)


packages/core/src/html/NodeProcessor.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 
 const cheerio = require('cheerio');
+const lodashHas = require('lodash/has');
```
ryoarmanda, line 4: We already have this require as `_.has` several lines below, let's just use that to remove duplication

ryoarmanda, line 4: We already have this require several lines below as `_.has`. Let's use that to minimise duplication

packages/core/src/html/NodeProcessor.js
```diff
@@ -314,7 +317,8 @@ class NodeProcessor {
     return node;
   }
 
-  process(file, content, cwf = file, extraVariables = {}) {
+  process(file, content, cwf = file, extraVariables = {},
+          shouldValidateIntraLinks = this.config.intrasiteLinkValidation.enabled) {
     const context = new Context(cwf, [], extraVariables, {});
```
ryoarmanda, line 322: You can leverage the use of the `Context` object to store the `shouldValidateIntralinks`. `Context` has a property called `processingOptions` that is an object intended to store any options for processing, and it looks like it fits for the situation, with the added bonus that you don't need to keep passing on the flag as a parameter.

ryoarmanda (changes requested):
Nice work @jovyntls ! First glance looks okay, I will get back on the overall process in a bit. Here are some change suggestions to minimise duplication and parameter drilling down:

packages/core/src/html/NodeProcessor.js
jovyntls: I see, I didn't catch that earlier, thank you!

packages/core/src/html/NodeProcessor.js
jovyntls: Thanks for the help! I've switched to using `processingOptions` for passing the flag from `process` to `_process`.

packages/core/src/html/NodeProcessor.js
```diff
@@ -264,7 +264,9 @@ class NodeProcessor {
     if (linkProcessor.hasTagLink(node)) {
       linkProcessor.convertRelativeLinks(node, context.cwf, this.config.rootPath, this.config.baseUrl);
       linkProcessor.convertMdAndMbdExtToHtmlExt(node);
-      if (this.config.intrasiteLinkValidation.enabled) {
+      const isSourceFile = context.callStack.length === 0
```
ang-zeyu, line 267: on the same lines, this may disable link checking entirely for files that are only `<include />`-ed (not used as a page)

ang-zeyu (commented):
> How about:
> 
> * report all broken links at once, at the end, once all pages have finished building (minimize disjointed messages, should be simpler to implement as well)
> * for any given source file, report all its broken links, and the count of broken links only once

Sorry for the late input.

If I'm not mistaken here (below comments), and in case I miscommunicated previously,

The solution I had in mind would entail creating some sort of *new*, *global* (shared across pages) abstraction (e.g. `SiteLinkChecker`) that stores broken link information for delayed logging.

Currently this is the "standard" message we log for all files, which we immediately do so on encountering the link:

```
const err = `You might have an invalid intra-link! Ignore this warning if it was intended.
'${resourcePath}' found in file '${cwf}'`;
```

- `resourcePath` is the `src/href/etc` attribute of the html tag as-is
- `cwf` can be many things (depending on boilerplate / external / layout / page), but is essentially "where relative file paths are to be computed from"


For this issue, you'll need to ensure that no 2 messages containing the same `resourcePath` has the same `cwf`, which can be done using the new global abstraction. Optionally you can also count the number of duplicates / references.

You may want to look through the original PR discussion https://github.com/MarkBind/markbind/pull/1375 which might help dispel any doubts about the original requirements too.

---

Also note our current message is by no means perfect but is simple enough to suffice for most use cases - ideally we should be logging the "full trace" of linkages to make it clear what caused the broken link for more complex `<include/panel (optionally)boilerplate>` chains. This is unfortunately much more complex due to boilerplates and externals, but you could also aim to accomplish this by the end of the semester! (this is not listed in #1418)

---

Another reason I had in mind for this abstraction is to facilitate the second item in #1418 (hash checking), by storing all encountered hashes in it, and doing the checks after all pages have finished building.

jovyntls:
Hi @ang-zeyu , thanks so much for the detailed explanation! Sorry I misunderstood the root issue previously, your clarification was very helpful. I'll look into creating the abstraction to deal with the different cases of link validation.

Should I close this PR and restart on a new one?

ryoarmanda:
Ooh yeah, I was a bit out of the loop on the issue discussion and just saw on the direction of the suggested fix. My bad as well 😅

An abstraction sounds like a nice plan (like what @ang-zeyu mentioned). Essentially it contains the broken links over time while checking is in progress, and can output the links as messages later on when everything is done.

@jovyntls , I guess we can close this PR as you are exploring the abstraction fix. You can revisit this PR to see whether some of the changes here would carry over to that fix. 👍

Issue #1699, opened 2021-12-28
Navbar overflows when there is a long list of items


jonahtanjz:
* **MarkBind Version:**
3.0.6

**What did you do? Please include the actual source code causing the issue.**
When a navbar contains many nav links or have a long list of content, the navbar will overflow and the overflowed items will be hidden.

Reference: https://github.com/MarkBind/markbind/pull/1690#discussion_r771133015

**What did you expect to happen?**
Should make each section scrollable if their content overflows.

**What actually happened? Please include the actual, raw output.**

![image](https://user-images.githubusercontent.com/47470981/147527905-06701115-58f0-4490-bbc7-8b6f4acf814f.png)

jonahtanjz:
Related to #1429

kaixin-hc:
Hi! I'll take a look at this next (and #1731 and #1429, since they're all related to mobile usability)

PR #1760, opened 2022-02-12
Fix navbar overflow on mobile

### LINKED ISSUE #1429: Small search bar and cropped dropdown menu on iPad

[CS2103T module website](https://nus-cs2103-ay2021s1.github.io/website/index.html) opened on an iPad Air 3. 

* **MarkBind Version:** MarkBind 2.15.2

I opened the CS2103T module website using Safari on my iPad Air 3 in portrait mode. It has been zoomed out all the way. 

I have not identified the actual source code causing this issue.

I expected to see a 'hamburger' dropdown button with additional options as shown below. This is the default option for smaller screens (emulated using Chrome dev tools)

image

I saw a tiny search bar and a dropdown menu that is partially cropped. Please refer to the screenshots below.

IMG_0013

IMG_0011

kaixin-hc:
Fixes #1429

Also related:  #1699
- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
- Set a minimum width for the scrollbar to be 8em. I tested this on the different device sizes of chrome and I think it looks fine on all; should guarantee at least least the whole placeholder attribute 'search' to be visible on the smallest screen sizes. 
- Make the section with items in the navigation bar scrollable in the case of overflow

search_scroll
Screenshot 2022-02-12 at 5 52 51 PM

- @jonahtanjz since this modifies your previous changes for mobile, I'd like your opinion!
- Hiding the scrollbar in the navigation bar looks nice, but I'm not sure if it's very accessible. 
- Used em instead of vw for accessibility reasons(as it scales if the user zooms the screen)

**Testing instructions:**
Add lots of elements to the navigation bar to see it scroll + use google chrome mobile.

Make navbar scrollable & optimise search input size

---

- [X] No unrelated changes


tlylt (commented):
> Hiding the scrollbar in the navigation bar looks nice, but I'm not sure if it's very accessible.

I agree with you that it looks nice without the scrollbar but I feel that indeed there might be valid concerns about accessibility. 

In an extreme case where the navbar has quite a few items and some of them are nicely hidden behind the search box, it might be frustrating for site visitors who then become unable to find the target nav item.
image

I think perhaps one way is to explore creating [custom css scrollbar](https://css-tricks.com/almanac/properties/s/scrollbar/)  which may be more visually appealing. 
Or, perhaps we can consider something like what Github is doing?
image

tlylt:
> What github extension are you using?

https://github.com/refined-github/refined-github

jonahtanjz:
> Hiding the scrollbar in the navigation bar looks nice, but I'm not sure if it's very accessible.

The issue with using the native scrollbar is that it can differ depending on OS and browser. For example this is how it looks on windows 10 Chrome

image

In this case I would agree with @tlylt to create a custom scrollbar for desktop to keep the design consistent.

> Or, perhaps we can consider something like what the Github extension that I am using is doing?
> I think putting a dropdown for excess navbar elements might be a better solution than a custom scrollbar.

I think this is a good idea. We can follow how Github is doing it 👍 

image

ong6:
Just throwing my opinion out here, but I think a native scrollbar is not the best solution, a dropdown like @tlylt and @jonahtanjz 
suggested is better. Especially when users want to see all the options in the navbar.

kaixin-hc:
Hmm, I'm also leaning towards the dropdown (which isn't how github does it but an extension) as opposed to the custom scrollbar. Considering this, I'll edit the description to just fixing the iPad issue? To keep changes separate, would it be possible to merge this first to fix the bug and then I make a separate PR for the dropdown?

jonahtanjz:
> Any thoughts?

#1768

jonahtanjz (commented):
@kaixin-hc, I think we can merge this first while waiting for the discussion on scoped styles.

One last thing, I noticed that by using `overflow-x: scroll` on the default navbar, it causes the dropdown to break.

https://user-images.githubusercontent.com/47470981/154031257-53735f15-2972-48fd-ad22-827010b429e0.mp4

It may take quite a bit of work to resolve this. Since we are intending to remove scroll for ipad/desktop and use dropdown for overflow, there is not much value in resolving this. We can limit this PR to just the searchbar and remove the changes made for the scrolling :)

kaixin-hc:
Sorry for the delay @jonahtanjz ! I've reverted the changes.

I was curious about the scrollbar issue, so I did a little digging for my own learning. After a few stack overflow posts and reading, I found this article: https://css-tricks.com/popping-hidden-overflow/ that explains that setting `overflow-x` sets `overflow-y` as well, and it's just not possible to have a dropdown peep out of a scrollable overflow without setting positions relatively. [This discussion](https://www.sitepoint.com/community/t/css-drop-down-menu-hidden-behind-horizontal-scrollbar/367783) with the offered [solution](https://codepen.io/paulobrien/embed/vYxWppv?) was also interesting. Not sure how to implement it cleanly in a vue component, but it is a non-issue.

tlylt:
@all-contributors please add @kaixin-hc for code

packages/vue-components/src/Navbar.vue
```diff
@@ -306,7 +306,6 @@ export default {
         .navbar-default {
             display: block;
             margin-top: 0.3125rem;
-            width: 100%;
```
jonahtanjz, line 309: @kaixin-hc I believe this is still needed so that the `default-navbar` will always go to the next line on mobile :)

Tried with 2 items in the `default-navbar` at about 700px:

image

kaixin-hc: Screenshot 2022-02-28 at 1 21 48 PM
Good catch! Though I've enclosed a picture for further reference, as I think that it looks quite strange with just two items. After some consideration I think it's still alright, since I expect this would be an edge case and it doesn't seem to have any usability issues. An alternative work around might be triggering the "go to next line" when the screen size is smaller or when the nav-bar default has a certain number of items, but I think that's more trouble than it's worth

jonahtanjz (approved):

PR #1762, opened 2022-02-12
Remove .mbdf and .mbd

### LINKED ISSUE #1672: Remove `.mbd / .mbdf`
I probably don't have the relevant context but this has been bugging me for a very long time now anyway =X
What's the purpose of these extra extensions?

`.mbdf` is used to denote "fragment" but really other than that, it dosen't provide any concrete value. My opinion thus is that file organisation / hierarchy here is a responsibility that should be left to the user. There are already many cases where you would need to `<include>` from a non `.mbdf` file.

`.mbd` files on the other hand don't really seem to have a use case.

I don't think we should be reinventing the wheel when it comes to file extensions here and should just stick with the canonical default (`.md`), for familiarity.

There's also the issue of syntax highlighting in ides that needs to be manually configured for `mbd / mbdf` files (which sometimes isn't even an option)

*This is a huge breaking change of course.

ong6:

- [x] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [x] Others, please explain:

Fixes #1672 

**Overview of changes:**

Removes allowing of `.mbdf` and `.mbd` as file extension types.
- Added new section in UG documentation under `reusing contents` called `Creating Custom Fragments`, [view here](https://deploy-preview-1762--markbind-master.netlify.app/userguide/reusingcontents#creating-custom-fragments)
- Updated all related core and cli files using `.mbdf` and `.mbd`
- Updated all existing documentation referring to `.mbdf` and `.mbd`.
- Updated existing test cases to remove `.mbdf` and `.mbd`
- Added new test for `custom-fragment.md`
- Fixed typos/errors in documentation found while updating docs (may be unrelated change)

Some context for this PR:
- `.mbd` was created with the intent of making our own extension called .markbind
- `.mbdf` was created to easily exclude fragments from page generation
- Since the addition of `pagesExclude` with PR #1328, the definition of fragments can be described by the users hence rendering the first 2 points obsolete.

**Testing instructions:**

- Run `npm run test` in `packages/cli` 
- All test cases should pass

Remove `.mbdf` and `.mbd` as markbind supported file extensions.

Syntax parsers do not work with `.mbdf` and `.mbd` and 
this functionality could be implemented using other means (e.g. pagesExclude)

Refer to this for the full disscussion #1672.

---

- [ ] No unrelated changes


damithc:
As this is a breaking change, we won't be able to do any v3.* releases (unless as hot patches to previous release) after merging this PR, right?

ong6:
@damithc Actually I'm not too sure about this maybe @ang-zeyu can respond? 

I think its good SWE practice to release this as a new version though? Since this will indeed remove all support for existing .mbdf and .mbd files. 

Perhaps I can include a new markbind command to fix all the old .mbdf and .mbd files? Thoughts on this?

damithc:
@ong6 at the same time, can you insert a tip somewhere in the user docs to guide authors on how to exclude fragment files from rendering?

jonahtanjz:
> As this is a breaking change, we won't be able to do any v3.* releases (unless as hot patches to previous release) after merging this PR, right?

Yup. Will need to create a branch for hot patches releases (if any) after this is merged.

jovyntls (commented):
Hi @ong6 ! I also noticed there's still one use of `.mbd` [here](https://github.com/ong6/markbind/blob/remove-mbd/mbdf/packages/core-web/src/VueCommonAppFactory.js#L22) - I realise quite a few sites still use .mbd as the index file, so I'm not sure how it should be handled in this case?

tlylt (changes requested):
@ong6 Good work!
I have highlighted some minor issues in the user guide. 
Since this PR is a breaking change, perhaps we can have a more detailed proposed commit message to explain a little bit more about the context?

docs/userGuide/siteJsonFile.md
```diff
@@ -124,7 +123,7 @@ _(Optional)_ **The styling options to be applied to the site.** This includes:
   * `src` can be used to specify a single file, or an array of files.<br>
   {{ icon_examples }} `docs/index.md` or `[ 'docs/index.md', 'docs/userGuide.md' ]` { .my-1 }
   * `glob` can be used alternatively to define a file pattern in the [_glob syntax_](https://en.wikipedia.org/wiki/Glob_(programming)), or an array of such file patterns.<br>
-  {{ icon_examples }} `**/*.md` or `[ '**/*.md', '**/*.mbdf' ]` { .my-2 }
+  {{ icon_examples }} `**/*.md` or `[ '**/*.md', '**/*.md' ]` { .my-2 }
```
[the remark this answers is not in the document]
ong6: Nice catch, missed those cases! Committed both suggested changes.

ong6:
> @ong6 Good work! I have highlighted some minor issues in the user guide. Since this PR is a breaking change, perhaps we can have a more detailed proposed commit message to explain a little bit more about the context?

Thanks for the catches, I've made the changes, perhaps instead of a more detailed explanation in the commit message, I will update the highlight/discussion part instead?

packages/cli/test/functional/test_site/sub_site/nested_sub_site/site.json
```diff
@@ -21,7 +19,7 @@
       "glob": "**/index.md"
     },
     {
-      "glob": "**/*.+(md|mbd)"
+      "glob": "**/*.+(md)"
```
jonahtanjz, line 22: I believe this can be simplified to
```suggestion
      "glob": "**/*.md"
```
Same for the other glob pattern.

packages/cli/test/functional/test_site/expected/sub_site/nested_sub_site/testNunjucksPathResolving.html
```diff
@@ -213,7 +213,7 @@ <h3 id="testing-site-nav"><span id="testing-site-nav" class="anchor"></span>Test
           The root directory in this case is the respective root directory of the root site or sub sites.</p>
         <hr>
         <p><strong>Test {% include %}</strong></p>
-        <p>This is content from testNunjucksPathResolvingInclude.mbdf located at <code class="line-numbers hljs inline no-lang" v-pre>sub_site/nested_sub_site</code></p>
+        <p>This is content from <a href="http://testNunjucksPathResolvingInclude.md">testNunjucksPathResolvingInclude.md</a> located at <code class="line-numbers hljs inline no-lang" v-pre>sub_site/nested_sub_site</code></p>
```
jonahtanjz, line 216: Looks like a bug here? This shouldn't be wrapped in `<a>` and the `href` attribute does not make sense.

jonahtanjz (commented):
Nice work @ong6!

Thanks @tlylt and @jovyntls for the reviews 👍 

> Since this PR is a breaking change, perhaps we can have a more detailed proposed commit message to explain a little bit more about the context?

> I've made the changes, perhaps instead of a more detailed explanation in the commit message, I will update the highlight/discussion part instead?

I think it's better to put it in the commit message as it will be easier to track from the commit history.

packages/core/src/utils/fsUtil.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require('fs-extra');
 const ensurePosix = require('ensure-posix-path');
 
-const markdownFileExts = ['.md', '.mbd', '.mbdf'];
+const markdownFileExts = ['.md'];
```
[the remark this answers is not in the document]
ong6: ok i'll make this change

ong6:
> Nice work @ong6!
> 
> Thanks @tlylt and @jovyntls for the reviews 👍
> 
> > Since this PR is a breaking change, perhaps we can have a more detailed proposed commit message to explain a little bit more about the context?
> 
> > I've made the changes, perhaps instead of a more detailed explanation in the commit message, I will update the highlight/discussion part instead?
> 
> I think it's better to put it in the commit message as it will be easier to track from the commit history.

Thanks for the review @jonahtanjz I've made all the suggested changes!

packages/cli/test/functional/test_site/sub_site/nested_sub_site/testNunjucksPathResolvingInclude.md
```diff
@@ -1,3 +1,3 @@
-This is content from testNunjucksPathResolvingInclude.mbdf located at `sub_site/nested_sub_site`
+This is content from testNunjucksPathResolvingInclude.md located at `sub_site/nested_sub_site`
```
jonahtanjz, line 1:
```suggestion
This is content from `testNunjucksPathResolvingInclude.md` located at `sub_site/nested_sub_site`
```

Might want to add a code fence here to prevent it from converting to a link in the expected test files :)

packages/core/src/utils/fsUtil.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require('fs-extra');
 const ensurePosix = require('ensure-posix-path');
 
-const markdownFileExts = ['.md', '.mbd', '.mbdf'];
+const markdownFileExts = '.md';
```
jonahtanjz, line 5: Will have to update [isMarkdownFileExt](https://github.com/MarkBind/markbind/blob/864900461b9bd1e46b069fecd8d5455a3da78ff7/packages/core/src/utils/fsUtil.js#L18) method as well since it is no longer an array. Can use `===` instead of `includes`.

jonahtanjz (commented):
Thanks for making the changes @ong6

Just 2 more comments :)

packages/core/src/utils/fsUtil.js
ong6: Includes will still work with strings, so there should be no issue with this particular case. Unless the change is for readability reasons?

image

jonahtanjz: I believe this will give a false positive? Although unlikely, but files with extension `.m` will be returned as `true`.

ong6: Ahh ok I see, will implement the change then!

packages/cli/index.js
```diff
@@ -118,10 +117,9 @@ program
     const logsFolder = path.join(rootFolder, '_markbind/logs');
     const outputFolder = path.join(rootFolder, '_site');
 
-    const defaultFiles = [INDEX_MARKDOWN_FILE, INDEX_MARKBIND_FILE];
-    const presentDefaultFile = defaultFiles.find(fsUtil.fileExists);
+    const presentDefaultFile = fsUtil.fileExists(INDEX_MARKDOWN_FILE);
```
jonahtanjz, line 120:
```suggestion
    const presentDefaultFile = fsUtil.fileExists(INDEX_MARKDOWN_FILE) ? INDEX_MARKDOWN_FILE : false;
```

Sorry my bad, this needs to be assigned the `index.md` if it exists 😓.

jonahtanjz (commented):
Thanks @ong6 :) You may want to update the commit message as well following [this template](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message).

Other than that the rest looks good 👍

ong6:
> Thanks @ong6 :) You may want to update the commit message as well following [this template](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message).
> 
> Other than that the rest looks good 👍

Oh nice spot, didn't realize that issue also! Thanks for checking my PR once again. I have updated the commit message as well!

Issue #1544, opened 2021-04-12
Setup testing for live preview features


ang-zeyu:
**Is your request related to a problem?**

With #1513, #1529, etc., there's quite a chunk of logic concerning live reload now.

**Describe the solution you'd like**

Would be good to explore and add in some e2e live-preview testing to prevent regressions.

tlylt:
Will Cypress be a good option to explore, in addition to Puppeteer? I have previously used Cypress for frontend testing and it is getting very popular in recent times.

I am a bit confused as to how to test a live-reload functionality. Do you have any rough idea as to how it can be implemented?
It seems to me that we need to perhaps:
- start running the server and serve the page on a browser
- start running a test case on a page
- somehow modify a file and save a file, and that triggers the reloading of the page
- cypress(or others) while the current test case is ongoing, picks up the changes in the page and verify that something has been updated

I will try to research more on this in the meantime :)

ang-zeyu:
My primary concern with cypress would just be cost; There's an open source plan but it I'd rather go for something free from the offset.
Puppeteer is also maintained by the chrome devtools team so we should be in good hands here =P

There's also the lesser issue of cypress startup being much [longer](https://blog.checklyhq.com/cypress-vs-selenium-vs-playwright-vs-puppeteer-speed-comparison/#scenario-1-single-end-to-end-test-against-a-static-website). I haven't used cypress in a while though so I'm not sure if that's changed.

Feel free to explore cypress still though, there might be some other benefits I'm not aware of. (e.g. if puppeteer dosen't support the necessary apis)

> I am a bit confused as to how to test a live-reload functionality. Do you have any rough idea as to how it can be implemented?
> It seems to me that we need to perhaps:

Pretty much what I had in mind for now!

A little clunky but since we are testing filesystem event handlers / websockets / etc. it might be best to replicate user action as much as possible. (e.g. `fs.writeFileSync` to trigger the change)

tlylt:
> My primary concern with cypress would just be cost; There's an open source plan but it I'd rather go for something free from the offset. Puppeteer is also maintained by the chrome devtools team so we should be in good hands here =P

[...]

Issue #1507, opened 2021-03-08
Live preview shows stale version of a page occasionally


damithc:
In recent times, I have notices the the live preview for smaller websites occasionally shows the stale version of the page (although the page refreshes i.e., shows the 'building' spinner and loads again). The new content appears only after a manual refresh.

But this can catch the user unawares. Not sure if there is a solution though. Anyway, the first step is to be able to reproduce it reliably.

ong6:
@damithc can I clarify if this is the issue that you are facing?
![StaleReload](https://user-images.githubusercontent.com/19247856/136562270-c0aa396d-aa79-470f-8526-1f1ed4295193.gif)

Or is it something else? I have not seen the page refresh automatically on the `markbind serve` command.

PR #1763, opened 2022-02-13
[WIP] Add e2e tests for live preview features

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Setup e2e tests for live preview features.
addresses #1544 

Libraries used:
- Jest
- Jest puppeteer (for jest to work well with puppeteer with its convenient setup & functions)
- puppeteer 

This PR hopes to set up the relevant dependencies and test frameworks for testing of live preview features. In the midst of working on this, it seems like the live preview feature for changes in `site.json` is particularly problematic. My test cases for changing of attributes in `site.json` are extremely flaky. My investigation shows that there are a few potential errors around this feature:
1. The update of the bootstrap theme seems to trigger some file action errors due to the removal/copying of the files
image
2. Sometimes the site.json file becomes unavailable
image
3. Sometimes the page does not reload/reload after a long delay
4. Sometimes the port is unavailable when ran in GitHub Actions and hence the server failed to start

After adjusting the timeout values and tweaking the test cases for a bit, I think the errors still occur sporadically. Hence, I am removing the flaky components and leaving the rest of the set-up changes + a simple test on live-reload for edits to a markdown file. 

P.S The flaky test file is [here](https://gist.github.com/tlylt/89007db20ac4a97267cfc85dfe581807), to include and run it locally, simply put this file in the directory `cli/test/e2e`. Note that some of the errors mentioned above are not always present, but can be seen in previous CI runs e.g. [here](https://github.com/MarkBind/markbind/runs/5218075311?check_suite_focus=true) and [here](https://github.com/MarkBind/markbind/runs/5209566643?check_suite_focus=true).

Would like to get some feedback on the approach and wonder if we should split the TODOs into a few PRs, with this being the first PR to set up the test environment.

The remaining TODOs:
- [ ] Add tests for changes to `site.json` (after fixing the underlying issues)
- [ ] Add more tests to verify the live-reload feature works for different kind of changes to a source file
- [ ] Add tests for different serve commands
  - "the -o option should not trigger a page build or reload the page if the file change is not a dependency of the pages being viewed" (as proposed by Ze Yu)

**Testing instructions:**
To run the e2e test cases added, cd into `packages/cli` and run `npm run test:e2e`
To view the browser instance, set `headless` to `false` in `packages\cli\jest-puppeteer.config.js`

Setup e2e tests for live preview features

---

- [x] No unrelated changes


tlylt:
Hi guys, would like to get some initial feedback on this PR if possible (details in the PR description).

ang-zeyu (commented):
Just a really brief look -- I think the overall setup looks ok

> Would like to get some feedback on the approach and wonder if we should split the TODOs into a few PRs, with this being the first PR to set up the test environment.

👍 let's work on the other test cases in separate PRs.

Although perhaps we could get to the bottom of the flaky test issue in case any setup changes are required here (but ok eitherway).

> P.S The flaky test file is [here](https://gist.github.com/tlylt/89007db20ac4a97267cfc85dfe581807.js), to include and run it locally, simply put this file in the directory cli/test/e2e

Not sure if I'm missing something here though, this looks like a frontend file but I'm not seeing the jest code orchestrating the testing.

tlylt:
> Not sure if I'm missing something here though, this looks like a frontend file but I'm not seeing the jest code orchestrating the testing.

Sorry I copied the wrong link. Updated [link](https://gist.github.com/tlylt/89007db20ac4a97267cfc85dfe581807)

packages/cli/jest-puppeteer.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  launch: {
+    headless: true, // set to false to view browser window
+  },
+  server: {
```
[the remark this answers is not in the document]
tlylt: I just checked the APIs for jest-puppeteer, which does allow launching of multiple servers of different setup [here](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-dev-server#usage)
> image

If this works then I think I will need to have two different sets of test sites to prevent side effects. So after tests on -f is done, proceed to test with -o which is serving content from another site. Do you think this is fine?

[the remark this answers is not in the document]
ang-zeyu: Ideally: setup -> teardown (also cleanup site edits, so we can keep to 1 test site perhaps) -> setup -> teardown, as there are quite a few combinations which might overload gh-actions. (e.g. `-o` `-o -b` `-b` `-f` ...)

ang-zeyu:
> Sorry I copied the wrong link. Updated [link](https://gist.github.com/tlylt/89007db20ac4a97267cfc85dfe581807)

> Sometimes the page does not reload/reload after a low delay

Hmm, I'm not seeing any issues apart from `3`. Specifically the tests seem to zoom past `pages.title` into `afterEach` far too quickly, this results in the second refresh never being triggered by the websocket.

It might be related to https://github.com/MarkBind/markbind/issues/1507.

Adding a little timeout after the `pages.title` test as such seems to solve the issue for me. (though I'm not sure we should be relying on this in the long run)

The underlying issue may be difficult to fix for now as the file change handlers and watchers in `live-server` and markbind's `cli/index.js` are currently separate; I think @ryoarmanda could advise whether there is a quick fix here as well.

I would be ok with a more ad-hoc fix for now otherwise:
- generous timeouts for the specific offending test
- another more reliable but hacky solution would be to launch the process independently and hook up the standard io, then detect for `"site data built"` like messages (then trigger the refresh using puppeteer manually)

Think we could also get a few more folks to verify the file you linked for the other issues I'm not getting too.

jonahtanjz:
> Think we could also get a few more folks to verify the file you linked for the other issues I'm not getting too.

@ang-zeyu @tlylt 

The linked file seems fine to me as well. For `1`, not able to replicate on my side but my guess is that the theme file somehow failed to copy over? Not sure what is causing this tho. For `2`, managed to replicate once after running the tests a few times. It could be caused by 2 different tests/processes is trying to write to the same `site.json` file, which corrupted the `site.json` file.

If this occurs, a re-run of the ci tests should resolve the issue from what I can see from my side as it only happens once in awhile but doesn't seem like a proper way to resolve this. My suggestion is to try to figure out the root cause first before adding the tests for `site.json`. What do you all think?

Note: There also seems to be some issue with the `faviconPath` test. Sometimes this error shows up but most of the time it seems fine.
image

ang-zeyu:
To clarify on `3`, we currently have 2 sets of file change handlers, one watching the source directory, the other belonging to the `live-server` package that watches the output directory.

An example likely series of events that causes the tests to get stuck at `pages.title` : 
1. `pages.title` change causes index.md to rebuild, at the same time, `page.waitForNavigation()` is called
1. index.md rebuilds, `waitForNavigation()` resolves. However, at the same time, the websocket for live-server hasn't established yet.
1. The tests proceed into `afterEach` far too quickly and restore the original site.json, which triggers another index.md rebuild. This triggers the file change handler in `live-server` but as the websocket is not yet established, the page is not refreshed. (manually refreshing with `headless: false` does the trick)

The change is specific to `pages.title` (at least on my system) likely as a much less costly branch is executed for this change (https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L955), resulting in step 3 (the second index.md` rebuild) proceeding far too quickly.

---

There are likely many other cases of race conditions from file handlers which only show up on certain occassions / systems due to processing speed / etc., resulting in the discrepancies here.

> My suggestion is to try to figure out the root cause first before adding the tests for site.json. What do you all think?

E.g. Likely, the pictures in `1` & `2` seem to indicate 2 *quick, successive* config file changes interpreted as one. (although, I'm still not able to reproduce this) This causes a whole bunch of races with file access / writes in turn.

We currently don't have a system for "stopping the world" / "ignoring file changes" while a build is progress (and queueing these changes for later), I think that could be implemented fairly quickly and would definitely be nice to have even in a non-test scenario (if user changes the config file successively and ctrl+S quickly). (PR would be welcome!) We could also limit this addition to site configuration file changes for now and see how it turns out.

---

Again, a more ad-hoc fix might also work for now otherwise, as these are not trivial to fix, and e2e tests for the huge amount of live-reload features we have now are fairly important.

packages/cli/jest-puppeteer.config.js
```diff
@@ -0,0 +1,11 @@
...
+    command: 'node index.js serve test/e2e/test_site -n -p 8888',
```
ang-zeyu, line 6: missing `-d` (to test the latest changes of the PR)

ang-zeyu (commented):
In view of the potential unreliability (if going with timeout fix for example) of e2e tests perhaps we could also set this up as a separate optional workflow for now, and see how things turn out.

packages/cli/jest-puppeteer.config.js
tlylt: I discovered that this is not possible at the moment and reported the bug in #1891. Nonetheless, I think it is still valuable to test the current serve command without the `-d` (good to test it as well, perhaps later on), since the test cases that I have currently written should not be impacted by latest changes

tlylt:
Hi @ang-zeyu,
Given the timeline, wonder if the following update would be sufficient for this PR to be merged:
- Create a separate workflow to run e2e tests, such that it continues on failure (hence "optional"), to see how things turn out
- Update the initialization of servers per your [comment](https://github.com/MarkBind/markbind/pull/1763#discussion_r816673781). I experimented a little and was not able to do it properly with jest-puppeteer, because somehow it does not tear down the server correctly. Will try a few other ways to see how it goes. Doing it without jest-puppeteer should be possible (just like how we do our snapshot tests), but seems a little troublesome to DIY.
- Add the flaky test file with an increased timeout, which hopefully resolves the issues mentioned most of the time. (I am also fine to leave it out, need some opinion on this)

packages/cli/jest-puppeteer.config.js
ang-zeyu: As a temporary workaround, try (in this example)

```
cd <test/e2e/test_site> && ... run the command
```

ang-zeyu:
> Hi @ang-zeyu, Given the timeline, wonder if the following update would be sufficient for this PR to be merged:
> 
> * Create a separate workflow to run e2e tests, such that it continues on failure (hence "optional"), to see how things turn out
> * Update the initialization of servers per your [comment](https://github.com/MarkBind/markbind/pull/1763#discussion_r816673781). I experimented a little and was not able to do it properly with jest-puppeteer, because somehow it does not tear down the server correctly. Will try a few other ways to see how it goes. Doing it without jest-puppeteer should be possible (just like how we do our snapshot tests), but seems a little troublesome to DIY.
> * Add the flaky test file with an increased timeout, which hopefully resolves the issues mentioned most of the time. (I am also fine to leave it out, need some opinion on this)

Sounds good @tlylt

jonahtanjz:
> I am experiencing some weird behavior with the current implementation that I am not sure if it's a bug or 🤯 .... So basically on CI and locally, the current branch should be working, e.g running npm run test:e2e in packages/cli should pass the added tests.
However, when merging the master back to the current branch to fix the merge conflict, then re-running the test will give immediate failures on all added e2e test cases, with little information on what's wrong. Changing the headless = false and checking the browser run, I can see that the page renders for a short moment and just closes immediately. The only two ways to fix this that I discovered are to either remove the pageNav: 2 in index.md of the test_site, or actually add a lot of content into index.md, some of which includes level 2 headings... Seems to suggest that there's a problem with page nav, but no idea why that is the case

Looks like the offending code snippet is the [`data-bs-spy` attribute from Bootstrap`]( https://github.com/MarkBind/markbind/blob/master/packages/core/src/Page/page.njk#L41) for the page nav. Removing that seems to resolve the issue. However, since we are using the Bootstrap scroll spying on the page nav, we cannot remove this. Will have to find a workaround or the root issue causing puppeteer to not work as expected :(

jonahtanjz:
@tlylt 

Found out that the underlying issue is that the Bootstrap's js will throw an error if the page nav is present but there is no header matching the heading level specified in the frontmatter. This causes jest-puppeteer to assume there is an error and closes the browser immediately.

image

Serve the below file and observe the error in the console. 
```md
<frontmatter>
  pageNav: 2
  pageNavTitle: Test Title
</frontmatter>

This is a test
```

This is a separate issue that will require not rendering the `data-bs-*` attributes if there are no headings matching the heading level (i.e if `#mb-page-nav` is not present on the page).

For a workaround on the test, you can either remove the page nav or add a h2 heading to the file and that should resolve the issue for now.

PR #1780, opened 2022-02-16
Implement src attribute for popovers

### LINKED ISSUE #59: Allow src attribute for popovers
e.g. `<popover src="definintions.md#xp">Extreme programming</popover>`

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #59

**Overview of changes:**
Implements the src attribute in popovers, which allows popover contents to be loaded from other files.

1. Static vs dynamic approach
	* See discussion on #59 on using a static approach (similar to Includes) vs dynamic approach (similar to Panels)
2. Behaviour on conflicting slots/attributes
	* Currently, popover contents can be given through 3 ways. I used the following 'priority' of overriding:
		1. `content` slot
		2. `src` attribute
		3. `content` attribute
	* Rationale: For other MarkBind components, slots are taken as the highest 'priority' and an attribute will be overwritten/ignored if the corresponding slot is provided. Though I'm not sure if `src` attribute should override content slots in this case?

**Testing instructions:**

```
<popover header="Content loaded from src" src="{{ baseUrl }}/userGuide/syntax/extra/loadContent.html#fragment">
	Hover
</popover>
```

```
Implement src attribute for popovers

Content can be supplied to popovers through either the content slot, or
content attribute.

Manually providing popover contents may be inconvenient if an author
wants to reference content from elsewhere on the site. 

Let's implement the src attribute for popovers, so that authors can
reference content from source files. This is more convenient for
authors, while helping to ensure consistency in website contents.

Using the src attribute maintains consistency with other components such
as Panels and Includes, which also use src.

The src attributed was implemented with a static approach to avoid
potential positioning issues. In the case that multiple contents are
supplied, the following priority is used: content slot > content
attribute > src attribute
```

---

- [x] No unrelated changes

### DIFF
--- a/docs/userGuide/syntax/popovers.md
+++ b/docs/userGuide/syntax/popovers.md
@@ -51,6 +51,8 @@
   </div>
   <button class="btn btn-secondary">Hover</button>
 </popover>
+<br />
+<br />
 <h4 class="no-index">Content using src</h4>
 <p>
   <popover header="From a HTML file" src="{{ baseUrl }}/userGuide/syntax/extra/loadContent.html#fragment">
--- a/packages/vue-components/README.md
+++ b/packages/vue-components/README.md
@@ -20,8 +20,6 @@ Some custom components and directives are also added for MarkBind's use.
 - Question.vue
 - QOption.vue
 - Quiz.vue
-- Popover.vue 
-- Tooltip.vue 
 - Trigger.vue (built on bootstrap-vue's popovers, modals, tooltips)
 
 ### MarkBind components ported from [Markbind/vue-strap](https://github.com/MarkBind/vue-strap):
@@ -52,6 +50,8 @@ Some custom components and directives are also added for MarkBind's use.
 ### BootstrapVue components included in the bundle
 
 - Modals
+- Popover.vue 
+- Tooltip.vue 
 
 ## Installation
 


tlylt (commented):
Great job!
One slight clarification:
Does the src allows for both a html page URL and a markdown file (just like the include mechanism)?
If so perhaps we could mention both and give one example each.

Also there's a conflict due to the latest PR merge.

PS. one more question in mind:
Just curious, does the importing of content from an external URL leads to a potential XSS attack, or has it already been handled accordingly?

jovyntls:
> Great job! One slight clarification: Does the src allows for both a html page URL and a markdown file (just like the include mechanism)? If so perhaps we could mention both and give one example each.

Yup! Good idea, I'll add it to the docs

> Also there's a conflict due to the latest PR merge.

Fixed, thanks for the catch :) 

> PS. one more question in mind: Just curious, does the importing of content from an external URL leads to a potential XSS attack, or has it already been handled accordingly?

By external URL, do you mean something like `src="{{baseUrl}}/hello.html"`? For this, XSS is possible though there's currently no warning given for users. I'm not sure if this will be a threat though, since these files are usually supplied by the user themselves?

tlylt:
> By external URL, do you mean something like `src="{{baseUrl}}/hello.html"`? For this, XSS is possible though there's currently no warning given for users. I'm not sure if this will be a threat though, since these files are usually supplied by the user themselves?

Ok, I think I had the wrong impression that the src could be any URL to an HTML page on the web but I think it doesn't seem to be the case. Hence maybe it's not that necessary to give any warnings at the moment.

ryoarmanda:
Thanks for the work, @jovyntls ! Review is underway.

> * Currently, popover contents can be given through 3 ways. I used the following 'priority' of overriding:
>		a. `content` slot
>		b. `src` attribute
>		c. `content` attribute
> * Rationale: For other MarkBind components, slots are taken as the highest 'priority' and an attribute will be overwritten/ignored if the corresponding slot is provided. Though I'm not sure if `src` attribute should override content slots in this case?

Let's discuss about the priority list. Just my 2 cents, how about if we put `src` attribute last? This is under the assumption that users would normally do just one of these three, which is a reasonable use-case and what is exampled in the issue.

It's a little bit unintuitive to think `src` will be overridden by `content`, but I like to think that just like a general scoping rule, the closest specified content should win. In this case, it means content that is supplied directly when writing the popover component, which is either through the `content` attribute or the `content` slot. The content specified in `src` is an externally-defined content and will lose to those. If the user wishes to display that however, they can just not write the `content` stuff (which is also reasonable). What do you think?

jovyntls:
@ryoarmanda 
I like the idea of using "scoping" here - I think while src overriding content is more intuitive, using some sort of rule (scoping) would be better than the intuitive approach in case we need to standardise behaviour in the future.

packages/core/src/html/includePanelProcessor.js
```diff
@@ -255,7 +255,99 @@ function processInclude(node, context, pageSources, variableProcessor, renderMd,
...
+
+    if (childContext.hasExceededMaxCallstackSize()) {
+      const error = new CyclicReferenceError(childContext.callStack);
+      logger.error(error);
+      cheerio(node).replaceWith(createErrorNode(node, error));
+    }
+  }
+
+  const attributeSlotElement = cheerio.parseHTML(`<template #content>${actualContent}</template>`, true);
+  node.children = node.children ? attributeSlotElement.concat(node.children) : attributeSlotElement;
```
ryoarmanda, line 342: I think we can abstract this to a function, e.g. `createSlotTemplateNode`, to minimize duplication against a part of `MdAttributeRenderer.processAttributeWithoutOverride`. (Refactoring of the latter function to use the former can be done in a separate work).

I'm not sure that we can put it on `MdAttributeRenderer` too as it doesn't make sense to couple this file to that one, but maybe either on `vueSlotSyntaxProcessor.js` (as it relates to slots), `elements.js` (nice file to put in nodes to use), or any other file you feel is relevant for this.

ryoarmanda (changes requested):
In the meantime, here are some improvements that can be made for your implementation  (the tests and docs reviews are to be done later when the implementation is solid).

Feel free to address this alongside implementing the priority scheme as discussed :)'

P.S. don't forget to sync with master!

packages/core/src/html/includePanelProcessor.js
```diff
@@ -255,7 +255,99 @@ function processInclude(node, context, pageSources, variableProcessor, renderMd,
...
+
+  const fileExistsNode = _getFileExistsNode(node, context, actualFilePath, pageSources);
+  if (fileExistsNode) {
+    return fileExistsNode;
+  }
+
+  // No need to process url contents
+  if (isUrl) {
+    delete node.attribs.src;
+    return node;
```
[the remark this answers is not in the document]
jovyntls: Thank you for the clarification! I was indeed confused about context/node, so this cleared things up.

packages/core/src/html/includePanelProcessor.js
```diff
@@ -255,7 +255,99 @@ function processInclude(node, context, pageSources, variableProcessor, renderMd,
...
+    actualContent = $(hash).html();
+
+    if (actualContent === null) {
+      actualContent = '';
+
+      const error = new Error(`No such segment '${hash}' in file: ${actualFilePath}\n`
+        + `Missing reference in ${context.cwf}`);
+      logger.error(error);
+
+      actualContent = cheerio.html(createErrorNode(node, error));
```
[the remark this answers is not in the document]
jovyntls: I used the same behaviour as `include` - i.e., when the `src` is invalid, it displays the error node:
image

Though on thinking further, I think it's a better idea to log the error, and not display the node instead 😅 
Is this behaviour desirable for `include` too?

jovyntls:
Thank you so much for the detailed look-through @ryoarmanda ! I've made the changes and re-synced with master.

packages/core/src/html/includePanelProcessor.js
```diff
@@ -255,7 +255,101 @@ function processInclude(node, context, pageSources, variableProcessor, renderMd,
...
+
+  // Process sources with or without hash, retrieving and appending
+  // the appropriate children to a wrapped include element
+  if (hash) {
+    const $ = cheerio.load(actualContent);
+    actualContent = $(hash).html();
+
+    if (actualContent === null) {
+      actualContent = '';
+
```
ryoarmanda, line 319: This part is unnecessary anymore since we will do an early return below

docs/userGuide/syntax/popovers.md
```diff
@@ -45,14 +45,25 @@
...
-<br />
-<br />
+<p>
+  <popover effect="scale" header="**Emoji header** :rocket:">
+    <div slot="content">
+      This is a long content...
+    </div>
+    <button class="btn btn-secondary">Hover</button>
+  </popover>
+</p>
```
ryoarmanda, line 55: This change, while good, is a bit unrelated as it is not within the scope of our issue (as this is not regarding to the `src` attribute)

docs/userGuide/syntax/popovers.md
```diff
@@ -74,29 +85,29 @@ This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.
...
-content{{slot_info_trigger}} | `String` | `''` | Popover content, supports MarkDown text.
-placement | `String` | `top` | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.
+\***\*Options\*\***
 
+| Name                         | Type     | Default | Description                                                                                             |
+| ---------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------- |
+| trigger                      | `String` | `hover` | How the Popover is triggered.<br>Supports: `click`, `focus`, `hover`.                                   |
+| header{{slot_info_trigger}}  | `String` | `''`    | Popover header, supports MarkDown text.                                                                 |
+| content{{slot_info_trigger}} | `String` | `''`    | Popover content, supports MarkDown text.                                                                |
+| src                          | `String` |         | The url to the remote page to be loaded as the content of the popover. Both .md and .html are accepted. |
```
ryoarmanda, line 95: We can stylize the file extensions to use inline code like so: `.md` and `.html` are accepted.

Also, we can break the `src` description into multiple lines with `<br>`, just like `placement`

docs/userGuide/syntax/popovers.md
```diff
@@ -74,29 +85,29 @@ This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.
...
+| placement                    | `String` | `top`   | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                             |
 
 <span id="short" class="d-none">
 
 ```html
-Hover over the <trigger for="pop:context-target">keyword</trigger> to see the popover.
+Hover over the <trigger for="pop:context-target">keyword</trigger> to see the
+popover.
```
ryoarmanda, line 102: Unrelated change

docs/userGuide/syntax/popovers.md
```diff
@@ -74,29 +85,29 @@ This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.
...
-
-description :+1:
-
-</div>
+	<div slot="content">
+		description :+1:
+	</div>
 </popover>
 ```
+
```
ryoarmanda, line 110: Unrelated change

packages/cli/test/functional/test_site/testPopovers.md
```diff
@@ -0,0 +1,67 @@
+**Popover with attributes**
```
ryoarmanda, line 1: We can add more tests on:
- ensuring the priority scheme discussed is followed
- ensuring that if `src` is an url it will error
- and other tests that you have come up with

docs/userGuide/syntax/popovers.md
```diff
@@ -74,29 +85,29 @@ This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.
...
 
-****Options****
-
-Name | Type | Default | Description
----- | ---- | ------- | ------
-trigger | `String` |	`hover`	| How the Popover is triggered.<br>Supports: `click`, `focus`, `hover`.
-header{{slot_info_trigger}} | `String` | `''` | Popover header, supports MarkDown text.
-content{{slot_info_trigger}} | `String` | `''` | Popover content, supports MarkDown text.
-placement | `String` | `top` | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.
+\***\*Options\*\***
```
ryoarmanda, line 88: This change breaks the styling of the `Options` text

docs/userGuide/syntax/popovers.md
```diff
@@ -74,29 +85,29 @@ This is the same <trigger for="pop:trigger_id">trigger</trigger> as last one.
...
 
+| Name                         | Type     | Default | Description                                                                                             |
+| ---------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------- |
+| trigger                      | `String` | `hover` | How the Popover is triggered.<br>Supports: `click`, `focus`, `hover`.                                   |
+| header{{slot_info_trigger}}  | `String` | `''`    | Popover header, supports MarkDown text.                                                                 |
+| content{{slot_info_trigger}} | `String` | `''`    | Popover content, supports MarkDown text.                                                                |
+| src                          | `String` |         | The url to the remote page to be loaded as the content of the popover. Both .md and .html are accepted. |
+| placement                    | `String` | `top`   | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                             |
```
ryoarmanda, line 97: We can add a short tipbox explaining the priority scheme, but also note that normally authors only use one of the different ways

ryoarmanda (commented):
Thanks for addressing the review! The implementation is pretty solid (with one very minor change below) and all that's left is coming up with the tests and cleaning up the documentation 😄

packages/core/src/html/includePanelProcessor.js
jovyntls: I missed this, thanks for the catch!

docs/userGuide/syntax/popovers.md
jovyntls: I see! I'll revert it

jovyntls:
Thanks again @ryoarmanda for the detailed lookthrough! I've made the changes mentioned, and also made an additional change to packages/vue-components/README.md since I realised it hadn't been updated yet.

tlylt:
Just checking, does this part in `coreesrc/html/linkProcessor.js` need to be updated as well?
image

jovyntls:
@tlylt Thanks for the catch!

You're right, `defaultTagLinkMap` is used in `convertRelativeLinks` so popovers should be included as well, otherwise relative links won't work for the `src` attribute. I originally thought it wasn't needed since `validateIntraLink` returns immediately, didn't realise that `convertRelativeLinks` uses this as well.

packages/vue-components/README.md
```diff
@@ -20,6 +20,8 @@ Some custom components and directives are also added for MarkBind's use.
 - Question.vue
 - QOption.vue
 - Quiz.vue
+- Popover.vue 
+- Tooltip.vue
```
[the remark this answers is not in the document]
jovyntls: Ok! In that case I'll create an issue to keep track of all the popover docs updates

ryoarmanda (approved):
LGTM! 🎉

ryoarmanda:
Hi @jovyntls , before I merge the changes, let's add more information to the commit message, as this is a considerable popover enhancement.

jovyntls:
Thanks for the reminder @ryoarmanda ! I've updated the commit message to include more information.

PR #1787, opened 2022-02-23
Fix tipBox content overflow

tlylt:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [x] Bug fix
- [ ] Feature addition or enhancement
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Prevent overflow when tipBox contains content that cannot fit within the box.
- fix is based on https://stackoverflow.com/questions/36230944/prevent-flex-items-from-overflowing-a-container

Before (on this page https://markbind.org/userGuide/siteJsonFile.html):
overflow1
overflow2
After (on this page https://deploy-preview-1787--markbind-master.netlify.app/userguide/sitejsonfile of the preview build):
image

**Anything you'd like to highlight / discuss:**
nil

**Testing instructions:**
visit user guide site.json file page
- inspect and change the dimension to iPhone SE or one of the mobile phones
- the content should now be within the box

**Proposed commit message: (wrap lines at 72 characters)**
Fix tipBox content overflow with min-width

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [ ] Linked all related issues
- [x] No unrelated changes

Issue #1741, opened 2022-01-29
Fixed headers should span only the viewport space


ang-zeyu:
Directly Related: #1731 (fixing this would fix the header issue here, but likely we should still try to scale embed as well)
Can also be observed here https://markbind.org/userGuide/siteJsonFile.html with a mobile device

Somewhat related: #1560 

Labelling this `toInvestigate` as I haven't really considered if there are any other user implications of `width: 100vw`

ang-zeyu:
Fixed in #1939

PR #1797, opened 2022-02-27
Fix embed overflow issues 

### LINKED ISSUE #1731: Embeds do not scale, causing overflow on mobile
Windows

**MarkBind Version:**
v3.1.0

From default template (via `markbind init`), serve the site and observe it on mobile dimensions (e.g. through Chrome's device emulation).

Everything is scaled appropriately

Embeds do not scale along, causing horizontal overflow, which is not ideal on mobile environment. The overflow negatively impacts the navbar, pushing the search bar out of the viewport and thus scrolling is required to access it.

This is largely due to the iframes that contain the embedded media have fixed width and height (default is 640x390). Making responsive iframe container is a little finicky, but [possible](http://jsfiddle.net/omarjuvera/8zkunqxy/2/) (found in [StackOverflow](https://stackoverflow.com/questions/17838607/making-an-iframe-responsive)). I think it's good to explore whether this can be done on our end, be it with the default or custom dimensions.

image

image

ong6:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #1731 

**Overview of changes:**

Implemented the CSS mentioned in issue #1731 which seemed to fix the problems.

Before: 
image

After:
image

**Testing instructions:**

View the site on mobile view

Fix embed scaling issues.

---

- [x] No unrelated changes


kaixin-hc:
LGTM, it works as described on the netlify preview

packages/core-web/src/styles/markbind.css
```diff
@@ -333,3 +333,14 @@ h5,
...
+.block-embed {
+    position: relative;
+    padding-bottom: 56.25%; /* 16:9 */
+}
+
+.block-embed iframe {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+}
```
ryoarmanda, line 346: We should add `top: 0` and `left: 0` to the absolute positioning, because the default might not be 0 as you thought (refer [here](https://stackoverflow.com/questions/19968979/what-are-the-default-top-left-botton-or-right-values-when-positionabsolute-is))

packages/core-web/src/styles/markbind.css
```diff
@@ -333,3 +333,14 @@ h5,
 h6 {
     margin-bottom: 1rem;
 }
+
+.block-embed {
+    position: relative;
+    padding-bottom: 56.25%; /* 16:9 */
+}
```
ryoarmanda, line 340: We can also add the `overflow: hidden` and `height: 0` for robustness like the sample fiddle in the issue (reasons why it's needed are outlined [here](https://www.benmarshall.me/responsive-iframes/) at Things to Remember)

ryoarmanda (commented):
Thanks for the work, @ong6 ! The fix should cover the embed YouTube case, but let's make it cover our entire use cases: YouTube and PowerPoint embeds.

The particular `56.25%` value suits well with the YouTube embed as the set dimension is `640x390`, which is `16:9` (see that `100 / (16 / 9) = 56.25`). However, PowerPoint embeds set its dimension to `610x481`, which requires a different percentage, but can be easily calculated by hand like above and put the approximate into the styling.

Along with the rules that you have written, let's write rules on embed service-specific classes to set different `padding-bottom` percentages on different services.

packages/core-web/src/styles/markbind.css
ong6: ohh interesting i will add it in thanks for the tip!

packages/core-web/src/styles/markbind.css
ong6: > The particular `56.25%` value suits well with the YouTube embed as the set dimension is `640x390`, which is `16:9` (see that `100 / (16 / 9) = 56.25`). However, PowerPoint embeds set its dimension to `610x481`, which requires a different percentage, but can be easily calculated by hand like above and put the approximate into the styling.

`640x390` is not 16:9 though? The percentage calculations ended up being just a rough estimate on how to reduce the blank space in the view area. 

e.g. 
image
vs 
image

I'm also not sure how the original author decided on each width/ height value also, but would it be ok to make changes to them?

ong6: yep i've made the changes to scale based on the width/height of the container

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
```diff
@@ -46,6 +46,14 @@ class VideoServiceBase {
     let escapedServiceName = this.env.md.utils.escapeHtml(this.name);
     containerClassNames.push(this.env.options.serviceClassPrefix + escapedServiceName);
 
+    let containerStyles = ["position: relative;", "height: 0;"];
+    // To set a specific style, add a style prop to the getDefaultOptions of any Service
+    if (this.options.style) {
```
ang-zeyu, line 51: Hmm what's this branch for? Couldn't find any use cases for it currently (if so let's remove it), but perhaps it could be used to solve the earlier issue.

ang-zeyu (commented):
Lgtm overall 👍, I think we'll just need to deal with the ppt exception:

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
```diff
@@ -46,6 +46,14 @@ class VideoServiceBase {
...
+      containerStyles.push(this.options.style);
+    } else {
+      containerStyles.push(`padding-bottom: ${this.options.height / this.options.width * 100}%`);
+    }
+
     let iframeAttributeList = [];
     iframeAttributeList.push([ "type", "text/html" ]);
     iframeAttributeList.push([ "src", this.getFilteredVideoUrl(videoID) ]);
```
ang-zeyu, line 59: One more tiny nit: the `width` / `height` additions to `iframeAttributeList` can probably be removed since you are scaling to `100%/100%`

packages/core-web/src/styles/markbind.css
```diff
@@ -333,3 +333,11 @@ h5,
 h6 {
     margin-bottom: 1rem;
 }
+
+.block-embed iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
```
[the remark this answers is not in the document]
ong6: ok, i'll look into setting a min here

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ong6: This is used to set the aspect ratio of the embed and then add padding to the bottom according to the aspect ratio. Without it, there would be no space for the embed to be displayed. So I cant remove this else nothing will be shown.

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ong6: Do you mean the first if (...) statement? If so then there is currently no use cases, but it could be used to custom style components if needed. I will see if its necessary.

packages/core-web/src/styles/markbind.css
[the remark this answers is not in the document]
ong6: upon doing more research into this issue, it seems as though the issue is caused by another component in the page shown in issue #1813. Once that overflow is fixed, the format starts working as intended.

image

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ang-zeyu: yup

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ang-zeyu: Untitled

Note the `padding-bottom` trick works by assuming the child element fills 100% of the container width (achieved by the stylesheet), therefore you can see it as the "height" itself.

Setting a fixed `width` / `height` is unnecessary in the iframe as it is overriden by the stylesheet.

Untitled

packages/core-web/src/styles/markbind.css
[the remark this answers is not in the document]
ang-zeyu: That issue is separate;

Try these settings (320px is a rather [standard](https://accessally.com/blog/mobile-responsive-screen-sizes/#:~:text=For%20example%2C%20the%20smallest%20screen,the%20Galaxy%20phones%20and%20tablets) "smallest screen size" we usually take into account):
Untitled


Note the text in the ppt ("we want to make logic a singleton...") should not be too blurry as well.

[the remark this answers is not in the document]
ang-zeyu: > upon doing more research into this issue, it seems as though the issue is caused by another component in the page shown in issue #1813. Once that overflow is fixed, the format starts working as intended.

(see https://github.com/MarkBind/markbind/pull/1787)

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ong6: Oh, sorry! My bad I completely misread your previous post, thought you wanted me to remove the container styles! 

OK anyways it seems like even though I can scale it properly, there is still blurriness, do you have any idea why this might be the case? 

width 320:
image

ang-zeyu: You'll have to dig into the ppt embed if you are interested; I'm unfortunately not familiar with its implementation details as well. 😐
Though the basic idea would be that the minimum dimensions of the slides decided internally is rather lax (in that the slide almost fills the embed viewport but readability is compromised)

ryoarmanda:
@ong6 @ang-zeyu 

After discussion with @damithc (may also check our blocker with PowerPoint embeds above), we thought that it is okay for the PowerPoint embed to be of reduced quality in mobile, as it is not in high usage now. The assumption is that users visiting the site are not limited to do so in just one device, so they can switch on larger screens.

I guess with that in mind, we have some options, with varying pros and cons:
- Make YouTube embeds responsive, but keep PowerPoint embeds fixed-size
  * Simple. Add a method that checks whether responsive styling should be applied. The `ServiceBase` can have it false as default and `YouTubeService` overrides it as true.
  * Not covering our use cases. Overflow is still possible if there is a page with PowerPoint embeds. (but low impact as its not highly used)
- Make PowerPoint embeds responsive too, but add accessibility warning on smaller screens just before the embed container
  * Overflows are mitigated on all cases and directed users to view small embeds on larger devices instead. (is landscape possible too?)
  * A bit involved, also adds extra HTML output. Likely have to write a `<script>` to detect window size and toggle the accessibility warning display (no Vue components to help here). So, the HTML output for the embed would be added with both the accessibility warning text element and the script.

If you have any other ideas feel free to explore them too

tlylt:
> * Make PowerPoint embeds responsive too, but add accessibility warning on smaller screens just before the embed container
>   
>   * Overflows are mitigated on all cases and directed users to view small embeds on larger devices instead. (is landscape possible too?)
>   * A bit involved, also adds extra HTML output. Likely have to write a `<script>` to detect window size and toggle the accessibility warning display (no Vue components to help here). So, the HTML output for the embed would be added with both the accessibility warning text element and the script.
> 
> If you have any other ideas feel free to explore them too

Perhaps for PowerPoint, we can use overflow-x: scroll to minimally tackle the overflow issue

ang-zeyu:
> width 320
> image

👀 I think there may be some miscommunication here; Do let me know if there are any particular parts I was unclear so I can improve on that.

`320px` is refering only to the viewport size (as per my reply [here](https://github.com/MarkBind/markbind/pull/1797#discussion_r823370571)), which is only meant for **reproducing** the issue given the prior difficulty @ong6 had doing so.

> may also check our blocker with PowerPoint embeds above

> OK anyways it seems like even though I can scale it properly, there is still blurriness, do you have any idea why this might be the case?

To clarify this part, my reply here isn't relevant, this seemed like an exploratory question on why the embed looks blurry with `320px` dimensions set in particular (per how it looks in the screenshot https://github.com/MarkBind/markbind/pull/1797#discussion_r823842739). That's out of our control, as the minimum dimensions of things rendered inside the ppt plugin are decided by the plugin. Setting too small an iframe width (which we can control) would cram too much things into a fixed width as you might expect, which is counterproductive.

---

> Make YouTube embeds responsive, but keep PowerPoint embeds fixed-size

Yes, this was the solution I'm suggesting (https://github.com/MarkBind/markbind/pull/1797#discussion_r822287264). Though, we can still see if we can make it responsive / get the best of both worlds by using minimum widths / heights. (fixed width/height would be fine as well if not)

---

> Perhaps for PowerPoint, we can use overflow-x: scroll to minimally tackle the overflow issue

That's another solution, although, we'd likely have difficulties with scroll/click propagation with the embed itself.

---

> Make PowerPoint embeds responsive too, but add accessibility warning on smaller screens just before the embed container

Let's avoid this for the reasons you mentioned; Its a little much work for little return.

ang-zeyu:
> Not covering our use cases. Overflow is still possible if there is a page with PowerPoint embeds. (but low impact as its not highly used)

Btw, linking this again in case you guys missed it @ong6 @damithc @ryoarmanda 

https://github.com/MarkBind/markbind/issues/1741

There are 2 parts to this overflow issue, both should be fixed. The latter one I've just linked (which makes the header look really out of place) should remove your concerns with this for the most part.

We'd be left with a page that has content flush along the viewport sides all the way except the powerpoint embed (which would then be accessible simply by zooming out / scrolling to the right).

ong6:
@ang-zeyu 
Ahh ok there might have been some miss com here, I thought there was a need to fix the blurriness issue as well. I will just push the fix for the scaling then!

@ryoarmanda 
Thanks for checking for me! I think I will try the overflow scroll like what @tlylt mentioned for those with screen sizes smaller than 360px (the point which it starts to get blurry).

ang-zeyu:
> @ang-zeyu Ahh ok there might have been some miss com here, I thought there was a need to fix the blurriness issue as well. I will just push the fix for the scaling then!
> 
> @ryoarmanda Thanks for checking for me! I think I will try the overflow scroll like what @tlylt mentioned for those with screen sizes smaller than 360px (the point which it starts to get blurry).

I think there's still some miscomm here 😕.

The fix for blurriness should come **naturally** (I think it should be fixed, in case it isn't clear) as a part of either of these solutions:
- minimum width / height (preserve responsiveness on larger screens, enable usability on smaller ones)
- fixed width / height (i.e. no change specifically for the ppt embed)
- @tlylt's suggested overflow -- would refer you again to my comment below as well, but its just speculation, you're welcome to experiment : )
  > That's another solution, although, we'd likely have difficulties with scroll/click propagation with the embed itself.
   - also note this is an *addition* on top of the above 2 solutions (you can't have overflow without the embed exceeding the viewport/container width)

Just for clarity as well my original issue wasn't with blurriness (as long as its somewhat readable, that's fine): https://github.com/MarkBind/markbind/pull/1797#discussion_r823370571

Notice the buttons are out of the embed, which makes this **unusable**.

Fixing the buttons issue however entails making the embed large enough in any solution, which also implies fixing the blurriness issue as side effect. (two birds one stone)

ryoarmanda:
Thanks for the clarification @ang-zeyu 👍 I was thinking that as one of the goals of the issue is to mitigate the overflow in mobile, but then introducing scaling to device width makes the PowerPoint embed blurry and unreadable as seen from your example (I see the blurriness is a bit relevant, because it now becomes a fix that takes the expense of other valuable things that we can't fix), so we are "blocked" on fixing overflow for powerpoint, unsure on whether to proceed with the responsiveness or keep it fixed-size. Just helping to nudge things along by discussing with @damithc , and giving some options for @ong6 to explore (thanks for your suggestions too!)

ang-zeyu:
> (may also check our blocker with PowerPoint embeds above)
> so we are "blocked" on fixing overflow for powerpoint

Thanks for clarifying and nudging things along. 👍 I misinterpreted this as "blocked on implementation details."

ang-zeyu:
> @ang-zeyu Ahh ok there might have been some miss com here, I thought there was a need to fix the blurriness issue as well. I will just push the fix for the scaling then!
> 
> @ryoarmanda Thanks for checking for me! I think I will try the overflow scroll like what @tlylt mentioned for those with screen sizes smaller than 360px (the point which it starts to get blurry).

To be clear as well, this comment still concerns me. Referring to my earlier reply, https://github.com/MarkBind/markbind/pull/1797#issuecomment-1079952821, note that the solution you're attempting *implicitly means* fixing blurriness, and includes either the earlier mentioned min-width / fixed width solutions. @ong6

tlylt (dismissed):
Overflow issues no longer appear on smaller screens. LGTM

packages/core-web/src/styles/markbind.css
```diff
@@ -341,3 +341,14 @@ h6 {
...
 }
+
+.block-embed-service-powerpoint {
+    position: relative;
+    padding-bottom: 78%;
+}
+
+@media screen and (max-width: 350px) {
+    .block-embed-service-powerpoint {
+        padding-bottom: 95%;
```
ang-zeyu, line 333: There is still some clipping (on the right side) of the slides at `320px` =\\
It also changes the aspect ratio discussed earlier here https://github.com/MarkBind/markbind/pull/1797#pullrequestreview-899134740 (I'm assuming we still want to keep this).

Consider the case the embed is even more deeply nested (say within 3 `<panel><panel><panel>`s), hardcoding the media query likely won't work as well, as it is still dependent on the parent width which has become extremely small (not the viewport's).

I really don't think there is any solution right now other from
1. fixed: easier - just means preserving the behaviour before this PR for powerpoint embeds specifically
   - you can also take some liberty in slightly reducing the values (preserving AR) to reduce overflow a little
2. minimum (harder) width
It's just physically impossible to fit even "properly designed" slides into too small and a finite number of pixels, and the powerpoint/prezi embed (unfortunately) isn't designed like the other embeds to be fully responsive. (which, would be extremely difficult given slides themselves can contain a wide range of contents)

Scrolling would be perfect, but this dosen't seem like something implementable from outside of the embed by us. (due to the embed already capturing mouse events) (again, part speculation)

tlylt: To clarify on the point of capturing mouse events for scrolling:
- Are you referring to the fact that to scroll "on" the slide itself may not work because it is being captured?

Adding a "overflow-x: scroll" in the container div like this:
image

will give the following effect:
- cannot scroll directly on the slides
- can scroll via the scroll bar
gif

ang-zeyu: Thanks for testing, yes, something like this.

ong6: @ang-zeyu Hmm it seems like there is a slight difference from what I'm seeing locally and what the Netlify preview is showing, just for reference:

Netlify:
image

Locally: 
image

I'm not sure which one will actually be represented on the website once its actually merged, but am I correct to say that the one I see locally would be ideal?

ong6: @ang-zeyu Sorry if I'm missing something but, I'm still not quite sure what you mean in point 2 where you set a min-width. From what I've seen, adding a min-width doesn't help at all. 

Also what do you mean by point 1? Isn't the behavior before the PR worse than hardcoding the media query. I think I can just adjust the aspect ratio (by using the padding) for the screens below 350px and that will be "good enough" for like 99% of the people.

ang-zeyu: > @ang-zeyu Sorry if I'm missing something but, I'm still not quite sure what you mean in point 2 where you set a min-width. From what I've seen, adding a min-width doesn't help at all.

Yes, it won't work as-is (per what I'm seeing in 7194a2c).

You'll need to consider how exactly the `padding-bottom` trick works. (by setting the height as a percentage of the _parent_ width) To introduce a minimum width therefore (while preserving aspect ratio) -- you need an additional parent container. (instead of leaving it up to whatever parent container is there)

> Also what do you mean by point 1? Isn't the behavior before the PR worse than hardcoding the media query. I think I can just adjust the aspect ratio (by using the padding) for the screens below 350px and that will be "good enough" for like 99% of the people.

The solution is as discussed [here](https://github.com/MarkBind/markbind/pull/1797#issuecomment-1079230382).

There was some clipping on the right side of the slides before 7194a2c.

> "good enough" for like 99% of the people.

> Consider the case the embed is even more deeply nested (say within 3 `<panel><panel><panel>`s), hardcoding the media query likely won't work as well, as it is still dependent on the parent width which has become extremely small (not the viewport's).

Do try to understand this part as well. The issue is *not* viewport dependent.

---

In case there's still any confusion, I've been suggesting right from the first comment [here](https://github.com/MarkBind/markbind/pull/1797#discussion_r822287264) we don't sacrifice usability on mobile, and allow overflow.

The same point is also mentioned by @ryoarmanda here https://github.com/MarkBind/markbind/pull/1797#issuecomment-1079230382.

I feel really strongly about this for a couple of reasons:
- It seems really strange that a content-oriented static site generator for education websites would compromise on readability on specific devices
- The external onedrive link (bottom right button in the embed) is even less usable on mobile than our previous behaviour before this PR. Were there a reasonable alternative (such as youtube's fullscreen mode), this would be different.
- > Not covering our use cases. Overflow is still possible if there is a page with PowerPoint embeds. (but low impact as its not highly used)

   As mentioned by @ryoarmanda here as well, overflow is of very low impact since powerpoint embeds are seldom used. In which case, I feel the case for usability is stronger.

... This is just my opinion. If there is consensus otherwise (which I have been repeatedly trying to get confirmation on thus far), I am *amenable*. (do post an update here as I am out of the loop)

tlylt: Hi @ang-zeyu, I think @ong6 and I had quite a few discussions over this PR previously, about how to create a good enough solution. However, we were really getting confused with the whole conversation (which we cross-reference earlier comments here and there 🤔) within this PR thread and some of the suggestions that we tried didn't really work out (e.g maybe works for Youtube and Box, but not for slides) as expected or as suggested.

Wonder if you can give any consolidated advice on how to proceed and what are the specific deliverables that we can aim to achieve for now and what can be set as a future TODO for a separate PR? Thanks!

ang-zeyu: I think something like this:
- #1741 - this has been fixed as a "freebie" with the sticky header PR. This was just a point that a huge part of the overflow problem you see here is unrelated to embeds specifically (e.g. fixed-size images). Hence why I'm really not concerned with overflow as much here too. (see latest netlify [preview](https://markbind-master.netlify.app/userguide/formattingcontents#embeds))
- 2 possible solutions we can do right now:
   - just keep the existing behaviour, styling specifically for ppt embeds, scale the rest
   - Keep the best of both worlds. That is, while *keeping the aspect ratio*, scale the ppt embed after some minimum width threshold (we can just use the existing fixed width as one), but otherwise, the embed should minimally be kept at that width. Hint in the first 2 paragraphs in this [comment](https://github.com/MarkBind/markbind/pull/1797#discussion_r846779449). I mentioned this is harder, but not much. Its a fairly interesting exercise in css and I really encourage trying it out.
- Future: originally, something along the lines of exploring the embed scrolling idea further. But with the sticky header PR's UX change, this is even less necessary, although that makes its own [compromises](https://github.com/MarkBind/markbind/pull/1939#discussion_r895126866). Wherever possible, we can aim to scale things to 100% container width (i.e. not something like the ppt embed that scales poorly, but for example images, or in the case of this PR youtube embeds), so page overflow becomes even more seldom. But this is really up for discussion individually.

Fwiw, I found this excerpt from mdn some time ago. I think it helps explain my view on usabililty v "looking nicer" with much better generality, which might have been more helpful:

> The problem with data loss is that you might not notice. Website visitors may not notice.
> 
> \- https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Overflowing_content#css_tries_to_avoid_data_loss

ong6: > * just keep the existing behaviour, styling specifically for ppt embeds, scale the rest

Hi @ang-zeyu thanks for giving the advice, I think I'll just go with this simpler approach and make another issue inviting people to take up the CSS challenge. Will try to push one PR tmr to clean up and make final changes.

ang-zeyu: Thanks for updating this!

> just keep the existing behaviour, styling specifically for ppt embeds, scale the rest

Existing behaviour meaning keeping the fixed width and height before this PR, I can't see any difference right now with the ppt embed from the other ones (let me know if I'm missing something!). Which is in line with the rest of the point on not compromising usability / data loss.

One simple way to do it is using the existing `block-embed-service-powerpoint` class to target ppt embeds only.

Some other nits below:

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/PowerPointOnlineService.js
```diff
@@ -5,15 +5,15 @@ const VideoServiceBase = require("./VideoServiceBase");
...
+    return { width: 500, height: 375 };
   }
 
   extractVideoID(reference) {
     return reference;
   }
 
   getVideoUrl(serviceUrl) {
-    return `${serviceUrl}&action=embedview&wdAr=1.3333333333333333`;
+    return `${serviceUrl}`;
```
ang-zeyu, line 16: Hmm what's the removal for? I would think the `wdAr` parameter is important for the powerpoint service to render things correctly, if the default isn't already `1.333...`. If it's not necessary, let's remove the change.

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/PowerPointOnlineService.js
```diff
@@ -5,15 +5,15 @@ const VideoServiceBase = require("./VideoServiceBase");
 class PowerPointOnlineService extends VideoServiceBase {
 
   getDefaultOptions() {
-    return {width: 610, height: 481};
+    return { width: 500, height: 375 };
```
ang-zeyu, line 8: Let's keep this change to https://github.com/MarkBind/markbind/issues/1952 too.

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
```diff
@@ -46,24 +46,22 @@ class VideoServiceBase {
...
-
-    if (this.env.options.outputPlayerSize === true) {
-      if (this.options.width !== undefined && this.options.width !== null) {
-        iframeAttributeList.push([ "width", this.options.width ]);
-      }
-      if (this.options.height !== undefined && this.options.height !== null) {
-        iframeAttributeList.push([ "height", this.options.height ]);
-      }
-    }
+    iframeAttributeList.push([ "scrolling", "yes" ]);
```
ang-zeyu, line 58: Missed out a leftover from @tlylt's suggestion, cmiiw.

One more strange empty line below I think we could remove too.

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/PowerPointOnlineService.js
ong6: Actually from my testing it seems as if the wdAr does not actually do anything to the aspect ratio of the PowerPoint, perhaps its due to the way things are implemented (but im not entirely sure why it doesn't affect anything). 

That's why I made the decision to remove it, since keeping it around did nothing anyways.

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/VideoServiceBase.js
ong6: yep, i missed this removing now! thanks

packages/core-web/src/styles/markbind.css
ong6: Ok i have reverted to the original, there is no easy way to do this without completely removing the block-embed css style from the ppt embed.

ang-zeyu: > there way no easy way

This is looks simple enough too. But fwiw I was thinking of these 2 ways off the top of my head, the second being the simplest:
- you can rely on the css cascade to override the `block-embed` styles, e.g.
   ```css
  .block-embed-powerpoint {
      overflow-x: visible;
      position: static;
      width: auto;
  }
   ```
- That's a little verbose; Alternatively you could simply add a [` :not`](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)
   ```css
  .block-embed:not(.block-embed-powerpoint) {
      overflow-x: auto;
      position: relative;
      width: 100%;
  }
   ```

packages/core/src/lib/markdown-it/plugins/markdown-it-block-embed/services/PowerPointOnlineService.js
ang-zeyu: Got it, while it can be tempting to clean these sort of things up, let's make sure we understand the original rationale behind them first.

I found this as well, giving a very brief description https://superuser.com/questions/1193572/how-to-run-powerpoint-off-onedrive-in-browser-in-loop. The documentation link is broken though, so not too helpful at the moment 😷

ang-zeyu (approved):
Lgtm 👍

packages/core-web/src/styles/markbind.css
ong6: ahh thanks for the info, i really need to work on my css 😄

PR #1807, opened 2022-03-07
Add <tree> component for folder structure visualisations

### LINKED ISSUE #1660: Support a way to visualize folder structures
In technical documentations, often we need to describe a folder structure, to explain where to put certain files for example.
image

It is nice if we can provide a syntax that MarkBind can convert to a nice visual.
More generally, it can even be used to explain other types of tree-like structures.

PlantUML doesn't seem to support folder visualizations.
https://jaredforsyth.com/treed/ seem to doing something along these lines but not sure if it can fit MarkBind.

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1660

Provides a convenient way to create lightweight tree-like textual output, suitable for folder structure visualisations.
Syntax:
```pre
<tree>
Edit me to generate
  a
    nice
      tree
        diagram!
        :)
</tree>
<tree>
Edit me to generate
  - a
    - nice
    - tree
        - diagram!
        - :)
</tree>
```
To
```pre
Edit me to generate
└── a
    └── nice
        └── tree
            ├── diagram!
            └── :)
```
More details in UG's [tree section](https://deploy-preview-1807--markbind-master.netlify.app/userguide/components/imagesanddiagrams#tree)

A summary of the work done:
- feature development in `core/src/plugins/default/markbind-plugin-tree.js`
- unit test in `core/test/unit/plugins/default/tree.test.js`
- functional test in `packages/cli/test/functional/test_site/testTree.md`
- UG inclusion under Using Component - Images & Diagrams - Tree
- Migration of the existing folder structure code in UG over to use the new tree syntax, mainly in Reusing Contents
- Update tests

- Initially thought of making a Vue component, but went with making a default plugin instead (since the feature itself is quite lightweight)
- The currently supported syntax uses spaces + unordered list bullets for indication of indent/level, I am not sure whether I should also support (chop off) ordered list bullets (basically, numbers like 1., 2. etc).
- The support for inline markdown/HTML elements works for most syntax, but not for tooltip/popover
```pre
<tooltip id="tt:trigger_id" content="This tooltip triggered by a trigger"></tooltip>

<tree>
C:/course/
  textbook/
    <tooltip content="hello">index.md</tooltip>
    <trigger for="tt:trigger_id">via trigger</trigger>
</tree>
```
Testing the above code will give:
```html
<div class="tree">C:/course/
├── <span data-mb-component-type="tooltip" tabindex="0" content="hello" class="trigger"><!----> <!----> index.md</span>
└── <span tabindex="0" class="trigger"><!----> via trigger</span></div>
```

image

Notice the white space added in front of the words "index.md" and "via trigger". While the direct tooltip approach does not work on hover, the trigger approach works. Not too sure if this is a bug with the tooltip/trigger element or is it due to the newly added tree component.

**Testing instructions:**
Tests cases are added in
- `packages/cli/test/functional/test_site/testTree.md`
- `packages/core/test/unit/plugins/default/tree.test.js`

User guide section
- [tree](https://deploy-preview-1807--markbind-master.netlify.app/userguide/components/imagesanddiagrams#tree)
- see some migration of the folder structure code in this [page](https://deploy-preview-1807--markbind-master.netlify.app/userguide/reusingcontents#include-inside-an-included-file)

Add <tree> component for folder structure visualisations

It is tedious to manually create tree-like structure diagrams.

Let's add a <tree> component to conveniently convert
an indented list of items into tree-like textual output.

This is especially useful for file system visualisations.

---

- [x] No unrelated changes


packages/core/src/plugins/default/markbind-plugin-tree.css
```diff
@@ -0,0 +1,6 @@
+.tree {
+    white-space: pre;
+    font-family: "Fira Code", "Fira Mono", monospace;
+    font-size: 0.9375rem;
+    line-height: 1.5;
+}
```
jonahtanjz, line 6: Looks like the `tree` can overflow the page similar to the embed issue:

image

Maybe can try adding `overflow-x: auto`?

Relevant code:
```.html
<tree>
Edit me to generate
  a
    nice
      tree
        diagram!
        :)
  b
    c
      d
        e
          f
            g
              h
                a very long file name
</tree>
```

docs/userGuide/syntax/tree.md
```diff
@@ -0,0 +1,115 @@
...
+    !!Underline with {text=danger}!!{.text-danger}
+    <span class="text-danger">RED</span>
+  __reading.md__
+  ++site.json++
+  ==hello==
+  :construction:
+  :fas-file-code:
+  :)
+</tree>
+<tooltip id="tt:trigger_id" content="This tooltip triggered by a trigger"></tooltip>
```
jonahtanjz, line 82: Is there supposed to be a tooltip example? Looks like it is missing the content and not showing up in the code output.

jonahtanjz (commented):
Very nicely done @tlylt 👍 Tried it out and so far it is working well :)

Just a quick review on the design and documentation first.

jonahtanjz:
> Notice the white space added in front of the words "index.md" and "via trigger". While the direct tooltip approach does not work on hover, the trigger approach works. Not too sure if this is a bug with the tooltip/trigger element or is it due to the newly added tree component.
> ```.html
> <div class="tree">C:/course/
> ├── <span data-mb-component-type="tooltip" tabindex="0" content="hello" class="trigger">  > index.md</span>
> └── <span tabindex="0" class="trigger"> via trigger</span></div>
> ```

Seems like an issue with the way the tree processes the node? Notice that the `content="hello"` attribute is still there on the HTML output which means the [`tooltip processor`](https://github.com/MarkBind/markbind/blob/9e950bc14355b2622f3112335bdd3c1bd228dcd2/packages/core/src/html/MdAttributeRenderer.js#L52) did not get to change the `content` into a `<slot>`.

If we try putting the tree code manually with the tooltip, it doesn't seem to have an issue.
```.html
<div class="tree">
  C:/course/
  └── textbook/
      └── <tooltip content="hello">index.md</tooltip>
</div>
```
image

Not entirely sure what is causing this but somehow after the `tree` plugin processes the code, the `tooltip` does not get processed.

tlylt:
> If we try putting the tree code manually with the tooltip, it doesn't seem to have an issue.
> 
> ```
> <div class="tree">
>   C:/course/
>   └── textbook/
>       └── <tooltip content="hello">index.md</tooltip>
> </div>
> ```
> 
> image
> 
> Not entirely sure what is causing this but somehow after the `tree` plugin processes the code, the `tooltip` does not get processed.

It seems like the problem of empty spaces added in front of the word 'index.md' is a separate issue. Is this a bug with the tooltip component? 🤔

jonahtanjz:
> It seems like the problem of empty spaces added in front of the word 'index.md' is a separate issue. Is this a bug with the tooltip component? 🤔

Seems like the the usual `tooltip` component will put in a white space before the text but this is usually ignored/removed 

image

However, since the `tree` component has the css rule `white-space: pre`, white spaces will be preserved.

I guess we can remove the white space generated from the `tooltip`/`trigger` components if possible.

tlylt:
> Seems like the the usual `tooltip` component will put in a white space before the text but this is usually ignored/removed with `white-space: normal` (default)

Seems like this is indeed the issue. Curious how did you find that out 😅? I was looking over the code for the Tooltip/Trigger Vue components and couldn't find something similar to that effect.

jonahtanjz:
> Seems like this is indeed the issue. Curious how did you find that out 😅? I was looking over the code for the Tooltip/Trigger Vue components and couldn't find something similar to that effect.

I noticed the white space by looking at the HTML code output on the browser. Since the `Tooltip`/`Trigger` component rely on the [`slot` component](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/Tooltip.vue#L17) for the content, which is transformed using the [`processAttributeWithoutOverride`](https://github.com/MarkBind/markbind/blob/9e950bc14355b2622f3112335bdd3c1bd228dcd2/packages/core/src/html/MdAttributeRenderer.js#L26) function, my guess is that the space is introduced during the processing of attributes here.

jonahtanjz (commented):
Other than the `tooltip`/`trigger` issue, I think this PR is good to merge.

Might also want to add a note in the documentation to show that `tooltip`/`trigger` is not supported in the `tree` component yet and open a new issue tracker for this.

EDIT: There seems to be some merge conflict as well.

packages/core/src/plugins/default/markbind-plugin-tree.js
```diff
@@ -0,0 +1,176 @@
...
+    },
+  },
+  getLinks: () => [`<link rel="stylesheet" href="${CSS_FILE_NAME}">`],
+  processNode: (pluginContext, node) => {
+    if (node.name !== 'tree') {
+      return;
+    }
+    node.name = 'div';
+    node.attribs.class = node.attribs.class ? `${node.attribs.class} tree` : 'tree';
+    node.children[0].data = TreeNode.visualize(node.children[0].data);
```
ang-zeyu, line 174: DomParser nodes returned by htmlparser have various `type`s, with the `isSpecial` option specified the (only) child is treated as text (much like script/style tags).

You could use cheerio apis here instead to attach the visualized string as html, which solves the main (non-whitespace) issue with #1831.

ang-zeyu: Debugger:
Untitled

damithc:
Minor: perhaps we can also add something like this to the examples given in the documentation? (it uses `:far-folder:` and `far-file`). I prefer to give examples that shows how the feature is used in realistic 'best-effect' scenarios.

image

packages/core/src/plugins/default/markbind-plugin-tree.js
tlylt: > You could use cheerio apis here instead to attach the visualized string as html

Do you mean changing it to something like:
```js
  processNode: (pluginContext, node) => {
    if (node.name !== 'tree') {
      return;
    }
    cheerio(node).text(TreeNode.visualize(cheerio(node).text()));
    cheerio(node).addClass('tree');
    node.name = 'div';
  },
```
I tried this but doesn't seem to resolve the issue with tooltips (still has white space + does not show up on hover).

ang-zeyu: yup, but this is the wrong method `.text(` (still attaching as text, not html)

ang-zeyu: reping in case this was missed @tlylt.
though not urgent, we can put up an issue if no time to test it yet.

tlylt: > reping in case this was missed @tlylt. though not urgent, we can put up an issue if no time to test it yet.

Thanks for the reminder. I have tested and setting it via .html (iirc) does achieve the effect as you mentioned. This resolves the white-space issue but triggers are still quite buggy/ not functional. I have #1831 to keep track of this and was thinking to investigate along with the comment https://github.com/MarkBind/markbind/issues/1831#issuecomment-1068871935 before pushing a fix.

PR #1826, opened 2022-03-13
Better default titles in search: Remove file format ending

### LINKED ISSUE #1803: UG/DG index.md filepath appears in search results
MacOS Monterey 12.0.1

* **MarkBind Version:** 3.1.1

Search for `userGuide/index.md` or `devGuide/index.md` in the MarkBind UG or DG search bar (respectively).

`userGuide/index.md` and `devGuide/index.md` should not appear as search results.

Screenshot 2022-03-06 at 7 00 38 PM

`userGuide/index.md` appears as a search result. Clicking on this result redirects to "Getting Started".

Same for the DG:
Screenshot 2022-03-06 at 7 06 21 PM

It seems weird that `userGuide/index.md` and `devGuide/index.md` appear as search results, since these are filepaths and not page headings or page titles.

kaixin-hc:
Fixes #1803 . 

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
When searching for a file that doesn't have a specified title, return an autogenerated title which is the file path minus the file format ending. For example, if the filepath is `userGuide/index.md`, it returns userGuide/index.
Screenshot 2022-03-13 at 11 44 42 AM

**Testing instructions:**
One way to test this is to remove `userGuide/index.md` from the unsearchable files in `site.json` on the markbind documentation site.

Improve default title for search by removing file format ending

---

- [X] No unrelated changes


packages/vue-components/src/Searchbar.vue
```diff
@@ -121,7 +121,7 @@ export default {
           headingKeywords,
         } = entry;
         const keywords = entry.keywords || '';
-        const displayTitle = title || src;
+        const displayTitle = title || src.substring(0, src.indexOf('.'));
```
tlylt, line 124: What happens if I have a file that includes '.' in the filename?
e.g.
```
const src = "test.hello.md"
console.log(src.substring(0, src.indexOf('.'))); // output "test"
```

kaixin-hc: brain glitch :') I forgot that files can have '.' in the name.

packages/vue-components/src/Searchbar.vue
```diff
@@ -121,7 +121,7 @@ export default {
           headingKeywords,
         } = entry;
         const keywords = entry.keywords || '';
-        const displayTitle = title || src;
+        const displayTitle = title || src.substring(0, -1 * src.reverse().indexOf('.'));
```
jovyntls: Would [`.lastIndexOf()`](https://www.w3schools.com/jsref/jsref_lastindexof.asp) be helpful here?

packages/vue-components/src/Searchbar.vue
```diff
@@ -121,7 +121,7 @@ export default {
           headingKeywords,
         } = entry;
         const keywords = entry.keywords || '';
-        const displayTitle = title || src;
+        const displayTitle = title || src.substring(0, src.lastIndexOf('.'));
```
ryoarmanda, line 124: Although not a major issue considering how the normal situation would be like, this would give unexpected results when the file has no extensions (like just `index`) or "double" extensions (things like `index.tar.gz`).

The latter seems a bit unlikely to be passed on to the search so I guess it's okay to not cover that (let me know if you feel otherwise though!), but it's good fault-tolerance if we can handle the former case.

ryoarmanda (commented):
A little nit below:

packages/vue-components/src/Searchbar.vue
ryoarmanda: Okay, I have delved in a little bit on how we include pages and the search, and looks like the latter case (and probably the former too) can be reasonably ignored 😅

Search data comes from `siteData.json`, which is written by `Site`, that includes data from `Site.pages`, which in turn is collected from `site.json` properties (`pages`, `globs`, and the likes). I guess we can reasonably believe that search results come from page files and not any arbitrary file. As our page files are of the normal extension form (`.md` and `.html`), then we can assume that the latter case would not happen.

By extension (heh), the former case shouldn't really happen as well. But I still think it's good to be safe rather than encountering a case where the result are empty strings one day. 😄

kaixin-hc: I investigated this suggestion more as well, and found the following code in NodeProcessor.

`      const fileExt = path.extname(file);
      if (fsUtil.isMarkdownFileExt(fileExt)) {
        const renderedContent = this.markdownProcessor.renderMd(content);
        // Wrap with <root> as $.remove() does not work on top level nodes
        parser.parseComplete(`<root>${renderedContent}</root>`);
      } else if (fileExt === '.html') {
        parser.parseComplete(`<root>${content}</root>`);
      } else {
        const error = new Error(`Unsupported File Extension: '${fileExt}'`);
        reject(error);
      }`

In other words, MarkBind only supports.html or .md . If you deliberately include a 'page' which doesn't have a supported file format ending, the `Unsupported File Extension:` error is thrown and the website cannot build at all.

So I think both the former and the latter case are invalid inputs that will safely never happen under the current implementation 😅

ryoarmanda: Thanks for the deeper investigation @kaixin-hc, we can safely use this implementation then 😄

ryoarmanda (approved):
LGTM 🎉

PR #1827, opened 2022-03-13
Include screenshot of dark themed code

### LINKED ISSUE #1811: Include an example/screenshot of code block in dark theme
Currently, the mentioning of code themes supported is very brief ([here](https://markbind.org/userGuide/formattingContents.html#themes) and [here](https://markbind.org/userGuide/siteJsonFile.html#style)). 

Given that the entire UG/DG uses the light theme for code blocks, would be nice to at least include a screenshot of the dark-themed code block in the user guide so that users can easily see what it looks like.

nil

**Additional context**
nil

kaixin-hc:
Fixes #1811.
- [X] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Adds an example of dark theme in a modal:

Screenshot 2022-03-13 at 10 51 39 PM

I considered a popover instead, but the size of the popover made it hard to see: 
popover alternative

I also considered just placing the image in directly, but since 1) images aren't centred 2) images aren't auto-resized, I felt it didn't look as good. 

**Testing instructions:**
http://127.0.0.1:8080/userGuide/formattingContents.html#fenced-code and click on the underlined text

Docs: Include screenshot of dark themed code and clean up page

---

- [ ] No unrelated changes


jovyntls:
I think using the modal is a good idea!

Do you think we should remove the extra text in the screenshot so that readers can focus on the code/output (red rectangle)? I'm not sure if the additional text (e.g. "Line numbering" section) will confuse readers 
158065371-6a37d8b0-c063-4d13-817f-bc797afa1c47

kaixin-hc:
I'm not sure! I cropped it like this because I wanted to show syntax coloured code & plain code as well as in-text code. Maybe it would be better to crop the section related to line numbering?

jovyntls:
@kaixin-hc ooh yeah I think it makes sense to show the different types of code, including the inline code. In that case, I'm neutral between cropping the line numbering section out or not. Maybe another approach could be to use lorem ipsum/placeholder hello world code?

ryoarmanda:
Nice work @kaixin-hc ! The modal looks nice.

I think to reduce confusion that the modal content being somehow related to the dark theme and still be able to present both inline and block highlights, we can cook up a custom content that is more direct in presenting the different code highlight types.

kaixin-hc:
Okay, I've changed the image as suggested!

tlylt:
Not sure if it is because of this PR or some issue with the modal, the following is what I see when browsing the preview on my phone:
image
image

ryoarmanda:
Ditto on @tlylt's issue. Looks like the image is fixed-size and overflows the modal in mobile. We should explore making the content responsive, if possible, and whether it is readable when so.

tlylt:
Using this solution https://stackoverflow.com/a/59360154/13725861 without specifying height or width of the img seems to work well. @kaixin-hc can take a look.

kaixin-hc:
@ryoarmanda @tlylt I've implemented it according to how yongliang suggested!

Will note that I also tried it with `Theme Example`, but found that it resized smaller than expected on small screen sizes. Testing this on small screen sizes with Chrome also seems to lead to unexpected behaviour, but I believe this is because of the embed overflow issues(#1797 ) which haven't been corrected yet.

docs/userGuide/syntax/code.md
```diff
@@ -2,15 +2,22 @@
...
 #### Themes
 
-MarkBind can present formatted code blocks, be it fenced or inline, with either **light** or **dark** themes. The default is dark.
+MarkBind can present formatted code blocks, be it fenced or inline, with either **light** or **dark** themes. The default is dark <trigger for="modal:code-dark-example" trigger="click">_(click for an example)_</trigger>.
+
+<modal header="Dark Code Theme" id="modal:code-dark-example" large >
+
+<div style="overflow: hidden;">
+  <img src="/images/codeDarkTheme.png" width="750" alt="Theme example" />
+</div>
```
ong6, line 11:
```suggestion
<div>
  <pic src="annotateSampleImage.png" />
</div>
```

Try replacing the inner img with a `pic` component that markbind supports :)

kaixin-hc: Seems to work as you say, I think this is a better solution than CSS magic.

One note though, on Surface Pro 7 dimensions, pic appears smaller than the div and img version and is a little worse. This is because the div version cuts off the image, but the pic version squeezes it in.

Pic version:
Screenshot 2022-04-19 at 12 14 40 AM

Div and img version: 
Screenshot 2022-04-19 at 12 14 40 AM

kaixin-hc:
I think that this isn't related to this PR? But I did notice this strange modal behaviour opening the modal causes the region taken up by text to decrease, and it increases again when the modal is closed.

https://user-images.githubusercontent.com/59471320/163839425-17984379-08e9-4a67-910f-f02bd73d3fca.mov

docs/userGuide/syntax/code.md
tlylt: > Try replacing the inner img with a `pic` component that markbind supports :). I think this might fix the overflow issues. The overflow issue seems to affect the whole page and doesn't seem to be the same issue as what tlylt was saying though.

The overflow issue that @kaixin-hc  is referring to is likely to be #1797, pending @ong6 update :)

kaixin-hc: So overall, content doesn't scale the best with width - but because the overall goal of showing that the colours are different comes across, I think while not ideal it might be okay?

ong6: I think its better with Pic, since the image is not getting cut off randomly (which might seem weird for users)

ryoarmanda:
A small suggestion to explore: how about changing the screenshot image such that the codeblock width is shorter (i.e. take the screenshot at a lower width)? I think the long "empty" space here just takes up width space, on smaller screens `<pic>` will have no choice but to squeeze it in to modal's width and reduce the quality.

ong6 (approved):
LGTM

kaixin-hc:
@ryoarmanda @ong6 Have tried your suggestion! I think it is better overall. Here are some quick screenshots of what it looks like. (It seems that on mobile though, this image seems to perform worse - but I think this is more due to the way modals are handled on mobile / the overflow issue that was raised... I'd take a picture, but its hard to demonstrate in an image.)
wider screen example
smaller screen example

tlylt:
Wonder if we should explore using a minimized panel instead? since there's quite a bit of problem with modal

kaixin-hc:
@tlylt I like modals because they are more hidden; an expandable panel seems to overstate the importance? If we avoid modals here, then we should probably go through all our docs and remove modals throughout as well. 

But I'm not too attached to this opinion, so I will change it if that's the consensus!

tlylt:
> @tlylt I like modals because they are more hidden; an expandable panel seems to overstate the importance? If we avoid modals here, then we should probably go through all our docs and remove modals throughout as well.

Er, I am referring to this specific case where the modal does not seem to play nice with images, not to say that the modal itself is not a good choice in this case or others.

> an expandable panel seems to overstate the importance? 

Personally, I thought it is actually quite important to show an example of what the dark theme looks like (which was why I raised the original issue 😅), so my take would be that the importance is justified. Also, using a minimized panel makes it take up less real estate so I think it is not too bad. I am happy to accept either solution.

jovyntls (approved):
LGTM! I think the modal works well and more clearly differentiates the example as a screenshot.

PR #1833, opened 2022-03-16
Make pop-ups accessible

### LINKED ISSUE #1745: Popover components are not accessible
MacOS Monterey 12.0.1

* **MarkBind Version:**
v3.1.1

Create a popover component without a trigger component, 
e.g. `<popover content="Lorem ipsum dolor sit amet" triggers="focus">Popover on top</popover>`

For accessibility purposes, the popover should be open-able from the keyboard to cater to keyboard-only users.

The popover content (in the above example, "Popover on top") is not focus-able by keyboard even when the `triggers` attribute is set to `focus`.
Furthermore, when a trigger method is not specified, currently MarkBind defaults to `hover`. However, this poses accessibility issues as `hover` is [impossible to trigger for keyboard-only users](https://bootstrap-vue.org/docs/directives/popover#making-popovers-work-for-keyboard-and-assistive-technology-users). 

Suggestion: Maybe we could set the default value of `triggers` to `"hover focus"`? This would maintain the expected hover behaviour, and also allow for accessibility for keyboard-only users. I think such a change would not be noticeable for most users, but would be helpful in terms of accessibility.

jovyntls:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #1745 

**Overview of changes:**
Change the default trigger from `hover` to `hover focus` 
- bootstrap-vue documentation mentions that [`hover` is impossible to trigger for keyboard users](https://bootstrap-vue.org/docs/components/popover#making-popovers-work-for-keyboard-and-assistive-technology-users)

N/A

**Testing instructions:**
Using the tab key to navigate the page should cause pop-ups to be displayed when they are in focus (i.e., when they are tabbed over)

```
Improve accessibility of pop-ups

According to the bootstrap-vue docs, the default value of `hover` for
popups is impossible to trigger for keyboard users, resulting in
accessibility issues.

Let's use `hover focus` as the default value, as this provides an
additional and unobtrusive way of displaying pop-up contents.
```

---

- [x] No unrelated changes


tlylt (commented):
> Using the tab key to navigate the page should cause pop-ups to be displayed when they are in focus (i.e., when they are tabbed over)

Tested on the preview [here](https://deploy-preview-1833--markbind-master.netlify.app/userguide/components/popups#popovers) and worked 👍

For this part:
> Name | Type | Default | Description
---- | ---- | ------- | ------
trigger | `String` | `hover focus` | How the overlay view is triggered.<br>Supports: `click`, `focus`, `hover`.

We say in the description that it supports 'click', 'focus', 'hover', but the value we use is 'hover focus', which is not part of the above. Seems a bit confusing?

jovyntls:
@tlylt That's true, thanks for the catch! I've changed it to:

> Supports: `click`, `focus`, `hover`, or any space-separated combination of these.

jonahtanjz (approved):
Thanks @jovyntls 

LGTM 👍

kaixin-hc:
It looks good! I'm wondering if these accessibility improvements can also be applied to modals. [On this page](https://deploy-preview-1833--markbind-master.netlify.app/userguide/components/popups#modals), I can't seem to activate the modals by focusing on them with the keyboard.

jovyntls:
> It looks good! I'm wondering if these accessibility improvements can also be applied to modals. [On this page](https://deploy-preview-1833--markbind-master.netlify.app/userguide/components/popups#modals), I can't seem to activate the modals by focusing on them with the keyboard.

It looks like the case for modals might be a bit tricky. When a modal is opened through `focus`, the focus moves away from the trigger. When the modal is closed, the focus automatically returns to the trigger, which makes makes the modal open again, which results in being unable to ever close the modal once it has been opened using `focus`, a bit like an infinite loop. (Popovers and tooltips don't have this problem because the focus remains on the popover/tooltip even after it is activated.)

One way to circumvent this could be to [manually return the focus](https://bootstrap-vue.org/docs/components/modal) by adding a non-visible dummy element after the trigger for modals? I'm not sure if this will have any implications though

PR #1848, opened 2022-03-22
Fix MarkBind internal package dependencies

### LINKED ISSUE #1845: Ensure referenced MarkBind packages are exact
When installing & using an older version of MarkBind (e.g. by npm i markbind-cli@3.1.0), the current configuration will install the referenced MarkBind core packages with version v3.1.1 instead, because they are compatible. This means the site that is built from markbind-cli@3.1.0 will have the version specified as v3.1.1, and it supports functionalities added in v3.1.1.

Thus, it becomes almost impossible for a user to specify an exact MarkBind version to use.

With reference to the [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies), use exact version number instead of a range.
E.g. in packages/cli/package.json
```
"@markbind/core": "^3.1.1",
"@markbind/core-web": "^3.1.1",
```
should be 
```
"@markbind/core": "3.1.1",
"@markbind/core-web": "3.1.1",
```

**Additional context**
https://github.com/MarkBind/markbind/pull/1839#issuecomment-1073998590 and the subsequent comments

tlylt:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1845

Not sure if `packages/core-web/package.json`'s dev dep on `vue-components` needs to be adjusted as well (may not be necessary). 

**Testing instructions:**
nil

Ensure the referenced versions of MarkBind packages are exact

The use of range versions for internal packages results in
incorrect versioning of MarkBind packages when using an
older version of MarkBind.

Let's remove the compatible versioning and use the exact match.

This ensures the versions of internal packages downloaded
will be exact and provide the expected functionalities.

- [x] No unrelated changes

### DIFF
--- a/docs/devGuide/projectManagement.md
+++ b/docs/devGuide/projectManagement.md
@@ -80,11 +80,12 @@ For general best practices, refer to the guide [_Working with PRs_ @SE-EDU](http
 
 1. **Make sure to start with a "clean slate"** by running `lerna clean` and then `npm run setup` in the root MarkBind directory.
 
-1. **Increment the version number** by running `lerna version --no-push`. Which to increment (`patch`, `minor` or `major`) depends on what PRs are merged for the new version, which means you must know beforehand about the changes.
+1. **Increment the version number** by running `lerna version --no-push --exact`. Which to increment (`patch`, `minor` or `major`) depends on what PRs are merged for the new version, which means you must know beforehand about the changes.
 
    <box type="info" seamless>
-  
-   The end result of this command is version commit with an appropriate tag. We will make use of the generated tag and commit message later.
+
+   * We will specify updated version numbers exactly to ensure that each version will consistently fetch the same versioned internal packages.
+   * The end result of this command is version commit with an appropriate tag. We will make use of the generated tag and commit message later.
    </box>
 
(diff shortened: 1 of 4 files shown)


jonahtanjz (commented):
Thanks @tlylt!

> Not sure if `packages/core-web/package.json`'s dev dep on `vue-components` needs to be adjusted as well (may not be necessary).

Don't think it's necessary but for consistency I think we should remove it as well.

Will also need to update the [developer guide for doing a release](https://deploy-preview-1837--markbind-master.netlify.app/devguide/projectmanagement#doing-a-release). Usually these versions are incremented using the command `lerna version --no-push` (see step 2), which automatically adds in the `^`. Will have to include the [`--exact`](https://github.com/lerna/lerna/tree/main/commands/version#--exact) option so that future releases will exclude the `^` as well.

tlylt:
> Will also need to update the [developer guide for doing a release](https://deploy-preview-1837--markbind-master.netlify.app/devguide/projectmanagement#doing-a-release). Usually these versions are incremented using the command `lerna version --no-push` (see step 2), which automatically adds in the `^`. Will have to include the [`--exact`](https://github.com/lerna/lerna/tree/main/commands/version?rgh-link-date=2022-03-22T11%3A10%3A10Z#--exact) option so that future releases will exclude the `^` as well.

Updated accordingly.

jonahtanjz (approved):
LGTM 👍

Issue #1702, opened 2022-01-08
Bootstrap 5


damithc:
Looks like Bootstrap has moved to v5

Some new things I noticed in passing: built-in support for icons, themes

ang-zeyu:
Similar to #1716. Can be worked on incrementally.

jovyntls:
Hello, I'm interested in working on this :)

I notice that Bootstrap 5 has quite a few breaking changes, which will break/change many of our styles. I think fixing the deprecated parts incrementally would be a good idea, otherwise the PR will be too big and hard to review. However, I realise that with one PR for each fix, MarkBind styles will be buggy/broken until the final PR is merged. I'm not sure what will be a good way to split up the migration?

jonahtanjz:
> I notice that Bootstrap 5 has quite a few breaking changes, which will break/change many of our styles. I think fixing the deprecated parts incrementally would be a good idea, otherwise the PR will be too big and hard to review. However, I realise that with one PR for each fix, MarkBind styles will be buggy/broken until the final PR is merged. I'm not sure what will be a good way to split up the migration?

Generally I don't think it's a good idea to split this up into multiple PRs. Like you have mentioned, this would mean having some functionalities/styles that will not work until the final PR is merged, which can take some time and also not able to test whether the newly introduced components are working as expected.

One way to do it incrementally is to open a single PR, then push changes to it periodically (with properly organised commits). Maybe after migrating a few components, you can ping someone to review it. Once it is good, you can start working on other components and continue this process until all the deprecated parts have been migrated. This should make it easier for reviewers as the changes that they will be looking at are much smaller and also ensure that all the changes can be properly tested before merging with the master branch :)

damithc:
A  _long-lived feature branch_ might be suitable for this. That is, create a separate branch (in the main repo) for this task and keep sending PRs against that branch (instead of the master branch). Once everything has been migrated, the feature branch can be merged back to the master branch.

jovyntls:
@jonahtanjz @damithc Thank you for the insights! A long-lived feature branch sounds like a good idea - I'll create a `bootstrap-v5-migration` branch in the main repo.

jonahtanjz:
@jovyntls @damithc @ang-zeyu 

[...]

Issue #1815, opened 2022-03-10
SSR leftovers


ang-zeyu:
Just a tracking issue, I think not much actionable here for now, especially as some of these might be fairly difficult / not worth the benefits to fix (the first 2 especially).

See https://github.com/MarkBind/markbind/pull/1534#issuecomment-827530954

> Just a checklist to track the follow-up tasks for this SSR PR:

* [x]  Integrate the re-rendering in SSR (due to source file changes) with [Lazy serve: Track opened pages for dependency changes #1513](https://github.com/MarkBind/markbind/pull/1513) (a dev-ops only issue)
       Also see: https://github.com/MarkBind/markbind/pull/1534#discussion_r611049244
* [ ]  Fix the regression of hot-reload to live-reload (by looking into live-server to see how we can prevent live-reload specifically upon SSR re-rendering and outputting of files) (a dev-ops only issue)
* [x]  [Find alternative to `createLocalVue`](https://github.com/MarkBind/markbind/pull/1534#discussion_r618183212)
* [x]  [Discuss and resolve inline rendering issue for `alt`](https://github.com/MarkBind/markbind/pull/1534#discussion_r619639681)
* [x]  [Resolve issue with mobile `site-nav`](https://github.com/MarkBind/markbind/pull/1534#discussion_r618212906)
* [x]  [Fix minimized panel's minor regression in behaviour](https://github.com/MarkBind/markbind/pull/1534#discussion_r620812687) 
* [x]  Hoist local script / style tags downward [SSR to resolve FOUC #1534 (comment)](https://github.com/MarkBind/markbind/pull/1534#discussion_r620902601) (p.high)
* [ ] Retarget / remove unnecessary or inappropriate selectors in the algolia plugin

Issue #1768, opened 2022-02-13
Scoped Styles - potential performance hit?


kaixin-hc:
**Is your request related to a problem?**

While not currently an issue, I'm not sure if vue components should rely primarily on scoped style. According to the documentation on scoped styles:

> Scoped styles do not eliminate the need for classes. Due to the way browsers render various CSS selectors, p { color: red } will be many times slower when scoped (i.e. when combined with an attribute selector). If you use classes or ids instead, such as in .example { color: red }, then you virtually eliminate that performance hit. https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind

However, scoped styles can reduce the chance of accidentally affecting other elements with the same class name.

Currently, newer vue components use scoped style wherever possible. Should we move out of using scoped style where it is not necessary?

jonahtanjz:
From the scoped CSS documentation, it does seem like vue-loader is recommending us to use global styles over scoped styles when possible.

> Currently, newer vue components use scoped style wherever possible. Should we move out of using scoped style where it is not necessary?

A brief look at the `vue-components` shows that most of our components are fully using scoped styles, even when the selectors are targeting classes that will only appear within their respective files. In these cases, using global styles should be preferred. I'm leaning towards moving these CSS out of scoped styles (except those that really need to be scoped) as there is not much benefit in using it.

@ang-zeyu would like to get your input on this.

ang-zeyu:
There hasn't really been a clear direction on this so far unfortunately, iirc we've just been mostly adopting the prior patterns of the original vue-strap project.

I'm open to going either way, tending toward global namespaced styles as well; It would definitely be nice to standardise this!

I don't think the performance impact is significant in our case however, at least we haven't really noticed anything, as our number of components / stylesheets are really small. The advice is likely targeted at general web apps relying on 101+ stylesheets in which case I would imagine the impact is actually measurable. So any attempt at this should be motivated by standardisation (also taking the opportunity to namespace our styles 😁).

My only concern otherwise is that any PR attempting to tackle this will need to be extremely careful, it'll likely be quite a rigorous conversion that can come in multiple prs. Some issues for example:
- Scoped styles isolate not only the application of styles to the component but also isolate child selectors to the same component (hence the need for "deep selectors" too)
- Scoped styles are internally achieved via an attribute selector, this affects its specificity. (may interfere with global styles such as bootstrap's)

gerteck:
[...]

PR #1849, opened 2022-03-23
Migrate modals from bootstrap-vue to vue-final-modal

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

Part of #1702 

**Overview of changes:**
Updates modals to use the [vue-final-modal](https://vue-final-modal.org/) package for modals instead of bootstrap-vue. 
- the modal transition CSS classes are moved to from markbind.css to Modal.vue with more specific selectors
- `bootstrapVueProcessor.js` is fully removed
	- the `_renameSlot` method is moved to `renameSlot` in `vueSlotSyntaxProcessor.js`
- Vue was upgraded from v2.6.11 to v2.6.14 to meet the dependencies of the latest version of vue-final-modal (v2.4.1)

DIscussion on using vue-final-modal and removing bootstrap-vue can be found on #1702.

Deprecation of `modal-footer` and `modal-header` slots: 
The docs state that these slot are deprecated their implementations had not been removed. 
It seems like deprecating these slots may be an unrelated change, hence these slots are re-implemented in Modals and can be removed in a future PR for fully deprecating modal-header and modal-footer.

**Testing instructions:**
The existing documented attributes for modals should work [as described](https://markbind.org/userGuide/components/popups.html#modals).

```
Use vue-final-modal for modal components

Modals are implemented using bootstrap-vue directives.

bootstrap-vue has dependencies on specific versions of Bootstrap and
Vue, making it hard for MarkBind to migrate to Bootstrap 5, or Vue 3.

vue-final-modal has minimal dependencies, and supports both Vue 2 and 3.

Let's use vue-final-modal for modals instead. Doing so allows for future
migrations to Bootstrap 5 or Vue 3, without having to wait for
bootstrap-vue to migrate first. 

Vue is also upgraded from v2.6.11 to v2.6.14, to meet the dependencies
of vue-final-modal v2.4.1.
```

---

- [x] No unrelated changes


packages/vue-components/package.json
```diff
@@ -41,6 +41,7 @@
     "jest": "^27.5.1",
     "jest-serializer-vue": "^2.0.2",
     "portal-vue": "^2.1.7",
+    "vue-final-modal": "2.3.2",
```
jonahtanjz, line 44: Just checking on this, I noticed that their latest version for Vue 2 is [`2.4.1`](https://www.npmjs.com/package/vue-final-modal). Is there a reason for using this specific version? 

Asking this because I saw in their [`2.4.0` release](https://github.com/vue-final/vue-final-modal/releases/tag/v2.4.0), they added the feature to import component directly. For example,
```js
import { vfmPlugin, $vfm, VueFinalModal } from 'vue-final-modal'
``` 
This should theoretically reduce the size of the bundle as only the required components are imported.

jovyntls: Vue Final Modal (VFM) v2.4.1 requires Vue v2.6.14 as a dependency. However, upgrading Vue from v2.6.11 (current) to v2.6.14 causes hydration errors in Navbars. The hydration errors should be unrelated to this issue - they are present even when upgrading on the master branch.
The latest version of VFM compatible with Vue v2.6.11 is VFM v2.3.2.

jonahtanjz: > The latest version of VFM compatible with Vue v2.6.11 is VFM v2.3.2.

Ahh ok got it :)

> However, upgrading Vue from v2.6.11 (current) to v2.6.14 causes hydration errors in Navbars. The hydration errors should be unrelated to this issue - they are present even when upgrading on the master branch.

Have you tried updating the dev version specified in [Site/index.js](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L319) to v2.6.14 as well to see if the hydration issue is still present? Might be cause by a mismatch of Vue's version. Will also need to update the [vue.min.js](https://github.com/MarkBind/markbind/blob/master/packages/core-web/asset/js/vue.min.js), which is used in production.

packages/vue-components/src/Modal.vue
```diff
@@ -0,0 +1,155 @@
...
+    close() {
+      this.show = false;
+    },
+  },
+  mounted() {
+    this.isMounted = true;
+  },
+};
+</script>
+<style scoped>
```
jonahtanjz, line 129: In view of #1768, we may want to consider using global styles instead since scoping may not be necessary here as these classes only appear in this file :)

jonahtanjz (commented):
Nice work @jovyntls 👍 Really clean implementation and good job on writing the tests. Will be nice to also include [Vue's snapshot test](https://github.com/MarkBind/markbind/tree/master/packages/vue-components/src/__tests__) for the newly added Modal component.

Just 1 suggestion, other than that the rest looks good :)

jovyntls:
Thanks for pointing it out @jonahtanjz , didn't realise the snapshot tests were a thing! I've changed the styles to global styles and added snapshot tests :)

jonahtanjz (approved):
Thanks @jovyntls! The tests look comprehensive :)

LGTM 👍

Will close and reopen this PR to restart the CI, looks like some issues with github recently 😅

packages/vue-components/src/Modal.vue
```diff
@@ -0,0 +1,155 @@
+<template>
+  <!-- NavBar has z-index of 1000, hence the z-index here should exceed that -->
+  <vue-final-modal
+    v-if="isMounted"
+    v-model="show"
+    ssr
```
ang-zeyu, line 6: Just to check, was `ssr=false` tried?

Presumably it prevents the modal from getting rendered during the ssr cycles, but I have been getting errors with it (might be a bug over there). The `isMounted` way seems like a good alternative.

Should we set `ssr=false` for clarity? (shouldn't be rendered during ssr) (maybe with a comment as well)

https://github.com/vue-final/vue-final-modal/blob/master/lib/VueFinalModal.vue#L3

ang-zeyu (commented):
Nice work, just 2 nits:

---

The number of package-lock.json changes introduced in merging conflicts https://github.com/MarkBind/markbind/pull/1849/commits/c3e678f4690cc75204d421ed4f7ab7bc1df94f0f seems odd (compared with the earlier 2 commits changing package-lock.json)

packages/vue-components/src/Modal.vue
jovyntls: Yup! When `isMounted` is used, both `ssr` and `ssr=false` work. (Without `isMounted`, both don't work.) I've added a comment to clarify.

Just to check my understanding, why should modals not be rendered during SSR cycles? I used browser devtools to check for SSR hydration issues, not sure if there's any other way to detect SSR issues.

jovyntls:
I tried checking out package-lock.json to the latest commit on master and then `npm run setup`, but it still gives 48k insertions and 25k deletions. O.o Not sure why there is such a large change since the only changes to the package.json files are adding `vue-final-modal` and upgrading the patch version of `vue` (along with `vue-server-renderer`, `vue-template-compiler`). 

Would this be an issue for the version history of package-lock.json?

packages/vue-components/src/Modal.vue
ang-zeyu: > EDIT: ssr=false breaks the snapshot tests, so I reverted the change

Ok, ugh, the issue seems to be that `$vfm` (being a singleton) isn't really testable. Wonder if we should transfer this once we have e2e tests setup in the other PR. @jonahtanjz  🤔

Ok with snapshot tests now though. Let's update the comment as appropriate too.

> Just to check my understanding, why should modals not be rendered during SSR cycles? I used browser devtools to check for SSR hydration issues, not sure if there's any other way to detect SSR issues.

That depends, in MarkBind's case we've decided to discard modals from indexing, per the updates you did in `algolia.js`.

For this implementation https://github.com/vue-final/vue-final-modal/blob/master/lib/VueFinalModal.vue#L3, the `ssr` option seems to render a `display: none` modal. This might be useful if you want to include modal content inside search engine results.

Another case is if you want to show the modal on-page-load (rarely, since popups are triggered by user action usually), you would need to render it during ssr as well. (by setting the initial value of `show=true`)

ang-zeyu:
> I tried checking out package-lock.json to the latest commit on master and then `npm run setup`, but it still gives 48k insertions and 25k deletions. O.o Not sure why there is such a large change since the only changes to the package.json files are adding `vue-final-modal` and upgrading the patch version of `vue` (along with `vue-server-renderer`, `vue-template-compiler`).
> 
> Would this be an issue for the version history of package-lock.json?

Hmm, just tried a fresh (wiped all `node_modules` -- could try this) install, I'm not getting any updates on the main branch.

packages/vue-components/src/Modal.vue
ang-zeyu: > per the updates you did in algolia.js.

EDIT on this: sorry, just realised with this discussion these selectors have to be retargeted / no longer necessary for SSR, something we missed out in #1534. (same rationale of excluding from indexing still applies though)

I've added it on to #1815.

jovyntls: I see, thank you so much for the explanation!
Have updated the comment too :)

jovyntls:
> Hmm, just tried a fresh (wiped all `node_modules` -- could try this) install, I'm not getting any updates on the main branch.

I tried:
- restore `package-lock.json` to the version on the master branch
- delete all `node_modules`, including those in `packages/*` 
- `npm run setup`

Still getting the excessive changes, could be that I've messed up my config somehow, though I recently uninstalled and reinstalled the newest version of npm.

ang-zeyu:
👀 Just did a fresh clone + install on another os (ubuntu) but still not getting any changes.

Just in case, try `npm -V`. Might be an issue with multiple node/npm installations and your `PATH` configuration.

If it still persists think we could get @jonahtanjz to verify as well

jonahtanjz:
> If it still persists think we could get @jonahtanjz to verify as well

I'm able to get the correct changes in `package-lock.json` by restoring from the master branch and running `npm run setup`. I've put up the one I got on github [here](https://github.com/jonahtanjz/markbind/blob/vue-final-modal/package-lock.json). Maybe you can try replacing it with this and run `npm run setup` again to see if that helps?

jovyntls:
Ah ok, a fresh clone and re-installation worked for me! Not sure what the problem with the previous clone was.

Thank you @ang-zeyu and @jonahtanjz for the help :)

packages/vue-components/src/Modal.vue
jonahtanjz: > Wonder if we should transfer this once we have e2e tests setup in the other PR. @jonahtanjz 🤔

I think we can keep the current snapshot tests for now since the e2e tests are focused on the live preview part (this may not be relevant there?). We can consider moving this once we have more e2e tests for the other frontend components as well.

ang-zeyu (approved):

PR #1858, opened 2022-03-27
Add Annotation Feature

### LINKED ISSUE #898: Support image annotation
Example expected outcome:
image

For each comment/annotation, the author specifies:
* the location to which the arrow should point to (e.g., as percentage of distance along the x and y axes from the top left corner)
* text for the comment

Perhaps this can be a part of the `pic` component

ong6:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Fixes #898

**Overview of changes:**

*Note that the implementation has changed completely after discussion with @damithc 

Added basic annotation feature into markbind with the idea of allowing users to add popovers and text to their existing images. Now works by making use of the existing floating-vue lib to help manage popovers over the image. Users will now specify the image to annotate then add child `<a-points>` to draw circles over the image which can function as markers for annotations. 

Example Code: 
```
<annotate src="https://www.researchgate.net/profile/Chaiwat-Sakul/publication/228949121/figure/fig2/AS:300654661783574@1448693065750/The-previous-square-root-circuit-1.png" alt="pic" eager type="point" width="500" >
  <a-point x="25%" y="30%" body="Lorem ipsum dolor sit amet" />
  <a-point x="50%" y="50%" label="1" size="50" header="Lorem ipsum"  body="Lorem ipsum dolor sit amet" legend="bottom" color="blue" opacity="0.1" size="100" />
</annotate>
```

Example Output:
image

The first `<a-point>` is the minimal template where an x, y and text body has to be entered. The second `<a-point>` showcases some of the changes users can make to the annotation. 

`<annotate>` functions very similarly to `<pic>` but after discussing with @damithc  we decided to split up the 2 classes to enable easier extension in the future. (E.g. adding the arrow implementation etc...)

Future improvements can be tracked at this issue #1911

Shoutout to @jovyntls for helping me with the reactivity of the points! 

**Testing instructions:**
Visit the Netlify Preview and head to `annotation` under images to test out the feature 

or 

1. inside a test markbind website run `markbind serve -d`
2. Copy the code snippet above and see if it works.
3. Play around with some of the other features

Adding a new Annotation Feature

Users want to annotate over images but have no way of doing so.

Implementing a `<annotate>` and `<a-point>` components to help users annotate over images.

This allows users to specify the image in `<annotate>`  and specify the points in `<a-point>`. 

Refer to issue #898 for initial suggestion.

---

- [x] No unrelated changes


damithc:
Good work @ong6 
Note that for this feature to be really useful (and different from simply editing the image to add text), the annotations should be dynamic and non intrusive (i.e., should not obscure/block the original image content), somewhat like popovers.

ong6:
> Good work @ong6 Note that for this feature to be really useful (and different from simply editing the image to add text), the annotations should be dynamic and non intrusive (i.e., should not obscure/block the original image content), somewhat like popovers.

Ah ok, that will have to be worked on as a future milestone as well 😅. Might need to research more on popovers. Issue is that my current implementation draws the whole image initially, so the lines are part of the image as well. I'll look into alternative implementations that might fix this issue though.

tlylt:
Good progress @ong6！

Some initial comments for the PR (since it's still a work in progress and has quite a bit of code that will probably be removed afterward:
- Just to be consistent, perhaps the tags can all be lowercase e.g. `<annotate>` instead of `<Annotate>`
- Besides specifying the color of the arrow, perhaps can consider the direction as well. I feel that it's more natural if the arrow is pointing towards the text instead.
- Looking at the implementation of the text and its styling, it seems pretty limited because of the choice to draw it via canvas. Perhaps we can explore an alternative so that future improvements on this can be easier. Not too sure if this is relevant, but I saw this [chartjs annotate plugin](https://github.com/chartjs/chartjs-plugin-annotation) that seems to be doing annotation over the canvas as well, perhaps can take a look at its implementation for inspiration.

jovyntls:
Thank you @ong6 , this looks like a really challenging and interesting issue!

Some other ideas/notes mostly on syntax:

- `annotate-wrapper` vs `annotatewrapper`: Personally I prefer kebab-case since "Annotate Wrapper" can be quite a long phrase, but I've noticed that Markbind components seem to use a mix of kebab-case (e.g. `page-nav`, `page-nav-button`, `site-nav-button`) and lowercase (e.g. `navbar`, `searchbar`)
- On supporting markdown styles: I couldn't find any solution that supports parsing markdown with HTML canvas. I'm not sure if we should manually parse the rest of markdown syntax, since there may be many edge cases. If only headers are supported, providing a `header` attribute could be another option.

More feature suggestions:
- Specifying the starting point of the arrow using percentages, e.g. `x="50%" y="50%"` refers to the centre of the image

damithc:
@ong6 given below is what I mentioned earlier today, as an alternative to arrows and text inside the image. The text can be just regular text (below) the image. Doesn't even need to be part of annotations as we can use regular MarkBind for it.
>image
>1. some explanation
>2. some **other** explanation

jovyntls (changes requested):
Cool work @ong6 ! Supporting different opacity/colours is a great idea since it circumvents issues with visibility on dark-coloured diagrams 👍  

Left some comments below!

---

Seems to have some issue with positioning the annotation points relative to the image: when the image size changes, the annotation points aren't scaled accordingly. E.g. on a large screen, with points at (0%, 0%), (50%, 50%) and (100%, 100%):
image

Same points on a smaller screen (absolute distance between points remains the same even though image size decreased):
image

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -0,0 +1,241 @@
...
+      <portal v-if="targetEl.id" :to="'popover:' + targetEl.id">
+        <h3 v-if="hasHeader" class="popover-header">
+          <slot name="header"></slot>
+        </h3>
+        <div class="popover-body">
+          <slot name="content"></slot>
+        </div>
+      </portal>
+
+      <v-dropdown
```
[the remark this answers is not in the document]
ong6: I just followed the implementation details from [the docs](https://floating-vue.starpad.dev/guide/component.html) page. From what I've seen, the implementation between v-dropdown and v-popover effectively perform the same way, so I can switch to either one.

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,110 @@
...
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+
+<annotate src="https://www.researchgate.net/profile/Chaiwat-Sakul/publication/228949121/figure/fig2/AS:300654661783574@1448693065750/The-previous-square-root-circuit-1.png" width="500">
+  <!-- Default Legend (popover only)-->
+  <a-point x="25%" y="50%" body="There is only text here" label="1"/>
+  <!-- Set Legend to bottom only -->
+  <a-point x="50%" y="50%" body="Hovering over this does nothing" label="2" legend="bottom"/>
+  <!-- Set Legend to both -->
+  <a-point x="75%" y="50%" body="There is text here and when you hover over it" label="3" legend="both"/>
```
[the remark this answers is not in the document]
ong6: yep, rephrasing this part

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,110 @@
...
+</annotate>
+</variable>
+</include>
+
+<br>
+
+\***\*`<a-point>` Options\*\***
```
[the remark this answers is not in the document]
ong6: nice spot

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,110 @@
...
+| trigger   | `String` | `click`   | How the Popover is triggered.<br>Supports: `click`, `focus`, `hover`, or any space-separated combination of these.        |
+| placement | `String` | `top`     | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                                               |
+| label     | `String` | `''`      | The content to be shown over the point.<br>The label will be omitted if this is not provided                              |
+| size      | `String` | `'40'`    | The size of the point in pixels.                                                                                          |
+| color     | `String` | `'green'` | The color of the point.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc... |
+| opacity   | `String` | `'0.3'`   | The opacity of the point.<br>Supports range of values from `0` to `1`.                                                    |
+| legend     | `String` | `'popover'`      | The presence of a legend to be displayed.<br>Supports: `popover`, `bottom`, `both`.                          |
```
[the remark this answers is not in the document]
ong6: > * `content` instead of `header` to match the rest of the pop-ups?

Did you mean `content` instead of `body`?

> * The name `legend` feels a little confusing to me (if it's at the bottom, I would call it a legend, but if it's a popover, I would think of it more of an annotation/label/pop-up). Though I'm not sure what would be a better alternative 🤔

Yeah, I also think that there could be a better name. But `legend` is indicative of what it actually is right haha.

> * Regarding the options for `legend`: `bottom` is one of the possible values of `placement`, so we may want to avoid having that as one of the options for `legend`. Maybe `caption` or `footnote` could be used instead of `bottom`? (Or any other suggestions)

Ok, maybe we can go with footnote.

> * We can add an `alt` attribute too for accessibility purposes (to set the `alt` attribute of the `` tag)! Would also allow an alternative text to be shown if the image fails to load for some reason.

alt is actually already supported, will have to add the options for the `<annotate>` class

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -0,0 +1,241 @@
...
+      return this.isBottomText();
+    },
+    hasPopover() {
+      return this.isPopover();
+    },
+  },
+  methods: {
+    isBottomText() {
+      return this.legend === 'bottom' || this.legend === 'both';
+    },
```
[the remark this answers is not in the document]
ong6: Ah initially I used these for another method that was removed later on, now I can move it up to computed! Nice spot.

ong6:
> Seems to have some issue with positioning the annotation points relative to the image: when the image size changes, the annotation points aren't scaled accordingly. E.g. on a large screen, with points at (0%, 0%), (50%, 50%) and (100%, 100%):

Did you refresh the page after making the screen size changes? Or is this by default?

docs/userGuide/syntax/annotations.md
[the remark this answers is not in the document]
jovyntls: > Yeah, I also think that there could be a better name. But `legend` is indicative of what it actually is right haha.

Ah ok, I just looked up "legend" and realised there are multiple meanings to the term. 
To me `bottom` is a legend, `popover` is a popover/pop-up, and collectively not sure what's the best term to call them.

packages/vue-components/src/annotations/AnnotatePoint.vue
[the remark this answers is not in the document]
jovyntls: From what I interpreted from their docs, they use `VDropdown` because they [import `Dropdown` as `VDropdown`](https://floating-vue.starpad.dev/guide/installation.html#vue-2) (though in our case, we import `Dropdown` as `VPopover`, since they don't have a `VPopover` component). So I'm confused as to why `v-dropdown` works. O.o (correct me if I'm wrong?)

e.g., `import { VDropdown } from 'floating-vue';` throws an error since there's no such component, but `import { Dropdown } from 'floating-vue';` works since there is a `Dropdown` component.

That aside, I think it may be better to use `v-popover` to follow the Popover implementation? This ensures that both the MarkBind Popover component and the MarkBind Annotations component are using the same implementation for popovers, since I'm not sure where `VDropdown` is coming from.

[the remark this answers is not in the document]
ong6: > e.g., `import { VDropdown } from 'floating-vue';` throws an error since there's no such component, but `import { Dropdown } from 'floating-vue';` works since there is a `Dropdown` component.

Honestly not sure why it works as well, but I did try just `dropdown` but it did not work so I switched to `v-dropdown` and it somehow started to work ;/.


> That aside, I think it may be better to use `v-popover` to follow the Popover implementation? This ensures that both the MarkBind Popover component and the MarkBind Annotations component are using the same implementation for popovers, since I'm not sure where `VDropdown` is coming from.

Ok sounds good, will change to v-popover instead then.

docs/userGuide/fullSyntaxReference.md
```diff
@@ -42,6 +42,7 @@
   tooltips : ['Tooltips', ['popups', 'reader-facing']],
   popovers : ['Popovers', ['popups', 'reader-facing']],
   modals : ['Modals', ['popups', 'reader-facing']],
+  annotations : ['Annotations', ['popups', 'reader-facing']],
```
jonahtanjz, line 45: Wondering if it is better to put this under Images & Diagrams instead of Pop-ups?

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,110 @@
...
+  <a-point x="66%" y="50%" content="Lorem ipsum dolor sit amet" trigger="hover focus"/>
+</annotate>
+</variable>
+</include>
+
+<br>
+
+**Displaying content as legends in Annotate Point** <br>
+
+Annotate point also allows users to display its content as a legend below the diagram. However, this only works when a label is provided for the point.
```
jonahtanjz, line 61: Maybe can also mention that when a `legend` attribute is used, the annotation will not be clickable.

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,110 @@
...
+| y         | `String` | `''`      | How to position the y-coordinate of the point.<br>Supports range of values from `0%` to `100%`.                           |
+| content      | `String` | `''`      | The content to be inserted into the body of the text.<br>The body will be omitted if this is not provided                 |
+| header    | `String` | `''`      | The content to be inserted into the header of the text.<br>The header will be omitted if this is not provided             |
+| trigger   | `String` | `click`   | How the Popover is triggered.<br>Supports: `click`, `focus`, `hover`, or any space-separated combination of these.        |
+| placement | `String` | `top`     | How to position the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                                               |
+| label     | `String` | `''`      | The content to be shown over the point.<br>The label will be omitted if this is not provided                              |
+| size      | `String` | `'40'`    | The size of the point in pixels.                                                                                          |
```
jonahtanjz, line 91: Since the size of the image is responsive (the width and height changes with the screen size), was wondering if we should support more dynamic sizes like `%` or `vw` as well so that it scales with the screen size. Using pixel will fixed the size of the annotation and it may become too big or too small when the image scales.

If it is too much work to implement this we can consider doing it in a follow-up PR as an extension to this :)

jonahtanjz (commented):
Good job on this feature @ong6! Mostly looks good. Just a couple of suggestions :)

> Did you refresh the page after making the screen size changes? Or is this by default?

Is it possible to make the annotation point responsive?

Since you are getting the position using `pointPosition` in `computed`, I think it is possible to move this into `method` instead and use an event listener to dynamically change the style.

Something like this:
```js
window.addEventListener("resize", () => {
    Object.assign(this.$refs.pointWrapper.style, this.pointPosition());
});
```
Assuming `pointWrapper` here is the `<span>` with the class `popover-annotation` in `AnnotatePoint.vue`.

docs/userGuide/syntax/annotations.md
ong6: This sounds like a good idea, I can definitely scale it to the size of the image. I will add this feature within this PR.

docs/userGuide/fullSyntaxReference.md
ong6: I was considering between the two as well, would like @damithc's thoughts on this.

packages/vue-components/src/annotations/Annotate.vue
```diff
@@ -0,0 +1,111 @@
...
+    <img
+      ref="pic"
+      :src="src"
+      :alt="alt"
+      :width="computedWidth"
+      class="annotate-image"
+      :loading="computedLoadType"
+      @load.once="getWidth"
+    />
+    <span class="point-wrapper">
```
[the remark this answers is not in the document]
ong6: Got it thanks for catching this!

docs/userGuide/syntax/annotations.md
ong6: > Is it possible to make the annotation point responsive?

Actually thanks to the help of @jovyntls and her godlike css skillz, it is now responsive, I'm still working on minor bug fixes and documentation cleanups!

ong6:
> Is it possible to make the annotation point responsive?

Actually thanks to the help of @jovyntls and her godlike css skillz, it is now responsive, I'm still working on minor bug fixes and documentation cleanups!

docs/userGuide/fullSyntaxReference.md
damithc: I think this is a better fit for images.

damithc:
Nice work @ong6 It's much better than I expected from the first MVP  👍 
1. I wonder what happens when the label is longer instead of just one character/number? If it is supported already, perhaps tweak the examples to include one such label as well?
2. I think we need to use some special formatting to distinguish legend from the rest of the text, perhaps show it as a bullet list, or a table?
3. Although we don't have it in other places (I think we should), it would be nice to round off the UG portion with a realistic example of the feature in use, as all the other examples use dummy content. It is a good way to showcase the _benefit_ (rather than explain just the _feature_ only). It's not good if we can't come up with even one good example of how the feature can benefit in real use.

ong6:
> Good job @ong6! Some comments on the documentation:
> 
> * Wonder if we can use a clearer image? The current one seems quite low quality
> * Wonder if we can use an image that is hosted by us to prevent it from going offline in the future?
> * Wonder if we can be more explicit and mention (or via the use of examples) that the `src` attribute can be a full URL or a relative link to the image file?

Ok I've implemented all changes!

ong6:
> Nice work @ong6 It's much better than I expected from the first MVP 👍
> 
> 1. I wonder what happens when the label is longer instead of just one character/number? If it is supported already, perhaps tweak the examples to include one such label as well?
> 2. I think we need to use some special formatting to distinguish legend from the rest of the text, perhaps show it as a bullet list, or a table?
> 3. Although we don't have it in other places (I think we should), it would be nice to round off the UG portion with a realistic example of the feature in use, as all the other examples use dummy content. It is a good way to showcase the _benefit_ (rather than explain just the _feature_ only). It's not good if we can't come up with even one good example of how the feature can benefit in real use.

Thanks prof here are some updates that I just pushed.

1. The text will have to be managed by the user, if the text is too long in the label, it will overflow out of the area. Thus I have given users the ability to adjust font size so that they can fit longer words into the clickable area.
2. I experimented with all these options but I cant really get them to look "natural" or nice looking. I tried to differentiate them by changing the weight of the font and the size so that it looks less like the generic markdown text. I will keep experimenting with different options though. 
3. Oki I'll try to think of some use cases, but if you have any ideas please let me know too 😅

damithc:
@ong6 does it work if src attribute is given a local `.puml` file (it works for ` 3\. Oki I'll try to think of some use cases, but if you have any ideas please let me know too

Perhaps in the context of explaining a diagram notation, like this one?
image

Another idea is to annotate a diagram such as an organizational chart or a grade breakdown e.g.,
image


BTW, do we have a good name for those clickable regions? hot spots? info spots? annotation markers?

ong6:
> @ong6 does it work if src attribute is given a local `.puml` file (it works for `` component, I can look into implementing it in a future PR

> BTW, do we have a good name for those clickable regions? hot spots? info spots? annotation markers?

I'm calling them annotation points or `<a-points>`

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
+{% from "userGuide/components/advanced.md" import slot_info_trigger %}
+
+## Annotations
+
+Annotation components allows you to easily add annotations over any image.
```
tlylt, line 5:
```suggestion
Annotation components allow you to easily add annotations over any image.
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <a-point x="50%" y="75%" content="Lorem ipsum dolor sit amet"  textColor="white" color="black" label="2" opacity="1"/>
+  <!-- Customize Font Size (default font size is 14) -->
+  <a-point x="75%" y="75%" content="Lorem ipsum dolor sit amet"  fontSize="30" label="3"/>
+</annotate>
+</variable>
+</include>
+
+**Using trigger for Annotate Point** <br>
+
+Similar to **[popovers](popups.md)**, annotation points also support different types of triggers
```
tlylt, line 46: link is broken

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <!-- Customize Point Color (default color is green) -->
+  <a-point x="33%" y="50%" content="Lorem ipsum dolor sit amet"  color="red"/>
+  <!-- Customize Point Opacity (default opacity is 0.3) -->
+  <a-point x="66%" y="50%" content="Lorem ipsum dolor sit amet"  opacity="0.7"/>
+  <!-- Customize Point Label (default is empty) -->
+  <a-point x="25%" y="75%" content="Lorem ipsum dolor sit amet" label="1"/>
+  <!-- Customize Text Color (default color is black) -->
+  <a-point x="50%" y="75%" content="Lorem ipsum dolor sit amet"  textColor="white" color="black" label="2" opacity="1"/>
+  <!-- Customize Font Size (default font size is 14) -->
+  <a-point x="75%" y="75%" content="Lorem ipsum dolor sit amet"  fontSize="30" label="3"/>
```
tlylt, line 39: For this part, we can consider having some meaningful annotation. For example: put an annotation in a spot where two lines cross: annotate that this is not an intersection. Another example: annotate the and gate and or gate

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+### Introduction
+
+Annotation wrappers (`<annotate>`) are used in conjunction with annotation
+points (`<a-point>`).
+
+Annotation wrappers are used to hold the image and set the width and height of
+said image.
+
+Annotation points are used to define the position, text and style of each point
+within the image. In use, the annotation points are directly inserted between
```
tlylt, line 16:
```suggestion
within the image. Insert them between
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+the annotate wrappers (`<annotate>...</annotate>`).
```
tlylt, line 17:
```suggestion
the annotation wrappers (`<annotate>...</annotate>`).
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <a-point x="75%" y="66%" content="Lorem ipsum dolor sit amet" placement="right"/>
+</annotate>
+</variable>
+</include>
+
+<br>
+
+**Displaying content as legends in Annotate Point** <br>
+
+Annotate point also allows users to display its content inside the popover or as
```
tlylt, line 70:
```suggestion
Annotate point allows users to display its content inside the popover or as
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <a-point x="25%" y="66%" content="Lorem ipsum dolor sit amet" placement="left"/>
+  <a-point x="50%" y="66%" content="Lorem ipsum dolor sit amet" placement="bottom"/>
+  <a-point x="75%" y="66%" content="Lorem ipsum dolor sit amet" placement="right"/>
+</annotate>
+</variable>
+</include>
+
+<br>
+
+**Displaying content as legends in Annotate Point** <br>
```
tlylt, line 68:
```suggestion
**Displaying content as legends in Annotate Point**
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <a-point x="66%" y="33%" content="Lorem ipsum dolor sit amet" trigger="hover focus"/>
+  <!-- Set Popover Placement (click)-->
+  <a-point x="25%" y="66%" content="Lorem ipsum dolor sit amet" placement="left"/>
+  <a-point x="50%" y="66%" content="Lorem ipsum dolor sit amet" placement="bottom"/>
+  <a-point x="75%" y="66%" content="Lorem ipsum dolor sit amet" placement="right"/>
+</annotate>
+</variable>
+</include>
+
+<br>
```
tlylt, line 66: I think can remove this br tag and the space

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+
+****`<a-point>` Options****
+
+| Name      | Type     | Default     | Description                                                                                                                              |
+| --------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| x         | `String` | `''`        | **This must be specified.**<br>How to position the x-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| y         | `String` | `''`        | **This must be specified.**<br>How to position the y-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| content   | `String` | `''`        | **This must be specified.**<br>The content to be inserted into the body of the text.<br>The body will be omitted if this is not provided |
```
tlylt, line 97: If they have a default known to the user, then it may not be true that "This must be specified"? Especially for `content`

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+  <a-point x="50%" y="50%" content="Clicking on this does nothing" label="2" legend="bottom" header="Headers are displayed as well"/>
+  <!-- Set Legend to both -->
+  <a-point x="75%" y="50%" content="There is text at both locations"  label="3" legend="both" header="Headers are displayed at both positions"/>
+</annotate>
+</variable>
+</include>
+
+<br>
+
+****`<a-point>` Options****
```
tlylt, line 91: Can consider using a `###`, so that the pageNav is utilized, same for the annotate options. Currently the pageNav is not very meaningful:
image

tlylt (changes requested):
Some documentation changes requested

tlylt:
Also noticed some strange shadow effects on the text, 
image

And another one that is more obvious [here](https://deploy-preview-1858--markbind-master.netlify.app/userguide/readerfacingfeatures#annotations)
image

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+| label     | `String` | `''`        | The content to be shown over the point.<br>The label will be omitted if this is not provided                                             |
+| size      | `String` | `'40'`      | The size of the point in pixels.                                                                                                         |
+| color     | `String` | `'green'`   | The color of the point.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc...                |
+| opacity   | `String` | `'0.3'`     | The opacity of the point.<br>Supports range of values from `0` to `1`.                                                                   |
+| fontSize  | `String` | `'14'`      | The font size of the label.<br>Supports any pixel size smaller than size of the point.                                                   |
+| textColor | `String` | `'black'`   | The color of the label.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc..                 |
```
jovyntls, line 106:
```suggestion
| textColor | `String` | `'black'`   | The color of the label.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc.                  |
```
Same for the table row for color!

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,145 @@
...
+| height | `string`  |`''`| The height of the image in pixels.                                                                                                                                                                                    |
+| width  | `string`  |`''`| The width of the image in pixels.<br>If both width and height are specified, width takes priority over height. It is to maintain the image's aspect ratio.                                                            |
+| eager  | `boolean` | false   | The `<pic>` component lazy loads its images by default.<br>If you want to eagerly load the images, simply specify this attribute.                                                                                     |
+
+</div>
+
+<div id="short" class="d-none">
+
+```
+<annotate src="../../images/annotateSampleImage.png" width="500" alt="sampleImage">
```
jovyntls, line 126:
```suggestion
<annotate src="../../images/annotateSampleImage.png" width="500" alt="Sample Image">
```
Since `alt` is usually the text displayed when the image fails to render :)

jovyntls (changes requested):
Thanks again @ong6  It looks like it's working well now!

docs/userGuide/syntax/annotations.md
ong6: This style is in line with every other option style so I just followed the existing format

docs/userGuide/syntax/annotations.md
ong6: this is done to ensure even spacing between two sections

docs/userGuide/syntax/annotations.md
ong6: That might make it harder to locate each point though. The goal here is just to demo how it works and not to show off the functionality, though I am working on a section that is focused on showing the practical aspects.

docs/userGuide/syntax/annotations.md
ong6: Hmm that's true, they could get away with omitting the content if they wished. Perhaps I should allow it... Any ideas on how to improve this?

packages/vue-components/src/annotations/Annotate.vue
```diff
@@ -0,0 +1,103 @@
...
+    hasHeight() {
+      return this.height !== '';
+    },
+    computedWidth() {
+      if (this.hasWidth) {
+        return this.width;
+      }
+      return this.widthFromHeight;
+    },
+    computedLoadType() {
```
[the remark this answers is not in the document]
ong6: Actually this was taken from the Pic.vue class, so I wanted to keep it as similar as possible since there's a possibility that we can merge the two classes in the future.

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -0,0 +1,284 @@
...
+            :arrow-padding="computeArrowPadding"
+            shift-cross-axis
+          >
+            <div class="hover-wrapper" @click.stop>
+              <button class="hover-point" :style="pointStyle">
+                {{ label }}
+              </button>
+              <div class="hover-label" :style="labelStyle">
+                {{ label }}
+              </div>
```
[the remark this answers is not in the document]
ong6: Nice spot, will nest the label inside button it makes more sense that way too.

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -0,0 +1,284 @@
...
+
+        <portal v-if="targetEl.id" :to="'popover:' + targetEl.id">
+          <h3 v-if="hasHeader" class="popover-header">
+            <slot name="header"></slot>
+          </h3>
+          <div class="popover-body">
+            <slot name="body"></slot>
+          </div>
+        </portal>
+      </div>
```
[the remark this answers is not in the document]
ong6: Yeah, I noticed that too, I think I'll go ahead and remove it then!

docs/userGuide/syntax/annotations.md
tlylt: I think as a user I want to know if it is required or can be omitted totally (by not specifying the attribute), so perhaps can reword to mention whether it's required. Also can consider removing the "How to position" to be more concise

docs/userGuide/syntax/annotations.md
tlylt: Perhaps the content of some of the points can be the location. E.g. `<a-point x="25%" y="25%" content="point at 25% 25%" />`. With the number of points in this diagram, I was unable to figure out which relates to which line of code, especially since there's no mention of how x, y works before this.

docs/userGuide/syntax/annotations.md
tlylt: In that case perhaps can consider removing the "introduction" title since there is nothing else besides it. Also can consider using a list here for clarity:
So instead of 
image

use

## Annotate
Annotate components allow you to easily annotate over any image.

- `<annotate>`: Annotate wrappers are used to hold the image and set its width and height.
-  `<a-point>`: Annotate points define the position, text and style of each point within the image. Insert them between the Annotate wrappers.

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,165 @@
...
+  <a-point x="50%" y="75%" content="Lorem ipsum dolor sit amet"  textColor="white" color="black" label="2" opacity="1"/>
+  <!-- Customize Font Size (default font size is 14) -->
+  <a-point x="75%" y="75%" content="Lorem ipsum dolor sit amet"  fontSize="30" label="3"/>
+</annotate>
+</variable>
+</include>
+
+**Using triggers for Annotate Point** <br>
+
+Similar to **[popovers](../components/popups.md)**, Annotate Points also support different types of triggers
```
tlylt, line 45:
```suggestion
Similar to **[popovers](../components/popups.md#popovers)**, Annotate Points also support different types of triggers
```

packages/vue-components/src/annotations/Annotate.vue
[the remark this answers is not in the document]
jovyntls: ah I see, in that case I agree that it's good to keep it as similar as possible

tlylt (changes requested):
When ready,
- you may want to improve the proposed commit message since this is a significant PR.
- can possibly change `Related to https://github.com/MarkBind/markbind/issues/898` to `fixes xxx` to close that issue. Improvement issues can be opened separately to track new goals.

Image broken in syntax cheatsheet https://deploy-preview-1858--markbind-master.netlify.app/userguide/syntaxcheatsheet and this one https://deploy-preview-1858--markbind-master.netlify.app/userguide/fullsyntaxreference and this one https://deploy-preview-1858--markbind-master.netlify.app/userguide/readerfacingfeatures
image

packages/vue-components/src/annotations/AnnotatePoint.vue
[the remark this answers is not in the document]
jovyntls: Can consider using background opacity instead of overall opacity. In that way only the background of the point will be affected and its children will not inherit any opacity :)

docs/userGuide/syntax/annotations.md
ong6: Ok based on the feedback I made x and y have no default value and made content optional!

damithc:
Minor:
image

`Annotations`? (and `Trees`?) -- to be consistent with existing sections

ong6:
> Minor: image
> 
> `Annotations`? (and `Trees`?) -- to be consistent with existing sections

oops, fixed!

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
...
+| header    | `String` | `''`        | The content to be inserted into the header of the text.<br>The header will be omitted if this is not provided                            |
+| trigger   | `String` | `click`     | Popover trigger type.<br>Supports: `click`, `focus`, `hover`, or any space-separated combination of these.                       |
+| placement | `String` | `top`       | Position of the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                                                              |
+| label     | `String` | `''`        | The content to be shown over the point.<br>The label will be omitted if this is not provided                                             |
+| size      | `String` | `'40'`      | The size of the point in pixels.                                                                                                         |
+| color     | `String` | `'green'`   | The color of the point.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc...                |
```
jonahtanjz, line 120:
```suggestion
| color     | `String` | `'green'`   | The color of the point.<br>Supports any color in the CSS color format. E.g. `red`, `#ffffff`, `rgb(66, 135, 245)`, etc.                |
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
...
+<br>
+
+****`<a-point>` Options****
+
+| Name      | Type     | Default     | Description                                                                                                                              |
+| --------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| x         | `String` |        | **This must be specified.**<br>The x-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| y         | `String` |        | **This must be specified.**<br>The y-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| content   | `String` | `''`        | The content to be inserted into the body of the text.<br>The body will be omitted if this is not provided |
+| header    | `String` | `''`        | The content to be inserted into the header of the text.<br>The header will be omitted if this is not provided                            |
```
tlylt, line 115: Do you mean the body/header of the popover? Not sure what the body of the text means here.

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
+{% from "userGuide/components/advanced.md" import slot_info_trigger %}
+
+## Annotations
+
+Annotate components allow you to easily annotate over any image.
```
jonahtanjz, line 5: For consistency with the other components' description.

```suggestion
**An `Annotate` component allows you to easily annotate over any images.**
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
...
+| x         | `String` |        | **This must be specified.**<br>The x-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| y         | `String` |        | **This must be specified.**<br>The y-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| content   | `String` | `''`        | The content to be inserted into the body of the text.<br>The body will be omitted if this is not provided |
+| header    | `String` | `''`        | The content to be inserted into the header of the text.<br>The header will be omitted if this is not provided                            |
+| trigger   | `String` | `click`     | Popover trigger type.<br>Supports: `click`, `focus`, `hover`, or any space-separated combination of these.                       |
+| placement | `String` | `top`       | Position of the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                                                              |
+| label     | `String` | `''`        | The content to be shown over the point.<br>The label will be omitted if this is not provided                                             |
```
jonahtanjz, line 118: Maybe can also include a note to let user know that labels should not be too long to prevent overflow?

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
...
+
+<br>
+
+****`<a-point>` Options****
+
+| Name      | Type     | Default     | Description                                                                                                                              |
+| --------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| x         | `String` |        | **This must be specified.**<br>The x-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| y         | `String` |        | **This must be specified.**<br>The y-coordinate of the point.<br>Supports range of values from `0%` to `100%`.           |
+| content   | `String` | `''`        | Annotate Point content.<br>The body will be omitted if this is not provided. |
```
jovyntls, line 114:
```suggestion
| content   | `String` | `''`        | Annotate Point content.<br>The annotation content will be omitted if this is not provided. |
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
...
+| header    | `String` | `''`        | Annotate Point header.<br>The header will be omitted if this is not provided.                            |
+| trigger   | `String` | `click`     | Popover trigger type.<br>Supports: `click`, `focus`, `hover`, or any space-separated combination of these.                       |
+| placement | `String` | `top`       | Position of the Popover.<br>Supports: `top`, `left`, `right`, `bottom`.                                                              |
+| label     | `String` | `''`        | The content to be shown over the point.<br>The label will be omitted if this is not provided.<br>Note that labels should not be too long as they might overflow out of the point.                                       |
```
jovyntls, line 118:
```suggestion
| label     | `String` | `''`        | The label shown on the point itself.<br>The label will be omitted if this is not provided.<br>Note that labels should not be too long as they might overflow out of the point.                                       |
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,162 @@
+{% from "userGuide/components/advanced.md" import slot_info_trigger %}
+
+## Annotations
+
+**An `Annotate` component allows you to easily annotate over any images.**
```
jovyntls, line 5:
```suggestion
**An `annotate` component allows you to easily annotate over any images.**
```

For consistency with the other subtitles

jovyntls (commented):
A few minor nits. Tests look good!

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+  <a-point x="50%" y="75%" content="This point is 50% from the left and 75% from the top"  textColor="white" color="black" label="2" opacity="1"/>
+  <!-- Customize Font Size (default font size is 14) -->
+  <a-point x="75%" y="75%" content="This point is 75% from the left and 75% from the top"  fontSize="30" label="3"/>
+</annotate>
+</variable>
+</include>
+
+**Using triggers for Annotate Point** <br>
+
+Similar to **[popovers](../components/popups.md#popovers)**, Annotate Points also support different types of triggers
```
jonahtanjz, line 42: Would be good to also mention positioning

```suggestion
Similar to **[popovers](../components/popups.md#popovers)**, Annotate Points also support different types of triggers and positions
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+  <a-point x="75%" y="66%" content="Popover on the right" placement="right"/>
+</annotate>
+</variable>
+</include>
+
+<br>
+
+**Displaying content as legends in Annotate Point**
+
+Annotate Point allows users to display its content inside the popover or as
```
jonahtanjz, line 66:
```suggestion
`<a-point>` allows users to display its content inside a popover or as
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+a legend below the diagram or both. However, this only works when if a label is
+also provided for the point.
```
jonahtanjz, line 68:
```suggestion
provided for the point.
```

jonahtanjz (commented):
Thanks @ong6!

Just 3 more nits. The rest looks good :)

jonahtanjz (commented):
Just noticed this

The images over at [Full syntax reference](https://deploy-preview-1858--markbind-master.netlify.app/userguide/fullsyntaxreference) and [Syntax cheat sheet](https://deploy-preview-1858--markbind-master.netlify.app/userguide/syntaxcheatsheet) are broken.

image


image

ong6:
> Just noticed this
> 
> The images over at [Full syntax reference](https://deploy-preview-1858--markbind-master.netlify.app/userguide/fullsyntaxreference) and [Syntax cheat sheet](https://deploy-preview-1858--markbind-master.netlify.app/userguide/syntaxcheatsheet) are broken.
> 
> image
> 
> image

Hmm its cause the relative links from annotations.md and full syntax reference are not the same (since annotations is nested within syntax). Is there any way to fix this?

ong6:
> Just noticed this
> 
> The images over at [Full syntax reference](https://deploy-preview-1858--markbind-master.netlify.app/userguide/fullsyntaxreference) and [Syntax cheat sheet](https://deploy-preview-1858--markbind-master.netlify.app/userguide/syntaxcheatsheet) are broken.
> 
> image
> 
> image

Nice spot, I just pushed a fix for it

jonahtanjz (approved):
Thanks @ong6!

Nice work on this new feature 👍

LGTM from my end :)

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+</variable>
+</include>
+
+<br>
+
+**Displaying content as legends in Annotate Point**
+
+`<a-point>` allows users to display its content inside a popover or as
+a legend below the diagram or both. However, this only works if a label is
+provided for the point.
```
tlylt, line 68:
```suggestion
`<a-point>` allows users to display its content !!inside a popover!! or !!as a legend below the diagram!! or !!both!!. However, a label must be provided for the point in order to display the content in both areas.
```

I think the "this" part is a bit confusing as to what it is referring to. I tried to reword it as above, but feel free to adjust as well.

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+
+<include src="codeAndOutput.md" boilerplate >
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+
+<annotate src="../../images/annotateSampleImage.png" width="500" alt="Sample Image">
+  <!-- Default Legend (popover only)-->
+  <a-point x="25%" y="50%" content="There is only text when you click me" label="1"/>
+  <!-- Set Legend to bottom only (popover is not clickable) -->
```
tlylt, line 77:
```suggestion
  <!-- Set Legend to bottom only (no popover) -->
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+  <a-point x="25%" y="75%" content="This point is 25% from the left and 75% from the top" label="1"/>
+  <!-- Customize Text Color (default color is black) -->
+  <a-point x="50%" y="75%" content="This point is 50% from the left and 75% from the top"  textColor="white" color="black" label="2" opacity="1"/>
+  <!-- Customize Font Size (default font size is 14) -->
+  <a-point x="75%" y="75%" content="This point is 75% from the left and 75% from the top"  fontSize="30" label="3"/>
+</annotate>
+</variable>
+</include>
+
+**Using triggers for Annotate Point** <br>
```
tlylt, line 40:
```suggestion
**Using triggers and positions for Annotate Point** <br>
```

docs/userGuide/syntax/annotations.md
```diff
@@ -0,0 +1,182 @@
...
+
+<annotate src="../../images/annotateSampleImage.png" width="500" alt="Sample Image">
+  <!-- Default Trigger (click)-->
+  <a-point x="33%" y="33%" content="Lorem ipsum dolor sit amet" />
+  <!-- Set Trigger to hover focus -->
+  <a-point x="66%" y="33%" content="Lorem ipsum dolor sit amet" trigger="hover focus"/>
+  <!-- Set Popover Placement (click)-->
+  <a-point x="25%" y="66%" content="Popover on the left" placement="left"/>
+  <a-point x="50%" y="66%" content="Popover on the bottom" placement="bottom"/>
+  <a-point x="75%" y="66%" content="Popover on the right" placement="right"/>
```
tlylt, line 57: Perhaps can add one example of both placement and trigger being specified, for completeness.

tlylt (changes requested):
Some minor nits from my end, otherwise looking good!

tlylt (approved):

jovyntls (approved):
LGTM 🎉 🎉 🎉  Thanks for your patience in making the changes @ong6 !

PR #1868, opened 2022-04-02
Migrate popovers and tooltips to floating-vue

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

Part of #1702 , by removing the dependency on bootstrap-vue
A copy of #1856 

**Overview of changes:**
- Popovers use the [`Menu` component from Floating Vue](https://floating-vue.starpad.dev/guide/component.html#hover-menu)
- Tooltips use the [`Tooltip` component from Floating Vue](https://floating-vue.starpad.dev/guide/component.html#tooltip)
- Default value of `placement` has been changed to `placement=top` to preserve the same behaviour. floating-vue has the option for `placement=auto`, but this tends to render popups to the right in small containers (where they previously would be on top). Since floating-vue automatically repositions popovers even for non-auto placements, `placement=top` is the closest to the original behaviour.

Should the full removal of bootstrap-vue be done in a separate PR? 

**Testing instructions:**
There should be no noticeable change in behaviour or UI for popovers and tooltips

```
Migrate popovers and tooltips to floating-vue

Popovers and tooltips are implemented using bootstrap-vue. 

bootstrap-vue has dependencies on both Vue and Bootstrap, and has not
migrated the latest versions. This is a blocker for MarkBind's
migration.

Let's use floating-vue's Menu and Tooltip components for Popovers and
Tooltips respectively. floating-vue supports both Vue 2 and Vue 3, and
has fewer dependencies. This allows us to migrate to Vue 3 and 
Bootstrap 5 more easily in the future. 
```

---

- [x] No unrelated changes


ang-zeyu (commented):
Some more issues on top of the earlier reviews:
- Some tooltips don't seem to be working:
   - This one in index.md 
      
   - These ones in tooltips.md (oddly, only on my phone)
      
- Let's add a simple fade or scale enter/leave animation similar to what we had for bootstrap-vue as well, so the activation becomes a little less jarring
- I'm unable to close this specific tooltip after activating it at times on my phone (desktop seems to be ok). Might be the same issue as the first one.
   - 
   - Reproduction (a little unreliable, might need a couple of tries):
      - double / spam click the "coupling" word a couple of times
      - try touching some area outside to close the tooltip

ryoarmanda:
>    - This one in index.md 
> 
>       

Ditto on the above.

> 
>    - These ones in tooltips.md (oddly, only on my phone)
> 
>       
>

I can access the above on my phone 🤔 not sure what happens here.

> - I'm unable to close this specific tooltip after activating it at times on my phone (desktop seems to be ok). Might be the same issue as the first one.
> 
>    - 

Sometimes thus happens on my device too. But I found out a sure way to close this, tap on the tooltip content, then tap outside it.

> Forgot to mention:
> 
> Saw that the previous PR was merged as well -- then reverted? 👀 

Few days ago me and Jonah gave the okay for 3281 students to approve and merge PRs. However, the previous PR for this one was merged with a regular merge commit. We decided to undo the merge in order to maintain atomic history, so this PR is created again.

jovyntls:
> > * This one in index.md
> >   
> 
> Ditto on the above.

Yep, this doesn't work for me too. Will look into it!

> > * These ones in tooltips.md (oddly, only on my phone)
> >   
> 
> I can access the above on my phone 🤔 not sure what happens here.

Same for me, it works fine on my phone

> > * I'm unable to close this specific tooltip after activating it at times on my phone (desktop seems to be ok). Might be the same issue as the first one.
> >   
> >   * 
> 
> Sometimes this happens on my device too. But I found out a sure way to close this, tap on the tooltip content, then tap outside it.

This one works on my phone too (can always be closed by clicking elsewhere).

ang-zeyu:
> I think popover/tooltip animations weren't implemented in v3.1.1 either (effect was in the docs but this was an error). Should I still add it to this PR?

Hmm not sure if we're looking at the same link https://markbind.org/userGuide/components/popups.html#tooltips
(not so much the presence of the `effect` prop, the animation is built into bootstrap-vue if I'm not mistaken, try hovering over it)

> Yep, this doesn't work for me too. Will look into it!

This seems fixed for me now

> Same for me, it works fine on my phone

Figured out a workaround -- holding down on it seems to work 👀

> But I found out a sure way to close this, tap on the tooltip content, then tap outside it.

It should be fixed of course (we can't expect the reader to know the holding down or above trick), hopefully both behaviours boil down to the same issue below.

> This one works on my phone too (can always be closed by clicking elsewhere).

I'm able to reproduce this very reliably now with the newest commit (previously, double/spam clicking seems to do it more reliably).
Just noticed it also occurs with the "tooltip on top/etc" buttons. (I think all the `hover` ones)

packages/vue-components/src/Popover.vue
```diff
@@ -58,16 +69,35 @@ export default {
       isMounted: false,
     };
   },
+  computed: {
+    triggers() {
+      return this.trigger.split(' ');
```
ang-zeyu, line 74: Might be missing the `touch` trigger here https://floating-vue.starpad.dev/guide/component.html#triggers

I'm not too sure of the implementation specifics in floating/bootstrap-vue as well but `hover` is generally clunky on mobile (https://stackoverflow.com/questions/22559756/changing-hover-to-touch-click-for-mobile-devices).
Likely, bootstrap-vue handled its specifics better. Could try investigating in this direction as well.

*ps: I'm using android, chrome

jovyntls:
Thanks for testing @ryoarmanda @ang-zeyu !  Sorry for the force-pushing, I was trying out a new fix by handling <input>s separately, which allows for more specific handling of triggers.

> Hmm not sure if we're looking at the same link [markbind.org/userGuide/components/popups.html#tooltips](https://markbind.org/userGuide/components/popups.html#tooltips)
(not so much the presence of the effect prop, the animation is built into bootstrap-vue if I'm not mistaken, try hovering over it)

It doesn't have the animation for me:

https://user-images.githubusercontent.com/61113575/161415358-11f6e591-9cb4-449b-86e2-7eb5d4ac4a15.mov

packages/vue-components/src/Popover.vue
jovyntls: I tried the `touch` trigger, and it displays the tooltip on "touch and hold" - when you release your finger, the tooltip disappears as well.
Should I document this behaviour in the docs?

ang-zeyu:
> It doesn't have the animation for me:

popoveranim

Hmm works for me (chrome, windows). Might be a browser issue with chrome on mac; See if you can get it working on something else (firefox / safari).

jovyntls:
> Hmm works for me (chrome, windows). Might be a browser issue with chrome on mac; See if you can get it working on something else (firefox / safari).

No animations on:
- Chrome (Mac)
- Safari (Mac/iOS)
- Firefox (Mac)

ang-zeyu:
👀👀  Tried a couple of other browsers and devices (could try this), still ok on my end

There is an https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion accessibility media query that could interfere as well (check if you have reduce motion)

jovyntls:
> 👀👀 Tried a couple of other browsers and devices (could try this), still ok on my end
> 
> There is an [developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) accessibility media query that could interfere as well (check if you have reduce motion)

You're right, turning off my reduced motion settings brought the fade back! 😭 Will add it back to the pop-ups implementations. Thanks for helping to investigate @ang-zeyu !

packages/vue-components/src/Popover.vue
```diff
@@ -21,7 +21,6 @@
       :hide-triggers="triggers"
       :placement="placement"
       :delay="0"
-      popper-class="v-popper__popper--skip-transition"
```
ang-zeyu: 👀 This does the trick for the previous issues, seems like everything was related. 

(just speculation: perhaps `v-popper__popper--skip-transition` isn't meant to be used externally? -- not sure)

packages/vue-components/src/Popover.vue
ang-zeyu: Ahh, thanks for trying. Not my intention (had expected it to be a substitute for `hover`).

https://github.com/MarkBind/markbind/pull/1868#discussion_r841177857

packages/vue-components/src/Popover.vue
ang-zeyu: Think we could @ryoarmanda to do another check too, since a lot of the issues were device specific.

ang-zeyu (approved):
Lgtm 👍

packages/vue-components/src/Popover.vue
jovyntls: > 👀 This does the trick for the previous issues, seems like everything was related.
> 
> (just speculation: perhaps `v-popper__popper--skip-transition` isn't meant to be used externally? -- not sure)

Hmm I believe so! floating-vue's [docs mention `v-popper__popper--skip-transition`](https://floating-vue.starpad.dev/guide/css.html) as a possible helper class though it seems like it's best used in conjunction with floating-vue's CSS themes as a dynamic class.

The hover issues were fixed in [`55a888`](https://github.com/MarkBind/markbind/pull/1868/commits/55a888916273b5c30f92acc5417a008a5c27cd57)—had to separate the handling of `<input>`s because wrapping `<input>` in a wrapper element results in floating-vue being unable to detect its focus events, but failing to wrap other tags in a wrapper element prevents floating-vue from detecting mouse events (which was what caused the un-openable tooltip in index.md).

ryoarmanda: The behaviours are now fixed on my end :+1:

ryoarmanda (approved):
:+1: The issues seem to be resolved now, looks good to me!

ang-zeyu (commented):
One more error, dosen't look like a hydration issue:

Untitled

jovyntls:
> One more error, dosen't look like a hydration issue:
> 
> Untitled

Thanks for the catch! It doesn't appear for me when serving locally, so I'll push some changes and check if it fixes it; will update this message when it's done

Issue #1009, opened 2020-01-31
Support site versioning


ang-zeyu:
**Is your request related to a problem?**

Site versioning is key for documentation use, and education websites may want to keep past versions for archival purposes as well.

**Describe the solution you'd like**

As discussed, we can support some form of `markbind archive <version name>` command that builds and archives the site in a say, '/archive' directory. 

The links in here would then be accessed with `${baseUrl}/archive/<version name>/<old link>`.

To add, perhaps to link it easily to the front end we can automatically insert a navigation dropdown `versions` in the sitenav ( or navbar ) through an option in `site.json`.

damithc:
Along the same vein, it's nice if we can support PR previews in `${baseUrl}/pulls/<PR number>/<old link>`
On a related note, if we can do all this using GitHub actions + gh-pages, we can shed our dependency on Travis and Netlify at the same time.

ang-zeyu:
A more comprehensive list of requirements for anyone interested:
- [ ] readers should be able to navigate to newer version from older versions - hence a single source of truth for versions is needed for _all versions_. (e.g. a `versions.json` output metadata file at the output (`_site`) and source folder root)
- [ ] the data file structure / design used to store versioning information should be extensible, so we can add more versioning information in the future as needed without breaking past versions
- [ ] a simple frontend versioning component should be provided, that allows readers to navigate to any version from any version (can be done separately)
- [ ] a `markbind archive "version name"` command
- [ ] re-executing this command should replace the version with a newly built one (see note on baseUrl below), so users can change the baseUrl
- [ ] a `markbind "delete-archive" "version name"` command (can be done separately)

Additional notes:
- The baseUrl may change across versions, if so, this would break the ability for past versions to navigate to newer versions since the single metadata file location would have changed

tlylt:
Saw a somewhat related project that might give some ideas/inspirations:
[mike,- Manage multiple versions of your MkDocs-powered documentation via Git](https://github.com/jimporter/mike)
- The `gh-pages` branch looks like [this](https://github.com/jimporter/bfg9000/tree/gh-pages)

kaixin-hc:
I'll be taking a crack at this issue in the coming weeks! 

Initial thoughts:

* We could implement this by having the archive command move all the files used to generate the current site into `archive/<version_name>`
    * All the files in `archive` can use a special layout file, which changes maybe the top level theming to make it clear which version it is. There also ought to be a way to change back to the latest version or change to other versions
[...]

Issue #1307, opened 2020-08-02
Flesh out the requirements of subsites


ang-zeyu:
Subsites work under the philosophy of - "you can create separate sites, then include and use one in another, but the included site should behave as if it was never included" (correct me if i'm wrong)

Currently however, only the following behaviours comply:
- [x] `{{baseUrl}}` resolves to subsites' root directories
- [x] nunjucks paths resolve **from** subsites' root directories for subsites' pages (#1263)
- [x] deploy config of subsites is correctly ignored

All other things here don't comply (based on some white box inferencing) (except baseUrl which shouldn't comply): https://markbind.org/userGuide/siteJsonFile.html
- [ ] subsites incorrectly share faviconpath
- [ ] subsites incorrectly share titlePrefix
- [ ] subsites incorrectly share layouts with the root site
- [ ] subsites incorrectly share theme configuration with the root site
- [ ] subsites incorrectly share page configuration with the root site
- [ ] subsites incorrectly share globalOverride
- [ ] subsites incorrectly share externalScripts
- [ ] subsites incorrectly share ignore config
- [ ] subsites incorrectly share plugins and plugin configuration
- [ ] subsites incorrectly share headingLevelIndexing

**share** - meaning subsites' pages should not be able to access the root site's properties / layouts / etc., only it's own. (it should be built as if it was built standalone)

All of this should be able to be easily achieved by creating a `Project` abstraction of sorts, which holds multiple `Site`s, although some refactors might have to be made before that.

damithc:
> **share** - meaning subsites' pages should not be able to access the root site's properties / layouts / etc., only it's own. (it should be built as if it was built standalone)

Keep in mind that the outer site should be able to override things in the sub-site (without modifying it) as that is one of the desired conditions for better reuse. That way, the sub-site can be adapted and integrated into the outer site better.

ang-zeyu:
> Keep in mind that the outer site should be able to override things in the sub-site (without modifying it) as that is one of the desired conditions for better reuse. That way, the sub-site can be adapted and integrated into the outer site better.

Hmm in that case, isn't the above behaviours actually desired? (**share** also meaning the root site overrides the sub site's configurations)

ang-zeyu:
I was thinking along the lines of:

```
root site - uses theme A
  sub site of a book - uses theme B suited for a book
```

when you build the root site, all pages in the sub site would use theme B, otherwise theme A.

There might be some usage benefits in keeping subsites independent like such.

[...]

Issue #267, opened 2018-06-18
Live preview does not use baseUrl for serving pages


yamgent:
In Jekyll, when using the live preview, the `baseurl` is taken into consideration and used as part of its serving URL. For example, if the `baseurl` is `/myblog`, then the live preview's server URL will be:

```
http://127.0.0.1:8080/myblog/
```

However, for MarkBind, we do not take `baseUrl` into consideration, so we serve everything in:

```
http://127.0.0.1:8080/
```

This makes it more difficult to determine whether the URLs inside the page's source are actually generated correctly, until you deploy to the actual server and find out that things are broken.

acjh:
If I recall correctly, the `baseUrl` works but the landing page that is initially opened is the root.

PR #1870, opened 2022-04-02
[Request for comments]: Site Versioning

kaixin-hc:
MVP for #1009 
- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**

* Add a command to archive a version under a given name. Archived versions are stored in HTML format, in a folder with the version name, which is inside the version folder(default name for the version folder is "version"). All links in the versioned site will link to other pages in the versioned site.

> To summarize the current solution, what it does is build the website and place it into a folder within the repository for it to be versioned (default location is version/, but you can customise this). When the site is built, by default the build action copies the versioned site over into _site, and it is deployed. If the user decides to store versions in different folders, then the URL to the version will simply be different.

_TODO:_
- [X] Amend the way that links have their baseUrl changed such that it works even if baseUrl != ''
- [X] Need to test for performance. (only a small difference: on MarkBind its 1/4th of a second, from ~ 4.75s to 5s.
- [x] Multiple versions -- saving a subsequent version should not keep the version folders
- [x] Clean up code
- [x] Add documentation
- [x] Add tests (?) How to test.

_Benefits of the solution proposed in this PR_
* Simple & customisable
* Since .html and asset files are kept, past changes in markbind version, assets used, etc will be no issue --> each version contains everything it needs to deploy as a site
* (Relatively) performant. I don't have proof of this yet, but I believe copying the files of html folders/asset folders should be a relatively small cost, as compared to generating from markbind files.

_Potential Drawbacks_
* The user does need to remember which folder they store their versions in, since they can change the name of the folder, etc.
* Additional manual(?) cost in implementing frontend components and creating backups of their versions in branches/repositories if they happen to use git (could be mitigated by future PRs targetting this)

_Thoughts_
* Is the site.json file copied over in the versioned files still being used? ETA: with further investigation, site.json is not copied over, a siteData.json is being created instead.

_CAVEATS:_
* If the baseURL is changed, the baseURL of the versioned site does not change. 

**Testing instructions:**

- `markbind archive <versionName> [versionFolderName]` --> builds and saves the site in a given version and folder name.
- `markbind archive <versionName>` builds and saves the site with a given version name in a folder called version

Access the site by manually appending `<versionFolderName>/<version name>` after the baseURL in the deployed site.

Implement a basic versioning CLI command

Site versioning is key for documentation use, and education websites 
may want to keep past versions for archival purposes as well.

Let's implement a markbind-cli command, markbind archive, to allow
users to easily version their website. When markbind build/serve is run,
all links within the versioned site will point to their versioned equivalent.

---

**Checklist: ☑️**

- [ ] No unrelated changes 

---

**Stretch goals/future improvements**

- [ ] Additional CLI commands to support renaming, moving versions, and deleting versions, so that users do not ever need to modify `versions.json` manually.
- [ ] Support live preview updating when the versions property of site.json is changed.
- [ ] Editing of previous versions. My preferred solution is _saving versions in separate branches_, holding all files, except previously archived versions. Update the deployed files by navigating to a separate branch (and take advantage of git; like say if you fix a bug which affects all versions, you can cherrypick the commit over to the versioned branch). This will also support reverting to previous versions (working and deploying off of the versioned branch...?). caveat: this is **dependent on git.**
    -  _Further support updating gh-pages from a versioned branch to update past versions._ I think this should be possible, as the [ghpages extension](https://github.com/tschaub/gh-pages) supports an "add" which would not remove existing files. We could remove the existing "version" directory and import the files from the branch back into main/the current branch, from which you can deploy as normal.
    - Believe this solution can account for MarkBind versions changing over time, as the master only holds the HTML files of each versioned site. Might require building and deploying to be done using NPM to ensure appropriate version control

- [ ] Front end component to easily navigate between versioned branches? Personally leaning towards a bannet, like in the 3281 site(note: banner component does not exist), but a component in the footer with a default message that you can change (like the "generated by markbind" message) might be sufficiently versatile to work as a default implementation. We could also have an auto-generated dropdown component which users can place in their navbar as desired. Key difficulty: making it versatile and customisable for all types of sites while also working "out of the box", since websites don't need to follow any set format, this may be difficult.


jonahtanjz:
Just some initial thoughts.

> Is the site.json file copied over in the versioned files still being used?

Can the user edit the archived site? From what I can see now, it only stores the built site (the `.html` files) so I am assuming once a site is archived, the user can't edit the site anymore? If that is the case, then I don't think the `site.json` needs to be copied over.

> If the baseURL is changed, the baseURL of the versioned site does not change.

Will this cause any issues? If the versioned sites are contained within the main site, then I suspect this would cause some issues with the links within the versioned sites.

> Archived versions are stored in HTML format, in a folder with the version name, which is inside the version folder(default name for the version folder is "version").

What happens when the user decides to store versions in different folders? We will need to keep track of where each versioned site is located. Maybe within the main `site.json` file? This will also help with the implementation of the frontend component to navigate between the versioned sites.

kaixin-hc:
Thanks for the review @jonahtanjz ! (also, theres been a lot of discussion in the issue #1009, but I'm not sure if I should reply mainly here or there) Some replies:

> Will this cause any issues? If the versioned sites are contained within the main site, then I suspect this would cause some issues with the links within the versioned sites.

It does cause issues in terms of deployment - if the baseUrl of the current website changes (for example, it being moved to a repository with a different name and deployed with github pages), the links in the previous website do not change accordingly. I can't see any clean workaround besides navigating back to a past commit of that version or the person keeping another copy of that version, and having the person rebuild the files. Otherwise you'd need to parse the HTML intelligently to figure out which URLs to change the baseUrl of(intra-links), and which to keep static(external links), which is not trivial. However, the changing of the baseUrl is also not a frequently expected use-case.

> We will need to keep track of where each versioned site is located.

I think we can track which folders are "version folders" in site.json - it will be necessary in order to ignore certain existing files when creating a new version. This can be done in two ways as @ang-zeyu suggested, a dedicated `versions.json` file like in docusaurus (simpler) or tracking the source and baseUrls for each version, if we need to change it in the future (developing it to allow versioned sites to point to other repos, etc -- cross origined versioning).

> What happens when the user decides to store versions in different folders? 

To summarize the current solution, what it does is build the website and place it into a folder within the repository for it to be versioned (default location is version/<versionName>, but you can customise this). When the site is built, by default the build action copies the versioned site over into `_site`, and it is deployed. If the user decides to store versions in different folders, then the URL to the version will simply be different.

From my discussion with @damithc today, the following features are features to discuss in future commits

* Front end UI component pointing to the version --> he prefers to leave it off for now, and leave it up to users to customize
* Using git branches to store previous versions (my proposal in the PR description to "Saving versions in separate branches"). To avoid a reliance on git (as users might not be using git to version control or deploy), this is not desirable as a built in feature and may be considered as an addon
    * Because of this, in response to your question "_Can we edit the versioned site_", the answer is not conveniently/through MarkBind. It is of course possible to edit the past versions manually (navigate to desired commit, re-archive the site under the desired version name, moving the created folder to replace the past copy of the past version files). But I also realised that site.json isn't kept, exactly; it's transformed into siteData.json, which is used to generate the titles and index the pages for search, for example. So I think all those files need to be created.

**Benefits of the solution proposed in this PR**

* Simple & customisable
* Since .html and asset files are kept, past changes in markbind version, assets used, etc will be no issue --> each version contains everything it needs to deploy as a site
* (Relatively) performant. I don't have proof of this yet, but I believe copying the files of html folders/asset folders should be a relatively small cost, as compared to generating from markbind files.

**Potential Drawbacks**
* The user does need to remember which folder they store their versions in, since they can change the name of the folder, etc.
* Additional manual(?) cost in implementing frontend components and creating backups of their versions in branches/repositories if they happen to use git (could be mitigated by future PRs targetting this)

kaixin-hc:
@ang-zeyu @ryoarmanda, while I still need to write tests, the basic implementation is done - would appreciate a review if you have time so that I can check I'm on the right track!

packages/core/src/Site/index.js
```diff
@@ -1467,6 +1482,84 @@ class Site {
...
+    archiveFolders.forEach((folder) => {
+      const filePath = `/${folder.output}/**`;
+      this.siteConfig.ignore.push(filePath);
+    });
+
+    // Used to get accurate intralinks within the archived site:
+    const archivedBaseUrl = this.siteConfig.baseUrl === ''
+      ? `/${versionPath}`
+      : `${this.siteConfig.baseUrl}/${versionPath}`;
+    this.siteConfig.baseUrl = archivedBaseUrl;
```
ang-zeyu, line 1508: Mostly looks like on the right track, I think the biggest missing thing may be:

> copy over the generated version of the website into a sub folder under the _sites folder for subsequent deployment

I realised with this note, perhaps we could also store the `baseUrl` used to archive the version (if we ever want to support changing baseurls) inside `versions.json`, as differing baseUrls shouldn't be copied over to `_site`.

docs/userGuide/cliCommands.md
```diff
@@ -167,6 +167,48 @@ The caveat is that not building all pages during the initial process, or not reb
 
 <hr><!-- ========================================================================== -->
 
+### `archive` Command
```
tlylt, line 170: Will need to include this command into the summary as well (~line 24)

docs/userGuide/cliCommands.md
```diff
@@ -167,6 +167,48 @@ The caveat is that not building all pages during the initial process, or not reb
...
+<div id="archiveWarning">
+<box type="warning">
+
+Warning: If the folder at `<archivePath>/<versionName>` already exists, the contents will be overwritten and your previous files may be lost. Only do so if you need to replace all the archived files with the current site files.
+</box>
+</div>
+
+* `[versionName]`<br>
+  The name of the version, and the folder which the generated HTML files and assets will be stored in.<br>
+  {{ icon_example }} `version_1`
```
tlylt, line 190: Perhaps can mention that version numbers can follow [semver](https://semver.org/), and give a more realistic version example, e.g. `v.1.1.1` or `v1`

docs/userGuide/cliCommands.md
```diff
@@ -167,6 +167,48 @@ The caveat is that not building all pages during the initial process, or not reb
...
+
+**Options** :fas-cogs:
+
+* `-s <file>`, `--site-config <file>`<br>
+  Specify the site config file (default: `site.json`)<br>
+  {{ icon_example }} `-s otherSite.json`
+
+**{{ icon_examples }}**
+* `markbind archive v1` - Stores the site in the directory `version/v1` (from the root)
+* `markbind archive version_1 custom_archive_path` - Stores the site in the directory `custom_archive_path/version_1` (from the root)
```
tlylt, line 206: Perhaps can use `./version/v1` instead of saying "(from the root)"

docs/userGuide/cliCommands.md
```diff
@@ -167,6 +167,48 @@ The caveat is that not building all pages during the initial process, or not reb
...
 
+### `archive` Command
+<br>
+
+**Format:** `markbind archive [options] [versionName] [archivePath]`
+
+**Alias:** `markbind ar`
+
+**Description:** Archives a version of the current site, which will not be affected by future changes to the site
+
```
tlylt, line 178: Perhaps can mention that a `version.json` will be created as well. Or, given the number of steps involved, perhaps can follow how the `serve` command is [described](https://markbind.org/userGuide/cliCommands.html#serve-command)

docs/userGuide/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+<div class="lead" id="overview">
+
+Site versioning is key for documentation use, and websites may want to keep past versions for archival purposes. MarkBind can help you easily archive your site.
+</div>
+
+## Archiving with a CLI command
+
+Markbind allows you to easily save a version of the site you've built to be hosted at the same site with a modified URL with a single CLI command. All intralinks within the archived site will point to the respective archived pages. By default, the archive folder is called `version`, but you may specify your own directory.
+
+For example, if your site's base url relative to your domain is 'my_site', and you archive a version named 'v1' in the default archive folder, then by navigating to the url `<domain>/my_site/version/v1/<someFile>` you would have accessed the archived version of someFile.
```
tlylt, line 25:
```suggestion
For example, if your site's base URL relative to your domain is `my_site`, and you archive a version named `v1` in the default archive folder, then by navigating to the URL `<domain>/my_site/version/v1/<someFile>` you would have accessed the archived version of `someFile`.
```

docs/userGuide/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+# {{ title }}
+
+<div class="lead" id="overview">
+
+Site versioning is key for documentation use, and websites may want to keep past versions for archival purposes. MarkBind can help you easily archive your site.
+</div>
+
+## Archiving with a CLI command
+
+Markbind allows you to easily save a version of the site you've built to be hosted at the same site with a modified URL with a single CLI command. All intralinks within the archived site will point to the respective archived pages. By default, the archive folder is called `version`, but you may specify your own directory.
```
tlylt, line 23: Would be great if intralink is added to navigate users to the cli command (and specify which command)

docs/userGuide/versioning.md
```diff
@@ -0,0 +1,32 @@
...
+
+For example, if your site's base url relative to your domain is 'my_site', and you archive a version named 'v1' in the default archive folder, then by navigating to the url `<domain>/my_site/version/v1/<someFile>` you would have accessed the archived version of someFile.
+
+A `versions.json` file will be created to track the archived sites you have made, and to exclude the archived sites from being re-archived the next time you make a new version. Modify it with caution, as it may result in unnecessary files being included or necessary files being excluded.
```
tlylt, line 27: Would be great if an example of the content of versions.json is included here

packages/cli/index.js
```diff
@@ -318,16 +318,38 @@ program
...
+      .catch(handleError);
+  });
+
 program
   .command('deploy')
   .alias('d')
   .description('deploy the site to the repo\'s Github pages')
   .option('-c, --ci [githubTokenName]', 'deploy the site in CI Environment [GITHUB_TOKEN]')
   .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
+  .option('-v, --versions <versionFolder>', 'specify the path where versions are stored')
```
tlylt, line 347: I suppose this is not documented yet? Not sure if it is ready for testing.

packages/cli/index.js
```diff
@@ -318,16 +318,38 @@ program
...
   .action((options) => {
     const rootFolder = path.resolve(process.cwd());
     const outputRoot = path.join(rootFolder, '_site');
-    new Site(rootFolder, outputRoot, undefined, undefined, options.siteConfig).deploy(options.ci)
+    const site = new Site(rootFolder, outputRoot, undefined, undefined, options.siteConfig);
+    site.deploy(options.ci)
```
tlylt, line 352: Not sure why this line is edited

packages/core/src/Site/index.js
```diff
@@ -1467,6 +1482,84 @@ class Site {
...
+   * Otherwise, create a new versions file to store information about archived versions
+   *
+   * @param {boolean} verbose Flag to emit logs of the operation
+   * @returns Returns the json object of the current versions data.
+   */
+  async writeVersionsFile(versionName, versionPath, verbose = true) {
+    const versionsPath = path.join(this.rootPath, VERSIONS_DATA_NAME);
+    const newVersionData = {
+      version_name: versionName,
+      build_ver: MARKBIND_VERSION,
```
tlylt, line 1533: Wonder if the variable names here should follow camelCase for consistency.

tlylt (changes requested):
Thanks @kaixin-hc for making good progress on this PR! I have tested the CLI command and functionality wise it seems to work as expected. Left some comments for discussion.

For reviewers, the following sections of UG describe this new feature, can go through them before starting
- [versioning](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning)
- [CLI command "archive"](https://deploy-preview-1870--markbind-master.netlify.app/userguide/clicommands#archive-command)

packages/core/src/Site/index.js
kaixin-hc: I'm not sure what you mean by "differing baseUrls sholudn't be copied over to `_site`" - from my understanding, the HTML files do away with references to baseURL by replacing references to baseURL with the baseURL. But yes, I can definitely store the archived site's baseUrl!

tlylt (commented):
One point on `versions.json` for discussion:
- Sample content of `versions.json` is as follows:
```json
{
  "versions": [
    {
      "version_name": "v1.2.2",
      "build_ver": "3.1.1",
      "output": "version/v1.2.2"
    },
    {
      "version_name": "v1.2.3",
      "build_ver": "3.1.1",
      "output": "version/v1.2.3"
    }
  ]
}
```
Wonder if we should have a root level attribute "archivePath" to keep track of the output instead of duplicating it in individual "output" (assuming that all versions will share the same root path)
So something like this:
```json
{
  "archivePath": "version",
  "versions": [
    {
      "version_name": "v1.2.2",
      "build_ver": "3.1.1",
    },
    {
      "version_name": "v1.2.3",
      "build_ver": "3.1.1",
    }
  ]
}
```

jonahtanjz:
> Wonder if we should have a root level attribute "archivePath" to keep track of the output instead of duplicating it in individual "output" (assuming that all versions will share the same root path)

@tlylt I think since we may want to extend this to allow having versions in different repos or branches, it may be better to have them as separate entries.

packages/cli/index.js
kaixin-hc: Ended up reverting the changes related to this, and forgot to change it back :')

tlylt:
> > Wonder if we should have a root level attribute "archivePath" to keep track of the output instead of duplicating it in individual "output" (assuming that all versions will share the same root path)
> 
> @tlylt I think since we may want to extend this to allow having versions in different repos or branches, it may be better to have them as separate entries.

In that case would a separate attribute "achivePath" within the individual object be cleaner? (instead of composing the output, list out the ingredients of the output path)

So something like:
```
{
  
  "versions": [
    {
      "version_name": "v1.2.2",
      "build_ver": "3.1.1",
	  "archivePath": "version",
    },
    {
      "version_name": "v1.2.3",
      "build_ver": "3.1.1",
	  "archivePath": "version",
    }
  ]
}
```

packages/core/src/Site/index.js
kaixin-hc: Hmm, I see...

It makes sense that differing baseUrls shouldn't be copied over (as they wouldn't be able to be deployed), but  I wonder if there are any cases where the baseUrls might differ, but you might still want to include it as part of the same deployment and it would work (something like a root baseUrl being 'root' and the subsite baseUrl being "root/subsite" - it still could be part of the same deployment (?)).

I think copying it over is not a significant cost or a bug, and if the user wants to it is easy to manually exclude by indicating that folder as a folder to ignore. But it shouldn't be too hard to exclude the files as well 🤔

kaixin-hc:
> In that case would a separate attribute "achivePath" within the individual object be cleaner?

Okay, I can do this! I was considering it originally, but thought output might be better as I think most of the time the use would use the composed output rather than the separate elements. Having the additional versatility is probably a good thing though

damithc:
@kaixin-hc minor: 
* Is `versions` (instead of `version`) a better name for the folder?
* Is the `versions.json` feature is already supported  in this PR? If so, I assume `versions.json` is generated automatically? I didn't see any mention of it in the documentation pages.

kaixin-hc:
> * Is `versions` (instead of `version`) a better name for the folder?

I decided on `version` because this is used to auto-generate the URL, and I think having a url that is `version/v1.1.1/` is nicer than `versions/v1.1.1/` 

> * Is the `versions.json` feature is already supported  in this PR? If so, I assume `versions.json` is generated automatically? I didn't see any mention of it in the documentation pages.

Yep, it is automatically generated and updated when new versions are created! You can find the reference to it in the [documentation](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning), but I'll be updating it with an example and perhaps a bit of information about the potential issues in modifying versions.json.

packages/cli/index.js
kaixin-hc: I added this originally when I thought it might be good to specify exactly which versions should be deployed, but now I think this should be left to a later PR to think about implementation! For now I think moving over all archived versions (or manually excluding in site.json) is the most natural solution and least prone to error.

packages/core/src/Site/index.js
ang-zeyu: > but I wonder if there are any cases where the baseUrls might differ, but you might still want to include it as part of the same deployment and it would work (something like a root baseUrl being 'root' and the subsite baseUrl being "root/subsite" - it still could be part of the same deployment (?)).

Subsites are meant to be entire sites able to be deployed independently.

Our support for subsites is more about content reuse https://markbind.org/userGuide/reusingContents.html#reusing-contents-across-sites, in particular revolving around resolving `{{baseUrl}}` properly

E.g.

Main site, baseUrl is '/cat'
Subsite , baseUrl is '/dog'

subsite index.md
```
[some link]({{baseUrl}}/foobar)
```

If you deploy the subsite independently, `baseUrl` resolves to `/dog`.

If you deploy the main site, `baseUrl` resolves to the main site base url + **sub folder** (i.e. `/cat/subfolder/of/subsite`).

This is important wherever there are links involved (link validaiton, relative link resolving, `<include>`s, ...).

> by indicating that folder as a folder to ignore.

If I'm getting this part right (ignore as in the `ignore` property), the versioning system's copying mechanism needs to be independent of asset copying (the `ignore` property).

e.g. if you have

```
ignore: ["xx.png"]
```

but at the point of archiving, `xx.png` wasn't present (meaning the author did indeed want to copy it over). The copying for versioning should still copy `xx.png` over.

kaixin-hc: > at the point of archiving, `xx.png` wasn't present (meaning the author did indeed want to copy it over). The copying for versioning should still copy `xx.png` over.

Yes, this is currently accounted for. At the point of archiving, site.json is checked to save the archived files, and is not updated unless you archive a newer version under the same name. So if `ignore: ["xx.png"]` is added afterwards, the archived site would still have `xx.png` but the current site would not.

docs/userGuide/cliCommands.md
kaixin-hc: I think whether or not users want to follow semantic versioning is up to them (this might not be so useful for a course website, for example) but I'll add the additional examples and change the example here like you suggest!

packages/core/src/Site/index.js
ang-zeyu: > the versioning system's copying mechanism needs to be independent of asset copying (the ignore property).

> The copying for versioning should still copy xx.png over.

Referring to `markbind build` copying (not `archive`), which hasn't been implemented yet.

kaixin-hc: (sorry I don't understand, could you explain that again?)

ang-zeyu: > Mostly looks like on the right track, I think the biggest missing thing may be:
> 
> > copy over the generated version of the website into a sub folder under the _sites folder for subsequent deployment

same part here ^ @kaixin-hc

ang-zeyu: From what I can see currently this hasn't been implemented, so I assume you're relying on asset copying / `ignore` property to this. Apologise as I think it might be clearer if I just assumed it hadn't been implemented 😅

ang-zeyu: The point on `ignore` however is this: If you had archived a site, say into `versions/v1`. Then, you ran `markbind build/serve`. MarkBind's asset copying would incorrectly / incompletely copy some parts of it over (since it is in the project source folder). This copying mechanism therefore needs to be implemented separately of this. The asset copying should also exclude said archive folders.

kaixin-hc: Oh :0 So the desired default behaviour, in your opinion, is to serve only the current site, and not any archived sites, because of the cost in copying over large sites? That makes sense. I had assumed the opposite, which is that if the user had archived the site previously they would want to check it when building and serving.

Currently, my understanding/implementation is:
 
* When the site is archived, the site is built with a changed baseUrl and placed into a certain folder, the location of which is tracked in `versions.json` - previously archived versions will not be re-archived.
    * (Currently working on subsite versions not being re-archived, should be done soon) 
* Every time the site is built/served, the contents of `_site` are wiped clean, and replaced by the newly built files + copied over assets (which now include the versioned sites).
    * To avoid deploying versioned sites, this is currently relying on the user manually specifying to ignore that folder.
    * Asset copying "correctly" moves over all the versioned sites, allowing the user to navigate to past versions in their live preview 
* Every time the site is deployed, the contents of `_site` and only those contents are transferred to `gh-pages` branch --> this will include all versioned sites which were not specifically ignored by the user.

And what I understand from your suggestion is that:
* when archiving, also move the archived site files into a subfolder in `_site`
* change the way the site is built and served such that only the current site is built and served, ignoring the new subfolders in `_site` which hold versions.
    * Hence we can't wipe `_site` clean, and will need to store the names of subfolders to "ignore"(either not overriding with versioning or not serving with markbind serve)
    * And I assume also adding options to allow the build and serve commands to also show archived sites, if specified. 
(One clarifying question: does this mean you suggest to store the archived site two places, both in `_site` and in the folder `<archivePath>/<versionName>` ?)

I can give it a shot, but do you think I can leave this as an improvement for another PR? I think it might be a little tricky because it affects big commands and features and might take more than a few days to implement carefully (might also make reviewing easier)

ang-zeyu: Nope, nothing so complex, ignore the additional point on `markbind serve` first, I'll come back to that later.
Your current understanding is mostly already correct.

> and replaced by the newly built files + copied over assets (which now include the versioned sites).

> Asset copying "correctly" moves over all the versioned sites, allowing the user to navigate to past versions in their live preview

This is the main issue, it is _incorrect to rely on asset copying_ for this. https://github.com/MarkBind/markbind/pull/1870#discussion_r851633960

You'll have to implement an independent copying mechanism.

If you need convincing, try archiving our documentation with `markbind archive v1`. Then, add `appveyorAddNewProject.png` to `ignore`, and run `markbind build`. The image will be missing from the archived site in `versions/v1/images`.

ang-zeyu: *Please make sure you get the above comment before reading this comment, otherwise it'll only be more confusing

> ~in which case (for `markbind serve`), I think to avoid copying the entire archived site over every file change/add/delete (expensive to copy large site over), we could also aim to do this only once 🤔 ~(realised its not possible to preview archived sites in `markbind serve` with https://github.com/MarkBind/markbind/issues/267 not done yet)~ (might be possible)

Back to this, my concern here was just with this:

1. author runs `markbind serve`
2. your yet-to-implement copying mechanism copies the versioned sites over
3. author edits files... continues with development
  3.1 Triggers your copying mechanism _everytime_ (file change/addition/deletion). This is rather disk intensive / expensive and might slow the development experience a fair bit.

> change the way the site is built and served such that only the current site is built and served, ignoring the new subfolders in _site which hold versions.

Not in the way you are thinking, but I realised with your comment you'll also need to add the _versioned site folders_ to the `pagesExclude` property here.
https://markbind.org/userGuide/siteJsonFile.html#pagesexclude

This is as:
- we support building html files as pages as well
- `.md` files may be copied to output folders as an asset in some projects

kaixin-hc: Thanks for explaining that again, and for the simple example - that's really useful to test my thoughts.

I hadn't realized that we "supported building html files as pages". I had the understanding that "building" a page was converting it from `.md` to a generated `.html` - will take a closer look at this.

kaixin-hc:
1. Have handled adding the subsite(s)' version(s) to siteConfig.ignore
2. Resolved all @tlylt 's comments
3. Saves the baseUrl at the point of archiving the site

Currently working on:
4. Independent copying mechanism for archived files - currently I'm thinking if you pass the option --versions it will copy over/build all versions (Might help with the issue of copying over large site). That means I need to exclude the copying over of versioned files during build and serve as well. This is just a preliminary idea though - I think I need to check it against the implementations of build and serve to see if it will work.
5. Correctly exclude copying over of versions with the same baseUrl

Notes: 
Might implement the version-related functions in a VersionManager class. I think having a PageManager class or similar might be the way to go to refactor the `markbind/packages/core/src/Site/index.js` file, since it already has a layout, plugin and sitelinkmanager.

kaixin-hc:
1. Have handled adding the subsite(s)' version(s) to siteConfig.ignore
2. Resolved all @tlylt 's comments
3. Saves the baseUrl at the point of archiving the site

Currently working on:
4. Independent copying mechanism for archived files - currently I'm thinking if you pass the option --versions it will copy over/build all versions (Might help with the issue of copying over large site). That means I need to exclude the copying over of versioned files during build and serve as well. This is just a preliminary idea though - I think I need to check it against the implementations of build and serve to see if it will work.
5. Correctly exclude copying over of versions with the same baseUrl

Notes: 
Might implement the version-related functions in a VersionManager class. I think having a PageManager class or similar might be the way to go to refactor the `markbind/packages/core/src/Site/index.js` file, since it already has a layout, plugin and sitelinkmanager.

ang-zeyu:
> * For build and serve, one way of implementing only 'building' or copying certain versions might be passing the cli command an option `--versions`. If passed this flag, it will copy over/build all versions (Might help with the issue of copying over large site), or maybe you can specify which options to build, otherwise it might just only serve the current site.
>   * That means I need to exclude the copying over of versioned files during build and serve as well. This is just a preliminary idea though ... I think I need to check it against the implementations of build and serve to see if it will work.

This sounds interesting, but also sounds like more work 👀 (in view of the timeline). You are welcome to experiment with this direction of course, if you have the time.

But otherwise, I would be ok with copying all versions for this PR. You can use a simple flag (e.g. `isFirstGeneration`) to tackle the issue with copying multiple times in `markbind serve`. 

Or, iirc (do double check), you might not even have to use a flag, as the function `build/serve` calls on first site build is the same one. Subsequent builds in `markbind serve` should be using separate functions, so you could take a look at implementing your copying at the end of that function.

> * Do we want to support only building/deploying selected versions? This makes sense to me, and implies to me that we might want to be able to "name" versions to refer to them later, necessitating that these names be unique: e.g. "current", "long term support". (Current "versionName" does not have to be unique, just a unique combination of archiveName and versionName)

Possibly, in view of the discussion on cross-origin support.

In this PR, we assume all sites are singly deployed for now (and we only support this), so I don't think we'll have to be worried too much.

But good to think ahead nonetheless:

>   1. We could implement this by changing the structure of versions.json so that the versions have a key which is their unique name

Agreed `version_name` could be unique.

It shouldn't be too much of a limitation for the author as we expect sites to be only periodically versioned, and should simplify future implementation greatly. (e.g. if we have 2 identical version names but one is stored in a local folder, one in a repo, we'd have to construct a composite key -- but this shouldn't be too difficult either.)

The downside is that the author loses some flexibility (no `/archivePath1/v1` can exist with `/archivePath2/v1`).
Keeping ⬇️ might be ok for now as such 

How about:

Instead of ⬇️ 

> All archived versions are stored in the folder <archivePath>/<versionName>

We let `archivePath` be the entire `<archivePath>/<versionName>` directly?

So the default value of `archivePath` would be `version/${versionName}` instead.

This way we get the benefits of both (url flexibility + simple, unique version name to specify in `--versions` or for our internal verification)

>(Current "versionName" does not have to be unique, just a unique combination of archiveName and versionName)

>   2. (Easier, potentially more practical?) We could **introduce another parameter into site.json, specifying the version file to use (default versions.json), and deploy only the versions in that version file**. But we still require versions.json to exist and to include all versions, otherwise the exclusion of version files won't work properly.

I'm favouring something like the `--versions` flag idea above to deploy/build specific versions, to keep the clunky lower level details of `versions.json` away from the user. Would also prevent any issues from incorrectly manually maintaining 2 versions.json versions.


> Tests

We could use a functional test with some "special" procedures for this (much like the special procedures for testing `markbind convert` vs other sites)

kaixin-hc:
> We let `archivePath` be the entire `<archivePath>/<versionName>` directly?
> So the default value of `archivePath` would be `version/${versionName}` instead.
> This way we get the benefits of both (url flexibility + simple, unique version name to specify in `--versions` or for our internal verification)

I do like this idea! So just to double check, archivePath becomes the URL, and both archivePath and versionName must be unique in `versions.json`. Hence for `/archivePath1/v1` and `/archivePath2/v1`, they could be specified with
```bash
markbind archive v1_take1 "archivePath1/v1" # version named v1_take1 at url {baseUrl}/archivePath1/v1
markbind archive v1_take2 "archivePath2/v1"# version named v1_take2 at url {baseUrl}/archivePath2/v1
```

ang-zeyu:
> I do like this idea! So just to double check, archivePath becomes the URL, and both archivePath and versionName must be unique in `versions.json`. Hence for `/archivePath1/v1` and `/archivePath2/v1`, they could be specified with
> 
> ```shell
> markbind archive v1_take1 "archivePath1/v1" # version named v1_take1 at url {baseUrl}/archivePath1/v1
> markbind archive v1_take2 "archivePath2/v1"# version named v1_take2 at url {baseUrl}/archivePath2/v1
> ```

yup

> archivePath and versionName must be unique in versions.json

(about `archivePath` particularly) for now at least, since we only support a single deployment / baseurl. (should be no need to verify this in your implementation)

kaixin-hc:
@ang-zeyu @MarkBind/active-devs  I still need to work on tests, however I think this is an MVP of everything discussed!

* See documentation: [cli Commands](https://deploy-preview-1870--markbind-master.netlify.app/userguide/clicommands), [site versioning](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning), [site.json file](https://deploy-preview-1870--markbind-master.netlify.app/userguide/sitejsonfile#versions)
    * build and serve now take an additional `--versions` or `-v` flag to specify version names. If you specify the versions, only those versions will be deployed. If you just use the flag, all versions are deployed. If you don't use the flag, it defaults to the `versions` property in `site.json` (which is set to an empty array by default when the SiteConfig is generated if it does not exist).
        * Have added the versions property to site.json because it may become tiring to repetitively specify the site to deploy - but it is optional.
        * This reflects the independent file copying of version files - which just copies the entire folder over into `_site`. I think this implementation should be fairly efficient as files are only copied once.
    * When archiving, the archived site is [stored as described](https://github.com/MarkBind/markbind/pull/1870#issuecomment-1101405826)
    * Sites are not copied over if baseUrls do not match
* Archive command
    * Archives the current site and not past versions or subsite versions, saving the baseUrl, markbind build, name, and path the site is stored in in `versions.json`. The current site is built, and its HTML files and assets are stored at archivePath.
    * `versions.json` is automatically updated and shouldn't need to be edited by the user. However, since delete archive / change names or location of archive commands are not implemented yet, for now the user can do so and then manually edit versions.json to ensure correct behaviour.

Note on subsites(taken from warning I wrote [here](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning#note-on-subsites)): 
> At present, when a site is archived and includes a subsite, it archives the subsite as it was at that point in time. Navigating to previous or future versions of the subsite from the parent site is not supported, though you can archive the subsite.

I decided not to look into supporting past versions of subsites within the parent subsite at present because
1. I think the use-cases are quite niche
2. At present, my only idea for implementing it is "injecting" the past versions of the subsite when every site is versioned - this will increase the time taken to archive and to build for limited utility, especially if there are many subsite versions or many parent site versions.

I believe that the archived sites are not rebuilt, recopied, or overwritten now when other files are changed (because the "generate" command and the other commands to rebuild site files seem to be separate). If this isn't true, please let me know!

packages/core/src/Site/index.js
```diff
@@ -651,17 +658,23 @@ class Site {
...
-      await this.buildAssets();
-      await (this.onePagePath ? this.lazyBuildSourceFiles() : this.buildSourceFiles());
-      await this.copyCoreWebAsset();
-      await this.copyBootstrapTheme(false);
-      await this.copyFontAwesomeAsset();
-      await this.copyOcticonsAsset();
-      await this.copyMaterialIconsAsset();
-      await this.writeSiteData();
+      this.ignoreVersionFiles('');
+      await this.buildSiteHelper();
```
jonahtanjz, line 662: A slight problem with building the main site first before copying the versions will result in the live preview refreshing multiple times.

https://user-images.githubusercontent.com/47470981/164440392-34914826-9ca5-4559-8d84-9d0509fe1e67.mp4

Steps to replicate:
1. Have about 5-6 versions so that it becomes obvious
2. `markbind serve -v`

I believe shifting `this.buildSiteHelper` to after the call to `this.addVersions` should resolve the issue.

packages/core/src/Site/index.js
```diff
@@ -1466,6 +1496,147 @@ class Site {
...
+   * @param {Array} versionFolders is the directory the versions are within
+   */
+  async addVersions(versionFolders) {
+    const versionFoldersArray = versionFolders.map(f => f.archivePath);
+    try {
+      versionFoldersArray.map(async (versionFolder) => {
+        fs.copy(versionFolder, path.join(SITE_FOLDER_NAME, versionFolder));
+      });
+      await Promise.all(versionFoldersArray);
+      logger.info('Pages copied');
```
jonahtanjz, line 1634:
```suggestion
      logger.info('Versioned sites copied');
```

So that it is clearer to the user on what is being copied.

packages/cli/index.js
```diff
@@ -311,13 +315,33 @@ program
...
   });
 
+program
+  .command('archive [versionName] [archivePath]')
+  .alias('ar')
+  .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
+  .description('archive a version of the site, which is not affected by later changes to the site')
+  .action((versionName, userSpecifiedArchivePath, options) => {
+    if (!versionName) {
+      logger.error('Please specify a name for the archived version.');
```
jonahtanjz, line 332: Maybe can throw an error here so that it will stop the archiving process if the version name is not specified?

Currently, it will be archived as `undefined`:
image

docs/userGuide/siteJsonFile.md
```diff
@@ -268,3 +269,9 @@ To disable this validation **entirely**, you may add the following to `site.json
   ```
 
 </div>
+
+#### **`versions`**
+
+**A list of version names to deploy by default when building and serving the site**. If this list is not present, by default no versions will be deployed. (You can override these settings by passing the `--versions` flag to build and serve -- see the [MarkBind CLI page](cliCommands.md) for more.)
+
+The version names to specify should be the same ones as in `versions.json`. Refer to the [Site Versioning](versioning.md) for more details.
```
jonahtanjz, line 277: Maybe can add a code example here? Similar to the other attributes.

docs/userGuide/cliCommands.md
```diff
@@ -116,13 +119,16 @@ The caveat is that not building all pages during the initial process, or not reb
    Force live reload to process all files in the site, instead of just the relevant files. This option is useful when you are modifying a file that is not a file type monitored by the <trigger trigger="click" for="modal:cliCommands-livePreview">live preview</trigger> feature.
 
 * `-p <port>`, `--port <port>`<br>
-    Serve the website in the specified port.
+  Serve the website in the specified port.
 
+* `-v [versionNames...]`, `--versions [versionNames...]` <br>
```
jonahtanjz, line 124: Perhaps can mention that this will override the attribute in `site.json`? Same for the `build` command

docs/userGuide/siteJsonFile.md
```diff
@@ -268,3 +269,9 @@ To disable this validation **entirely**, you may add the following to `site.json
   ```
 
 </div>
+
+#### **`versions`**
```
jonahtanjz, line 273: Need to update the [live preview page](https://deploy-preview-1870--markbind-master.netlify.app/userguide/glossary#live-preview) to mention that `versions` is not yet supported. This can be a future improvement :)

jonahtanjz (commented):
Nice work @kaixin-hc!

I've tested it and it seems to be working as described. Good job on that 👍 

Overall it looks good. Just a couple of suggestions that I've listed below.

ang-zeyu (commented):
> * See documentation: [cli Commands](https://deploy-preview-1870--markbind-master.netlify.app/userguide/clicommands), [site versioning](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning), [site.json file](https://deploy-preview-1870--markbind-master.netlify.app/userguide/sitejsonfile#versions)
>   
>   * build and serve now take an additional `--versions` or `-v` flag to specify version names. If you specify the versions, only those versions will be deployed. If you just use the flag, all versions are deployed. If you don't use the flag, it defaults to the `versions` property in `site.json` (which is set to an empty array by default when the SiteConfig is generated if it does not exist).
>     
>     * Have added the versions property to site.json because it may become tiring to repetitively specify the site to deploy - but it is optional.
>     * This reflects the independent file copying of version files - which just copies the entire folder over into `_site`. I think this implementation should be fairly efficient as files are only copied once.

Hmm I'm actually thinking we copy all versions over by default (since if the user archives a site, they probably want to include it in always given this PR assumes single deployment). The use case for `-v` might be to retroactively "hide" a version of the site. (though something like a future `markbind archive delete` command might also work). There's also this quote "Might help with the issue of copying over large site", which assumes all archived sites are copied over by default.

Could / should we modify this slightly to do so?

Sidenote: The use case I was originally thinking for a `--versions` flag is actually more for `markbind deploy` (more for in the future, if we really want cross-origin/repo support). In that sense, I'm not quite sure if we should include this feature (`--versions`) right now just for "Might help with the issue of copying over large site". It may not be too huge of an issue given the current system assumes fairly small (e.g. documentation sites) since it is single deployment. Hiding a version can also be done via the `.json` or maybe somethine like `markbind archive ... --delete`. Not sure if there any use cases I'm missing 🤔 

> Note on subsites(taken from warning I wrote [here](https://deploy-preview-1870--markbind-master.netlify.app/userguide/versioning#note-on-subsites)):
> 
> > At present, when a site is archived and includes a subsite, it archives the subsite as it was at that point in time. Navigating to previous or future versions of the subsite from the parent site is not supported, though you can archive the subsite.
> 
> I decided not to look into supporting past versions of subsites within the parent subsite at present because
> 
> 1. I think the use-cases are quite niche
> 2. At present, my only idea for implementing it is "injecting" the past versions of the subsite when every site is versioned - this will increase the time taken to archive and to build for limited utility, especially if there are many subsite versions or many parent site versions.

No need, this is actually in line with our current implementation of subsites. Its nothing more than a means for content reuse of one site to another easily. 

If you're interested, I raised a few suggestions / clarifications in https://github.com/MarkBind/markbind/issues/1307, but have come to think the pre-existing implementation makes more sense, since subsites are meant to be sites to be deployed independently. (but again, at the current its just more about resolving `{{baseUrl}}` correctly)

packages/cli/index.js
```diff
@@ -311,13 +315,33 @@ program
...
-      .generate(baseUrl)
+      .generate(baseUrl, options.versions)
       .then(() => {
         logger.info('Build success!');
       })
       .catch(handleError);
   });
 
+program
+  .command('archive [versionName] [archivePath]')
```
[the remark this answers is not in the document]
kaixin-hc: Good idea, I agree with this! Will do so unless anyone else disagrees.

docs/userGuide/siteJsonFile.md
kaixin-hc: Sorry, could you clarify what you mean by this! Thank you!

kaixin-hc:
> Hmm I'm actually thinking we copy all versions over by default (since if the user archives a site, they probably want to include it in always given this PR assumes single deployment). The use case for -v might be to retroactively "hide" a version of the site. (though something like a future markbind archive delete command might also work). There's also this quote "Might help with the issue of copying over large site", which assumes all archived sites are copied over by default.

> Could / should we modify this slightly to do so?
from @ang-zeyu 's comment

I've taken a bit to think about this, and I think I prefer the current implementation of 1) Specify desired versions in site.json 2) Override as specifically you want with the versions flag. Reasons being:

- Site.json holds most of the "global" site information. But some of these/defaults can be overridden already (like baseUrl, or specifying which siteconfig to use). Following this structure for versions just makes sense to me.
    - If all archived sites are copied over by default, in most cases you will need to do nothing. If you want a specific version and to exclude others, you can just specify the versions that you want and leave out the version you don't want? (I think this number of versions you need to type in will be relatively small, because how many versions will you check at once? Probably either all versions, or a specific version that you're testing).
- I think the main question here is "are there usecases where you might want to override the data in site.json"? and I think we have already thought of some, although they aren't the most common or typical
    - multiple domains
    - large sites
    - Any case where a site is archived but not intended to be deployed. I don't think this will be that common for documentation, but I could see people doing so for their own records, teachers doing so for versions of courses they are no longer teaching but think they might want to reference one day, read-only copies of notes
    - Prof raised that we shouldn't assume that our audience is necessarily using git - so I think in rare cases where one isn't using a version control tool, having a snapshot of the website might be useful. A bit like wayback machine...?
- I don't think keeping this feature makes the overall implementation that much more complex.
    - The default use would just be to specify the desired version in site.json, and not think about it after that
    - So it can just be an additional feature that is there if you need it.
- I'm not convinced implementing the reverse leads to that many convenience benefits
    - I don't think there are that many uses where you'd deliberately want to hide versions when deploying
    - Also this is a bit nitpicky but I'm not sure what to call the flag? Specifying -v definitely implies to me that that's a version I want to include, rather than a version i want to exclude.

In summary: I think there's no harm, potentially a benefit for a niche group of users, and I'm not sure -v to hide single versions is more convenient.

packages/core/src/Site/index.js
kaixin-hc: I'm having difficulty replicating this : ( I'll try it on a different computer, but I seem to get the same behaviour with both orders of this line, with 5 versions of the markbind docs website (which I think should be big enough(?). I've moved it for now - if you or anyone else has time I'd appreciate if you could check if that fixes the issue!

jonahtanjz: > I've moved it for now - if you or anyone else has time I'd appreciate if you could check if that fixes the issue!

Yup this seems to have fixed the issue :)

docs/userGuide/siteJsonFile.md
jonahtanjz: We currently have live preview (hot reload) for changes made to the attributes in `site.json` except for `baseUrl`. I'm assuming the `versions` attribute does not support hot reload yet from what I can see (if I'm already serving a site using `markbind serve`, adding/removing a new version to `site.json` will not add/remove this to the site unless I restart the preview)? 

If that is the case, we will have to document this exception here
image

Don't have to implement this feature in this PR as it can be a future improvement :)

ang-zeyu:
> I've taken a bit to think about this, and I think I prefer the current implementation of 1) Specify desired versions in site.json 2) Override as specifically you want with the versions flag. Reasons being:

> I think the main question here is "are there usecases where you might want to override the data in site.json"? and I think we have already thought of some, although they aren't the most common or typical

> I'm not convinced implementing the reverse leads to that many convenience benefits

Thanks for the explanation, I think we are already on the same page. I'm also ok with `site.json` followed by `--versions` override.
Meant altering the defaults for the more common use case like how you're doing it for `--versions` currently:

```
// site.json
  "versions": null // by default (or if the property dosen't exist) -- deploy all versions.
```

From what I see in `SiteConfig.js` it currently seems to be `[]` by default. (no versions copied)

> Any case where a site is archived but not intended to be deployed. I don't think this will be that common for documentation, but I could see people doing so for their own records, teachers doing so for versions of courses they are no longer teaching but think they might want to reference one day, read-only copies of notes

> - I don't think keeping this feature makes the overall implementation that much more complex.
>    - The default use would just be to specify the desired version in site.json, and not think about it after that

ok 👍, convinced of the use case.

Though, to add some perspective on the second argument for generality, even "small" features like this add a commitment, which can introduce other non-implementation complexity issues down the line (deprecation, breaking changes, integration with some other feature, contradiction with other yet-to-be-added features). Which makes supporting rare use cases not too worthwhile in the first place. From a user standpoint, adding features does increase product complexity as well, in some cases that make things less appealing. On the contrary, an appealing feature should also not be forgone in view of implementation complexity, if possible with the resources.

The best example here might be bootstrap themes, which while initially simple in implementation, has many contradictions with https://github.com/MarkBind/markbind/issues/903, and a whole bunch of unforseen implementation issues that are too tedious to maintain / resolve.

ang-zeyu (commented):
while you're working on the rest, some documentation nits + one more follow up on the above!

docs/userGuide/versioning.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Working with sites with multiple versions
+
+You may not always want to build all your saved versions. To specify the "default versions to build", add a [versions property](siteJsonFile.md#versions) in your `site.json` file.
+
+You may also specify which versions to build when using the build and serve cli commands([more information](cliCommands.md)).
+
+## Note on subsites
+
+At present, when a site is archived and includes a subsite, it archives the subsite as it was at that point in time. Navigating to previous or future versions of the subsite from the parent site is not supported, though you can archive the subsite.
```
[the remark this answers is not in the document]
kaixin-hc: Do you think this section still has any use? I was thinking about the current example of the CS2103T websites having the textbook embedded when I wrote this section. While you might archive versions of the textbook, if you also archived versions of the 2103 website, would it be important to note that when navigating to the textbook files it would show the textbook as it was at the point of archiving and you might not see the most recent version of the textbook?

It's kind of niche, but I wasn't sure if it would be an unexpected behaviour for the user.

docs/userGuide/versioning.md
```diff
@@ -0,0 +1,78 @@
...
+
+Markbind allows you to easily save a version of the site you've built to be hosted at the same site with a modified URL with a [single CLI command](cliCommands.md#archive-command). All intralinks within the archived site will point to the respective archived pages. By default, the archived site is stored in a folder `version/<versionName>`, but you may specify your own archivePath.
+
+For example, if your site's base URL relative to your domain is `my_site`, and you archive a version named `v1` in the default archive folder, then by navigating to the URL `<domain>/my_site/version/v1/<someFile>` you can have accessed the archived version of `someFile`.
+
+A `versions.json` file will be created to track the archived sites you have made, and to exclude the archived sites from being re-archived the next time you make a new version. This file is **automatically updated** every time you archive a version.
+
+<box type="warning">
+
+Modify versions.json with caution as it may result in unnecessary files being included or necessary files being excluded.
```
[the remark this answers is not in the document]
kaixin-hc: Mm, I agree with you, its just that I haven't implemented CLI commands to manage this yet so it might become necessary to edit it as a stopgap measure (changing the name of versions, for example, or deleting versions). I think using CLI commands to support renaming, deletion and maybe moving would be ideal, but might be outside the scope of this PR? Though I don't think it would be too difficult!

docs/userGuide/versioning.md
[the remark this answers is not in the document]
kaixin-hc: That's a good point, it sounds like it might be more confusing than useful in its current form. I'll take it out, thank you!

packages/core/src/Site/index.js
```diff
@@ -1466,6 +1496,147 @@ class Site {
...
+    // Do not transfer the versions file into the archived site
+    this.siteConfig.ignore.push(path.join(pathToVersionFromRootDir, VERSIONS_DATA_NAME));
+
+    const pathsToVersionFiles
+     = walkSync(pathToDirWithVersion, { directories: false, ignore: this.siteConfig.ignore })
+       .filter(x => x.endsWith(VERSIONS_DATA_NAME))
+       .map(x => path.relative(pathToDirWithVersion, x)) // assumes versions files are in the 'root' of site
+       .map(x => path.dirname(x));
+
+    pathsToVersionFiles.forEach(p => this.ignoreVersionFiles(p));
```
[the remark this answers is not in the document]
kaixin-hc: I believe I did reference this section of baseUrlMap, but because I needed an additional map, and was referencing different paths and everything, I thought the additional complexity of extracting it and abstracting this function for re-use would actually make it less maintainable 😅 But if I'm misunderstanding do let me know!

docs/userGuide/cliCommands.md
```diff
@@ -158,10 +170,65 @@ The caveat is that not building all pages during the initial process, or not reb
...
+  Warning: If the folder at `<archivePath>` already exists, the contents will be overwritten and your previous files may be lost. Only do so if you need to replace all the archived files with the current site files.
+
+  (Also note that you cannot save a version with the same name into a different archive path.)
+  </box>
+</div>
+
+**{{ icon_examples }}**
+
+* `markbind archive v1`: Stores the archived site in the directory `./version/v1` as the version named 'v1'
+* `markbind archive version_1 -ap custom_archive_path`: Stores the archived site in the directory `./custom_archive_path`, and the version is named version_1.
```
tlylt, line 229:
```suggestion
* `markbind archive version_1 -ap custom_archive_path`: Stores the archived site in the directory `./custom_archive_path`, and the version is named 'version_1'.
```

tlylt (changes requested):
Good work @kaixin-hc ! Left some documentation comments for consideration pls

tlylt (commented):
Further comments added for consideration pls

packages/core/test/unit/Site.test.js
```diff
@@ -676,3 +682,212 @@ siteJsonPageExclusionTestCases.forEach((testCase) => {
...
+    expected:
+      {
+        v1: false,
+        v2: false,
+        differentBaseUrl: false,
+      },
+  },
+];
+
+// TODO:
```
ang-zeyu, line 827: > We could use a functional test with some "special" procedures for this (much like the special procedures for testing markbind convert vs other sites)

Thanks for this. I was actually just expecting functional tests `test/e2e/functional` which are easier to maintain.
For example,

```js
// testSites.js

const testVersioningSites = [
  'test_site_versioning/1',
];

// test.js
testVersioningSites .forEach((siteName) => {
  console.log(`Running ${siteName} versioning tests`);
  const nonMarkBindSitePath = path.join(siteName, 'non_markbind_site');
  try {
    execSync(`node ../../index.js version versionName`, execOptions);
    compare(siteName, 'expected1', '_site');

    // The downside specific to MarkBind's functional tests is that we need the expected directory (which if there are too many of might begin to slow git a little)
   // But you don't have to create one site for every behaviour you want to test,
   // e.g. get creative with something like this

    execSync(`node ../../index.js version versionName --versions`, execOptions);
    expect versions dir to be empty

    ... more ...
    
  } catch (err) {
    printFailedMessage(err, siteName);
    cleanupConvert(path.resolve(__dirname, siteName));
    process.exit(1);
  }
  cleanupConvert(path.resolve(__dirname, siteName));
});
```

In general, note that we aim to test public interfaces instead of private ones. In the case of markbind (a cli app), especially about `Site.js`'s functionalities which integrate multiple things, that's predominantly e2e behaviour.

You may have noticed the tests are quite brittle (e.g. what if we reordered the execution of `readSiteConfig` / `writeVersionsFile`), these are whimsical implementation details that once changed would require changing the tests too.

**A lot of our existing unit tests** frankly don't adhere to this well too, and definitely need improving.

Though for clarification -- this is not me making a case for e2e tests, just "which is easier to write and maintain" _in this case_ (and still tests what we want to test), which varies from project to project. For example, deeper implementation details that are sufficiently complex and well encapsulated, e.g. a hypothetical pure function like `replaceInvalidCharacters`, can and probably should be unit-tested, if we expect them to work a certain way.

I'm ok with keeping the tests here since you've already done quite a bit of work here, but **if you have the time**, I think we can adapt these to functional tests. But we can definitely work on porting the tested behaviours otherwise in the future :)

ang-zeyu (commented):
Thanks for continuing work on this! @kaixin-hc 

I think we're almost there, a few more minor implementation details:

packages/core/src/Site/index.js
```diff
@@ -641,7 +648,7 @@ class Site {
    * @param baseUrl user defined base URL (if exists)
    * @returns {Promise}
    */
-  async generate(baseUrl) {
+  async generate(baseUrl, versionsToGenerate = false) {
```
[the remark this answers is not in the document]
kaixin-hc: Ah yeah, its not only a boolean so I ended up choosing this name. It's not perfect but I couldn't think of anything better - suggestions appreciated!

packages/core/src/Site/index.js
```diff
@@ -135,6 +136,12 @@ class Site {
     this.siteConfig = undefined;
     this.siteConfigPath = siteConfigPath;
 
+    /**
+     * Archived version information
+     * @type {undefined | SiteConfig}
```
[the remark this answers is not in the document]
kaixin-hc: good catch >///> yeah its a json object

packages/core/src/Site/index.js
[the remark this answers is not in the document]
tlylt: Hmm, understand that the variadic option is used here, just slightly uncomfortable with the [options-boolean-or-value](https://github.com/tj/commander.js/blob/master/examples/options-boolean-or-value.js) aspect 🤯. I think perhaps a suggestion would be to process the `options.versions` and split the versionTogenerate into two variables `shouldGenerate` and `versions` before passing them into the generate function. The existing solution is fine too.

A separate nit that I happen to realize: line 648, need to add the new parameter into the jsdoc comment for this `generate` function

packages/core/test/unit/Site.test.js
kaixin-hc: Whoops... Maybe for a future PR 😓  I don't think I'll have the time in the coming months.

kaixin-hc:
@MarkBind/active-devs @damithc Hi all! Sorry for no progress on this for so long. Can I check if this feature is still relevant to figure out if this should be revised or just closed?

damithc:
I think it is still relevant, @kaixin-hc

kaixin-hc:
Okay! Just to update that I've revisited the code and these comments to some extent and I think the original principles were sound, but the refactoring done since when I first worked on this and now is making the merge quite painful, which is why there's been some delay; but I'll make the fixes either here or in a fresh PR.

Issue #1421, opened 2020-12-13
Adopt typescript for the backend


ang-zeyu:
**Is your request related to a problem?**

The readability of the backend (`core` package) codebase can really benefit from some strict typing.

**Describe the solution you'd like**

Migrate to typescript

Requirements:
- The migration should be able to be done incrementally
- Adjustments should be made to the dev workflow such that it isn't more cumbersome than before. We may provide npm scripts for:
   - auto recompilation of typescript source files
   - auto restart of some common dev commands (e.g. `markbind s -d` in the `docs/` folder) run after recompilation
   - can look into: https://www.npmjs.com/package/tsc-watch
   - ❔ 
- Other workflows should be adapted as necessary: release, automated site deployment, ...

**Additional context**

The frontend may not see as much benefit, since we only develop mostly independent components. Also, typescript support for vue v2 apis isn't known to be strong.

JSDoc typing: https://github.com/MarkBind/markbind/pull/1039#pullrequestreview-386546173

le0tan:
Looks like `cli` is now an independent npm package that depends on `core` - is there any way to make `require('@markbind/core/src/...')` depends on the local version?

ang-zeyu:
@le0tan should be already, we have some updated setup instructions :-)
https://markbind.org/devdocs/devGuide/settingUp.html#setting-up-the-dev-environment

le0tan:
@ang-zeyu Looks like the dev guide should replace running `npm link` under `packages/cli` with running `lerna link` under the repo root?

ang-zeyu:
`npm link` is only for executing `markbind xx` commands globally for development, [`lerna link`](https://github.com/lerna/lerna/tree/main/commands/link) does not do the same

`npm run setup` executes [`lerna bootstrap`](https://github.com/lerna/lerna/tree/main/commands/bootstrap) which does installation + also cross-package symlinking

ang-zeyu:
labelling this as `high` now that 3.0 is rather stable, we could really use better type safety going forward to prevent silly regressions and make variables / fields a little more readable

ang-zeyu:
@wxwxwxwx9 @jonahtanjz @raysonkoh @ryoarmanda If anyone wants to tackle this issue in the upcoming semester, do drop a note here. Would advise to start now as well as its a massive change that will conflict with other PRs! (would be much smoother to iterate on it now)

ryoarmanda:
[...]

PR #1877, opened 2022-04-05
Adopt TypeScript for core package

ryoarmanda:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [x] Others, please explain: Project development upgrade

First phase of #1421

**Overview of changes:**
(Edited as of 17/04)
Some of the `.js` code files in `core` have been renamed as `.ts` files and adapted into TypeScript as a starter.

See the documentation on migrating to TypeScript: https://deploy-preview-1877--markbind-master.netlify.app/devguide/migratingtotypescript

Workflow of `core` and by extension the whole project are slightly changed: 
- Development 
  - Three new scripts are added in root, `npm run build:backend` to build the TypeScript files, `npm run clean` to remove built files, and `npm run dev` to start the development environment which builds the files and automatically rebuilds on changes
  - IDEs like WebStorm, VSCode, natively supports rebuild on change too, so developers who use those IDEs can sidestep `npm run dev`.
  - Three new scripts are added in core, `npm run build` to build the TypeScript files, `npm run clean` to remove built files, and `npm run prepare` which is explained in Installation and Publishing below
- Installation
  - A new lifecycle script `prepare` is added in core which does `npm run build` and automatically runs as part of installation of the package (`npm run setup`). 
- Debugging
  - Builds will emit sourcemaps (`.js.map`) which provides the ability to directly debug `.ts` files (e.g. put breakpoints, etc)
- Linting
  - New package `typescript-eslint` is integrated to `eslint` to enable lint on `.ts` files
  - `npm run lint` now lints `.ts` files
- Testing
  - New package `ts-jest` is integrated to `jest` to perform direct unit tests to `.ts` files without the need to manually build them
  - However, functional tests with `cli` will require for the files to be built first
- Publishing
  - Builds will emit declaration files (`.d.ts`) and their maps (`.d.ts.map`) for type-checking and definitions support across projects
  - The lifecycle `prepare` script in core will automatically run as part of publishing of the package (`lerna publish`) 

- Comments on the approach taken are appreciated.

**Testing instructions:**
Setup test: `npm run setup` should compile the TypeScript files.
Development test: `npm run test` should be fine.
Development usage test (with linked `markbind-cli`): Any `markbind` commands should be fine.

Adopt TypeScript for core package

The dynamic typing nature of JavaScript can create avenues for subtle
bugs during runtime related to parameter types. Strict typing system,
like TypeScript, can address such cases and benefit development of
MarkBind.

Let's introduce TypeScript to the core package to improve the quality
and security of our codebase. Furthermore, let's adapt starter files
to TypeScript as an example, and document on the steps to migrate other
files to TypeScript for developers.

---

- [x] No unrelated changes


.gitignore
```diff
@@ -50,3 +50,7 @@ packages/cli/test/functional/*/_site
 # vscode configuration
 .vscode/
 *.iml
+
+# Compiled files from TypeScript
+*.js.map
```
ang-zeyu, line 55: > The .js files are still kept in repository for easier transitional change, but after this PR those files will slowly be removed and excluded.

I think we should directly *rename* the files for the following reasons:
- duplicating wipes `git blame/annotate` information
- the diff in the PR becomes much more reviewable (from what I can see, most of the files seem to be duplicates, but there are some minor differences like in the test files which are very difficult to spot) and valuable

> easier transitional change

I'm not getting this part though, let me know if I'm missing something 🤔

packages/core/package.json
```diff
@@ -21,6 +22,9 @@
     "access": "public"
   },
   "scripts": {
+    "build": "tsc",
+    "clean": "tsc --build --clean",
+    "dev": "tsc-watch --noClear",
```
ang-zeyu, line 27: should we utilise `tsc-watch` as intended and string up some command development commands in the root `package.json` for use?

e.g.
- `tsc-watch .... --onSuccess node ......./cli/index.js ./docs` (docs development)
- maybe also one for the default template site

to reduce the slightly more cumbersome 2-step process to one

otherwise, also somewhat equivalent to just using [`tsc --watch`](https://www.typescriptlang.org/docs/handbook/compiler-options.html). (`tsc-watch` is meant as a typescript substitute for [nodemon](https://www.npmjs.com/package/nodemon))

ang-zeyu (commented):
Nice work on this huge environment change 👍, just a brief look:

---

> Some files are not yet duplicated to .ts, would like some assistance on whether we should do it or keep it as-is:

Ditto on the 2nd.

As for the other files, I'm thinking - as long as we've adopted it from somewhere considerably in its original code style, we keep it as is for now, since we may need to cross-refer the original external implementation. (e.g. convert `HighlightRule` but not `double-delimiter.js`)

As for the "midway" ones like `markdown-it/index.js` which are more subjective, I'm leaning toward converting them for now as type annotations may be helpful for development.

> Publishing requires an additional manual build step because the build returns an error exit code (due to the files not adopting the type system yet), meaning we can't put this as a part of the npm lifecycle methods (it will abort as error). Added documentation on projectManagement.md for this.

👍

.gitignore
ryoarmanda: The decision to duplicate is based from https://github.com/MarkBind/markbind/issues/1421#issuecomment-1055225798. Originally I thought to just swap (rename) the `.js` to `.ts`, but maybe in the interest of reducing merge conflicts on other PRs as raised in the discussion, the `.js` files are still kept for now, and would slowly be deleted as the migration goes.

However, I also agree with your point on preserving git annotates, and I think workflow wise it should be better to just swap it as `.ts` so we don't have to keep maintaining the `.js` and `.ts` to perfectly match. Would it be okay for me to just rename them?

jonahtanjz: Agree with this. I think renaming is better as well for the reasons above. This will also be less confusing when working on the core packages (whether to make changes in the `.js` or `.ts` files).

packages/core/package.json
ryoarmanda: I have thought of this, but I am not sure on how taxing would it be when MarkBind commands are repeatedly invoked in short bursts over the course of development. For example, `serve` needs time to build, and it's not ideal for the script to call it in every added line when I am in the middle of implementing a function that takes some time to develop, it might just hog resources with only a little convenience benefit.

However, I am admittedly not too experienced in how `tsc-watch` does things (or `nodemon` too), so I am also not sure whether this is a real issue 🤔 but I have tried this with just `markbind --version` and it is invoked multiple times as I write some lines at a relaxed pace (the cutoff time delay is ~1 second).

.gitignore
ang-zeyu: > The decision to duplicate is based from https://github.com/MarkBind/markbind/issues/1421#issuecomment-1055225798.

I see, I actually meant leveraging typescript's js inter-op (much like you're doing for the patched files currently) if possible.

- rename one/a few starter files to `.ts` to test tsc in this PR
- the rest in subsequent PRs (where we can even proceed with the typescript code conversion too)

This reduces conflicts with other PRs since we're still dealing with the same `.js` files. Duplicating won't help here as the updates from other PRs need to be transferred to the `.ts` files here constantly (and vice versa in other PRs once this was merged).

... This comment was in the middle of the semester though, (thankfully) not much PRs left touching the core code at this juncture so I don't think we need to worry about this too much. 😌 

> Some files are not yet duplicated to .ts, would like some assistance on whether we should do it or keep it as-is:

Renaming all at once sounds good too, especially since you've already given quite some thought to which files should be converted.

> Would it be okay for me to just rename them?

Let's do that 👍

packages/core/package.json
ang-zeyu: Good point, I think we can keep the 2 commands separate for now and see how things go. (also just realised `--onSuccess` and the like won't work without having resolved all the typescript errors)

In that sense, would `tsc --watch` be sufficient for now? (over `tsc-watch`)

docs/devGuide/projectManagement.md
```diff
@@ -88,6 +88,17 @@ For general best practices, refer to the guide [_Working with PRs_ @SE-EDU](http
    * The end result of this command is version commit with an appropriate tag. We will make use of the generated tag and commit message later.
    </box>
 
+1. **Build the core package** by executing `npm run build:core` in the root directory.
+
+   The script would compile the TypeScript files into JavaScript files (`.js`), sourcemaps (`.js.map`), and type declarations (`.d.ts`), which would be packed when `npm` publishes this package.
```
jonahtanjz, line 93: Should all these generated files be committed to the repo? If not, we need some ways to ignore all these files when doing the release commit.

docs/devGuide/settingUp.md
```diff
@@ -43,6 +43,9 @@ We recommend the **WebStorm IDE** for working with MarkBind code.
    <popover content="Under the hood, this calls `npm ci` and `lerna bootstrap`">`npm run setup`</popover>
    in the **root folder** of your cloned repo.
 
+1. **Start the development environment** with <popover content="This will compile the `core` package for local use and automatically recompile if there are any changes in there.">`npm run dev`</popover>
+   in the **root folder** of your cloned repo.
```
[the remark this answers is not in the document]
ryoarmanda: That would also be good, but newly-cloned projects actually need to do a one-time build of the TypeScript files in order to properly run the linked MarkBind (the `cli` package cannot directly execute `.ts` files), so minimally `npm run build:core` must be done during setting up.

I guess it's better to put `npm run build:core` here and move `npm run dev` to the workflow page, what do you think?

docs/devGuide/projectManagement.md
ryoarmanda: The generated files are already in `.gitignore` so hopefully it won't get committed.

docs/devGuide/settingUp.md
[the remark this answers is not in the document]
jonahtanjz: Yup sounds good 👍

packages/core/package.json
ryoarmanda: Yeah, it would probably be `--onCompilationComplete` if we want to achieve this in the middle of migration.

I use the package `tsc-watch` just for the `--noClear` option for now, though it can be discussed whether it would be necessary for us. Just my 2 cents, in my command prompt, I don't quite like how `tsc --watch` clears the screen when it starts, so the previous command results and all are gone. I think it's better for the history to still be viewable by devs. What do you think?

docs/devGuide/projectManagement.md
jonahtanjz: Oh i was mainly referring to the generated `.js` files from their corresponding `.ts` files. I noticed that these are not ignored in the current `.gitignore`. Another potential issue is that we can't simply ignore all `.js` files in the `core` folder as there will be some files that will not be converted to typescript (as discussed above). Wondering if there is a better way to exclude these files?

ryoarmanda: I see :+1: I just updated the `.gitignore` to compile, the rules are the same as what I have input in `.eslintignore` (ignore all JS files except those we know are still directly written in JS).

ryoarmanda:
@jonahtanjz @ang-zeyu I found a way to suppress the error exit code when running the build, so now I can put the build script as part of the npm's lifecycle scripts, but it's admittedly quite hacky (just string the command with `exit 0` haha). It will run in both `npm install` (which is invoked by our `npm run setup`) and `npm publish`, so there should be no manual builds needed in both setting up and publishing. I'm not entirely happy of the workaround, but I think it's the best we can do in the middle of migration.

However, it still outputs a ton of TypeScript errors as part of the job (see the `npm run setup` log in this PRs checks). I was thinking to pipe output to null to discard it, but seems there is no easy cross-platform way to do it (in Windows we use NUL, Linux we use /dev/null, etc)...

jonahtanjz:
> I found a way to suppress the error exit code when running the build, so now I can put the build script as part of the npm's lifecycle scripts, but it's admittedly quite hacky (just string the command with exit 0 haha)

Will this suppress legitimate errors from typescript as well (e.g some files failed to compile)? If it does, the developer doing the release may not be aware of the errors and broken/missing files may get published? Not sure if this is a potential issue?

ryoarmanda:
> Will this suppress legitimate errors from typescript as well (e.g some files failed to compile)? If it does, the developer doing the release may not be aware of the errors and broken/missing files may get published? Not sure if this is a potential issue?

Good concern. I just found out that even if you write a broken `.ts` file, the compiler will still compile it to `.js`, but the error will be carried to the compiled file.

```
const test: string = [].map(x => ;     // compiled as "const test = [].map(x => ;"
```

To guard against this, I think we can chain the lifecycle scripts with `npm run test` for safety. (As an aside, given that we have some manual steps in our current publish workflow, I think we can enhance `lerna publish` to add safety lifecycle scripts when running it to catch any things that might have gone past us, and maybe add a script to do a [dry run](https://github.com/lerna/lerna/issues/51#issuecomment-705820466) of the publish too)

---

For context, the reason why I went with the hack is that our checks fail when core is not yet compiled (see https://github.com/MarkBind/markbind/pull/1877/commits/94b09a9f445800bff678c7c7e56b83807ef3223a). There are several places that I might insert the build script in, I choose to put it as an npm lifecycle script as it's a "two birds one stone" solution, `prepare` runs on both `npm install` and `npm publish` (refer [here](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install)). But on either choice, the checks would still fail as the build would give a non-zero exit code. Suppressing the exit code may be necessary.

ryoarmanda:
An aside comment here, I think some of our issues can be resolved if we can migrate `cli` into TypeScript too, so we can be in a fully TypeScript environment (well, except the patched libraries)

.gitignore
ryoarmanda: Ahh I understand your point. Sorry that I read it a bit differently 😅 

This makes me rethink my approach though. My current strategy is to rename everything and fix the errors in separate PRs as we go along. While it's easier to fully change the environment first and adapt to it incrementally, this leaves us with issues to handle in the middle of the adaptation (from files riddled with red squiggles, build does not give a zero exit code, publishing concerns, etc).

I think from reading your point, I found a different way to approach it. In separate PRs, we incrementally rename the files _and_ fix the errors, so we make sure that it's successfully compiled without errors and avoiding the issues we have here.

I feel like I would revise the strategy to that, so this PR sets up the workflow and also converts (and fixes) some starter files, like the tests and utils. Would this be a better way to go for this?

(cc @jonahtanjz so everyone is on the same page)

packages/core/package.json
ang-zeyu: Would all compilation errors be reported again by `tsc --watch`? (or just incremental ones) I personally don't have much preference if so, since we should only need one `tsc` process for all packages. 

If not, I think `tsc-watch` is definitely the way to go 👍 (fine with either though, if it might be useful to some developers let's leave it as is)

ryoarmanda:
> Would this work?
> 
> * commit A - do the renaming
> * commit B - do the changes

Sorry, can you clarify about this? Not sure which this is targeted to (in the working branch / target branch). I mentioned that splitting the rename and changes into separate commits can still retain the history in the working branch, but when diffed to the target branch, the history is lost.

Assuming you mean those commits are in the target branch, it could be possible. I just tried to play around in my code playground, I noticed that the history is lost only when using the squash strategy. When merged with a normal merge commit, the history is retained in the target branch.

I think the key here is that we must keep those two commits separate in the target branch. I guess this calls for a multi-commit PR with a normal merge commit? I might need to get this organized if that's the case.

> * Dependency setup (tsconfig, package.json, ......)
> * Rename ts starter files
> * Adapt ts starter files
> * Adapt / add test dependencies and some starter test files
> * Development docs

:+1: This structure is a good start. Should I follow the rename-adapt commits for the test files as well (assuming we also want to retain the history for test files)? Or I can also put the renaming of test files on the same commit as the renaming of code files?

Also just to confirm, I would structure this through soft reset and commits, but then I have to force push to here, right? Or do I need to create another PR for the clean one?

ang-zeyu:
> Sorry, can you clarify about this? Not sure which this is targeted to (in the working branch / target branch). I mentioned that splitting the rename and changes into separate commits can still retain the history in the working branch, but when diffed to the target branch, the history is lost.
> 
> Assuming you mean those commits are in the target branch, it could be possible. I just tried to play around in my code playground, I noticed that the history is lost only when using the squash strategy. When merged with a normal merge commit, the history is retained in the target branch.

No, I missed this part.
 
> I think the key here is that we must keep those two commits separate in the target branch. I guess this calls for a multi-commit PR with a normal merge commit? I might need to get this organized if that's the case.

Not getting this part with the above though, would a merge commit be subject to the same issue if so? 👀 (working -> target)

> Should I follow the rename-adapt commits for the test files as well (assuming we also want to retain the history for test files)? Or I can also put the renaming of test files on the same commit as the renaming of code files?
 
Feel free to try it out. Anything reasonable and clear enough (not too much in one commit) is fine : )

> Also just to confirm, I would structure this through soft reset and commits, but then I have to force push to here, right? Or do I need to create another PR for the clean one?

Yup. I've bumped the relevant "guide" from last year in telegram as well if you need it. (note the part on fixups too)

ryoarmanda:
> Not getting this part with the above though, would a merge commit be subject to the same issue if so? 👀 (working -> target)

I found out that it's not the case, that is the history is retained when the separate commits are merged with a merge commit. Here's a small demo I cooked up, where I tried

The file history in the working branch is like so ("Slowly adapt..." is the renaming commit, "Fully adapt..." is the fix commit, and the rest is history from master):

image

If I merge this branch to master with a squash commit, the history is lost (see the blame and git log --follow):

image

While if I merge it with a merge commit, the history is retained (see the blame and git log --follow):

image

ang-zeyu:
> I found out that it's not the case, that is the history is retained when the separate commits are merged with a merge commit. Here's a small demo I cooked up, where I tried to do both strategies.

Thanks for testing 👍, ok let's go for that then.

Fyi I was suggesting merge commit (otherwise no point in structuring the commits here). Though, still not too sure on it (`rename-threshold` note below)

There are a couple of alternatives we can try if that (github merge) fails too -- reverting the merge and do it manually:
- rebase (without squash) - not as neat as a merge commit, but should preserve history
- `git merge -X rename-threshold Y` (though, haven't researched `rename-threshold` heavily)

Hopefully github's merge commit is sufficient. :-)

packages/core/package.json
ryoarmanda: On further consideration, I think `tsc-watch` is not needed for now, and can opt for `tsc --watch` first. Turns out the clearing the screen thing is more on the terminal and not on the `tsc`'s shortcomings 😃

docs/devGuide/workflow.md
```diff
@@ -126,6 +126,22 @@ The sections below has more information about various stages of submitting a PR.
 
   </modal>
 
+#### Editing backend features
```
ang-zeyu, line 129: let's update the release guide as well with the relevant additional steps

tsconfig_dev.json
```diff
@@ -0,0 +1,6 @@
+{
```
ang-zeyu, line 1: hmm to check my understanding -- what's the separate `_dev.json` for?

tried replacing `npm run dev: "tsc --build --watch"`, seems to work too

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,131 @@
...
+   ------------------------|--------------------------|-----------------------
+   `const ... = require()` | `import ... = require()` | `import ... from ...`
+   `module.exports = {}`   | `export = {}`            | `export ...`
+   `module.exports = ...`  | `export = ...`           | `export default ...`
+
+   Start with the TypeScript equivalent syntax, as it is compiled to be exactly the same as the CommonJS syntax, making it the safest choice when adapting the file. You can read more about this syntax in the [official documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require).
+
+   There are some limitations to the TypeScript equivalent syntax, such as:
+   - There can only be one `export =` in a file
+   - You cannot reference what is not (unlike `module.exports`)
```
ang-zeyu, line 82: 👀

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,131 @@
...
+
+2. **Change the CommonJS module import/export syntax** to TypeScript's equivalent (use ES6 only if necessary).
+
+   A common error in newly renamed files comes from the way modules are imported/exported. The syntax for those can be converted according to the table below:
+
+   CommonJS                | TypeScript equivalent    | ES6
+   ------------------------|--------------------------|-----------------------
+   `const ... = require()` | `import ... = require()` | `import ... from ...`
+   `module.exports = {}`   | `export = {}`            | `export ...`
+   `module.exports = ...`  | `export = ...`           | `export default ...`
```
ang-zeyu, line 76: > Do not use the ES6 syntax if not necessary, as it compiles into something more than just the CommonJS syntax when compiled, and might lead to unexpected behaviours when another `.js` file imports the compiled file, to the point of failing established tests in some cases.

This particular line may not be equivalent (commonJs dosen't have concept of `default`), and might be the sole cause of your issues with above quote.

Try inserting the following in the playground [here](https://www.typescriptlang.org/play?module=1#code/PTAEAEGcAsHsHcCiBbAlgFwFAgs2ATAVwBsBTALlAGNZk8A7AK0mzHHtkQCcvYuXUyAA590oAN6gAbgENihUgHkAZgAVUoAL6hlvZKABEAOmA16kdDPrpIBgNyZMpAB4iuYsxdDp4sdaABeaTkFFX8AKlAAJgcnV1FQfFJlGRIxHz9UByA)

```
// @showEmit
// @module: commonjs
// @noErrors
import { valueOfPi } from "./constants";

export const twoPi = valueOfPi * 2;

export default twoPi;
```

ang-zeyu: I think one solution may be to advise against using `export default` in our **intra** (external libraries should be fine) exports / imports altogether (we've been mostly using named exports too which should make this easier)

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,131 @@
...
+   There are some limitations to the TypeScript equivalent syntax, such as:
+   - There can only be one `export =` in a file
+   - You cannot reference what is not (unlike `module.exports`)
+   - A module exported with `export =` must be imported with the TypeScript equivalent syntax (for `.ts` files) or the CommonJS syntax (for `.js` files).
+
+   Do not use the ES6 syntax if not necessary, as it compiles into something more than just the CommonJS syntax when compiled, and might lead to unexpected behaviours when another `.js` file imports the compiled file, to the point of failing established tests in some cases.
+
+   However, changing to the ES6 syntax might be necessary, e.g. you cannot use the TypeScript equivalent syntax if you want to import/export types or interfaces. When doing so, make sure the files pass the tests first before proceeding.
+
+3. **Adapt the files** to fully using TypeScript.
```
ang-zeyu, line 89:
```suggestion
3. **Adapt the files** fully to TypeScript.
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,131 @@
...
+
+   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Some tips on adapting to TypeScript:
```
ang-zeyu, line 91:
```suggestion
   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Here are some tips:
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,131 @@
...
+
+The solution to this is to make the rename and adapt commits separate, in order for Git to recognize first that the file has changed into `.ts`, and so the changes are compared not against the old `.js` file, but to the renamed `.ts` file instead.
+
+### Why do we need a merge commit
+
+Even if the migration developer has kept the history intact through the separate "Rename" and "Adapt" commits, this is only intact at their working branch. At the end of the day, the totality of the changes in the working branch is compared against the target branch, in which the same consideration would apply.
+
+If we do the usual squash commit, the two separate commits are combined into one, and so the problem with losing history occurs again, but now in the context of the target branch. Again, the solution is to keep those two commits separate, even in the target branch, and therefore what we must merge the PR with a merge commit to achieve this.
```
ang-zeyu, line 131: 👀 
Dosen't a merge PR still squash everything into one commit in the target branch? (hence the need for `rename-threshold`)

But preserving the history in the working branch, from which git can _possibly_ infer the rename.
(In contrast, squash removes the information from the working branch completely)

ang-zeyu (commented):
Nice work organizing the commits, a breeze to read through 👍 

Just in case the "guide" is unclear: you can deliver the changes using "fixups" in this way for a reviewer

E.g. if "add initial typescript support" and "add typescript migration and development docs" had comments,

```
Add initial typescript support
Rename starter files to TypeScript
Adapt starter files to TypeScript
Add TypeScript migration and development docs
# Add these 2 commits
Fixup: Add initial typescript support
Fixup: Add TypeScript migration and development docs
```

Once the fixups are reviewed again, you can use git's fixup feature or interactive rebase (my personal preference) to squash them back into 4 commits.

ryoarmanda:
Thanks for the weekend reviewing @ang-zeyu 🙇  I pushed some WIP stuff earlier haha. Some parts of the docs and the scripts are a little bit different now as I discovered more quirks along the way. Feel free to do another read through, and I will check out your comments in the meantime.

tsconfig_dev.json
ryoarmanda: If I can summarize, `tsconfig.json` is intended for production as it does not want to compile if there are errors as a [consequence](https://www.typescriptlang.org/docs/handbook/project-references.html#caveats) to the [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) approach, and `tsconfig_dev.json` is intended for development as it allows that to happen, facilitating tests in WIP (e.g. testing if the import/export syntax plays well with other files before moving on to fixing the types), albeit with a performance hit.

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+## Migration Steps
+
+The TypeScript migration process is a little different than normal development work, as maintainers would not follow the normal procedure of creating a squash commit. Instead, they will do a **normal merge commit**, and therefore migration developers will need to structure the commits in a particular way.
+
+We have decided on structuring the commits as two commits: a "Rename" commit and an "Adapt" commit. If you need further context on the approach, feel free to read the [Explanation Notes](#explanation-notes) section below.
+
+### Step 1: Preparation
+
+1. **If running, stop automatic compilation on change/save**. More on this at [Editing backend features](workflow.md#editing-backend-features) section in the Workflow page.
+   <br>
```
ryoarmanda, line 26: Will remove this

docs/devGuide/migratingToTypeScript.md
ryoarmanda: Agree with `export default`. I have already disabled `prefer-default-export` in ESLint, maybe we can strengthen it by enabling `no-default-export`. Does external libraries here mean the ones that will be consumed by other packages (e.g. `cli`)? If so, we still have to follow how `cli/index.js` imports `core/index.js` to see our options right now.

As a note, I was just about to write that I remembered I couldn't make ES6-style named exports (`export ...`) work in `CyclicReferenceError.ts` as it will always fail the tests, suggesting that it's not always possible to use ES6-style. But, as I investigated further to write this comment, I found that the root cause is that there is a file (`Context.js`) that imports the `CyclicReferenceError` class as a *default* export. Because `export default` is out of the question, therefore the only way for `CyclicReferenceError.ts` to fit in with other files is to use the TypeScript-style default export (`export = ...`).

I can update the text so that we can start from TypeScript-style, then move to ES6 *if possible*, but do not use `export default`.

docs/devGuide/workflow.md
ryoarmanda: As far as I remember, we don't need to do any additional things for release.

I have set core's npm lifecycle script `prepare` which will run as part of [`npm publish`](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-publish) (which is actually invoked by [`lerna publish`](https://github.com/lerna/lerna/tree/main/commands/publish#lifecycle-scripts)) to build the TypeScript files. This means that as we do `lerna publish`, it will build the files beforehand for the package before it is packed and published.

docs/devGuide/migratingToTypeScript.md
ryoarmanda: Hmm yeah, now that I think about it, I felt like I was only basing it like a fast-forward merge rather than an actual merge commit haha. Will update into it.

docs/devGuide/migratingToTypeScript.md
ang-zeyu: > Does external libraries here mean the ones that will be consumed by other packages (e.g. cli)?

Nope, just regular external libraries (e.g. `simple-git`).

> I can update the text so that we can start from TypeScript-style, then move to ES6 if possible, but do not use export default.

Sounds good 👍
(at least for now, while we're still migrating, default exports won't play so well with js)

> I found that the root cause is that there is a file (Context.js) that imports the CyclicReferenceError class as a default export.

> default export.

Rather, the `module.exports` object. (commonjs has no "default" export, by default 😅)

Just to be clear, this would actually work, but rather clunky-ly:

```
var x = require('ts-module').default;
```

Once we have almost all files converted we can possibly revisit if `export default` would be a good fit (if really needed).

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+2. **Change the CommonJS module import/export syntax** to TypeScript's equivalent syntax (use ES6 syntax only if possible and/or necessary).
+
+   A common error in newly renamed files comes from the way modules are imported/exported. The syntax for those can be converted according to the table below:
+
+   Case           | CommonJS                   | TypeScript equivalent                                            | ES6
+   ---------------|----------------------------|------------------------------------------------------------------|-----------------------
+   Default import | `const x = require(y)`     | `import x = require(y)`                                          | `import x from y`
+   Named import   | `const { x } = require(y)` | N/A, workaround:<br>`import z = require(y)`<br>`const { x } = z` | `import { x } from y`
+   Named export   | `module.exports = { x }`   | `export = { x }`                                                 | `export { x }`
+   Default export | `module.exports = x`       | `export = x`                                                     | `export default x`
```
ang-zeyu, line 78: referring https://github.com/MarkBind/markbind/pull/1877#discussion_r851785006

Just one nit as such while you're still touching things up -- let's make it clear the regular `import x = require(y)` / `export = x` isn't default import export. (Maybe just give a warning, and link the typescript docs / playground)

E.g. `import x from y` actually compiles to this:

ts

---

Remember to push fixups as well! https://github.com/MarkBind/markbind/pull/1877#pullrequestreview-943841714

docs/devGuide/workflow.md
```diff
@@ -126,6 +126,22 @@ The sections below has more information about various stages of submitting a PR.
...
+
+Some of our backend code files in `packages/core` are written in TypeScript, and you will need to compile those into JavaScript for local execution with `packages/cli`.
```
ang-zeyu, line 131:
```suggestion
Some of our backend code files in `packages/core` are written in TypeScript, and you will need to compile those into JavaScript for local execution with our command-line module `packages/cli`.
```

docs/devGuide/workflow.md
```diff
@@ -126,6 +126,22 @@ The sections below has more information about various stages of submitting a PR.
...
+
+You can run `npm run build:backend` in the root directory to compile the files, but in some cases, it might be tedious to manually execute the command many times. We recommend you to either:
+
+1. Run `npm run dev` in the root directory. (_Recommended for TypeScript migration_)
+
+   This command starts the compiler's file watcher which will detect file changes and rebuild all of the files if there are changes.
```
ang-zeyu, line 137:
```suggestion
   This command starts the compiler's file watcher which will rebuild the relavant files when file changes are detected.
```

docs/devGuide/workflow.md
```diff
@@ -126,6 +126,22 @@ The sections below has more information about various stages of submitting a PR.
...
+
+1. Configure your IDE to perform automatic compilation on file change/save. (_Recommended for general development_)
+
+   Refer to your IDE's guides to set this up. For instance, here are the guides for [WebStorm](https://www.jetbrains.com/help/webstorm/compiling-typescript-to-javascript.html#ts_compiler_compile_code_automatically) and [Visual Studio Code](https://code.visualstudio.com/docs/typescript/typescript-compiling#_step-2-run-the-typescript-build).
+   
+   Note the TypeScript configuration used for development (`tsconfig_dev.json`) is different than the one for production (`tsconfig.json`), so you might need to modify the IDE's `tsc` command to include the option `-p tsconfig_dev.json`.
```
ang-zeyu, line 143:
```suggestion
   Note that the TypeScript configuration used for development (`tsconfig_dev.json`) is different than the one for production (`tsconfig.json`), so you might need to modify your IDE's `tsc` command to include the option `-p tsconfig_dev.json`.
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+   However, in some cases, changing to the ES6 syntax might be necessary. For instance, you cannot use the TypeScript equivalent syntax if you want to import/export TypeScript specific constructs, such as types, interfaces, and enums. Again, when doing so, make sure that all tests pass.
+
+   <box type="info">
+
+   We will perform a mass change into ES6 syntax after all files in core have been adapted into TypeScript.
+   </box>
+
+3. **Adapt the files** to fully use TypeScript.
+
+   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Only fix what is necessary and do not modify any code functionality. Avoid using `any` as best as you can.
```
ang-zeyu, line 95:
```suggestion
   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Only fix what is necessary and be careful with accidentally modifying any code functionality. Avoid using `any` as best as you can.
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+
+   If you happen to encounter type errors related to using names from an internal dependency that is still in JavaScript (`.js`), you can try to infer simple types in your file first as a "stand-in" of the more robust types when that internal dependency is eventually migrated.
```
ang-zeyu, line 97:
```suggestion
   If you happen to encounter type errors related to using names from another MarkBind import that is still in JavaScript (`.js`), you can try to infer simple types in your file first as a "stand-in" of the more robust types when that internal dependency is eventually migrated.
```

> related to using names

👀 what's this ("names")?

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+
+   On the flip side of the above situation, as you now have developed a robust type for your own file, you might have to **adjust how other TypeScript (`.ts`) files type names referenced from your files**. Some type declarations might need to be made stricter, or can be removed if it's not necessary anymore, and so on. This will ensure that the files will be incrementally stricter.
```
ang-zeyu, line 99:
```suggestion
   On the flip side of the above situation, once you have developed a robust type for your own file, **adjust how other TypeScript (`.ts`) files type names referenced from your files** as applicable. Some type declarations might need to be made stricter, or can be removed if it's not necessary anymore, and so on. This will ensure that the files will be incrementally stricter. 
```

ang-zeyu (commented):
wording nits:

docs/devGuide/migratingToTypeScript.md
ryoarmanda: Apologies if my definitions might lead to some confusion, but to clarify, the "default" import/export I'm mentioning here is rather more to the concept of importing/exporting a single (non-object) thing, such a constant, a function, a class, etc. It is why I associate `const x = require(y)`, `import x = require(y)`, and `import x from y` together.

I have also noted in the docs that, as exampled in the image above, the ES6 syntax will compile into something more than just CommonJS syntax, and thus proceed with ES6 only if possible. I guess what I'm trying to say is that while they are doing the same job at high-level (import/export a single thing), they are not identical at low-level.

I will update the docs to clarify this more.

ang-zeyu: maybe:
- "module.exports object"
- "Entire export"? (still seems ambiguous)
- "`export` object"

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+
+   Case           | CommonJS                   | TypeScript equivalent                                            | ES6
+   ---------------|----------------------------|------------------------------------------------------------------|-----------------------
+   Default import | `const x = require(y)`     | `import x = require(y)`                                          | `import x from y`
+   Named import   | `const { x } = require(y)` | N/A, workaround:<br>`import z = require(y)`<br>`const { x } = z` | `import { x } from y`
+   Named export   | `module.exports = { x }`   | `export = { x }`                                                 | `export { x }`
+   Default export | `module.exports = x`       | `export = x`                                                     | `export default x`
+
+   Similar to the process of refactoring a function, you might have to **adjust how these files are imported by other TypeScript (`.ts`) files**. If you are changing it to the TypeScript equivalent syntax, then it must be imported with so as well. The same goes for the ES6 syntax.
```
jonahtanjz, line 80: Not sure if there is a typo here?
```suggestion
   Similar to the process of refactoring a function, you might have to **adjust how these files are imported by other TypeScript (`.ts`) files**. If you are changing it to the TypeScript equivalent syntax, then it must be imported the same way as well. The same goes for the ES6 syntax. 
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+     It's recommended to try to match the `@types/<lib-name>` version with the associated module. In general, try to match the major and minor version as closely as possible.
+     </box>
+
+3. **Stash the changes** (`packages/core/package.json` and root `package-lock.json`).
+   <br>
+   This will be added back during the "Adapt" commit later.
+
+### Step 2: "Rename" Commit
+
+1. **Add the paths to the `.js` files to `.gitignore` and `.eslintignore`**.
```
jonahtanjz, line 51:
```suggestion
1. **Add the paths of the `.js` files to `.gitignore` and `.eslintignore`**.
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Only fix what is necessary and do not modify any code functionality. Avoid using `any` as best as you can.
+
+   If you happen to encounter type errors related to using names from an internal dependency that is still in JavaScript (`.js`), you can try to infer simple types in your file first as a "stand-in" of the more robust types when that internal dependency is eventually migrated.
+
+   On the flip side of the above situation, as you now have developed a robust type for your own file, you might have to **adjust how other TypeScript (`.ts`) files type names referenced from your files**. Some type declarations might need to be made stricter, or can be removed if it's not necessary anymore, and so on. This will ensure that the files will be incrementally stricter. 
+
+4. **Make sure everything is in order** by running `npm run test` in the root directory.
+
+   The `core` package test directly tests `.ts` files (powered by `ts-jest`), and the `cli` package test uses the compiled files from `core`. Therefore, passing the `cli` tests means that it is very likely the compiled files work as expected.
```
jonahtanjz, line 103:
```suggestion
   The `core` package test can directly run on the `.ts` files (powered by `ts-jest`), and the `cli` package test uses the compiled files from `core`. Therefore, passing the `cli` tests means that it is very likely the compiled files work as expected.
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Example of Migrated Works
+
+You can see these pull requests to observe the finished migration works:
+
+- [#1877: Adopt TypeScript for core package](https://github.com/MarkBind/markbind/pull/1877)
+
+## Explanation Notes
+
+### Why do we need separate "Rename" and "Adapt" commits
```
jonahtanjz, line 121:
```suggestion
### Why do we need to separate "Rename" and "Adapt" commits
```

docs/devGuide/migratingToTypeScript.md
```diff
@@ -0,0 +1,135 @@
...
+
+We want to keep the file history intact throughout this migration, meaning that we should still be able to see the change history of the `.ts` files even from the time when they were still in `.js`.
+
+When we rename files, we have to keep the Git similarity index between the old file and the new file to be above a certain threshold (50%) to keep the history, or else it will be regarded as a different file with its own history.
```
jonahtanjz, line 125:
```suggestion
When we rename files, we have to keep the Git similarity index between the old file and the new file above a certain threshold (50%) to keep the history, or else it will be regarded as a different file with its own history.
```

jonahtanjz (commented):
Couple of suggested changes

docs/devGuide/migratingToTypeScript.md
ryoarmanda: I think there is no agreeable term for the concept. Most just simply say one is to "export one thing" and the other to "export many things". There are developer-written articles (like [this](https://www.freecodecamp.org/news/node-module-exports-explained-with-javascript-export-function-examples/) and [this](https://www.sitepoint.com/understanding-module-exports-exports-node-js/)) kind of use the same definition as I wrote earlier, "export one thing" = "default export" in CommonJS context.

docs/devGuide/migratingToTypeScript.md
ryoarmanda: It's my attempt to generalize referring to constants, properties, members, etc. The case that I saw that , was that a file in migration is trying to access a static class member of another file that is still in `.js`. We have to make do for now inferring simple types like `Record<string, any>` for example while we're waiting for the other `.js` file to be migrated.

docs/devGuide/migratingToTypeScript.md
ryoarmanda: I just updated the docs to clarify this, and since it is becoming too long I put the syntax reference on a separate section to be referred to from the migration steps. I opt for the simpler "export one thing" vs "export many things" to describe the different cases.

ryoarmanda:
I'd like to ask for advice about the migration specifics:

- (Managing) Should the migration be done in-order or free-to-pick? It would be easier to migrate things in topological order so we don't have to spend time guessing types from other files that we haven't migrated yet (which could also be hard at times like for complex objects), but it might take longer time as now we have a bottleneck (have to migrate x first before y).

As a reference, here's our internal dependency graph for core. It's a huge image so you can click on it and zoom on it. If going for topological order, the green ones are our starting points.

test

- (Adapting) Some types from external libraries often have optional properties declared, like for htmlparser2's nodes the `name` property is actually optional (i.e. could be undefined). Right now we just use `node.name` in many of our functions as if it's a defined property. When adapting to TypeScript, this will cause many errors. To fix it, we have three choices: 
  1. Have type validation on every function call (the most secure way, but leads to repeated checks)
  2. Have validation on a high-level call, then assert that the properties are non-null (the easiest way, but is not good for security)
  3. (Just thought of this) Devs selectively remove the optional condition on properties which we have validated are non-null with some TypeScript magic. Not sure on this.

docs/devGuide/migratingToTypeScript.md
ang-zeyu: > I think there is no agreeable term for the concept. Most just simply say one is to "export one thing" and the other to "export many things". There are developer-written articles (like [this](https://www.freecodecamp.org/news/node-module-exports-explained-with-javascript-export-function-examples/) and [this](https://www.sitepoint.com/understanding-module-exports-exports-node-js/)) kind of use the same definition as I wrote earlier, "export one thing" = "default export" in CommonJS context.

No, do refer to a canonical, or at least peer reviewed source. https://nodejs.org/api/modules.html, https://stackoverflow.com/questions/21117160/what-is-export-default-in-javascript, https://stackoverflow.com/questions/40294870/module-exports-vs-export-default-in-node-js-and-es6

There is no concept of "default import/export" in commonJs, this is a very common misconception that the first article you linked has clearly made as well. The second article refers to the vague concept of a "default value", not "default import/export" (which is a technical term formalised in es6 proposals).

This is important as its a very common misconception, one that I made when I started learning about this as well, and as you mentioned has very real implementation implications. We have contributors who aren't as familiar with module systems coming in; Developer documentation above all else should aim to be technically accurate and unambiguous.

The canonical term here would simply be `module.exports` or `exports` object (https://nodejs.org/api/modules.html).

Default exports, are in fact, closer to the concept of named exports in the context of commonJS. https://stackoverflow.com/questions/40294870/module-exports-vs-export-default-in-node-js-and-es6

> I opt for the simpler "export one thing" vs "export many things" to describe the different cases.

Sounds better, and good work here 👍 

Some minor suggestions dropped below

docs/devGuide/migratingToTypeScript.md
```diff
@@ -66,41 +65,27 @@ We have decided on structuring the commits as two commits: a "Rename" commit and
...
-
-   Start with the TypeScript equivalent syntax, as it is compiled to be exactly the same as the CommonJS syntax (except the named import case), making it the safest choice when adapting the file. You can read more about this syntax in the [official documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require).
+2. **Change the CommonJS module import/export syntax** to TypeScript's equivalent syntax (use ES6 syntax only if possible).
 
-   The ES6 syntax compiles into something more than just the CommonJS syntax when compiled, and might lead to unexpected behaviours when another `.js` file imports the compiled file, to the point of failing tests in some cases. You can try changing to the ES6 syntax, but only proceed if there are no failing tests. The TypeScript equivalent syntax is enough for the initial adaptation.
+   A common error in newly renamed files comes from the way modules are imported/exported. Those statements can be converted according to the [Import/Export Syntax Reference](#importexport-syntax-reference) section below. The sequence is as follows:
```
ang-zeyu, line 70: > ~sequence is~ steps are as follows

(for clarity that these are steps that should be followed, not a sequence of X)

docs/devGuide/migratingToTypeScript.md
```diff
@@ -66,41 +65,27 @@ We have decided on structuring the commits as two commits: a "Rename" commit and
...
 
-   However, in some cases, changing to the ES6 syntax might be necessary. For instance, you cannot use the TypeScript equivalent syntax if you want to import/export TypeScript specific constructs, such as types, interfaces, and enums. Again, when doing so, make sure that all tests pass.
+   1. Start by changing the import/export statements into TypeScript equivalent syntax. Be careful with the imports as you have to match them with the export syntax that the modules use, be it TypeScript equivalent or ES6.
```
ang-zeyu, line 72: > Be careful with ~the~ imports, as you have

docs/devGuide/migratingToTypeScript.md
```diff
@@ -66,41 +65,27 @@ We have decided on structuring the commits as two commits: a "Rename" commit and
...
 
-   <box type="info">
+   2. Check whether it is possible to change the export use ES6 syntax. As mentioned in the syntax reference, ES6 should not be used if the module only exports a single thing. Change to ES6 syntax only if possible, otherwise keep with TypeScript equivalent syntax.
```
ang-zeyu, line 74: > export use ES6 syntax

> export to use ES6 syntax

docs/devGuide/migratingToTypeScript.md
```diff
@@ -66,41 +65,27 @@ We have decided on structuring the commits as two commits: a "Rename" commit and
...
 
-   We will perform a mass change into ES6 syntax after all files in core have been adapted into TypeScript.
-   </box>
+   3. You might have to **adjust how the files are imported by other TypeScript (`.ts`) files**. Before, you would see that some If you are changing the export to the TypeScript equivalent syntax, then it must be imported the same way as well. The same goes for the ES6 syntax.
```
ang-zeyu, line 76: > You might have to **adjust how the files are imported by other TypeScript (`.ts`) files**. Before, you would see that some If you are changing the export to the TypeScript equivalent syntax, then it must be imported the same way as well. The same goes for the ES6 syntax.

> You might have to **adjust how the files are imported by other TypeScript (`.ts`) files**. ~Before~Previously, you would see that ~some~ if you are changing the export to the TypeScript equivalent syntax, then it must be imported the same way as well. The same goes for the ES6 syntax.

docs/devGuide/migratingToTypeScript.md
```diff
@@ -66,41 +65,27 @@ We have decided on structuring the commits as two commits: a "Rename" commit and
...
 
-3. **Adapt the files** to fully use TypeScript.
+3. **Adapt the files** fully to TypeScript.
 
-   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Only fix what is necessary and do not modify any code functionality. Avoid using `any` as best as you can.
+   The errors from TypeScript and `typescript-eslint` can guide you on what to fix. Only fix what is necessary and be careful with accidentally modifying any code functionality. Avoid using `any` as best as you can.
 
-   If you happen to encounter type errors related to using names from an internal dependency that is still in JavaScript (`.js`), you can try to infer simple types in your file first as a "stand-in" of the more robust types when that internal dependency is eventually migrated.
+   If you happen to encounter type errors related to using names from another MarkBind import that is still in JavaScript (`.js`), you can try to infer simple types in your file first as a "stand-in" of the more robust types when that internal dependency is eventually migrated.
```
ang-zeyu, line 82: > related to using names from another MarkBind import

> related to using names (of constants, properties, members, etc.) from another MarkBind import that is still written in JavaScript

docs/devGuide/migratingToTypeScript.md
```diff
@@ -116,13 +101,49 @@ You can see these pull requests to observe the finished migration works:
 
 - [#1877: Adopt TypeScript for core package](https://github.com/MarkBind/markbind/pull/1877)
 
+## Import/Export Syntax Reference
+
+Here is the reference on changing the import/export statements from CommonJS syntax into those that TypeScript support: TypeScript equivalent syntax, and ES6 syntax.
+
+The TypeScript equivalent syntax is designed to compile to the exact same code as CommonJS. You can read more about this syntax in the [official documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require).
+
+The ES6 syntax is the newer JavaScript syntax that TypeScript recommends, but in certain cases it does not compile to the exact same code as CommonJS.
```
ang-zeyu, line 110: > ~The~ ES6 syntax is ~the~a newer ... exact same code as CommonJS. We simplify the concepts here for illustration, refer to the StackOverflow post [here](https://stackoverflow.com/questions/40294870/module-exports-vs-export-default-in-node-js-and-es6) for more details!

docs/devGuide/migratingToTypeScript.md
```diff
@@ -116,13 +101,49 @@ You can see these pull requests to observe the finished migration works:
...
+
+In the tables below, a "thing" is defined as a JavaScript construct that can be exported, such as constants, functions, classes, etc. In the ES6 syntax, a "thing" can also be TypeScript constructs like types, interfaces, enums, etc.
```
ang-zeyu, line 112: > In ~the~ ES6 syntax

docs/devGuide/migratingToTypeScript.md
```diff
@@ -116,13 +101,49 @@ You can see these pull requests to observe the finished migration works:
...
+
+!!**Exports**!!
+
+| What to Export              | CommonJS                         | TypeScript equivalent   | ES6                   |
+|-----------------------------|----------------------------------|-------------------------|-----------------------|
+| One thing only              | `module.exports = x`             | `export = x`            | `export default x`*   |
```
ang-zeyu, line 118: 👀  would footnotes look better / ok? (on the "*" and "^"'s)

could also try bolding it if not to make it stand out from normal text a little bit more

docs/devGuide/migratingToTypeScript.md
```diff
@@ -116,13 +101,49 @@ You can see these pull requests to observe the finished migration works:
...
+| Object of one/more things   | `module.exports = { x, y, z }`   | `export = { x, y, z }`  | `export { x, y, z }`  |
+
+*: For compatibility, **do not use** `export default` during migration. It is compiled differently, and does not play well with dependants that are still in `.js` (unless some modifications are made in the `.js` files, which would be quite clunky).
```
ang-zeyu, line 121: > For compatibility, **do not use** `export default` during migration. It is compiled differently, and does not play well with dependants that are still in `.js`, unless some clunky modifications are made in the `.js` files (refer to the StackOverflow post previously [linked](https://stackoverflow.com/questions/40294870/module-exports-vs-export-default-in-node-js-and-es6)!).

docs/devGuide/migratingToTypeScript.md
```diff
@@ -132,4 +153,6 @@ The solution to this is to make the rename and adapt commits separate, in order
 
 Even if the migration developer has kept the history intact through the separate "Rename" and "Adapt" commits, this is only intact at their working branch. At the end of the day, the totality of the changes in the working branch is compared against the target branch, in which the same consideration would apply.
 
-If we do the usual squash commit, the two separate commits are combined into one, and so the problem with losing history occurs again, but now in the context of the target branch. Again, the solution is to keep those two commits separate, even in the target branch, and therefore what we must merge the PR with a merge commit to achieve this.
+If we do the usual squash commit, the changes from the two commits are combined into a new commit and only that commit will be pushed into the target branch. The original two commits are omitted, therefore the history of the working branch that we have tried to keep intact is stripped away.
+
+The normal merge commit also creates another commit (the merge commit), but the merge process interleaves the commits from the working branch to the target branch, therefore retaining the changes history.
```
ang-zeyu, line 158: > change~s~ history

ang-zeyu (commented):
on the first commit [300dbcf](https://github.com/MarkBind/markbind/pull/1877/commits/300dbcf55b1716dd357fff4ae540bb8c078f6a7d)

ang-zeyu:
No comments on the second fixup, nice work with tailoring all the linting configurations 👍

docs/devGuide/migratingToTypeScript.md
ryoarmanda: Some excess statement there 😅 removing the `Before, you would see that some` part instead.

ang-zeyu:
> I'd like to ask for advice about the migration specifics:
> 
> * (Managing) Should the migration be done in-order or free-to-pick? It would be easier to migrate things in topological order so we don't have to spend time guessing types from other files that we haven't migrated yet (which could also be hard at times like for complex objects), but it might take longer time as now we have a bottleneck (have to migrate x first before y).

Agreed. Though, I don't really have a preference on enforcing this. It might become apparent to anyone who tries to do some conversion that we should proceed in topological order. There might also be better ways at this we can't think of right now 🤔 

> As a reference, here's our internal dependency graph for core at the current point in this PR (only js files, leaving out lib, patches, ~and plugins~ edit: plugins should be included as well). It's a huge image so you can click on it and zoom on it. If going for topological order, the green ones are available files to migrate immediately after this PR.
> 
> test
> 
> * (Adapting) Some types from external libraries often have optional properties declared, like for htmlparser2's nodes the `name` property is actually optional (i.e. could be undefined). Right now we just use `node.name` in many of our functions as if it's a defined property. When adapting to TypeScript, this will cause many errors. To fix it, we have three choices:
>   
>   1. Have type validation on every function call (the most secure way, but leads to repeated checks)
>   2. Have validation on a high-level call, then assert that the properties are non-null (the easiest way, but is not good for security)
>   3. (Just thought of this) Devs selectively remove the optional condition on properties which we have validated are non-null with some TypeScript magic. Not sure on this.

> Right now we just use `node.name` in many of our functions as if it's a defined property.

I think one of the main motivations for adopting typescript is avoiding mistakes from this.

I'm favouring something along options 1 & 3 as such, to deal with this on a case-by-case basis.

```ts
// e.g.
const x = node.name as string; // forcibly "recast"
```

I think we might figure out better solutions as we go along in other conversion PRs though, I can't really say for sure right now.

One alternative worth sharing (but likely not applicable in this case) I came across in my previous internship was to override @types declarations in a custom `.d.ts` file. However, in that case, we had *full, logical verification* that the property was always going to be defined for our use cases. Something that might be an option too. (depends on the issue)

docs/devGuide/migratingToTypeScript.md
ryoarmanda: I'm not sure on the footnotes. I feel like these are important info that readers should understand, and putting it in our footnotes component hides it from view unless the reader intentionally hovers to it.

My intention for the symbols is to contextualize which syntax the info is referring to. In a way, it's like footnotes, but I just wanted to make the content more prominent than actual footnotes.

ang-zeyu: formatting nit to make the asterixes appear more prominent:

`<span style="font-weight:bold;">&nbsp;*</span>`

could also stick to `*` / one symbol to make it consistent.

ryoarmanda: :+1: I'll take the `^` symbol as it's more visible than `*`

ang-zeyu (approved):
Lgtm 👍 

Let's squash the fixups in and rebase this over the latest commit on the main branch (so it goes in "immediately after" - just to keep things neat)

ang-zeyu:
You can it merge in after if @jonahtanjz has nothing else to add : )

jonahtanjz (approved):
Nice work @ryoarmanda 👍 

LGTM :)

ryoarmanda:
Okay, will do the merge tonight. Will update on the merge results on whether we need `rename-threshold` (cc @ang-zeyu)

ryoarmanda:
@ang-zeyu To report: I believe the merge process is working well, but the matter of displaying the history differs from the situations below. GitHub history does not follow renames, WebStorm's history and annotate follow the rename, and VS Code (+ GitLens)'s has to modify a setting in order to follow renames.

Here it is in WebStorm checking `fsUtil`'s history and annotate (it is one of the adapted files the PR diff regards as "added" and not "renamed"):

image

image

In VS Code (+ GitLens):

image

image

(still not sure how to configure the annotate)

And in GitHub:

image

ang-zeyu:
Thanks for investigating @ryoarmanda, this is new to me as well.

I think not much we can do here since it depends on the git client involved. (unavoidable for github; for vscode since its a preference, we can leave it up to individual developers)

Noticed as well with cli `git show --diff-merges=on --summary d81eaee1` that certain files are not detected as renames. (though its ok with `git log -- filename`, up until the rename commit at least (`--follow` required for beyond that)).

Other git clients:
- gitkraken - ok by default
- sourcetree - same situation with cli (also has a "`--follow`" option)

For later merges, we could try a manual merge with --rename-threshold option mentioned earlier to see if it fixes the "`git show` issue". I think its ok to leave the current merge as-is since its a small offending number (3) of small files, and there's already work done over it from other PRs.

Issue #1903, opened 2022-04-14
PUML not working on CI deployment (Teammates) due to lack of graphviz


kaixin-hc:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

See [Teammates issue](https://github.com/TEAMMATES/teammates/issues/11732) and [reposense issue](https://github.com/reposense/RepoSense/issues/1759)

### Tell us about your environment

Live Deployed version

### MarkBind version

3.1.1

### What did you do? Describe the bug

After deployment, puml diagrams seem to be missing and have thrown an error.
![image](https://user-images.githubusercontent.com/59471320/163318797-3e6c8fef-e9e8-41a1-bc7a-441b32b55c13.png)

### Steps to reproduce

Unable to reproduce manually on local serve or on manual deployment on a forked branch

### Expected behavior

.puml diagrams should display

### Anything else?

Hypothesis: This is related to CI, and building .puml files during CI is the problem. CI may not have graphviz.

Observation: For MarkBind docs puml diagrams, we **actually never use the <puml> tags within the website**. Instead, we have saved the .puml as equivalent .pngs, and displayed the .png instead (see `diagrams.md`) Found this PR #917 explaining the reason is to avoid installing graphviz.

Graphviz related .puml errors: #1459, and the reposense: [#1406 Replace images as PUML diagrams PR](https://github.com/reposense/RepoSense/pull/1406) leading to #916 which was marked as fixed. Also consider reposense [1427 on the PUML diagrams not rendering](https://github.com/reposense/RepoSense/issues/1427)

Suggestions? What should we do about this issue?

tlylt:
Putting some details of the investigation that I have done here for reference:
- Graphviz does not seem to be required for sequence and activity diagrams, so it should be fine for anyone to refer to .puml files directly in CI (They might still need to ensure that Java is available for puml plugin to work). This is mentioned in the original [puml PR](https://github.com/MarkBind/markbind/pull/887#pullrequestreview-254595590), but was requested to change to "all diagrams", which is what we state in our UG at the moment. Perhaps we should consider specifying it as is (that some diagrams work out of the box)
  - ![image](https://user-images.githubusercontent.com/41845017/163331949-bf92fccc-fa66-41e4-b499-cb9c65499700.png)

damithc:
Thanks for the investigation so far @tlylt 
[...]

Issue #1863, opened 2022-03-29
Incorrect report of success when CLI commands actually fail


tlylt:
**Tell us about your environment**
Windows 10
* **MarkBind Version:**
3.1.1

**What did you do? Please include the actual source code causing the issue.**
When there is an error in site.json, or due to a mismatch of markbind version, the CLI commands output a warning but still run to "completion". This may not be ideal because for example in CI checks, it is not obvious that a (critical error) has occurred and because of that, the build step in CI may still be deemed successful (when it is not).

For example, when there is an error in the `site.json` file. e.g just remove a comma:
```
{
  "baseUrl": "/test-md",
  "titlePrefix": "",
```
to
```
{
  "baseUrl": "/test-md"
  "titlePrefix": "",
```

1. markbind serve output (which is ok since it actually reports an error)
![image](https://user-images.githubusercontent.com/41845017/160549312-8a62218d-f73e-4614-8907-7cd1a6538588.png)

2. markbind build output (no files being built but still says build success)
![image](https://user-images.githubusercontent.com/41845017/160549451-d68afbd5-025f-4268-8f6f-a3ba01b612be.png)

Additionally, if there is mismatch between markbind version, markbind build behaves similar to the above (which is not ideal), markbind serve also does not actually stop.

e.g. trying to serve the markbind site of cs3218 using the latest master branch markbind(there is a problem because the site uses .mbd files). The markbind serve is unable to serve any pages, but it runs to completion
![image](https://user-images.githubusercontent.com/41845017/160550003-3e1f2e57-0a25-4b9c-936e-0accebc9f5b2.png)
![image](https://user-images.githubusercontent.com/41845017/160550216-0393a8f4-6192-4835-8223-657f8dd1b58c.png)

**What did you expect to happen?**
I expect the command to fail so that it's failure can be picked up in CI.

**What actually happened? Please include the actual, raw output.**
See above, does not always fail and only show warnings.

**Additional context**
I tested build and serve of a docusaurus site and both will fail immediately if there is a problem with the `docusaurus.config.js`
![image](https://user-images.githubusercontent.com/41845017/160550691-87298c0e-b59d-4c39-b41c-2f80a86d1465.png)

This problem is related to the teammates devdocs migration. The issue is that given the fact that the build might fail, it should be included in the CI to ensure this is checked everytime a PR is merged.

[...]

PR #1912, opened 2022-04-19
Warn users if PlantUML diagrams are found but Graphviz is not installed

tlylt:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Related to #1903
- Add a check to verify that Graphviz is installed before processing PlantUML diagrams.
- This will be done once per build/refresh (in the case of markbind serve)
- A warning will be logged to the console
	- image
- I am leaving the decision and the work to fail hard (or not) to development in #1863, as that will involve changing existing test cases and how we decide what kinds of errors to totally stop the build. This PR simply logs the warning but does not error out.
- Include a way to disable the warning:
	- https://deploy-preview-1912--markbind-master.netlify.app/userguide/sitejsonfile#plantumlcheck
	- image

- Lastly, make the warning in our UG even more explicit:
	- https://deploy-preview-1912--markbind-master.netlify.app/userguide/components/imagesanddiagrams#diagrams
	- image

nil

**Testing instructions:**
- Checkout to this PR
- insert the following (or any other puml code) in a .md file:
```md
<puml width="300">
@startuml
alice -> bob ++ : hello
bob -> bob ++ : self call
bob -> bib ++  #005500 : hello
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
</puml>
```
- Run `markbind build` or `markbind serve`, if Graphviz is installed, no warnings should be found, else, a warning should be found (exactly once even if u have multiple PlantUML diagrams)

Add Graphviz check when PlantUML diagrams are used

Without Graphviz, PlantUML diagrams generated may
contain errors. Users may not be aware that diagrams are
generated incorrectly if they do not serve up the pages.

Let's add a check and warn users in the console if the
installation of Graphviz is missing.

This helps to keep users informed in the console when
the needed installation is missing. Users are also informed
of a way to disable the warning in `site.json`.

---

- [x] No unrelated changes


damithc:
This is a good feature to add @tlylt 

Perhaps the CI info is a better info for a minimized panel rather than a popover, as it contains substantial amount of info, and stuff the user might want to copy-paste? Alternatively, it can simply link to another page that has the full CI details.

As for the error message, should we mention that only some (or which) puml diagrams require Graphviz? I'm thinking of the situation in which the site has puml diagrams that don't actually need Graphviz and yet this warning keeps showing up every time. I suppose making the error message smart enough to figure out if Graphviz is actually needed is not feasible or worth the effort?

tlylt:
> Perhaps the CI info is a better info for a minimized panel rather than a popover, as it contains substantial amount of info, and stuff the user might want to copy-paste? Alternatively, it can simply link to another page that has the full CI details.

Thanks for the suggestion, I have updated the docs to use a panel and included more details. (see PR description). Is this ok?

> As for the error message, should we mention that only some (or which) puml diagrams require Graphviz? I'm thinking of the situation in which the site has puml diagrams that don't actually need Graphviz and yet this warning keeps showing up every time. I suppose making the error message smart enough to figure out if Graphviz is actually needed is not feasible or worth the effort?

I have updated the docs to mention what's stated in the PlantUML website (that sequence and activity diagram work without Graphviz). For the warning, as I am doing this check before processing any of the puml diagrams, this implementation would not be able to identify the diagram type. This warning is therefore more like our broken link reporting. Do you think adding something like "Ignore this warning if it was intended." will be better? 

Ideally, we should have the warning be smarter, which could be done with some adjustments to how we process the PlantUML diagrams. I have something in mind to try out but because of the other considerations involved (e.g. performance, adjustments needed to the test cases), might take some time before it's done. Hence I am also ok if we want to remove this warning log in this PR and keep the documentation updated for now.

damithc:
> Thanks for the suggestion, I have updated the docs to use a panel and included more details. (see PR description). Is this ok?

Yes, this looks better. A slight worry is that a sizeable chunk of the yml file is duplicated here (I expect it is also given somewhere else in our docs?). Can go out of sync if we update the CI documentation later. 
Also, not strictly a 'tip', right? More like an 'example' of how to configure a CI, especially because we don't explain how to do it in other CI environments?


> Ideally, we should have the warning be smarter, which could be done with some adjustments to how we process the PlantUML diagrams. I have something in mind to try out but because of the other considerations involved (e.g. performance, adjustments needed to the test cases), might take some time before it's done. Hence I am also ok if we want to remove this warning log in this PR and keep the documentation updated for now.

Yes, we certainly don't want to lose performance over this. We _can_ consider a flag e.g., `requireGraphviz` in site.json to switch off this warning, although not sure if that would be going too far.

tlylt:
> Yes, this looks better. A slight worry is that a sizeable chunk of the yml file is duplicated here (I expect it is also given somewhere else in our docs?). Can go out of sync if we update the CI documentation later. Also, not strictly a 'tip', right? More like an 'example' of how to configure a CI, especially because we don't explain how to do it in other CI environments?

Updated to mention only the required steps.
image

> Yes, we certainly don't want to lose performance over this. We _can_ consider a flag e.g., `requireGraphviz` in site.json to switch off this warning, although not sure if that would be going too far.

Hmm in that case I will leave this PR here for the time being.

damithc:
> Hmm in that case I will leave this PR here for the time being.

If the flag is easy to add, go ahead and add it. I don't think we want a case where users see a warning that technically should not be there (because their PUML diagrams don't actually need Graphviz) and have no way to get rid of it. We can remove the flag when (and if) we make the error message 'smarter'? What do you think?

tlylt:
> > Hmm in that case I will leave this PR here for the time being.
> 
> If the flag is easy to add, go ahead and add it. I don't think we want a case where users see a warning that technically should not be there (because their PUML diagrams don't actually need Graphviz) and have no way to get rid of it. We can remove the flag when (and if) we make the error message 'smarter'? What do you think?

Sure I will add the option to suppress the warning.

packages/core/src/Site/index.js
```diff
@@ -950,7 +951,8 @@ class Site {
         || !_.isEqual(oldSiteConfig.enableSearch, this.siteConfig.enableSearch)
         || !_.isEqual(oldSiteConfig.timeZone, this.siteConfig.timeZone)
         || !_.isEqual(oldSiteConfig.locale, this.siteConfig.locale)
-        || !_.isEqual(oldSiteConfig.intrasiteLinkValidation, this.siteConfig.intrasiteLinkValidation);
+      || !_.isEqual(oldSiteConfig.intrasiteLinkValidation, this.siteConfig.intrasiteLinkValidation)
+      || !_.isEqual(oldSiteConfig.plantumlCheck, this.siteConfig.plantumlCheck);
```
jonahtanjz, line 955: I think these 2 lines should be indented by 2 more spaces?

packages/core/src/plugins/default/markbind-plugin-plantuml.js
```diff
@@ -84,12 +86,23 @@ module.exports = {
...
     processedDiagrams.clear();
+    graphvizCheckCompleted = false;
   },
 
   processNode: (pluginContext, node, config) => {
     if (node.name !== 'puml') {
       return;
     }
+    if (config.plantumlCheck && !graphvizCheckCompleted) {
+      exec('dot -V', (error, stdout, stderr) => {
```
jonahtanjz, line 97:
```suggestion
      exec(`java -jar "${JAR_PATH}" -testdot`, (error, stdout, stderr) => {
```

Would this be better since it is the recommended testing method on the [plantuml site](https://plantuml.com/graphviz-dot#fa6a5a3224d7da66)? The original should work as well for most cases (except when Graphviz is installed in a custom path).

jonahtanjz (commented):
Thanks @tlylt

Just 2 comments, the rest looks good :)

PR #1920, opened 2022-05-03
Upgrade nodejs to v14

### LINKED ISSUE #1862: Upgrade node.js to v14
According to the [schedule](https://nodejs.org/en/about/releases/), v12 is coming to the end of life at the end of April (about one month later). I think we can put this issue up for tracking (& discussion if any) before we move to v14.

Upgrade node.js and affected documentation.

nil
**Additional context**
nil

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1862

Update CI scripts and documentation mention of the node.js version to 14

- **Will likely need a senior dev to update the netlify build script to use node.js 14 if not done already.**
- I have this error when using node v14.15, which does not seem to work with the required packages after typescript migration. I think it is not critical as I am still able to run all the code normally, but wonder if we should specify this condition in our DG
	- image
- Any other places that I may have missed

**Testing instructions:**
I have been using nodejs v14 for quite a while locally and doesn't seem to have any issues, but feel free to test it if possible.

Upgrade nodejs to v14

---

- [x] No unrelated changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '12'
+          node-version: '14'
       - name: Install Graphviz
         uses: tlylt/install-graphviz@v1
       - name: Install Java
@@ -45,7 +45,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '12'
+          node-version: '14'
       - name: Install Graphviz
         uses: tlylt/install-graphviz@v1
       - name: Install Java
--- a/docs/_markbind/variables.md
+++ b/docs/_markbind/variables.md
@@ -12,7 +12,9 @@
 <variable name="icon_ticked">:far-check-square:</variable>
 
 <variable name="node_version_number">14</variable>
+<variable name="node_dev_version_number">14.17.0</variable>
 <variable name="node_version"><tooltip content="MarkBind aims to support up to the last maintenance lts release as outlined [here](https://nodejs.org/en/about/releases/)">v{{ node_version_number }}</tooltip></variable>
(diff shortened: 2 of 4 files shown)


jonahtanjz (commented):
Thanks @tlylt 

> I have this error when using node v14.15, which does not seem to work with the required packages after typescript migration. I think it is not critical as I am still able to run all the code normally, but wonder if we should specify this condition in our DG

Seems fine to me as well. Maybe can get @ryoarmanda to double check on this?

> Will likely need a senior dev to update the netlify build script to use node.js 14 if not done already.

Will need @ang-zeyu's assistance on this :)

ryoarmanda:
> I have this error when using node v14.15, which does not seem to work with the required packages after typescript migration.

Looks like the error might be resolved if your local node version is upgraded to a higher v14 version. As noted in the image, the packages support Node v14 among other versions (they also keep support for Node v12 too), but with a lower bound of `14.17.0`. Your local node version is `14.15.1`, which is below that. You can try upgrading it to a higher v14 version if need be, or jump to v16 LTS if you'd like.

---

Notwithstanding the above, I think the move to v14 should be done as it has reached end of life on 30 April.

tlylt:
> Your local node version is 14.15.1, which is below that. You can try upgrading it to a higher v14 version if need be, or jump to v16 LTS if you'd like.

I see. Do you think it will be good to mention in our DG that we require `14.17.0` and above? Currently, it will only say `v14` and above, which is slightly misleading. (I assume that this is fine for end-users as long as they have v14 so need no specify for UG)

ryoarmanda:
> I see. Do you think it will be good to mention in our DG that we require 14.17.0 and above? Currently, it will only say v14 and above, which is slightly misleading. (I assume that this is fine for end-users as long as they have v14 so need no specify for UG)

Yeah, let's do that :+1:

ang-zeyu (approved):

PR #1925, opened 2022-05-06
Exclude node_modules from copying and page rendering in default site.json

### LINKED ISSUE #1844: Exclude node_modules files by default when setting up via markbind init
Windows 10
* **MarkBind Version:**
v3.1.1
I think it is pretty complicated for someone new with MarkBind to do a simple task such as ignoring files from `node_modules`. This is a common scenario when starting a js project and adding MarkBind to it. I took some time myself in order to figure out how to do it and initially thought that it was bug 😢

A typical project initialization process might look like this:

1. Start a new js project `npm init`
2. Install the latest version of markbind-cli `npm i markbind-cli`
3. Add in the `package.json`'s script that `"build": "markbind build"`, `"init": "markbind init"`
4. initialize a markbind project `npm run init`
5. try building the project by `npm run build`
	- will copy and build all the files in node_modules
	- not what is expected
	- read the docs to discover the `ignore` attribute
6. try adding "node_modules/*" in `site.json`'s `ignore` attribute
7. try building the project by `npm run build`
   - doesn't quite work, still tries to build 500 plus pages, and ended with an error
   - possibly due to the glob pattern of `"**/*.+(md|mbd)"`
   - read the docs again to figure out that there's a globExcludes that works **with glob** (I initially thought that it is defined in a separate object and that doesn't work) 
8. try adding "node_modules/*" in a globExcludes:
Note that this doesn't work, but it's the first thing I tried:
```
    {
      "glob": ["**/index.md", "**/*.+(md|mbd)"],
    },
	{
      "globExclude": ["node_modules/*"]
	}
```
This works
```
    {
      "glob": ["**/index.md", "**/*.+(md|mbd)"],
      "globExclude": ["node_modules/*"]
    }
```

10. possibly also realize that this will work:
```
"pagesExclude": ["node_modules/*"],
```

I expect the common task to be simple.

Currently, it requires a new user to understand a few attributes such as 'ignore', 'glob', 'globExcludes', 'pagesExclude' just to get started.
This could be due to the fact that we only mention how to globally install markbind-cli in our user guide. While that works, it may not work well for existing projects or those that already have the setup of a package.json with node_modules. My proposal is as follows:
- include a guide on how to set up a project via normal installation e.g. `npm i markbind-cli@3.1.1`. Related: #1772
- in `site.json`, ignore`node_modules/*` and also set `"pagesExclude": ["node_modules/*"]`
- In tandem, work on #1782

tlylt:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1844
- exclude node_modules from page rendering as the default pages include a glob of
```
	{
      "glob": ["**/index.md", "**/*.md"]
    }
```
- the above does not play nice if MarkBind is installed locally, or if there are any other node_modules (because most of them include README.md), which will have to be excluded explicitly.

nil

**Testing instructions:**

1. Check out this PR
2. Run `markbind init`
3. Verify the `site.json`
4. Install a node_module by doing something like `npm init -y` and then `npm install happy-birthday`
5. Run `markbind serve` or `markbind build`
6. Verify that the node_modules are not copied over to _site and things are rendered properly

Exclude node_modules from page rendering in default site.json

If there are node modules installed along with MarkBind, the
README.md files within these folders will be processed due to
our default glob pattern in `site.json`, and they will be copied
over to the production site.

The users will typically want to avoid the above situation, but
they may not be familiar with MarkBind configurations early on
and can be troubled by this problem.

Let's exclude node_modules from copying over to the
production site and add them into `pagesExclude` to prevent
any processing, by default.

This change will not affect users negatively but provides a
sensible default to deal with node_modules.

---

- [x] No unrelated changes


jonahtanjz (approved):
Thanks @tlylt. This makes it so much easier for MarkBind users to work with projects that require `node_modules` 💯 

LGTM 👍

PR #1939, opened 2022-05-25
Fix heading padding FOUC (option 3 with modifications: sticky header)

### LINKED ISSUE #1560: Improve the fixed header approach to remove "header padding fouc"

ang-zeyu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] Others, please explain:

Resolves #1560 

Follow up on #1929, #1930, #1931 and discussion in #1560

**Overview of changes:**
- Adds this feature: https://deploy-preview-1939--markbind-master.netlify.app/userguide/tweakingthepagestructure#option-2-sticky-header-dynamically-hidden
- Changes this https://deploy-preview-1939--markbind-master.netlify.app/userguide/tweakingthepagestructure#option-1-fixed-header to be dynamically hidden (see yellow warning box)

- Is the feature set o-k? (too complex? lacking in configuration?)
- This is somewhat breaking change as it requires making some small adjustments in the default user layout files (move initial padding-top to margin). ❗  Should be noted in the release changelog as it requires user modification.

**Testing instructions:**
- `<header fixed>` (anywhere in the userguide except `userGuide/stickyDemo.html`) should behave as before, except without the FOUC. This includes the mobile site / page nav.
   - https://deploy-preview-1939--markbind-master.netlify.app/userguide/tweakingthepagestructure
- `<header sticky>` (new feature) performs as described in the documentation. The mobile site / page nav should perform identically to that of `<header fixed>` (~**currently broken** - tbd~ done!)
   - https://deploy-preview-1939--markbind-master.netlify.app/userguide/stickydemo

Implement sticky header

Since no assumptions of the user's header's height can be made at build
time, runtime style detection is required to calculate the header
height used in the fixed-header-padding class. This causes a flash of
unstyled content briefly before this detection is done
(the content is hidden behind the header).

Let's implement a sticky header instead, which "natively" offsets
content below it.

It should be noted that header height detection is still required
however, particularly for offsetting the sticky positioning of the
navigation menus on the side (top: var(--sticky-header-height)), since
they are (usually) much shorter than the main content element.

This header height is now exposed using the css variable
var(--sticky-header-height) for use in the user layout files.

---

- [ ] Added tests for bug fixes or features (we need e2e tests)
- [x] No unrelated changes


packages/core-web/src/index.js
```diff
@@ -21,21 +21,25 @@ function scrollToUrlAnchorHeading() {
...
+  const isSticky = stickySelector.length !== 0;
+  const headerSelector = isSticky ? stickySelector : fixedSelector;
+  if (!(isSticky || isFixed)) {
     return;
   }
 
+  let isHidden = false;
   window.addEventListener('resize', () => {
-    if (window.innerWidth > 767 && headerSelector.hasClass('hide-header')) {
+    if (window.innerWidth > 767 && isHidden) {
```
jonahtanjz, line 42: Just a heads up, I think will have to update the `fixed-header-padding` height one more time during resize.

Currently there will be a large padding if the header reappears during resize:

https://user-images.githubusercontent.com/47470981/170835030-f8f08933-7558-476a-9d4d-84344f77504e.mp4

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -148,20 +147,46 @@ If you wish insert scripts at the bottom, before MarkBind's scripts, simply inse
...
-****To fix the `<header>`****
+#### Option 1: Fixed header
+
+****Steps****
 1. Add the `fixed` attribute to your `<header>` element in the layout per the above example.
 
-2. Then, to add the necessary top padding for the main content, add the `fixed-header-padding` class to **elements that should be shifted down** in accordance with the fixed header.
+2. Then, to add the necessary top padding for the main content, add the `fixed-header-padding` class to **elements that should be shifted down** in accordance with the height of the fixed header.
+   
+   This class should generally be added to the elements that immediately below the header. You may also refer to the default template generated by `markbind init`, or the above code example (which is used for this documentation).
```
tlylt, line 161:
```suggestion
   This class should generally be added to the elements that are immediately below the header. You may also refer to the default template generated by `markbind init`, or the above code example (which is used for this documentation).
```

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -148,20 +147,46 @@ If you wish insert scripts at the bottom, before MarkBind's scripts, simply inse
...
+
+<box type="warning" seamless>
+
+As this feature does not make any assumptions of your header height, runtime height detection is performed to populate the `fixed-header-padding` class. This results in improperly formatted content on page load for a brief period.
+
+To avoid this, the implementation here switches the `<header>` element to use `fixed` positioning, and populates the `fixed-header-padding` class's padding *only once* the user has scrolled past the top of the header. Thus, ensure your `<header>` element is the first visible element.
+</box>
```
tlylt, line 168: May I check what you intend to explain here?

> As this feature does not make any assumptions of your header height, runtime height detection is performed to populate the `fixed-header-padding` class. This results in improperly formatted content on page load for a brief period.

This says that using the `fixed-header-padding` class will result in improperly formatted content... but
> To avoid this, the implementation here switches the `<header>` element to use `fixed` positioning, and populates the `fixed-header-padding` class's padding *only once* the user has scrolled past the top of the header. Thus, ensure your `<header>` element is the first visible element.

does this say that it actually won't? So is there anything that the user needs to do if he/she chooses to go with option 1 with `fixed` + `fixed-header-padding`?

Perhaps can rephrase this section to make it clearer:)

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -148,20 +147,46 @@ If you wish insert scripts at the bottom, before MarkBind's scripts, simply inse
...
 
 <box type="tip" seamless>
 
-If you are not sure where to put the `fixed-header-padding` attribute, you may also refer to the default template for `markbind init`, which already has this setup.
+To preserve screen real estate, the header is hidden on devices with a width of `<  767px` when the user scrolls down, and automatically re-shown when the page is scrolled up.
```
tlylt, line 172:
```suggestion
To preserve screen real estate, the header is hidden on devices with a width smaller than 767px when the user scrolls down and automatically re-shown when the page is scrolled up.
```

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -148,20 +147,46 @@ If you wish insert scripts at the bottom, before MarkBind's scripts, simply inse
 
 ---
 
-### Fixing the header to the top
+### Sticking the header to the top
 
-Headers are commonly included inside the HTML `<header>` tag. In encouraging this, a convenient interface to implement <tooltip content="Headers that stick to the top of the page while scrolling the content">fixed headers</tooltip> surrounding the `<header>` tag is provided that ensures page anchors work correctly.
+Headers are commonly included inside the html `<header>` tag. In encouraging this, 2 convenient options to implement fixed/sticky headers using the `<header>` tag are provided, which primarily ensures that the page **scrolls to anchors** correctly.
```
tlylt, line 152: When reading this, I am confused as to why we explain that the convenient options "encourage" the use of headers within the `<header>` tag 😓, since they are provided **primarily** to ensure 
> that the page **scrolls to anchors** correctly.

ang-zeyu:
@jonahtanjz @tlylt @damithc Thanks for the reviews!

After letting this sit for a while I realised there was a very simple way I missed previously to achieve the best result:
- in terms of reader functionality, this results in the benefits of the previously discussed option 2.
- resulting implementation is also no where as complex as option 2

The gist:
- I've removed option 1 (dynamically fixed header). This works, but again the drawback is that non-sticky announcements cannot be placed.
- I've adapted option 3. The easy fix here is to offset the sticky positioning of the site / page nav in the layout files by the sticky header height. This has all the benefits of option 2 (no fouc, allows things on top of the header, ...)
   - I've also exposed the `--sticky-header-height` css variable in the user guide for advanced usage. This can be seen as the same slightly awkward `fixed-header-padding` class, so its still slightly complex in terms of author usability however. 
- In terms of documentation best reviewed from scratch. 😓 There are a lot of test file updates (from `fixed-header-padding` removal mainly) so github's file explorer should really help

Concerns:
- Mainly whether the user guide is clear enough due to exposing the css variable.
https://deploy-preview-1939--markbind-master.netlify.app/userguide/tweakingthepagestructure#sticking-the-header-to-the-top

packages/core-web/src/index.js
```diff
@@ -21,50 +21,61 @@ function scrollToUrlAnchorHeading() {
...
 
-  const headerSelector = jQuery('header[fixed]');
-  const isFixed = headerSelector.length !== 0;
-  if (!isFixed) {
+  const headerSelector = jQuery('header[sticky]');
+  if (headerSelector.length === 0) {
     return;
   }
 
-  window.addEventListener('resize', () => {
```
ang-zeyu, line 37: > Just a heads up, I think will have to update the fixed-header-padding height one more time during resize.

I've the "merged" this fix from the other PR into the resizeObserver changes below

packages/core-web/src/index.js
```diff
@@ -21,50 +21,61 @@ function scrollToUrlAnchorHeading() {
...
-    const headerMaxHeight = headerSelector.css('max-height');
+  let isHidden = false;
+  function showHeader() {
+    isHidden = false;
+    headerSelector.removeClass('hide-header');
+  }
+  headerSelector[0].addEventListener('transitionend', () => {
     // reset overflow when header shows again to allow content
     // in the header such as search dropdown etc. to overflow
-    if (headerMaxHeight === '100%') {
```
ang-zeyu, line 53: most of the changes here (resizeObserver, this line, isHidden) are necessary due to switching to `transform translate` for hiding the header (height change dosen't work with `position: sticky`), which should be simpler (I think) to deal with too.

packages/core-web/src/styles/markbind.css
```diff
@@ -127,30 +127,21 @@ code.hljs.inline {
...
 /* Header */
 
-header[fixed] {
-    max-height: 100%;
-    position: fixed;
-    transition: max-height 0.6s ease-in;
+header[sticky] {
+    position: sticky;
+    top: 0;
+    transition: transform 0.3s ease-in-out;
```
ang-zeyu, line 133: height transition won't work now since its sticky

ang-zeyu (commented):
some self-commenting on the changes to hopefully make reviewing easier :)

damithc:
>Concerns:
>
>Mainly whether the user guide is clear enough due to exposing the css variable.
https://deploy-preview-1939--markbind-master.netlify.app/userguide/tweakingthepagestructure#sticking-the-header-to-the-top

Perhaps give one more example of how the variable can be used? Having more than one example helps to identify which part of the code the user needs to change when applying to another context.

ang-zeyu:
> Perhaps give one more example of how the variable can be used? Having more than one example helps to identify which part of the code the user needs to change when applying to another context.

Added, though - I avoided this previously as it requires referencing some of our internal implementation details: `"auto-generated, invisible page anchors"`. 🤔 

We only have one more example use case in place otherwise which is even more obscure (setting up the header hiding transition for mobile devices).

I could also remove the tip and second example entirely and just focus on the first one to avoid confusion. Wdyt? @damithc

damithc:
> I could also remove the tip and second example entirely and just focus on the first one to avoid confusion. Wdyt? @damithc

Is it possible to give examples of how _they_ can use it (rather than how we use it)? Yes, if it is unlikely users will want to use that variable, we can skip it altogether.

ang-zeyu:
> Is it possible to give examples of how _they_ can use it (rather than how we use it)? 

I can't think of any right now, at least. =X

Maybe @jonahtanjz @tlylt have some ideas. I can go ahead with skipping the tip + second example altogether otherwise.

> Yes, if it is unlikely users will want to use that variable, we can skip it altogether.

To be clear though, its still important for the primary use case (the first example), or at least, for the author to take note of it (e.g. accidental removal) when altering layout files. Otherwise, the site / page nav will be hidden behind the header as you scroll down. Similar to the slightly awkward `fixed-header-padding` class currently.

ang-zeyu:
Sorry, realised I haven't handled a few more implementation edge cases. Will re-ping once its ready.

jonahtanjz:
> Maybe @jonahtanjz @tlylt have some ideas. I can go ahead with skipping the tip + second example altogether otherwise.

I don't think there are any good examples other than if the user decides to not use a `site/page-nav` and have their own sticky sidebar? Agree, I think it's good for the users to be aware of the use of `--sticky-header-height` in general if they are using sticky header. The second example on page anchors can be a little confusing to the users so I think this can be removed.

jonahtanjz:
> Maybe @jonahtanjz @tlylt have some ideas. I can go ahead with skipping the tip + second example altogether otherwise.

I don't think there are any good examples other than if the user decides to not use a `site/page-nav` and have their own sticky sidebar? Agree, I think it's good for the users to be aware of the use of `--sticky-header-height` in general if they are using sticky header. The second example on page anchors can be a little confusing to the users so I think this can be removed.

jonahtanjz (commented):
Just noticed this, will also need to collapse the dropdown when the header is hidden on mobile. Seems like the dropdown will auto hide on `blur` so maybe we can trigger a `blur` event inside the `hideHeader` method? 

https://user-images.githubusercontent.com/47470981/173217141-d295318b-a782-4ca2-9300-01a77659548c.mp4

docs/userGuide/tweakingThePageStructure.md
```diff
@@ -148,20 +147,43 @@ If you wish insert scripts at the bottom, before MarkBind's scripts, simply inse
...
-### Fixing the header to the top
+### Sticking the header to the top
 
-Headers are commonly included inside the HTML `<header>` tag. In encouraging this, a convenient interface to implement <tooltip content="Headers that stick to the top of the page while scrolling the content">fixed headers</tooltip> surrounding the `<header>` tag is provided that ensures page anchors work correctly.
+A sticky header can be implemented by simply adding the `sticky` attribute to a `<header>` element.
 
-****To fix the `<header>`****
-1. Add the `fixed` attribute to your `<header>` element in the layout per the above example.
+Using this attribute as opposed to setting `position: sticky` manually in your stylesheets comes with several conveniences:
+- MarkBind will take care of **aligning the viewport with page anchors** when scrolled to.
```
tlylt, line 155: Is this point about not letting elements be covered by the header when clicking on anchor links? Perhaps can reword it to state just that? I feel that a bit confusing to say it in a current way, what do you think?

tlylt (commented):
Not really related to this PR, but wondering if you guys also experience the weird behavior of having the header reappearing even if I am scrolling downwards? It happens sporadically when browsing the site on a mobile (actual phone, not on desktop inspector). I tried on iPhone Safari & Chrome and both have this effect. It contradicts the point:
> To preserve screen real estate, the header is hidden on devices with a width of less than 767px when the user scrolls down, and automatically re-shown when the page is scrolled up.

...but I think it has been like this all along.

> The second example on page anchors can be a little confusing to the users so I think this can be removed.

Agree that in the current state, the example is a little confusing (because it is how we use it and not how they can use it). I think either removing it or putting it in a closed panel is fine. Can't really think of any other common use cases either.

docs/css/main.css
```diff
@@ -56,28 +56,28 @@ mark {
     margin: 0 auto;
     min-width: 0;
     max-width: 1000px;
+    overflow-x: auto;
```
ang-zeyu, line 59: Documenting the layout file changes for review + future release changelog noting (should be simplified though I think =P)

1st of 3 important changes in the layout file:

This might be somewhat controversial, but is unfortunately necessary with a sticky header. At least, I haven't found another solution.


The issue (without this) occurs on pages that overflow (e.g. "formatting contents" page or "Using html javascript and css"). The [nearest scrolling ancestor](https://developer.mozilla.org/en-US/docs/Web/CSS/position) which is the page / root element expands to accomodate the overflown content, which in turn does 2 things:
- `window.innerWidth/Height` are erroneous (although there is an easy alternative for this `document.documentElement.clientWidth/Height`)
- More important is that the way sticky positioning is defined, the sticky positioning goes haywire (with or without our header height detection)

Gif demo (open image in new tab to view in order): there are 3 settings here.

1. The first shows proper function of sticky positioning when the viewport zooms out to the entire page.
2. The second showcases how the page loads on a mobile device normally, observe how the sticky positioning completely goes haywire as the page is "zoomed" (although it dosen't quite appear like that to the reader).
3. The third is same as the second, except I temporarily toggled off the header hiding (the `translateY`) to show the issue is really unrelated to header hiding / anything we are doing.

stickyissues

You can try removing the `overflow-x` when testing this PR to see the issue better, its a little hard to present on a gif.

The fix: I've set `overflow-x` on the content wrapper which significantly changes ux on mobile devices (but just pages that overflow).

Pros:
- As a nice side effect, this change "fixes" https://github.com/MarkBind/markbind/issues/1741.
- Overflown content is still usable by swiping left / right.

Cons:
- On desktop devices, although _very very_ unlikely (since there is enough space for overflown content on mobile to fully "expand" - including our embeds), if there is overflow, I would still like to point out the horizontal scroll bar is only visible at the bottom of the content wrapper. (which can be confusing)
   - add to that that site development should usually be done on a desktop device (so authors will likely limit content sizes to a reasonable degree (e.g. images))
- You'll no longer be able to "zoom out" on mobile (though arguably this could also be a plus for mobile ux)

Not sure if there's a better solution =\, but imo its a good enough compromise.

docs/css/main.css
```diff
@@ -56,28 +56,28 @@ mark {
...
     -webkit-transition: 0.4s;
 }
 
 #site-nav,
 #page-nav {
+    display: flex;
+    flex-direction: column;
     position: sticky;
-    top: 0;
+    top: var(--sticky-header-height);
```
ang-zeyu, line 70: 2nd change per the docs

docs/css/main.css
```diff
@@ -56,28 +56,28 @@ mark {
...
     flex: 0 0 auto;
-    padding-top: 1rem;
     max-width: 300px;
+    max-height: calc(100vh - var(--sticky-header-height));
```
ang-zeyu, line 73: last change: `100vh` worked before as it took into account `padding-top` from `fixed-header-padding`.
With the shift to `top: var(--sticky-header-height);` the content boxes actually do shift down, so we need to minus that off.

The other changes (`90vh`) is just to standardise the way this is done in the layout stylesheet between the site and page nav to make things simpler.

packages/core-web/src/index.js
```diff
@@ -75,31 +88,36 @@ function detectAndApplyFixedHeaderStyles() {
     if (isEndOfPage) { return; }
 
     if (currentOffset > lastOffset) {
-      const headerEnd = headerSelector.height() + headerSelector[0].getBoundingClientRect().top;
-      const isBeforeHeader = currentOffset < headerEnd;
+      /*
+       1) Bounding box is calculated as if 'position: fixed' when sticky is "activated".
+       Revert the position to 'static' temporarily to avoid this.
+
+       Seems to be harmless UX wise, even on extremely slow devices.
```
ang-zeyu, line 95: The change here is pretty much as explained in the comment, although I would like to get some help with testing this ^ in particular on more devices. (whether the header ever "flickers" right before it is being hidden)

ang-zeyu (commented):
@jonahtanjz @tlylt  I've removed the second example since everyone is in agreement, and touched up the relevant parts from your reviews. Also edited the example a little.

The implementation should be ready now too :)

> Just noticed this, will also need to collapse the dropdown when the header is hidden on mobile. Seems like the dropdown will auto hide on blur so maybe we can trigger a blur event inside the hideHeader method?

ouch, nice catch =X. Miscopied one of the `headerSelector.css('overflow', 'hidden');` -> `headerSelector.css('overflow', '')`.


Some more self comments to help reviewing:

ang-zeyu:
> Not really related to this PR, but wondering if you guys also experience the weird behavior of having the header reappearing even if I am scrolling downwards? It happens sporadically when browsing the site on a mobile (actual phone, not on desktop inspector). I tried on iPhone Safari & Chrome and both have this effect. It contradicts the point:


I've encountered this as well, have a hunch as to why this occuring (still is too, I think). As the debounce duration is pretty short, and touchscreens are quite sensitive (as opposed to chrome emulator operated by a mouse), there's a chance this is functioning  correctly, but the change in page offset isn't "thresholded" enough. (currently we just do `current > lastOffset`)
Will investigate in a separate PR along with a possible ux improvement to the header hiding overall.

packages/core-web/src/index.js
jonahtanjz: Tested and works fine on my end 👍

jonahtanjz (approved):
LGTM on my end 👍 

Will merge once the temporary announcement is removed :)

jonahtanjz:
@ang-zeyu I think will need you to update the commit message as well since the implementation has changed :)

ang-zeyu:
> @ang-zeyu I think will need you to update the commit message as well since the implementation has changed :)

Thanks for the reminder, updated

PR #1949, opened 2022-06-15
Update copy/wrap button to be visible on mouse hover

### LINKED ISSUE #1842: Code copy button: overlay on hover rather than show outside of code block
Current:
image

Problem:
Robs us of a some horizontal space that we can use for code otherwise.

Suggestion:
Show overlaid on top-right corner on hover, similar to how GitHub does it.
image

IIRC, we decided to show it outside to prevent it covering code, but if we show the button only on hover, that problem can be avoided.

Applies to code wrapping button as well.
image

elroygohjy:

- [ ] Bug fix
- [ ] Documentation update
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1842
I updated the css of the buttons to be only visible on hover. 
Before (copied from issue 1842):
image
After:
(Only wrapped button plugin enabled)
image
(Both plugins enabled)
image

**Testing instructions:**
1. Include `codeBlockCopyButtons`/`codeBlockCopyButtons` under plugins in site.json
2. Hover any code block.
 
**Proposed commit message:** Update wrap and copy button to be visible on hover


packages/core-web/src/styles/markbind.css
```diff
@@ -247,22 +247,34 @@ pre {
...
 }
 
-.function-btn-container {
-    display: grid;
-    height: max-content;
-}
-
 button.function-btn {
-    border: none;
     background-color: transparent;
```
jonahtanjz, line 251: Would be nice to give it a background color instead of `transparent` to avoid the below issue

image

packages/core-web/src/styles/markbind.css
```diff
@@ -247,22 +247,34 @@ pre {
...
+}
+
+pre:hover > .function-btn-container > button.function-btn {
+    display: block;
+}
+
+.function-btn-container {
+    display: grid;
+    height: max-content;
+}
```
jonahtanjz, line 276: Just want to check, can these be removed? I tried removing them and it doesn't seem to affect the buttons.

jonahtanjz (commented):
Thanks @elroygohjy! Just a couple of suggestions.

For single line code block, the buttons are not "centered" vertically.

image

Perhaps you can reduce the top margin of the buttons or slightly increase the paddings of the code block?

packages/core-web/src/styles/markbind.css
elroygohjy: Agreed! I will remove it

Issue #1855, opened 2022-03-26
Update regular expressions to use literals


ryoarmanda:
Follow up to #1852.

With the bumped version of `eslint-config-airbnb-base`, it has enabled `prefer-regex-literals` rule which I turned off at the moment. However, there is little reason on why we do not prefer regex literals so let's update our regex declarations in our code to use literals and remove the disable on it.

PR #1956, opened 2022-07-03
Update regex to enable prefer-regex-literals rule

EltonGohJH:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves https://github.com/MarkBind/markbind/issues/1855

~~Other than the prefer-regex-literal, I also updated default-param-last which is also included in the air bnb default config~~.

**Testing instructions:**

Update regex to enable prefer-regex-literals

---

- [x] No unrelated changes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,7 @@ module.exports = {
   "rules": {
     "array-bracket-newline": ["error", { "multiline": true }],
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
-    "default-param-last": "off",
+    "default-param-last": "error",
     "func-names": "off",
     "no-else-return": ["error", { "allowElseIf": true }],
     "no-underscore-dangle": "off",
@@ -33,7 +33,7 @@ module.exports = {
     "max-len": ["error", { "code": 110 }],
     "no-param-reassign": ["error", { "props": false }],
     "operator-linebreak": ["error", "before"],
-    "prefer-regex-literals": "off",
+    "prefer-regex-literals": "error",
     // override airbnb-base dev dependencies, latest version does not white list __mocks__
     "import/no-extraneous-dependencies": [
       "error", {
@@ -65,7 +65,7 @@ module.exports = {
         "import/no-default-export": "error",
         "lodash/prefer-lodash-method": [0],
         "lodash/prefer-noop": [0],
-        "@typescript-eslint/default-param-last": "off",
+        "@typescript-eslint/default-param-last": "error",
         "@typescript-eslint/indent": [
           "error",
           2,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
   "rules": {
     "array-bracket-newline": ["error", { "multiline": true }],
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
+    "default-param-last": "off",
     "func-names": "off",
     "no-else-return": ["error", { "allowElseIf": true }],
     "no-underscore-dangle": "off",
@@ -63,6 +64,7 @@ module.exports = {
         "import/no-default-export": "error",
         "lodash/prefer-lodash-method": [0],
         "lodash/prefer-noop": [0],
+        "@typescript-eslint/default-param-last": "off",
         "@typescript-eslint/indent": [
           "error",
           2,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,6 @@ module.exports = {
   "rules": {
     "array-bracket-newline": ["error", { "multiline": true }],
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
-    "default-param-last": "error",
     "func-names": "off",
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
     "max-len": ["error", { "code": 110 }],
     "no-param-reassign": ["error", { "props": false }],
     "operator-linebreak": ["error", "before"],
-    "prefer-regex-literals": "off",
+    "prefer-regex-literals": "error",
```
tlylt, line 36: Since this is fixed by your PR, we can safely remove this line entirely and allow it to be included by default from the airbnb config

.eslintrc.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   "rules": {
     "array-bracket-newline": ["error", { "multiline": true }],
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
-    "default-param-last": "off",
+    "default-param-last": "error",
```
tlylt, line 15: Hmm for this change, it is not mentioned in the original issue. As such, could you also describe the change in the "Overview of changes" section of the PR description? Similarly, since this is part of airbnb config, we can safely remove this line

packages/core/src/Site/index.js
```diff
@@ -107,8 +107,9 @@ const ABOUT_MARKDOWN_DEFAULT = '# About\n'
 const MARKBIND_LINK_HTML = `<a href='${MARKBIND_WEBSITE_URL}'>MarkBind ${MARKBIND_VERSION}</a>`;
 
 class Site {
-  constructor(rootPath, outputPath, onePagePath, forceReload = false,
-              siteConfigPath = SITE_CONFIG_NAME, dev, backgroundBuildMode, postBackgroundBuildFunc) {
+  constructor(rootPath, outputPath, onePagePath, dev,
```
tlylt, line 110: With this change, I would think that the calls to this constructor, via `new Site(...` should also be updated, which I think there are a few usages in `packages\cli\index.js`. May I check if you are aware and if there's any reason that those are not updated?

I guess our existing test cases are not comprehensive enough that the CI is still passing without the above changes 😓

tlylt (changes requested):

packages/core/src/Site/index.js
EltonGohJH: I used my IDE refactor method (to refactor all usages and I guess it was not working) and since it passed all test cases I tot it was working fine.
Sorry. I will look into this.

tlylt (commented):
Hi @EltonGohJH, thank you for making the changes!

I think everything looks good, but just one point on the `default-param-last` that I want to check with @MarkBind/active-devs on whether we want this rule enabled. This is because it wasn't in the scope of the mentioned issue and while it is part of the Airbnb standard, not surer if we do want to enforce this. (some discussions about the utility of this rule here https://github.com/standard/standard/issues/1414). Looking at the change you made after adjusting the parameters in `cli/index.js`, it seems like we now pass in more `undefine` in order to invoke the `new Site(...)` method.
@ryoarmanda and the rest, any thoughts?

jonahtanjz:
@tlylt @EltonGohJH 

> I think everything looks good, but just one point on the `default-param-last` that I want to check with @MarkBind/active-devs  on whether we want this rule enabled.

> it seems like we now pass in more `undefine` in order to invoke the new Site(...) method.

For readability, I am leaning towards leaving this rule disabled. From the discussion https://github.com/standard/standard/issues/1414 here, it seems like the goal of this rule is to omit optional tail arguments but in our use case, it is creating tail arguments for the `Site` method 😅.

tlylt (approved):
LGTM! @EltonGohJH sorry for the revision required:) I think we could have avoided that, perhaps by discussing the details beforehand...something that we ought to do prior to raising a PR.
(I have also adjusted the PR description slightly to clean up the outdated info)

packages/core/src/lib/markdown-it/index.js
```diff
@@ -14,7 +14,7 @@ const logger = require('../../utils/logger');
 
 const { HighlightRule } = require('./highlight/HighlightRule');
 
-const HIGHLIGHT_LINES_DELIMITER_REGEX = new RegExp(',(?![^\\[\\]]*])');
+const HIGHLIGHT_LINES_DELIMITER_REGEX = /,(?![^[\]]*])/;
```
tlylt, line 17:
```suggestion
const HIGHLIGHT_LINES_DELIMITER_REGEX = /,(?![^\[\]]*])/;
```
@EltonGohJH this place might be missing a slash
image
image

EltonGohJH: @tlylt 
I get an eslint error when adding the extra escape character.
image

tlylt: Hi @EltonGohJH , looks like the original regex
`new RegExp(',(?![^\\[\\]]*])');` could have been
`new RegExp(',(?![^[\\]]*])');`

In that case, let's go with your updated one. Thanks for checking!

tlylt (approved):

PR #1957, opened 2022-07-04
Set search bar width to placeholder text

weiquu:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [x] Bug fix
- [ ] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

Resolves #665 

**Overview of changes:**
Binds the width of the search bar to the length of the placeholder text so as to avoid clipping of the placeholder text.

Before:
search bar before

After:
search bar after

**Anything you'd like to highlight/discuss:**
The width is set using the `ch` unit, which is defined according to the width of the `o` character in a given typeface. Hence, it might run into issues with certain non-standard typefaces, should the MarkBind typeface be changed in the future.

**Testing instructions:**
Use a long placeholder text to check if the width of the search bar expands accordingly.

**Proposed commit message: (wrap lines at 72 characters)**
Set search bar width to placeholder text

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [ ] Linked all related issues
- [ ] No unrelated changes

PR #1958, opened 2022-07-04
Add new question type for questions with multiple blanks

weiquu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Add a new question type, the `multiple blanks` question type. The syntax is very similar to other question types. An example is provided below:
```
<question type="multiBlanks" header="This question _______ is a '_______ blanks' question." threshold = 0.5 showIntermediateResult>
  <q-option keywords="type, class">
  </q-option>
  <q-option keywords="multiple, inline" reason="That's the name of the question type.">
  </q-option>
</question>
```
Here is what someone might see after answering a question correctly:
multi blanks answered

The threshold works the same way as the `text` question type (checks the percentage of correct blanks). The keywords are checked for exact matches. The `showIntermediateResult` option allows for users to see the results for individual blanks even if the threshold has yet to be met .

My original intention was for the inputs to be placed in line with the question, but based on my knowledge, this requires a fairly significant change in how the current question files are written/structured.

**Testing instructions:**

Add new question type for questions with multiple blanks

---

- [ ] No unrelated changes


tlylt:
@weiquu Same as #1957, may I check if all the file changes included in this PR are intended?

packages/vue-components/src/questions/Question.vue
```diff
@@ -267,6 +284,35 @@ export default {
         this.markAsWrong(markAsAnsweredIfWrong);
       }
     },
+    checkMultiBlanksAnswer(markAsAnsweredIfWrong) {
+      let numMatching = 0;
+      for (let i = 0; i < this.answers.length; i += 1) {
+        let ansIsCorrect = false;
+        const ansProps = this.answers[i]._props;
```
ang-zeyu, line 291: And just one minor nit in the implementation that jumped out:

Is the use of this private property documented in Vue? I'm not sure if it it would be wise to use an internal representation.
Perhaps we could use the documented `$props` https://v2.vuejs.org/v2/api/#vm-props instead.

ang-zeyu (commented):
@weiquu thanks for contributing, and good work here.

Was this feature addition discussed in private with @damithc? I think this is really nice to have if not, but in either case good practice put up an issue for feature additions / enhancements in case anyone has any objections / suggestions, to avoid wasted work. 😀 

I'll try to get to this in-depth on the weekend (although I think it mostly looks ok at a glance), a few high level points you can work on in the meantime otherwise:
- needs documentation
- \+ test updates
- ux wise, should we use a smaller `<textarea>`, or better regular `<input>` here? since we don't the blanks to get very large.

   Feel free to tweak other things too and see what's best, like removing the horizontal divider to reduce the footprint. You don't have to follow the free text question design strictly. 🙂 

   Roping in @damithc for any other suggestions.

damithc:
>Was this feature addition discussed in private with @damithc? I think this is really nice to have if not, but in either case good practice put up an issue for feature additions / enhancements in case anyone has any objections / suggestions, to avoid wasted work. 😀

IIRC, this is the experimental feature @weiquu did to get into the IWM module. I'm OK with adding this feature.

@weiquu Update the user guide, if you haven't done so already. When giving examples, use a clean and realistic example. Don't go 'meta' as you have done in the current example as it makes the example harder to understand  :-)

weiquu:
Hi, I've made a few changes to the UI as well as changed the code a bit (shifted checking of answers from the main question to the options themselves since each has its own keywords). Please do let me know if you have any comments (:

In addition, I've ran into an issue while writing tests where I'm unable to properly set the value of the QOption input. Using `await wrapper.findAllComponents(QOption).at(2).find('input').setValue('key');` seems to set the data of the input (it is marked correctly) but not the value in the input box (in the snapshot, the input box is empty). I've also tried `await wrapper.findAllComponents(QOption).at(2).setData({ inputText: 'key' });` as well as doing `await wrapper.findAllComponents(QOption).at(2).find('input').trigger('input');` after both lines above. I'll continue looking for a solution, but would greatly appreciate any suggestions or links to resources if possible!

ang-zeyu:
> in the snapshot, the input box is empty

hmm do you mean (snapshot):

```html
<input
  class="form-control"
  value="key"
/>
```

? That's expected, the `value` [attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#value) only specifies the initial value. 

> it is marked correctly

If snapshots (how they "marked correctly" in there) aren't sufficient to test the behaviours you're seeking to verify, one alternative is to assert the value of the `inputText` property manually, since it is directly tied to the input with `v-model`.

packages/vue-components/src/questions/QOption.vue
```diff
@@ -214,4 +314,26 @@ export default {
...
+        margin-bottom: 0;
+        width: 50%;
+    }
+
+    input.form-control:disabled,
+    .multiBlanks-keywords {
+        margin-bottom: 0.5rem;
+    }
+
+    .multiBlanks-div {
```
ang-zeyu, line 331: We're following [this](https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters) ..which dosen't say how to handle compound words =(.. but [this](https://developers.google.com/style/word-list#compounds) does.

i.e. would be `multiblanks-div`.

Or perhaps you could get rid of `multi` in the code references too, per the earlier suggestion.

A clearer suffix might be `multiblanks-form-control` / `blanks-form-control` as well so the reader immediately knows this overrides `form-control` for the case of blanks questions.

packages/vue-components/src/questions/QOption.vue
```diff
@@ -96,21 +96,94 @@
...
+
+  <!-- multiple blanks option -->
+  <div
+    v-else-if="qOptionType === 'multiBlanks'"
+    :class="['form-control', 'multiBlanks-div', hintClass]"
+    @mouseover="hover = true"
+    @mouseleave="hover = false"
+  >
+    <label :class="['row', 'm-0', { 'disabled': qState.answered }]" @click.stop>
+      <!-- <textarea
```
ang-zeyu, line 108: 👀

ang-zeyu (commented):
On the qoption styling:
- Should we remove the hover styling?

   It introduces some UI flickering and dosen't make too much sense since the entire row can't be clicked.

   hoverblanks
- Let's make sure we're using `cursor: text` for the `<input>` element too, since the input isn't actually a button.

`showIntermediateResult`: is this meant to allow seeing which answers you got correct "halfway"? I think that's a good idea, and can even be made the default. 
- Currently I'm seeing this behaviour in the docs (no difference from not specifying it), I think as a result of `threshold=0.5` in the example, which might be confusing:
   - type "elias" in the first box, click check -> the entire question is immediately answered (second box's answer is also displayed), without the opportunity to retry
- perhaps a better default for blanks (vs free-text) might be: `showIntermediateResult: true, threshold: 1`, especially since the probability of getting it right (vs free-text) is higher. (per your test case)
   - docs can be updated accordingly with another example to show the other case too

Some other minor nits below:

weiquu:
Thanks @ang-zeyu for the comments! I've edited accordingly, and also changed `showIntermediateResult` to `hideIntermediateResult`, so that the default is to show, and then they can specify the `hideIntermediateResult` / `hide-intermediate-result` option should they not want this behaviour.

packages/vue-components/src/questions/QOption.vue
```diff
@@ -214,4 +307,27 @@ export default {
...
+        cursor: text;
+    }
+
+    input.form-control:disabled,
+    .blanks-keywords {
+        margin-bottom: 0.5rem;
+    }
+
+    .blanks-form-control {
+        border: none;
```
ang-zeyu, line 326:
```suggestion
        border: none;
        cursor: default;
```

(pointer around the input edge)

docs/userGuide/syntax/questions.md
```diff
@@ -217,6 +217,83 @@ correct | `Boolean` | `false` | Whether this option (placed under either a MCQ o
...
+#### Fill-in-the-Blanks Questions {.mt-4 .mb-3}
+
+Fill-in-the-blanks questions are specified with the `type="blanks"` attribute.
+
+Unlike MCQ and checkbox questions, answer checking is performed for each blank by providing keywords to check for in the user's answer through the `keywords` attribute in each `q-option`.
+If no keywords are provided, the answer for that blank will always be marked as correct.
+
+<box type="warning" seamless>
+
+Keywords are validated by checking if the keyword matches the user's answer exactly (ignoring case)!
```
ang-zeyu, line 229:
```suggestion
Keywords are validated by checking if the keyword matches the user's answer exactly (ignoring letter casing).
```

docs/userGuide/syntax/questions.md
```diff
@@ -217,6 +217,83 @@ correct | `Boolean` | `false` | Whether this option (placed under either a MCQ o
...
+  <q-option keywords="figuration"></q-option>
+</question>
+{% endset %}
+
+<include src="codeAndOutput.md" boilerplate>
+<variable name="highlightStyle">html</variable>
+<variable name="code">{{ blanksQuestion }}</variable>
+</include>
+
+You can also specify a `hide-intermediate-result` attribute to hide the result of each blank upon submitting their attempt.
```
ang-zeyu, line 257: Let's give some context first (e.g. "by default, ..... Specify the .... to avoid this."), as the reader might not have reproduced this case.

docs/userGuide/syntax/questions.md
```diff
@@ -217,6 +217,83 @@ correct | `Boolean` | `false` | Whether this option (placed under either a MCQ o
...
+<box type="tip" seamless>
+
+Since the validation is imperfect, the minimum proportion of correct blanks needed for the entire question to be marked as correct can also be changed using the `threshold` attribute.
+
+If you don't want to validate the answer at all, you may set the `threshold` attribute to `0`. Doing so always marks the blank as correct, and users will be able to see the intended answers.
+
+</box>
+
+{% set blanksQuestion2 %}
+<question type="blanks" hint="What properies would you want these database transactions to have?" threshold=0.75 hideIntermediateResult>
```
ang-zeyu, line 268:
```suggestion
<question type="blanks" hint="What properies would you want these database transactions to have?" threshold=0.75 hide-intermediate-result>
```

docs/userGuide/syntax/questions.md
```diff
@@ -217,6 +217,83 @@ correct | `Boolean` | `false` | Whether this option (placed under either a MCQ o
...
+<include src="codeAndOutput.md" boilerplate>
+<variable name="highlightStyle">html</variable>
+<variable name="code">{{ blanksQuestion2 }}</variable>
+</include>
+
+****Fill-in-the-Blanks Question specific Options and Slots****
+Name | Type | Default | Description
+--- | --- | --- | ---
+threshold | `Number` | `0.5` | Minimum proportion of keywords that have to be matched in the user's answer for the answer to be marked as correct.
+hide-intermediate-result | `Boolean` | `False` | Does not show the result of each blank after an attempt.
```
ang-zeyu, line 288:
```suggestion
hide-intermediate-result | `Boolean` | `False` | Hides the result of each blank after an incorrect attempt.
```

ang-zeyu (commented):
few more nits:

docs/userGuide/syntax/questions.md
ang-zeyu: imo:
- `hide-interim` seems a little too short / vague
- `feedback` seems longer

Some more suggestions:
- `no-intermediate-result` (following @tlylt's `no` suggestion above + combining your original one =P)
- `no-intermediate-hints`
- `no-incorrect-hints`
- `no-partial-hints`

weiquu: oki! will likely go with `no-intermediate-result` so there's no confusion with the `hint` attribute

ang-zeyu (approved):
Lgtm 👍

*edited the commit message a little for conciseness

packages/vue-components/src/__tests__/Questions.spec.js
```diff
@@ -274,6 +274,213 @@ describe('Checkbox Questions and QOptions', () => {
...
+          option,
+        ],
+        header: 'unanswered blanks question header',
+        hint: 'blanks question hint',
+      },
+      provide: DEFAULT_INJECTIONS,
+      stubs: DEFAULT_STUBS,
+    });
+
+    // Click hint
```
tlylt, line 302: @weiquu Sorry for the late review, just a minor nit on maintaining consistency if possible.
In your comments within this file and the next, you have:
- `// Click hint`,
- `// click 'check'`,
- `// click 'hint'`,
- `// Click check of blanks question`

Sometimes there's a quotation, sometimes the comment is capitalized, and sometimes not. Perhaps can standardize a little in future PRs.

tlylt:
@all-contributors please add @weiquu for code, doc

PR #1959, opened 2022-07-06
Set search bar width to placeholder text

### LINKED ISSUE #665: Allow setting the width of search bar
The problem:

image

weiquu:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Resolves #665 
Continuation of PR #1957 (was resetting the dist file changes)

**Overview of changes:**
Binds the width of the search bar to the length of the placeholder text so as to avoid clipping of the placeholder text.

Before:
search bar before

After:
search bar after

The minimum width of the search bar is set to its current width (i.e. the width in the "Before" screenshot above).

The width is set using the `ch` unit, which is defined according to the width of the `o` character in a given typeface. Hence, it might run into issues with certain non-standard typefaces, should the MarkBind typeface be changed in the future.

**Testing instructions:**
Use a long placeholder text to check if the width of the search bar expands accordingly.

Set search bar width to placeholder text

---

- [ ] No unrelated changes

### DIFF
--- a/packages/vue-components/src/Searchbar.vue
+++ b/packages/vue-components/src/Searchbar.vue
@@ -17,7 +17,7 @@
       @keydown.esc="reset"
       @blur="showDropdown = false"
     />
-    <div ref="placeholderDiv" class="form-control">
+    <div ref="placeholderDiv" class="form-control hidden">
       {{ placeholder }}
     </div>
     <ul ref="dropdown" :class="dropdownMenuClasses">
@@ -96,7 +96,7 @@ export default {
       noResults: true,
       current: 0,
       items: [],
-      placeholderDivWidth: 1,
+      placeholderDivWidth: 12.7, // initially set to min-width
     };
   },
   computed: {
@@ -264,7 +264,7 @@ export default {
 <style scoped>
     .form-control {
         min-width: 12.7em;
-        max-width: 25.4em;
+        max-width: 25.4em; /* twice of min-width, to accommodate a range of lengths */
     }
 
     .table-active {
@@ -274,6 +274,10 @@ export default {
     .dropdown-menu-end {
         right: 0;
         left: auto;
+    }
+
+    .hidden {
+        visibility: hidden;
     }
 </style>
 
--- a/packages/vue-components/src/Searchbar.vue
+++ b/packages/vue-components/src/Searchbar.vue
@@ -8,6 +8,7 @@
       type="text"
       class="form-control"
       :placeholder="placeholder"
+      :style="{ width: placeholder.length + 'ch', 'min-width': '12.7em' }"
       autocomplete="off"
       @input="update"
       @keyup.up="up"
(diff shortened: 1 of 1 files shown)


weiquu:
Hello! I've switched to another method for setting the width which I think works better. It uses a temporary div and sets the text inside to the placeholder:
```
<div ref="placeholderDiv" class="form-control">
    {{ placeholder }}
</div>
```
After the page has loaded, it gets the width of the div and sets the search bar width to be the same as that. Then, the temporary div is removed:
```
mounted() {
    const { placeholderDiv } = this.$refs;
    this.placeholderDivWidth = placeholderDiv.offsetWidth;
    placeholderDiv.remove();
}
```
As suggested, I've also added in a minimum width of 12.7em and a maximum width of 25.4em.

This results in this behaviour for a short placeholder:
very short placeholder
This behaviour for a normal placeholder:
normal placeholder
And this behaviour for a long placeholder:
very long placeholder

Is this method fine, and should I commit the changes?

tlylt:
> Hello! I've switched to another method for setting the width which I think works better. It uses a temporary div and sets the text inside to the placeholder:
> 
> ```
> <div ref="placeholderDiv" class="form-control">
>     {{ placeholder }}
> </div>
> ```
> 
> After the page has loaded, it gets the width of the div and sets the search bar width to be the same as that. Then, the temporary div is removed:
> 
> ```
> mounted() {
>     const { placeholderDiv } = this.$refs;
>     this.placeholderDivWidth = placeholderDiv.offsetWidth;
>     placeholderDiv.remove();
> }
> ```
> 
> As suggested, I've also added in a minimum width of 12.7em and a maximum width of 25.4em.
> 
> This results in this behaviour for a short placeholder: very short placeholder This behaviour for a normal placeholder: normal placeholder And this behaviour for a long placeholder: very long placeholder
> 
> Is this method fine, and should I commit the changes?

Hi, @weiquu sounds good to me! Just want to confirm if it works with live reload? i.e. when we are developing locally and after changing the placeholder, will the browser reloads correctly with the correct width?

weiquu:
> Hi, @weiquu sounds good to me! Just want to confirm if it works with live reload? i.e. when we are serving the site locally and after changing the placeholder, will the browser reload correctly with the correct width?

Yups, the browser will reload correctly with the correct width!

ong6 (approved):
LGTM, works on mac too

packages/vue-components/src/Searchbar.vue
```diff
@@ -253,7 +263,8 @@ export default {
 
 <style scoped>
     .form-control {
-        min-width: 8em;
+        min-width: 12.7em;
+        max-width: 25.4em;
```
jovyntls, line 267: What's the rationale for the max-width? I think it could help to add a comment to explain in case someone would want to change it in the future :)

jovyntls (changes requested):
The code looks good! I'm getting some FOUC errors, search bar appears twice then once:

https://user-images.githubusercontent.com/61113575/179404160-e8518ad6-1750-4b2a-9121-4e7ad3ff23d8.mov

weiquu:
Thanks @jovyntls for the comments! I've added an explanation to the max-width line, as well as hidden the temporary div so that only 1 search bar is seen.

Issue #1937, opened 2022-05-23
Include caveats for updating functional testcases in DG


tlylt:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

#1130

### What is the area that this feature belongs to?

Documentation

### Is your feature request related to a problem? Please describe.

Due to issue #1130, there might be unrelated changes to png files (generated by plantuml) when updating functional test cases. Before the issue is resolved, perhaps it will be good to mention the need to ignore .png files in our developer guide so newcomers will know how to update functional tests properly.

### Describe the solution you'd like

Put in a note somewhere here: https://markbind-master.netlify.app/devguide/workflow#updating-and-writing-tests

### Describe alternatives you've considered

Manual reminder, e.g https://github.com/MarkBind/markbind/pull/1652#issuecomment-907950780

ang-zeyu:
Perhaps to unblock that issue we can resort to the following:
- add the puml outputs to gitignore / untrack them
- manually check, in our test scripts for these files (should be easy to maintain the list since just 9/10 files)

tlylt:
> Something like this, the ability can be moved to our test scripts:
> 
> ```
> const pumlFiles = [ ... ] // maintain this manually
> pumlFiles.forEach(() => {
>   assert( ... exists ...)
> })
> ```

I assume you mean:
1. Delete the puml files
2. Add them to gitignore
3. Add assertions to check for existence in test scripts

I think this works fine except for the case whereby someone who has never seen the generated puml files (perhaps he just cloned the repo...actually also applies to any dev but more serious for newcomers) and work on some features which affected the puml files. He then runs the tests and sees that the tests passed, yet
- the puml files are generated but could be erroneous, but he is not aware and won't be checking the content to find out
- the puml files look "different" (which may be undesirable), but he is not aware because he does not have a reference to the previous version even if he manually inspects the puml files

ang-zeyu:
> I assume you mean:

yup

> the puml files are generated but could be erroneous, but he is not aware and won't be checking the content to find out
> the puml files look "different" (which may be undesirable), but he is not aware because he does not have a reference to the previous version even if he manually inspects the puml files

Hmm I don't see how this would be any different, we are only checking for existence of files at the current too. (via expected - generated folder difference)

> the puml files look "different" (which may be undesirable), but he is not aware because he does not have a reference to the previous version even if he manually inspects the puml files

[...]

Issue #486, opened 2018-12-22
Apply smooth scrolling when collapsing a panel from the bottom


damithc:
Current: When a panel is collapsed using the button at the bottom of a panel (example below), it can often result in all content visible to the reader being replaced by some new content. The reader is teleported to a whole new location of the page and will feel 'lost'.

![bottom collapse](https://user-images.githubusercontent.com/1673303/50371607-12af8880-05f9-11e9-9257-963e53e8ed77.gif)

Suggested: make the transition smoother so that the reader does not 'lose the place' in the page. For example, the bottom edge of the panel should remain visible to the reader throughout the entire transition and the scrolling can be smooth rather than 'jump'.

PR #1965, opened 2022-07-17
Add conditional panel closing transition based on height

### LINKED ISSUE #1836: "Cap" / remove panel closing animations

When the user closes the panel, the viewport is "dragged" back up to the panel's start.

This works well for smaller panels but becomes rather disorienting for larger ones.

Either:
- Only activate the closing transitions if the panel contents are less than some fixed height (I think preferably as the transition isn't too big)
- Completely remove the _closing_ transitions

eeliana:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Resolves #1836 

**Overview of changes:**
Only activate closing transition for panels when the height is below a certain height.
Kapture 2022-07-17 at 10 12 17

Only adjusted closing transitions, opening transition remains the same.
Fixed height is 300 for now, can be adjusted.

**Testing instructions:**
Create a test site and put chunkier placeholder text into the panel to see the difference in closing transitions.

Add conditional panel closing transition based on height

---

- [x] No unrelated changes


tlylt:
Hi @eeliana, thank you for contributing!

Wonder if it is possible for you to add the two cases you showed in the gif in https://github.com/MarkBind/markbind/tree/master/packages/cli/test/functional/test_site so that this behavior is properly noted down in our functional test suite?

tlylt:
@ang-zeyu do you have any examples of disorienting panels when closing? Looking at some of the panels in 
https://markbind.org/userGuide/deployingTheSite.html#generating-a-github-personal-access-token and I feel that the closing animation is not too bad even when the content is long.

ang-zeyu:
@tlylt was trying out the panels in the 2103 pages https://nus-cs2103-ay2122s2.github.io/website/schedule/week3/topics.html, which can be really long. The problem is worse on slower + mobile devices as well where the animation gets dragged out beyond the specified duration / becomes really prominent.

Might just be me though (I'm not a huge fan of large animations in regular day-to-day things - e.g. mac's mission control), perhaps we can get @jonahtanjz's input too.

I should also add that its more so disorienting for me only when the scroll animation takes place (the top of the panel is above the viewport), combined with the panel's height transition.

If in agreement here perhaps, instead of just a fixed height, we can also add:

```
if top of panel beyond viewport (I think there's some existing checks for this too) && this.$refs.panel.scrollHeight > 300
  don't animate
```

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -141,6 +141,9 @@ export default {
           Thus, we need to reset the maxHeight to its current height for collapse transition to work.
         */
         this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
+        if (this.$refs.panel.scrollHeight > 300) {
+          this.$refs.panel.style.transition = 'max-height 0s ease-in-out';
```
ang-zeyu, line 145: On that note, I noticed the jQuery scroll animation isn't disabled too, was thinking it should (sorry that the issue description wasn't clear enough!)

jonahtanjz:
> If in agreement here perhaps, instead of just a fixed height, we can also add:
>
>```
>if top of panel beyond viewport (I think there's some existing checks for this too) && this.$refs.panel.scrollHeight > 300
>  don't animate
>```

Agree with this. I think the panel transition is fine as long as the header is within the viewport. I feel that it would be nice to have some transition when the panel is closed rather than it closing abruptly.

The transition only becomes disorienting when the panel is really long and causes the page to scroll when it is collapsed:

https://user-images.githubusercontent.com/47470981/179396296-72b32c22-5e1b-42b7-aa65-99951b80413e.mp4

tlylt:
Hi @eeliana, you may need to address the functional test failures (See this part https://github.com/MarkBind/markbind/runs/7374887371?check_suite_focus=true)

Read https://markbind-master.netlify.app/devguide/workflow#updating-and-writing-tests to find out how to update the test snapshot.

ong6:
Just a tip for updating tests: 

1. To update tests you will have to run `npm run updatetest` after adding the new tests.
2. Following this remove all the files which are not relevant to the changes (like picture files and files with no actual changes) 
3. Run `npm run test` again to see if it works

eeliana:
> Great work with this PR!
> 
> Just a tip for updating tests:
> 
> 1. To update tests you will have to run `npm run updatetest` after adding the new tests.
> 2. Following this remove all the files which are not relevant to the changes (like picture files and files with no actual changes)
> 3. Run `npm run test` again to see if it works

Hi, I have done this and I don't see the same issue after `npm run test`. I'm not too sure what the issue is here. Can I get some help?

tlylt:
> > Great work with this PR!
> > Just a tip for updating tests:
> > 
> > 1. To update tests you will have to run `npm run updatetest` after adding the new tests.
> > 2. Following this remove all the files which are not relevant to the changes (like picture files and files with no actual changes)
> > 3. Run `npm run test` again to see if it works
> 
> Hi, I have done this and I don't see the same issue after `npm run test`. I'm not too sure what the issue is here. Can I get some help?

i don’t think you have updated the snapshot, have you tried running npm run updatetest? What @ong6 was referring to is that you should git add only the relevant generated files, and discard the changes to image files (and any other files that at the moment are incorrectly flagged)

eeliana:
> > > Great work with this PR!
> > > Just a tip for updating tests:
> > > 
> > > 1. To update tests you will have to run `npm run updatetest` after adding the new tests.
> > > 2. Following this remove all the files which are not relevant to the changes (like picture files and files with no actual changes)
> > > 3. Run `npm run test` again to see if it works
> > 
> > 
> > Hi, I have done this and I don't see the same issue after `npm run test`. I'm not too sure what the issue is here. Can I get some help?
> 
> i don’t think you have updated the snapshot, have you tried running npm run updatetest? What @ong6 was referring to is that you should git add only the relevant generated files, and discard the changes to image files (and any other files that at the moment are incorrectly flagged)
> 
> Also, you should be able to run “npm run test” locally to check if the tests are passing locally , if they aren’t then it’s unlikely that they will pass in CI

  Yup, I ran `npm run updatetest` before `npm run test`. Is the issue because I did not commit the generated files? I can't seem to reproduce the `Unequal number of files` error locally although the check is failing because of it.

tlylt:
> > > > Great work with this PR!
> > > > Just a tip for updating tests:
> > > > 
> > > > 1. To update tests you will have to run `npm run updatetest` after adding the new tests.
> > > > 2. Following this remove all the files which are not relevant to the changes (like picture files and files with no actual changes)
> > > > 3. Run `npm run test` again to see if it works
> > > 
> > > 
> > > Hi, I have done this and I don't see the same issue after `npm run test`. I'm not too sure what the issue is here. Can I get some help?
> > 
> > 
> > i don’t think you have updated the snapshot, have you tried running npm run updatetest? What @ong6 was referring to is that you should git add only the relevant generated files, and discard the changes to image files (and any other files that at the moment are incorrectly flagged)
> > Also, you should be able to run “npm run test” locally to check if the tests are passing locally , if they aren’t then it’s unlikely that they will pass in CI
> 
> Yup, I ran `npm run updatetest` before `npm run test`. Is the issue because I did not commit the generated files? I can't seem to reproduce the `Unequal number of files` error locally although the check is failing because of it.

Yes you will need to commit the generated files, they are generally in packages/cli/test/functional/test_site/expected

take note of the caveat in https://github.com/MarkBind/markbind/issues/1937

packages/vue-components/src/panels/PanelBase.js
ang-zeyu: @tlylt @eeliana in case you missed this.

It's nowhere near as disorienting with the panel height transition removed, but I'm not entirely sure if this should be removed as well. (personally, still favouring removal as there is a slight time lag till the scroll takes place)

If it looks ok / better to you guys feel free to keep it :)

tlylt:
@ang-zeyu I am not sure if the disorienting issue can be resolved even by removing the transition effect. Trying it out locally with the changes from this branch:
0X6B3y1jCq

For pages like the one below, with a panel that is not too long, I prefer the previous closing animation since it looks fine to me. With the new changes, the closing is more abrupt and it may be a little surprising if another shorter panel nearby has an animation and suddenly the next panel doesn't have it
- Before:
  - https://markbind.org/userGuide/deployingTheSite.html#deploying-via-github-actions
- After:
  - https://deploy-preview-1965--markbind-master.netlify.app/userguide/deployingthesite#deploying-via-github-actions

What do you think?

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -141,6 +141,9 @@ export default {
           Thus, we need to reset the maxHeight to its current height for collapse transition to work.
         */
         this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
+        if (this.$refs.panel.scrollHeight > window.innerHeight) {
```
ang-zeyu, line 144: > @ang-zeyu I am not sure if the disorienting issue can be resolved even by removing the transition effect. Trying it out locally with the changes from this branch:

I don't think this implementation is correct, @tlylt the case in your gif shouldn't occur ⬇️ 

> if top of panel beyond viewport (I think there's some existing checks for this too) 

the relevant existing check is here https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/PanelBase.js#L147

ang-zeyu: > For pages like the one below, with a panel that is not too long, I prefer the previous closing animation since it looks fine to me. With the new changes, the closing is more abrupt and it may be a little surprising if another shorter panel nearby has an animation and suddenly the next panel doesn't have it

If corrected, does it account for the 2 links you linked too?

If not, we could try https://github.com/MarkBind/markbind/pull/1965#discussion_r922768335, or even remove the closing transitions completely for UX consistency (⭐I think this might be better based on what you've mentioned⭐).

Again this is all very subjective, and I'm not huge fan of large animations (fwiw I'm the original implementor of this feature, haven't found the closing transitions anything apart from dizzying)
Our primary [issue](https://github.com/MarkBind/markbind/issues/486) back then was also that the page did not scroll back up to the top of the panel *at all*, much less the transitions which was a cool "afterthought" =P

If agreed here perhaps we should just remove closing transitions completely (including the jquery scroll).
@damithc @MarkBind/active-devs can chime in too.

maybe give a thumbs up 👍 or 👎 on this comment to get a consensus. :-)

damithc: I'm OK to remove the transition.

ang-zeyu: @eeliana based on the 👍 so far, I think we can go ahead with replacing any and all closing transitions (including the jquery scroll) with an "instant snap" back to the top of the panel. The opening ones can be kept.

packages/cli/test/functional/test_site/testPanelsClosingTransition.md
```diff
@@ -0,0 +1,39 @@
...
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+</panel>
+
+**Nested Panel with height below 300px**
+<panel header="Expanded panel with closing transition" type="info" >
```
tlylt, line 16: Test cases for this update need to be updated

tlylt (changes requested):

packages/vue-components/src/panels/PanelBase.js
tlylt: @ang-zeyu 
> If corrected, does it account for the 2 links you linked too?

Yes if all closing transitions are removed, then I think the behavior is uniform and I am ok with that.

> I don't think this implementation is correct, @tlylt the case in your gif shouldn't occur ⬇️

Re looking at this again, may need your advice on if there's anything special going on here... So the case in my gif is one that the closed panels are near the bottom of the page. When opening or closing, the height of the page could reduce or increase such that the panel moves away from the original position. Base on my observation, whether before or after the latest change done by @eeliana, it seems like at times the panel could move away from the original position and result in slight disorienting effect, or somehow it won't. Here's a few recordings:
1. You can see that the first closing of "Guideline: Maximum readability" result in the panel heading moving, while the open and closing of the subsequent times didn't cause that (which is what we want to achieve?). This is done with the change in this PR applied
2OJTbHhksz

2. You can also experiment with this on https://nus-cs2103-ay2021s1.github.io/website/se-book-adapted/chapters/codeQuality.html which is the original behavior:
a) works fine (seems like I am able to trigger this, by closing the panels one by one, avoiding any scrolling before I close all the panels)
3KeCqFsO4U
b) doesn't really work well (seems like I am able to trigger this, by scrolling around, before closing the panels)
sGEKNWxtzh

@eeliana Please feel free to experiment a little and follow up on this since it's your PR :)

ang-zeyu: > Re looking at this again, may need your advice on if there's anything special going on here...

I actually thought the issue you wanted to show in your gif was that the closing transition was not activating previously, in that case it should since the top of the panel is in view. (the check isn't correct for that)

> Yes if all closing transitions are removed, then I think the behavior is uniform and I am ok with that.

But not so much relevant since we're going with this now. 😶 

> Re looking at this again, may need your advice on if there's anything special going on here...

This seems separate, let's put up a another issue for this.

Since its occuring only on the first open, likely:
- there is some dynamically loaded content (e.g. `<panel src="...">`) inside that panel
- something to do with how we handle these updates (see [here](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/PanelBase.js#L188) and [here](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/NestedPanel.vue#L98)) and/or how chrome handles these viewport changes (on that note, maybe we could try different browsers too)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -135,26 +135,13 @@ export default {
...
-          Panel's maxHeight is 'none' at the moment, as event listener set it to 'none' after expansion.
-          Thus, we need to reset the maxHeight to its current height for collapse transition to work.
-        */
-        this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
-
-        requestAnimationFrame(() => {
-          // To enable behaviour of auto window scrolling during panel collapse
-          if (this.$el.getBoundingClientRect().top < 0) {
-            const headerHeight = jQuery('header[sticky]').height() || 0;
-            jQuery('html').animate({
```
ang-zeyu, line 149: let's update this with an instant scroll; this was the key issue we had back then. Again the transitions are just a "cool afterthought" =P

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -135,26 +135,13 @@ export default {
...
-          }
-          this.$refs.panel.style.maxHeight = `${this.collapsedPanelHeight}px`;
-        });
+        // Remove closing transition entirely
+        this.$refs.panel.style.transition = 'none';
+        this.$refs.panel.style.maxHeight = `${this.collapsedPanelHeight}px`;
       } else {
         // Expand panel
         this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
+        this.$refs.panel.style.transition = 'max-height 0.5s ease-in-out';
```
ang-zeyu, line 144: 👀  could this go before the previous line? I don't think there's any browser guarantees here but it would be best to ensure the sequence of operations on our side too.

ang-zeyu (commented):
the logic here needs to be updated too https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/MinimalPanel.vue#L123 (try collapsing a minimal type panel, the header would disappear)

packages/vue-components/src/panels/PanelBase.js
```diff
@@ -135,25 +135,19 @@ export default {
...
-        this.$refs.panel.style.maxHeight = `${this.$refs.panel.scrollHeight}px`;
-
         requestAnimationFrame(() => {
           // To enable behaviour of auto window scrolling during panel collapse
           if (this.$el.getBoundingClientRect().top < 0) {
             const headerHeight = jQuery('header[sticky]').height() || 0;
             jQuery('html').animate({
               scrollTop: window.scrollY + this.$el.getBoundingClientRect().top - headerHeight - 3,
-            }, 500, 'swing');
+            }, 0, 'swing');
```
ang-zeyu, line 144: Why not remove the `animate`?

`.scrollTop = ...`

tlylt:
Hi @eeliana, great work on this PR. Checking if you are still keen to implement the last few comments/fixes as requested?

tlylt:
cc @MarkBind/contributors pls help support this if possible, it's near completion.

yucheng11122017:
@tlylt I don't seem to have permission to edit the PR.

tlylt:
> @tlylt I don't seem to have permission to edit the PR.

Pls checkout from this branch and raise a follow-up PR instead.

tlylt:
Closing after work is merged in #2159, thank you @eeliana and @yucheng11122017!

Issue #205, opened 2018-04-25
Auto-search content of pages


damithc:
Current: only page titles and specified keywords in the frontmatter appear in search results.

Suggested: also include other content in pages for search results

damithc:
Raising priority as full-text search can greatly enhance the usefulness of a content-heavy website.

I don't mind full-text search is a separate page altogether and takes some time to load (i.e., if the full search index needs to be downloaded to the Browser first)

amad-person:
Are we open to integrating existing solutions for full-text search? 

[Docsearch](https://github.com/algolia/docsearch) (free, open-source):
> DocSearch will crawl your documentation website, push its content to an Algolia index, and allow you to add a dropdown search menu for your users to find relevant content in no time.

damithc:
> Are we open to integrating existing solutions for full-text search?

Ideally, we should have a decent built-in solution _and_ the ability to integrate other third-party solutions.

yamgent:
As discussed with @marvinchin today, Marvin is planning to explore using the [Lunrjs](https://lunrjs.com/) library to implement a built-in full text search. This library is also used by MkDocs.

marvinchin:
Are we still looking to have built-in full text search for V2? 😅  I'm not sure that I can finish it by the end of the semester.

damithc:
> Are we still looking to have built-in full text search for V2? 😅 I'm not sure that I can finish it by the end of the semester.

Good to have, but not necessary. Same for the FOUC problem. Both have a good-enough workaround but not a full-fledged solution.

ang-zeyu:
I've just published an almost year long project originally motivated by this issue:
- https://github.com/ang-zeyu/morsels
- https://ang-zeyu.github.io/morsels/

It consists of a cli file indexer (integratable by copying the binary similar to what we do for `plantuml.jar`), a search library powered by wasm (rust), and search ui (typescript).

It deals with the issue in 2 aspects:
- Scalability
  > I don't mind full-text search is a separate page altogether and takes some time to load (i.e., if the full search index needs to be downloaded to the Browser first)

  I found this issue to be common to many static site generators using lunrjs / some other client side search solution; This was my primary motivation in creating this project (see `https://github.com/olivernn/lunr.js/issues/222` and discussion here `https://github.com/rust-lang/mdBook/issues/51` for example), although it turned out to be a secondary plus in the end.

  The indexer is also created in rust as such (:star: indexes the entire 2103 site in `0.5s`!).
[...]

PR #1968, opened 2022-07-19
Avoid duplicating search results

### LINKED ISSUE #311: Avoid duplicating search results if main heading of the page matches page title
v1.8.4

Context: a page's heading (i.e., the heading that appears at the top of page) is very similar to the page title in the frontmatter

Problem: when the user search for a keyword that is in one (i.e., heading or the title), each appears as a separate result. 

Suggestion: show both as one result

Thoughts: while this sounds like a unavoidable problem, it is good if we can find a smart way (even if it takes a significant effort) to avoid this problem as it happens a lot and makes a significant impact on search UX

### LINKED ISSUE #1972: Search results show the same thing in both columns

### Any related issues?

_No response_

n/a

### MarkBind version

v4.*

The two columns of the search results almost always show the exact same content? Is that intended?
image

### Steps to reproduce

Go to https://markbind.org/ and search for some keywords using the search feature of the site.

### Expected behavior

-

### Actual behavior

-

### Anything else?

-

weiquu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Resolves #311 
Resolves #1972 

**Overview of changes:**
Add in a check to ensure that headers at the very top of the page which are an exact match of the title are not displayed in the search results. Also changed the design of the results dropdown to avoid duplicating results and to make clear which results are page titles and which are headers within the page.

Screenshots are in the comments below!

**Testing instructions:**
For the main test of the implementation, create a page with the page title the same as the first and second heading. The first heading should not show up, while the second heading does. 

For testing of heading keywords, create a page with the page title the same as the first heading, and with the first heading having some keyword not contained in heading/page title. Upon searching the keyword, only the page title should turn up (i.e. without the first heading below it).

For testing of UI/UX, use long headings and keywords, with some having no spaces within the words as well.

For page keywords and heading indexing, please see [this comment](https://github.com/MarkBind/markbind/pull/1968#discussion_r939712691) for the issues.

Update displaying of search results

Search results currently show a duplicate column for headings.
Topmost headings are also included in the search results even if
they are exactly the same as the title.

This duplication is redundant and could appear to the user as
being buggy due to the repetition.

To avoid these issue, let's
* remove the duplicate results column for headings
* remove the duplication of search results if the topmost heading
matches the title exactly

---

- [x] No unrelated changes


packages/vue-components/src/Searchbar.vue
```diff
@@ -166,10 +166,120 @@ export default {
         }
       });
 
+      // using distance with adjacent transpositions
```
ong6, line 169: Would be nice if you could explain what exactly this function does, since not everyone may know this algo.

ong6 (commented):
Looks pretty good, nice work here! Just clean up the comments and it'll LGTM.

weiquu:
Thanks @ong6! Have removed the commented out code and added an explanation.

ong6 (approved):
LGTM!

jonahtanjz (commented):
Nice work @weiquu!

You might want to add a note in the user guide (under [Making the Site Searchable](https://markbind-master.netlify.app/userguide/makingthesitesearchable#making-the-site-searchable) section) so that users are aware of this behaviour :)

packages/vue-components/src/Searchbar.vue
```diff
@@ -166,10 +166,86 @@ export default {
...
+          matrix[i + 1][0] = maxDist;
+          matrix[i + 1][1] = i;
+        }
+        for (let j = 0; j <= s2.length; j += 1) {
+          matrix[0][j + 1] = maxDist;
+          matrix[1][j + 1] = j;
+        }
+
+        let cost = 0;
+        let db = 0;
```
tlylt, line 194: Might be a dumb question: is `db` supposed to mean something?

packages/vue-components/src/Searchbar.vue
```diff
@@ -166,10 +166,86 @@ export default {
...
+          return true;
+        }
+
+        // For anything else, we check if the edit distance is less than or equals to the limit set above.
+        const dist = calcDamerauLevenshtein(heading, title);
+        if (dist <= limit) {
+          return true;
+        }
+
+        return false;
```
tlylt, line 240:
```suggestion
        return dist <= limit;
```

packages/vue-components/src/Searchbar.vue
```diff
@@ -166,10 +166,86 @@ export default {
...
+        return matrix[s1.length + 1][s2.length + 1];
+      }
+
+      function isDuplicateResult(page) {
+        // If there is more than 1 header, we keep all headers in the results.
+        if (page.headings.length !== 1) {
+          return false;
+        }
+
+        const heading = page.headings[0].heading.text.toLowerCase().trim();
```
tlylt, line 225: I think we can be more defensive here and check for the existence (right at the beginning) of headings and their length before proceeding

packages/vue-components/src/Searchbar.vue
```diff
@@ -166,10 +166,86 @@ export default {
...
+            const transposition = matrix[k][l] + (i - k - 1) + 1 + (j - l - 1);
+            matrix[i + 1][j + 1] = Math.min(substitution, insertion, deletion, transposition);
+          }
+          alphabet[s1[i - 1].charCodeAt(0)] = i;
+        }
+
+        return matrix[s1.length + 1][s2.length + 1];
+      }
+
+      function isDuplicateResult(page) {
```
tlylt, line 219: Perhaps we can give this a better name? Or minimally a JSdoc, because reading it now "isDuplicateResult(page)" is a little vague, what does "result" mean here?
What do you think?

Perhaps what I can suggest is: hasSimilarPageTitleAndHeading(page)

tlylt (changes requested):

ang-zeyu:
Thanks for the PR @weiquu. Is this what you had in mind? (see OP description) @damithc 

I'm not sure if these checks (in particular edit distance) would be generally applicable enough, some examples:

multilingual:

```
title: any 3 chinese characters

# any other 3 chinese characters
```

even with latin / ascii text, there are issues:

```
title: Today

# Todo // semantics are very different, both should be presented
```

We can scale the edit distance limit by the string length, but that is still susceptible:

```
// if we adopted limit <= 1 for strings of length <= 5,
title: Today

# Koday // don't really have the time now to think of a proper example, but imagine this is a product name, etc. =|
```


I'm supposing this issue was raised because in many sites, the *level 1 heading* in particular is often a "repeat" of the title, so I wonder if we should instead be supporting a range for the [heading indexing level](https://markbind.org/userGuide/siteJsonFile.html#headingindexinglevel).

e.g. `"headingIndexingLevel": "2-4"`,


exceptions can be dealt with by using the `always/no-index` class we already [have](https://markbind.org/userGuide/makingTheSiteSearchable.html#including-or-excluding-headings)

tlylt:
> Thanks for the PR @weiquu. Is this what you had in mind? (see OP description) @damithc
> 
> I'm not sure if these checks (in particular edit distance) would be generally applicable enough, some examples:
> 
> multilingual:
> 
> ```
> title: any 3 chinese characters
> 
> # any other 3 chinese characters
> ```
> 
> even with latin / ascii text, there are issues:
> 
> ```
> title: Today
> 
> # Todo // semantics are very different, both should be presented
> ```
> 
> We can scale the edit distance limit by the string length, but that is still susceptible:
> 
> ```
> // if we adopted limit <= 1 for strings of length <= 5,
> title: Today
> 
> # Koday // don't really have the time now to think of a proper example, but imagine this is a product name, etc. =|
> ```
> 
> I'm supposing this issue was raised because in many sites, the _level 1 heading_ in particular is often a "repeat" of the title, so I wonder if we should instead be supporting a range for the [heading indexing level](https://markbind.org/userGuide/siteJsonFile.html#headingindexinglevel).
> 
> e.g. `"headingIndexingLevel": "2-4"`,
> 
> exceptions can be dealt with by using the `always/no-index` class we already [have](https://markbind.org/userGuide/makingTheSiteSearchable.html#including-or-excluding-headings)

I agree with the point raised..have similar concerns about the effectiveness of the edit distance check. On a separate note, seems like the current implementation won’t be able to support languages other than English (iirc)

damithc:
> I'm supposing this issue was raised because in many sites, the _level 1 heading_ in particular is often a "repeat" of the title.

Yes, that's the specific problem I'm trying to fix. 

Here is an example:
image

code:
```html
<frontmatter>
  title: Binary Search Tree
</frontmatter>

# Binary Search Tree

> More content to be added
```

I think the problem case occurs when the page title is **repeated exactly** as the very first heading, at the **very top of the page** (as the case above). In such a case, the two search results goes to the exact same place in the site. Hence, the second result is redundant and an eyesore too (at first glance it might even look like search results are buggy as the same words are repeated thrice in the same place).

damithc:
On a separate note, these two columns of the search results almost always show the exact same content? Is that intended?
image

weiquu:
Thanks everyone for the feedback! Would a better solution be to simply check if the level 1 heading is contained within the page title and remove it if so (regardless of whether multiple headings are found by the search)? Then we can do away with the edit distance, which removes the language issues raised by @ang-zeyu and also solves the issue that @damithc wants to fix.

Not too sure about the separate issue as the dropdown looked like that previously! I can try making it into a single column if needed (:

damithc:
> Not too sure about the separate issue as the dropdown looked like that previously! I can try making it into a single column if needed (:

If this is an issue, it is likely a separate issue, in which case we should create a separate issue and fix it separately.



> Would a better solution be to simply check if the level 1 heading is contained within the page title and remove it if so

_contained within_ or _same as_?
Also, the problem occurs only when the offending heading is at the top of the page. If it is in a different location, the two search results point to two different places in the page, and hence, not technically duplicates.

ang-zeyu:
> which removes the language issues raised by @ang-zeyu

not quite unfortunately, a substring check is also problematic. In general I don't think we can avoid making assumptions (which we should avoid) about the page content with the current solutions in the PR:

```
// e.g.
title: 3D Procedural Game Engine Using OpenGL

# OpenGL
```

weiquu:
Oki I'll change it to checking if the heading is at the top of the page and is exactly the same as the title!

ang-zeyu:
> I'll change it to checking if the heading is at the top of the page

Hmm that's certainly one way, but detecting if the heading is at the top may be slightly difficult. (html comments, frontmatter), and/or not very performant (extracting the html text then doing something like a `startsWith` check)

> so I wonder if we should instead be supporting a range for the [heading indexing level](https://markbind.org/userGuide/siteJsonFile.html#headingindexinglevel).

What do you think about this?

ping @damithc @MarkBind/active-devs as well

damithc:
> > so I wonder if we should instead be supporting a range for the [heading indexing level](https://markbind.org/userGuide/siteJsonFile.html#headingindexinglevel).
> 
> What do you think about this?

Because the problem is specific to some pages only, omitting all H1 is not an option. We can let the author configure it at page level but that is a lot of work for the author. Plus the author might not be aware of this problem at all, or which pages are affected. So, a 'it just works' solution that the author doesn't even have to know about is best but I realize performance can be affected.

As @ang-zeyu noted, it's good if the solution works for Algolia as well, but seems like Algolia search results are not affected by this issue. Apparently, the page title is not shown in Algolia search results. Should we do something similar?

image

image

Also note that #1972 makes this problem appear worse (i.e., repeats the same thing thrice). Perhaps we tackle that one first?

ang-zeyu:
> > I'll change it to checking if the heading is at the top of the page
>
> Hmm that's certainly one way, but detecting if the heading is at the top may be slightly difficult. (html comments, frontmatter), and/or not very performant (extracting the html text then doing something like a `startsWith` check).
>
> *just a rough estimate though, if there's an easy way to do it that would be great.

@weiquu To be clear again I might _very, very_ well be wrong here, feel free to dig around the html processing [source](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/NodeProcessor.js).

> but seems like Algolia search results are not affected by this issue.

That might be another solution. I think Algolia always uses `h1` as its top level item, iirc Google does this to some degree as well.

But I think this comes after the first solution (detecting if heading is at the top of the page). We can also prioritise top-level, first `h1` but fallback to `title`.

---

On a slightly unrelated note... I liked Algolia docsearch v2's UI better. Having level 2 / 3 items (anchors) displayed on the same level as level 0 items (h1) is visually confusing for me and rather unconventional to understand the SERP hierarchy. The hashtag icon does not help. =|
*To be clear, this random rant is separate from the idea of using `h1` as a level 0 item.

weiquu:
Fixed the issue in #1972 by expanding out the left column and adding a hashtag to indicate that the result is a heading within the page. Also added a quick check to ensure that headings at the very top of the page which match the title exactly do not get shown in the search results. Below is an example with the page title being `Using Plugins` and 4 headings (in order) `Using Plugins`, `Using Plug-ins`, `Using Plugins`, and `Using Plug-in`. As seen, the first heading is removed from the search results:
Search Results

Here is the search results with long headings:
Long Heading

ang-zeyu:
> Also added a quick check to ensure that headings at the very top of the page which match the title exactly do not get shown in the search results. Below is an example with the page title being Using Plugins and 4 headings (in order) Using Plugins, Using Plug-ins, Using Plugins, and Using Plug-in. As seen, the first heading is removed from the search results:

I was thinking of a few other edge cases we can deal with later on by looking at the page content at build time instead, although the implementation might be a little more complex.


e.g.
```
lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem lorem 

... very long content ...

# Ipsum
```

```
# Lorem {.no-index}

...

# Ipsum
```

Unless there's any objections, I think this simple check is a good enough stopgap for most cases, and since document recall (which is more critical) ultimately isn't affected by the exact check 🙂👍 

> Fixed the issue in https://github.com/MarkBind/markbind/issues/1972 by expanding out the left column and adding a hashtag to indicate that the result is a heading within the page. 

Thanks for the fix. Let's move this to another PR though, the scope of the change is very different.

You may also need to check if this plays correctly and nicely with [keywords](https://markbind.org/userGuide/makingTheSiteSearchable.html#keywords).

packages/vue-components/src/Searchbar.vue
```diff
@@ -134,6 +134,7 @@ export default {
 
         if (totalPageMatches > 0) {
           const pageHeadings = [];
+          let first = true;
           Object.entries(headings).forEach(([id, text]) => {
```
ang-zeyu, line 138: would `idx === 0` (from `forEach`) work too?

weiquu:
Okay have updated the code to use index instead! I've also tested it with keywords (page and headings), and based on my tests it works fine.

tlylt:
> Okay have updated the code to use index instead! I've also tested it with keywords (page and headings), and based on my tests it works fine. Have also updated the commit message (:

Hi @weiquu, a quick note on the commit message in case you missed it:
image

Feel free to read up on https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message for our commit message format. I think this change is non-trivial, perhaps we can update the commit title and body to summarize the change done 🦾

damithc:
> Feel free to read up on https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message for our commit message format.

Moved to https://se-education.org/guides/conventions/git.html (OSS-generics is now defunct).

weiquu:
Thanks, have updated the commit message accordingly!

packages/vue-components/src/SearchbarPageItem.vue
```diff
@@ -1,8 +1,10 @@
 <template>
   <div v-if="item.heading" class="heading">
+    <i class="fa fa-hashtag"></i>
     <div class="heading-text">
-      <span v-html="item.heading.text"></span>
+      <span v-html="highlight(item.heading.text, value)"></span>
     </div>
+    <!--
```
ang-zeyu, line 7: > You may also need to check if this plays correctly and nicely with [keywords](https://markbind.org/userGuide/makingTheSiteSearchable.html#keywords).
> > I've also tested it with keywords (page and headings), and based on my tests it works fine.

I actually meant moving the keywords to a separate location (e.g. below the heading as smaller, muted text), not removing them entirely.

But this looks ok to me as well (as long as it turns up).

Let's remove the dead code (one more below as well)

On a related note https://github.com/MarkBind/markbind/issues/205 would change the design completely again... but don't think I'm going to be able to complete that anytime soon.

weiquu: Ah sorry for misunderstanding! I've added in the keywords to the search results, looks like this:

keywords long

Have also removed the dead code (:

tlylt: Hi @weiquu, thanks for the update. A gentle reminder to provide testing instructions in the future, so that reviewers can get back to you quicker:) For the screenshot you posted above, I tried it on the PR preview in https://deploy-preview-1968--markbind-master.netlify.app/ and then realized that it won't work because you have added some keywords locally to show this behavior.

Here's some screenshots of existing keywords and how they look like with this update:
- hide minimized panels
image
- regress
image

Are the words suppose to start on a newline?

Lastly, may I double check if the content here is still valid after your change @weiquu ?
https://markbind.org/userGuide/makingTheSiteSearchable.html
Specifically:
- does keyword still work for heading and an entire page?
- does the `.always-index or .no-index` still work,  in particular, if the heading is the very first one?

weiquu: Thanks for the reminder! Will update the testing instructions.

For the newline issue, can I check if you went into the PR preview and directly typed in the search words? I tried and there's no newline for me (e.g. regress below).
regress

I've checked through the content in the page linked, responding to the 2 pointed out:

For the `.always-index` and `.no-index`, I've tested with the code below:
```
<span id="title" class="d-none">Using Plugins</span>

<frontmatter>
  title: "Using Plugins"
</frontmatter>

###### Using Plugins {.always-index}

# Using Plugins {.no-index}
```
With this code, both don't show up (as expected), but if the positions of the 2 headings are swapped around (h1 then h6), then both don't show up again since the h6 heading is taken as the first indexed heading.

For the keywords, they do also work with the heading. However, I'm not too sure if I'm doing something wrong with the  page heading since it doesn't seem to work even if I revert back to the main build. With this code, I'm not seeing anything coming up with search word `plugin` even though I believe the expected result should be seeing the page title `Formatting Contents`:
```
<span id="title" class="d-none">Formatting Contents</span>

<frontmatter>
  title: "Formatting Contents"
  keywords: plugin
</frontmatter>

# Formatting Contents (Advanced)
```

Would appreciate any advice on the above 2 issues!

ang-zeyu: > does the .always-index or .no-index still work, in particular, if the heading is the very first one?
>
> > With this code, both don't show up (as expected), but if the positions of the 2 headings are swapped around (h1 then h6), then both don't show up again since the h6 heading is taken as the first indexed heading.

I might be lacking some context here but imo this change should have nothing to do with these attributes. (current behaviour is fine)

E.g.

```
<frontmatter>
  title: "Using Plugins"
</frontmatter>

// Should not show. What we index =/= user friendliness of the SERP we generate.
# Using Plugins {.always-index}
```

> does keyword still work for heading and an entire page?
>
> > For the keywords, they do also work with the heading. However, I'm not too sure if I'm doing something wrong with the page heading since it doesn't seem to work even if I revert back to the main build. With this code, I'm not seeing anything coming up with search word plugin even though I believe the expected result should be seeing the page title Formatting Contents:

That's a good catch. I think we can add to the check something along the lines of `!(idx === 0 && ... && !keywords.length)`.

ang-zeyu: On a separate note, a 🔑 icon conveys to me *security* - e.g. a password, authentication, secret key, etc. Might be confusing to readers what exactly the words following are.

I think we ought to swap up or remove the icon. 🤔 

If there's another suitable icon, let's limit it to displaying just once too so it dosen't get too repititive.

tlylt: > For the newline issue, can I check if you went into the PR preview and directly typed in the search words? I tried and there's no newline for me (e.g. regress below).

@weiquu see gif below, seems like it depends on window size:
OQyxxAH6uv

> On a separate note, a 🔑 icon conveys to me security

How about a 🎯 ?

> I might be lacking some context here but imo this change should have nothing to do with these attributes.

Was checking if these attributes will be affected by this implementation, because of how we  ignore duplicates (what if the duplicates have been marked with the always-index, no-index, does it still work correctly)

> For the keywords, they do also work with the heading. However, I'm not too sure if I'm doing something wrong with the page heading since it doesn't seem to work even if I revert back to the main build. 

Yes seems like it is an existing issue.

ang-zeyu: > 🎯

This conveys target - which means a goal, okr to me personally.

Would a muted search icon work?

On further thought I'm favouring removing the icon more strongly as well, rationale is that the author can also add unicode emojis themselves iirc.

ang-zeyu: > Was checking if these attributes will be affected by this implementation, because of how we ignore duplicates (what if the duplicates have been marked with the always-index, no-index, does it still work correctly)

🙆‍♂️

weiquu:
> I think we can add to the check something along the lines of `!(idx === 0 && ... && !keywords.length)`

Have added this in!

> Yes seems like it is an existing issue.

Okay, I'll leave it as is for this PR, should I open a new issue for it?

> On further thought I'm favouring removing the icon more strongly as well, rationale is that the author can also add unicode emojis themselves iirc.

Have removed the icon as well, do let me know if it works (or if I should try making the text smaller)

tlylt:
@weiquu 

> Okay, I'll leave it as is for this PR, should I open a new issue for it?

Sure, sounds good to me.

On a separate note, though there's no longer the icon issue (because there's no more icon ,-,), the `#` symbol also faces the same problem. Is this display of symbol and text on different line the expected behavior?
image

weiquu:
Okay have submitted the issue! I think I've also fixed the line break issue (wrong calculation of width)

packages/vue-components/src/SearchbarPageItem.vue
```diff
@@ -1,15 +1,13 @@
...
-      <span v-html="item.heading.text"></span>
-    </div>
-    <div class="heading-text-items">
-      <small v-html="highlight(item.heading.text, value)"></small>
-      <br />
+      <span v-html="highlight(item.heading.text, value)"></span>
       <small v-for="(keyword, index) in item.keywords" :key="index">
-        <span v-html="highlight(keyword, value)"></span>
-        <br />
+        <div class="heading-text">
```
tlylt, line 8: I see you have this div and class "heading-text" to ensure that the keyword starts on a new line. Wonder if you have tried simplifying it to `<div v-html="highlight(keyword, value)"></div>` and do without the class? Seems like it works for most cases. What do you think @weiquu?

weiquu: Thanks I think it works well! Have changed it

tlylt: Hi @weiquu, sorry for going back and forth on this one (the rest of the PR LGTM). On second thought I feel that this update might not be very semantic because small is an inline element while div is a block element, so inline should technically not contain a block.

 I am suggesting to change to the following to avoid possible complication down the line:

```
    <div class="heading-text">
      <span v-html="highlight(item.heading.text, value)"></span>
      <div
        v-for="(keyword, index) in item.keywords"
        :key="index"
        class="keyword-text"
        v-html="highlight(keyword, value)"
      ></div>
    </div>

.keyword-text {
  font-size: smaller;
}
```

(P.S then again, the rule is broken in the dropdown level outside of this, where an inline `<a>` wraps the `<div>`, but things still work fine...) pinging @ang-zeyu @jonahtanjz for advice on this 👁️  
image

ang-zeyu: > (P.S then again, the rule is broken in the dropdown level outside of this, where an inline <a> wraps the <div>, but things still work fine...) pinging @ang-zeyu @jonahtanjz for advice on this 👁️

This is actually ok (*not* the `<small><div>...` though).

Some links: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties 
https://html.spec.whatwg.org/#transparent-content-models
 to validate the markup - https://validator.w3.org/nu/#textarea 
You can also check out how algolia docsearch structures its html - https://docusaurus.io/ (just using docusaurus as a convenient example)

I am nowhere an expert in this area too 😅😂  but in general also note browsers don't comply with the spec completely; Failing gracefully is as is the case here is often the better option for the end user. https://stackoverflow.com/questions/64821044/how-browser-restructure-the-invalid-html-structure-is-there-any-rules-or-approa

ang-zeyu: SSR hydration assertions also do not check everything, likely just some of the most common ones like `<table>`s that browsers autocorrect aggressively

weiquu: Thanks for the links! Have edited according to @tlylt's suggestion

tlylt (approved):
Thanks @weiquu, LGTM

PR #1976, opened 2022-07-28
Update seamless panels to inherit background color from parent

### LINKED ISSUE #1697: Seamless panels should inherit all colours

Note the seamless panel standing out in a not-so-good way:

seamless

This seems like an intentional design decision, as far back as 2019 at least:

https://github.com/MarkBind/vue-strap/blob/c61d132f8bf7e97e7fe7db3452ddb7dfcdc2ced4/src/Panel.vue#L189

The solution below may break some existing edge case usages that depend on this behaviour as such
(e.g. if placing a seamless panel with some `color: black` content, inside another container with `background-color: black`)

As "seamless" panels (barring the minimized form) should be, I think:
- the background should be transparent
- `color`s of parents should be inherited
  - e.g., if its placed in a container with `color: white`, it should also automatically use white colours

seamless2

eeliana:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [x] Others, please explain:
ReaderUsability

Resolves #1697 

**Overview of changes:**
Updated CSS for child components to inherit background color of parent component.

_Before_
image

_After_
image

nil

**Testing instructions:**
Check the User Guide --> Authoring Contents --> Presentation --> Panels for how the different panels would look like.
Confirm that the seamless panels inherit the colour of their parent container.
Check Home --> More control to the reader, without duplicating code. The nested panel there should inherit the box colour.

Update CSS for seamless panels to inherit parent's background color

---

- [x] No unrelated changes


eeliana:
> Thanks for the PR addressing this issue! It seems that the documentation does not have an example of this use case, perhaps it would be good to include it, that way, users would know that this is a possibility.
> 
> Also, is there a reason you choose to use `!important` in the css?

In response to the last question, I used it because there might be styles from libraries which I can't seem to access

tlylt:
> > Also, is there a reason you choose to use `!important` in the css?
> 
> In response to the last question, I used it because there might be styles from libraries which I can't seem to access

May I know what styles/libraries are you referring to and if you have any concrete examples of these cases that required the use of `!important`?

tlylt (changes requested):
# Before
https://markbind.org/userGuide/components/presentation.html#panels
image
# After
image

Something seems off ❓

eeliana:
Updated CSS styling to account for non-seamless nested panels

image

I tested with the just the `<box>... </box>` without specifying the class (see Home for docs) with and without `!important`. 
Without !important, the header does not inherit the box colour (which is grey by default). However, if the class for the `<box>` tag is specified, the inherit works with/without !important.

ang-zeyu:
(separate issue)

image

Should we update the bottom bar and caret's colours as well? (and the minimize button on the top right - not shown here)
(I've added a `text-white` to your example in this screenshot)

Take note of the button hover colours too (I think we can remove the `btn-secondary-outline:hover` background / anything else to make it more independent of the parent's background)

ang-zeyu:
Closing as stale, but we'd very much like to have this still.

PR #1988, opened 2022-08-03
Support a simple font colouring markdown syntax

### LINKED ISSUE #1475: Support a simple font coloring markdown syntax?

Inspiration from the `lineSpec` parameter matlab's plot function https://www.mathworks.com/help/matlab/ref/plot.html

I personally found being able to edit various plot properties with a single character really convenient.

We could support something along the lines of:

```
#r#...some text...## // red color
#g#...some text...## // green color
...
```

Should save a lot of typing `<span style="font-color:red">...</span>` or `<span class="your-red-class / bootstrap text classes">...</span>` too.

Concerns:
- not very semantic=P
- ❔ 

**Additional context**

weiquu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Resolves #1475 

**Overview of changes:**
Add support for a simple font colouring inline markdown syntax. The colours supported are red, green, blue, cyan, magenta, yellow, black, and white. Using red as an example, syntax is: #r#texttext## (with a space before #r# and space after ##).

I've made the changes as a preprocessor method before the content is passed to the markdown processor. Essentially it checks for the starting and ending syntax, and replaces them with the appropriate span once found. The preprocessed content is then returned.

**Testing instructions:**
I've done up 4 tests that covers normal use, escaped syntax, syntax without matching start/end, and syntax within a code block. The code for the tests is:
````
#### Test 1
#r# Red text #b# Blue text within red text ## Escaped ending: \## Real end. ##

#### Test 2
\#g# <- Escaped green starting. #g# Green text ## End without close: ## and Start without close: #g# green text

new line, not green text

#### Test 3
#c#cyantext##

#### Test 4
```
code block #y#yellowtext##
```
````
and the result is:
tests

Support a simple font colouring markdown syntax

---

- [ ] No unrelated changes

### CODE REVIEW

**Comment by damithc** [comment]
@weiquu Currently, Bootstrap styles can be used to get a similar effect, like so:
```
# My Heading{.text-danger}
```

**Comment by tlylt** [comment]
Thanks, @weiquu! 
The original issue seems to want a general text coloring utility -> both heading and normal paragraph of text, any specific reason why the proposed change only addresses the latter?

**Comment by weiquu** [comment]
Hi @tlylt and @damithc, I originally did just the header colourings as a test, but later on forgot about expanding it to normal text. Given that Bootstrap styles can be used to add colour quite easily, is there still a need to address the issue, or should it be closed?

**Comment by tlylt** [comment]
> Hi @tlylt and @damithc, I originally did just the header colourings as a test, but later on forgot about expanding it to normal text. Given that Bootstrap styles can be used to add colour quite easily, is there still a need to address the issue, or should it be closed?

I think there's some merit in that a simpler syntax makes it easy to set fonts for text, especially for text within a paragraph, in that case then I think doing something like `My Heading{.text-danger}` won't work. Pinging @ang-zeyu for further advice on this.

File: packages/core/src/html/NodeProcessor.js [hunk 1]
```diff
@@ -315,6 +315,55 @@ class NodeProcessor {
...
+          // Now that we've got the lines which match the syntax, we just
+          // replace the last character with <span> ... </span> and recreate the line
+          let span = '';
+          if (lastChar === 'r') {
+            span = '<span style="color: red;">';
+          } else if (lastChar === 'g') {
+            span = '<span style="color: green;">';
+          } else if (lastChar === 'b') {
+            span = '<span style="color: blue;">';
+          }
```
ong6, line 339: Could we abstract this part out and instead use a map to represent this logic? It would make it easier to add new colors in the future and be more versatile overall.

ang-zeyu:
> > Hi @tlylt and @damithc, I originally did just the header colourings as a test, but later on forgot about expanding it to normal text. Given that Bootstrap styles can be used to add colour quite easily, is there still a need to address the issue, or should it be closed?
> 
> I think there's some merit in that a simpler syntax makes it easy to set fonts for text, especially for text within a paragraph, in that case then I think doing something like `My Heading{.text-danger}` won't work. Pinging @ang-zeyu for further advice on this.

Yes this is the use case.

markdown-it-attrs dosen't allow things like that.

To add, I'm thinking the above example would become (support only the _inline_ syntax):

```
# #r#Using Plugins##
```

This goes with how `**...**` `*...*` ... works in markdown.

I didn't spend too much time thinking about the syntax as well (`#r/g/b#...##` ), its possible there's something even more usable. Feel free to suggest some variants. 🙂 

Nor the colours -- if we want we can even go with bootstrap's ones `i(nfo)/d(anger)/...`.

weiquu:
Thanks all for the feedback, I've updated the implementation to an inline one with the syntax suggested by @ang-zeyu (handling of syntax not being properly closed and escaping of the syntax are detailed in the code). I'll add in support for more colours in a bit and update the initial comment, but in the meantime here's a few tests that I ran:

colour syntax test

with code as follows:
```
#### Test 1
#r# Red text #b# Blue text within red text ## Escaped ending: \## Real end. ##

#### Test 2
\#g# <- Escaped green starting. #g# Green text ## End without close: ## and Start without close: #g#
```

ang-zeyu:
> Not sure if it is worth the effort and additional verbosity though

Personally it seems a little too verbose, close to the point of just writing `<span style="...">...</span>` which is more flexible.

> Thanks all for the feedback, I've updated the implementation to an inline one with the syntax suggested by @ang-zeyu (handling of syntax not being properly closed and escaping of the syntax are detailed in the code). I'll add in support for more colours in a bit and update the initial comment, but in the meantime here's a few tests that I ran:

I think the colours are good for now. We can test out the utility of this feature bit by bit. If its worth it let's add more down the line. @damithc @tlylt

damithc:
> I think the colours are good for now. We can test out the utility of this feature bit by bit. If its worth it let's add more down the line. @damithc @tlylt

👍

tlylt:
> Personally it seems a little too verbose, close to the point of just writing `<span style="...">...</span>` which is more flexible.

Agree with this. Also, I think perhaps we can work towards a generic & simpler syntax to add class to inline text in future PRs. So that we don't have to do `<span class="a">xxxx</span>` and can just do `#a# xxxx #a#`.

> > I'll add in support for more colours in a bit and update the initial comment

I think we can either go with covering the [bootstrap theme](https://getbootstrap.com/docs/5.0/customize/color/) colors, 
image

or the [colors from the original issue referencing MATLAB](https://www.mathworks.com/help/matlab/ref/plot.html#:~:text=Hexagram-,Color%20Name,-Short%20Name)
image

So that we have a minimally complete set of colors for this PR.

With that, we will also need to
- add documentation in UG
- add a functional test to cover the usage

packages/core/src/html/NodeProcessor.js
```diff
@@ -315,6 +321,60 @@ class NodeProcessor {
...
   }
 
+  /*
+   * Syntax is (space)#char#(any text)##(space)
+   * If we have #char# without ## or ## without #char# we don't replace
+   * To escape start do \#char#, so it will show up as #char# (escaping also works with any front char)
+   * To escape end do \##, which will show up as ##
+   */
+  preprocessMarkdown(content) {
+    const lines = content.split('\n');
```
tlylt, line 331: Perhaps we can early return if the content is empty

packages/core/src/html/NodeProcessor.js
```diff
@@ -315,6 +321,60 @@ class NodeProcessor {
...
+    let amendedContent = '';
```
tlylt, line 332: "Amend" feels like it's saying the original content is "incorrect", perhaps we can use a more neutral name here, like "result", or "preprocessedContent"? Ok to leave it as is too.

packages/core/src/html/NodeProcessor.js
```diff
@@ -315,6 +321,60 @@ class NodeProcessor {
...
+    for (let i = 0; i < lines.length; i += 1) {
+      const lineArray = lines[i].split(' ');
+      const stack = [];
+      for (let j = 0; j < lineArray.length; j += 1) {
+        // Check for #char#
+        const syntaxStartChecker = lineArray[j].slice(0, 3);
```
tlylt, line 338: There's a magic number here, could be difficult for future extension

tlylt (commented):
Sorry, I think we forgot to discuss the syntax. Midway through my review, it seems like the handling is that there is a space between the identifiers, so `#r# some text ##`. Apart from whether it is CommonMark compliant, I think it is slightly different from how typical markdown inline syntax works, for example `**some text**` (notice there is no space between them for it to work, `** some text**` fails). 

Could we perhaps go for the same approach here, something like `#r#some text#r#`. @ang-zeyu @damithc feel free to chime in as well on the details of the syntax.

(Lastly, forgot to mention that @weiquu I think it will be good to add some **unit** test for this feature, to cover the various type of inline situations. You can see examples in `core/test/unit/html/NodeProcessor.test.js`)

ang-zeyu:
> or the [colors from the original issue referencing MATLAB](https://www.mathworks.com/help/matlab/ref/plot.html#:~:text=Hexagram-,Color%20Name,-Short%20Name)

I prefer MATLAB. (with moving away from bootstrap (#903) in mind)

ang-zeyu:
> Could we perhaps go for the same approach here, something like #r#some text#r#. @ang-zeyu @damithc feel free to chime in as well on the details of the syntax.

If possible, I personally prefer to loose the `r` in the ending `#r#`.

packages/core/src/html/NodeProcessor.js
weiquu: Have pulled it out to the front and left comments explaining!

weiquu:
Okay have added tests and updated the documentation! Lmk if I missed anything out (:

> Sorry, I think we forgot to discuss the syntax. Midway through my review, it seems like the handling is that there is a space between the identifiers, so `#r# some text ##`. Apart from whether it is [CommonMark compliant](https://spec.commonmark.org/0.30/#inlines), I think it is slightly different from how typical markdown inline syntax works, for example `**some text**` (notice there is no space between them for it to work, `** some text**` fails).

Hmm `#r#some text##` works for me though (see test 3 in the [initial PR comment](https://github.com/MarkBind/markbind/pull/1988#issue-1326848503)), the spacing mentioned in the code comments is for before the start syntax and after the ending syntax

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,10 +26,14 @@ Syntax added by MarkBind:
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">markdown</variable>
 <variable name="code">
-****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~
+****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~, #r#Coloured Text##
```
tlylt, line 29: Quick review: doesn't seem to work here
image

tlylt: Also, quick check, does it work in a code block? e.g. `some code then #r#text## `(not supposed to)

weiquu: > Also, quick check, does it work in a code block? e.g. `some code then #r#text## `(not supposed to)

Ah it still changes to the span, let me go and fix it!

tlylt:
> I've made the changes as a preprocessor method before the content is passed to the markdown processor. Essentially it checks for the starting and ending syntax, and replaces them with the appropriate span once found. The preprocessed content is then returned.

@weiquu On second thought, perhaps you may want to look into markdown-it plugins and implement this feature similar to this one here
-  (center div) packages\core\src\lib\markdown-it\plugins\markdown-it-center-text.js

Or see how the other text style additions that we did before, via double delimiter:
- packages\core\src\lib\markdown-it\index.js
- packages\core\src\lib\markdown-it\plugins\markdown-it-double-delimiter.js
image

This could help avoid all the parsing and weird edge cases.

(P.S really sorry to raise this later, great work so far!)

docs/userGuide/syntax/textStyles.md
damithc: I think coloring should be explained as a separate feature, as the syntax is different, and not part of markdown.  `Tip` boxes are not suitable for giving regular instructions anyway (they are meant to be to be optional info).

packages/core/src/html/NodeProcessor.js
```diff
@@ -315,6 +326,68 @@ class NodeProcessor {
     return node;
   }
 
+  /*
+   * Syntax is (space)#char#(any text)##(space)
+   * If we have #char# without ## or ## without #char# we don't replace
+   * To escape start do \#char#, so it will show up as #char# (escaping also works with any front char)
+   * To escape end do \##, which will show up as ##
+   */
+  preprocessMarkdown(content) {
```
ang-zeyu, line 335: @weiquu Sorry I didn't manage to take a look at the implementation earlier. I assumed this was implemented similarly to one of our current double-delimiter plugins.

I don't think this approach is quite right, writing a markdown parser / processor isn't trivial. We should be tapping directly into `markdownIt`'s parsing / [plugin](https://github.com/markdown-it/markdown-it/tree/master/docs) system.

There are far too many edge cases from trying to parse / handle the file from scratch (other than just the few pointed out above...)

e.g. comments, for one

```
/*
#r#...##
*/
```

ang-zeyu: Tldr basically the same as @tlylt's [comment ](https://github.com/MarkBind/markbind/pull/1988#issuecomment-1209207496)above, I haven't read that when I posted this 😶

weiquu:
Hey sorry I realised I forgot to leave a comment here! I've changed the implementation to use the markdown-it plugin, similar to the center-text one.

> I think coloring should be explained as a separate feature, as the syntax is different, and not part of markdown.

Regarding this, should I put in a new section in Formatting Contents between Text Styles and Blockquotes?

packages/core/src/lib/markdown-it/plugins/markdown-it-colour-text.js
```diff
@@ -0,0 +1,151 @@
...
+ EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// Process #r# red text ##
```
tlylt, line 26: Is this the correct description? I am not sure if you are attaching this because you modify the code from the original author or?

packages/core/test/unit/html/NodeProcessor.test.js
```diff
@@ -123,6 +123,43 @@ test('processNode processes dropdown with header slot taking priority over heade
                            testData.PROCESS_DROPDOWN_HEADER_SLOT_TAKES_PRIORITY_EXPECTED);
 });
 
+// test('preprocessor converts inline colour syntax to the appropriate span', async () => {
```
tlylt, line 126: Could we repurpose the test cases here into unit tests for the plugin that you wrote? Quite a waste to remove and would be great to have these behaviors checked for the plugin

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,10 +26,14 @@ Syntax added by MarkBind:
...
 ->Center-align<-
 </variable>
 </include>
+<box type="tip" seamless>
```
tlylt, line 33: As mentioned by a previous comment:
> I think coloring should be explained as a separate feature, as the syntax is different, and not part of markdown. Tip boxes are not suitable for giving regular instructions anyway (they are meant to be to be optional info).

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,10 +26,14 @@ Syntax added by MarkBind:
...
+Colours supported are red (r), green (g), blue (b), cyan (c), magenta (m), yellow (y), black (k), and white (w).
+To escape the syntax, simply put a backslash in front of it (e.g. \#b#, \##).
```
tlylt, line 35:
```suggestion
To escape the syntax, simply put a backslash in front of it (e.g. `\#b#, \##`).
```

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,10 +26,14 @@ Syntax added by MarkBind:
 <include src="codeAndOutput.md" boilerplate >
 <variable name="highlightStyle">markdown</variable>
 <variable name="code">
-****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~
+****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~, #r#Coloured Text##
 ->Center-align<-
 </variable>
 </include>
+<box type="tip" seamless>
+Colours supported are red (r), green (g), blue (b), cyan (c), magenta (m), yellow (y), black (k), and white (w).
```
tlylt, line 34: Suggest to use a list to list out the options instead

docs/userGuide/syntax/textStyles.md
tlylt, line 29: On the website this line is getting too long and being hidden for smaller screen, could we split the lines into two, so perhaps
```
****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--

Super^script^, Sub~script~, #r#Coloured Text##

```

weiquu:
> Hi @weiquu sorry is this PR ready for review? I am seeing quite a few blocks of commented-out code in the changed files.

Ah I was referencing it while changing the implementation, will remove it in a moment.

Regarding the comments on documentation, I think @damithc previously mentioned to explain it as a separate feature. In this case, should I put in a new section in Formatting Contents between Text Styles and Blockquotes? Or should I just explain it separately under the Text Styles section?

packages/core/src/lib/markdown-it/plugins/markdown-it-colour-text.js
weiquu: Hmm I adapted the implementation from the center-text one which had this description at the top, hence I left it in

packages/core/test/unit/html/NodeProcessor.test.js
weiquu: Okay repurposed!

ang-zeyu:
> Regarding the comments on documentation, I think @damithc previously mentioned to explain it as a separate feature. In this case, should I put in a new section in Formatting Contents between Text Styles and Blockquotes? Or should I just explain it separately under the Text Styles section?

Separate section under `Text Styles` would do as its still altering text styles :)
I think @damithc's point was that the syntax is quite different from the other ones in that section.


(Identical delimiters on left and right)
```
****Super Bold****, !!Underline!!, ==Highlight==, %%Dim%%, ++Large++, --Small--, Super^script^, Sub~script~
```

damithc:
> Separate section under `Text Styles` would do as its still altering text styles :)
> I think @damithc's point was that the syntax is quite different from the other ones in that section.

Yup. But I'm OK to keep the current way too. It's a minor cosmetic issue.

weiquu:
Sorry I missed this! Have updated the documentation

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,12 +26,43 @@ Syntax added by MarkBind:
...
+To escape the syntax, simply put a backslash in front of it (e.g. \#b#, \##).
+</box>
```
tlylt, line 36: @weiquu with the content below, should we consolidate the repeated content here down as well? What do you think?
(ps also some review comments given above in case you missed it)

weiquu:
Okay made some changes! Is the table for the letters to colours alright?

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,12 +26,40 @@ Syntax added by MarkBind:
...
+g | green
+b | blue
+c | cyan
+m | magenta
+y | yellow
+k | black
+w | white
+
+<box type="tip" seamless>
+To escape the syntax, simply put a backslash in front of it (e.g. `\#b#`, `\##`).
```
tlylt, line 58: Likely need to leave a blank line as the markdown is not rendered properly
image

docs/userGuide/syntax/textStyles.md
```diff
@@ -26,12 +26,40 @@ Syntax added by MarkBind:
...
+</box>
+</div>
+
```
tlylt, line 61: Is the blank line here on purpose?

tlylt (changes requested):
Some more nits on the docs, rest LGTM

weiquu:
Okay made the changes!

tlylt (approved):
LGTM

packages/core-web/src/styles/markbind.css
```diff
@@ -175,6 +175,38 @@ footer {
     text-decoration: underline;
 }
 
+.red {
+    color: red;
```
ang-zeyu, line 179: this is a good case for using `!important`.

we can make the class name something more specific / conflict free too, e.g. follow bootstrap's `text-xx` but prefix something more. 👀

ang-zeyu: Made these tiny changes and merged. Nice work here! @weiquu

PR #1993, opened 2022-08-07
Handle generated plantuml files in functional testcases

### LINKED ISSUE #1130: PlantUML diagrams always "update" on updating test sites

### LINKED ISSUE #1937: Include caveats for updating functional testcases in DG

tlylt:

- [x] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1937
Fixes #1130 (arguably, this is an "expedient" solution such that if we include more/adjust plantuml files in our test sites in the future, we will need to update the handling code accordingly to ignore them. Let me know if I should change the keyword here to leave that issue open)

Also:
- update a warning in our DG -> workflows to mention the caveat, just in case
- add a note about how to update plantuml files that are ignored

~~Besides deleting the plantuml images and git ignore them, I realized I will also need to delete them from the `expected` folder again, before running the functional test. This is because these files which will still be generated after an `npm run updatetest` will affect testing.~~
The plantuml images will still be generated in the `expected` folder, in case devs want to look at them.

Wonder if we should do the same thing for the font files, as they appear to be problematic as well.
- at the moment: no

**Testing instructions:**
For docs update:
https://deploy-preview-1993--markbind-master.netlify.app/devguide/workflow#updating-and-writing-tests
image
image

For testing code:
- `npm run test` should succeed
- Running `npm run updatetest` should no longer show plantuml files as uncommitted changes, but those files should actually be generated in the `expected` folder  (because they are added to .gitignore)
- Re running `npm run test` should pass again, and those generated plantuml files should still appear in the `expected` folder

Handle generated plantuml files in test sites

Unrelated changes to png files (generated by plantuml) appear
when updating functional test cases.

Let's mention the need to ignore .png files (and binary files in general)
in our developer guide. Let's also update functional tests such that
some files can be ignored, to reduce the need to deal with 
unrelated changes.

This helps newcomers to know how to update functional tests
properly and reduce unnecessary dev work to maintain the test cases.

---

- [x] No unrelated changes


packages/cli/test/functional/testUtil/compare.js
```diff
@@ -24,26 +25,43 @@ function _readFileSync(...paths) {
...
+
+  // Check for file existence of ignoredPaths and that they are present in actualPaths
+  if (ignoredPaths.length !== 0 && !_.isEqual(_.intersection(ignoredPaths, actualPaths), ignoredPaths)) {
+    throw new Error('Ignored paths are not present in actual paths!');
+  }
 
   let error = false;
-  if (expectedPaths.length !== actualPaths.length) {
-    throw new Error('Unequal number of files');
+  if ((expectedPaths.length + ignoredPaths.length) !== actualPaths.length) {
```
jonahtanjz, line 49: > Besides deleting the plantuml images and git ignore them, I realized I will also need to delete them from the expected folder again, before running the functional test. This is because these files which will still be generated after an npm run updatetest will affect testing.

What do you think about filtering the `expectedPaths` first before comparing so that we do not need to delete the generated puml images? From the perspective of the developers, it feels slightly off to have the images generated after `updatetest` and then only deleted after running `npm run test`. 

Would also be nice to keep the images in the expected folder in case the developers want to have a look at them (they don't have to run `updatetest` to regenerate).

packages/cli/test/functional/testSites.js
```diff
@@ -14,8 +14,27 @@ const testTemplateSites = [
...
+  '9c9e77fc0a983cb6b592e65733787bec.png',
+  'inline-output.png',
+  'diagrams/activity.png',
+  'diagrams/class.png',
+  'diagrams/component.png',
+  'diagrams/object.png',
+  'diagrams/sequence.png',
+  'diagrams/state.png',
+  'diagrams/usecase.png',
+  'sub_site/inline-output/inline-puml-image.png',
```
jonahtanjz, line 32: Might also want to add in the dev docs the workflow of adding/removing puml files to the test sites (updating this array and `.gitignore` etc).

jonahtanjz (commented):
> Wonder if we should do the same thing for the font files, as they appear to be problematic as well.

Some of the fonts need to be updated together with the dependency (Bumping font-awesome etc). Not sure about the rest but the fonts on my end don't really change during `updatetest` (Happened only a few times as far as I can remember). If this is a rare issue, we can do the manual reminder method as it seems more complicated to maintain a long list of ignored fonts.

packages/cli/test/functional/testUtil/compare.js
tlylt: Hmm, thanks for the feedback. I'm ok with the filtering of expectedPaths. When doing so, I realized that there's a mismatch between CI and our local environment. The CI environment will never contain the generated images, but the local environment will first contain none of the images, and then have the generated images after one `npm run updatetest`. 

I guess it's ok but just want to point that out.

jonahtanjz: An alternative would be to remove the generated png files in the `updatetest` step. This would ensure consistency between CI and local environment but will make it very difficult for devs to manually check the generated images for correctness. What do you think?

tlylt: Yup that works too, I think weighing the cost, we can go with the previous suggestion so that there's a probability of checking the generated files if needed

jonahtanjz (approved):
Nice job @tlylt!

LGTM 👍

Issue #2002, opened 2022-08-11
Search bar overflows on mobile devices


jonahtanjz:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Tell us about your environment

Windows 10

### MarkBind version

4.0.0

### What did you do? Describe the bug

Go to https://markbind-master.netlify.app/index.html on mobile devices, the search bar exceeds the screen width.

![image](https://user-images.githubusercontent.com/47470981/184059867-80c0893a-c6b7-45e6-a77d-60609fbfe21f.png)

Let's try to resolve this before releasing 4.0.1.

### Steps to reproduce

Most likely due to setting a [`min-width` in `Searchbar.vue`](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/Searchbar.vue#L266)

### Expected behavior

Search bar should not overflow

### Actual behavior

Search bar exceeds the screen width as shown above.

domlimm:
Hello, may I please pick this issue up?

jonahtanjz:
@domlimm Sure. Thanks for taking this :)

domlimm:
Hey @jonahtanjz, is https://markbind-master.netlify.app/index.html the deployed `master` branch? I think the prod one is https://markbind.org/? Thank you!

PR #2004, opened 2022-08-11
Fix FOUC in navbar

### LINKED ISSUE #2003: FOUC in navbar due to search bar

### Any related issues?

_No response_

Windows 10

### MarkBind version

4.0.0

When navigating between pages, you can see that the navbar height expands first and then revert back to original size.

https://user-images.githubusercontent.com/47470981/184063267-20ea4c6b-7bd1-42eb-b392-62361b6e7d34.mp4

### Steps to reproduce

1. Go to https://markbind-master.netlify.app/userguide/gettingstarted
2. Navigate between the pages (especially between developer guide and user guide)
3. See the FOUC in the navbar

### Expected behavior

Navbar height remains the same

### Actual behavior

Navbar height changes during loading

### Anything else?

Most likely related to https://github.com/MarkBind/markbind/pull/1959#pullrequestreview-1041138318

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #2003 

The FOUC is caused by the `placeholderDiv` in `Searchbar.vue`. When it is set to `visibility: hidden`, the placeholder div will still take up space and hence increases the height of the navbar during loading. Setting it to `display: none` will ensure that the div does not occupy space which resolves the FOUC.

NIL

**Testing instructions:**
1. Go to https://deploy-preview-2004--markbind-master.netlify.app/index.html
2. Navigate between the pages
3. FOUC should be resolved

Fix FOUC in navbar and search bar

The placeholderDiv in Searchbar.vue is hidden using the 
visibility property. The placeholderDiv will still take up space
on the navbar which will increase its height during loading.
Furthermore, the width of the search bar is set dynamically
during page load which results in FOUC in the search bar as
well.

Let's change the approach such that the placeholderDiv is hidden
and does not take up any height to prevent FOUC in the navbar.
The approach of setting the search bar width should also be 
changed such that it is not set during page load to prevent 
FOUC in the search bar.

---

- [x] No unrelated changes

### DIFF
--- a/packages/vue-components/src/Searchbar.vue
+++ b/packages/vue-components/src/Searchbar.vue
@@ -1,24 +1,25 @@
 <template>
   <div style="position: relative;" class="dropdown">
     <div v-if="algolia" id="algolia-search-input"></div>
-    <input
-      v-else
-      v-model="value"
-      data-bs-toggle="dropdown"
-      type="text"
-      class="form-control"
-      :placeholder="placeholder"
-      autocomplete="off"
-      @input="update"
-      @keyup.up="up"
-      @keyup.down="down"
-      @keydown.enter="hit"
-      @keydown.esc="reset"
-      @blur="showDropdown = false"
-    />
-    <div class="form-control placeholderDiv-hidden">
-      {{ placeholder }}
-    </div>
+    <template v-else>
+      <input
+        v-model="value"
+        data-bs-toggle="dropdown"
+        type="text"
+        class="form-control"
+        :placeholder="placeholder"
+        autocomplete="off"
+        @input="update"
+        @keyup.up="up"
+        @keyup.down="down"
+        @keydown.enter="hit"
+        @keydown.esc="reset"
+        @blur="showDropdown = false"
+      />
+      <div class="form-control placeholderDiv-hidden">
+        {{ placeholder }}
+      </div>
+    </template>
     <ul ref="dropdown" :class="dropdownMenuClasses">
       <li
         v-for="(item, index) in items"
--- a/packages/vue-components/src/Searchbar.vue
+++ b/packages/vue-components/src/Searchbar.vue
@@ -16,7 +16,7 @@
         @keydown.esc="reset"
         @blur="showDropdown = false"
       />
-      <div class="form-control placeholderDiv-hidden">
+      <div class="form-control placeholder-div-hidden">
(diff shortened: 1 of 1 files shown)


tlylt:
Hi @jonahtanjz, quick question: does this affect the ability of the previous PR by @weiquu to retrieve the width of the div to help determine the overall width of the search bar?

jonahtanjz:
> Hi @jonahtanjz, quick question: does this affect the ability of the previous PR by @weiquu to retrieve the width of the div to help determine the overall width of the search bar?

Thanks for the catch @tlylt!

Missed out on the part where the placeholder div is used to determine the overall width.

I also noticed that apart from the navbar FOUC, the search bar itself also has FOUC during loading. Notice how the search bar expands after page is loaded.

https://user-images.githubusercontent.com/47470981/184140410-47f561fd-51c8-4096-a71d-f9ad5317baaa.mp4

This is due to the width changing while the page has already loaded.

I've went with another approach to prevent the search bar input width from changing. Instead of setting the width of the search bar to the placeholder div's width when it is mounted, I let the parent div's (`div.dropdown`) width be determined by the placeholder div. Since the search bar is set to `width: 100%`, it should take the parent div's width as well. The placeholder div is then hidden from view using some css which will prevent it from taking up space and removing the FOUC issue altogether.

tlylt (approved):
Thanks @jonahtanjz, LGTM. The overflow issue is not resolved yet, but the FOUC seems to be gone now.

jonahtanjz:
> The overflow issue is not resolved yet

Yup that will be done in another PR (Issue #2002)

packages/vue-components/src/Searchbar.vue
```diff
@@ -276,7 +274,13 @@ export default {
         left: auto;
     }
 
-    .hidden {
+    .placeholderDiv-hidden {
```
ang-zeyu, line 277:
```suggestion
    .placeholder-div-hidden {
```

I don't have much context here, just dropping a not-so-helpful minor nit 🙈

jonahtanjz: Thanks! Updated :) Also added `overflow: hidden` to prevent the placeholder div from overflowing the navbar.

PR #2012, opened 2022-08-17
Remove modals with duplicate id

### LINKED ISSUE #1909: Modal not working correctly in UG Page Navigation Menus section

### Any related issues?

_No response_

Windows 10

### MarkBind version

3.1.1

Click on the modal trigger in https://markbind-master.netlify.app/userguide/syntaxcheatsheet page-navigation-menus, two modals opened for that.

### Steps to reproduce

- Enter https://markbind-master.netlify.app/userguide/syntaxcheatsheet
- Go to Page Navigation Menus
- Click on the `Example usage of the <page-nav /> component`

### Expected behavior

Expected to behave normally like in https://markbind.org/userGuide/syntaxCheatSheet.html

### Actual behavior

Open two models

### Anything else?

Other pages seem to work fine. Wonder if it's due to the Bootstrap update, @jovyntls any idea?

### LINKED ISSUE #2008: The page becomes unscrollable after opening and closing modal

### Any related issues?

_No response_

n/a

### MarkBind version

4.0.0

See https://github.com/nus-cs2103-AY2223S1/forum/issues/1

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Anything else?

_No response_

jonahtanjz:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #2008 
Fixes #1909 

The issue is caused by having modals of the same id on the same page (usually because the modals are included more than once using `<include>`). Removing modals with ids that already exist fixes the issue. The issue with duplicated modals opening is also resolved.

Previous implementation allows for modals with the same id to be included/present on the page more than once. Having just 1 copy of the modal still allows the included `<trigger>`s to open the modal. Not sure if I missed out on any edge cases where more than 1 modal of the same id needs to be on the page.

**Testing instructions:**
1. Create multiple `<modal>`s and `<trigger>`s with the same id
2. Serve the site
3. All triggers should still work but only 1 modal shows up

Remove modals with duplicate id

Modals with duplicate id on the same page will result in
triggers opening multiple modals and can cause the page
to be unscrollable after the modal is closed.

Let's remove modals with duplicate id so that only one
copy of the modal for each id will exist on the page.

---

- [x] No unrelated changes


packages/core/src/html/NodeProcessor.js
```diff
@@ -166,11 +181,13 @@ class NodeProcessor {
         this.mdAttributeRenderer.processTooltip(node);
         break;
       case 'modal':
-        // Transform deprecated slot names; remove when deprecating
-        renameSlot(node, 'modal-header', 'header');
-        renameSlot(node, 'modal-footer', 'footer');
+        if (!this._removeDuplicateModals(node)) {
```
tlylt, line 184: Should we explore extracting this logic (including line 186-189) into a process modal function instead? So
```js
    this._processModal(node)
   break;
```
and then something like:
```js
  _processModal(node) {
    if (this.processedModals[node.attribs.id]) {
       cheerio(node).remove();
    } else {
    this.processedModals[node.attribs.id] = true;
    // Transform deprecated slot names; remove when deprecating
          renameSlot(node, 'modal-header', 'header');
          renameSlot(node, 'modal-footer', 'footer');
          this.mdAttributeRenderer.processModalAttributes(node);
    }
  }
```

packages/core/src/html/NodeProcessor.js
```diff
@@ -57,6 +57,8 @@ class NodeProcessor {
     this.footnoteProcessor = new FootnoteProcessor();
     this.mdAttributeRenderer = new MdAttributeRenderer(this.markdownProcessor);
     this.pageNavProcessor = new PageNavProcessor();
+
+    this.processedModals = {};
```
tlylt, line 61: Double checking whether this will be handled correctly during live reloading?

tlylt (commented):
Thanks @jonahtanjz for the fix, since it also resolves the modal issue #1909  on our UG, perhaps we can link it in the PR description to close it as well.

I was thinking about this problem for a bit, and I wonder whether there are other cases of this problem due to how include is implemented, which causes the id collision issue. For example, anchor scroll handling and scroll spy for pageNav. Will try to look into these in the coming weeks...

packages/core/src/html/NodeProcessor.js
jonahtanjz: Thanks @tlylt. Sounds good 👍

packages/core/src/html/NodeProcessor.js
jonahtanjz: During live reload, a new instance of `NodeProcessor` should be created hence this should be updated accordingly for each of the regenerated pages.

tlylt (approved):
LGTM on my end

damithc:
@jonahtanjz Thanks for fixing this. Perhaps do a release, so that I can start using the bug fix in production sites?

jonahtanjz:
> Thanks for fixing this. Perhaps do a release, so that I can start using the bug fix in production sites?

Just did a release :)

Issue #1913, opened 2022-04-20
Fully adapt core package and associated tests to TypeScript


ryoarmanda:
**(Updated as of 21/04/2022)**

This is an epic-type issue to track of our TypeScript migration of the `core` package. Feel free to edit this as we go along.

Interested migration developers can get started by checking out the [Migrating to TypeScript](https://markbind.org/devdocs/devGuide/migratingToTypeScript.html) page in our developer's guide.

## Dependency graph of remaining `js` files

(There are some files not presented that are still under consideration on whether to be adapted to TypeScript or leave as-is)

Click the image to view it all full resolution:

![graph](https://user-images.githubusercontent.com/24468110/164284427-54953286-eede-4c62-9a02-31c25508e1fc.png)

The green files are ones that do not have any _static_ dependencies to another file and are **highly recommended** to migrate first. However, the graph is not 100% accurate, as there are green files that have _dynamic_ imports (like in Plugin.js) or have constants documented to be typed as other files' classes (like PageConfig.js and SiteConfig.js)

## Which one should I adapt first?

As mentioned, it is **highly recommended** that we go through files that do not have any dependencies, and collectively work through the dependency graph upwards until the root, meaning that we perform the migration in topological order.

> Though possible to migrate not in topological order, the effort will be much more as you have to be able to infer simple types as a "stand-in" of the dependencies that are not yet migrated.

You can adapt multiple _related_ files residing in the same folder in one migration. For example, you might adapt Plugin.js and PluginManager.js together.

Note that when you adapt a file, you also have to adapt the associated tests as well. For example, adapt linkProcessor.js together with  linkProcessor.test.js.

## Files checklist

(There are some files not presented that are still under consideration on whether to be adapted to TypeScript or leave as-is)

### package root

- [x] index.js

### `src`

- [x] errors (#1877)
  - [x] CyclicReferenceError.js
  - [x] index.js
- [x] External
  - [x] External.js
  - [x] ExternalManager.js
- [x] html
  - [x] codeblockProcessor.js
  - [x] constants.js
  - [x] Context.js
  - [x] elements.js
  - [x] FootnoteProcessor.js
  - [x] headerProcessor.js
  - [x] includePanelProcessor.js
  - [x] linkProcessor.js
  - [x] MarkdownProcessor.js
  - [x] MdAttributeRenderer.js
  - [x] NodeProcessor.js
  - [x] scriptAndStyleTagProcessor.js
  - [x] siteAndPageNavProcessor.js
[...]

PR #2029, opened 2022-09-03
Migrate nodeprocessor and its minor dependencies to typescript

ang-zeyu:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Tackles part of #1913 

**Overview of changes:**
- Migrate the entrypoint to the `src/html` folder, `NodeProcessor`
- Some of its minor dependencies referenced (e.g. `Context` class)

This should enable migration of the rest of the `src/html` folder more smoothly/independently

See commit message

**Testing instructions:**
na

NA (in the commits themselves), let's do a rebase here to preserve the file rename

---

- [x] No unrelated changes


packages/core/src/html/NodeProcessor.ts
```diff
@@ -83,23 +99,24 @@ class NodeProcessor {
...
       // Retrieves the front matter from either the first frontmatter element
       // or from a frontmatter element that includes from another file
       // The latter case will result in the data being wrapped in a div
-      const frontMatterData = frontMatter.find('div').length
-        ? frontMatter.find('div')[0].children[0].data
-        : frontMatter[0].children[0].data;
+      const frontMatterIncludeDiv = frontMatter.find('div');
+      const frontMatterData = frontMatterIncludeDiv.length
+        ? ((frontMatterIncludeDiv[0] as DomElement).children as DomElement[])[0].data
+        : ((frontMatter[0] as DomElement).children as DomElement[])[0].data;
```
ang-zeyu, line 119: I feel like there should be better ways to configure the typing here, let me know if any ideas 🙂 

Or we can defer to another PR to investigate, since functionally it is the same.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -116,34 +133,41 @@ class NodeProcessor {
    * Layout element collection
    */
 
-  _collectLayoutEl(node, property) {
+  private static collectLayoutEl(node: DomElement): string | null {
     const $ = cheerio(node);
-    this[property].push($.html());
+    const html = $.html();
     $.remove();
+    return html;
```
ang-zeyu, line 140: had to change this part functionally since `this[property]` is not going to pass tsc.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -116,34 +133,41 @@ class NodeProcessor {
...
   /**
    * Removes the node if modal id already exists, processes node otherwise
    */
-  _processModal(node) {
-    if (this.processedModals[node.attribs.id]) {
-      cheerio(node).remove();
-    } else {
-      this.processedModals[node.attribs.id] = true;
+  private processModal(node: DomElement) {
+    if (node.attribs) {
```
ang-zeyu, line 147: only change here is this. This lets tsc know `node.attribs` (which is defined as `?: {}`) always exists

packages/core/src/html/NodeProcessor.ts
```diff
@@ -116,34 +133,41 @@ class NodeProcessor {
...
     }
   }
 
   /*
    * API
    */
-  processNode(node, context) {
+  processNode(node: DomElement, context: Context) {
     try {
+      if (!node.name || !node.attribs) {
```
ang-zeyu, line 167: same hint to tsc.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -291,12 +327,20 @@ class NodeProcessor {
...
+      ) as unknown as DomElement[];
       break;
     case 'markdown':
       node.name = 'div';
       node.children = cheerio.parseHTML(
-        this.markdownProcessor.renderMd(cheerio.html(node.children)), true);
+        // TODO figure this typing out
+        this.markdownProcessor.renderMd(cheerio.html(node.children as unknown as cheerio.Element)),
+        undefined, true,
+        // TODO figure this typing out
```
ang-zeyu, line 342: a more horrendous typing issue with cheerio/htmlparser2. Functionally the same but dosen't look pretty.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -308,30 +352,37 @@ class NodeProcessor {
...
 
     this.postProcessNode(node);
 
     addSitePageNavPortal(node);
 
-    const isHeadingTag = (/^h[1-6]$/).test(node.name);
-    if (isHeadingTag && !node.attribs.id) {
-      setHeadingId(node, this.config);
-    }
+    if (node.name) {
```
ang-zeyu, line 363: same sort of hint to tsc to let it know `node.name` always exists.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -308,30 +352,37 @@ class NodeProcessor {
...
+      const isHeadingTag = (/^h[1-6]$/).test(node.name);
+      if (isHeadingTag && !node.attribs?.id) {
+        setHeadingId(node, this.config);
+      }
 
-    // Generate dummy spans as anchor points for header[sticky]
-    if (isHeadingTag && node.attribs.id) {
-      cheerio(node).prepend(`<span id="${node.attribs.id}" class="anchor"></span>`);
+      // Generate dummy spans as anchor points for header[sticky]
+      if (isHeadingTag && node.attribs?.id) {
```
ang-zeyu, line 370: Added `?.` here (and line 365)

packages/core/src/html/NodeProcessor.ts
```diff
@@ -353,7 +404,9 @@ class NodeProcessor {
         });
         mainHtmlNodes.forEach(d => NodeProcessor._trimNodes(d));
 
-        const footnotesHtml = this.footnoteProcessor.combineFootnotes(node => this.processNode(node));
+        const footnotesHtml = this.footnoteProcessor.combineFootnotes((node: DomElement) => this.processNode(
+          node, new Context(cwf, [], extraVariables, {}),
```
ang-zeyu, line 408: Another functional change. But `context` is unused inside `processNode` here.

For standardising the typing of `processNode` here (2nd parameter is nullable). The other option is to make the 2nd parameter of `processNode(..., context?: Context)`, but that requires some hinting to tsc about whether `context` exists. This seemed like the cleaner option.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -353,7 +405,9 @@ class NodeProcessor {
         });
         mainHtmlNodes.forEach(d => NodeProcessor._trimNodes(d));
```
jonahtanjz, line 406: Looks like there may be some typing issue here? `mainHtmlNodes` here has the type `(DomElement | DomElement[])[]` which makes `d` have the type `DomElement | DomElement[]`. `_trimNodes` however only accepts a `DomElement` as param.

Also seems like `_trimNodes` method will have some issues if an array of `DomElement`s is passed to it.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -343,8 +395,8 @@ class NodeProcessor {
         const mainHtmlNodes = dom.map((d) => {
           let processed;
           try {
-            processed = this._process(d, context);
-          } catch (err) {
+            processed = this.traverse(d, context);
```
jonahtanjz, line 398: `this.traverse` can return a `DomElement | DomElement[]`

packages/core/src/html/NodeProcessor.ts
```diff
@@ -343,8 +395,8 @@ class NodeProcessor {
...
+          } catch (err: any) {
             err.message += `\nError while rendering '${file}'`;
             logger.error(err);
             processed = createErrorNode(d, err);
```
jonahtanjz, line 402: I think you can update this to `import { createErrorNode } from './elements'` since `element` has been converted to typescript. This should also let typescript check the typing of `mainHtmlNodes` which will cause the error.

packages/core/src/html/NodeProcessor.ts
```diff
@@ -268,9 +305,9 @@ class NodeProcessor {
     }
   }
 
-  _process(node, context) {
+  private traverse(node: DomElement, context: Context): DomElement | DomElement[] {
     if (_.isArray(node)) {
```
jonahtanjz, line 309: From what I can see, it seems like this check is not necessary since `node` will not be an array? If that is the case, we can remove this block and `traverse` should just return a `DomElement`.

jonahtanjz (commented):
Thanks @ang-zeyu 👍 

Just one concern, other than that the rest looks good :)

packages/core/src/html/NodeProcessor.ts
ang-zeyu: Nice catch, changed to just `DomElement`. Also resolves the other issue you pointed out with `_trimNodes`

ang-zeyu:
Changes are in the two fixups, have rebased on the latest master branch.

I also found a solution for the horrible typing issue [here](https://github.com/MarkBind/markbind/pull/2029#discussion_r962142464).
 Just left with this one https://github.com/MarkBind/markbind/pull/2029#discussion_r962142184 if you have any ideas. =C

packages/core/src/html/NodeProcessor.ts
jonahtanjz: I guess we can defer this to another PR to investigate. Can't think of a good idea right now...

jonahtanjz (approved):
LGTM 👍

ang-zeyu:
Hi @jonahtanjz! Didn't expect you to merge in so quickly. It's been very long since we had a pr requiring a[ fixup workflow](https://dev.to/koffeinfrei/the-git-fixup-workflow-386d), sorry for not reminding you of the intended process earlier.

It usually goes like this (the fixups should be squashed into the main commits once reviewed; only the main commits should be merged).
This PR isn't very big, but we're using a similar workflow for preserving renames.

```
1. Contributor sends up a rather big PR that should have a multi commit structure
  - If it is already structured, skip to step 4
2. Maintainer informs contributor its too big for one commit 
3. Contributor soft resets git branch, recommits and pushes with structure
4. The PR is reviewed
5. Contributor acts on the review by delivering "fix ups" (**incremental** commits that are reviewable, each one tied to a separate commit in the structure earlier) (https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash)
6. Reviewers review these fixups
  - All good -> step 9, otherwise step 7
7. Contributor does an interactive rebase / uses the git fixup feature to squash each fixup commit earlier with its respective commit, and adds **new** fixups to tackle step 6's issues.
8. Go back to step 6
9. Maintainer merges using a **merge commit**
```

The intention is to keep a bunch of well-organized, semantic commits throughout the review, and after merge, but also make incremental changes easy to review (via fixups).

ang-zeyu:
My opinion is that its fine to leave some derps in the commit history; But if you wish, you can also overwrite the relevant parts with an interactive rebase, since its fairly recent, and the changes once squashed are identical. (shouldn't be any issues with merging other prs on top)

Issue #2021, opened 2022-08-21
Consider running tests/lints before pre push


ang-zeyu:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### What is the area that this feature belongs to?

DevOps

### Is your feature request related to a problem? Please describe.

I'm always frustrated when I push my code up only to have it fail in CI 10mins later, due to forgetting to run tests / lints.

### Describe the solution you'd like

Consider running tests / lints before push.

A more controversial one would be pre commit. 

Notes:
- This can still be prevented using `git push --no-verify`.
- (maybe overkill for this project, pointing out this for ideas / research) https://pre-commit.com/

tlylt:
Related to #1966 on precommit.

EltonGohJH:
I will give this a try

tlylt:
> I will give this a try

Previously played around with it for some time and realized there are a few things to take note (some details are in https://github.com/MarkBind/markbind/issues/1966)

- for linting, it would be faster if we only lint files that have changed, which requires some configuration in order to make it work. As we have a few different eslint configs for different packages, there's a need to ensure that the current configs are still respected when linting via hooks
- for testing, there might be a problem when the backend is not built (due to our typescript setup), this means if the backend is not ready and we run the test via hooks, the test will fail unless we do a `npm run build:backend`. If somehow the backend is not automatically built, then running the tests before pre-push might fail unexpectedly.

tlylt:
Fixed by #2035

Issue #2020, opened 2022-08-21
Streamline the (typescript) development experience with some comprehensive commands


ang-zeyu:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

#1934, #1994, #1921

### What is the area that this feature belongs to?

DevOps

### Is your feature request related to a problem? Please describe.

Currently we heavily rely on the `post-checkout` git hook to do the following:
- Cleanup typescript build artifacts from the previous state that may cause inconsistent/erroneous assumptions when developing
- Compile typescript source files

### Describe the solution you'd like

Lots of suggestions were brought up in the above discussions.

In essence:
- Git hooks are not comprehensive. (e.g. git stash pop) This leads to inconsistent compilation behaviour.
- Each of us likely follows slightly different workflows, we should support something that is independent of this
  > we can let the developer decide based on their individual preferences, workflow (did you do a git stash? merge? ...), PC capabilities (referring the point on tsc watcher being expensive).

1. I prefer supporting these 3 separate commands:

```
// Cli parameter will be the site. index.js refers to the cli package's index.js entrypoint

// Run like "npm run buildAndDev docs", "npm run buildAndDev packages/core/template/default"
"buildAndDev": "npm run clean && npm run build:backend && node ...index.js s -d -o"

// Run like "npm run watchAndDev docs", "npm run buildAndDev packages/core/template/default"
"watchAndDev": "npm run clean && run the tsc watcher ... && node ...index.js s -d -o",

// And one, if you don't need to recompile source files. (e.g. if just iterating on some frontend feature)
"dev": "node ...index.js s -d -o""
```

This:
- Accounts for differing PC capabilities (if you think running tsc watcher is too heavy) / Typescript workflows (if you use ide build on save features)
- Independent of any git events / hooks
- Removes the need for `npm link` to get started. (we can still put this somewhere in the docs, but ideally we can even just do `npm run buildAndDev ../somefolder/outside/the/markbind/project/directory`)

<br>

2. Removing git hooks for `npm run clean` and `npm run build:backend`.

Reasons:
- Consistency, git hooks are not comprehensive.
- Remove unnecessary builds/cleans from checkouts (e.g. checkout to commit on different branch referencing same `HEAD`)

PR #2035, opened 2022-09-23
Running tests/lints before commit and push

### LINKED ISSUE #1966: Run linter via pre-commit hook

### Any related issues?

_No response_

Code Quality, DevOps

It is less effective for CI to capture trivial errors when they can be (automatically) caught and addressed locally before a PR is proposed. One example is the need to fix linting errors such as "Strings must use singlequote"
image
[The fix by dev](https://github.com/MarkBind/markbind/pull/1965/commits/5304b50a8b73b32ee68cd8669c00fdf0387130ee)

These sorts of errors are only caught at the CI stage because linting (and testing) do not get executed automatically. Also, we actually do mention the need to run linters in our DG:
image

Make use of git hooks (pre-commit) to run lint automatically. To reduce friction (time needed to execute), we should explore running our linters only on changed files.
- this seems possible with https://github.com/okonet/lint-staged (https://medium.com/@qjli/daily-coding-tips-8-how-to-only-eslint-only-changed-files-c5b0f6e3a31c)
- with this, we will still need to experiment and verify that the time taken is short enough

If we go with git hooks, also possible to execute at the post-commit stage.

### Additional context

I am raising this issue because, for the particular case of first-time contributors, CI won't be run automatically. This means they might have submitted a change, and only when senior devs approve the CI run (sometimes after a few hours or even a day), will they be notified that there is a need to make trivial adjustments to the code. While this is ok, I feel like we can help improve this process slightly with the proposed solution.

EltonGohJH:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves  #1966 #2021

Git hooks are not executable by default.  (For linux and macbook only I think)
I needed to make it executable using chmod +x.
Not sure, whether there is a way we can change the permission by default. 
Alternatively, we add it to the developer docs to ask developers to make their githooks executable.

**Testing instructions:**
(Sample image to showcase how it looks like)
image
image
image

**Add pre-commit and pre-push hooks to run lint and test**

---

- [x] No unrelated changes


ang-zeyu:
> Git hooks are not executable by default. (For linux and macbook only I think)
> I needed to make it executable using chmod +x.

We're currently relying on the `npm install:hooks` script for this (implicitly run when you run `npm run setup`), there should be no need to `chmod +x`. (was the post-checkout hook running fine for you previously?)

> Alternatively, we add it to the developer docs to ask developers to make their githooks executable.

But I think it would be good to document nonetheless especially with these few extra additions.

.githooks/pre-push
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
+RED="\033[1;31m"
+BLUE="\033[34m"
+GREEN="\033[1;32m"
+NOCOLOR="\033[0m"
+
+exit_code=1
+echo "${BLUE}(pre-push) Building backend...${NOCOLOR}"
+npm run build:backend
```
ang-zeyu, line 9: Think we could also add the `npm run clean` script to ensure there aren't any undeleted files when running the tests

https://github.com/MarkBind/markbind/blob/b10e424352cb3041e8216444aae2cfa2debf6181/.githooks/post-checkout#L12

.githooks/pre-commit
```diff
@@ -0,0 +1,30 @@
...
+RED="\033[1;31m"
+BLUE="\033[34m"
+GREEN="\033[1;32m"
+
+linter_exit_code=0
+eslint_files=$(git diff --cached --diff-filter=d --name-only --line-prefix=`git rev-parse --show-toplevel`/ | grep '.js$\|.ts$\|.vue$')
+stylelint_files=$(git diff --cached --diff-filter=d --name-only --line-prefix=`git rev-parse --show-toplevel`/ | grep '.css$\|.vue$')
+
+echo "${BLUE}(pre-commit) Running linters..."
+if [ "$eslint_files" ]
```
ang-zeyu, line 11: Running lint pre commit might need the typescript files need to be built first. (there are some eslint checks for imports)

I don't think it would be as much an issue once we've migrated as much as possible to typescript though.

Should we add `npm run clean` and `npm run build:backend` on every commit as well for now? Seems like a huge hindrance to the development experience but also a temporary one.

EltonGohJH:
@ang-zeyu 
 Was the post-checkout hook running fine for you previously? 
 Nope. I am using WSL2 btw.
 
 I will write some basic documentation to the developer guide workflow.

ang-zeyu:
I see, I was under the impression `git config core.hooksPath .githooks` would set the right permissions. TIL!

I think we could even think about running the prepush hook manually in CI (versus the npm scripts) to test for this.

.githooks/pre-commit
EltonGohJH: @ang-zeyu 
Personally, I think that would be fine.  Experienced developers that want to skip the check can always use the no verify flag.

ang-zeyu: Also a bunch of git UIs that don't have this option unfortunately. 😢

ang-zeyu: (let's add it then)

.githooks/pre-commit
```diff
@@ -0,0 +1,37 @@
...
+npm run build:backend
+
+linter_exit_code=0
+eslint_files=$(git diff --cached --diff-filter=d --name-only --line-prefix=`git rev-parse --show-toplevel`/ | grep '.js$\|.ts$\|.vue$')
+stylelint_files=$(git diff --cached --diff-filter=d --name-only --line-prefix=`git rev-parse --show-toplevel`/ | grep '.css$\|.vue$')
+
+echo "${BLUE}(pre-commit) Running linters...${NOCOLOR}"
+if [ "$eslint_files" ]
+then
+  eslint --fix $eslint_files
```
ang-zeyu, line 20:
```suggestion
  npx eslint --fix $eslint_files
```

.githooks/pre-commit
```diff
@@ -0,0 +1,37 @@
...
+  linter_exit_code=$?
+fi
+
+if [ "$stylelint_files" ]
+then
+  stylelint --fix $stylelint_files
```
ang-zeyu, line 26:
```suggestion
  npx stylelint --fix $stylelint_files
```

.githooks/pre-commit
ang-zeyu: would the `lint:fix` script work as well? running on the changed files is certainly faster but might falsely pass in our case (e.g. if file A is deleted, but file B references A an import lint check should fail)

same for `stylelint`-ing

.githooks/pre-commit
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env sh
+RED="\033[1;31m"
+BLUE="\033[34m"
+GREEN="\033[1;32m"
+NOCOLOR="\033[0m"
```
ang-zeyu, line 5: The colours seem to be printed out literally on powershell (e.g. `\033[34m(pre-commit) Running linters...\033[0m`). Should we remove these for now? (imo not a priority, but nice to have!)

This is the library we use for colouring our test scripts https://github.com/chalk/chalk, you can take a look at its implementation details if you would like to work this out as well!

docs/devGuide/workflow.md
```diff
@@ -256,6 +256,15 @@ It is also possible to auto-fix some (not all) style errors, using `npm run lint
 ESLint has [integrations with popular editors](https://eslint.org/docs/user-guide/integrations). They offer features such as "fix errors on save", which will make development smoother.
 </box>
 
+## Git hooks
+We have three git hooks in our project. 
+
+To skip running the pre-commit hook or pre-push hook, you can use the --no-verify flag (e.g. git push --no-verify fork branch).
```
ang-zeyu, line 262: minor nit:
```suggestion
To skip running the pre-commit hook or pre-push hook, you can use the --no-verify flag (e.g. git push --no-verify fork_branch).
```

tlylt:
> as the post-checkout hook running fine for you previously?)

iirc existing githooks were working fine for me without further action (I'm using Windows). My experience is that if I add a new executable script on Windows I should change the file permission correctly before I git commit it. Else I will have to force update the file permission for git to record the change. Afterwards the permission should work fine

.githooks/pre-commit
```diff
@@ -0,0 +1,22 @@
...
+echo "(pre-commit) Cleaning compiled files..."
+npm run clean
+
+echo "(pre-commit) Building backend..."
+npm run build:backend
+
+linter_exit_code=0
+
+echo "(pre-commit) Running linters..."
+npm run lintfix
```
ang-zeyu, line 12: Hmm I think the right behaviour here should be to `npm run lint` first, then `lintfix` if that fails, so erroneous changes don't get committed at all. (better yet if we can get lintfix to return non-0 code in one go if something is fixed)

On the same lines, we can stash local, unstaged changes before running the checks, then pop them back out once done, to ensure the checks are run on committed history. (consider how this should interact with `lintfix` too)

ang-zeyu: > stash

(and the case there is nothing to stash)

EltonGohJH: On my end lintfix does return non-zero code if something cannot be fixed. Is that alright?

But I do agree that we should stash the files first before we run the linters and even the tests when pushing.

ang-zeyu: > On my end lintfix does return non-zero code if something cannot be fixed.

This is the case for me too, its alright since the fix needs to come from the developer. What about the case something is fixed? (imo this should fail, as `lintfix` does not automatically add/commit the fixes for you)

EltonGohJH: I seee. I will just run npm run lint instead then.

ang-zeyu: That's another option!

ang-zeyu: We can put up the `lintfix` as another item for someone to work on -- not a priority but nice to have.

.githooks/pre-push
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env sh
 
+git stash -q --keep-index
```
ang-zeyu, line 3: what if there are no items to stash?

EltonGohJH: Do you mean that we should stash untracked files also? 
If yes, I think we need to do that.

ang-zeyu: nope -- but yes we should stash untracked files too, e.g. there may be issues with import checks / running tests from newly added files. Nice catch 💯 

More so simply: if there's nothing to stash, no stash would be created, in which case we shouldn't try running `pop` at all (e.g. it might pop an older manual stash)

EltonGohJH: The flag --keep-index cancels removal of the cached changes from the working directory (but they are still included in the stash). (Explained by [stackoverflow](https://stackoverflow.com/questions/69110050/in-git-pre-commit-hook-temporarily-remove-all-changes-that-are-not-about-to-be))

So, a stash would still be created and I tested it on my end. (That seems to be the case)

ang-zeyu: Hmm to check we're on the same page, here's an example, I don't think the `keep-index` option is for this:

1. I'm working on XX branch, I run `git stash`
2. I switch to YY branch, I `git add .` *all* changes
3. `git commit -m "..."`
4. In the script, you're running `git stash -q --include-untracked --keep-index`. This does nothing since I've `git add .` everything. (That seems to be the case on my end at least as well)
5. `git stash pop -q` accidentally pops my old stash from XX branch.

EltonGohJH: There would be things to stash as the staged files would be stashed. With --keep-index, we will stash the staged files but  not remove it. So, something will still be stashed.
Does this behaviour still occur when you run attempt to commit. It seems working on my end hmmm.

ang-zeyu: Managed to get it to work, I needed to actually make some modifications (was testing with no changes at all). Thanks for the link!

> was testing with no changes at all

The XX branch stash is still popped here. Should we do something simple like naming the stash (with something very unique) and popping only that named stash to prevent this?

ang-zeyu: Another issue seems to be file renames, in this case the rename itself is stashed (`--keep-index` dosen't seem to work literally). Is this occuring for you as well?

ang-zeyu: > prevent this

I think could also help prevent other things we haven't accounted for in the discussion here. 🙂

ang-zeyu: one downside/potential issue of naming might be if the script gets in a bad state (leftover stash from previous precommit run)

or we can try to make the name unique only within the same precommit run.. 💭

EltonGohJH: @ang-zeyu 
I tried doing that by making the name unique but then I realised that you cannot drop/pop base on stash name. You can only drop or pop by index.

So, what I am doing will be checking the stash count before and after stashing to determine whether to pop.

ang-zeyu: > cannot drop/pop base on stash name

Does this work? https://stackoverflow.com/questions/11269256/how-do-i-name-and-retrieve-a-git-stash-by-name

> So, what I am doing will be checking the stash count before and after stashing to determine whether to pop.

That works too!

EltonGohJH: `git stash pop stash^{/my_stash_name}`
There are quite a lot of guide that states this but this does not actually works.

ang-zeyu: > --keep-index

There's an issue here:
1. Change some line to trigger a lint fail (e.g. add trailing whitespace), stage it
2. Change it again, don't stage it
3. Run your precommit script
4. On stash pop, a merge conflict occurs

Would the alternative pointed out here work? https://stackoverflow.com/questions/20028507/git-stash-uncached-how-to-put-away-all-unstaged-changes/34681302#34681302

(make a temp commit without running pre commit, stash, soft reset, continue)

iirc (it's been a while since I used it, do double check 😅) [precommit tool](https://pre-commit.com/) dosen't have this issue too, could also reference how it does it.

ang-zeyu: > merge conflict

presumably this occurs as the staged changes are also stashed.

ang-zeyu: >You can only apply but you cannot pop.

That makes sense, per https://git-scm.com/docs/git-stash.

I think apply has to be followed by a `drop <stash>` (if this works too)

ang-zeyu: > Change some line to trigger a lint fail (e.g. add trailing whitespace), stage it

Screenshot 2022-10-03 205951

> Change it again, don't stage it

Screenshot 2022-10-03 210020

> On stash pop, a merge conflict occurs

Screenshot 2022-10-03 210104

ong6:
Hi @EltonGohJH, just checking up on your progress for this PR, do request a review from one of us once you're completed with the changes that were mentioned by zeyu.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
+default_install_hook_types: [pre-commit, post-checkout, pre-push]
+repos:
+  -   repo: local
+      hooks:
+        -   id: run lint
+            name: run lint
```
ang-zeyu, line 6: As discussed, we can probably run lintfix now that precommit is used 🙂

docs/devGuide/settingUp.md
```diff
@@ -22,11 +22,17 @@ This page explains how to set up your development environment to start contribut
    **Graphviz** v2.38 or later<br>
    %%(The above two are required for one of the third-party libraries used by MarkBind)%%
 
+1. **Python** 3 or later, and<br>
+   **pre-commit** v2.20 or later<br>
+   %%(The above two are required for python pre-commit hooks)%%
```
ang-zeyu, line 27: Should we be using the `.pyz` instead? I think some might not be comfortable installing a global dependency.

docs/devGuide/settingUp.md
```diff
@@ -43,15 +49,15 @@ We recommend the **WebStorm IDE** for working with MarkBind code.
    <popover content="Under the hood, this calls `npm ci` and `lerna bootstrap`">`npm run setup`</popover>
    in the **root folder** of your cloned repo.
 
+1. **Install git hooks** by running `npm run install:hooks` in the **root folder** of your cloned repo.
```
ang-zeyu, line 52: Should we move all the changes to a separate section since hooks are optional convenience features? (point on python we discussed) maybe after `Setting up the dev environment` as a separate heading.

pre-commit-scripts/pre-commit.sh
```diff
@@ -0,0 +1,30 @@
...
+then
+  git stash pop -q
+fi
+
+if [ $linter_exit_code -ne 0 ]
+then
+  echo "❌ (pre-commit) Please fix linting errors before committing!"
+  exit 1
+else
+  echo "✔ (pre-commit) ESlint and Stylelint did not find any errors!"
```
ang-zeyu, line 28: >  run lintfix

Just make sure to adapt the messages accordingly so what's happening is clear, e.g.

```
echo "✔ (pre-commit) ESlint and Stylelint did not find any unfixable errors!"
echo "✔ (pre-commit) Some files may be automatically lint fixed, you will need to ... in order for pre-commit ..."
```

docs/devGuide/settingUp.md
EltonGohJH: Sounds like a good idea and that will reduce the dependency to just python.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+            entry: ./pre-commit-scripts/pre-commit.sh
+            pass_filenames: false
+            language: system
+            stages:
+              - commit
+        -   id: clean and build
+            name: clean and build
+            entry: ./pre-commit-scripts/post-checkout.sh
```
ang-zeyu, line 14: 👍 nice! Thanks for adapting this as well.

docs/devGuide/settingUp.md
EltonGohJH: Lets do that

ang-zeyu (commented):
Some context in slack discussions as to why we switched to precommit:
- Proven solution that handles "stashing" unstaged changes [nicely](https://github.com/MarkBind/markbind/pull/2035#discussion_r985756094)
- Most developers should have python available
  - Hooks are ultimately still optional convenience features otherwise

.pre-commit-config.yaml
EltonGohJH: I think lintfix wont work since it modifies the file right.

ang-zeyu: > modifies the file right.

That's a good thing, pre-commit will fail in this case so that we can verify and stage the modifications before recomitting

Many existing [hooks](https://pre-commit.com/hooks.html) rely on this behaviour.

package.json
```diff
@@ -13,8 +13,8 @@
     "csslint": "stylelint **/*.css **/*.vue",
     "csslintfix": "stylelint **/*.css **/*.vue --fix",
     "dev": "tsc --watch",
-    "install": "npm run install:hooks",
-    "install:hooks": "git config core.hooksPath .githooks",
+    "install:hooks": "python3 ./pre-commit/pre-commit-2.20.0.pyz install",
+    "uninstall:hooks": "python3 ./pre-commit/pre-commit-2.20.0.pyz uninstall",
```
ang-zeyu, line 17: Would remove these 2 npm commands completely (see `settingUp.md` comment for rationale)

docs/devGuide/settingUp.md
```diff
@@ -45,13 +49,16 @@ We recommend the **WebStorm IDE** for working with MarkBind code.
 
 1. **Congratulations!** Now you are ready to start modifying MarkBind code.
 
+
+## Setting up the git hooks (optional but recommended)
+
+1. **Install git hooks** by running `npm run install:hooks` in the **root folder** of your cloned repo.
```
ang-zeyu, line 55: It may be better to move and document the commands here directly as `python` vs `python3` alias availability is very system setup dependent.

```suggestion
There are a few [Git hooks](./workflow.md#git-hooks) implemented using the [precommit](https://pre-commit.com/) tool that runs common tasks like linting, automated tests and compiling typescript files.

**Install the hooks** by running `python3 ./pre-commit/pre-commit-2.20.0.pyz install` in the **root folder** of your cloned repo.

If you ever need to uninstall the hooks, simply run `uninstall` instead of `install`.
```

pre-commit/pre-commit-scripts/pre-commit.sh
```diff
@@ -0,0 +1,31 @@
...
+  git stash pop -q
+fi
+
+if [ $linter_exit_code -ne 0 ]
+then
+  echo "❌ (pre-commit) Please fix linting errors before committing!"
+  exit 1
+else
+  echo "✔ (pre-commit) ESlint and Stylelint did not find any unfixable errors!"
+  echo "✔ (pre-commit) Some files may be automatically lint fixed, please review and commit again!"
```
ang-zeyu, line 29: Adding some rationale:

```suggestion
  echo "✔ (pre-commit) ESlint and Stylelint did not find any unfixable errors!"
  echo "               Some files may have been automatically lint fixed, which will cause precommit to fail."
  echo "               Verify and stage these changes, then recommit as needed."
```

pre-commit/pre-commit-scripts/post-checkout.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
```
ang-zeyu, line 2: I think this should run on all files, staged and unstaged contrary to the other two. But not a huge issue as we are going to be moving away from post-checkout hooks https://github.com/MarkBind/markbind/issues/2020.

If there's an easy way, feel free to add it in though.

ang-zeyu (commented):
last few nits:

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+            entry: ./pre-commit/pre-commit-scripts/post-checkout.sh
+            pass_filenames: false
+            language: system
+            always_run: true
+            stages:
+              - post-checkout
+        -   id: run tests
+            name: run tests
+            entry: ./pre-commit/pre-commit-scripts/pre-push.sh
+            pass_filenames: false
```
ang-zeyu, line 23: should this have `always_run: true` as well? Its not trigerring on my end without it.

pre-commit/pre-commit-scripts/post-checkout.sh
EltonGohJH: Hmm I dun think pre-commit supports this 😢

.pre-commit-config.yaml
ang-zeyu: The `verbose: true` flag might be helpful here too so we can ensure the output of the tests are correct.

pre-commit/pre-commit-scripts/post-checkout.sh
ang-zeyu: 🙆‍♂️

.pre-commit-config.yaml
EltonGohJH: I think no harm adding always_run: true. We can add that in. But it was working on my end, in fact this hook ran when I pushed the latest commit hahahha.

EltonGohJH: We can enable verbose: true too.

ang-zeyu: > I think no harm adding always_run: true. We can add that in. But it was working on my end, in fact this hook ran when I pushed the latest commit hahahha.

Yes, sorry - correction.

> Its not trigerring on my end without it **sometimes**.

I'm not entirely sure what's the cause, but agreed having `always_run: true` would be no harm.

ang-zeyu: 🤔🤔🤔 Please chime back in if you happen to find out what that option does in the context of push! @EltonGohJH

ang-zeyu (approved):
Lgtm 👍

PR #2053, opened 2022-12-12
Include repeated attributes warning for panel elements

### LINKED ISSUE #1764: Inconsistent behaviour for repeated slots/attributes
MacOS 12.0.1

* **MarkBind Version:** 3.1.1

* Create a dropdown component with the same slot and attribute. 
* Create any other component with the same slot and attribute

(Source code causing the issue:  [link](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.js#L109-L119))

E.g. 
```html
### Dropdowns
This triggers a logger warning:
<dropdown header="Does not appear, overwritten by slot">
  <span slot="header">Dropdown header slot</span>
  <li><a href="#dropdown" class="dropdown-item">Lorem ipsum</a></li>
</dropdown>

### Panels
This **does not** trigger a warning:
<panel header="Does not appear, overwritten by slot">
  <span slot="header">Panel header slot</span>
  Lorem ipsum
</panel>

### Popover
This **does not** trigger a warning:
<popover header="Does not appear, overwritten by slot" content="Does not appear">
  <span slot="header">Popover header slot</span>
  <span slot="content">Popover content slot</span>
  Lorem ipsum
</popover>
```

I expect that repeating an attribute as a slot either always triggers a warning, or never triggers a warning.

A warning is triggered for dropdowns, but not for any other component.
Screenshot 2022-02-13 at 3 34 08 PM

Dropdowns has [some code](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.js#L109-L119) that checks against having both the header slot and header attribute, and logs a warning.

I'm wondering if there is a reason why Dropdowns have this specific logger warning, but not any other component? 
I tried removing the check for the duplicated slot/attribute in Dropdowns to check if this code exists because failing to delete the header attribute will cause issues, but it does not seem like the case, as Dropdowns continue to work correctly without the check.

lhw-1:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1764
Related to #1780 which partially resolved this issue but was undocumented

While this pull request resolves the original issue of repeated attributes in `panel` elements not logging a warning message, there may be additional work required to streamline the process of adding warnings instead of declaring a warning case for every element as is currently being done: [link](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.js)

Three further issues have been proposed thus far: 

1. Streamline the process of adding warnings as opposed to declaring a warning for each attribute of every element that requires a warning, as highlighted above.
2. Implement a way to have unit tests for logging output from the winston logger. This will allow us to have unit tests for warnings and log outputs as well, which should increase robustness of the system as a whole.
3. Following (2), unit / functional tests for warnings (for the rest of the elements with warnings) should be written.

**Testing instructions:**

`Include repeated attributes warning for panel elements`

---

- [X] No unrelated changes


tlylt:
Hi @lhw-1, thank you for contributing!

Will you be adding unit tests and functional tests for this bug fix? Would be great to have them in place to keep this bug in check. (You can get more information on our testing practices [here](https://markbind-master.netlify.app/devguide/workflow#testing))

lhw-1:
@tlylt Thanks for the heads up! I will add in some tests for this issue.

lhw-1:
Looks like relevant unit tests for panel elements have been previously added under `packages/core/test/unit/html/NodeProcessor.data.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.data.js#L3-L56)), and `packages/core/test/unit/html/NodeProcessor.test.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.test.js#L43-L48)).

Functional tests for panel elements have been added in commit [c489112](https://github.com/MarkBind/markbind/pull/2053/commits/c489112fd18c1a970e9b901c2144c7930f17b7fe).

tlylt:
> Looks like relevant unit tests for panel elements have been previously added under `packages/core/test/unit/html/NodeProcessor.data.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.data.js?rgh-link-date=2023-01-11T00%3A46%3A41Z#L3-L56)), and `packages/core/test/unit/html/NodeProcessor.test.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.test.js?rgh-link-date=2023-01-11T00%3A46%3A41Z#L43-L48)).
> 
> Functional tests for panel elements have been added in commit [c489112](https://github.com/MarkBind/markbind/pull/2053/commits/c489112fd18c1a970e9b901c2144c7930f17b7fe).

If suitable, you can add tests to the existing test files related to the panel. Have you added new test cases specifically for triggering/not triggering the warning?

lhw-1:
> Have you added new test cases specifically for triggering/not triggering the repeated attributes warning? (Possibly with relevant text mentioning what's being tested for the test cases)

I could not find test cases specifically for triggering / not triggering the repeated attributes warnings for the Popover / Dropdown elements (warnings for these elements have been previously implemented) after a quick search, but I could have missed them, so I'll give it a more thorough search!

On another note, the logger currently uses the winston logger to output such warnings for elements, but I did not find any test cases testing the winston logger (again, I could have missed them...). If so, it may be worthwhile to look into implementing test cases for the logger, which seems reasonably feasible (but out of scope for this particular pull request, so I'll open another issue for this if this happens to be the case).

tlylt:
> I could not find test cases specifically for triggering / not triggering the repeated attributes warnings for the Popover / Dropdown elements (warnings for these elements have been previously implemented) after a quick search, but I could have missed them, so I'll give it a more thorough search!

Probably means they are not written, can file an issue for those separately.

> On another note, the logger currently uses the winston logger to output such warnings for elements, but I did not find any test cases testing the winston logger (again, I could have missed them...). If so, it may be worthwhile to look into implementing test cases for the logger, which seems reasonably feasible (but out of scope for this particular pull request, so I'll open another issue for this if this happens to be the case).

I don't think we have the mechanism to match console output at the moment, but we still want to have the cases captured so that minimally on visual inspection we can see the warnings being logged.

Good suggestion for having such an implementation... yes we can file another issue to take that forward 👍

lhw-1:
To clarify, the case in `packages/core/test/unit/html/NodeProcessor.data.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.data.js#L21-L30)) seems to be the only one that is relevant to this issue under `packages/core/test/unit`, and there do not seem to be other relevant unit tests specific to warnings implemented for other elements either. Should I add in a statement that specifies that a warning will be thrown if repeated attributes are used under the above test case? Though in that case, I think similar statements should also be added in for all the other elements (which may be out of scope for this pull request).

packages/core/src/html/MdAttributeRenderer.js
```diff
@@ -105,8 +105,12 @@ class MdAttributeRenderer {
    */
 
   processPanelAttributes(node) {
-    this.processAttributeWithoutOverride(node, 'alt', false, '_alt');
-    this.processAttributeWithoutOverride(node, 'header', false);
+    if (!this.hasSlotOverridingAttribute(node, 'alt')) {
+      this.processAttributeWithoutOverride(node, 'alt', false, '_alt');
+    }
```
jovyntls, line 110: We are not currently supporting the `alt` attribute as a slot. I believe the `processAttributeWithoutOverride` for `alt` is because `alt` clashes with an existing HTML attribute, so we want to convert it to `_alt` instead. So we won't need to throw a warning for this :)

packages/cli/test/functional/test_site/testPanels.md
```diff
@@ -0,0 +1,55 @@
...
+
+<panel header="Should not appear: Overwritten header">
+  <span slot="header">Correct header</span>
+  Lorem ipsum...
+</panel>
+
+<br>
+
+<panel alt="Should not appear: Overwritten alternative header">
+  <span slot="alt">Alternative header</span>
```
jovyntls, line 42: We can remove the `alt`s in these tests since `alt` is not supported as a slot (see below)

jovyntls (changes requested):
Thank you @lhw-1 , the tests look pretty comprehensive in testing slot/attribute/overrides! Just some comments on the `alt` slot. 

On a side note, we could use more tests for all the features we support with panels (e.g. peek, no-close, etc). That's out of the scope of the PR though so don't worry about it :)

jovyntls:
> Should I add in a statement that specifies that a warning will be thrown if repeated attributes are used under the above test case? 
> Though in that case, I think similar statements should also be added in for all the other elements (which may be out of scope for this pull request).

Agreed on this, I think for this PR we can add a comment for just the panels. 😃

tlylt:
> To clarify, the case in `packages/core/test/unit/html/NodeProcessor.data.js` ([here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.data.js?rgh-link-date=2023-01-11T01%3A23%3A07Z#L21-L30)) seems to be the only one that is relevant to this issue under `packages/core/test/unit`, and there do not seem to be other relevant unit tests specific to warnings implemented for other elements either. Should I add in a statement that specifies that a warning will be thrown if repeated attributes are used under the above test case? Though in that case, I think similar statements should also be added in for all the other elements (which may be out of scope for this pull request).

Do those unit tests have their individual cases that they are checking for? If so I think we might want to create another similar one (somewhere below perhaps) with the sole focus on the conditions that will trigger the warning. This might help make tests isolated and avoid unnecessary changes in the future. What do you think, @lhw-1?

packages/core/src/html/MdAttributeRenderer.js
lhw-1: Noted! I have removed the warning in [25f4250](https://github.com/MarkBind/markbind/pull/2053/commits/25f42501e8da056540cb26e7b51f23ce0746c88f) :+1:

packages/cli/test/functional/test_site/testPanels.md
lhw-1: Also noted! I have removed this in [25f4250](https://github.com/MarkBind/markbind/pull/2053/commits/25f42501e8da056540cb26e7b51f23ce0746c88f) as well :+1:

lhw-1:
> Do those unit tests have their individual cases that they are checking for? If so I think we might want to create another similar one (somewhere below perhaps) with the sole focus on the conditions that will trigger the warning. This might help make tests isolated and avoid unnecessary changes in the future. What do you think, @lhw-1?

Ah, I see what you mean. Yes, I think that would be a good solution in the meantime as well - I'll get to it!

packages/cli/test/functional/test_site/expected/testPanels.html
```diff
@@ -0,0 +1,288 @@
...
+        <panel><template #header>
+            <p>Correct header</p>
+          </template>
+          Lorem ipsum...
+        </panel>
+        <br>
+        <panel><template #header>
+            <p>Correct header</p>
+          </template><template #_alt>
+            <p>Alternative header</p>
```
jovyntls, line 220: Thanks for removing the `alt` attribute from the tests! Remember to update these files as well :)

jovyntls (changes requested):

packages/cli/test/functional/test_site/expected/testPanels.html
lhw-1: Thank you for taking note of this! I will remember to update tests next time :sweat_smile: 

Fixed in [eef8681](https://github.com/MarkBind/markbind/pull/2053/commits/eef8681472a7f46b39bb401ae7138e0172e8787b) :+1:

lhw-1:
I have added unit tests for panels [here](https://github.com/MarkBind/markbind/pull/2053/files#diff-482ac5d29bff5c29c1bde6124b8a8162c2ea7312c422df5aa01decce32155d6bR21-R38). It seems that there were actually unit tests for this case but only for dropdown elements [here](https://github.com/lhw-1/markbind/blob/master/packages/core/test/unit/html/NodeProcessor.data.js#L304-L316), for future reference.

Once this is good to go & hopefully gets merged, I will open up new issues as mentioned in the "things to highlight / discuss" section above (I've updated it) :+1:

jovyntls (approved):
LGTM! 👍 

What you've mentioned under "things to highlight/discuss" sound like good ideas to me as well! Thanks for opening the issues :)

tlylt:
@all-contributors please add @lhw-1 for code

PR #2054, opened 2022-12-13
Fix hasPageNav logic to only apply data-bs attributes if there are navigable headings

tlylt:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [x] Bug fix
- [ ] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1954

**Anything you'd like to highlight/discuss:**
nil

**Testing instructions:**
- check functional test written
- check reproduction steps in the original issue no longer generates an error when running `markbind serve -d`

**Proposed commit message: (wrap lines at 72 characters)**
Fix hasPageNav logic

Bootstrap data-bs attributes are applied to the body element when
there is pageNav specified in the frontmatter and the layout.

Let's fix the hasPageNav logic to account for the fact that if there 
are no navigable headings, the Bootstrap attributes are not needed.

This removes the read null error in the console.

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [x] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No unrelated changes

Issue #1954, opened 2022-06-29
Remove `data-bs-*` attributes if page nav has no anchor links


jonahtanjz:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Tell us about your environment

Windows 10

### MarkBind version

3.1.1

### What did you do? Describe the bug

Bootstrap's utility js will throw an error if there are no headings on the page that match the specified heading level.

![image](https://user-images.githubusercontent.com/47470981/176066394-1fcc4648-7cd8-4acd-a64d-1ca911552d0d.png)

Since there are no anchor links in the page nav, the `#mb-page-nav` id will not be present and hence the `data-bs-target` is unable to find the element. 

Will have to not render the `data-bs-*` attributes in [`page.njk`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Page/page.njk#L41) for this particular case.

### Steps to reproduce

For example, serve the below code in a new `md` file and observe the error in the console. 

```md
<frontmatter>
  pageNav: 2
  pageNavTitle: Test Title
</frontmatter>

This is a test
```

### Expected behavior

Should not have any errors.

### Actual behavior

Error can be seen in the console.

PR #2058, opened 2022-12-16
Fix bug with dangling separator in `pageNav` when no title and nav headings given

### LINKED ISSUE #2055: Omit pageNav element if there are no navigable heading to remove unnecessary separator

### Any related issues?

_No response_

Windows 10

### MarkBind version

Master branch

image

### Steps to reproduce

- `markbind init`
- go to `contents/topic1.md`
- add the following:
```
<frontmatter>
  layout: default.md
  pageNav: 2
</frontmatter>
```
- remove the default `# Topic 1`
- run `markbind serve -d`
- view the topic1 page

### Expected behavior

- If there are no navigable headings and no page title, the separator should not appear (the whole pagenav should be omitted)
- If there are page title, just show the page title

### Actual behavior

See screenshot above

### Anything else?

_No response_

domlimm:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #2055.

Add a strict check on the title and nav headings which are both type `string` to see if they are available from the `*.md` file. This allows for an early return which in turn omits the entire `pageNav` component on the page.

Related to Issue #1954 and its PR #2054.

**Testing instructions:**

Fix bug with dangling separator in pageNav section

The separator by CSS styling, border, appears when there is no title
and nav headings specified in the markdown files. This should not
appear.

Let's fix the logic when rendering the pageNav section to check
if there is a title and nav headings provided.

This logic will remove the unwanted separator.

---

- [ ] No unrelated changes


ang-zeyu:
Does this enlarge the width of the main content area? We had some issues with that in the past, and decided to always render the page nav to ensure a consistent layout.

domlimm:
Hello @ang-zeyu! Thanks for reviewing the PR!

I have just tested it on this branch, it seems like the entire page still has the `page-nav` component. So, I suppose the main content area width stays the same. Not sure if this is the best way to solve this issue. You mentioned CSS?

With the current change:
image

Before the current change:
image

As for #2054 was discussing with yongliang about this issue and the relation to #1954. But yes it seems as though omitting the element fixes the error in console. Maybe @tlylt can chime in on this. Not entirely certain also..

ang-zeyu:
> CSS

Presumably the border is there as the content box of the page nav is taking up space, so we could try shifting the margins and paddings around to achieve the same look when there are headings then this fix when there aren't. But this might requires
 breaking changes to the user layout css file again... 😞 (but if not, it would be best)

This solution looks to be better in that case.

packages/core/src/Page/index.js
```diff
@@ -432,6 +432,12 @@ class Page {
       this.collectNavigableHeadings(content);
       const pageNavTitleHtml = this.generatePageNavTitleHtml();
       const pageNavHeadingHTML = this.generatePageNavHeadingHtml();
+
+      if ((!pageNavTitleHtml || pageNavTitleHtml.length === 0)
+        && (!pageNavHeadingHTML || pageNavHeadingHTML.length === 0)) {
```
ang-zeyu, line 437: my understanding with #2054 is that the fix there is still ideal (dont render the data-bs attribute if we dont need it since we dont know whta might happen), even if this fixes it as a side effect as well.

---

the second .length checks here are unnecessary, empty strings are falsy.

tlylt: @domlimm yup as mentioned above, my PR should still stand as it tackles the bs attribute part. Your PR should be targeting just the #2055 part

domlimm: thanks for highlighting that! will make the change now

domlimm: @ang-zeyu @tlylt also, thanks so much for the help guys!

ang-zeyu (approved):

packages/core/src/Page/index.js
ang-zeyu: > Proposed commit message: (wrap lines at 72 characters)

Let's add a message for the squashed commit

packages/core/src/Page/index.js
```diff
@@ -432,6 +432,11 @@ class Page {
       this.collectNavigableHeadings(content);
       const pageNavTitleHtml = this.generatePageNavTitleHtml();
       const pageNavHeadingHTML = this.generatePageNavHeadingHtml();
+
+      if (!pageNavTitleHtml && !pageNavHeadingHTML) {
+        return '';
```
ang-zeyu, line 437: We can add in a simple functional test for this as well https://github.com/MarkBind/markbind/tree/master/packages/cli/test/functional
(there's a guide somewhere in the dev docs too if you need help writing one!)

ang-zeyu (changes requested):

packages/core/src/Page/index.js
domlimm: Will do!

packages/core/src/Page/index.js
domlimm: Thanks zeyu, will do that asap 👍🏻

domlimm:
hello guys, does anyone know why does this happen? i tried adding the functional test (could look at the stash at the bottom left) to this feat branch and running `npm run updatetest` was okay. but after rebasing the feat branch on the latest master it is not working anymore. not sure what's wrong 😓 

image

tlylt:
> hello guys, does anyone know why does this happen? i tried adding the functional test (could look at the stash at the bottom left) to this feat branch and running `npm run updatetest` was okay. but after rebasing the feat branch on the latest master it is not working anymore. not sure what's wrong 😓

Are you saying after rebasing and then adding functional test and `npm run updatetest` won't work? After rebasing did you run `npm run setup` to just make sure all the setup stuff is correct? Can also try merging instead. I think if there's any conflict to the generated files its fine, I will typically do a `npm run updatetest` and then check in the latest generated files, which should be the correct version

domlimm:
thank you so much for the help again yongliang! yes i missed that out.. will keep all these in mind!

i have added the test relevant to the bug fix. ready for your review please, and thank you!

packages/core/src/Page/index.js
tlylt: @domlimm could we add one more test for having only the pageNavTitle and it still works as expected?

domlimm: sure!

packages/cli/test/functional/test_site/testPageNavWithOnlyTitle.md
```diff
@@ -0,0 +1,11 @@
+<frontmatter>
+  layout: default.md
+  pageNav: 1
```
tlylt, line 3: Since pageNav is 1 and there's a H1 heading, is this example containing only the title?

domlimm: That's a good point. It contains title and the H1 heading. Sorry about that! Let me fix it now

domlimm: thanks yongliang!

It should look like this:

image

tlylt (approved):
LGTM. @domlimm thanks for the fixes

PR #2098, opened 2023-01-17
Migrate PageVueServerRenderer to TypeScript

raysonkoh:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Part of #1913.

**Overview of changes:**
As per title, followed the migration steps as per the developer docs.

Needed to add `eslint-disable import/no-import-module-exports` because of the `requireFromString` function, that uses nodejs `module` to load an in-memory `string` as a javascript in memory. I can't seem to find a reasonable alternative to the `requireFromString` function. I'm open to discussions on this.

**Testing instructions:**

```
Migrate PageVueServerRenderer to TypeScript
```

---

- [x] No unrelated changes


packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -78,8 +89,8 @@ async function renderVuePage(compiledVuePage) {
  *
  * Credits to Dominic
  */
-function requireFromString(src, filename) {
-  const m = new module.constructor();
+function requireFromString(src : string, filename : string) {
+  const m = new (module.constructor as any)();
```
raysonkoh, line 93: If I don't do this, typescript will complain: `This expression is not constructable.   Type 'Function' has no construct signatures.`

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -1,15 +1,25 @@
-const Vue = require('vue');
-const path = require('path');
-const fs = require('fs-extra');
-
-const VueCompiler = require('vue-template-compiler');
-const { renderToString } = require('vue-server-renderer').createRenderer();
+/* eslint-disable import/no-import-module-exports */
```
ang-zeyu, line 1: Is this due to the false positives from the `requireFromString` method? Let's document this in a comment.

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -1,15 +1,25 @@
...
+import { PageConfig, PageAssets } from './PageConfig';
+/* eslint-enable import/no-import-module-exports */
 
 let bundle = require('@markbind/core-web/dist/js/vueCommonAppFactory.min');
 
-const logger = require('../utils/logger');
+const { renderToString } = createRenderer();
+
+interface PageEntry {
+  page : any;
```
ang-zeyu, line 16: I'm surprised the airbnb base config dosen't enforce this, let's follow `var: type` spacing to be consistent with the other files.

It's presumably a small change so feel free to add this rule into our eslint configs here too if you'd like. Or we can create a follow up.

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/type-annotation-spacing.md

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -111,6 +123,6 @@ const pageVueServerRenderer = {
   pageEntries,
 };
 
-module.exports = {
+export = {
```
ang-zeyu, line 126: Since this file only uses named exports, this could be a good place to use es6 syntax instead:

```
export async function updateMarkBindVueBundle(newBundle : string) : Promise<void>
```

packages/core/src/Page/PageVueServerRenderer.ts
raysonkoh: Sure, done!

packages/core/src/Page/PageVueServerRenderer.ts
raysonkoh: I've updated the eslint rules since it's a pretty small change. Thanks for pointing this out!

packages/core/src/Page/PageVueServerRenderer.ts
raysonkoh: I think doing 
```
export const pageVueServerRenderer = { ... }
```
would be better than exporting the individual functions since it would be more obvious to the reader when using functions from `PageVueServerRenderer`. Wydt?

ang-zeyu: sounds good, I misread the original code as `module.exports = pageVueServerRenderer`. 🤦

.eslintrc.js
```diff
@@ -88,6 +88,7 @@ module.exports = {
           },
         ],
         "@typescript-eslint/return-await": "off",
+        "@typescript-eslint/type-annotation-spacing": "warn",
```
ang-zeyu, line 91: how about `error`-ing?

raysonkoh: Oops, done!

ang-zeyu (approved):

Issue #2075, opened 2023-01-15
Investigate any benefits of switching to parse5 library


ang-zeyu:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### What is the area that this feature belongs to?

Other

### Is your feature request related to a problem? Please describe.

We are currently using `htmlparser` for HTML parsing both directly and indirectly (cheerio).
This is a rather forgiving parser, but it is not always spec compliant. This causes issues for things that rely on strict spec compliance (e.g. SSR hydration, frontend libraries making such assumptions, ...).

The primary concern here is SSR hydration.

### Describe the solution you'd like

This ticket is for investigating what benefits parse5 might bring in resolving the string of SSR hydration issues we've encountered (do a search in the issues for "hydration"), versus htmlparser2.

For example (hypothetically), if parse5 returns a list of parsing errors, we can fail soft/hard so that the author is alerted and can correct the HTML markup.
Another hypothetical example is that parse5 could autocorrect invalid markup to be fully spec compliant, in which case no action would be required.

Document the findings in a comment here.

If doing the switch brings enough benefits, please create a follow up issue for implementing the switch.

Other notes:
- As parsing touches many components of the application, the implementation issue should have a fairly high LoE required. One example is adapting the patch we have for htmlparser2 to allow any tags to be parsed like `script/style` tags.
- cheerio 1.0.0 by default now uses parse5 instead of htmlparser2 in cheerio 0.2.2 (what we're using), so this could ease that migration.
- In any case, performing the investigation for this issue would be a good start.

PR #2112, opened 2023-01-27
Migrate codeBlockButton plugins to TypeScript

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Part of #1913 

**Overview of changes:**
Migrate codeBlockButtonsContainer to TS

Should we standardise a type for `node` that combines its possible types (e.g. `DomElement & cheerio.Element`)? I'm not sure if it's necessary (would having these different types for nodes cause issues as we continue the migration?).

**Testing instructions:**
N/A

(merge commit)

---

- [x] No unrelated changes


raysonkoh:
>  Should we standardise a type for node that combines its possible types (e.g. DomElement & cheerio.Element)? I'm not sure if it's necessary (would having these different types for nodes cause issues as we continue the migration?).

I think it's a good idea, we could maybe have our own type `MarkbindNode` (ideally immutable), and have some functions that wrap the common methods that we call on `DomElement` or `cheerio.Element`.

My thoughts are to do the migration fully first, before implementing this as this is potentially a breaking change. Also there are not enough unit tests for the core package that I'm confident that it will detect any subtle bugs introduced. Wdyt?

jovyntls:
> I think it's a good idea, we could maybe have our own type `MarkbindNode` (ideally immutable), and have some functions that wrap the common methods that we call on `DomElement` or `cheerio.Element`.

Yeah agree with having an interface for Node! Standardising all the common methods would make the codebase more readable and less confusing, especially since we have code that looks similar with significant differences (e.g. `cheerio.html(node)` vs `cheerio(node).html`

> My thoughts are to do the migration fully first, before implementing this as this is potentially a breaking change. Also there are not enough unit tests for the core package that I'm confident that it will detect any subtle bugs introduced. Wdyt?

My thoughts:
- Creating the types first could help with the TS migration by making it easier to type `node` arguments/parameters. We may also be able to reduce the number of checks for undefined/null attributes if we have a type for nodes.
- I agree that having more unit tests would good, since we may end up introducing some unintended effects with such a large scale migration
- Also, if we further refactor the code to use this new Node interface/type, the changes might be so significant that our git history detects the `.ts` file as a new file rather than a renaming of a previous `.js` file

tlylt:
> My thoughts:
> 
> * Creating the types first could help with the TS migration by making it easier to type `node` arguments/parameters. We may also be able to reduce the number of checks for undefined/null attributes if we have a type for nodes.
> * I agree that having more unit tests would good, since we may end up introducing some unintended effects with such a large scale migration
> * Also, if we further refactor the code to use this new Node interface/type, the changes might be so significant that our git history detects the `.ts` file as a new file rather than a renaming of a previous `.js` file

Jumping in with some brief comments since I have not done any TS migration myself:
- will be good to create an issue to keep track of the TODO
- in the issue, also would be good to mention an expedient solution when encountering the type of DomElement vs cheerio.Element for the moment so that we can deal with this uniformly (as much as possible).
  - `DomElement & cheerio.Element` seems ok, perhaps @raysonkoh can give a second opinion on this
- upon checking, `cheerio` seems to have migrated to TypeScript with the [latest `v1` release](https://github.com/cheeriojs/cheerio/releases/tag/v1.0.0-rc.12), might be useful to take a look there
  - could be related to #2075

packages/core/src/plugins/codeBlockCopyButtons.ts
```diff
@@ -45,9 +46,9 @@ const copyCodeBlockScript = `<script>
     }
     </script>`;
 
-module.exports = {
+export = {
   getScripts: () => [copyCodeBlockScript],
-  processNode: (pluginContext, node) => {
+  processNode: (_: any, node: cheerio.Element & DomElement) => {
```
raysonkoh, line 51: Interesting use of intersection types!

raysonkoh (approved):
LGTM

packages/core/src/plugins/codeBlockCopyButtons.ts
tlylt, line 51: one small question: is `pluginContext` marked as `any` here because it's a user defined json and hard to type?

jovyntls: I marked it as `any` because it's not used anywhere in the function. Not sure if `_:any` or `_: PluginContext` would be better?

ong6: why not just use `_:PluginContext` since we're only taking that in anyways

jovyntls: Thanks for the input guys, created a `PluginsContext` class in types.ts - let me know what you think

packages/core/src/plugins/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface PluginContext {
```
raysonkoh, line 1: Nit: can we place this in an existing file instead of creating a new file? Maybe `Plugin.js`?

jovyntls: Just left a [comment](https://github.com/MarkBind/markbind/issues/2114#issuecomment-1413805848) on #2114 related to this - my concern was that putting types in an existing file could create cyclic dependency issues. We shouldn't have cyclic dependencies in this case for plugins, but if we choose to standardise how the types are handled, a `types.ts` may be a solution. What do you think? For now I could leave it in Plugin.js and see how it goes?

tlylt: Perhaps can also consider using the `d.ts` extension? Somehow `types.ts` feels like an anti-pattern to me.

raysonkoh: I think as long as we standardize how we organize the types, it _should_ be fine. I would prefer something more descriptive than `types.ts` though. 

> For now I could leave it in Plugin.js and see how it goes?

Yup! I think you can do this for now. Let's finish up with migration first since it is higher priority and we can come back later to reorganize the types.

jovyntls: Speaking of Plugin.js - should we put this PR on hold until plugin.js is migrated, since the types will be defined there?

tlylt: I think merging this in first is fine, can refactor it later if needed.

tlylt (approved):

Issue #1851, opened 2022-03-24
Alternative syntax for frontmatter


damithc:
Currently, the frontmatter part appears at the top, when a MarkDown file is viewed on GitHub (through GitHub markdown preview).
Not a big problem, but not pretty either.

e.g.,
![image](https://user-images.githubusercontent.com/1673303/159849473-12e09df0-ced1-4716-84dd-99688d331c46.png)

I wonder if supporting an alternative syntax that is ignored by GitHub rendering is useful. e.g.,

```html
<!-- frontmatter>
title: "Python Basics: Error Handling"
layout: python-layout.md
pageNav: 4
</frontmatter -->
```

jonahtanjz:
We can also consider [markdown-it-front-matter](https://github.com/ParkSB/markdown-it-front-matter), which uses three dashes, `---`, to indicate frontmatter.
```
---
  title: "Python Basics: Error Handling"
  layout: python-layout.md
  pageNav: 4
---
```
Although this will not hide the content of the frontmatter, it will show up nicely in a table format.

![image](https://user-images.githubusercontent.com/47470981/159958817-ce6832f2-ea3f-457c-b969-43bc8140e7fa.png)

damithc:
> We can also consider [markdown-it-front-matter](https://github.com/ParkSB/markdown-it-front-matter), which uses three dashes, `---`, to indicate frontmatter.
> 
> ```
> ---
> title: "Python Basics: Error Handling"
> layout: python-layout.md
> pageNav: 4
> ---
> ```

I like it :+1:  Probably better than what I suggested above.

tlylt:
>

> We can also consider [markdown-it-front-matter](https://github.com/ParkSB/markdown-it-front-matter), which uses three dashes, `---`, to indicate frontmatter.
> 
> ```
> ---
> title: "Python Basics: Error Handling"
> layout: python-layout.md
> pageNav: 4
> ---

[Gatsby](https://www.gatsbyjs.com/docs/how-to/routing/adding-markdown-pages/#frontmatter-for-metadata-in-markdown-files) also uses this syntax.

tlylt:
Somehow managed to find the previous discussion on `<frontmatter>` vs `---` when it was first introduced in this [PR](https://github.com/MarkBind/markbind/pull/175#issuecomment-375567474). Not sure if the previous concerns are still valid.

damithc:
[...]

PR #2115, opened 2023-01-29
WIP: Create alternative syntax for frontmatter

EltonGohJH:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves issue #1851 

This is quite a tricky problem to resolve.
Initially, I wanted to write it as a markbind plugin as suggested by @jonahtanjz.  I tired this but it does not work as markbind follow the commonmark spec (correct me if I am wrong). 
If we look at [Commonmark Docs](https://spec.commonmark.org/0.30/#setext-headings ), we can see that `-` this is under setext headings and when I tested it, it indeed made the tag into `h2`. At the entry point of `processNode`, the tag has already been changed to `h2` so I think I have no choice but to write a markdown-it plugin instead (which is at an earlier entrypoint before it is changed to `h2`).<br>
I realised that there are quite a lot of test cases that used `---` syntax. What should I change it to?
For instance,
in `packages/cli/test/functional/test_site_special_tags/index.md` we have 
```
---

There should be no text between this and the next `<hr>` tag in the browser, since it is a `<style>` tag.

<style>

.bigger-text {
  font-size: 10em;
}

</style>

---
```
Moreover, I have no idea how to resolve the potential interaction with the commonmark spec. (Some people may use --- for `h2`). **Do let me know if this is okay before I add new testcases and fixed existing test cases.**
**Testing instructions:**
Replace `<frontmatter>` tags with `---`
This is my first time writing a markdown-it plugin so please do a sanity check that it is correct.

Create alternative syntax for frontmatter

---

- [x] No unrelated changes


tlylt:
> Initially, I wanted to write it as a markbind plugin as suggested by @jonahtanjz. I tired this but it does not work as markbind follow the commonmark spec (correct me if I am wrong).

MarkBind tries to follow CommonMark, but I think certain deviations are acceptable. The following of commonmark also stems from our use of markdown-it, which is commonmark compliant.


> I realised that there are quite a lot of test cases that used --- syntax. What should I change it to?

Do you mean when `---` is used for horizontal rules in test cases? They should not be changed to frontmatter since they are not frontmatter.

> If we look at [Commonmark Docs](https://spec.commonmark.org/0.30/#setext-headings), we can see that - this is under setext headings and when I tested it, it indeed made the tag into h2. At the entry point of processNode, the tag has already been changed to h2 so I think I have no choice but to write a markdown-it plugin instead (which is at an earlier entrypoint before it is changed to h2).

> Moreover, I have no idea how to resolve the potential interaction with the commonmark spec. (Some people may use --- for h2).

Wondering does the referenced https://github.com/ParkSB/markdown-it-front-matter from #1851 help address the above? Perhaps can take inspiration from the implementation there.

PR #2121, opened 2023-01-30
Deprecate TipBox

### LINKED ISSUE #1804: `tip-box` is not documented and can be deprecated
MarkBind currently supports using both `<tip-box>...</tip-box>` and `<box>...</box>` used to make boxes. 
`tip-box` is not documented, but is used in multiple places in our docs. Both support the same attributes and have the same behaviour. I think this is caused by vue-components registering both `box: tipBox` and `tipBox` as components:
	Screenshot 2022-03-06 at 8 21 39 PM

Since `box` is documented and more widely used, we could deprecate `tip-box` by 
* Keeping only `box: tipBox` in `vue-components/src/index.js` 
* Change all usages of `tip-box` to `box`

Alternatively, we could document tip-box and officially support it, but I think it may be confusing and unnecessary to have two different names for the same component.

itsyme:

- [x] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #1804. Changes all appropriate instances of tipbox to box.

**Testing instructions:**

Deprecate TipBox

Both tip-box and box are supported with box being an alias for
tipbox.

This could be confusing for users and developers as the use of
tip-box is not documented.

Let's deprecate tipbox, remove support for the tip-box tag and
replace all instances of tipbox to box.

This prevents confusion between the use of variations of tipbox in
both code and documentation, making any reference to or use of this
this component consistent. Replacing tipbox with box removes the
need for box being an alias for tipbox, deprecating tipbox completely.

---

- [x] No unrelated changes


yucheng11122017 (commented):
I found some other instances of Tip Box which I'm not sure if should be renamed as well. 

packages/vue-components/src/__tests__/TipBox.spec.js 

packages/vue-components/src/utils/utils.js
Comment ```// Used in the TipBox component to classify the different styles used by bootstrap from the user input.```

itsyme:
> I found some other instances of Tip Box which I'm not sure if should be renamed as well.
> 
> packages/vue-components/src/**tests**/TipBox.spec.js
> 
> packages/vue-components/src/utils/utils.js Comment `// Used in the TipBox component to classify the different styles used by bootstrap from the user input.`

Hi @yucheng11122017! Thanks for looking through my PR, I intentionally left these as I felt they were referring to the TipBox component which is still currently named as `TipBox.vue` under `vue-components`. I am currently still unsure about whether to change this and would be open to discussion about this!

lhw-1:
> I intentionally left these as I felt they were referring to the TipBox component which is still currently named as `TipBox.vue` under `vue-components`. I am currently still unsure about whether to change this and would be open to discussion about this!

If we're going by the original intent of fully deprecating the `tip-box` component as the `box` component, then I think we can rename these as well. Referring to all instances of `tip-box` as `box` in both development and documentation should make things easier!

Of course, the other concern would be backwards compatibility, so while we can make `Box` as the main component, perhaps any usage of `tip-box` should still be supported (as an alias of `box` or otherwise)?

jovyntls:
> Of course, the other concern would be backwards compatibility, so while we can make box as the main component, perhaps any usage of tip-box should still be supported (as an alias of box or otherwise)?

We shouldn't need to support any usage as an alias since it's been undocumented since quite a while back (see [this comment](https://github.com/MarkBind/markbind/issues/1804#issuecomment-1059990537)). Removing it altogether is fine :)

As for renaming `tipBox.vue` -> `Box.vue`, I'm fine with either way - I originally thought using Box.vue is a common name and could cause naming conflicts in the future but upon further thought I don't think this is so (and even if it does it shouldn't be too big of a problem).

itsyme:
Alright! I will rename all variants of `tipbox` to `box` and rename the component in the next commit!

jovyntls (approved):
LGTM! 🎊

itsyme:
Hi @jovyntls I have edited the commit message! Let me know if you would like any additions or changes to it.

tlylt:
> Hi @jovyntls I have edited the commit message! Let me know if you would like any additions or changes to it.

Hi @itsyme, I think @jovyntls (and I too) would think that a more detailed commit message following the following structure (as noted in the link in https://github.com/MarkBind/markbind/pull/2121#pullrequestreview-1278623774) would be more suitable, could you update the commit message accordinly to it?

```
{current situation} -- use present tense

{why it needs to change}

{what is being done about it} -- use imperative mood

{why it is done that way}

{any other relevant info}
```

itsyme:
Hi @tlylt! Thank you for the clarification! I interpreted it as a more detailed commit header. I will make the relevant changes now.

jovyntls:
Thanks for making the changes @itsyme 😄 
I see that you're following the `scope: change` [commit message guidelines](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message) for the commit header. Can we go back to the non-scope format (e.g. `Deprecate TipBox`) for the commit header? This is to be consistent with the rest of the commit messages and because the scope of `tip-box tag` and `TipBox` is relatively small. The differentiation between the tag and component can be detailed in the message body, if needed

jovyntls (approved):
LGTM, thanks for all the changes @itsyme 😄

PR #2123, opened 2023-01-30
Update bug-report template

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Consolidates the "bug description", "steps to reproduce" and "actual behavior" into one place, which may work better as they tend to be mentioned at the same time.

nil

**Testing instructions:**
Preview [here](https://github.com/tlylt/markbind/blob/update-form/.github/ISSUE_TEMPLATE/bug-report.yml)

Update bug-report template

---

- [x] No unrelated changes

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -42,9 +42,10 @@ body:
   attributes:
     label:  Describe the bug and the steps to reproduce it
     description: |
-      A clear and concise description of the bug.
-      - What actually happened? Please include the actual, raw output. You are encouraged to submit a PR that reproduces this in `packages/cli/test/functional/test_site/bugs/index.md`.
-      - Please also include the actual source code causing the issue.
+      Please include the following, if availble:
+      - actual, raw output
+      - source code causing the issue
+        - you are encouraged (though not required) to submit a PR that reproduces this in `packages/cli/test/functional/test_site/bugs/index.md`.
   
       Tip: You can attach images or log files by clicking this area to highlight it and then dragging files in.
   validations:
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -40,32 +40,20 @@ body:
 - type: textarea
   id: Description
   attributes:
-    label:  What did you do? Describe the bug
+    label:  Describe the bug and the steps to reproduce it
     description: |
       A clear and concise description of the bug.
-
+      - What actually happened? Please include the actual, raw output. You are encouraged to submit a PR that reproduces this in `packages/cli/test/functional/test_site/bugs/index.md`.
+      - Please also include the actual source code causing the issue.
+  
       Tip: You can attach images or log files by clicking this area to highlight it and then dragging files in.
   validations:
     required: true
-- type: textarea
-  id: RepoSteps
-  attributes:
-    label:  Steps to reproduce
-    description: Please include the actual source code causing the issue.
-  validations:
-    required: false
 - type: textarea
   id: expected
   attributes:
     label:  Expected behavior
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -40,34 +40,22 @@ body:
 - type: textarea
   id: Description
   attributes:
-    label:  What did you do? Describe the bug
+    label:  Describe the bug and the steps to reproduce it
     description: |
       A clear and concise description of the bug.
-
+      - What actually happened? Please include the actual, raw output. You are encouraged to submit a PR that reproduces this in `packages/cli/test/functional/test_site/bugs/index.md`.
```
jovyntls, line 46: > You are encouraged to submit a PR that reproduces this in `packages/cli/test/functional/test_site/bugs/index.md`

Should we consider removing this? I don't recall anyone doing so recently and the steps below ask for code for reproduction, though I can see how the PR might be easier for investigation

jovyntls (approved):
LGTM! This template is much shorter 👍

.github/ISSUE_TEMPLATE/bug-report.yml
tlylt: I agree with your observation =>
- a PR reproduction might aid in the investigation
- but people don't seem to be doing it now, most probably because it's too troublesome

Do you think it will be good to improve this aspect by writing up a note in our DG (section: how to create bug report)?
- will be a separate issue

jovyntls: I'm think we shouldn't have need a "how to create bug reports" section in the DG since the purpose of the issue templates is to teach/enforce creating bug reports in a constructive way, and we should minimise the friction of creating a bug report (i.e. shouldn't need to read DG). 

other considerations:
- most of the time since code snippets are usually sufficient for a minimal reproduction. 
- too many test PRs could clog up the PRs tab
- complex issues might need a PR to reproduce though - is there any value to having the reproduction PR in our Markbind repo vs inspecting code on the bug reporter's repo directly?

damithc: Also keep in mind that bug reporters may be end users who are not necessarily programmers.

In general, use a philosophy that the most important thing is to get the user to initiate a communication with us (i.e., post _something_) because our assumption is that every such communication benefits us (i.e., it can potentially help us improve the product) more than the person posting the issue. A partial bug report is better than no bug report. If we can figure out the missing information ourselves, we do that. But if we can't, we can always respond with a request for more information.

tlylt: I see where you're coming from.

I feel there's value in reproduction PR because I do see that being suggested in some of the big OSS projects out there, but obviously, this takes effort on the bug reporter's end.

> most of the time since code snippets are usually sufficient for a minimal reproduction.
> complex issues might need a PR to reproduce though - is there any value to having the reproduction PR in our Markbind repo vs inspecting code on the bug reporter's repo directly?

Indeed only complex cases require something more than code snippets. In some cases, bug reporters may not have a public repo.

> too many test PRs could clog up the PRs tab

This might be ok as devs can filter and see the latest list of PRs by date (similar to what we can do for issues).

My understanding from this convo so far is that it's not necessary most of the time and we should probably not impose upon devs to do this reproduction.

As currently the instruction is stated as an "encouragement" and it's not prominent, do you think it's fine to keep it that way for advanced devs who need it? What do you think?

tlylt: Update: I have adjusted the wording to make it clear that it's not necessary to reproduce the bug via PR.
image

jovyntls: Looks good! Thanks @tlylt :)

PR #2125, opened 2023-01-31
Fix bug where dash cannot be highlighted in code blocks

yucheng11122017:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Trying to highlight code which contained a '-' did not work since the highlight rules were parsed according to '-'. 
Eg. {highlight-lines="1['block-level']"}
Change to parse using regex in which '-' cannot be within a quotation. 

nil

**Testing instructions:**
````
 ```{highlight-lines="1['block-level']"}
<markdown> elements allow block-level markdown without needing a leading newline.
Hence, the contained markdown should be parsed and output as is, without any parsing errors.
</invalidhtml>
 ```
````

```Fix bug to allow highlighting of code with '-'```

---

- [X] No unrelated changes

### CODE REVIEW

File: packages/core/src/lib/markdown-it/highlight/HighlightRule.js [hunk 1]
```diff
@@ -1,6 +1,8 @@
 const { HighlightRuleComponent } = require('./HighlightRuleComponent');
 const { splitCodeAndIndentation } = require('./helper');
 
+const HIGHLIGHT_LINES_RULES_DELIMITER_REGEX = /(?<!'.*)-|-(?!.*')/;
```
lhw-1, line 4: Since the other regex-based highlighting rules are defined in `HighlightRuleComponent.js` as seen [here](https://github.com/MarkBind/markbind/blob/master/packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js#L3-L6), would it make more sense to declare it there instead of here? Just a suggestion!

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
```diff
@@ -10,7 +12,7 @@ class HighlightRule {
   }
 
   static parseRule(ruleString, lineOffset, lines) {
-    const components = ruleString.split('-')
+    const components = ruleString.split(HIGHLIGHT_LINES_RULES_DELIMITER_REGEX)
```
lhw-1, line 15: As a follow-up to my comment above, if it's better to move the regex to `HighlightRuleComponent.js`, perhaps we can incorporate the splitting part into the `HighlightRuleComponent.parseRuleComponent` method [here](https://github.com/MarkBind/markbind/blob/master/packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js#L24-L80) as well? That might make the process more straightforward as parsing the string once using the `HighlightRuleComponent`!

lhw-1 (commented):
Left some comments on the code! The changes seem to resolve the bug on my end as well :+1: 

Not sure if relevant, but as I was testing the code, I realized that there may be cases where the user wants to highlight square brackets following commas in certain situations, e.g.
````
```{highlight-lines="1['[4, 5, 6], and [7, 8, 9]']"}
<markdown> I have 3 arrays: [1, 2, 3], [4, 5, 6], and [7, 8, 9].
</invalidhtml>
```
````

Expected: Only "[4, 5, 6], and [7, 8, 9]" part is highlighted.
Actual: The entire line is highlighted.

Since this Pull Request is not tied to a specific issue, would it be worth it to expand the scope of this PR if this bug can be solved concurrently? Otherwise, we can alternatively just open a new issue for this!

File: packages/core/src/lib/markdown-it/highlight/HighlightRule.js (hunk 1 above)
Comment (yucheng11122017) [reply packages/core/src/lib/markdown-it/highlight/HighlightRule.js:4]: Hmmm I'm not too sure about this. Because the ```HighlightRuleComponent``` is about parsing the components of the rules. So the splitting of the components should be done before the components are made.

Comment (lhw-1) [reply packages/core/src/lib/markdown-it/highlight/HighlightRule.js:4]: Alright, that makes sense. In that case I have no further issues :+1:

**Review by raysonkoh — APPROVED** [review]
Thanks @yucheng11122017 for the PR! Seems to be working to my end too 👍 . 

I'm wondering if there's a better way to handle highlighting substrings 🤔, since this PR doesn't address the example from @lhw-1. I think generally the other highlighting syntax is pretty structured, so parsing it would be fine. But parsing substrings could be tricky since the substring may contain reserved keywords etc. This can be addressed in a separate issue.

**Comment by yucheng11122017** [comment]
> I'm wondering if there's a better way to handle highlighting substrings 🤔, since this PR doesn't address the example from @lhw-1. I think generally the other highlighting syntax is pretty structured, so parsing it would be fine. But parsing substrings could be tricky since the substring may contain reserved keywords etc. This can be addressed in a separate issue.

Yeah it's true. Trying to write a regex in which it will not parse if these characters are within quotes. But its a bit harder especially since you might get an example where 
```
1['some text'], 2['other text']
```
in which case the "," in the middle might be dropped.
Am currently working on it and will open a PR when its done :)

PR #2133, opened 2023-02-05
Migrate Plugin and PluginManager to TypeScript

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Part of #1913 

**Overview of changes:**
Migrate Plugin and PluginManager to TypeScript
Also removes `plugins/types.ts` and moves the types to `Plugin.js`

(rebase commit)

---

- [x] No unrelated changes


packages/core/src/plugins/PluginManager.ts
```diff
@@ -55,8 +73,6 @@ class PluginManager {
    * Load all plugins of the site
    */
   _collectPlugins() {
-    module.paths.push(path.join(this.config.rootPath, 'node_modules'));
```
jovyntls, line 58: Removed this line because the `module` object is not compatible with ES6 imports. This is manually implemented below around line 146.

packages/core/src/plugins/PluginManager.ts
```diff
@@ -127,7 +143,18 @@ class PluginManager {
...
 
-    return require.resolve(pluginName);
+    // Check the environment's node_modules folders
+    try {
+      const resolvedPluginPath = require.resolve(pluginName);
+      return resolvedPluginPath;
+    } catch (err) {
+      // Also check the node_modules of the project
+      if (_.isError(err) && (err as NodeJS.ErrnoException).code === 'MODULE_NOT_FOUND') {
+        return require.resolve(pluginName, { paths: [path.join(projectRootPath, 'node_modules')] });
```
jovyntls, line 153: This addition is a manual implementation of `module.paths.push(path.join(this.config.rootPath, 'node_modules'));`, which adds the root path's `node_modules` to the search path. 

Since that line was removed, this section manually searches the root path's `node_modules` if `require.resolve` via the usual search paths does not find the plugin module. It should be the same order of checks as before.

raysonkoh: Would there be any issues with substituting this whole block with
```
return require.resolve(pluginName, { paths: [path.join(projectRootPath, 'node_modules')] });
```

? It seems to be working on my end.

packages/core/package.json
```diff
@@ -82,6 +82,7 @@
     "@types/jest": "^27.4.1",
     "@types/js-beautify": "^1.13.3",
     "@types/lodash": "^4.14.181",
+    "@types/node": "^18.11.19",
```
raysonkoh, line 85: Is this necessary?

raysonkoh (commented):
Thanks for the PR, the TS migration is going at a good pace! I have added a few comments.

packages/core/src/plugins/PluginManager.ts
jovyntls: That works on my end too! I added the isError and `MODULE_NOT_FOUND` check to differentiate between the case where the module is not found (and so we need to search the next path) vs the case where there is some other error (which we probably want to throw?)

I think there are no issues with substituting if we don't need to throw errors arising from the first `require.resolve`. What do you think? @raysonkoh

raysonkoh: > I think there are no issues with substituting if we don't need to throw errors arising from the first require.resolve

I think it's fine as long as the error (if any) when trying to resolve in the project root is thrown. 

Based on the [docs](https://nodejs.org/api/modules.html#loading-from-node_modules-folders):
> Node.js starts at the directory of the current module, and adds /node_modules, and attempts to load the module from that location. ... If it is not found there, then it moves to the parent directory, and so on, until the root of the file system is reached.

So it seems that it will still search the local `node_modules` folder first before searching the project root's `node_modules`. Would it still work if you remove the `paths` parameter? 🤔

raysonkoh: I see, this seems like too deep of a rabbit hole to dive into 😅. I'm fine with either substituting it, or leaving it in this `try catch` form but with a comment as to why it was done this way.

raysonkoh (approved):
LGTM

packages/core/src/plugins/PluginManager.ts
jovyntls: Added some documenting comments! Thanks @raysonkoh

PR #2138, opened 2023-02-06
Fix parsing for code highlight rules

yucheng11122017:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
As pointed out by @lhw-1 in #2125, the code highlight is unable to deal with edge cases like 
````
```{highlight-lines="1['[4, 5, 6], and [7, 8, 9]']"}
<markdown> I have 3 arrays: [1, 2, 3], [4, 5, 6], and [7, 8, 9].
</invalidhtml>
```
````

Changed the splitting of highlight rules to only split when special characters (such as ' and -) are not within quotes. 
Changed the order in which parseRuleComponent parses to deal with linepart first. 

nil

**Testing instructions:**
Some edge cases:
````
```{highlight-lines="1['[4, 5, 6], and [7, 8, 9]']"}
<markdown> I have 3 arrays: [1, 2, 3], [4, 5, 6], and [7, 8, 9].
</invalidhtml>
```
````

````
```java {.line-numbers highlight-lines="1['1[:],3[\'Inventory\'],4[\'It\'s designed\'],5,6[8:18],8[0::2],12[:]-14,16-18,20[12:]-22,24[1::]-26']"}
1[:],3['Inventory'],4['It's designed'],5,6[8:18],8[0::2],12[:]-14,16-18,20[12:]-22,24[1::]-26
```
````

```Fix parsing of code highlight rules```

---

- [X] No unrelated changes

### CODE REVIEW

File: packages/core/src/lib/markdown-it/highlight/HighlightRule.js [hunk 1]
```diff
@@ -11,8 +9,36 @@ class HighlightRule {
...
+    return highlightLines
+      .map(ruleStr => HighlightRule.parseRule(ruleStr, lineOffset, strArray))
+      .filter(rule => rule); // discards invalid rules
+  }
+
+  static splitByChar(allRules, splitter) {
+    const highlightRules = [];
+    let isWithinQuote = false;
+    let currentPosition = 0;
+    for (let i = 0; i < allRules.length; i += 1) {
```
lhw-1, line 24: The convention for `for` loops in MarkBind seem to be using `i++` instead of `i += 1` when incrementing by 1 (as far as I can see)!

lhw-1, line 24: On that note, I think we can consider using a `split` method instead of a `for` loop. For instance, we can initially split the rules based on unescaped quotation mark using something like `.split(\(?<!\\)\'\)`, and then further split the substrings using the `splitter` given to the `splitByChar` method (not sure if this will fully remove the need for `currentPosition` but looks like it). What do you think @yucheng11122017?

lhw-1 (commented):
Just some nits on the code!

On a separate note (yet again, sorry), if i fail to close the quotation mark e.g.

````
```{highlight-lines="1['I have]"}
<markdown> I have 3 arrays: [1, 2, 3], [4, 5, 6], and [7, 8, 9].
</invalidhtml>
```
````
It seems to highlight the entire line:

Screenshot from 2023-02-07 09-54-55

Since this is an erroneous use case by the user, I wonder if we should either highlight it in the documentation or have it throw an error...?

**Review by jovyntls — COMMENTED** [review]
Nice work catching the edge cases!

File: packages/core/src/lib/markdown-it/highlight/HighlightRule.js (hunk 1 above)
Comment (yucheng11122017) [reply packages/core/src/lib/markdown-it/highlight/HighlightRule.js:24]: Hmmm I'm not too sure how that will work. The point of this function is to split by the splitter except for when the splitter is within a quote Eg. ```','```
If I split using the unescaped quotes, then split by the splitter won't it give broken outputs?
Eg. ```['hi'], 1['sd']``` will give ```'[', 'hi', ']', '1[', 'sd', ']'```

Am I interpreting you correctly?

**Comment by yucheng11122017** [comment]
> On a separate note (yet again, sorry), if i fail to close the quotation mark e.g.
> 
> ```
> ```{highlight-lines="1['I have]"}
> <markdown> I have 3 arrays: [1, 2, 3], [4, 5, 6], and [7, 8, 9].
> </invalidhtml>
> ```
> ```
> 
> It seems to highlight the entire line:
> 
> Screenshot from 2023-02-07 09-54-55
> 
> Since this is an erroneous use case by the user, I wonder if we should either highlight it in the documentation or have it throw an error...?

Hmm that is a good point. I don't think there is any error handling right now in the highlightLines, it will just not highlight if it is wrong or in this case, it just got interpreted as a line instead. I will look into it thanks @lhw-1

File: packages/core/src/lib/markdown-it/highlight/HighlightRule.js (hunk 1 above)
Comment (yucheng11122017) [reply packages/core/src/lib/markdown-it/highlight/HighlightRule.js:24]: It seems like its mostly += 1 when I ctrl F and Eslint is giving me an error when ```i++``` is used. Shouldn't it be += 1?

**Comment by yucheng11122017** [comment]
@lhw-1 I'll probably open a new PR on error checking for highlight lines. Thanks for the suggestion!

File: packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.js [hunk 2]
```diff
@@ -22,6 +22,24 @@ class HighlightRuleComponent {
   }
 
   static parseRuleComponent(compString, lineNumberOffset, lines) {
+    // Match line-part syntax
+    const linepartMatch = compString.match(LINEPART_REGEX);
+    if (linepartMatch) {
+      // There are four capturing groups: [full match, line number, quote type, line part]
+      const groups = linepartMatch.slice(1); // discard full match
```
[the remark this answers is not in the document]
yucheng11122017: Fixed in [2d775f0](https://github.com/MarkBind/markbind/pull/2138/commits/2d775f0d0171b0e2156fa32ca6ea3df41ec55f51)!

packages/core/src/lib/markdown-it/highlight/HighlightRule.js
lhw-1: Given that the regex I mentioned above will not consider `\'`, I was thinking that we could use this to filter out the outputs that are not within the quotation marks, and only split the parts that were inside the quotation marks by the splitter. My reasoning was that if we only split by `'` then the split should alternate between the segments that were inside the quotation marks and segments that were outside the quotation marks. E.g. `['hi'], 1['sd']` will result in `'[', 'hi', '], 1[', 'sd', ']'`. We can then split the alternating substrings of `'hi'` and `'sd'` using the `splitter`, etc.

yucheng11122017: Hmm but the point of the function is to split the function according to the splitter, not the quotation though. Like it doesn't need the split according the quotation mark

lhw-1: Ah, I see! Piecing it back together would indeed be a problem. In that case the current implementation should be better than the suggestion :sweat_smile:

yucheng11122017: Alright thanks for all the suggestions though! @lhw-1

jovyntls (approved):
LGTM! Would be good to document (code comments) the `splitByChar` function a little more since there's some potentially confusing logic involved in the computations there.

yucheng11122017:
> Would be good to document (code comments) the `splitByChar` function a little more since there's some potentially confusing logic involved in the computations there.

Fixed in [aee07d5](https://github.com/MarkBind/markbind/pull/2138/commits/aee07d5b04fd6d70535d494298826d95f0be6191)!

ong6 (approved):
LGTM

tlylt (changes requested):
In the git record, an empty new file `packages/vue-components/src/ScrollToTopButton.vue` has been added. May I know what is that about?

yucheng11122017:
@tlylt Fixed in [4999280](https://github.com/MarkBind/markbind/pull/2138/commits/499928008130a1f49e09f2978ec6ff9a2b6cc6f8)

tlylt (approved):

PR #2143, opened 2023-02-07
Migrate template to TypeScript

lhw-1:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [ ] Feature addition or enhancement
- [X] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

Part of #1913.

Migrates `packages/core/template/template.js` to TypeScript following the workflow specified [here](https://markbind.org/devdocs/devGuide/development/migratingToTypeScript.html).

This PR also builds on the changes made in PR #2148.

**Overview of changes:**

**Anything you'd like to highlight/discuss:**

**Testing instructions:**

**Proposed commit message: (wrap lines at 72 characters)**

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No unrelated changes

PR #2148, opened 2023-02-08
Convert Template function to class

### LINKED ISSUE #2147: Convert `template.js` to a class

### Any related issues?

#2143

Code Quality

Currently, [`packages/core/template/template.js`](https://github.com/MarkBind/markbind/blob/master/packages/core/template/template.js) exports a function `Template`, which is then used in [`packages/core/src/Site/index.js`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L182) to start the initialization process when `markbind init` is called in the CLI.

However, the way the code is currently structured could be improved by converting the `Template` function into a class instead, which would allow us to have the `init` method be part of the `Template` class rather than have it be declared as `Template.prototype.init()` as it currently is being done.

This also simplifies the current migration process for `template.js` into TypeScript in #2143.

As a reference, the `Template` function was never a class to begin with, and was introduced in [this commit](https://github.com/MarkBind/markbind/commit/9e50fc06a5ac3b9cef920ba08979ec40df359c0a#diff-0e37c9b8a9badb5fb366b1b9237a0c66020ebf2679f047d6fcb136776f4a54da) as part of #935, meaning that there was never an issue where we had to convert it from a class to a function.

Convert the current `Template` function into a class, which would not change the way it is currently being called.

Currently, there is a `prototype.init` function that implements the `Template` function, which causes the calling statement to be:

```js
return await new Template(rootPath, templatePath).init();
```

as seen [here](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L182).

If the `Template` function is instead converted to a class (and the `init` function into a method of the class), then the way it is called will not change, meaning that it should not cause external issues - it will only affect how the code is organized within `template.js`. Moreover, this exact `Template` function is only used within [`packages/core/src/Site/index.js`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Site/index.js#L182) and nowhere else in the codebase currently, and so there should not be any unintended effects.

Regardless, any review would be welcome so as to make sure that the change is standalone and doesn't result in unintended consequences!

An example of what the conversion would look like can be seen in the PR addressing this issue: #2148.

_No response_

### Additional context

_No response_

lhw-1:

- [ ] Documentation update
- [ ] Bug fix
- [X] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Resolves #2147.

Converts the `Template` function inside [`packages/core/template/template.js`](https://github.com/MarkBind/markbind/blob/master/packages/core/template/template.js) into a class.

This also simplifies the migration process for `template.js` into TypeScript in #2143.

**Testing instructions:**

Convert Template function to class

---

- [X] No unrelated changes


packages/core/template/template.js
```diff
@@ -5,46 +5,50 @@ const fsUtil = require('../src/utils/fsUtil');
...
+  constructor(rootPath, templatePath) {
+    this.root = rootPath;
+    this.template = templatePath;
+  }
 
-function validateTemplateFromPath(templatePath) {
-  for (let i = 0; i < requiredFiles.length; i += 1) {
-    const requiredFile = requiredFiles[i];
-    const requiredFilePath = path.join(templatePath, requiredFile);
+  static validateTemplateFromPath(templatePath) {
```
tlylt, line 14: Should this be static?

packages/core/template/template.js
```diff
@@ -5,46 +5,50 @@ const fsUtil = require('../src/utils/fsUtil');
...
+        return false;
+      }
     }
+
+    return true;
   }
 
-  return true;
-}
+  static generateSiteWithTemplate(rootPath, templatePath) {
```
tlylt, line 27: Should this be static?

packages/core/template/template.js
```diff
@@ -5,46 +5,50 @@ const fsUtil = require('../src/utils/fsUtil');
...
-  return new Promise((resolve, reject) => {
-    fs.access(rootPath)
-      .catch(() => fs.mkdirSync(rootPath))
-      .then(() => fsUtil.copySyncWithOptions(templatePath, rootPath, { overwrite: false }))
-      .then(resolve)
-      .catch(reject);
-  });
-}
+  init() {
+    const templatePath = path.join(__dirname, this.template);
```
tlylt, line 38: How about moving this up to constructor and refactor accordingly?

tlylt (commented):
Interesting changes proposed! 

Curious if there are opportunities here to improve the code further. For example, I wonder whether `template.js` really needs to be in this folder, or perhaps it is more suited in the `Site` folder. What do you think?

packages/core/template/template.js
lhw-1: Ah right, I'll change them back (they were changed to `static` while I was trying to debug some cases where the functions weren't being referenced properly). Same for the other method.

Thanks for pointing it out!

packages/core/template/template.js
lhw-1: Good suggestion - thank you!

packages/core/template/template.js
lhw-1: Looks like there was a reason why I didn't change them back... Having them not be `static` methods will cause the Eslint to complain, as detailed [in the Eslint docs](https://eslint.org/docs/latest/rules/class-methods-use-this), and shown in the screenshot below:

Screenshot from 2023-02-09 23-22-54

The reasoning in the Eslint docs seem to be that as long as a method does not make use of the `this` keyword to reference the object, there is no reason why it should be an instance method.

packages/core/template/template.js
lhw-1: Updated in [38c7676](https://github.com/MarkBind/markbind/pull/2148/commits/38c76764fed6da189b9c3cc5a76ef64d8f2a40b0)!

lhw-1:
> Curious if there are opportunities here to improve the code further. For example, I wonder whether `template.js` really needs to be in this folder, or perhaps it is more suited in the `Site` folder. What do you think?

On one hand, I think keeping `template.js` in the `template` folder makes sense since this is acting as a sort of "interface" for other files to create and initialize a template (though only `Site/index.js` is doing that right now). 

But on the other hand, I also see the benefit in moving it over to the `Site` folder to (1) keep the JavaScript / TypeScript files all in once place (within `src`), and (2) since `Site` contains the files for initializing a MarkBind site / converting to a MarkBind site, it makes sense that we should have the `Template` class within this folder. And of course, the only file using the `Template` function / class right now is `Site/index.js`, which is one more point towards moving `template.js` to the `Site` folder.

Yes, I think moving it over may make more sense than keeping it where it currently is.

packages/core/template/template.js
tlylt: And my suggestion follows: how about making the parameters instance variables?

lhw-1: Updated according to the suggestion in [ffa06c4](https://github.com/MarkBind/markbind/pull/2148/commits/ffa06c4634c60a65c7aebc0d9adf46b6152df31c)!

packages/core/template/template.js
lhw-1: Same as above - updated according to the suggestion in [ffa06c4](https://github.com/MarkBind/markbind/pull/2148/commits/ffa06c4634c60a65c7aebc0d9adf46b6152df31c)!

lhw-1:
**Update in [931271a](https://github.com/MarkBind/markbind/pull/2148/commits/931271a8a1f79070798d3df4897cdece793f4998)**:

`template.js` has been moved from `packages/core/template/` to `packages/core/src/Site/` as per the suggestion from @tlylt & the following discussion.

As the template files reside in another directory outside `packages/core/src/Site`, an additional correction to the path needed to be declared [here](https://github.com/MarkBind/markbind/pull/2148/files#diff-f9831becb42de391e9c938d4a178d548568313b1ccfec08ccbc3708dfe05a977R8). I am wondering if it would be a good idea to additionally move all the template files into the `packages/core/src/Site` folder to make it more direct, or if it would make more sense to keep the template in a separate folder as is currently being done. I would appreciate thoughts on this matter as I am uncertain if this makes sense :+1:

PR #2160, opened 2023-02-12
Migrate includePanelProcessor to TypeScript

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Part of #1913 

**Overview of changes:**
Migrate includePanelProcessor to TS

N/A

**Testing instructions:**
N/A

N/A - rebase commit

---

- [x] No unrelated changes


packages/core/src/html/includePanelProcessor.ts
```diff
@@ -110,11 +121,12 @@ function processPanelSrc(node, context, pageSources, config) {
...
     filePath,
     actualFilePath,
-  } = _getSrcFlagsAndFilePaths(node, config);
+    // We can typecast here as we have checked for src above.
+  } = _getSrcFlagsAndFilePaths(node as DomElement & { attribs: { src: string } }, config);
 
   const fileExistsNode = _getFileExistsNode(node, context, actualFilePath, pageSources);
   if (fileExistsNode) {
-    return fileExistsNode;
+    return context;
```
jovyntls, line 129: This change in logic is because 
- Doing so allows `processNode` to return a `Context` object. Otherwise, it would return `Context | Document | DomElement` and this is not desirable because all 3 of these types are very different.
- The returned `fileExistsNode` is not used anywhere - it triggers an early return later on

raysonkoh: I think standardizing the return type of `processNode` is great! But I want to check

> it triggers an early return later on

Which part of the code can I find this behavior?

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -141,33 +153,38 @@ function processPanelSrc(node, context, pageSources, config) {
...
     }
   });
 
-  delete node.attribs.boilerplate;
-  delete node.attribs.src;
-  delete node.attribs.inline;
-  delete node.attribs.trim;
-  delete node.attribs.optional;
-  delete node.attribs.omitFrontmatter;
+  delete nodeAttribs.boilerplate;
```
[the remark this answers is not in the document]
jovyntls: Yup! Just did a check, it's passed by reference and gets deleted as intended

packages/core/src/html/includePanelProcessor.ts
jovyntls: The final return (in the case where we return `fileExistsNode`) is [here](https://github.com/jovyntls/markbind/blob/1dcef01f393e79aa3dedfede63de3db94ef43835/packages/core/src/html/NodeProcessor.ts#L316) - essentially it gets passed to the `traverse` function (recursively), and then nothing happens because it gets returned early as a text node. The returned node is never processed - the red error message we see when an `include` fails is from the original node's children, not the returned node.

raysonkoh: Gotcha! Thanks for the explanation. In that case, can we update the `_getFileExistsNode` function to return a boolean instead of the error node since we are not doing anything with it? And also maybe rename it to `_checkFileExists` to be clearer.

jovyntls: Good point, will refactor it to a boolean checks function

raysonkoh (approved):
LGTM

PR #2165, opened 2023-02-13
Clean up default init template

lhw-1:

- [ ] Documentation update
- [ ] Bug fix
- [X] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Cleans up the default template for `markbind init`, consisting of minor changes in wording, etc.

As this PR is not tied to a specific issue, any suggestions on improving the current template (in terms of pure cosmetics without changing the functionalities) are welcome.

**Testing instructions:**

Clean up default init template

---

- [X] No unrelated changes


lhw-1:
Update in [b7c8cbd](https://github.com/MarkBind/markbind/pull/2165/commits/b7c8cbdac91ae4f0af1700dea08808333beed275):

The placeholder text in the `topicx.md` files were changed to be:

```
# Topic XX

> This is a placeholder page - more content to be added.
```

Some of the whitespaces in `404.md` were removed to be standardized with the rest of the `.md` files. Additionally, the text was changed from "File not found ..." to "Page not found ...".

itsyme:
Hi @lhw-1! Thanks for the PR! Just a small nit in the home page under "Heading 3". I noticed that for the third panel: "Expanded panel", the panel is meant to be expanded on load, however, it seems to be missing the `expanded` property in `index.md`. Perhaps you could include that to make the panel expand on load as intended!

lhw-1:
Thanks for looking through the changes @itsyme!

> Just a small nit in the home page under "Heading 3". I noticed that for the third panel: "Expanded panel", the panel is meant to be expanded on load, however, it seems to be missing the `expanded` property in `index.md`. Perhaps you could include that to make the panel expand on load as intended!

This is the current output for `index.md` under "Heading 3":

Screenshot from 2023-02-17 14-56-36

Essentially, the minimized panel at the top is set to change its header back to "Expanded panel" once it is no longer minimized (by clicking on it at least once):

Screenshot from 2023-02-17 14-56-53

Additionally, the expanded panel at the bottom will become a minimized panel as well after clicking the "X" button:

Screenshot from 2023-02-17 14-57-14

So I believe that the heading "Expanded panel" refers not to having the panel already expanded on load, but to differentiate between a minimized and an expanded panel - I don't think including the property for this bottom panel is a necessity, since it could potentially further obfuscate the meaning of "expanded" for new users (if it already isn't doing so...).

Regardless, what you've pointed out is still a good example to showcase for panels (especially to differentiate between these two meanings of "expanded"), so it may be worth including another panel at the bottom that showcases this property specifically! Though maybe this could be "out of scope" since it would be adding new features to the `index.md`. 

On that note, it might also be worth looking into improving the template to be more comprehensive to a new user, instead of having filler texts. With reference to the above example, some brief explanation of the features could be more handy than the standard `Lorem ipsum` :smile: .

packages/core/template/default/404.md
```diff
@@ -1,7 +1,7 @@
 <frontmatter>
-    title: Page not found
-    layout: 404.md
+  title: Page not found
+  layout: 404.md
 </frontmatter>
 
 -><p style="font-size: 10rem">404</p><-
--><p style="font-size: 1.5rem">File not found<br>Click <a href="/">here</a> to go back to the home page.</p><-
```
tlylt, line 7: I think this is fine. Is there any reason that you changing it? I'm guessing the reason why @EltonGohJH initially have this wording is because it mirrors the existing [github pages 404 page](https://httpwg.org/specs/hjkhk.image)

tlylt:
> > Just a small nit in the home page under "Heading 3". I noticed that for the third panel: "Expanded panel", the panel is meant to be expanded on load, however, it seems to be missing the `expanded` property in `index.md`. Perhaps you could include that to make the panel expand on load as intended!
> 
> Essentially, the minimized panel at the top is set to change its header back to "Expanded panel" once it is no longer minimized (by clicking on it at least once):
> 
> Screenshot from 2023-02-17 14-56-53
> 
> Additionally, the expanded panel at the bottom will become a minimized panel as well after clicking the "X" button:
> 
> Screenshot from 2023-02-17 14-57-14
> 
> So I believe that the heading "Expanded panel" refers not to having the panel already expanded on load, but to differentiate between a minimized and an expanded panel - I don't think including the property for this bottom panel is a necessity, since it could potentially further obfuscate the meaning of "expanded" for new users (if it already isn't doing so...).
> 
> Regardless, what you've pointed out is still a good example to showcase for panels (especially to differentiate between these two meanings of "expanded"), so it may be worth including another panel at the bottom that showcases this property specifically!

If you add the expanded property in, does it look better/make the example clearer? I think this is a good point - I re-read the conversation above at least twice to understand what this is all about, it's definitely not the clearest example that we can give. Would be great if we can modify this part slightly to make the example clear.

FYI the deployed output of the standard template can be viewed [here](https://markbind-init-typical.netlify.app/).

packages/core/template/default/404.md
lhw-1: I see. Initially I changed the wording for "file not found" to "page not found" (both seem to be widely used for 404 pages) as the wording felt slightly awkward to me (since the deployed MarkBind site would be dealing with pages). If we want to mirror the existing github pages, then we can keep the "file not found" but add in the "The site configured at this address does not contain the requested file. If this is your site, make sure that the filename case matches the URL." Does this sound alright?

lhw-1:
> If you add the expanded property in, does it look better/make the example clearer? I think this is a good point - I re-read the conversation above at least twice to understand what this is all about, it's definitely not the clearest example that we can give. Would be great if we can modify this part slightly to make the example clear.

Right now, the wording we are using in the template for "Expanded panel" vs "Minimized panel" is referring to the panel header, and not the panel body itself, which is what the `expanded` property is for. So I think that we can add in a third panel to showcase the expanded panel using the `expanded` property, and also change the wording for the two panels to "Expanded panel header" and "Minimized panel header". How does that sound?

packages/core/template/default/404.md
tlylt: Just to clarify, I'm not suggesting we should copy the content from GitHub page 404 wholesale. I agree with the part about "file not found" could be awkward for pages that do not exist. On the other hand, this 404 also covers legitimate file not founds, i.e. if you go to https://markbind.org/image.png, this 404 is going to show up as well. If so, I think mentioning "file not found" is perfectly valid. In fact, now "page not found" is mentioned in the title, and "file not found" is mentioned in the content. I think this somewhat ensures that we covered all angles 🤣

Given that this is a template that is general enough at the moment, we can keep it as is. Users can make further modifications to customize the style and phrasing of their site.

What do you think?

tlylt:
> > If you add the expanded property in, does it look better/make the example clearer? I think this is a good point - I re-read the conversation above at least twice to understand what this is all about, it's definitely not the clearest example that we can give. Would be great if we can modify this part slightly to make the example clear.
> 
> Right now, the wording we are using in the template for "Expanded panel" vs "Minimized panel" is referring to the panel header, and not the panel body itself, which is what the `expanded` property is for. So I think that we can add in a third panel to showcase the expanded panel using the `expanded` property, and also change the wording for the two panels to "Expanded panel header" and "Minimized panel header". How does that sound?

Sure let's try that 🚀

packages/core/template/default/404.md
lhw-1: Agreed! And fixed in [4ebf484](https://github.com/MarkBind/markbind/pull/2165/commits/4ebf4845b49cb128b3f2812be191fd2f508c8c87) :smile:

lhw-1:
As per suggestions above, I've included a new panel in [4ebf484](https://github.com/MarkBind/markbind/pull/2165/commits/4ebf4845b49cb128b3f2812be191fd2f508c8c87):

image

Which makes more sense to me :smile:

lhw-1:
Also, given that this PR has been up for quite some time & there has not been more suggestions, I think it's fine to consider this PR as complete. 

(Requesting for approving reviews!)

tlylt (approved):

PR #2172, opened 2023-02-17
Add release-drafter

### LINKED ISSUE #1713: Automated release changelog
Filler issue for project board sticking.

Ups for research / discussion 😀

tlylt:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #1713

Add the release-drafter action to aid in the drafting of release notes. It does not fully automate the process due to the following reasons:
- in order to know what categories the PR is in, you will either need to codify some rules (e.g based on file changes, but may not be accurate), or enforce labeling on the PR (which, is not very easy to ensure 100% compliance due to forgetfulness)
- it's not easy to tell what "issue" the PR is resolving automatically and release-draft doesn't have the ability to do so, but in our template, we are supposed to fill in the issues targeted by each PR

In conclusion, this inclusion does very limited things, but since it's a very low-risk and low-effort integration, we can try it out and make further improvements in the future. The limited benefits include:
- no need to manually copy the release note template from dev docs
- PRs of the form "title" + "PR number" is already pre-filled, so you just need to fill in the issue number
  - You will also need to move the PR items into the correct category, which is exactly the same as what you need to do now
- as for version control, our release process dictates that we would already know the version number when we want to release. Nonetheless, I have added labels (r.major, r.minor, r.patch) and the respective rules such that release-drafter can help calculate the next version number and use it in the draft release note, if the labels are added to the PRs. If not added, the version will simply defaults to patch. Again, this is optional and can be edited easily.

Comments on whether you think we should add release-drafter, or not. (and anything else)

TODO: update devdocs if the answer to the above is yes.

**Testing instructions:**
Hard to test in this PR, can check out https://github.com/tlylt/test-release-drafter for example and sample release note

TODO

---

- [ ] No unrelated changes

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -47,12 +47,12 @@ change-template: "* $TITLE (#$NUMBER)"
 version-resolver:
   major:
     labels:
-      - 'r.major'
+      - 'r.Major'
   minor:
     labels:
-      - 'r.minor'
+      - 'r.Minor'
   patch:
     labels:
-      - 'r.patch'
+      - 'r.Patch'
   default: patch
 name-template: v$RESOLVED_VERSION
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,58 @@
+template: |
+  # markbind-cli
+  <!-- List out each of the PR for this version in the following format: -->
+  <!-- #ISSUE_NUMBER ISSUE_TITLE (#PR_NUMBER) -->
+  ## User Facing Changes
+
+  ### Breaking Changes
+  <!-- Any feature that is made obsolete -->
+
+  > Also give a brief explanation note about:
+  >   - what was the old feature that was made obsolete
+  >   - any replacement feature (if any), and
+  >   - how the author should modify his website to migrate from the old feature to the replacement feature (if possible).
+
+  ### Features
+  <!-- Features enable users (authors/readers) to do something new. -->
+
+  ### Enhancements
+  <!-- Enhances any existing features -->
+
+  ### Fixes
+  <!-- Fixes correct a programming error/assumption. -->
+
+  ### Documentation
+  <!-- Pure changes to the documentation, such as typo, restructuring, etc -->
+
+  ## Other Changes
+
+  ### Code Quality
+  <!-- Refactoring, etc. -->
+
+  ### DevOps Changes
+  <!-- Tooling, etc. -->
+
+  ### Dependencies
+  <!-- Dependency version upgrades of the main or any subpackages -->
+
+  ### Miscellaneous
+  <!-- Any other changes -->
+
+  <!-- When it's time to release, move the items below into the respective categories -->
+
+  $CHANGES
+
+tag-template: 'v$RESOLVED_VERSION'
+change-template: "* $TITLE (#$NUMBER)"
+version-resolver:
+  major:
+    labels:
+      - 'r.major'
+  minor:
+    labels:
+      - 'r.minor'
+  patch:
+    labels:
+      - 'r.patch'
+  default: patch
+name-template: v$RESOLVED_VERSION
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,25 @@
+name: Release Drafter
+
+on:
+  push:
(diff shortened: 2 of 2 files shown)


.github/release-drafter.yml
```diff
@@ -0,0 +1,58 @@
...
+
+  ### Dependencies
+  <!-- Dependency version upgrades of the main or any subpackages -->
+
+  ### Miscellaneous
+  <!-- Any other changes -->
+
+  <!-- When it's time to release, move the items below into the respective categories -->
+
+  $CHANGES
```
raysonkoh, line 43: > You will also need to move the PR items into the correct category, which is exactly the same as what you need to do now

More concretely, does this mean that whatever is in `$CHANGES` have to be manually copied to the appropriate section above?

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,25 @@
+name: Release Drafter
+
+on:
+  push:
```
raysonkoh, line 4: What happens if someone force pushes to the master branch or reverts a PR? Would it mess up the draft release?

.github/release-drafter.yml
```diff
@@ -0,0 +1,58 @@
...
+
+tag-template: 'v$RESOLVED_VERSION'
+change-template: "* $TITLE (#$NUMBER)"
+version-resolver:
+  major:
+    labels:
+      - 'r.major'
```
raysonkoh, line 50: Is the label here case-sensitive? I see that the labels you have added are `r.Major`, `r.Minor`, `r.Patch`.

.github/release-drafter.yml
tlylt: Yes, all the PR items will be introduced here. At the point of release, you will grab items from here and move them accordingly into their categories above.

.github/release-drafter.yml
tlylt: could be case-sensitive, thanks will update

.github/workflows/release-drafter.yml
tlylt: I think it won't mess up anything because it always regenerates the release note based on the current state, so an updated master will mean that the release note will be generated based on the updated state

.github/release-drafter.yml
```diff
@@ -0,0 +1,58 @@
...
+  minor:
+    labels:
+      - 'r.minor'
+  patch:
+    labels:
+      - 'r.patch'
```
tlylt, line 56:
```suggestion
      - 'r.Major'
  minor:
    labels:
      - 'r.Minor'
  patch:
    labels:
      - 'r.Patch'
```

.github/workflows/release-drafter.yml
raysonkoh: Got it, thanks!

jovyntls:
Thank you @tlylt , I like the idea of labelling PRs with r.patch/minor/major!

Does this mean that we would no longer need to add milestones to PRs, Since the milestone would naturally be the next major/minor/patch increment? 

I think overall this seems like a good direction in organising PRs/issues in releases :)

kaixin-hc:
Just read this thread and the accompanying one -- While I think release-drafter looks cool and the existing functionalities you describe also sound interesting, I am very happy with the idea of relying on the format by GitHub. After all, both would require a slight change in our workflow, and a change to use something closer to default GitHub release notes feels more "default" and intuitive, if that makes sense. I am not sure we lose much by doing away with the categories -- and if we need the issue number we could click on the PR and to the issue through that?

raysonkoh:
> Instead of doing this, what if we just use GitHub's existing feature that autogenerates a release note based on the PRs merged after the previous release? 

Thanks for the investigation @tlylt! I think the autogeneration of the release note is good enough for our use-case. 

> This means abandoning (or at least, greatly simplifying) our existing release template to do away with all the categories and the inclusion of issue numbers, in favor of the default format provided by GitHub. With the default as the base, we will still add the section on user-facing breaking changes, such that we still provide a detailed guide on migration.

👍 to simplifying the release notes. Other than breaking changes I think it would be good to include features/enhancements as well as bug fixes.

tlylt:
Thanks all for the feedback, will close this and make a docs update

PR #2184, opened 2023-02-23
Fix Quiz expanding between questions

### LINKED ISSUE #2126: Quiz component expanding when transitioning to next question

### Any related issues?

#2083 

MacOS Monterey

### MarkBind version

4.0.2

When doing a quiz, going from one question to the next causes the quiz expands greatly before shrinking again.

Example:
ezgif com-gif-maker (1)

### Steps to reproduce

In the User Guide -> Using Components -> Others -> Quizzes, start the quiz and go from one question to the next.

### Expected behavior

The quiz should go to the next question and expand/shrink to the size of the next question.

### Actual behavior

When going from one question to the next, the quiz expands greatly before shrinking again.

### Anything else?

_No response_

itsyme:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #2126. Forces intro/outro cards and questions in Quiz to wait for the previous card to leave before allowing the next card to enter.

As a side effect of this PR, `gotoNextQuestion` in `Quiz.vue` had to be refactored into two functions,`gotoNextQuestion` and `showNextQuestion`, in order for the hiding of the current question and the showing of the next question to be called in two separate places. The transition duration was also reduced to keep the same effect as before.

**Testing instructions:**

Fix expanding of Quiz during transition

Cards in Quiz expands during transition from one to the next due to
both cards being rendered in the Quiz.

This could be undesirable as it is an unexpected behaviour for users
and could lead to confusion when doing the Quiz.

Let's make the next card in the Quiz wait for the previous card to 
leave before entering.

This will prevent the expansion of the Quiz component as it is
guaranteed that the Quiz will only have one card rendered at any
one time.

---

- [ ] No unrelated changes


packages/vue-components/src/questions/Quiz.vue
```diff
@@ -128,12 +128,13 @@ export default {
   },
   methods: {
     gotoNextQuestion() {
+      const DELAY = 525;
```
yucheng11122017, line 131: Might be better to define this outside the method since I see it also used inline 153 :)

packages/vue-components/src/__tests__/Quiz.spec.js
```diff
@@ -3,6 +3,12 @@ import Quiz from '../questions/Quiz.vue';
 import Question from '../questions/Question.vue';
 import QOption from '../questions/QOption.vue';
 
+// Function to delay test until question is rendered
+export default function quizTimeout() {
```
yucheng11122017, line 7: I think the export default is unecessary since the function is only used in this file :)

yucheng11122017 (commented):
Hi @itsyme, thanks for the work on this! 
I think the reason why the things in ```Quiz.spec.js.snap``` is being deleted is because it hasn't rendered yet after the triggers are triggered. I added ```await quizTimeOut()``` to lines 198 and 237 in Quiz.spec.js and it seems to work. 
Also I was wondering how did you get 525 for the value of delay?

packages/vue-components/src/__tests__/Quiz.spec.js
itsyme: Good catch! I was trying to export it from another utils.js file but decided not to in the end and forgot to delete it!

packages/vue-components/src/questions/Quiz.vue
itsyme: Yep I agree! I was trying to do that under methods but only functions can be declared in methods. I also tried implementing a getDelay() function but eslint threw me an error so I decided on this instead! Maybe I can try putting it under a different hook. What do you think?

itsyme:
Thanks @yucheng11122017! I got 525 by experimenting. I tried using 500 since the transitions were 0.5s long under `styles` -> `.question` in `Question.vue`, however it seemed to be too short so I increased it till it achieved the desired result!

packages/vue-components/src/questions/Quiz.vue
yucheng11122017: I thinking more along the lines of pulling out the constant DELAY with other constants like STATE_QUIZ_FRESH in line 96 so other methods can reference this as well. What do you think?

itsyme: Ahh I see. I think thats a great idea! I will update that in the next commit!

itsyme:
> Hi @itsyme, thanks for the work on this! I think the reason why the things in `Quiz.spec.js.snap` is being deleted is because it hasn't rendered yet after the triggers are triggered. I added `await quizTimeOut()` to lines 198 and 237 in Quiz.spec.js and it seems to work. Also I was wondering how did you get 525 for the value of delay?

Thank you for the feedback! I added `await quizTimeOut()` to those two lines and it works perfectly now!

ang-zeyu (commented):
flicker

There are some flickering issues introduced by this change.

Presumably because the timeout is timed to be slightly after the transition actually ends.

Consider trying:
- something better guaranteed https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event (what Vue internally uses also actually)
- if that alone dosen't work.. (i have no time to browse the spec to check if the new card is introduced right before or after the transition ends...) consider if you really want to be attaching the new card only after the transition of the previous card has finished. (I am personally fine with this, though this changes the UX)

   I believe there are just 2 main cases to consider:
   1. New card's height > old card's height
   2. Old card's height > new card's height

   What can you do in these scenarios?

   The implementation details should be clear after.

itsyme:
Hi @ang-zeyu! Thank you for the feedback! I will try out your suggestions in the next commit!

itsyme:
I managed to overcome this issue by using `position: relative` for the animation when entering, this allows the incoming question to render in the same place as the previous question. However, since the intro and outro cards are not questions and were not animated, I had to add an animation to them in order to remove the expansion from intro -> first question and last question -> outro. I also realised that the progress bar is being blocked again during the animation from intro -> first question and outro -> first question (when retrying) and am currently working to fix that.

ang-zeyu:
Not sure if it has been updated but there is still some flicker currently:

paneltransitionflicker

itsyme:
Hi @ang-zeyu! Hmm that is odd. Since I made the change I have not observed this behaviour. May I know which browser you are using so that I may further investigate this issue? Thank you!

ang-zeyu:
> Hi @ang-zeyu! Hmm that is odd. Since I made the change I have not observed this behaviour. May I know which browser you are using so that I may further investigate this issue? Thank you!

Desktop chrome/firefox + mobile chrome on android, it occurs about 50% of the time. Unlike the gif it also occurs on mcq/radio questions (unrelated to a specific type of question).

lhw-1:
I have tested this on Ubuntu 20.04, on Firefox 109.0.1 and Chromium 110.0.5481.100. Both seem to work fine without flickering.

jovyntls:
Chipping in here on a different environment, I'm getting flickering as well on MacOS with Chrome:

https://user-images.githubusercontent.com/61113575/221511067-76bd8233-6797-4e67-9bb4-8a70ddc16dff.mov

And MacOS with FireFox:

https://user-images.githubusercontent.com/61113575/221511904-3a63ecba-051e-4bfa-91f2-64b7d7b2f67f.mov

But it's on and off - can't reproduce it consistently.

jovyntls:
@itsyme Try setting throttling using Chrome dev tools (open a mobile view and adjust the throttling option)? I can reproduce the flickering on this branch much more easily when I set throttling to "low-end mobile" 
image


Uploading Screen Recording 2023-02-27 at 5.02.57 PM.mov…

itsyme:
Hi @lhw-1 @jovyntls! Thanks for testing it out! My guess is that the previous question's leave transition finishes slightly before the next question's enter transition finishes, leading to the div becoming height 0 for a brief moment. I will try to fix this else I will explore a different approach!

itsyme:
I managed to solve the issue by changing my approach. However as a side effect of this, `gotoNextQuestion` in `Quiz.vue` had to be refactored into two functions, `gotoNextQuestion` and `showNextQuestion`, in order for the hiding of the current question and the showing of the next question to be called in two separate places. Also, the animation duration was shortened in order to keep a similar effect to the previous animation, since the next question now waits for the previous question to fully leave before starting its animation.

ang-zeyu:
@itsyme Sorry for the late response.

This UX is completely fine to me, I don't see any flickering anymore as well.

> My guess is that the previous question's leave transition finishes slightly before the next question's enter transition finishes, leading to the div becoming height 0 for a brief moment. I will try to fix this else I will explore a different approach!

If you'd still like to experiment with the original UX outside of this PR, 

If I recall previously you were trying to:
- Introduce [absolute](https://github.com/MarkBind/markbind/pull/2184/commits/1a64885f47c244c399c7bc405ff57f3c0f1a078a#diff-3bcc5ce549c558f30435a786ffc4c2c80bc10e4415f2553083f5e354618bdbd1L188) positioning [before](https://vuejs.org/guide/built-ins/transition.html#css-based-transitions) the element is inserted.
- Remove the absolute positioning [after the transition](https://vuejs.org/guide/built-ins/transition.html#css-based-transitions) has finished


I think there's a few other options:

1. Let the quiz expand/shrink to the next card's height just as the "previous card's transition ends".  
   To be exact this can be either:
   - previous card's transition ends - it's still in the DOM so may not be the best option 
   - previous card leaves the DOM - worth a try if you can sync the removal of `position: absolute` on the next card to this.
   - the next card's transition ends - I believe were going for this previously, it is difficult since the previous card leaving the DOM is not synced to this.
2. A much easier option (I think it would be better UX wise as well) is to immediately let the quiz expand/shrink to the next card's height *just as* the user clicks "next". Because you can sync `position: absolute` on the previous card and the introduction of the next card right at the start.
3. Make all child cards `position:absolute` inside the quiz, manage the `height` at the `<quiz>` level through the transition.

packages/vue-components/src/__tests__/Quiz.spec.js
```diff
@@ -51,6 +56,12 @@ const BLANKS_QUESTION = `
 </question>
 `;
 
+// Function to delay test until question is rendered
+function timeoutRenderQuestion() {
+  // eslint-disable-next-line no-promise-executor-return
+  return new Promise(resolve => setTimeout(resolve, 350));
```
ang-zeyu, line 62: Is there no way to work around using a timeout?

Something like `waitForSelector(...).toGoMissing()` (imaginary).
Or creating your own simple transition stub.

There are 17 calls here so resolving this should help improve the speed of tests a fair bit. Unless the runner is parallel, I can't remember..

But don't spend too much time on it, 6s to me is still an ok degradation for now. (just create a follow up issue if we can't resolve this).

ang-zeyu: @jovyntls could you help to merge in this MR if it looks good to you? 🙇

ang-zeyu (approved):

jovyntls (approved):
LGTM! Nice fix @itsyme, let me know if you'll be adding the tests now or opening a future issue for it - will merge if it's the latter

itsyme:
> LGTM! Nice fix @itsyme, let me know if you'll be adding the tests now or opening a future issue for it - will merge if it's the latter

I will open a new issue for this, thank you!

PR #2186, opened 2023-02-23
Remove empty `title` tags in generated tests

### LINKED ISSUE #2166: Remove empty `title` tags in generated tests

### Any related issues?

#2128

Ubuntu 20.04

### MarkBind version

4.0.2

The issue was brought up during the code review in PR #2128 where functional tests generated for template files will have an empty `title` tag if it has not been defined in `site.json`, even if it was defined in the `frontmatter` of the markdown file. An example can be seen [here](https://github.com/MarkBind/markbind/blob/master/packages/cli/test/functional/test_site_convert/test_basic_convert/expected/index.html#L9).

### Expected behavior

The `title` tag should not be generated in the first place if there is no title declared.

### Anything else?

_No response_

lhw-1:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #2166.

~~As pointed out in the issue thread, instead of removing the `title` tag altogether when empty, it would be a better practice to simply declare a default title for sites, e.g. "Default", as is currently being done for the `layout` property (with the default value declared [here](https://github.com/MarkBind/markbind/blob/master/packages/core/src/Layout/index.js#L7)). This PR consists of a straightforward fix by implementing a default value for the `title` tag.~~

As per the discussion in the original issue, after comparing approaches taken by other static site generators, we decided to stick with the default behavior, and additionally address the original concern of removing any unnecessary empty `title` tags generated (as well as prevent any empty `title` tags from being generated in the future).

~~The default value is currently set to "Default", but suggestions on a better title are needed.~~

~~(Setting the `title` as the name of the `.md` file could be a better idea)~~

~~Once the fix has been finalized, documentation (primarily the user guide) will be updated as well.~~

**Testing instructions:**

Remove empty `title` tags in generated tests

---

- [X] No unrelated changes


packages/core/src/Page/index.ts
```diff
@@ -371,7 +373,7 @@ export class Page {
       ...this.pageConfig.frontmatterOverride,
     };
 
-    this.title = this.title || this.frontmatter.title || '';
+    this.title = this.title || this.frontmatter.title || TITLE_DEFAULT;
```
yucheng11122017, line 376: Can be changed to ```this.pageConfig.src``` to get the name of the file.
Can add some cleaning up to remove the extension and maybe convert the name from camelCase / snake_case to Capitalised Words

yucheng11122017 (commented):
Hi @lhw-1, thank you for the work on this! 
I agree with the point that you raised that maybe using the title of the file would be a better way to generate the default title. Default is urm.. not the best :joy:

packages/core/src/Page/index.ts
lhw-1: Yep, I suppose having the name of the file as the stand-in title is more conventional :grin: 

I am wondering if removing the extension + name conversion is strictly necessary though. If a new user deploys a site that they have built & the title of the site shown on the browser tab is something like "index.md", then they would know that something needs to be changed (and will subsequently look into it... hopefully), but it won't be a stretch for them to assume that it is taking it directly from the file name. If it is instead something cleaner like "Index", then they might be confused on where it came from, and also, it might be  difficult to capture all the edge cases when it comes to cleaning the file name :sweat_smile:

lhw-1: > This can be documented so maybe not so much of an issue?

Agreed! My initial thought process was that having three different ways to declare the `title` could be messy, especially if the final `title` is not even the same as the file name. But I suppose it would then be up to the user to read the documentation & up to us to write clear documentation :smile: 

> It could be something simple like [remove file extension](https://stackoverflow.com/questions/4250364/how-to-trim-a-file-extension-from-a-string-in-javascript) and splitting the words. If there are edge cases, users can fall back on using the frontmatter.title.

Yes, I suppose frontmatter could be the backup option for cases that cannot be covered - and the documentation should reflect this clearer (I'll update the documentation in this PR, of course). 

The major change from this update would then be that now the primary way to declare titles would be through the file name, which can be overriden to have custom titles via the frontmatter, which can again be overriden through `site.json`.

tlylt (approved):
LGTM

PR #2191, opened 2023-02-27
Contact form plugin

### LINKED ISSUE #2095: Explore integrating contact form as a plugin

### Any related issues?

_No response_

Author Usability, Reader Usability

A contact form will allow readers to submit feedback or questions directly to the site owner/maintainers. A contact form can also ensure that relevant information about the enquirer can be captured (such as email address, name etc.) if the site owner/maintainers intend to communicate back.

https://web3forms.com/ seems like a suitable third party API to integrate with. Feel free to explore other possible APIs.  

We can implement the contact form as a plugin. Users should be able to toggle between a few pre-configured inputs (e.g name, email, phone number, dropdown with custom options, text field and radio buttons).

image

image

_No response_

### Additional context

_No response_

yucheng11122017:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Closes #2095

**Overview of changes:**
Add a web3Form plugin to allow users to create forms which uses the web3Form api 

Supports 
- Default Contact Us form 
- Standard inputs (name, email, message, submit button)

In terms of number of submissions per month, web 3 form free plan seems to be the best so far with 250 submissions per month. However it only allows one email target vs [Formbold ](https://formbold.com/pricing) with 2 email target but 100 submission / month.
For now, I think web3form offers the best free plan. 

**Testing instructions:**

```Add web 3 form plugin```

---

- [X] No unrelated changes

### CODE REVIEW

File: packages/core/src/plugins/web3Form.ts [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+  if (node.type !== 'tag') {
+    return;
+  }
+  node.name = 'box';
+  const $node = cheerio(node);
+  if (!_.has(node.attribs, 'type')) {
+    $node.attr('type', 'info');
+  }
+  $node.attr('no-icon', '');
+  let header = '<h2>Contact Us</h2>';
```
jovyntls, line 69: Hmm I'm not sure about having a (undocumented) default header here. On one hand it makes sense since this form is mostly used for "Contact Us" purposes; on the other hand, it's slightly weird to have to put `header=""` if you don't want a header. 
Either direction sounds good to me, wdyt?

jovyntls (commented):
Tested and the form works smoothly for me! Nice implementation 🎉 

Mostly nits on documentation! 
The form submissions emails ended up in junk/spam for me -  not sure if adding a note on this in the UG would be suitable?

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,89 @@
...
+
+```js {heading="site.json"}
+{
+  ...
+  "plugins": [
+    "web3Form"
+  ],
+  "pluginsContext": {
+    "web3Form": {
+      "accesskey": "YOUR_WEB_3_FORM_ACCESS_KEY", // replace with your Web3Form access key
```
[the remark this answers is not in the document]
jovyntls: Also is this code block language is labelled `js` instead of `json` to support comments? 🤔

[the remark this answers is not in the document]
yucheng11122017: Opps sorry I copied it from documentation on other plugins which was using js.

packages/core/src/plugins/web3Form.ts
yucheng11122017: Oh that's a good point. I think it would make more sense to remove the header in that case. Thank you!

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,89 @@
+### Plugin: Web3Forms 
+
+This plugin allows you to create forms whose response will be sent directly to your email, using the [Web3Form](https://web3forms.com/) API. 
+
+To set it up, get an access key from [Web3Forms](https://web3forms.com/). Then add`web3Form` to your site's plugin, and add the `accessKey` parameter via the `pluginsContext`.
```
damithc, line 5: @yucheng11122017 I assume this access key is not a 'secret' key?

packages/core/src/plugins/web3Form.ts
yucheng11122017: I think it would be good to have it only if 1. its a default contact form and users do not specify a header. If they do, then the default header will be overriden

docs/userGuide/plugins/web3Form.md
yucheng11122017: Urm not exactly? It's linked to the user's email. web3forms uses it to redirect the form submissions to the user's email.

yucheng11122017:
Thank you for reviewing the PR @jovyntls :)
> The form submissions emails ended up in junk/spam for me - not sure if adding a note in the UG on checking spam folders would be suitable?
I have added a warning the documentation. Also added a warning that they are limited to 250 submissions per month if they are using the free plan

docs/userGuide/plugins/web3Form.md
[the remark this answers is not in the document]
yucheng11122017: I believe so. Changing it to JSON will get warnings regarding comments and regarding the trailing commas in the start

docs/userGuide/plugins/web3Form.md
jovyntls: Is there potential for misuse of the key? My current understanding is that if someone wants to abuse this key, they could use it to direct the submissions from a different form to the user's email address, but this would have the same effect as simply spamming junk submissions to the original form so it doesn't give a malicious user any additional tool for abuse. 
Unless the key can be used to reveal the user's actual email (which the user might have intended to keep private)?

damithc: Should be fine, if the key is not meant to be a 'secret' key that should not be made public.

docs/userGuide/plugins/web3Form.md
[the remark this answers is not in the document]
jovyntls: Ah I see, in that case I think it's fine to keep it as `js` to maintain consistency with the other plugins especially since the difference in syntax highlighting would be visible to users

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,94 @@
+### Plugin: Web3Forms 
+
+This plugin allows you to create forms whose response will be sent directly to your email, using the [Web3Form](https://web3forms.com/) API. 
+<box type="warning" seamless>
+Emails from Web3Forms often end up in spam / junk mail.
```
tlylt, line 5:
```suggestion
Emails from Web3Forms may end up in spam/junk mail.
```

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,94 @@
...
+</variable>
+</include>
+
+**Web3Form plugin also supports a default 'Contact Us' form.**
+
+<include src="codeAndOutput.md" boilerplate>
+<variable name="highlightStyle">html</variable>
+<variable name="code">
+<web-3-form default>
+</web-3-form>
```
tlylt, line 50:
```suggestion
<web-3-form default></web-3-form>
```

Does this work if we write `<web-3-form default />` ? If so perhaps we can simplify this and other similar cases in the docs.

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,94 @@
...
+<web-3-form header="### Form with submit button">
+<submit-button></submit-button>
+</web-3-form>
+
+</variable>
+</include>
+
+**Web3Form plugin supports customised header and styles**
+
+Options from the [box component](https://markbind.org/userGuide/components/presentation.html#boxes) are supported here.
```
tlylt, line 86: use a relative reference to avoid linking out of the currently deployed site, i.e from deploy preview to markbind.org, or from markbind-master to markbind.org

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,94 @@
...
+This plugin allows you to create forms whose response will be sent directly to your email, using the [Web3Form](https://web3forms.com/) API. 
+<box type="warning" seamless>
+Emails from Web3Forms often end up in spam / junk mail.
+</box>
+
+<box type="warning" seamless>
+The free plan from Web3Form only allows 250 submissions per month.
+</box>
+
+To set it up, get an access key from [Web3Forms](https://web3forms.com/). Then add `web3Form` to your site's plugin, and add the `accessKey` parameter via the `pluginsContext`.
```
tlylt, line 12: We should mention whether the access key is ok to be exposed or at least make known that it is exposed, given that it's a working form without a backend. I tried looking at the web3forms docs and it doesn't seem to mention it at all. Perhaps it's due to the fact that the project seems really new...... Nonetheless, we should do a bit of research and put a note about this in our docs, explaining:
- whether this key is exposed to public
- whether such exposure is risky and has any consequences

tlylt (changes requested):
Thank you @yucheng11122017 for the PR. Left some comments. Also, I think the suggestions from Jovyn are not yet resolved (FYI, in case you have overlooked them).

docs/userGuide/plugins/web3Form.md
yucheng11122017: The key will be exposed to the public since users can view the API call to web3forms and see the access key 
ie Developer tools -> network -> payload 
image
It is [mentioned](https://docs.web3forms.com/getting-started/installation#:~:text=%3C!%2D%2D%20REQUIRED%3A%20Your%20Access%20key%20here.%20Don%27t%20worry%20this%20can%20be%20public%20%2D%2D%3E) that its ok for the access key to be public. I emailed them to ask why and they said 

> Yes, it can be public.
> 
> The access key is an alias to your email. This is same as revealing email in public but one step harder. 
> 
> The people who get the access key can only used to email you. Nothing else. We also have spam protection + reCaptcha & domain protection in the Pro plan.

The problem with exposing the key is that other forms can use it and attackers might use it to DOS especially since only 250 submission are allowed per month as mentioned by @jovyntls. 

I will add the above to the documentation!

packages/core/src/plugins/web3Form.ts
```diff
@@ -0,0 +1,150 @@
...
+  const $replaceNode = cheerio('<form onsubmit="submitForm(this)"></form>');
+  $replaceNode.append(`<input type="hidden" name="access_key" value=${pluginContext.accessKey}>`);
+  return $replaceNode;
+}
+
+function deleteWeb3FormAttributes(node: cheerio.Element & DomElement) {
+  const nodeAttribs = node.attribs;
+  if (!nodeAttribs) return;
+  delete nodeAttribs.default;
+  delete nodeAttribs.header;
```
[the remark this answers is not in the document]
yucheng11122017: Hi @jovyntls, is the current implementation ok? 
I can't remove the guard clause without triggering warnings

```
function deleteWeb3FormAttributes(node: cheerio.Element & DomElement) {
  if (!node.attribs) return;
  delete node.attribs.default;
  delete node.attribs.header;
}
```

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,108 @@
+### Plugin: Web3Forms 
+
+This plugin allows you to create forms whose response will be sent directly to your email, using the [Web3Form](https://web3forms.com/) API. 
+<box type="warning" seamless>
+Emails from Web3Forms may end up in spam/junk mail.
+</box>
+
+<box type="warning" seamless>
+The free plan from Web3Form only allows 250 submissions per month.
```
jovyntls, line 9:
```suggestion
The free plan from Web3Forms only allows 250 submissions per month.
```
Nit, can we standardise to use "Web3Forms" throughout? Same for the rest of the docs

docs/userGuide/plugins/web3Form.md
```diff
@@ -0,0 +1,108 @@
+### Plugin: Web3Forms 
+
+This plugin allows you to create forms whose response will be sent directly to your email, using the [Web3Form](https://web3forms.com/) API. 
+<box type="warning" seamless>
+Emails from Web3Forms may end up in spam/junk mail.
```
jovyntls, line 5: Thanks for adding this in! Could we rephrase this as an actionable tip for users? e.g. "Don't forget to check your spam/junk folder if your form responses aren't appearing in your inbox!" 
We could also use a `tip` type for the `<box>` to reduce the number of consecutive warning messages :)

docs/userGuide/plugins/web3Form.md
jovyntls:
```suggestion
The free plan from Web3Form allows up to 250 submissions per month.
```
Another nit, to sound more optimistic 😄

packages/core/src/plugins/web3FormAssets/web-3-form.css
```diff
@@ -0,0 +1,18 @@
+.form-container {
+    border-radius: 5px;
+    background-color: #00b0f0;
```
jovyntls, line 3: I'm guessing this background colour is meant to mirror the `primary` type box's background colour (correct me if I'm wrong!). Is it possible to apply this using the `box` container instead of hardcoding it in the CSS? This would have the benefit of being explicit about the link between the `primary` theme and this form's default colour, and also avoid future issues if there are changes in the Bootstrap themes (e.g. Bootstrap migrations).

packages/core/src/plugins/web3Form.ts
```diff
@@ -0,0 +1,153 @@
...
+  delete node.attribs.default;
+  delete node.attribs.header;
+}
+
+function transformFormInputs(child: cheerio.Element) {
+  if (child.type !== 'tag') {
+    return;
+  }
+  const $node = cheerio(child);
+  switch (child.name.toLowerCase()) {
```
jovyntls, line 41: Valid edge case, but I believe conversion to lowercase is not necessary as it's already done in [NodeProcessor](https://github.com/yucheng11122017/markbind/blob/f12518e6281aceae8fca50d9e336cb17df1cdf90/packages/core/src/html/NodeProcessor.ts#L321) before the pluginManager calls `processNode` :)

jovyntls (changes requested):
Thanks for making all the previous changes @yucheng11122017 🔥  
Added some more comments! And nits on documentation

packages/core/src/plugins/web3FormAssets/web-3-form.css
yucheng11122017: Sorry removed! I did this before moving over to markbind's box :sweat_smile:

packages/core/src/plugins/web3FormAssets/web-3-form.css
```diff
@@ -0,0 +1,12 @@
+input,
+select,
+textarea {
```
jovyntls, line 3: One last nit, do the Bootstrap form classes work well here or would it be more cumbersome than using CSS? https://getbootstrap.com/docs/5.1/forms/overview/#overview

yucheng11122017: Mmm they definitely work but if bootstrap is used the problem is for custom inputs (ie not using `name-input` or `message-input` etc) they would not be styled the same way since users may not have added the bootstrap class. 

Eg. something like this 
```
<web-3-form>
  <label for="fname">First name:</label><br>
  <input type="text" id="fname" name="fname"><br>
  <label for="lname">Last name:</label><br>
  <input type="text" id="lname" name="lname">
</web-3-form>
```

I can change it to add the bootstrap class for the custom inputs if that is better?

docs/userGuide/plugins/web3Form.md
tlylt: Is this in response to the question above:
image

?

Can we re-word it such that it says something along the line of
- generally ok to be public as shared by the team behind it (the reason you gave here)
- however, some potential problem is xxxx

This is because, at the moment, I'm not sure what to make of the warning after reading it...-> it seems to suggest that I shouldn't expose the access key given all the negatives

tlylt:
image
The page nav is corrupted by the example

docs/userGuide/plugins/web3Form.md
yucheng11122017: Thanks for the comments @tlylt, updated in [eef1b38](https://github.com/MarkBind/markbind/pull/2191/commits/eef1b38665d42a54440dfd7bb1d401def31fef98)

yucheng11122017:
> image The page nav is corrupted by the example

@tlylt I changed some of the headings to bold instead in [7a149d9](https://github.com/MarkBind/markbind/pull/2191/commits/7a149d92e3b963b8b3c7a75056c4fc534ef6a0f9). But for the default contact form the heading is `<h2>` which will be picked up by the page nav. Is that ok?

packages/core/src/plugins/web3FormAssets/web-3-form.css
jovyntls: Sorry I missed this! I see, in that case the Bootstrap classes are quite cumbersome - it seems that CSS is a better solution 👍

tlylt:
> > image The page nav is corrupted by the example
> 
> @tlylt I changed some of the headings to bold instead in [7a149d9](https://github.com/MarkBind/markbind/pull/2191/commits/7a149d92e3b963b8b3c7a75056c4fc534ef6a0f9). But for the default contact form the heading is `<h2>` which will be picked up by the page nav. Is that ok?

Hi @yucheng11122017, I just discussed with @jovyntls, I think if we can change the “contact us" default from h2 to a div with the relevant bootstrap classes to style it, it would prevent the issue from occurring. What do you think about this?

jovyntls (dismissed):
LGTM! Thanks again @yucheng11122017 

> Yeah ok sounds good! There is nothing that disables the page-nav from rendering some headings right haha

Nope as far as I know! Might be a possible feature request but I don't think there's much utility to doing so (especially since using Bootstrap styles would be a possible quick workaround for authors who wish to do so)

(Edited the PR description to close the original issue with keywords!)

packages/core/src/plugins/web3FormAssets/web-3-form.css
```diff
@@ -0,0 +1,12 @@
+input,
```
tlylt, line 1: Hi @yucheng11122017, these CSS selectors are too generic, it causes any input etc elements to have unwanted effects. The most obvious one that caught my attention is the increased in margin for our header. See the preview:
image

Current working version is:
image

yucheng11122017:
Hi @tlylt, thanks for bringing that up! I changed the selector to only affect the inputs in the web3Form.

tlylt (dismissed):

packages/core/src/plugins/web3Form.ts
jovyntls, line 1: One last nit, can we use the newly introduced node types (from #2201 - MbNode, NodeOrText etc) wherever possible instead of `cheerio.Element` and `DomElement`?

jovyntls (commented):
Didn't catch this earlier (my bad)!

packages/core/src/plugins/web3Form.ts
yucheng11122017: Added in [ac99277](https://github.com/MarkBind/markbind/pull/2191/commits/ac99277ae0285f48252113d63b809fda4079eead)! :)

jovyntls (approved):
LGTM!! 🥳🥳🥳

packages/core/src/plugins/web3Form.ts
tlylt: Hi @yucheng11122017, just want to share that besides changing the node type, there might be a need to adjust the code accordingly. So in this case, some of the lines do not make sense anymore:
e.g. these few lines are not needed anymore because `attribs` will always exist
image

In the future do take note of the general lesson here as well: the review comment may be specific and points out one thing, that doesn't mean other relevant code changes will not be needed. So relooking at what other potential changes are necessary would help a lot.

tlylt (approved):
@yucheng11122017 Try merging this (note that this is not typescript PR, so it should be a squash commit strategy)

PR #2193, opened 2023-03-01
Add autogenerated breadcrumbs component

### LINKED ISSUE #2094: Support auto generating breadcrumbs

### Any related issues?

_No response_

Reader Usability

Many documentation websites can automatically generate breadcrumbs at the top of the page to allow readers to easily navigate between topics/sub-topics.

Consider auto generating breadcrumbs at the top of each page. Users should be able to toggle this either in the `frontmatter` or in the site config file. 

Can consider using [bootstrap's breadcrumb component](https://getbootstrap.com/docs/5.0/components/breadcrumb/) to implement this:

image

_No response_

### Additional context

_No response_

itsyme:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #2094.

**Testing instructions:**

Add autogenerated breadcrumbs component

---

- [x] No unrelated changes


ong6:
@itsyme Good work with the PR so far! 

>For the sitenav sections with only a dropdown (i.e. clicking on it only expands the dropdown), the breadcrumb is returning the first link instead of the dropdown. I can fix this if I can find a way to differentiate between only dropdown sections and dropdown with page sections

Maybe we can try to check this by looking if the dropdown has a link or not? if the dropdown has a link just link to it, else we just look at the first link. 

>The breadcrumbs do not show for some pages and I cannot explain why. When I print the items array used to store the data for the breadcrumbs, it prints an empty array. I will have to look further into this to find out the root cause.

Not sure what could be causing this issue but it might be the following?  
1. We're using the `document.querySelectorAll` to find the site navigation. But what if the site nav is somehow not in the `ul` element with the class `site-nav-list-root`. 
2. The site navigation structure is not consistent across all pages

itsyme:
> @itsyme Good work with the PR so far!
> 
> > For the sitenav sections with only a dropdown (i.e. clicking on it only expands the dropdown), the breadcrumb is returning the first link instead of the dropdown. I can fix this if I can find a way to differentiate between only dropdown sections and dropdown with page sections
> 
> Maybe we can try to check this by looking if the dropdown has a link or not? if the dropdown has a link just link to it, else we just look at the first link.
> 
> > The breadcrumbs do not show for some pages and I cannot explain why. When I print the items array used to store the data for the breadcrumbs, it prints an empty array. I will have to look further into this to find out the root cause.
> 
> Not sure what could be causing this issue but it might be the following?
> 
> 1. We're using the `document.querySelectorAll` to find the site navigation. But what if the site nav is somehow not in the `ul` element with the class `site-nav-list-root`.
> 2. The site navigation structure is not consistent across all pages

Hi @ong6! Thanks for the suggestions! I managed to fix the second issue with some breadcrumbs not appearing. It seems that for some `site-nav`s in different pages, there seems to be two `ul`s with `site-nav-list-root`, even if they are in the same tier in the `site-nav`. 

e.g. UG -> Authoring Contents -> Reusing Contents has 1 `site-nav-list-root`s but UG -> Authoring Contents -> Tweaking the page structure has 2. Similarly UG -> Authoring Contents -> Using Components -> Navigation has 2 but  UG -> Authoring Contents -> Using Components -> The Rest has 1. 

Do you think this could be a potential bug in the `site-nav` or was it intentionally made this way?

itsyme:
The breadcrumbs works now! Just some small issues to iron out. I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not. Currently, I do not distinguish between them as it made the breadcrumbs look messy.

Also, do we want to have a toggle for breadcrumbs in the frontmatter or just have the user add the breadcrumb component into their layout as the toggle?

jovyntls:
> I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not. 

Ideally yes, but I agree that it may look messy. I can't think of a cleaner way to go about this besides having the usual cursor-hover behaviour on links.

> Also, do we want to have a toggle for breadcrumbs in the frontmatter or just have the user add the breadcrumb component into their layout as the toggle?

I think both are viable! Adding the breadcrumbs component into their layout that could offer users more flexibility and allow them to include it in the default layout. On the other hand, 90% of the time the breadcrumbs are probably in the same place, so having it as a frontmatter toggle could be more convenient. 

I'm leaning towards having it as a component to include in layouts since this aligns more with how we support other features like pageNavs, etc.

itsyme:
> > I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not.
> 
> Ideally yes, but I agree that it may look messy. I can't think of a cleaner way to go about this besides having the usual cursor-hover behaviour on links.
> 
> > Also, do we want to have a toggle for breadcrumbs in the frontmatter or just have the user add the breadcrumb component into their layout as the toggle?
> 
> I think both are viable! Adding the breadcrumbs component into their layout that could offer users more flexibility and allow them to include it in the default layout. On the other hand, 90% of the time the breadcrumbs are probably in the same place, so having it as a frontmatter toggle could be more convenient.
> 
> I'm leaning towards having it as a component to include in layouts since this aligns more with how we support other features like pageNavs, etc.

Thanks for the feedback! I am doing the documentation for Breadcrumbs now and am wondering if I should put it under "Navigation" or "Others". I am leaning towards "Others" as the "Navigation" section is already very long and breadcrumbs is more a peripheral component rather than a core navigation component. What do you think?

lhw-1:
> I am doing the documentation for Breadcrumbs now and am wondering if I should put it under "Navigation" or "Others". I am leaning towards "Others" as the "Navigation" section is already very long and breadcrumbs is more a peripheral component rather than a core navigation component. What do you think?

Hi @itsyme, thanks for the PR! I personally think that since breadcrumbs fundamentally still has to do with navigation between pages, I think putting it under the "Navigation" section still makes more sense than "Others". (As a side note, I feel like we can afford to also rename the "Others" into "Questions and Quizzes" since there's only those components described in the "Others" section)

docs/userGuide/syntax/breadcrumbs.md
```diff
@@ -0,0 +1,15 @@
...
+
+#### Usage
+
+Simply include the breadcrumb component (`<breadcrumb />`) into the page you want and breadcrumbs for that page will be automatically generated.
+
+You can also insert the breadcrumb component into a layout file to generate breadcrumbs for all pages using that layout.
+
+<box type="tip">
+  Breadcrumb components refer to the Site Navigation for the hierarchy of pages. <strong>Breadcrumbs will not appear if there is no Site Navigation present on the page!</strong>
+</box>
```
lhw-1, line 15: Would it be useful to show an example of what breadcrumbs look like (perhaps as shown in the original issue #2094)?

itsyme: Hi @lhw-1! I tried doing that using the `codeAndOutputSeparate` boilerplate but it doesn't show because i think the breadcrumb does not work with includes. I will add an example without using the boilerplate!

damithc:
@itsyme For new features, also good to try them out on CS2103/T website as that's the heaviest MarkBind site we have now.
e.g., https://github.com/nus-cs2103-AY2223S2/website
Check how the new feature affects various pages, including the the textbook (https://nus-cs2103-ay2223s2.github.io/website/se-book-adapted/index.html), printer-friendly pages in particular.

ong6:
> The breadcrumbs works now! Just some small issues to iron out. I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not. Currently, I do not distinguish between them as it made the breadcrumbs look messy.

Great work so far! What do you think about leaving those unclickable links out of the breadcrumbs completely. (don't show them all together.)

> Also, do we want to have a toggle for breadcrumbs in the frontmatter or just have the user add the breadcrumb component into their layout as the toggle?

I agree with Jovyn on this point. Leaving it in layouts seems right.

tlylt:
Can you also share what it looks like for longer-than-usual titles (hence a longer breadcrumb), to see if it handles that well?

itsyme:
> Great work so far! What do you think about leaving those unclickable links out of the breadcrumbs completely. (don't show them all together.) My reasoning for this is because breadcrumbs are used mainly for navigation and if there is no link, perhaps there is no need to show them since they can't navigate to it anyways.

I think that leaving the unclickable links out might reduce the value that the breadcrumbs bring to the user.

I find the unclickable links being displayed helps the user to track which section they currently are more easily, even if the links do not direct them to a page on click.

Also, much of the sitenav dropdowns (that I have seen so far) do not contain a page. This could lead to our breadcrumb displaying only the current page majority of the time. What do you think?

itsyme:
@tlylt In my opinion it handles this case well.

Screenshot 2023-03-15 at 11 53 45 PM

I also tested out the Breadcrumbs for a SiteNav with deep nested dropdowns and it seems to handle that well too.

Screenshot 2023-03-15 at 11 55 44 PM

ong6:
> Screenshot 2023-03-15 at 11 53 45 PM

Should the /overview be in the same line as the long text above though?

tlylt:
> Hi @ong6! Thanks for the suggestions! I managed to fix the second issue with some breadcrumbs not appearing. It seems that for some `site-nav`s in different pages, there seems to be two `ul`s with `site-nav-list-root`, even if they are in the same tier in the `site-nav`.
> 
> e.g. UG -> Authoring Contents -> Reusing Contents has 1 `site-nav-list-root`s but UG -> Authoring Contents -> Tweaking the page structure has 2.
> 
> Similarly, UG -> Authoring Contents -> Using Components -> Navigation has 2 but UG -> Authoring Contents -> Using Components -> The Rest has 1.
> 
> Do you think this could be a potential bug in the `site-nav` or if it intentionally made this way?

Is this in HTML or? I tried checking the HTML for the site nav and I'm not sure what you are referring to:
image

tlylt:
> The breadcrumbs works now! Just some small issues to iron out. I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not. Currently, I do not distinguish between them as it made the breadcrumbs look messy.

I think UI wise it would be great to have it differentiated. Especially since we are using blue color for clickable links, having even the non-clickable ones to be styled this same way is confusing.
image
 i.e:, visually there's no difference between "Authoring Contents" and "Using Components", but one is clickable and the other is not. 

Not too sure what alternatives would be most desirable, might need to explore and compare a few possible options.

itsyme:
> > Screenshot 2023-03-15 at 11 53 45 PM
> 
> Should the /overview be in the same line as the long text above though?

That might be better! I will look into this to try and change it!

itsyme:
> > Hi @ong6! Thanks for the suggestions! I managed to fix the second issue with some breadcrumbs not appearing. It seems that for some `site-nav`s in different pages, there seems to be two `ul`s with `site-nav-list-root`, even if they are in the same tier in the `site-nav`.
> > e.g. UG -> Authoring Contents -> Reusing Contents has 1 `site-nav-list-root`s but UG -> Authoring Contents -> Tweaking the page structure has 2.
> > Similarly, UG -> Authoring Contents -> Using Components -> Navigation has 2 but UG -> Authoring Contents -> Using Components -> The Rest has 1.
> > Do you think this could be a potential bug in the `site-nav` or if it intentionally made this way?
> 
> Is this in HTML or? I tried checking the HTML for the site nav and I'm not sure what you are referring to: image

Yep it is in the HTML! It occurs when you are at page "Reusing Contents" or "Tweaking the page structure". I too, could not find the other site-nav-root while inspecting the page, however when I queried the DOM for `ul` elements that contained `site-nav-list-root` and appended them to an array, I get an array of length 2 when in these two pages.

itsyme:
> > The breadcrumbs works now! Just some small issues to iron out. I was wondering if we should visually distinguish between breadcrumbs that are clickable and those that are not. Currently, I do not distinguish between them as it made the breadcrumbs look messy.
> 
> I think UI wise it would be great to have it differentiated. Especially since we are using blue color for clickable links, having even the non-clickable ones to be styled this same way is confusing. image i.e:, visually there's no difference between "Authoring Contents" and "Using Components", but one is clickable and the other is not.
> 
> Not too sure what alternatives would be most desirable, might need to explore and compare a few possible options.

Alright! I will try different variations and post them here for further discussion!

itsyme:
> Thank you for exploring the options @itsyme !
> 
> > I changed the colour of the non-links to match the links as I felt the breadcrumbs looked very messy with 3 very distinct colours. What do you think about using a more faded blue for non-links to differentiate between them? This would differentiate them without causing too much variance in the colours.
> 
> Hmm, personally I think the contrast between blue and faded blue is not strong enough for me to recognise that the colours are different at first glance. Blue is also a common colour for hyperlinks, so faded blue might still be misrecognised as a hyperlink.
> 
> imo the lesser of two evils would be using blue only for clickable elements, and grey for all others?

I agree! Having only 2 colours would make it less messy as well.

tlylt:
> > Thank you for exploring the options @itsyme !
> > > I changed the colour of the non-links to match the links as I felt the breadcrumbs looked very messy with 3 very distinct colours. What do you think about using a more faded blue for non-links to differentiate between them? This would differentiate them without causing too much variance in the colours.
> > 
> > 
> > Hmm, personally I think the contrast between blue and faded blue is not strong enough for me to recognise that the colours are different at first glance. Blue is also a common colour for hyperlinks, so faded blue might still be misrecognised as a hyperlink.
> > imo the lesser of two evils would be using blue only for clickable elements, and grey for all others?
> 
> I agree! Having only 2 colours would make it less messy as well.

I think all grey + blue for clickable looks great

itsyme:
> @itsyme For new features, also good to try them out on CS2103/T website as that's the heaviest MarkBind site we have now. e.g., https://github.com/nus-cs2103-AY2223S2/website Check how the new feature affects various pages, including the the textbook (https://nus-cs2103-ay2223s2.github.io/website/se-book-adapted/index.html), printer-friendly pages in particular.

Hi @damithc! The breadcrumbs work as expected on the site. When printing, the breadcrumbs are printed as seen on the site. Do you think this is alright? Or would you like the breadcrumbs to not be shown when printing?

Screenshot 2023-03-26 at 10 34 15 AM

damithc:
> Hi @damithc! The breadcrumbs work as expected on the site. When printing, the breadcrumbs are printed as seen on the site. Do you think this is alright? Or would you like the breadcrumbs to not be shown when printing?

@itsyme check how they appear in this page: ;
image

That's the one I use to generate the textbook PDF file.

itsyme:
> > Hi @damithc! The breadcrumbs work as expected on the site. When printing, the breadcrumbs are printed as seen on the site. Do you think this is alright? Or would you like the breadcrumbs to not be shown when printing?
> 
> @itsyme check how they appear in this page: ; image
> 
> That's the one I use to generate the textbook PDF file.

Hi @damithc! This page is actually unchanged as breadcrumbs will only render when there is a sitenav present in the page, which is not the case for this page.

damithc:
> Hi @damithc! This page is actually unchanged as breadcrumbs will only render when there is a sitenav present in the page, which is not the case for this page.

@itsyme Got it. It's all good then.

packages/vue-components/src/Breadcrumb.vue
```diff
@@ -0,0 +1,95 @@
...
+        return;
+      }
+      if (el.classList.contains('site-nav-list-root')) {
+        siteNav = el;
+        firstRootFound = true;
+      }
+    });
+
+    siteNav.querySelectorAll('a[href]').forEach((el) => {
+      window.console.warn(el.textContent);
```
jovyntls, line 54: Is this console statement here intentional? If not, let's remove it :)

packages/vue-components/src/Breadcrumb.vue
```diff
@@ -0,0 +1,95 @@
...
+      if (firstRootFound) {
+        return;
+      }
+      if (el.classList.contains('site-nav-list-root')) {
+        siteNav = el;
+        firstRootFound = true;
+      }
+    });
+
+    siteNav.querySelectorAll('a[href]').forEach((el) => {
```
jovyntls, line 53: This is a pretty long block - for readability and future maintenance purposes, can we add some comments to explain what's going on here, or extract some operations into functions (whichever makes more sense)?

jovyntls (changes requested):
Nice work @itsyme 🥳 

Is it possible to add some tests? e.g. generating breadcrumbs with different sitenavs or possible edge cases

packages/vue-components/src/Breadcrumb.vue
itsyme: > Nice work @itsyme 🥳
> 
> Is it possible to add some tests? e.g. generating breadcrumbs with different sitenavs or possible edge cases

Thanks for reviewing my PR @jovyntls! I'm currently working on the tests!

itsyme:
> Nice work @itsyme 🥳
> 
> Is it possible to add some tests? e.g. generating breadcrumbs with different sitenavs or possible edge cases

Thanks for reviewing my PR @jovyntls! I'm currently working on the tests!

packages/vue-components/src/Breadcrumb.vue
```diff
@@ -0,0 +1,102 @@
...
+          if (aElement === null) {
+            this.items.unshift({
+              'title': divElement.textContent,
+              'link': null,
+            });
+          }
+          this.items.unshift({
+            'title': aElement.textContent,
+            'link': aElement.getAttribute('href'),
+          });
```
jovyntls, line 88: I think an `else` block may be needed here, or the elements will get added twice

jovyntls (changes requested):
Thanks @itsyme , the component is working well!

Is it possible to add tests for generating the `items` array from the siteNav, since the current tests stub the `items` array?

I'm also wondering if we should add this to the default template 🤔  What do you think?

jovyntls:
Not a big issue, but I'm getting a weird jerky change of the cursor when moving to an `.active` class:

 I'm fine with removing the class entirely since having the cursor as a caret when hovering over text seems reasonable to me, given that that's the standard cursor behaviour.


https://user-images.githubusercontent.com/61113575/228275386-20f409e1-b47d-45a4-a8a4-4cfc3fa2bdc0.mov

docs/userGuide/syntax/breadcrumbs.md
```diff
@@ -0,0 +1,45 @@
...
+%%OUTPUT:%%
+
+<div class="indented">
+
+> <breadcrumb />
+
+</div>
+
+<box type="tip" seamless>
+  Breadcrumb components refer to the Site Navigation for the hierarchy of pages. <strong>Breadcrumbs will not appear if there is no Site Navigation present on the page!</strong>
```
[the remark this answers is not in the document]
itsyme: Hi @jovyntls! I tried this but it rendered the asterisks instead of making the contents bold. I have updated it to have an empty line between the bolded part and not bolded part which makes it work!

packages/vue-components/src/Breadcrumb.vue
```diff
@@ -0,0 +1,102 @@
...
+        currentEl = currentEl.parentElement;
+      }
+    });
+  },
+};
+</script>
+
+<style scoped>
+    /* Make cursor default when there is no link */
+    .active:hover {
```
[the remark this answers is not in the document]
itsyme: The `active` class is from bootstrap to indicate the current item, I think I will create a new class with a more appropriate name that does what `active` does!

packages/vue-components/src/Breadcrumb.vue
```diff
@@ -0,0 +1,102 @@
...
+      return index === length - 1;
+    },
+  },
+  mounted() {
+    let siteNav = null;
+    let firstRootFound = false;
+
+    // Identify the first site-nav-list-root
+    // In the ideal case, there is only one site-nav-list-root
+    // however this is not the case for all pages
```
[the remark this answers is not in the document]
itsyme: From my experience, the other site-nav-list-root is a subset of the current sitenav. I feel like this is unexpected behaviour as ideally there should only be one site-nav-list-root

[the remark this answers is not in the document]
itsyme: I have found out why the bug is happening! I realised that the navigation page has an example sitenav in the page, resulting in the weird behaviour.

itsyme:
> Thanks @itsyme , the component is working well!
> 
> Is it possible to add tests for generating the `items` array from the siteNav, since the current tests stub the `items` array?
> 
> I'm also wondering if we should add this to the default template 🤔 What do you think?

Thanks for the review @jovyntls! I was having some issues with the tests with siteNav and might need some time to work it to get them to work. I will try to get it done ASAP! As for the default template, I think it would be a great addition! I will add it in the next commit!

itsyme:
> Thanks @itsyme , the component is working well!
> 
> Is it possible to add tests for generating the `items` array from the siteNav, since the current tests stub the `items` array?
> 
> I'm also wondering if we should add this to the default template 🤔 What do you think?

HI @jovyntls! I have made the relevant changes including generating the breadcrumbs from the sitenav component. However, because it requires a short timeout, I have skipped the tests because of issue #2238 just to be sure!

jovyntls:
I see, why does it require a setTimeout? Is it to wait for the breadcrumbs to be generated after mounting?

itsyme:
> I see, why does it require a setTimeout? Is it to wait for the breadcrumbs to be generated after mounting?

Yup! There needs to be a very short delay for the breadcrumbs to generate the data from the sitenav, else the snapshot will return an empty breadcrumb!

itsyme:
> Would the `nextTick()` method fix this? 🤔 Something like `await nextTick()` to wait for the component to mount and run the generating breadcrumbs method.
> 
> For docs, ctrl-F 'nextTick' here: https://test-utils.vuejs.org/guide/advanced/async-suspense.html#a-simple-example-updating-with-trigger
> 
> Unfortunately this won't work for the other tests in #2238 since those components implicitly use setTimeouts 😢

Thanks for the suggestion! It worked well and I have implemented it in the latest commit and unskipped the tests!

jovyntls (approved):
LGTM! This would be a nice feature addition 🎉  Thanks @itsyme !

PR #2200, opened 2023-03-08
Fix panel transition abruptly snapping to the end

### LINKED ISSUE #2185: Fix panel transition abruptly "snapping" under certain circumstances

### Any related issues?

_No response_

Desktop Chrome + Mobile Andoid Chrome

### MarkBind version

Latest main branch

panelheighttransitionbug

Under certain circumstances, the panel expansion transition abruptly snaps to the end.

### Expected behavior

The transition should continue smoothly all the way.

### Anything else?

_No response_

yucheng11122017:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Fixes #2185. 

Current problem: 
1. The panel is opened,
2. The max height of the panel is set to the [scroll height](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/PanelBase.js#:~:text=in%2Dout%27%3B-,this.%24refs.panel.style.maxHeight%20%3D%20%60%24%7Bthis.%24refs.panel.scrollHeight%7Dpx%60%3B,-%7D) of the panel which does not include the [bottom margin](https://github.com/MarkBind/markbind/blob/master/packages/vue-components/src/panels/NestedPanel.vue#:~:text=margin%2Dbottom%3A%2013px%3B) of the panel-switch. The transition uses this value.
3. When it finishes the transition, the max height is set to none which is when the entire panel with the scroll button is displayed. This does not have a transition, resulting in the abrupt transition

**Overview of changes:**
Added collapse button's bottom margin to the panel's maxHeight when panel is transitioning. 

An alternative I thought of would be to scale up the maxHeight (eg. the maxHeight = 1.5 * scrollHeight)
Pros would be that if other components that are added in the future to the panels also have a bottom margin, there would be a need to keep including those elements in the calculation
Cons are the transition might not be as consistent since the ratio of the actual height (including the bottom margin) to calculated maxHeight would defer. And if a really large bottom margin is used, the problem would occur again

**Testing instructions:**

```Fix panel transition abruptly snapping to the end```

---

- [X] No unrelated changes

### CODE REVIEW

**Comment by yucheng11122017** [comment]
I'm having this issue that the bottom margin for the collapse button is not being computed correctly during testing. Currently just returns an empty string since the CSS is not being inherited correctly from the nested panel. So the maxHeight it not being set properly with the margin of the collapse button. 

If anyone has any advice on how to deal with this, would be much appreciated!

**Comment by lhw-1** [comment]
Thanks for the PR @yucheng11122017!

While the changes made seems to have solved the original issue, it seems introduces a new problem - the panels are snapping back when the panel body is minimized:

https://user-images.githubusercontent.com/25716529/225216646-41a17326-759c-4ba7-911a-143b953c2c5c.mp4

---

This is different from the original behavior where the closing transition would be smooth. Might want to look into this!

As a side note, I've realized that the original bug of the abrupt snapping does not occur on FireFox Ver. 110.0.1 (on Windows 10), but does occur on Google Chrome Ver. 111.0.5563.65 (also on Windows 10), while the new issue of snapping closing transition occurs on both. (I don't know if this is useful information, just something I noticed!)

**Comment by yucheng11122017** [comment]
Hi @lhw-1, thanks for the comments! I think the snapping back was the desired behavior as per #2159 which was just recently merged. It removes the closing transition for panels

**Comment by lhw-1** [comment]
> Hi @lhw-1, thanks for the comments! I think the snapping back was the desired behavior as per #2159 which was just recently merged. It removes the closing transition for panels

Oh I see, I didn't realize this was now the intended behavior (though to be honest, having the closing transition seems more natural). Regardless, I suppose that means all's working as intended 👍

**Review by raysonkoh — APPROVED** [review]
LGTM! Tested on Chrome, Firefox and Safari. Interestingly, I didn't really notice the original issue on Firefox and Safari. Good job!

File: packages/vue-components/src/panels/PanelBase.js [hunk 1]
```diff
@@ -196,7 +196,28 @@ export default {
...
+      const bottomSwitch = document.querySelector('.card-body > .collapse-button');
+      if (bottomSwitch == null) {
+        return this.$refs.panel.scrollHeight;
+      }
+      const bottomSwitchStyle = window.getComputedStyle(bottomSwitch);
+      const bottomSwitchBottomMargin = parseFloat(bottomSwitchStyle.marginBottom);
+      if (Number.isNaN(bottomSwitchBottomMargin)) {
+        return this.$refs.panel.scrollHeight;
+      }
+      return this.$refs.panel.scrollHeight + bottomSwitchBottomMargin;
```
ang-zeyu, line 220: would creating a `<div>` wrapper with bottom padding instead work?

PR #2201, opened 2023-03-10
Create unified types for nodes

### LINKED ISSUE #2114: Consider creating unified types for common variables
As we continue with the TypeScript migration (#1913), we may want to create unified types to be used across common method arguments. One example could be creating a `Node` type that is used throughout different methods/classes when the `node` parameter is passed around. 

We'll need to decide:
- When we want to do it — after we have fully migrated, or while the migration is currently underway, or if we even want to standardise these types at all
- What the types should be (e.g. for Node, `cheerio.Element & DomElement` could be an option)
- How to handle applying types across the whole codebase

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Fixes #2114 (the creation part)

**Overview of changes:**
Creates MbNode, TextElement and NodeOrText types. This avoids unnecessary and unsafe typecasting later on, and creates clearer distinction between how we handle nodes.
This PR implements a minimal "demo" to preview how the changes will look like. Further refactoring may be done in future PRs.

- ~~Naming of the type `Node` - seems to clash with many other types. Would `NodeElement` be a better name?~~ Renamed `MbNode`
- General feedback/suggestions are welcome

**Testing instructions:**
N/A, there should be no difference in behaviour

```
Create unified types for nodes

There is no standardised type for node variables, resulting in several
undesirable issues:
* `as unknown as` typecasts throughout the codebase 
* unnecessary checks on node attributes to satisfy the linter
* incompatible types, treated as interchangeable in conversions

Let's add common MbNode, TextElement and NodeOrText types that enforce
attributes for nodes throughout the codebase. Doing so avoids the
abovementioned issues, and formalises our view of the relationship
between cheerio.Element, DomElement, and all the other attributes
expected from node variables.
```

---

- [x] No unrelated changes

### DIFF
--- a/packages/core/src/utils/node.ts
+++ b/packages/core/src/utils/node.ts
@@ -1,12 +1,22 @@
-import cheerio from 'cheerio';
 import { DomElement } from 'htmlparser2';
 
+/*
+ * A TextElement is a simple node that does not need complex processing.
+ */
 export type TextElement = DomElement;
 
+/*
+ * MbNode (MarkbindNode) is an element that can be operated on by cheerio and our own node processing
+ * methods. It must have a name (used to identify what kind of node it is), attributes (possibly empty),
+ * and children nodes (possibly empty). This type allows us to assert that these attributes exist.
+ */
 export type MbNode = DomElement & cheerio.Element & {
   name: string,
   attribs: { [key: string]: any },
   children: NodeOrText[],
 };
 
+/*
+ * NodeOrText is used before a node can be casted to either TextElement or MbNode.
+ */
 export type NodeOrText = TextElement | MbNode;
--- a/packages/core/src/utils/node.ts
+++ b/packages/core/src/utils/node.ts
@@ -6,7 +6,7 @@ import { DomElement } from 'htmlparser2';
 export type TextElement = DomElement;
 
 /*
- * MbNode (MarkbindNode) is an element that can be operated on by cheerio and our own node processing
+ * MbNode (MarkBindNode) is an element that can be operated on by cheerio and our own node processing
  * methods. It must have a name (used to identify what kind of node it is), attributes (possibly empty),
  * and children nodes (possibly empty). This type allows us to assert that these attributes exist.
  */
--- a/packages/core/src/utils/node.ts
+++ b/packages/core/src/utils/node.ts
@@ -3,10 +3,10 @@ import { DomElement } from 'htmlparser2';
 
 export type TextElement = DomElement;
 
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/node.ts
```diff
@@ -0,0 +1,12 @@
+import cheerio from 'cheerio';
+import { DomElement } from 'htmlparser2';
+
+export type TextElement = DomElement;
+
+export type Node = DomElement & cheerio.Element & {
```
ong6, line 6: As @raysonkoh suggested, maybe we should rename this to something like `NodeElement` or `MarkbindNode`.

jovyntls: I'd suggest `NodeElement`, wdyt?

raysonkoh: I think `NodeElement` might be a little wordy. How about `MbNode`?

jovyntls: 👍  renamed to `MbNode`

ong6 (approved):
LGTM

packages/core/src/utils/node.ts
```diff
@@ -0,0 +1,12 @@
+import cheerio from 'cheerio';
+import { DomElement } from 'htmlparser2';
+
+export type TextElement = DomElement;
+
+export type MbNode = DomElement & cheerio.Element & {
```
tlylt, line 6: Would be nice to document the rationale for this in the code, as it might cause some confusion as to why such an abstraction exists. Same for the one below.

packages/core/src/utils/node.ts
```diff
@@ -0,0 +1,22 @@
+import { DomElement } from 'htmlparser2';
+
+/*
+ * A TextElement is a simple node that does not need complex processing.
+ */
+export type TextElement = DomElement;
+
+/*
+ * MbNode (MarkbindNode) is an element that can be operated on by cheerio and our own node processing
```
tlylt, line 9:
```suggestion
 * MbNode (MarkBindNode) is an element that can be operated on by cheerio and our own node processing
```

tlylt: Typo here, I suppose MbNode is still ok, or do you think we should change it to MBNode?

jovyntls: Hmm I'd prefer MbNode (MBNode feels like too many capital letters). Open to suggestions!

tlylt:
LGTM, @raysonkoh want to take a last look?

PR #2038, opened 2022-11-15
Tabgroup dropdown button size

petermonky:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [x] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

Resolves #2037

**Overview of changes:**

Adjusts size of tabgroup buttons in tabset to match that of non-tabgroup buttons.

| **Clickable region for non-tabgroup button** | **Clickable region for tabgroup button** |
|---|---|
| image | image |

**Anything you'd like to highlight/discuss:**

nil

**Testing instructions:**

Confirm clickable region by hovering over tabgroup and non-tabgroup buttons in rendered page.

**Proposed commit message: (wrap lines at 72 characters)**

Unify tabset sizing across tabgroup and non-tabgroup buttons

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No unrelated changes

PR #2202, opened 2023-03-11
Standardise size of tabgroup buttons in tabset to match that of non-tabgroup buttons.

### LINKED ISSUE #2037: Tabset dropdown button should be more easily clickable

### Any related issues?

_No response_

Reader Usability

The clickable area of a tabgroup dropdown button in a tabset is smaller than that of a non-tabgroup button.

**Clickable region for non-tabgroup button**

image

**Clickable region for tabgroup button**

image

A bigger clickable region for tabgroup buttons that matches that of non-tabgroup buttons.

_No response_

### Additional context

_No response_

yucheng11122017:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Duplicate of #2038, hence closes #2038
Resolves #2037 

**Overview of changes:**
Standardized the size of tab group
\

Initially: Clickable region for tab group is only at the arrow. 
image
This is different from the clickable region for the non tab group which is the entire tab including the white space
image

Changed the clickable region for the tab group to the entire button (arrow and white space)

This deals with [problem ](https://github.com/MarkBind/markbind/pull/2038#discussion_r1024744659) raised in #2038 regarding the additional padding in nav-page by setting padding to be 0 in `ul`. 
An alternative is to remove the nav-link in page-nav suggested in #2038 but this would not be consistent with bootstrap nav components. This would also be considered a breaking change

**Testing instructions:**
Confirm clickable region by hovering over tabgroup and non-tabgroup buttons in rendered page.

Unify tabset sizing across tabgroup and non-tabgroup buttons

---

- [X] No unrelated changes


tlylt:
Hi @yucheng11122017, could you explain the changes made in this PR? We are ok with a breaking change if necessary. Just a side note I think not removing nav-link in the navbar would be best, as that's how bootstrap nav component works.

yucheng11122017:
Hi @tlylt, 
Basically what @petermonky did was to shift the nav-link from tabset to dropdown instead to enable the entire region to be clicked instead of just the tabset button (which is only the arrow) to standardise it with the tab buttons. 
But because nav bar also uses nav-link to wrap dropdowns, he suggested to remove the navlink so that the double padding will not be applied since the nav-link class is already applied within the dropdown itself. 

But personally I think it's not the best idea as well. So instead I implemented the workaround suggested by @jonahtanjz to just remove the padding `nav-link` on the outer `<li>` element

tlylt:
> Anything you'd like to highlight/discuss:
This deals with https://github.com/MarkBind/markbind/pull/2038#discussion_r1024744659raised in https://github.com/MarkBind/markbind/pull/2038 regarding the additional padding in nav-page by setting padding to be 0 in ul. This should be a temporary workaround until a release with breaking changes is ready to be released.

@yucheng11122017 is this part in the description still relevant?

yucheng11122017:
> > Anything you'd like to highlight/discuss:
> > This deals with [#2038 (comment)](https://github.com/MarkBind/markbind/pull/2038#discussion_r1024744659raised) in #2038 regarding the additional padding in nav-page by setting padding to be 0 in ul. This should be a temporary workaround until a release with breaking changes is ready to be released.
> 
> @yucheng11122017 is this part in the description still relevant?

imo no and updated the description accordingly. I'll leave it there as an alternative if anyone believes otherwise.

packages/vue-components/src/Dropdown.vue
```diff
@@ -230,7 +230,10 @@ export default {
     }
 
     .navbar .dropdown-toggle {
-        color: inherit;
         text-decoration: none;
     }
+
+    li {
```
tlylt, line 236: Is this still necessary? Can you show screenshots to illustrate the effect of this?

tlylt (dismissed):
Currently on markbind-master: 
image
This PR:
image

The previous discussion in the original PR seems to be valid here as well.

Lastly, can you write about how one might migrate an existing MarkBind site to accommodate this change after this is merged? (Or is this not breaking? then the above example should not produce problematic output when no changes are made to the source?)

packages/vue-components/src/Dropdown.vue
yucheng11122017: Yup still necessary!

This is without the padding 
image

This is with the padding 
image

yucheng11122017:
> Currently on markbind-master: image This PR: image
> 
> The previous discussion in the original PR seems to be valid here as well.


Hi @tlylt, could I check where did you see this error? The [PR deploy review](https://deploy-preview-2202--markbind-master.netlify.app/userguide/components/navigation#dropdowns:~:text=%3C/navbar%3E-,OUTPUT%3A,-MarkBind) seems to be show the padding correctly. Its from the nav bar right
 
> Lastly, can you write about how one might migrate an existing MarkBind site to accommodate this change after this is merged? (Or is this not breaking? then the above example should not produce problematic output when no changes are made to the source?)

There should be no breaking changes.

tlylt:
> > Currently on markbind-master: image This PR: image
> > The previous discussion in the original PR seems to be valid here as well.
> 
> Hi @tlylt, could I check where did you see this error? The [PR deploy review](https://deploy-preview-2202--markbind-master.netlify.app/userguide/components/navigation#dropdowns:~:text=%3C/navbar%3E-,OUTPUT%3A,-MarkBind) seems to be show the padding correctly. Its from the nav bar right
> 
> > Lastly, can you write about how one might migrate an existing MarkBind site to accommodate this change after this is merged? (Or is this not breaking? then the above example should not produce problematic output when no changes are made to the source?)
> 
> There should be no breaking changes.

Thanks for checking, I might have looked at the PR preview of the original PR.

packages/vue-components/src/Dropdown.vue
```diff
@@ -230,7 +230,10 @@ export default {
     }
 
     .navbar .dropdown-toggle {
-        color: inherit;
```
lhw-1, line 233: Is there a particular reason why we are removing this property?

lhw-1 (approved):
LGTM aside from a nit! This should, though small, improve the user experience for dropdowns :)

lhw-1:
Also, just highlighting that due to this fix, the area that a `dropdown` occupies in a navbar is now larger:

image

packages/vue-components/src/Dropdown.vue
yucheng11122017: I think @petermonky removed it because the color inherit will be overridden so it doesn't make a difference

yucheng11122017:
> Also, after this fix, the area that a `dropdown` occupies in a navbar is now larger than normal navbar elements (just a cosmetic change, so I have no issues, just highlighting it):
> 
> Before:
> 
> image
> 
> After:
> 
> image

Hi @lhw-1, thanks for the comments! I changed the selector to nav-link > nav-link to remove the padding for the inner nav link in dropdowns. This seems to solve the issue.

EltonGohJH (approved):
LGTM

PR #2204, opened 2023-03-11
Tweak sitenav design

### LINKED ISSUE #2085: Improve site nav coloring/font schemes

### Any related issues?

_No response_

_No response_

This comes in 2 parts:
1. Super bold usage in docs - imo this is too bold
2. Default font design, etc.

I'm not much of a UX designer but the current site nav design still feels a little lacking. Could just be that I've been staring at it for too long. Feel free to input here.

Remove the super bold usage and any other tweaks.

_No response_

### Additional context

_No response_

itsyme:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolves #2085. Removed the bold in the `site-nav` in user guide. Updated the paddings to make the text vertically centered. Updated the font sizes of the different tiers (values were chosen by setting the first tier to an appropriate size and scaling the subsequent tiers by a factor of 1.16 (golden ratio) down from the previous tier).

Even after editing the design of `site-nav-top`, I still feel it can improved further. Seeking feedback/suggestions to improve `site-nav-top` or any other part of `site-nav`.

**Testing instructions:**

Tweak sitenav design

---

- [x] No unrelated changes


tlylt:
Vertical align looks weird to me, I think original left align is fine. Otherwise LGTM

itsyme:
Hi @tlylt! I totally agree, I changed it back to be left aligned but increased the font size slightly. I initially centered it to try and differentiate the `site-nav-top` from the first item in the `site-nav` as I felt that it could be slightly confusing. I feel that this increase in font size would achieve a similar effect.

tlylt (approved):
LGTM

itsyme:
Hi @tlylt! The tiers after third tier of dropdowns seem to not have their font sizes scaled down. Would you think it's better to continue scaling it down further or to keep it at the smallest font size?
Screenshot 2023-03-13 at 4 40 42 PM

itsyme:
Hi @tlylt! I managed to align them but I had to introduce a small transition to mask the solution. The issue was that it was using `em` for width which made the width of the button change as the font size decreased. I changed it to `rem` which fixed the issue. However, this introduced a new issue of the background becoming oval when the button is hovered. I added a transition to make the width back to `em` on hover. This creates a small transition that is rather aesthetic in my opinion. The larger buttons move less and the smaller buttons move more, creating a sense of inertia to the buttons. What do you think?

ong6 (commented):
LGTM @tlylt for the final say

tlylt:
> Hi @tlylt! I managed to align them but I had to introduce a small transition to mask the solution. The issue was that it was using `em` for width which made the width of the button change as the font size decreased. I changed it to `rem` which fixed the issue. However, this introduced a new issue of the background becoming oval when the button is hovered. I added a transition to make the width back to `em` on hover. This creates a small transition that is rather aesthetic in my opinion. The larger buttons move less and the smaller buttons move more, creating a sense of inertia to the buttons. What do you think?

Thanks @itsyme, I tried the animation. I think sometimes at certain positions the transition feels like a shaking animation such that the button just shakes from left to right continuously. This is not very obvious and only occurs when I put my cursor near the edge of the button. I could not capture this.

Ping @MarkBind/active-devs if anyone has any opinion on the aesthetic of the transition. I find it slightly out of place but ok to add it.

docs/css/main.css
```diff
@@ -81,8 +81,7 @@ mark {
 }
 
 .site-nav-top {
-    margin: 0.8rem 0;
-    padding: 0 12px 12px 12px;
+    padding: 0.7rem 0 0.5rem 0.8rem;
```
tlylt, line 84: This makes this PR a breaking change, because you don't just need to fix it here, but also in https://github.com/MarkBind/markbind/blob/21620cb2d527ba7cb33a4d569ce442b83194f9c2/packages/core/template/default/stylesheets/main.css
and I believe all the test sites will need this update. Is this change very obvious? I tried undoing this and somehow the title looks about the same? If we can omit this then let's revert this change.

itsyme: Alright! The change was mainly to reduce the spacing below the `site-nav-top`. However, I think we can omit this change if this change makes the PR a breaking change.

itsyme:
> Thanks @itsyme, I tried the animation. I think sometimes at certain positions the transition feels like a shaking animation such that the button just shakes from left to right continuously. This is not very obvious and only occurs when I put my cursor near the edge of the button. I could not capture this.

Ping https://github.com/orgs/MarkBind/teams/active-devs if anyone has any opinion on the aesthetic of the transition. I find it slightly out of place but ok to add it.

Thanks for testing it out @tlylt! I had another transition of just expanding the button (below). However, only the 0-th, 2-nd and 3-rd tier will expand as the 1-st tier has the same font size as the root, making it inconsistent. I added the movement to the right to make the transition consistent with all the buttons.

Screen Recording 2023-03-14 at 9 22 01 AM

jovyntls (commented):
@tlylt is this what you were referring to? 

https://user-images.githubusercontent.com/61113575/224882259-bddaf4bc-9fff-4b6b-899b-ed9f2636ba70.mov

I would also prefer if the transition didn't make the button move towards the right since it feels like the button is moving _away_ from my cursor (especially when my cursor approaches from the left). Personally I think a minimal aesthetic would be nice for the sitenav, but no strong opinions on this.

tlylt:
> @tlylt is this what you were referring to?
> 
> I would also prefer if the transition didn't make the button move towards the right since it feels like the button is moving _away_ from my cursor (especially when my cursor approaches from the left). Personally I think a minimal aesthetic would be nice for the sitenav, but no strong opinions on this.

Yes and agreed!

itsyme:
These are other possibilities we can explore:
- Without margin transition:  
Screen Recording 2023-03-14 at 11 43 48 AM

- Without any transition (background visibly rotates with icon):
Screen Recording 2023-03-14 at 11 47 38 AM

- Move to the right instead:
Screen Recording 2023-03-14 at 11 50 40 AM

Do let me know your thoughts!

itsyme:
2. Without any transition (background visibly rotates with icon):

https://user-images.githubusercontent.com/71922282/224893390-93afb95e-ef51-4991-ac06-bd14dcf2d8fb.mov

damithc:
Be careful with using rotation to direct attention, as the direction of the chevron has a meaning.

itsyme:
Hi @tlylt! I managed to fix the button alignments successfully by wrapping them in a `div`. There is one more potential change that could improve consistency of `site-nav`. When the scrollbar appears because the contents exceed the height of `site-nav`, it causes the dropdown buttons to move to the right. Do you think this is an issue?

https://user-images.githubusercontent.com/71922282/225170705-49bae94b-bf2a-4f35-bf01-91aba56ced30.mov

docs/_markbind/layouts/devGuide.md
```diff
@@ -3,7 +3,7 @@
 <div id="flex-body">
   <nav id="site-nav">
     <div class="site-nav-top">
-      <div class="fw-bold mb-2" style="font-size: 1.25rem;">Developer Guide</div>
+      <div class="fw-bold mb-2" style="font-size: 1.4rem;">Developer Guide</div>
```
tlylt, line 6: Same as previously commented, if we don't change this, does it still look ok?
By the way I think it's fine if it becomes a breaking change, just that we might need to delay this to the end of the semester or some time later.

tlylt:
> Hi @tlylt! I managed to fix the button alignments successfully by wrapping them in a `div`. There is one more potential change that could improve consistency of `site-nav`. When the scrollbar appears because the contents exceed the height of `site-nav`, it causes the dropdown buttons to move to the right. This existed even before this PR. Do you think this is an issue?

Good point, I think it would be great if there are some space allocated for the scrollbar such that its appearance and disappearance does not shift the layout. If it's easy enough to implement then pls feel free to add it in this PR. A separate PR would be fine as well.

docs/_markbind/layouts/devGuide.md
itsyme: Yup it does! Just a smaller difference in font size between the `site-nav-top` and 0th tier of the `site-nav`. I would prefer this PR to not be a breaking change for grading purposes. Perhaps I could open a new PR in the future for these breaking changes instead?

tlylt (approved):
LGTM, thanks @itsyme. For the breaking changes let me confirm again and we can think about adding them in if we decide to do a major version bump.

Can you help me:
- file the issue on the scrollbar
- update this branch with latest master to avoid any potential conflicts since I just merged another PR dealing with test files changes.

PR #2221, opened 2023-03-17
Refactor core package to use Node types

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Implements #2114 on the rest of the core package

**Overview of changes:**
- Use the newly introduced MbNode/TextElement/NodeOrText types throughout the core package. This allows us to remove unnecessary checks for attributes on nodes and enforce better typing throughout the package.
- Add a `parseHTML` utility method. Wrapping `cheerio.parseHTML` allows us to avoid further `as unknown as` typecasts throughout the codebase.

N/A

**Testing instructions:**
Should have no changes in behaviour

```
Refactor core package to use Node types

In #2201, we introduced a unifying type for nodes. Most of the core
package is not yet using the new types.

Let's refactor the core package to utilise these new types, and add a
utility parseHTML method that avoids further `as unknown as` typecasts
throughout the codebase. Doing so allows us to make full use of
TypeScript's type-checking abilities while avoiding code duplication.

For more information on the rationale of the unifying types, refer to
#2201.
```

---

- [x] No unrelated changes


packages/core/src/html/siteAndPageNavProcessor.ts
```diff
@@ -38,18 +39,15 @@ const SITE_NAV_DROPDOWN_ICON_ROTATED_HTML = '<div class="site-nav-dropdown-btn-c
  * otherwise it is replaced with one until it is unique.
  */
 export class PageNavProcessor {
-  uuidTextNode: DomElement | undefined;
-  constructor() {
-    this.uuidTextNode = undefined;
-  }
+  uuidTextNode?: DomElement;
```
EltonGohJH, line 42: Do we want to replace the type of this to be NodeOrText since parseHtml returns that?

packages/core/src/html/siteAndPageNavProcessor.ts
```diff
@@ -38,18 +39,15 @@ const SITE_NAV_DROPDOWN_ICON_ROTATED_HTML = '<div class="site-nav-dropdown-btn-c
...
 
-  renderPageNav(node: DomElement) {
-    [this.uuidTextNode] = cheerio.parseHTML(uuidv4()) as unknown as [DomElement];
-    cheerio(node).replaceWith(this.uuidTextNode as unknown as cheerio.Element);
+  renderPageNav(node: MbNode) {
+    [this.uuidTextNode] = parseHTML(uuidv4());
+    cheerio(node).replaceWith(this.uuidTextNode as MbNode);
   }
 
   finalizePageNavUuid(mainHtml: string | null, mainHtmlNodes: DomElement[], footnotesHtml: string) {
```
EltonGohJH, line 53: Same here.

EltonGohJH (approved):
LGTM!

packages/core/src/html/siteAndPageNavProcessor.ts
jovyntls: Nice catch! Amended to TextElement

packages/core/src/plugins/PluginManager.ts
```diff
@@ -205,13 +205,13 @@ export class PluginManager {
       .reduce((renderedContent, plugin) => plugin.postRender(frontmatter, renderedContent), content);
   }
 
-  processNode(node: DomElement) {
+  processNode(node: NodeOrText) {
```
raysonkoh, line 208: Just curious. Seems like in some places, `DomElement` was replaced by `MbNode` and in other places `DomElement` was replaced by `NodeOrTest`. How do you decide which to replace `DomElement` by?

jovyntls: @raysonkoh MbNode is a stricter type than NodeOrText, so I try to use MbNode wherever possible so that we can assert that the node contains some attributes using types instead of using the code logic (which is repetitive since the linter would enforce that we do so repeatedly). NodeOrText is used when we cannot guarantee what kind of node it is - i.e. it may be a TextElement, which is simpler/looser than MbNode and doesn't contain the same assertions. 
May be best illustrated with the example from parseHTML:
```typescript
export function parseHTML(html: string) {
  const node = cheerio.parseHTML(html);
  return node as unknown as NodeOrText[];
}
```
NodeOrText is used as we don't know what HTML string this method will be called with, which means that the parsed node might be a TextElement or an MbNode. Using NodeOrText allows us to do a typecast nearer the caller when we can guarantee that the node will be a MbNode:
```typescript
  const test = '<panel><div #header>test</div></panel>';
  const testNode = parseHTML(test)[0] as MbNode;
```
We can typecast here as we know that `test = '<panel><div #header>test</div></panel>'` is a node with a name, attributes, etc, and not a TextElement.

raysonkoh: Got it, thanks! I think it would be good to document this somewhere in the devdocs otherwise people might use `NodeOrText` or `MbNode` in inappropriate places.

jovyntls: Much of this is summarised in [node.ts](https://github.com/jovyntls/markbind/blob/ts-unify-node-types/packages/core/src/utils/node.ts); do you think it is sufficient, or it would be helpful to add the above example to the devdocs as well? @raysonkoh

PR #2234, opened 2023-03-25
Clean up & standardise TypeScript conventions

jovyntls:

- [x] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
As we've been doing quite a large-scale file-by-file migration to TypeScript (#1913), there is some inconsistency and redundancy in how we handle some types, exports, etc. With #2201 and #2221, we now have a more organised typing system and a rough "convention" that can be followed by TS files. 

This PR:
- standardises class exports
- standardises how we represent generic key-value types to use `Record`
	- `Record<K,V>` is interchangeable with `{ [key: K]: V }` syntax, but is more extensible, generalisable, and is an officially supported utility type in TS.
- fixes some typos in our docs, as many `js` files are now `ts` files
- removes unnecessary `as unknown as` and `any` typecasts, now that we have a more mature typing system

N/A

**Testing instructions:**
N/A

```
Clean up & standardise TypeScript conventions

The core package is almost fully migrated to TypeScript. There are
several inconsistencies and redundancies in how we represent common
types, handle typecasts, and export/import files. These inconsistencies
may cause confusion, reduce code quality, and prevent us from fully
taking advantage of static typing. 

As of #2201 and #2221, we have a more mature typing system that we can
utilise to fix these inconsistencies as well as a general set of
conventions that can be followed across TS files. 

Let's,
- Standardise class exports
- Standardise key-value types to use `Record<K, V>`
- Remove unnecessary `as unknown as` typecasts
- Reduce the use of `any` types
- Fix outdated docs that refer to `.js` files

Using Record is preferable as `Record<K,V>` is interchangeable with
`{ [key: K]: V }` syntax, but is more extensible and arguably more
readable. `unknown` and `any` types also introduce uncertainty and
unsafe typecasts into the codebase.
```

---

- [x] No unrelated changes


.eslintignore
```diff
@@ -9,69 +9,27 @@ packages/cli/src/lib/live-server/*
...
-packages/core/test/unit/plugins/default/plantuml.test.js
-packages/core/test/unit/plugins/default/shorthandSyntax.test.js
-packages/core/test/unit/plugins/default/tree.test.js
-packages/core/src/plugins/algolia.js
-packages/core/src/plugins/disqus.js
-packages/core/src/plugins/filterTags.js
-packages/core/src/plugins/googleAnalytics.js
-packages/core/src/plugins/mathDelimiters.js
+packages/core/src/lib/markdown-it/highlight/*.js
+packages/core/src/lib/markdown-it/utils/*.js
```
yucheng11122017, line 20: I think the packages/core/src/lib/markdown-it/plugins/markdown-it-icons.js needs to be added also! :)

jovyntls: There's a section on the markdown-it plugins below (around lines 34-37), would that give the desired behaviour of `.eslintignore` ?

yucheng11122017: Oh yes you are right!

ong6 (approved):
LGTM

tlylt (commented):
I briefly looked through and looks good to me. Curiously most of the changes in the form of`{ [key: string]: string };` to `Record<string, string>;` will make the type work better or? What's the advantage of Record

jovyntls:
> * `Record<K,V>` is interchangeable with `{ [key: K]: V }` syntax, but is more extensible, generalisable, and is an officially supported utility type in TS.

@tlylt 
- both `Record<K,V>` and `{ [key: K]: V }` are currently used in our codebase
- `Record<K,V>` is interchangeable with { [key: K]: V } syntax
- `Record<K,V>` is more extensible, and is an officially supported utility type in TS
- `Record<K,V>` is more flexible, and can support more types for the key (`K`). It supports generic types which is not supported in `{ [key: K]: V }`. 
	- e.g. if we have `type ColorName = "red" | "blue";` we can do `Record<ColorName, ColorVal>` but not `{ [key: ColorName]: ColorVal }`

packages/core/src/Page/PageConfig.ts
```diff
@@ -74,7 +74,7 @@ export class PageConfig {
     baseUrlMap: Set<string>;
     dev: boolean;
     faviconUrl?: string;
-    frontmatterOverride?: { [frontmatterName: string]: string },
+    frontmatterOverride?: Record<string, string>,
```
raysonkoh, line 77: While I do agree that `Record<K,V>` is a cleaner way to represent key-value types, I'm not sure if simply substituting `{ [key: K]: V }` with `Record<K,V>` would be more readable. 

In this case, I would actually prefer the previous type declaration because I would know that the key is actually `frontmatterName`, whereas if I saw it as `Record<string, string>`, I would not know what the meaning of the key is other than the fact that it is a `string`. Wdyt?

jovyntls: > In this case, I would actually prefer the previous type declaration because I would know that the key is actually `frontmatterName`, whereas if I saw it as `Record<string, string>`, I would not know what the meaning of the key is other than the fact that it is a `string`. Wdyt?

I agree with this point on the semantic value of the key!
It feels that most of the time the key is `id:` or `name:` which can be omitted without losing information, but in cases like this the key conveys information that may be valuable to the developer. What do you think of leaving a comment in such cases? e.g.
```typescript
{
	frontmatterOverride?: Record<string, string>,  // key: frontmatter name
}
```

packages/core/src/variables/VariableProcessor.ts
```diff
@@ -182,18 +182,16 @@ class VariableProcessor {
...
-    const includeElementAttribs = includeElement.attribs as { [s: string]: string };
+    const includeChildVariables: Record<string, string> = {};
 
     includeElement.children.forEach((child) => {
       if (child.name !== 'variable' && child.name !== 'span') {
         return;
       }
-      const childAttribs = child.attribs as { [s: string]: string };
-      const variableName = childAttribs.name || childAttribs.id;
+      const variableName = child.attribs && (child.attribs.name || child.attribs.id);
```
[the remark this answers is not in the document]
jovyntls: > if `child.attribs` exists, then `variableName` will be `child.attribs.name` (if it exists) or `child.attribs.id` (otherwise)?

Yup! `child` is of type DomElement so we can't assert that it has `name` and `attribs` attributes. If `child.attribs` doesn't exist then it will be caught in the `variableName` check in the next line.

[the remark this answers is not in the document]
tlylt: Can we improve the readability here? Is variableName a bool?

packages/core/src/Page/PageConfig.ts
jovyntls: No problem, just concerns about consistency between the two declarations. In that case should we leave it as such in these cases?

packages/core/src/variables/VariableProcessor.ts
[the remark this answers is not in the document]
jovyntls: Ah just realised a better way for this: `const variableName = child.attribs?.name || child.attribs?.id;`
Thanks for highlighting!

docs/userGuide/usingPlugins.md
```diff
@@ -76,7 +76,7 @@ MarkBind has a set of built-in plugins that can be used immediately without inst
 Only run plugins from sources that you trust. Do not run the plugin if the source/origin of the plugin cannot be ascertained.
 </box>
 
-Plugins come as `.js` files. To install an external plugin, simply put it in the `_markbind/plugins` folder. To use the plugin, update the `site.json` file the same way you did for built-in plugins.
+Plugins come as `.ts` (or `.js`) files. To install an external plugin, simply put it in the `_markbind/plugins` folder. To use the plugin, update the `site.json` file the same way you did for built-in plugins.
```
tlylt, line 79: Clarifying on usage of typescript plugin, otherwise ok.

jovyntls: Thanks for looking out for this - it looks like `.ts` plugins are not supported (ultimately a `.js` file is needed). Updated the docs!

tlylt (approved):

PR #2248, opened 2023-03-30
Generate .gitignore on init command

### LINKED ISSUE #1782: Generate sensible .gitignore when running `markbind init`
Given that most people who use markbind would typically be using git to version control a markbind project, currently it's a tiny bit of a hassle to manually create a `.gitignore` when generating a new site using `markbind init`.

Would be nice if a sensible .gitignore file is added. In particular, to ignore log files and maybe `.env` as well as `_site`.

Minimally, would be nice to provide a suitable example of a .gitignore file, perhaps as part of [GitHub .gitignore templates](https://github.com/github/gitignore)

**Additional context**
Might also want to consider generating a `package.json` file that will be able to track the `markbind-cli` version.
  - If `package.json` is added, the `.gitignore` file will need to ignore `/node_modules`

---

**TODO for PRs targeting this issue**
- [ ] generating a .gitignore when markbind init is called
- [ ] either skip or possibly override for the --convert flag (Preferably skip)
- [ ] update UG docs on this behavior

This is possibly breaking change

lhw-1:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Resolves #1782, and is a breaking change.

TODOs as listed by the original issue:

- [x] Generating a `.gitignore` when `markbind init` is called
- [X] Skip `.gitignore` generation for the `--convert` flag 
- [x] Update UG docs on this behavior
- [x] Add tests as necessary

This adds a `.gitignore` to the `default` template (only).

The proposed implementation is to skip the `.gitignore` generation when `markbind init` is called with `--convert` flag, but in the future, there may be a need for another action (e.g. merge / overwrite). In this case, please refer to the discussions in #1782 for the rationale on why merging / overwriting may not be a good idea. Also refer to the implementation [here]([4389512](https://github.com/MarkBind/markbind/pull/2248/commits/43895126a74b63543a12fc7894319606c11629a8)) if needed.

The original issue suggests a potential extension to generate a `package.json` as well.

**Testing instructions:**

Pull this PR and call `markbind init`. Check that `.gitignore` is generated when calling `markbind init`, but not when using the `--convert` flag.

```
Generate .gitignore on init command

Currently, when users initialize a site from scratch, we do not
generate .gitignore automatically.

This may cause unnecessary hassle for users using git as they
need to manually create a .gitignore, or add to a pre-existing
.gitignore, all the build files created by markbind.

Generating a .gitignore including common log and build files
from markbind allows users to use our .gitignore without
having to manually find which generated files are unnecessary.

Skipping .gitignore when --convert flag is used prevents
accidentally overwriting a pre-existing .gitignore file.
```

---

- [x] No unrelated changes


packages/core/src/Site/template.ts
```diff
@@ -3,16 +3,21 @@ import path from 'path';
...
 const requiredFiles = ['index.md', 'site.json', '_markbind/'];
+const omitOnConvertFiles = ['.gitignore'];
 
 const PATH_TO_TEMPLATE = '../../template';
 
 export class Template {
   root: string;
   template: string;
+  toConvert: boolean;
+  omitFiles: Array<string>;
```
jovyntls, line 14: Are these fields needed since we pass an `overwrite: false` option passed to the `copySyncWithOption` function, which may automatically avoid overwriting an existing .gitignore? (And we only want to omit copying a .gitignore if it already exists)

jovyntls (changes requested):
Hi @lhw-1 , just one clarification!

packages/core/src/Site/template.ts
lhw-1: Good point!

This implementation ensures that regardless of pre-existing files, if the `--convert` flag is used, we will entirely skip `.gitignore` (and potentially other files declared in `omitOnConvertFiles` if we want to extend to `package.json` as well later on).

If we instead rely on `overwrite: false` alone, this would mean that we will generate a `.gitignore` when converting documents without a `.gitignore` in their root repository. 

I don't think we came to a conclusion for this as mentioned by @tlylt [here](https://github.com/MarkBind/markbind/issues/1782#issuecomment-1488331079), but you're right that this is a more complex implementation. I can revert these for now, and if we later decide to implement skipping files for `--convert` altogether, we can refer back to these changes 👍 

Does that sound good @jovyntls?

jovyntls: Ah I see, sounds good to me - I think it's good to have add gitignore in convert (if there isn't an existing one)

tlylt:
Would be good to add a note in https://markbind-master.netlify.app/userguide/gitignorefile to explain this new behavior

packages/core/src/Site/index.js
```diff
@@ -177,9 +177,9 @@ class Site {
    * @param rootPath
    * @param templatePath
```
tlylt, line 178: add jsdoc param here, and i would suggest `isConvert`

lhw-1: Just opened another PR #2267 that will remove the need for JSDocs here!

packages/core/src/Site/template.ts
lhw-1: Changes reverted in [1297fbe](https://github.com/MarkBind/markbind/pull/2248/commits/1297fbe23477cd06027f0446a03a5d63c91f0548)!

docs/userGuide/gitignoreFile.md
```diff
@@ -45,3 +45,11 @@ _site/
 .idea/*
 *.iml
 ```
+
+<box type="info">
+
+All pre-existing `.gitignore` file will not be overwritten if you are converting using the `markbind init --convert` command.
```
jovyntls, line 51:
```suggestion
A pre-existing `.gitignore` file will not be overwritten if you are using the `markbind init --convert` command. 
```

jovyntls (commented):
LGTM, just documentation nits!

docs/userGuide/gitignoreFile.md
lhw-1: Updated in [6f09122](https://github.com/MarkBind/markbind/pull/2248/commits/6f091229ca498782376f4be82f5e6b3b1257c035)!

jovyntls (approved):
LGTM, thanks @lhw-1 !

tlylt (approved):

Issue #2211, opened 2023-03-14
Includes does not render markdown when fragment is a variable


yucheng11122017:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Tell us about your environment

Windows 10

### MarkBind version

V4.1.0

### Describe the bug and the steps to reproduce it

When using includes with hash, markdown is not rendering correctly

This is seen in the UG reader facing feature for classes attributes and identifiers and questions. 

![image](https://user-images.githubusercontent.com/77204346/224975430-ca1fd150-8170-48d0-b182-8b30f41d6f71.png)
![image](https://user-images.githubusercontent.com/77204346/224975468-be64606d-e178-4b50-b3c6-16ecbbd52ba4.png)

### Expected behavior

Expected markdown to render correctly when using includes

### Anything else?
Remove documentation in UG introduced in #2258 regarding this bug if fixed
_No response_

damithc:
@yucheng11122017 The common reason for this problem is the lack of a blank line just above the start of the markdown text. I assume that's not the case here?

damithc:
@yucheng11122017 I see. Thanks for checking. Looks like a regression that was introduced after the last release (i.e., I have not encountered this before).

yucheng11122017:
@jovyntls Oh yes I think you are right! I think the markdown doesn't render because the code is within the variable for the include. Hence when it is extracted, it remains as text instead of markdown. 

Eg. 
```
<include src="codeAndOutput.md" boilerplate >
<variable name="highlightStyle">markdown</variable>
<variable name="code" id="list-example">
* Apply to the list item itself like so {.text-success #list-item-id}
  * Curly groups after newlines apply to the closest nested list {.text-danger}
{.bg-light}
* Curly groups two lines after the last line apply to the top most list

{.alert-info}
</variable>
</include>
```

The `list-example` will not get rendered and hence when it is extracted out, it will just remain as text. 

I don't believe that we would want to change this behavior as well. 

I've been trying to change the order to do the markdown rendering after the taking out the part of the hash but this will result in some issues 

- Trim does not working properly with this ordering 
- An additional `<p>` tag is added if inline option is not enabled

Would this issue be worth fixing?

jovyntls:
> Would this issue be worth fixing?

I believe so! It would be nice to polish how we handle `includes` since it's one of our more powerful features.

[...]

PR #2258, opened 2023-04-02
Add documentation that markdown is not rendered when including variables

yucheng11122017:

- [X] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Related to #2211

**Overview of changes:**
Adding documentation about the bug where markdown is not rendered when including variable
Included it in a panel in includes so users don't have to view it if not relevant

I did not include the reason for this error because its a bit complicated to explain and users wouldn't need to understand it.

**Testing instructions:**

Document bug where including variables does not render markdown

---

- [X] No unrelated changes


docs/userGuide/syntax/includes.md
```diff
@@ -61,6 +61,62 @@ The `<include>` mechanism can be used inside any MarkBind source file (even insi
 </span>
 </include>
 
+<panel header="Markdown not rendered when including a variable">
```
damithc, line 64: The heading should let the reader decide quickly between expanding and reading the panel or skipping it. This one is hard to decide (is this a tip, advanced technique, caveat, etc.?), and the reader will have to expand the panel to figure out what it's about.

yucheng11122017: Hmmm how about "If trying to include a variable with markdown" or something like that?

docs/userGuide/syntax/includes.md
```diff
@@ -61,6 +61,62 @@ The `<include>` mechanism can be used inside any MarkBind source file (even insi
...
+<variable id="fragment-id">
+  
+  ### I am a header
+</variable>
+```
+
+Code that uses include fragment:
+```html
+<include src="index.md#fragment-id" />
+```
```
damithc, line 85: Something is not right here. I don't think variables are meant to be used as include fragments in the first place.  Although it would be nice 'trick' to be able to do so, I don't think it is an advertised feature. Variable values are inserted using the `{{ v_name }}` instead. Does our UG say variables can be included in this way?

image

yucheng11122017: Oh I see I see. In that case, I will close this PR and the corresponding issue :)

PR #2270, opened 2023-04-08
Migrate Site/index.js to TypeScript

jovyntls:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Part of #1913

**Overview of changes:**
Migrates Site/index.js to TS

If we want to refactor how the methods at the end of Site/index.js are handled, I'd suggest we do a separate refactor of these methods after migrating this file to TS. Rationale of having these methods defined with `Site.prototype`: see #979

**Testing instructions:**
N/A

(rebase commit)

---

- [x] No unrelated changes


raysonkoh (commented):
Thanks for the hardwork! Added some preliminary comments.

packages/core/src/Site/index.ts
```diff
@@ -235,33 +314,20 @@ class Site {
       return this.siteConfig;
     } catch (err) {
       throw (new Error(`Failed to read the site config file '${this.siteConfigPath}' at`
-        + `${this.rootPath}:\n${err.message}\nPlease ensure the file exist or is valid`));
+        + `${this.rootPath}:\n${(err as Error).message}\nPlease ensure the file exist or is valid`));
     }
   }
 
-  listAssets(fileIgnore) {
+  listAssets(fileIgnore: Ignore) {
```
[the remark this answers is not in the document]
jovyntls: It's the type of the manager used by the [`ignore` package](https://www.npmjs.com/package/ignore) which we're currently using to ignore files. Kind of like the entrypoint class that we can call methods on, e.g. has an `add()` method to add files to a set of ignored paths

packages/core/src/Site/index.ts
```diff
@@ -101,9 +121,85 @@ const ABOUT_MARKDOWN_DEFAULT = '# About\n'
 const MARKBIND_WEBSITE_URL = 'https://markbind.org/';
 const MARKBIND_LINK_HTML = `<a href='${MARKBIND_WEBSITE_URL}'>MarkBind ${MARKBIND_VERSION}</a>`;
 
-class Site {
-  constructor(rootPath, outputPath, onePagePath, forceReload = false,
-              siteConfigPath = SITE_CONFIG_NAME, dev, backgroundBuildMode, postBackgroundBuildFunc) {
+/*
+ * A page configuration object.
+ */
+type PageCreationConfig = {
```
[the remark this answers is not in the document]
jovyntls: Not sure about moving the types to a file, because they largely fall into 3 different purposes (page creation, page generation, deploy options) which are all quite different so extracting them to the same file may be a bit weird unless we're going for a `types.ts` kind of file for all types. We could use multiple files for the types with 1-2 types per file, which would reduce the length of Site/index.ts, but I feel that it doesn't add much in terms of understandable-ness especially since most of these types are only used once or a few times in Site/index.ts. But then again I don't think it hurts either, besides having the types further from where they are defined 🤔  Wdyt?

packages/core/src/Site/index.ts
```diff
@@ -887,15 +959,18 @@ class Site {
...
    */
-  async handlePageReload(oldAddressablePages, oldPagesSrc, oldSiteConfig) {
+  async handlePageReload(oldAddressablePages: AddressablePage[], oldPagesSrc: string[],
+                         oldSiteConfig: SiteConfig) {
     this.collectAddressablePages();
 
     // Comparator for the _differenceWith comparison below
-    const isNewPage = (newPage, oldPage) => _.isEqual(newPage, oldPage) || newPage.src === oldPage.src;
+    const isNewPage = (newPage: AddressablePage,
+                       oldPage: AddressablePage) => _.isEqual(newPage, oldPage)
```
raysonkoh, line 968: Minor nit but I would prefer if the RHS expression of `=>` is all on the same line.

packages/core/src/Site/index.ts
```diff
@@ -1611,6 +1686,21 @@ class Site {
     const time = new Date().toLocaleTimeString(this.siteConfig.locale, options);
     this.variableProcessor.addUserDefinedVariableForAllSites('timestamp', time);
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, class-methods-use-this
+  rebuildPagesBeingViewed(_currentPageViewed: string) {
```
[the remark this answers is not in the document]
raysonkoh: Yup, makes sense. That can be done in a separate PR.

packages/core/src/Site/index.ts
```diff
@@ -1184,7 +1256,7 @@ class Site {
     context.numPagesGenerated += 1;
 
     if (pageGenerationQueue.length) {
-      pageGenerationQueue.pop()();
+      pageGenerationQueue.pop()!();
```
raysonkoh, line 1259: Reason for this change?

packages/core/src/Site/index.ts
```diff
@@ -571,10 +642,12 @@ class Site {
       this.variableProcessor.addUserDefinedVariable(base, 'MarkBind', MARKBIND_LINK_HTML);
 
       const $ = cheerio.load(content, { decodeEntities: false });
-      $('variable,span').each((index, element) => {
+      $('variable,span').each((_index, element) => {
         const name = $(element).attr('name') || $(element).attr('id');
 
-        this.variableProcessor.renderAndAddUserDefinedVariable(base, name, $(element).html());
+        if (name) {
```
raysonkoh, line 648: Could this result in a different behavior since this is a possible change in logic?

packages/core/src/Site/index.ts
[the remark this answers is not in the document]
raysonkoh: Ahh okay. I think it is fine to leave it then since the types are only used in this file.

packages/core/src/Site/index.ts
jovyntls: `pageGenerationQueue.pop()` may be undefined if the array is empty, which TS will complain about.
We check that this isn't the case using the `.length` check and so we can assert that it's not undefined using `!`

packages/core/src/Site/index.ts
jovyntls: Previously a non-user-facing error would be thrown if the user provides a variable without a name; this will instead silently ignore the variable that has no name.
We could either fail silently, or explicitly throw a user-facing error. Which do you think is better?

jovyntls: Yup! I think logging a warning would be helpful. This would likely without specifying which file it came from though, since that would involve passing down a context object to wherever the addVariables method is called which may add a lot more complexity. Does that sound good?

raysonkoh: Sounds good!

packages/core/src/Site/index.ts
```diff
@@ -914,18 +983,14 @@ class Site {
         || !_.isEqual(oldSiteConfig.plantumlCheck, this.siteConfig.plantumlCheck);
 
     if (isGlobalConfigModified() || !_.isEmpty(addedPages) || !_.isEmpty(removedPages)) {
-      await this.removeAsset(removedPages);
+      this.removeAsset(removedPages);
```
tlylt, line 986: Do we not need to await this?

jovyntls: Ah you're right, we need to await it. Added `async` to the 3 method signatures to reflect the asynchronous status!

packages/core/src/Site/index.ts
```diff
@@ -1611,6 +1678,21 @@ class Site {
     const time = new Date().toLocaleTimeString(this.siteConfig.locale, options);
     this.variableProcessor.addUserDefinedVariableForAllSites('timestamp', time);
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, class-methods-use-this
+  async rebuildPagesBeingViewed(_currentPageViewed: string) {
+    throw new Error('Method not implemented.');
```
tlylt, line 1684: Are these added for the sake of the type? The message here is a little confusing to understand

jovyntls: Yup for types - reasoning [here](https://github.com/MarkBind/markbind/pull/2270#discussion_r1164425215)

tlylt: I see, will need to open an issue to KIV this. if not we could use a more suitable message, such as "Method implemented via prototype".

ang-zeyu:
@tlylt as it will be quite a bit of work to undo the revert while keeping track of new PRs editing the file https://github.com/MarkBind/markbind/pull/2308#issuecomment-1646846693, I suggest adding a small follow up comment on top that prohibits any changes to this file until this is fixed (and bumping the priority).

The alternative is letting a portion of the typescript migration work go to waste (also an ok approach).
I'll leave this up to you.

Issue #1866, opened 2022-03-31
Markbind init --convert improvements


kaixin-hc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Any related issues?

Other suggestions for MarkBind init command #1782, #1844

### Tell us about your environment

MacOS Monterey

### MarkBind version

3.1.1

### What did you do? Describe the bug

Created a repository with .md files (https://github.com/kaixin-hc/animal-behaviour-LSM1303) and ran `markbind init --convert` 

### Steps to reproduce

Clone my repository, navigating to the branch without MarkBind  (https://github.com/kaixin-hc/animal-behaviour-LSM1303/tree/toTestInitOn), and run `markbind init --convert`. Or make your own repository with file names that have spaces.

### Expected behavior

Site is created and all elements are natural and intuitive.

### Actual behavior

* Default files(the contents folder) are still created and included in the sitenav by default; since it is a conversion of an existing repo, you would expect these files not to be created and included
* The sitenav does not work as expected (Files with spaces are not rendered correctly)
* Instead of the default navigation sidebar name being "Template", it makes more sense for it to be something like "Contents"
* Titles are not auto-generated for content pages.

Nits
* Personally, I feel like the default brand image makes it look like there was an error in rendering the favicon, rather than a default placeholder being selected by default.
* The auto-generated default.md file has a few blank lines after the navbar; I think most linters will flag this as an issue and we can do without them.

ong6:
I agree with the points that you have raised. I believe that the convert feature is quite important since most new users would be using this to help generate a site from markdown files. 

> * Default files(the contents folder) are still created and included in the sitenav by default; since it is a conversion of an existing repo, you would expect these files not to be created and included

I think there's an option to convert with a template? Does this still happen if you use the Minimal template?

tlylt:
> I agree with the points that you have raised. I believe that the convert feature is quite important since most new users would be using this to help generate a site from markdown files.
> 
> > * Default files(the contents folder) are still created and included in the sitenav by default; since it is a conversion of an existing repo, you would expect these files not to be created and included
> 
> I think there's an option to convert with a template? Does this still happen if you use the Minimal template?

Conversion with a template doesn't seem to be documented yet. I believe we worked on it in https://github.com/MarkBind/markbind/pull/1718

@kaixin-hc any reason the above draft PR was closed? 🤔

kaixin-hc:
[...]

PR #2278, opened 2023-04-12
Fix minor UI issues with default.md

lhw-1:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Addresses two of the issues related to `_markbind/default.md` brought up in #1866.

Specific improvements:
- [X] Change the sidebar name from "Template" to "Contents" in `default.md`
- [x] Remove generation of blank lines after the navbar in `default.md`

**Testing instructions:**
Try `markbind init` or `markbind init --convert` on a pre-existing set of documents, and ensure that changes are reflected.

Fix minor UI issues with default.md

---

- [ ] No unrelated changes


tlylt:
@lhw-1 need resolve conflict

packages/core/src/Site/index.js
```diff
@@ -366,7 +366,7 @@ class Site {
     let footer;
     if (fs.existsSync(wikiFooterPath)) {
       logger.info(`Copied over the existing ${WIKI_FOOTER_PATH} file to the converted layout`);
-      footer = fs.readFileSync(wikiFooterPath, 'utf8');
+      footer = `\n${fs.readFileSync(wikiFooterPath, 'utf8')}`;
```
tlylt, line 369: What's this about? Possible to show screenshots of before/after for your changes?

lhw-1: (Side note: This change is shown as outdated because `Site/index.js` has been converted to `Site/index.ts`.)

In order to remove unnecessary newlines before the footer in `default.md`, I included nunjucks whitespace removers into the `set` and `endset` tags, like this, in `Site/siteConvertLayout.njk`:

```html
{%- set defaultFooter -%}
...
{%- endset -%}
```

This will ensure proper behavior for the default footers:

```html
<footer>
  <!-- Support MarkBind by including a link to us on your landing page! -->
  <div class="text-center">
    <small>[Generated by {{MarkBind}}]</small>
  </div>
</footer>
```

But for custom footers, as we are stripping the newlines from before the `set` tags as well (there are two), it results in this behavior:

```html
<footer>Custom footer
</footer>
```

image

You can check this as well by pulling this branch, removing this line, and then updating the tests to see the results!

Adding the newline here will ensure that the current behavior is not modified, i.e. this will only affect the custom footer to have an extra newline, producing this:

```html
<footer>
  Custom footer
</footer>
```

lhw-1: All in all, this change is quite minor, so if we are fine with leaving generated custom footers to be without the newline then I can revert this newline addition - the user can modify the footer later on anyways, since this is only on `init`. I initially added this line just to keep to the current behavior :+1:

tlylt: If there's a problem we should fix it, doesn't matter whether this is minor or not

tlylt: For the files/screenshots that you mentioned above, could you indicate the filename/where they are from? Are they from .md or .html?

Could you provide a summary of what was the original behavior/output, what's the behavior/output without this newline, what's the final output with all the changes that you have made? I'm confused reading the explanation above.

packages/core/src/Site/siteConvertLayout.njk
```diff
@@ -38,13 +38,12 @@
   </nav>
 </div>
 
-{% set defaultFooter %}
```
tlylt, line 41: Is there any issue with making the `{%-` change here? I see that you moved this into `<footer>`

packages/core/src/Site/index.js
tlylt: Also, is this change fixing `Remove generation of blank lines after the navbar in default.md` ? So the blank lines after the navbar is due to the footer? That's seems confusing.

lhw-1:
Here is a summary of the changes made targeting "Remove generation of blank lines after the navbar in `default.md`".

In the current version of MarkBind, when the user calls `markbind init --convert` with a pre-existing set of documents, the section between the navbar and the footer in the generated `default.md` looks like this:

```nunjucks
  ...
  </nav>
</div>



<footer>
  
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {{MarkBind}}]</small>
</div>

</footer>
```

The issue brought up in #1866 is that there are unnecessary blank lines here that are generated during our `convert` process. After some investigation, the blank spaces are generated because we are currently using a nunjucks variable in our `siteConvertLayout.njk`, which dictates what the resulting `default.md` should look like:

```nunjucks
  ...
  </nav>
</div>

{% set defaultFooter %}
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {% raw %}{{MarkBind}}{% endraw %}]</small>
</div>
{% endset %}

<footer>
  {{ footer or defaultFooter }}
</footer>
```

---

In the case where the `defaultFooter` declared above is used, the blank lines before and after the `{% set %}` tags will remain in the file after the nunjucks variable is substituted into the `<footer>` tag below. This is why I directly moved it into the `<footer>` tag like this in the `siteConvertLayout.njk`:

```nunjucks
  ...
  </nav>
</div>

<footer>
{% set defaultFooter %}
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {% raw %}{{MarkBind}}{% endraw %}]</small>
</div>
{% endset %}
{{ footer or defaultFooter }}
</footer>
```

And this will result in the following `default.md` generated:

```nunjucks
  ...
  </nav>
</div>

<footer>


<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {{MarkBind}}]</small>
</div>

</footer>
```

The blank lines between the navbar and the footer are now reduced to a single blank line (which is acceptable), but there are now new unnecessary blank lines introduced within the `<footer>` itself. These are leftover newline characters from after the nunjucks variables are rendered. Hence, I added "outer `-`" to the `{% set %}` tags to remove all blank lines before the `{% set %}` and after the `{% endset %}`. Here is what I tried with `siteConvertLayout.njk`:

```nunjucks
  ...
  </nav>
</div>

<footer>
{%- set defaultFooter %}
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {% raw %}{{MarkBind}}{% endraw %}]</small>
</div>
{% endset -%}
{{ footer or defaultFooter }}
</footer>
```

And the resulting `default.md` generated:

```nunjucks
  ...
  </nav>
</div>

<footer>
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {{MarkBind}}]</small>
</div>

</footer>
```

Since there is still one newline within the `<footer>` itself, I added "inner `-`" to the `{% set %}` tags to further remove all the newlines encased between the `{% set %}` and the `{% endset %}` tags. Here is what I tried with `siteConvertLayout.njk`:

```nunjucks
  ...
  </nav>
</div>

<footer>
{%- set defaultFooter -%}
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {% raw %}{{MarkBind}}{% endraw %}]</small>
</div>
{%- endset -%}
{{ footer or defaultFooter }}
</footer>
```

And the resulting `default.md` generated:


```nunjucks
  ...
  </nav>
</div>

<footer>
<!-- Support MarkBind by including a link to us on your landing page! -->
<div class="text-center">
  <small>[Generated by {{MarkBind}}]</small>
</div>
</footer>
```

This is the behavior we would like to see, without unnecessary blank lines. 

---

Unfortunately, because we remove the inner blank lines, this will affect custom footers that users may use. When I updated our functional tests, the `footer` for many of them were modified. For example, in the [`<footer>` section of `packages/cli/test/functional/test_site_convert/test_basic_convert/expected/Home.html`](https://github.com/MarkBind/markbind/blob/master/packages/cli/test/functional/test_site_convert/test_basic_convert/expected/Home.html#L68-L70), the code was modified from:

```
<footer>
  Custom footer.
</footer>
```

To:

```
<footer>Custom footer.
</footer>
```

In order to keep the behavior of custom `<footer>` consistent with previous behavior (which adds a newline), I modified `Site/index.ts` [here](https://github.com/MarkBind/markbind/pull/2278/files#diff-1579d51321badfb0388ed10d49bfa7ee76f2983bcbc4efdfb7e4733a8871700eR432) to have a newline added when there is a pre-existing footer in the documents being converted.

packages/core/src/Site/index.js
lhw-1: I've listed out the changes and the reason for these changes in a separate comment, hopefully that should be more comprehensible.

packages/core/src/Site/siteConvertLayout.njk
lhw-1: As shown in the summary of changes I made, `{%-` is intentional and is to remove the newlines before the `{% set %}` tag.

tlylt (approved):
LGTM

tlylt:
@lhw-1 possible to update the outdated content in here? https://markbind-master.netlify.app/devguide/bootcamp/exploremarkbind#modify-site-structure-and-configuration

Should have done so in your previous PR updating the template.

packages/core/src/Site/siteConvertLayout.njk
tlylt, line 1: line 23 
>       <div class="fw-bold mb-2" style="font-size: 1.25rem;">Template</div>

tlylt (commented):
Can you have another look at packages/core/src/Site/siteConvertLayout.njk to ensure it's up-to-date?

packages/core/src/Site/siteConvertLayout.njk
lhw-1: Fixed in [62bc902](https://github.com/MarkBind/markbind/pull/2278/commits/62bc9022664cf4addbe5b47a668e9458e342e7fb).

lhw-1:
> Can you have another look at packages/core/src/Site/siteConvertLayout.njk to ensure it's up-to-date?

Thanks for the catch! There does seem to be inconsistencies between `default.md` and `siteConvertLayout.njk`. I've just updated it in [09e7264](https://github.com/MarkBind/markbind/pull/2278/commits/09e72646f0d8c5bf6a899135046bbb33f0ef7706).

On another note, I wonder if we can work on a further refactor to have `siteConvertLayout.njk` dictate the behavior for the `default.md` generated during both `init` and `init --convert` (the former is dictated by the `default.md` in the templates directory, while the latter is dictated by `siteConvertLayout.njk`). Given that default values can be specified in `siteConvertLayout.njk`, it might be a better idea to consolidate them into a single file. What do you think?

tlylt:
> On another note, I wonder if we can work on a further refactor to have `siteConvertLayout.njk` dictate the behavior for the `default.md` generated during both `init` and `init --convert` (the former is dictated by the `default.md` in the templates directory, while the latter is dictated by `siteConvertLayout.njk`). Given that default values can be specified in `siteConvertLayout.njk`, it might be a better idea to consolidate them into a single file. What do you think?

Good suggestion and yes can explore. I wonder if there will be a case where the two (default and convert) will deviate further in the future. From the perspective of the functionality, I think it feels more natural to have "default.md" dictate "siteConvertLayout.njk"

PR #2225, opened 2023-03-18
Update default template with UG links and useful information

lhw-1:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [X] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

**Overview of changes:**

Addresses a few of the main outcomes of the discussions in #2214, and updates the `default` template to include links to the main sections of the User Guide. Also possibly addresses #519.

Here is a preview of what it currently looks like:

image

I've kept the new look to be as simple as possible, and have modelled it based partially on the MarkBind [User Guide - Authoring Contents](https://markbind.org/userGuide/authoringContents.html) section and partially on the [Gatsby initial site](https://egghead.io/lessons/gatsby-use-npm-init-gatsby-to-create-an-initial-gatsby-site).

This is just a possible direction we could take the default template towards - if there are any suggestions to improve this current look, or if there are any concerns / issues with this, please highlight as needed!

**Anything you'd like to highlight/discuss:**

We should keep in mind to keep the template as simple as possible while including the main features that a user using `markbind init --convert --template default` would want to have in their site. (E.g. siteNav and pageNav are essential for this - any other suggestions are welcome.)

**Testing instructions:**

Run `markbind init` in a new directory to initialize a site built using the `default` template. 

Alternatively, you can view the deployed version of the template [here](https://lhw-1.github.io/markbind-2225/).

**Proposed commit message: (wrap lines at 72 characters)**
```
Update default template with UG links and useful information

The current default template contains filler content and
component usage that is not well explained.

This could cause confusion to users who are starting with the
default template from scratch for their site.

Updating the default template allows us to help guide the newer
users, and since the content is overriden when being converted,
will not affect users who are converting pre-existing documents.

Let's add links to our User Guide, and re-organize the information
given to the users in a more succinct manner, while also
showcasing MarkBind components.
```

---

**Checklist:** :ballot_box_with_check:

- [X] Updated the documentation for feature additions and enhancements
- [X] Added tests for bug fixes or features
- [X] Linked all related issues
- [X] No unrelated changes

PR #2288, opened 2023-04-16
Update DG to be consistent with default template

lhw-1:

- [X] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Updates the [Onboarding Bootcamp](https://markbind.org/devdocs/devGuide/bootcamp/intro.html) to be consistent with changes made in #2225. Mainly wording and image updates to the Developer Guide.

As the `pageNav` of the frontmatter in the default template was changed from 2 to 4, I've modified the instructions to accommodate this change: instead of changing it from 2 to 1, the readers are now expected to try changing it from 4 to 5, which will show the sub-headings in the panel. I've also included a "before" image for this section to better highlight the differences. Following this, should we include a before vs after for all the relevant sections?

Also, after the removal of `title` from `site.json` in #2128, the `title` property is not generated by default, so I've tweaked the wording to be more suitable (and included a panel to the `site.json` page in the UG for reference).

**Testing instructions:**

Update DG to be consistent with default template

---

- [ ] No unrelated changes


docs/devGuide/bootcamp/exploreMarkBind.md
```diff
@@ -327,7 +331,9 @@ The `site.json` file is autogenerated and located in the root directory of the s
 
 1. Open the `site.json` file in your `mb-dev-xxx` folder.
 1. Change `"titlePrefix": "",` to `"titlePrefix": "mb-dev-xxx",`.
-1. Change `"title": "Landing Page"` to `"title": "Home"`.
+1. Under `pages`, add `"title": "Home"`.
+
+<panel src="{{baseUrl}}/userGuide/siteJsonFile.html#pages" header="**`site.json` File: `pages`**" type="info" minimized></panel>
```
tlylt, line 336: image
https://deploy-preview-2288--markbind-master.netlify.app/devguide/development/workflow#adding-intra-site-links-to-documentation

docs/images/bootcamp/logo_and_nav.png
tlylt, line 1: arent you changing "template" to "contents" in the other PR?

tlylt (changes requested):

docs/images/bootcamp/logo_and_nav.png
lhw-1: Thanks for the catch! Updated the images (will update once more to have the highlighting) in [e4e217c](https://github.com/MarkBind/markbind/pull/2288/commits/e4e217c5143b343c3a078a52acd7107d8a0a9b3e).

docs/devGuide/bootcamp/exploreMarkBind.md
lhw-1: Updated in [786dc91](https://github.com/MarkBind/markbind/pull/2288/commits/786dc91c2eb79d973ee2866080f35e9749bf5f57).

docs/images/bootcamp/logo_and_nav.png
lhw-1: Updated with highlighting in [50b4d8f](https://github.com/MarkBind/markbind/pull/2288/commits/50b4d8ff2183b7bf944f25cddaacc32e0c651f9d).

docs/devGuide/bootcamp/exploreMarkBind.md
tlylt: Sorry I think my previous reply is not accurate. It seems like somewhere below in this page I did use a panel referencing content in the userguide and it works fine. So using a panel should be ok for some reason.

lhw-1: The original `panel` I included was like this:

```html
<panel src="{{baseUrl}}/userGuide/siteJsonFile.html#pages" header="**`site.json` File: `pages`**" type="info" minimized></panel>
```

Which didn't work because I was using `html` instead of `md` for the `src` - the panel you used (I think it's this one) is using `md`, which is why I think it works:

```html
<panel src="{{baseUrl}}/userGuide/deployingTheSite.md#markbind-action-tip" header="**MarkBind Action**" type="info" expanded/>
```

But using `src="{{baseUrl}}/userGuide/siteJsonFile.md#pages"` didn't display properly either, so I changed it to a normal anchor. I felt that either way as long as the developer has some convenient access to the link it shouldn't matter for this particular PR. (Of course, if a panel is better for readability, I can look into making sure the panel version works)

Regardless, I will be investigating this issue - I'll either post it as an issue or address it somewhere else (potentially as a Documentation update if this is another way to link between DG and UG).

tlylt: Interesting. I think the problem is this:
- include only works if you include a fragment that is properly enclosed (meaning you created a div/span to wrap the content)
  - you used the heading 'pages' but that is not supported (because our heading is not auto-wrapping the content below it)
- as for .md vs .html, I think needs further clarification. I see that there's one instance of .html being used in the panel docs:
image
but that is actually referencing a .html document. In other cases, it seems to be that panel needs to reference .md, or whatever actual type of file the src is. But this is not clear from the docs. On the flip side I vaguely remember .md and .html is interchangeable when using 'include', so we might be able to support that as well, for the special case of '.md'

Yes, please help to raise an issue for this!

tlylt (approved):

Issue #1756, opened 2022-02-08
General Refactoring work 


ong6:
**Is your request related to a problem?**

Feels like the current core/site/index.js file is getting a little long (over 1.5k lines), making it hard to read and find important information, what is the general consensus around refactoring it?

**Describe the solution you'd like**

I propose we refactor the core/site/index.js file into all the different commands such as deploy, convert etc. Each command can be a separate class that contains its own functionality, while site can still exist as a the main driver class. This way, we can easily view all the code related to each command and make adding new commands easier. 

Though I'm not sure how doable this is since everything looks very tightly coupled. Any suggestions or ideas on how to go about refactoring this would be great!

**Describe alternatives you've considered**

Nil

**Additional context**

Nil

jonahtanjz:
Related to #1495

lhw-1:
Let me try tackling this issue - I'll post any updates / findings if I run into issues with decoupling and refactoring :sweat_smile: 

(Keeping this in mind: Completing TypeScript migration of `Site/index.js` first might be preferrable as mentioned in #1495 ([comment](https://github.com/MarkBind/markbind/issues/1495#issuecomment-1207165754)) by @ang-zeyu)

PR #2308, opened 2023-06-14
Update Site to generate default.md based on template

### LINKED ISSUE #2293: Consider refactoring generation of `default.md` into a single file

### Any related issues?

#2278

CLI, Code Quality

Initially brought up during a discussion in #2278, we can consider refactoring and combining the two ways of generating `default.md` in the site generated by `markbind init` commands.

Currently, there are two ways that a site is generated: `markbind init` generates a MarkBind site from a given template (`default` template is used if none is specified), and `markbind init --convert` converts a pre-existing set of documents into a MarkBind site. 

For the former, we simply use the `default.md` specified in the default template, which can be currently found in `packages/core/template/default/_markbind/layouts/default.md`. This behavior is dictated in `packages/core/src/Site/template.ts`, where we directly copy over files from the default template to the generated site.

For the latter, we use the `siteConvertLayout.njk` found in `packages/core/src/Site/siteConvertLayout.njk`, where custom values from the user's pre-existing site can be inserted into the layout. These values are given default values to fall back on if the user does not have custom values. 

The core issue is that much of these two files are similar, and given that we can specify default values within `siteConvertLayout.njk`, it feels redundant to have the current `default.md` from a design perspective. An example of an inconvenience is highlighted within #2278, where some of the more recent changes (at the time of writing) were specified in `default.md` but not in `siteConvertLayout.njk`, leading to unexpectedly different behaviors for `markbind init` and `markbind init --convert`.

The low-hanging fruit solution here would be to simply document the need to update both within the developer guide.

Alternatively, and a better solution from a design perspective, would be to combine these two files into a single file that dictates the behavior for a generated `default.md`.

Of course, this is up for discussion as we should also consider (1) whether this is more hassle than necessary if this was well-documented, and (2) whether behaviors for these two files will deviate further in the future (pointed out by @tlylt) that could necessitate having these two as separate files.

_No response_

### Additional context

_No response_

WillCWX:

> _edited to follow PR template more accurately_

- [ ] Documentation update
- [ ] Bug fix
- [X] Code maintenance
- [ ] DevOps
- [X] Improve developer experience
- [ ] Others, please explain:

Attempts to:
Resolve #2293 

**Overview of changes:**

Feature / Behaviour Changes:
- `markbind init --convert` 
  - On an empty folder now produces a website that is exactly the same as `markbind  init` 
    - This also means no more `about.md` as it does not appear in `template/default`
    - The icon is also changed into `Your Logo` as it is in `default.md`
  - No longer adds `404` as a navigation link, `404.html` is also now displayed exactly the same as in `markbind init`
  - Is now able to be used with `-t minimal` and has `default.md` be exactly the same as a minimal template
  - generated `site.json` is now more readable with indentations
  - `footer` and `siteNav` nunjucks variables no longer replace their defaults and instead adds to their defaults.
- `markbind init`
  - `default.md` is now generated by `Site.generateDefaultLayout` rather than copied over

Code Maintenance Changes:

- Updating `default.md` layout:
  - As `default.md` should no longer exist, update the respective template `.njk` nunjucks files instead.
  -  New `TemplateConfig.ts` contains template configurations for both default and minimal.
       -  `njkFile: string` dictates which nunjucks file to use.
       -  `njkSubs: NjkVars[]` allows for using nunjucks to copy existing files to incorporate into the layout `default.md`
       -  `layoutObjs: SiteConfigPage[]` allows for edits to `config`
       -  `hasAutoSiteNav: boolean` toggles whether `Site.buildSiteNav` is called during a conversion
       -  `siteNavIgnore: string[]` contains files that `Site.buildSiteNav` will ignore when making links 

I have no idea whether this was worth it.
The scope of `markbind init --convert` is broad and its functions could potentially be used for easy migrations / upgrades.
However, this increase in scope may also unduly burden the `Site` class as it is the only class currently able to inherently do everything needed to convert. Also `Site/index.ts` got even longer...

`TemplateConfig.ts` may not be needed if we decide that templates should be left with just `default` and `minimal` as it is.

**Testing instructions:**
`npm run test` passes. The functional tests for `test_site_templates` have not changed but should still pass because the generated default.md for both minimal and default is exactly the same.

Fast-ish Manual Tests:
  1. Copy the `non-markbind-site` from the conversion tests
  2. `npm link` with this branch
  3. `markbind init --convert` the copied `non-markbind-site` folders
  4. `markbind serve` to see that they look `markbind init` but with texts different in the front page and additional / inherited links at the side. Additional texts to footer should exist if `_Footer.md` is present.

Update Site to generate default layouts

---

- [X] No unrelated changes


packages/core/src/Site/index.ts
```diff
@@ -407,60 +416,55 @@ export class Site {
...
+          fileCopy = `\n${fs.readFileSync(subFilePath, 'utf8')}`;
+        }
+      });
+      njkObject[njkSub.variableName] = fileCopy;
+    });
 
-    const wikiSiteNavPath = path.join(this.rootPath, WIKI_SITE_NAV_PATH);
-    let siteNav;
-    if (fs.existsSync(wikiSiteNavPath)) {
-      logger.info(`Copied over the existing ${WIKI_SITE_NAV_PATH} file to the converted layout\n`
```
lhw-1, line 438: I wonder if we can fit in this log somewhere...? I believe this log would still be helpful for the user since `siteNav` is one of our key navigation features.

lhw-1 (changes requested):
Good work and thanks for the PR @WillCWX! The changes look quite promising.

Some preliminary overall comments:
- As you mentioned, `TemplateConfig.ts` might not be necessary. However, it still adds a nice avenue for us to easily keep track of new templates if we end up needing more than just `default` and `minimal`.
- As mentioned in [this comment](https://github.com/MarkBind/markbind/issues/1866#issuecomment-1504757627) in #1866, we may want to remove default file generation for conversion - for that, we'll want to replace instead of adding to `siteNav` and `footer`. But for now, adding instead of replacing should be fine to work with the default files being generated.

Also, your addition of in-line documentations have been noted & are appreciated 😄

packages/core/src/Site/index.ts
```diff
@@ -380,13 +378,24 @@ export class Site {
...
+  }
+
+  /**
+   * Creates the default layout of the website based on a template.
+   */
+  async generateTemplateDefault(templateConfig: TemplateConfig) {
+    await this.readSiteConfig();
+    this.collectAddressablePages();
+    this.addDefaultLayoutFiles(templateConfig.njkFile, undefined);
+    await this.addTemplateLayoutSettingsToSiteConfig(templateConfig);
```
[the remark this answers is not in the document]
WillCWX: Sure, I have merged both functions together.

packages/core/src/Site/index.ts
```diff
@@ -380,13 +378,24 @@ export class Site {
...
+    Site.printBaseUrlMessage();
+  }
+
+  /**
+   * Creates the default layout of the website based on a template.
+   */
+  async generateTemplateDefault(templateConfig: TemplateConfig) {
+    await this.readSiteConfig();
+    this.collectAddressablePages();
+    this.addDefaultLayoutFiles(templateConfig.njkFile, undefined);
```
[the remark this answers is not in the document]
WillCWX: Ok, fixed!

packages/core/src/Site/index.ts
WillCWX:
```javascript
  createNjkObjects({ njkSubs, hasAutoSiteNav, siteNavIgnore }: TemplateConfig) {
    const njkObject: NunjuckObj = {};
    njkSubs.forEach((njkSub) => {
      let fileCopy;
      njkSub.fileSubstitutes.forEach((fileName) => {
        const subFilePath = path.join(this.rootPath, fileName);
        if (fs.existsSync(subFilePath)) {
          logger.info(`Copied over the existing ${fileName} file to the converted layout`);
          fileCopy = `\n${fs.readFileSync(subFilePath, 'utf8')}`;
        }
      });
      njkObject[njkSub.variableName] = fileCopy;
    });
```
[lines](https://github.com/WillCWX/markbind/blob/45d0ead81ab4bb4814d04c39fbc570d84632783c/packages/core/src/Site/index.ts#L439-L451)

I have placed this log inside a more general `nunjucks variable` copy function, will that be ok?

packages/core/src/Site/index.ts
```diff
@@ -378,15 +376,22 @@ export class Site {
   }
 
   /**
-   * Converts an existing GitHub wiki or docs folder to a MarkBind website.
+   * Creates the default layout of the website based on a template.
    */
-  async convert() {
+  async generateTemplateDefault(templateConfig: TemplateConfig, isConversion: boolean) {
```
lhw-1, line 381: Since this method now handles the default file generation for `markbind init` and conversion for `markbind init --convert`, a name like `initSite` might be more suitable (and self-explanatory), as compared to `generateTemplateDefault` - what do you think?

packages/core/src/Site/index.ts
lhw-1: Ah I should have been clearer, I was also pointing to:

```
+ 'Check https://markbind.org/userGuide/tweakingThePageStructure.html#site-navigation-menus\n'
+ 'for information on site navigation menus.');
```

But looking at the current setup of your functions, it may be more hassle than worth, so the current changes should be fine 👍 

We can re-visit this should we decide that a link to the UG for Site Navigation during conversion is helpful.

lhw-1 (commented):
Just a nit on naming:

packages/core/src/Site/index.ts
lhw-1: Ah I should have been clearer, I was also pointing to:

```
+ 'Check https://markbind.org/userGuide/tweakingThePageStructure.html#site-navigation-menus\n'
+ 'for information on site navigation menus.');
```

But looking at the current setup of your functions, it may be more hassle than worth, so the current changes should be fine 👍 

We can re-visit this should we decide that a link to the UG for Site Navigation during conversion is helpful.

packages/core/src/Site/index.ts
WillCWX: Sure, `generateTemplateDefault` is quite a mouthful and is difficult to remember. `initSite` is definitely much better.

lhw-1 (approved):
LGTM - Thanks for the PR @WillCWX!

packages/core/src/Site/TemplateConfig.ts
```diff
@@ -0,0 +1,61 @@
+import { SiteConfigPage } from './SiteConfig';
+
+type NjkVars = {
+  variableName: string,
+  fileSubstitutes: string[]
```
yucheng11122017, line 5: Why is an array used here? There should only be one file for each substitution
Changes below shows that the last file will be used anyway.

packages/core/src/Site/index.ts
```diff
@@ -401,66 +406,66 @@ export class Site {
...
-    }
+  createNjkObjects({ njkSubs, hasAutoSiteNav, siteNavIgnore }: TemplateConfig) {
+    const njkObject: NunjuckObj = {};
+    njkSubs.forEach((njkSub) => {
+      let fileCopy;
+      njkSub.fileSubstitutes.forEach((fileName) => {
+        const subFilePath = path.join(this.rootPath, fileName);
+        if (fs.existsSync(subFilePath)) {
+          logger.info(`Copied over the existing ${fileName} file to the converted layout`);
+          fileCopy = `\n${fs.readFileSync(subFilePath, 'utf8')}`;
```
yucheng11122017, line 447: What I'm referring to in another comment: 
Only the last file is used as a substitute here. In that case, would be better to make it a string instead of an array of string

yucheng11122017 (commented):
Hi @WillCWX thanks so much for the work! Sorry for the late review as well. 
Some comments regarding the fileSubstitutions :)

packages/core/src/Site/TemplateConfig.ts
WillCWX: I used an array so that possible variations of file names that could be used as substitutes.
Kind of like how `['README.md', 'Home.md']` is used in `addIndexPage()`
([link](https://github.com/WillCWX/markbind/blob/053d3b85a37744232396eae3b73012eb3ff2e6c0/packages/core/src/Site/index.ts#L396))

Though the current nunjucks variables only have one file name as a substitute, I thought that using an array may allow a greater range of possible file names. (at the cost of some complexity and readability)

yucheng11122017: Ah I see where you are coming from. That;s ok then :) 
Also could you change the variable name to something more obvious that its just for files? Eg. `NjkFileVars`

lhw-1:
Also, very minor comment: do make sure to follow the PR template (e.g. the checklist at the beginning for the purpose of PR)

packages/core/src/Site/minimalLayout.njk
```diff
@@ -0,0 +1,14 @@
+<header>
```
ang-zeyu, line 1: Do we still need `packages\core\template\minimal\_markbind\layouts\default.md`? (and the default one)

packages/core/src/Site/TemplateConfig.ts
```diff
@@ -0,0 +1,61 @@
...
+  ],
+  njkFile: 'defaultLayout.njk',
+  layoutObjs: [
+    { glob: '**/*.md', layout: 'default.md' },
+    { glob: '404.md', layout: '404.md' },
+  ],
+  hasAutoSiteNav: true,
+  siteNavIgnore: [
+    'index.md', '404.md',
+    'contents/topic1.md', 'contents/topic2.md', 'contents/topic3a.md', 'contents/topic3b.md',
```
ang-zeyu, line 35: (not your pr but possible future work)

For a _conversion_ feature, what do you think of including `'contents/topic1.md', 'contents/topic2.md', 'contents/topic3a.md', 'contents/topic3b.md'`? (personally only just realised its a little odd we are adding pages when doing conversion)

`index.md` and `404.md` makes some sense but I don't think we have `file.exists` checks or fallback plans if the site already has these files too.

ang-zeyu (commented):
> I have no idea whether this was worth it.
> The scope of markbind init --convert is broad and its functions could potentially be used for easy migrations / upgrades.
However, this increase in scope may also unduly burden the Site class as it is the only class currently able to inherently do everything needed to convert. Also Site/index.ts got even longer...

I think this is still worth it alone for eliminating `siteConvertLayout.njk`.


I like the idea of `TemplateConfig.ts` as well, it sets the ground for adding future templates easily.

> this increase in scope may also unduly burden the Site class

As a next step (if you are interested), would it be possible to extract all site initialisation and conversion functionality out of `Site`?
I can't imagine initialisation needs much stateful logic (to be contained within a `Site`), I see a few references but don't think they should be un-workable around.

It should help out with https://github.com/MarkBind/markbind/issues/1756 greatly

---

Just a few nits on the PR itself:

packages/core/src/Site/minimalLayout.njk
WillCWX: No we don't need it anymore, I'll delete it now that we are okay with how each template `.njk` looks in comparison

WillCWX: Sorry, I need to keep `packages\core\template\minimal\_markbind\layouts\default.md` to make sure `git` tracks the empty `layouts` folder. I couldn't get git to track the empty folder otherwise.

I made `default.md` empty with a message to edit the `.njk` file instead.

ang-zeyu:
@WillCWX (cc @damithc) for any purposes I think we can consider this PR "merged".

As site/index.ts was recently reverted https://github.com/MarkBind/markbind/issues/2333 to a `js` file, I suggest leaving the actual merging until that is fixed as there's quite a number of changes to this file here. (After which, rollback the merge commit here including that revert and replay only the code review's changes on top)

ang-zeyu:
Sorry as I do not have time currently to rereview this PR due to the large diffs from the merge commit including the revert. But if anyone else can pick things up here please feel free.

WillCWX:
@ang-zeyu , thanks for the heads up and for the review previously

> As a next step (if you are interested), would it be possible to extract all site initialisation and conversion functionality out of Site?
> I can't imagine initialisation needs much stateful logic (to be contained within a Site), I see a few references but don't think they should be un-workable around.

Would it be better if I extract all Site initialisation and conversion functionality out of Site first (in a different pull request) before coming back here?

This would ensure that 
- process of fixing https://github.com/MarkBind/markbind/issues/2333 would not be affected / easier to parse with less functions
- https://github.com/MarkBind/markbind/issues/1756 would progress
- no large diffs from the merge commit
- this pull request would be able to be merged independently of #2333 as it would edit `template.ts` instead of `Site.js / Site.ts`

tlylt:
Hi @WillCWX are you keen to continue working on this? If we can fix the conflicts and also outline the remaining points for discussion, I'm happy to support reviewing this PR and get a closure on this.

WillCWX:
@tlylt 

Apologies for the late reply, I would still like to work on this issue but I think the conflicts found in this pull request (and on my branch) are a bit too much. So I would like to close this pull request as it is.

I thinking starting fresh on a new branch/pull request will be easier for me to write new tests for the proposed modified `init` and `convert` functions (with codecov added, it gets angry when I move previously untested functions to new files without writing new tests). 
It will also be easier to avoid overwriting other people's contributions / creating new bugs as I try to fix the conflicts. 

Once again, apologies for the late reply and thanks to everyone for reviewing and helping with this pull request.

tlylt:
> @tlylt 
> 
> 
> 
> Apologies for the late reply, I would still like to work on this issue but I think the conflicts found in this pull request (and on my branch) are a bit too much. So I would like to close this pull request as it is.
> 
> 
> 
> I thinking starting fresh on a new branch/pull request will be easier for me to write new tests for the proposed modified `init` and `convert` functions (with codecov added, it gets angry when I move previously untested functions to new files without writing new tests). 
> 
> It will also be easier to avoid overwriting other people's contributions / creating new bugs as I try to fix the conflicts. 
> 
> 
> 
> Once again, apologies for the late reply and thanks to everyone for reviewing and helping with this pull request.

Sure, please aim to make incremental changes (across multiple PRs) if possible.

PR #2345, opened 2023-07-23
Fix `puml` not syncing Resolve #1617

### LINKED ISSUE #1617: PlantUML diagrams require a manual refresh when live-previewing in lazy loading mode
When using `markbind server -o`, if a page has a PlantUML diagram, the diagram appears as a broken image link when the page is auto-refreshed (due to an edit to the source code). Only after manually refreshing the browser I can see the diagram in the previewed page.
After auto-refresh by the live preview:
image

After another manual refresh:
image

Not sure if it is general issue or related to my local set up though.

SPWwj:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Resolve #1617
By introducing a lock service to wait for asynchronous functions to complete:
```
// Importing LockManager module from the specified directory
const LockManager = require('../../utils/LockManager');

// Start an async function
const lockId = LockManager.createLock();

// Stop of an async function
LockManager.deleteLock(lockId);

// Wait for all async functions to complete
await LockManager.waitForLockRelease();
```

Fix the issue where all distinct PUML diagrams were being reprocessed on every hot reload by introducing a hashmap tracker to keep track of the diagrams.
- The hash key comes from the content and the file path of the PUML diagram.

Performance Analysis:

The implementation of the lock service has minimal impact on PUML processing performance. 

Testing was conducted on 70 PUML diagrams:

- Original Configuration (without lock service)
    `- o`
    image
    - CPU Utilization: 12% to 93%
    - RAM Usage: 60% to 68%
    - Hot Reload CPU Utilization: 12% to 94%
    - Hot Reload RAM Usage: 60% to 69%
  `- d`
    image

    - Initial CPU Utilization: 12% to 97%
    - Initial RAM Usage: 63% to 78%
    - Hot Reload CPU Utilization: 6% to 94%
    - Hot Reload RAM Usage: 63% to 75%

- Singleton Implementation
   `-o`
    image
  Initial Performance Statistics:
    - CPU Utilization: 12% to 95%
    - RAM Usage: 63% to 72%
    - On every single diagram hot reload: RAM shows no change, CPU increases by 6%

   `-d`
   image

  Initial Performance Statistics:
  - CPU Utilization: 7% to 96%
  - RAM Usage: 63% to 74%
  - On every single diagram hot reload: RAM shows no change, CPU increases by 6%

These results indicate that the lock service does not significantly affect PUML processing performance.
In fact, it greatly reduces the CPU computation on every hot reload, leading to a more efficient processing system.

**Testing instructions:**
1. Check out the PR locally.
2. Create a new MarkBind project using `markbind init`.
3. Run `markbind serve -o`.
4. Experiment with the sample:

```markdown
<puml width="300">
@startuml
alice -> bob ++ : hellos
bob -> bob ++ : self call
bob -> bib ++  #005500 : a
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
</puml>
```
5. Edit the sample and wait for the hot reload to see the effect.

Introduce LockManager for Async Handling and Optimize PUML Diagrams

Introduce the LockManager module to manage the lifecycle of asynchronous
functions, ensuring robust and consistent handling within synchronous
methods.

Utilize the createLock, deleteLock, and waitForLockRelease methods for
starting, stopping, and waiting for async functions respectively.

Code snippet:

```javascript
const LockManager = require('path_to/LockManager');
const lockId = LockManager.createLock();
LockManager.deleteLock(lockId);
await LockManager.waitForLockRelease();
```

Fix the issue where all distinct PUML diagrams were being reprocessed on
every hot reload by introducing a hashmap tracker.

- The hash key comes from the content and the file name of the PUML diagram.
- The original \`processedDiagrams.clear()\` has been updated to
  \`cleanDiagrams(processedDiagrams)\`.

By implementing these changes, the code is made more maintainable,
asynchronous functions are properly managed, and PUML diagram processing
is optimized.

---

- [X] No unrelated changes


ang-zeyu:
The performance hit from `execSync` is quite significant (2103 site is a good demo ground).

The fix needs to be more involved https://github.com/MarkBind/markbind/issues/1617#issuecomment-951839872.

SPWwj:
> The performance hit from `execSync` is quite significant (2103 site is a good demo ground).
> 
> The fix needs to be more involved [#1617 (comment)](https://github.com/MarkBind/markbind/issues/1617#issuecomment-951839872).

I have an alternative solution using the lock file. Not sure if this could fix the issue.

tlylt:
@SPWwj is there a real need to implement it as a lock file? Can we do it in memory?

SPWwj:
> @SPWwj is there a real need to implement it as a lock file? Can we do it in memory?
> 
> Also pls update your PR description

Yes, a similar idea would be to use a singleton class instead of files for this lock service. 

Another way is to use Vue components. 

Alternatively, we could make `NodeProcessor.ts` asynchronous, but this would require a lot of work.

tlylt:
> > @SPWwj is there a real need to implement it as a lock file? Can we do it in memory?
> > Also pls update your PR description
> 
> Yes, a similar idea would be to use a singleton class instead of files for this lock service.
> 
> Another way is to use Vue components.
> 
> Alternatively, we could make `NodeProcessor.ts` asynchronous, but this would require a lot of work.

pros & cons? rationale behind your decision? could you elaborate to aid review plz

SPWwj:
> pros & cons? rationale behind your decision? could you elaborate to aid review plz

1. Use a singleton class instead of files for this lock service.
   - Pro: More efficient and faster than IO operations. Easy to implement without affecting the current implementation.
   - Cons: Extra utility class needed for asynchronous operations.

2. Use Vue components.
   - Pro: Easy to implement.
   - Cons: Results in a larger client-side library. Need to re-implement the component.

3. Make `NodeProcessor.ts` asynchronous. 
   - Pro: Supports the asynchronous operation for the NodeProcessor pipeline.
   - Cons: Need to re-implement the pipeline. This is similar to a TypeScript migration and can be very tedious.

The first option (singleton) is recommended as it does not affect the current project structure and provides a way to support asynchronous operations.

packages/core/test/unit/plugins/default/_plantuml/a31b4068deea63d65d1259b4d54bcc79.png
tlylt, line 1: If the only thing that this PR changes is to fix puml not syncing Resolve, why are these images added?

.gitignore
```diff
@@ -89,6 +89,7 @@ packages/core/src/Site/*.js
 packages/core/src/utils/*.js
 packages/core/src/variables/*.js
 packages/core/test/unit/**/*.js
+packages/core/_site/*
```
tlylt, line 92: This should not be necessary, why are we creating a markbind site in core?

tlylt:
> > pros & cons? rationale behind your decision? could you elaborate to aid review plz
> 
> 1. Use a singleton class instead of files for this lock service.
>    
>    * Pro: More efficient and faster than IO operations. Easy to implement without affecting the current implementation.
>    * Cons: Extra utility class needed for asynchronous operations.

> The first option (singleton) is recommended as it does not affect the current project structure and provides a way to support asynchronous operations.

Sounds like a plan. What about performance (time, and maybe cpu/memory?) before and after introducing this change? How much longer(or shorter or no diff) would it take, especially for sites with a lot of  puml diagrams?

packages/core/src/utils/lockFileManager.ts
tlylt, line 1: interesting 👀

tlylt, line 1: Btw perhaps a dump question: if you are creating this lock file and removing it when it's done, and waiting till the lock files have been removed, why don't we just wait till (by checking that) the image files are present in their folders?

packages/core/test/unit/plugins/default/_plantuml/a31b4068deea63d65d1259b4d54bcc79.png
SPWwj: > If the only thing that this PR changes is to fix puml not syncing, why are these images added?

These files were generated by the test case. After applying the lock service, the test case is able to finish generating the diagram before the program exits.

.gitignore
SPWwj: This happens because some test cases fail to generate the diagram before the program hits a fatal error, leaving the lockfile unremoved. As a result, there are some dummy files inside the `packages/core/_site/*`. If we apply the Singleton implementation, this would be redundant.

SPWwj:
> Sounds like a plan. What about performance (time, and maybe cpu/memory?) before and after introducing this change? How much longer(or shorter or no diff) would it take, especially for sites with a lot of puml diagrams?

I don't expect much performance difference if the lock service does not change the current implementation. The only thing it does is to wait for all asynchronous operations to complete, much like a `join`.

The compiling time may slightly increase after applying this lockfile, due to PlantUML is able to finish generating the file, just like in the test case mentioned above.

packages/core/src/utils/lockFileManager.ts
SPWwj: > Btw perhaps a dumb question: if you are creating this lock file and removing it when it's done, and waiting till the lock files have been removed, why don't we just wait till (by checking that) the image files are present in their folders?

Implementing your suggestion is feasible, but the current implementation serves as a general-purpose lock service to handle all kinds of asynchronous operations. Therefore, we can directly use this service to get an asynchronous output rather than waiting for a file; for instance, waiting for an SVG string from the `mermaid` API.

This service can act as an interim solution for asynchronous operations before we decide to make `NodeProcessor.ts` asynchronous.

tlylt:
> > Sounds like a plan. What about performance (time, and maybe cpu/memory?) before and after introducing this change? How much longer(or shorter or no diff) would it take, especially for sites with a lot of puml diagrams?
> 
> I don't expect much performance difference if the lock service does not change the current implementation. The only thing it does is to wait for all asynchronous operations to complete, much like a `join`.
> 
> The compiling time may slightly increase after applying this lockfile, due to PlantUML is able to finish generating the file, just like in the test case mentioned above.

Sure, have some numbers when you are done so that we can make a fair judgment 👍

packages/core/src/utils/lockFileManager.ts
tlylt: Interesting. If we await/join asynchronous operations properly, would this lock service still be necessary?

I'm not sure what you mean by making NodeProcessor.ts asynchronous. You might need to raise an issue and discuss this in further detail. For example:
> Pro: Supports the asynchronous operation for the NodeProcessor pipeline.

Aren't asynchronous operations already possible (just not working well when not awaited properly)? What's your definition of support?

damithc:
> Sure, have some numbers when you are done so that we can make a fair judgment

Good idea. Use a decent number of puml diagram in the test case.

SPWwj:
> > Sure, have some numbers when you are done so that we can make a fair judgment
> 
> Good idea. Use a decent number of puml diagram in the test case.

@damithc @tlyltI have updated the PR description to include some test statistics.

packages/core/src/utils/lockFileManager.ts
SPWwj: I don't think there's a way to properly await/join asynchronous operations under the current setting. Node.js doesn't allow us to join the asynchronous operations.
We can only use `Promise` and `async` for asynchronous operations. The `Promise` won't be waited on if the program exits first. Meanwhile, `async` can only work if the function is async, thus the return type will be a promise. This requires changing the function signature and updating the return type.

tlylt:
@SPWwj is this ready for review? There's still a merge conflict

SPWwj:
> @SPWwj is this ready for review? There's still a merge conflict

Fixed merged conflict

.gitignore
tlylt: So is it still redundant or do we need to keep this?

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -16,28 +16,35 @@ import { PluginContext } from '../Plugin';
...
 import { MbNode } from '../../utils/node';
 
+const LockManager = require('../../utils/LockManager');
+
 const JAR_PATH = path.resolve(__dirname, 'plantuml.jar');
 
 const processedDiagrams = new Set();
 
 let graphvizCheckCompleted = false;
+// Generate a lock file
```
tlylt, line 26: ?

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -16,28 +16,35 @@ import { PluginContext } from '../Plugin';
...
 
 /**
  * Generates diagram and returns the file name of the diagram
  * @param imageOutputPath output path of the diagram to be generated
  * @param content puml dsl used to generate the puml diagram
  */
 function generateDiagram(imageOutputPath: string, content: string) {
+  const lockId = LockManager.createLock();
```
tlylt, line 34: would it work if we move this to before line 48? to avoid the need for the change at line 38?

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -66,6 +74,8 @@ function generateDiagram(imageOutputPath: string, content: string) {
   childProcess.on('exit', () => {
     // This goes to the log file, but not shown on the console
     logger.debug(errorLog);
+    // Delete the lock file after generating the diagram
```
tlylt, line 77: ?

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,47 @@
+import { v4 as uuidv4 } from 'uuid';
+
+class LockManager {
+  private static _instance: LockManager;
+  private locks: Map<string, boolean>;
+
+  private constructor() {
+    this.locks = new Map();
```
tlylt, line 8: Curious what are your thoughts on choosing Map vs Set?

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.locks.clear();
+  }
+
+  waitForLockRelease(): Promise<void> {
+    return new Promise((resolve) => {
+      const checkLocks = () => {
+        if (this.locks.size === 0) {
+          resolve();
+        } else {
+          setTimeout(checkLocks, 100);
```
tlylt, line 39: How is `100` chosen? By experiment or? Any specific reason?

tlylt:
> The implementation of the lock service has minimal impact on PUML processing performance.

Does it have any impact on Page processing itself? Especially for a site with a lot of pages, i.e. https://github.com/nus-cs2113-AY2324S1/website

packages/core/src/utils/LockManager.ts
SPWwj: map is more flexible and intuitive in our case, as we might need to add the status of the lock.

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
SPWwj: Yes it will work.

packages/core/src/utils/LockManager.ts
SPWwj: Lowering the `setTimeout` increases the frequency of lock checks, which may improve responsiveness but increases computation. I have decreased the value from 100 to 10, and the CPU usage for 70 diagrams increased by around 3%, with a 0.2s decrease in compilation time.

SPWwj:
> > The implementation of the lock service has minimal impact on PUML processing performance.
> 
> Does it have any impact on Page processing itself? Especially for a site with a lot of pages, i.e. https://github.com/nus-cs2113-AY2324S1/website

This should not have a significant impact on page processing, and it should have no effect on pages without PlantUML diagrams. This is because all the added operations are almost as good as O(k), where k is the number of diagrams. The `setTimeout(checkLocks, 10);` might take a bit longer, but as it's asynchronous and the delay is within an acceptable range, it should be manageable.

packages/core/src/utils/LockManager.ts
tlylt: > as we might need to add the status of the lock

Are there any use cases that you foresee?

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
tlylt: Is the comment still up-to-date?

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  public static get instance() {
+    if (!LockManager._instance) {
+      LockManager._instance = new LockManager();
+    }
+
+    return LockManager._instance;
+  }
+
+  createLock(): string {
```
tlylt, line 19: Would it be more flexible to allow passing in a lockId, and only generate one if not provided? What do you think?

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,47 @@
+import { v4 as uuidv4 } from 'uuid';
+
+class LockManager {
```
tlylt, line 3: Adding comment/overview/description of this class would be nice, to help retain the context

tlylt:
Also, a commit message per https://se-education.org/guides/conventions/git.html plz

packages/core/src/utils/LockManager.ts
tlylt: ok we can keep 100

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
SPWwj: Oops, forget to remove.

packages/core/src/utils/LockManager.ts
SPWwj: I've made updates to the function to accept an optional `lockId`. This allows for more flexible operations.

packages/core/src/utils/LockManager.ts
SPWwj: **Status of the Lock**:
The current setup serves as a sample and doesn't include this feature. However, it can be integrated if needed. An example use-case would be as follows:

```
  childProcess.stdin?.write(
    //Update lock status
    content,
    (e) => {
      if (e) {
        logger.debug(e as unknown as string);
        logger.error(`Error generating ${imageOutputPath}`);
      }
      childProcess.stdin?.end();
    },
  );

  childProcess.on('error', (error) => {
    logger.debug(error as unknown as string);
    logger.error(`Error generating ${imageOutputPath}`);
    LockManager.deleteLock(lockId);
  });

  childProcess.stderr?.on('data', (errorMsg) => {
     //Update lock status
    errorLog += errorMsg;
  });

  childProcess.on('exit', () => {
    // This goes to the log file, but not shown on the console
    logger.debug(errorLog);
    LockManager.deleteLock(lockId);
  });
```

packages/core/src/utils/LockManager.ts
SPWwj: Added

tlylt:
Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
image

for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.

I'm using Windows 10, Chrome, on your PR branch.

SPWwj:
> Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
> 
> image
> 
> 
> 
> for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.
> 
> 
> 
> I'm using Windows 10, Chrome, on your PR branch.

Can you provide the command you use.
I notice that serve ｛-o/d｝ behave differently.

tlylt:
> > Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
> > image
> > for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.
> > I'm using Windows 10, Chrome, on your PR branch.
> 
> Can you provide the command you use. I notice that 'serve ｛-o/d｝' behave differently.

Tested on all three.

`markbind serve -o` : both initial and edit didn't work
`markbind serve` and `markbind serve -d`: initial render works, edit didn't work

SPWwj:
> > > Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
> > > image
> > > for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.
> > > I'm using Windows 10, Chrome, on your PR branch.
> > 
> > 
> > Can you provide the command you use. I notice that 'serve ｛-o/d｝' behave differently.
> 
> Tested on all three.
> 
> `markbind serve -o` : both initial and edit didn't work `markbind serve` and `markbind serve -d`: initial render works, edit didn't work

I mistakenly worked on the layout instead of the pages. I have now corrected and tested it for all three different commands on both my PC and laptop. It should now work as expected.

tlylt:
> > > > Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
> > > > image
> > > > for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.
> > > > I'm using Windows 10, Chrome, on your PR branch.
> > > 
> > > 
> > > Can you provide the command you use. I notice that 'serve ｛-o/d｝' behave differently.
> > 
> > 
> > Tested on all three.
> > `markbind serve -o` : both initial and edit didn't work `markbind serve` and `markbind serve -d`: initial render works, edit didn't work
> 
> I mistakenly worked on the layout instead of the pages. I have now corrected and tested it for all three different commands on both my PC and laptop. It should now work as expected.

Thanks will check on my end. BTW, does that change the performance analysis you conducted before?

SPWwj:
> > > > > Hi @SPWwj, following the testing instructions provided, I'm still getting broken images:
> > > > > image
> > > > > for both initial render, and after edit, the image remains broken after automatic refresh. A manual refresh is needed to load the images.
> > > > > I'm using Windows 10, Chrome, on your PR branch.
> > > > 
> > > > 
> > > > Can you provide the command you use. I notice that 'serve ｛-o/d｝' behave differently.
> > > 
> > > 
> > > Tested on all three.
> > > `markbind serve -o` : both initial and edit didn't work `markbind serve` and `markbind serve -d`: initial render works, edit didn't work
> > 
> > 
> > I mistakenly worked on the layout instead of the pages. I have now corrected and tested it for all three different commands on both my PC and laptop. It should now work as expected.
> 
> Thanks will check on my end. BTW, does that change the performance analysis you conducted before?
> 
> Quick update: the hot-reload works, but it takes 7-10 seconds to auto refresh for a single edit.

I've updated the test result. The time difference is minimal as shown above. Have you compared the time between the original and the lock?

I believe the entire page's reload and reprocessing is causing this. We could resolve it using a buffer-like technique.

tlylt:
> The `processedDiagrams.clear();` is called every time, causing the regeneration of the distinct diagram whenever there is a reload. Thus, almost the entire page will be reloaded. I'm not sure why it has been added.

If you don't clear the processedDiagrams, wouldn't it result in stale images since any processed diagram will not be generated a second time?

> I've updated the test result. The time difference is minimal as shown above. Have you compared the time between the original and the lock?

Is this because of the above reason? That the images are not being generated a second time.

SPWwj:
> > The `processedDiagrams.clear();` is called every time, causing the regeneration of the distinct diagram whenever there is a reload. Thus, almost the entire page will be reloaded. I'm not sure why it has been added.
> 
> If you don't clear the processedDiagrams, wouldn't it result in stale images since any processed diagram will not be regenerated?
> 
> * After editing the .puml source, it causes a reload but I don't see any changes to the images rendered
> 
> > I've updated the test result. The time difference is minimal as shown above. Have you compared the time between the original and the lock?
> 
> Is this because of the above reason? That the images were not being generated a second time.

I see the reason why there is a `clear` before the `generate`:

```
      if (node.attribs.name) {
        const nameWithoutBaseUrl = urlUtil.stripBaseUrl(node.attribs.name, config.baseUrl);
        const nameWithoutLeadingSlash = nameWithoutBaseUrl.startsWith('/')
          ? nameWithoutBaseUrl.substring(1)
          : nameWithoutBaseUrl;
        pathFromRootToImage = fsUtil.ensurePosix(fsUtil.setExtension(nameWithoutLeadingSlash, '.png'));

        delete node.attribs.name;
      } else {
        const normalizedContent = pumlContent.replace(/\r\n/g, '\n');
        const hashedContent = cryptoJS.MD5(normalizedContent).toString();
        pathFromRootToImage = `${hashedContent}.png`;
      }
```
If the image uses the hashed name `pathFromRootToImage = `${hashedContent}.png`;`, it will be updated correctly (unless a hash collision occurs, which is not in the scope of this discussion), due to the distinct file name.


However, if the name of the image is specified  `node.attribs.name`, and `clear` is used, the image will not be able to regenerate.

This is because the program cannot differentiate the difference before and after the content changes. Using `clear` is not a correct way to enforce the regeneration of the content, as it will remove all trackers and cause a regeneration of images, even if the images did not change.
image

The correct way to handle this would be to use a hashmap with key-value pairs, where the key is the hashed content of the file, and the value is the name of the file. This approach would be more efficient and precise compared to using a `set()`.

tlylt:
> This is because the program cannot differentiate the difference before and after the content changes. Using clear is not a correct way to enforce the regeneration of the content, as it will remove all trackers and cause a regeneration of images, even if the images did not change.

Say if the .puml file imports a global style.puml file for styling, and I make a change to style.puml, do you think it should trigger the regeneration of the image that imports style.puml? In this case, the .puml file has no change in content.

SPWwj:
> > This is because the program cannot differentiate the difference before and after the content changes. Using clear is not a correct way to enforce the regeneration of the content, as it will remove all trackers and cause a regeneration of images, even if the images did not change.
> 
> Say if a .puml file imports a global style.puml file for styling, and I make a change to style.puml, do you think it should trigger the regeneration of the image that imports style.puml? In this case, the .puml file has no change in content.

If that's the case, the import tracker should handle the `clear` action for the content tracker.

tlylt:
> The correct way to handle this would be to use a hashmap with key-value pairs, where the key is the hashed content of the file, and the value is the name of the file. This approach would be more efficient and precise compared to using a set().

Could you help me understand why it would be better to use a hashmap? We already use the hashed content of the puml as the file name, so
- changing the filename will result in a regeneration
  - are you saying this is not necessary as we could have done a file rename operation?
- changing the content of the puml file will change the hashed filename, which will result in a regeneration
  - Will still need to handle the global import issue I mentioned (as we don't seem to currently track that), if we remove the clear() before site generates

tlylt:
> However, if the name of the image is specified node.attribs.name, and clear is used, the image will not be able to regenerate.

Is there a typo? If clear is used, the image will be able to regenerate as the name won't be in the set?

SPWwj:
> > However, if the name of the image is specified node.attribs.name, and clear is used, the image will not be able to regenerate.
> 
> Is there a typo? If clear is used, the image will be able to regenerate as the name won't be in the set?

Yes should be clear is not used

SPWwj:
> > The correct way to handle this would be to use a hashmap with key-value pairs, where the key is the hashed content of the file, and the value is the name of the file. This approach would be more efficient and precise compared to using a set().
> 
> Could you help me understand why it would be better to use a hashmap? We already use the hashed content of the puml as the file name, so
> 
> * changing the filename will result in a regeneration
>   
>   * are you saying this is not necessary as we could have done a file rename operation?
> * changing the content of the puml file will change the hashed filename, which will result in a regeneration
>   
>   * Will still need to handle the global import issue I mentioned (as we don't seem to currently track that), if we remove the clear() before site generates

We can use a hash that combines both the content and the filename to generate a unique identifier. This way, any change in either the filename or the content will trigger a regeneration of the diagram.

For global imports, we can either add a trigger to clear the tracker or rebuild the file, since all the diagrams would be affected.

tlylt:
> We can use a hash that combines both the content and the filename to generate a unique identifier. This way, any change in either the filename or the content will trigger a regeneration of the diagram.

Sure, sounds gd.
 
> For global imports, we can either add a trigger to clear the tracker or rebuild the file, since all the diagrams would be affected.

It may not be true that all diagrams would be affected, i.e. diagrams that don't import that file. Anyway, we can have some leeway for this special case for now, if it's a blocker that cannot be resolved cleanly.

Keep a lookout on CPU usage and any significant delay after implementing the solution 👍

SPWwj:
> It may not be true that all diagrams would be affected, i.e. diagrams that don't import that file. Anyway, we can have some leeway for this special case for now, if it's a blocker that cannot be resolved cleanly.

Just learn how to import the style for puml.

`global_style.puml`:
```
!define AWESOME
skinparam backgroundColor #EEEBDC
skinparam handwritten true
skinparam class {
  BackgroundColor PaleGreen
  ArrowColor SeaGreen
  BorderColor SpringGreen
}
skinparam interface {
  BackgroundColor Yellow
  ArrowColor DarkSeaGreen
  BorderColor DarkOliveGreen
}
```
`main_diagram.puml`:
```
!include global_style.puml

class MyClass
interface MyInterface

MyClass --> MyInterface : implements
```

We can create a data structure for tracking files with the following attributes:
- `Filename`: The name of the file.
- `Hash`: A hash of the file's content to detect changes.
- `importList`: A list of files that are imported or dependent on this file.
- `outdate`: A boolean flag indicating if the file is outdated.

The algorithm could operate as follows:
1. If an import is present, then look for it in the `importList`.
   - Check if `filename` and `hash` match.
   - If they match, do nothing; otherwise, set `outdate` to `true`.
2. If the `hash` is not present or `outdate` is `true`, regenerate the diagram.

SPWwj:
> > We can use a hash that combines both the content and the filename to generate a unique identifier. This way, any change in either the filename or the content will trigger a regeneration of the diagram.
> 
> Sure, sounds gd.

Now, Markbind is able to only regenerate the edited PUML. For RAM and CPU usage, I don't see much difference from the original implementation.

tlylt:
@SPWwj
- for inline puml seems to work fine, great job!
- for puml src from a separate .puml file, it seems to not regenerate properly, one visible case is when you delete something (which will trigger an update in the image) and then undo that deletion (which will reload the page, but the image remains the same and not reverted back)
crRXvmXLXL

E.g.
have the following in `test.puml`
```
@startuml
alice -> bob ++ : hellos
bob -> bob ++ : self call
bob -> bib ++  #005500 : a
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
```
then have a index.md that contains:
```
<puml src="test.puml" width="574" />
```

- First delete `bob -> george !! : delete` and save
- then undo the delete and save

SPWwj:
> @SPWwj
> 
> * for inline puml seems to work fine, great job!
> * for puml src from a separate .puml file, it seems to not regenerate properly, one visible case is when you delete something (which will trigger an update in the image) and then undo that deletion (which will reload the page, but the image remains the same and not reverted back)

Above issue is fixed.

For the global style import, it looks like it needs to be done with the live-server, as editing the style file does not update the current page.

image

tlylt:
> Above issue is fixed.

Thank you @SPWwj, seems to work fine now. Will review the code next. Meanwhile, could you update the PR description as it does not detail your new changes? I assume the performance analysis is also stale?

packages/core/test/unit/plugins/default/_plantuml/a31b4068deea63d65d1259b4d54bcc79.png
tlylt, line 1: @SPWwj Running the tests via `npm run test` locally and I don't see these 3 files being generated anymore. Any potential issues for this phenomenon? If it is not supposed to be generated now, let's delete these files.

packages/core/src/Site/index.ts
```diff
@@ -1359,6 +1359,7 @@ export class Site {
     try {
       await this.runPageGenerationTasks(pageGenerationTasks);
       await this.writeSiteData();
+
```
tlylt, line 1362:
```suggestion
```

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -90,14 +118,15 @@ export = {
...
-    processedDiagrams.clear();
+    cleanDiagrams(processedDiagrams);
     graphvizCheckCompleted = false;
   },
 
   processNode: (_pluginContext: PluginContext, node: MbNode, config: NodeProcessorConfig) => {
     if (node.name !== 'puml') {
       return;
     }
+
```
tlylt, line 129:
```suggestion
```

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,86 @@
...
+  private constructor() {
+    this.locks = new Map();
+  }
+
+  /**
+   * Provides a way to access the single instance of the LockManager.
+   * If it doesn't exist, it creates one.
+   * @returns {LockManager} The single instance of LockManager.
+   */
+  public static get instance() {
```
tlylt, line 33:
```suggestion
  public static get instance(): LockManager {
```

packages/core/src/utils/LockManager.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Waits until all locks are released and then resolves.
+   * @returns {Promise<void>} A promise that resolves when all locks are released.
+   */
+  waitForLockRelease(): Promise<void> {
+    return new Promise((resolve) => {
+      const checkLocks = () => {
+        if (this.locks.size === 0) {
+          resolve();
+        } else {
+          setTimeout(checkLocks, 1);
```
tlylt, line 77: Will `1` be too intense? How about `100`?

SPWwj: Tested both `1` and `100` settings for processing approximately 70 diagrams:

- **Setting 1:** (Website generation complete! Total build time: 10.997s)
  - Speed: Faster 
  - CPU Intensity: More CPU-intensive

- **Setting 100:**
  - Speed: Slower (info: Website generation complete! Total build time: 11.587s)
  - CPU Intensity: Less CPU-intensive

Change in CPU Only around 3%.

tlylt: Thank you. I think we can go with 100 first and reassess in the future 🙆‍♂️

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -16,9 +16,32 @@ import { PluginContext } from '../Plugin';
...
-const processedDiagrams = new Set();
+const processedDiagrams = new Map<string, DiagramStatus>();
+
+function clearDeadDiagrams(diagrams: Map<string, DiagramStatus>) {
+  Array.from(diagrams.entries())
+    .filter(([, value]) => !value.isAlive)
+    .forEach(([key]) => diagrams.delete(key));
+}
+
+function initDiagrams(diagrams: Map<string, DiagramStatus>) {
```
tlylt, line 35: Could you explain why we are initializing the diagrams and setting their `isAlive` to false?
- what's the definition of `isAlive` ? Is it equivalent to `isLatest`?

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
```diff
@@ -28,16 +51,23 @@ let graphvizCheckCompleted = false;
  * @param content puml dsl used to generate the puml diagram
  */
 function generateDiagram(imageOutputPath: string, content: string) {
+  const hashKey = cryptoJS.MD5(imageOutputPath + content).toString();
+
   // Avoid generating twice
-  if (processedDiagrams.has(imageOutputPath)) { return; }
-  processedDiagrams.add(imageOutputPath);
+  if (processedDiagrams.has(hashKey)) {
```
tlylt, line 57: should we check the status of `isAlive` here? Is there different treatment for either value?

tlylt (commented):
Hey @SPWwj, looks mostly good on my end, thank you for your quick response. Have one more concern as I read through the diagram status tracking code:
- Could you add in comments to explain the functions? the definition of "dead" and "alive" weren't as clear. I initially thought you meant "stale" and "fresh", but the logic doesn't add up.
- Could you illustrate the flow of events for the typical cases:
  - on first generation, what happens? (especially the state of processedDiagrams)
  - after edit of a puml file, what happens? (especially the state of processedDiagrams)
  - after edit of a non puml file, what happens? (especially the state of processedDiagrams)

I'm not sure whether processedDiagrams is working as advertised. Clarification from your end would help!

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
SPWwj: I have renamed it to isStale, used to track whether the diagram has been used by the site or latest diagram .

packages/core/src/plugins/default/markbind-plugin-plantuml.ts
SPWwj: Don't need to check; we just need to set it to isFresh to ensure that the diagram is still being used by the site.

SPWwj:
> Hey @SPWwj, looks mostly good on my end, thank you for your quick response. Have one more concern as I read through the diagram status tracking code:
> 
> * Could you add in comments to explain the functions? the definition of "dead" and "alive" weren't as clear. I initially thought you meant "stale" and "fresh", but the logic doesn't add up.
> * Could you illustrate the flow of events for the typical cases:
>   
>   * on first generation, what happens? (especially the state of processedDiagrams)
>   * after edit of a puml file, what happens? (especially the state of processedDiagrams)
>   * after edit of a non puml file, what happens? (especially the state of processedDiagrams)
> 
> I'm not sure whether processedDiagrams is working as advertised. Clarification from your end would help!


On first generation, all the diagrams will be added to the map, and whenever a diagram is generated,
it will be marked as fresh.
Upon editing a PUML file or a non-PUML file that triggers a hot reload on the site,
the stale diagram will be removed from the map, and the map will be reset to all stale.
If the diagram is inside the map, the generateDiagram function will not regenerate the diagram.
This is to prevent generating the same diagram twice.

tlylt:
> Upon editing a PUML file or a non-PUML file that triggers a `hot reload` on the site, the stale diagram will be removed from the map, and the map will be reset to all stale.

The stale diagram may not be removed immediately right?

Example A, B are diagrams

1. Initial generation
    - processedMap = {A=true, B=true}
2. Edit A (which creates C as a new diagram)
   - This will trigger a hot reload on the site
       - before regeneration: no stale diagrams, so no diagrams are deleted
       - before regeneration: now set A, B as stale, C is not in the map yet
         - processedMap = {A=false, B=false}
      - during regeneration: A in the map remains stale, B mark as fresh processedMap = {A=false, B=true}
      - during regeneration: C is added to the map and marked as fresh processedMap = {A=false, B=true, C=true}

And only after a further save operation, will A be removed from the map?

Also, those diagrams that are removed from the map, but they are not deleted right?
- there wasn't an issue previously because we override the same diagram? Now with this new approach we need to perform cleanup to avoid accumulating a ton of diagrams in _site folder?

---

Finally, have you tried observing the actual values of processedDiagrams during initialization and after edit?
I console log the values and they seem quite strange.
For example, having two images in one page (and no more puml in other pages) by adding the following to an index.md of a newly generated site:
```
<puml width="300">
@startuml
A -> bob ++ : hellos
bob -> bob ++ : self call
bob -> bib ++  #005500 : a
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
</puml>

<puml width="300">
@startuml
B -> bob ++ : hellos
bob -> bob ++ : self call
bob -> bib ++  #005500 : a
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
</puml>
```
I would assume that processedDiagrams would have both files after markbind serve, but apparently when I console log, it only contains one:
(to illustrate)
image

Then, even though I just delete something unrelated in an index.md, the processedDiagrams values are updated for some weird reason to one true, one false:
image

FYI I added the log here:
image

SPWwj:
> The stale diagram may not be removed immediately right?

Yes, it will be removed in a future update.

> Example A, B are diagrams

1. Initial generation processedDiagrams {}
    * generateDiagram =>  processedDiagrams Add A and B to processedDiagrams   {A=true, B=true}
2. Edit A (which creates C as a new diagram)
    
    * This will trigger a hot reload on the site
      
      * before regeneration: no stale diagrams, so no diagrams are deleted
          *  now set A, B as stale, C is not in the map yet
      * during regeneration: A in the map remains stale, B mark as fresh generateDiagram => {A=false, B=true, C = true}
          *  Add C = true
          *  Update B = True
          *  Unchange A = stale
3 Edit B (Which will create D as a new diagram)
    * This will trigger a hot reload on the site
      
      * before regeneration:
          *  A will be remove
          *  now set B, C as stale, D is not in the map yet
      * during regeneration: A in the map remains stale, B mark as fresh generateDiagram => { B=false, C = true, D = true}
          *  Add D = True
          *  Update C = True
          *  Unchange B = False
4 Future Edit repeat above steps. 

Yes answered above

> Also, those diagrams that are removed from the map, but they are not deleted right?

Yes 

> * there wasn't an issue previously because we override the same diagram? Now with this new approach we need to perform cleanup to avoid accumulating a ton of diagrams in _site folder?

We can add a feature here to remove diagrams.

SPWwj:
> 



> > Upon editing a PUML file or a non-PUML file that triggers a `hot reload` on the site, the stale diagram will be removed from the map, and the map will be reset to all stale.
> 
> The stale diagram may not be removed immediately right?
> 
> Example A, B are diagrams
> 
> 1. Initial generation
>    
>    * processedMap = {A=true, B=true}
> 2. Edit A (which creates C as a new diagram)
>    
>    * This will trigger a hot reload on the site
>      
>      * before regeneration: no stale diagrams, so no diagrams are deleted
>      * before regeneration: now set A, B as stale, C is not in the map yet
>        
>        * processedMap = {A=false, B=false}
>      * during regeneration: A in the map remains stale, B mark as fresh processedMap = {A=false, B=true}
>      * during regeneration: C is added to the map and marked as fresh processedMap = {A=false, B=true, C=true}
> 
> And only after a further save operation, will A be removed from the map?
> 
> Also, those diagrams that are removed from the map, but they are not deleted right?
> 
> * there wasn't an issue previously because we override the same diagram? Now with this new approach we need to perform cleanup to avoid accumulating a ton of diagrams in _site folder?
> 
> Finally, have you tried observing the actual values of processedDiagrams during initialization and after edit? I console log the values and they seem quite strange. For example, having two images in one page (and no more puml in other pages) by adding the following to an index.md of a newly generated site:
> 
> ```
> <puml width="300">
> @startuml
> A -> bob ++ : hellos
> bob -> bob ++ : self call
> bob -> bib ++  #005500 : a
> bob -> george ** : create
> return done
> return rc
> bob -> george !! : delete
> return success
> @enduml
> </puml>
> 
> <puml width="300">
> @startuml
> B -> bob ++ : hellos
> bob -> bob ++ : self call
> bob -> bib ++  #005500 : a
> bob -> george ** : create
> return done
> return rc
> bob -> george !! : delete
> return success
> @enduml
> </puml>
> ```
> 
> I would assume that processedDiagrams would have both files after markbind serve, but apparently when I console log, it only contains one: (to illustrate) image

This is because the diagram is only added to the map before the generation of a new diagram:
image


> Then, even though I just delete something unrelated in an index.md, the processedDiagrams values are updated for some weird reason to one true, one false: image
> 
> FYI I added the log here: image

This is correct because the console log is placed at the beginning of the function. However, the following code is in charge of preventing the regeneration of the existing diagram:

image

tlylt:
> This is correct because the console log is placed at the beginning of the function. However, the following code is in charge of preventing the regeneration of the existing diagram:

I see, thank you for that!

I adjusted the comments slightly as well as code (made unlink sync).

For inline use of `puml` diagrams via e.g. 
```
<puml width="300">
@startuml
B -> bob ++ : hellos
bob -> bob ++ : self call
bob -> bib ++  #005500 : a
bob -> george ** : create
return done
return rc
bob -> george !! : delete
return success
@enduml
</puml>
```
seems alright now, the only issue is the delayed deletion of the stale diagrams. Could be improved in the future but ok for now.

For use of `src` in `puml`, via e.g. `<puml src="diagrams/ArchitectureSequenceDiagram.puml" width="574" />`, I am encountering some problems. It seems like we might either have some collision or accidentally delete the wrong files. 
To reproduce:
- clone https://github.com/tlylt/ab3-docs-demo
- `cd docs`
- `markbind serve` using your branch of markbind
- go to http://127.0.0.1:8080/DeveloperGuide.html#design
- now, add `<puml src="diagrams/ArchitectureDiagram.puml" width="280" />` to the end of `index.md`
  - save the file (i.e. ctrl + s)
  - save the file again (i.e. ctrl + s)
- refresh the http://127.0.0.1:8080/DeveloperGuide.html#design
- see that besides the first image, the rest are not found


image


(On initial markbind serve, the _site images are present)
image

(After double save, only the first diagram remains, hence resulting in error)
image

@SPWwj could you investigate this, please?

SPWwj:
> > This is correct because the console log is placed at the beginning of the function. However, the following code is in charge of preventing the regeneration of the existing diagram:
> 
> I see, thank you for that!
> 
> I adjusted the comments slightly as well as code (made unlink sync).
> 
> For inline use of `puml` diagrams via e.g.
> 
> ```
> <puml width="300">
> @startuml
> B -> bob ++ : hellos
> bob -> bob ++ : self call
> bob -> bib ++  #005500 : a
> bob -> george ** : create
> return done
> return rc
> bob -> george !! : delete
> return success
> @enduml
> </puml>
> ```
> 
> seems alright now, the only issue is the delayed deletion of the stale diagrams. Could be improved in the future but ok for now.
> 
> For use of `src` in `puml`, via e.g. `<puml src="diagrams/ArchitectureSequenceDiagram.puml" width="574" />`, I am encountering some problems. It seems like we might either have some collision or accidentally delete the wrong files. To reproduce:
> 
> * clone https://github.com/tlylt/ab3-docs-demo
> * `cd docs`
> * `markbind serve` using your branch of markbind
> * go to http://127.0.0.1:8080/DeveloperGuide.html#design
> * now, add `<puml src="diagrams/ArchitectureDiagram.puml" width="280" />` to the end of `index.md`
>   
>   * save the file (i.e. ctrl + s)
>   * save the file again (i.e. ctrl + s)
> * refresh the http://127.0.0.1:8080/DeveloperGuide.html#design
> * see that besides the first image, the rest are not found
> 
> image
> 
> (On initial markbind serve, the _site images are present) image
> 
> (After double save, only the first diagram remains, hence resulting in error) image
> 
> @SPWwj could you investigate this, please?

The hot reload scans only the PUML diagram on the modified page, causing diagrams on other non-modified pages to become stale. Currently, there's no good solution, as there's no way to identify inline PUML. Every change to inline PUML generates a new diagram without a common identifier to the stale diagram.

Since this is for production, one option is to clean up the PUML diagrams in the directory when no diagram is in the tracker list, to remove the redundant diagrams.

tlylt:
Does making the plantuml status tracking hashmap persist through site rebuild help to fix the issue? Either somehow make it global or write to file

SPWwj:
> Does making the plantuml status tracking hashmap persist through site rebuild help to fix the issue? Either somehow make it global or write to file

The issue won't be resolved, as we have no way to identify and remove the inline PUML diagram tracker from the tracking list. This is because there's no common identifier for the inline PUML diagram, unlike the .puml file where we can use the filename as an identifier.

One solution is to remove the diagram during the initial load with 'Markbind serve'. This ensures that our tracker list only captures the required diagrams.

tlylt:
Hey @SPWwj could you help me understand your replies further?

> Every change to inline PUML generates a new diagram without a common identifier to the stale diagram.

Aren't we currently making the identifier depend on both the output file path and the puml content, meaning every change to any PUML diagram will generate a new diagram without a common identifier to the stale diagram? This includes both inline and file puml diagrams?

> Since this is for production, one option is to clean up the PUML diagrams in the directory on init when no diagram is in the tracker list, to remove the redundant diagrams.

"what" is for production? What do you mean by "production" here?


> One solution is to remove the diagram during the initial load with 'Markbind serve'. This ensures that our tracker list only captures the required diagrams.

Do you mean we now **ONLY** remove the diagram during the initial load with "markbind serve", and stop removing the diagrams afterwards? I think that is a viable solution.

SPWwj:
> Hey @SPWwj could you help me understand your replies further?
> 
> > Every change to inline PUML generates a new diagram without a common identifier to the stale diagram.
> 
> Aren't we currently making the identifier depend on both the output file path and the puml content, meaning every change to any PUML diagram will generate a new diagram without a common identifier to the stale diagram? This includes both inline and file puml diagrams?
> 

The goal is to remove unused, outdated diagrams.

Here's the relevant code snippet:
image

If `node.attribs.name` is present, a new diagram can overwrite the old one using the same `imageOutputPath`.
`imageOutputPath` also can be used to update the current tracking list.

```
    const imageOutputPath = path.resolve(config.outputPath, PUML_DIR, pathFromRootToImage);
    generateDiagram(imageOutputPath, pumlContent);
```
If node.attribs.name is missing, the old diagram can't be removed or replaced by a new one,
and tracking list will expand because we have no information on which diagram has been updated.

> > Since this is for production, one option is to clean up the PUML diagrams in the directory on init when no diagram is in the tracker list, to remove the redundant diagrams.
> 
> "what" is for production? What do you mean by "production" here?
> 

When the site is ready to be deployed and go live.

> > One solution is to remove the diagram during the initial load with 'Markbind serve'. This ensures that our tracker list only captures the required diagrams.
> 
> Do you mean we now **ONLY** remove the diagram during the initial load with "markbind serve", and stop removing the diagrams afterwards? I think that is a viable solution.


Yes, because during the implementation of the website, I don't think there's a need to worry about deleting unused diagrams while editing PUML source code. This will also improve the undo operation during PUML code editing.

tlylt:
> If node.attribs.name is missing, the old diagram can't be removed or replaced by a new one, and tracking list will expand because we have no information on which diagram has been updated.

Hmm, now it reminds me that we should ensure that our solution works for the case where the user edits the plantuml files without running `markbind serve`.

> Yes, because during the implementation of the website, I don't think there's a need to worry about deleting unused diagrams while editing PUML source code. This will also improve the undo operation during PUML code editing.

Sure, sounds like a plan 👍.
To track, there are a few use case to check in order to ensure the functionality works:
- [ ] unused diagrams don't end up accumulating and occupying disk space (should at least be cleared after stopping and rerunning markbind serve
- [ ] updates to inline puml should work
- [ ] updates to puml with src to a file should work
- [ ] updates to puml with user-defined name should work
- [ ] updates to puml done without running markbind serve should work

SPWwj:
> ```
> <puml src="test.puml" width="574" />
> ```

Actually, I just realized that MarkBind already takes care of deleting redundant files during build time. Thus, there's no need to worry about accumulating and occupying disk space.

tlylt (approved):
Thank you @SPWwj, LGTM 🚀 

I updated the relevant docs to:
- use the `<puml src="...puml">` directly in our UG
- remove the generated diagrams since they will be generated on the fly
- update instructions in DG regarding puml files in UG
- add a note on duplicated names may result in invalid output

And finally some minor refactoring/adjusting comments.

PR #2356, opened 2023-08-17
Implement Site.Prototype methods in ES6 syntax #2280

### LINKED ISSUE #2280: Refactor methods implemented using Site.prototype

Currently, we implement a few methods using `Site.prototype` ([here](https://github.com/jovyntls/markbind/blob/63ed8520b1208ead38162628dd055bbd9f83d8c3/packages/core/src/Site/index.ts#L1698-L1736)) due to incompatibility with ES6 class syntax. 
This is slightly confusing, especially since we need to [stub out the method signatures](https://github.com/jovyntls/markbind/blob/63ed8520b1208ead38162628dd055bbd9f83d8c3/packages/core/src/Site/index.ts#L1682-L1695) to maintain TS compatibility.

If it is possible, let's implement these as class methods for consistency and readability.

### Additional context

- See #979 for the reasoning behind using Site.prototype.xxx
- [Relevant context](https://github.com/MarkBind/markbind/pull/2270#discussion_r1164425215) on the reasoning for stubbing out the method signatures

jmestxr:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Resolves #2280 according to solution proposed in #979

**Overview of changes:**

- Refactor `Site.prototype` methods implemented in `Site/index.ts` using ES6 class syntax

**Testing instructions:**

`npm run test`

Refactor Site.prototype methods in ES6 syntax

---

- [ ] No unrelated changes


tlylt:
> solution proposed in https://github.com/MarkBind/markbind/pull/979

Which solution are you referring to?

jmestxr:
> > solution proposed in #979
> 
> Which solution are you referring to?

I was referring to this comment: https://github.com/MarkBind/markbind/pull/979#issuecomment-574157038
using this syntax: `rebuildAffectedSourceFiles = delay(this._rebuildAffectedSourceFiles, 1000);`

ESlint doesn't show any parsing errors for me so I assumed it was updated alr.

tlylt:
> > > solution proposed in #979
> > 
> > 
> > Which solution are you referring to?
> 
> I was referring to this comment: [#979 (comment)](https://github.com/MarkBind/markbind/pull/979#issuecomment-574157038) using this syntax: `rebuildAffectedSourceFiles = delay(this._rebuildAffectedSourceFiles, 1000);`
> 
> ESlint doesn't show any parsing errors for me so I assumed it was updated alr.

I see, so because class fields proposal is accepted in TC39 hence there is no more syntax error complained by Eslint?
Could you investigate the browser support for class fields since it's only accepted last year (2022)?

I briefly searched it and found that class fields seem to not work in IE at all and have about ~92% support globally
https://caniuse.com/?search=class%20field

It's not a big issue, but good to find out more before we make this change.

jmestxr:
> 

I did some search and the browser support for both constructor and class fields are similar, so logically speaking, MarkBind should be able to continue supporting the browsers that we are currently supporting.

Javascript class is not supported in IE at all: https://caniuse.com/?search=class. So anyway, our current MarkBind is not supported in IE at all

packages/core/src/Site/index.ts
```diff
@@ -690,6 +687,7 @@ export class Site {
       await this.writeSiteData();
       this.calculateBuildTimeForGenerate(startTime, lazyWebsiteGenerationString);
       if (this.backgroundBuildMode) {
+        // @ts-ignore
```
tlylt, line 690: could you explain why this is necessary?

jmestxr: From what I understand, `_backgroundBuildNotViewedFiles` doesn't take in any args and isn't compatible with the `func` param in `delay` technically (not sure why `func` in `delay` is given its type `(arg: T[]) => Promise<unknown>` in the first place, could have been more flexible?). So TS gave me an error as it thought `backgroundBuildNotViewedFiles` should be receiving an arg when it didn't have to.

It was a little confusing here as TS has accepted `_backgroundBuildNotViewedFiles` as its `func` param in the first place, not giving any error. Not really sure why..

tlylt: Do you think it will be better if we assert the function signature as without argument at line 1706:
```
  backgroundBuildNotViewedFiles = delay(this._backgroundBuildNotViewedFiles as () => Bluebird<unknown>, 1000);
```
to make it more explicit that we know this function takes in no extra argument? https://www.scaler.com/topics/typescript/typescript-as/

jmestxr: Thanks for the tip @tlylt 

On a related note, currently in `serveUtil.js`, in line 29 and 62, we have:

`return site.rebuildSourceFiles(filePath);`

even though rebuildSourceFiles does not take in any argument. Is this is mistake and should I rectify it?

packages/cli/src/util/serveUtil.js
```diff
@@ -26,7 +26,7 @@ const addHandler = (site, onePagePath) => (filePath) => {
   }
   Promise.resolve('').then(async () => {
     if (site.isFilepathAPage(filePath) || site.isDependencyOfPage(filePath)) {
-      return site.rebuildSourceFiles(filePath);
```
tlylt, line 29: hmm, maybe there's something wrong here. I'm not sure why we are rebuilding all source files when one file is added/removed. Anyhow, it's not related to this PR.

packages/core/src/Site/index.ts
```diff
@@ -1678,58 +1675,35 @@ export class Site {
...
-/**
- * Build/copy assets that are specified in filePaths
- * @param filePaths a single path or an array of paths corresponding to the assets to build
- */
-Site.prototype.buildAsset = delay(Site.prototype._buildMultipleAssets as () => Bluebird<unknown>, 1000);
+  /**
+   * Build/copy assets that are specified in filePaths
+   * @param filePaths a single path or an array of paths corresponding to the assets to build
+   */
+  buildAsset = delay(this._buildMultipleAssets as () => Bluebird<unknown>, 1000);
```
tlylt, line 1682: Might cause a type problem in the future when serveUtil.js is migrated to TypeScript, but should be ok for now.

tlylt (approved):
LGTM

PR #2366, opened 2023-09-09
Reorganize MarkBind installation instructions in UG

tlylt:

- [x] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Re-organize the installation instructions into three methods. Add a note in CLI command docs.

Request came from: https://github.com/se-edu/addressbook-level3/pull/156#issuecomment-1705903643

In the future, we might want to move away from using global installation, which will require adjustments to the setup instructions again.

**Testing instructions:**
See UG:
- https://deploy-preview-2366--markbind-master.netlify.app/userguide/gettingstarted
- https://deploy-preview-2366--markbind-master.netlify.app/userguide/clicommands

Reorganize MarkBind installation instructions in UG

---

- [x] No unrelated changes

### DIFF
--- a/docs/userGuide/cliCommands.md
+++ b/docs/userGuide/cliCommands.md
@@ -10,6 +10,13 @@
 
 ### Overview
 
+<box type="info" seamless>
+
+If you do not have MarkBind installed globally, you can still use MarkBind commands by prefixing the commands with `npx`. For example, `markbind init` becomes `npx markbind-cli init`.
+
+If you have MarkBind installed locally, you may also refer to the scripts section of your `package.json` file for the commands you can use. For example, `npm run init` for `markbind init`.
+</box>
+
 An overview of MarkBind's Command Line Interface (CLI) can be referenced with `markbind --help`:
 ```
 $ markbind --help
--- a/docs/userGuide/gettingStarted.md
+++ b/docs/userGuide/gettingStarted.md
@@ -33,7 +33,7 @@ This method is recommended for most users. It allows you to use MarkBind command
 Run the following command to install MarkBind globally. This will make the `markbind` command available in your terminal.
 
 ```
-$ npm install -g markbind-cli
+npm install -g markbind-cli
 ```
 
 Next, run the command `markbind`. If MarkBind has been installed correctly, you should see the MarkBind ascii logo followed by a summary of MarkBind commands as the output.
@@ -58,7 +58,7 @@ Usage: ...
 Navigate into an empty directory and run the following command to initialize a skeletal MarkBind site in that directory. It will create several new files in the directory e.g., `index.md`, `site.json`.
 
 ```
-$ markbind init
+markbind init
 ```
 
 <include src="tip.md" boilerplate >
@@ -90,7 +90,7 @@ Run the following command in the same directory. It will generate a website from
 </modal>
 
 ```
-$ markbind serve
+markbind serve
 ```
 
 Do some changes to the `index.md` and save the file. The live preview in the Browser should update automatically to reflect your changes.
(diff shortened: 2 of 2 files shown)


damithc:
@tlylt How about at the very top we mention something like the following?

> The steps below explains how to install ... If you prefer to .... instead, please scroll to .... or, if you prefer to ... scroll to ...

And then, we put the two alternatives as separate sections on their own (not as panels), at the bottom. That way, the two alternatives appear as peers of the original rather than being tucked away under step 1 of the original.

Also, in the cli page, do we need to mention how to adjust the commands if they are using the alternatives?

tlylt:
> @tlylt How about at the very top we mention something like the following?
> 
> > The steps below explains how to install ... If you prefer to .... instead, please scroll to .... or, if you prefer to ... scroll to ...
> 
> And then, we put the two alternatives as separate sections on their own (not as panels), at the bottom. That way, the two alternatives appear as peers of the original rather than being tucked away under step 1 of the original.

ok.

> Also, in the cli page, do we need to mention how to adjust the commands if they are using the alternatives?

Hmm, I think there's no need because ultimately the CLI is still the same. For global you can invoke `markbind` directly because it's available globally in your terminal. For local and npx, underlying calls are still `markbind serve` etc. I can make a note if you think it helps clarify? For users who understand npx and npm packages, I think the confusion is quite minimal.

damithc:
> For users who understand npx and npm packages, I think the confusion is quite minimal.

@tlylt As our target users are not regular node/JavsScript users, it is best to assume no knowledge of node/npm.

tlylt:
> The steps below explains how to install ... If you prefer to .... instead, please scroll to .... or, if you prefer to ... scroll to ...

I added a note at the top. For every method, I begin the section with a note about the recommendation. Hope that clarifies.

damithc:
> I added a note at the top. For every method, I begin the section with a note about the recommendation. Hope that clarifies.

@tlylt Looks good. So, I guess option 2 is the one that aligns with AB3? In that case, do we need to update that AB3 branch every time we release a new MarkBind version (due to package.json and package-lock.json)?

tlylt:
> > I added a note at the top. For every method, I begin the section with a note about the recommendation. Hope that clarifies.
> 
> @tlylt Looks good. So, I guess option 2 is the one that aligns with AB3?

Yes

> In that case, do we need to update that AB3 branch every time we release a new MarkBind version (due to package.json and package-lock.json)?

If we want to update to the new version then yes. Again, I strongly recommend this approach because pinning the dependency is so much safer than going with option 1, which might just be slightly more convenient.

Some reasons already discussed here https://github.com/se-edu/addressbook-level3/pull/156#discussion_r1105298426

damithc:
@tlylt Sounds good. I asked because I anticipate a new release soon (to fix the incorrect scrolling issue) which AB3 adopters need to upgrade as well. On a related note, should we also add one more item under each of the three alternatives, to explain how to upgrade MarkBind later?

tlylt:
> @tlylt Sounds good. I asked because I anticipate a new release soon (to fix the incorrect scrolling issue) which AB3 adopters need to upgrade as well.

Yup, for every necessary update at the moment, I will need to bump the version there 😰.

> On a related note, should we also add one more item under each of the three alternatives, to explain how to upgrade MarkBind later?

Added.

damithc:
Looks good. For method 3, I thought the latest version is used by default? If that's the case, the only thing that needs to be mentioned in item 5 is how to use a specific version, right?

tlylt:
> Looks good. For method 3, I thought the latest version is used by default? If that's the case, the only thing that needs to be mentioned in item 5 is how to use a specific version, right?

Yes if you run it for the first time I believe. Then, there will be a cached version of the package that may or may not be cleared automatically, which will give you a stale version if you don't specify fetching the latest. https://stackoverflow.com/questions/73184553/npx-does-not-look-for-latest-version-of-package

damithc (approved):
Thanks for the quick fix @tlylt 
No further comments from me.

PR #2395, opened 2024-01-28
[Request for early review on functional code] Allow `markbind serve` to specify custom host #2382

### LINKED ISSUE #2382: Allow `markbind serve` to specify custom host

### Any related issues?

_No response_

Author Usability, CLI, DevOps

According to the [documentation](https://markbind.org/userGuide/cliCommands.html#serve-command), `markbind serve` starts a webserver at `127.0.0.1` aka `localhost`, but there does not seem to be a way to customise this.

My use case is that I have a separate machine as a dev server and I remote into this machine from my other devices to edit code.

However, as `localhost` does not expose the port to the local network, I can't access the previewed site remotely.

I would ideally like to set the host to `0.0.0.0` to expose on all interfaces.

Other potential use cases:

* Expose only from a specific interface
* Multi-person collaborative editing off a single codebase

Support passing `--host` and/or `-h` to the `markbind serve` command, e.g. `--host 0.0.0.0`.

~~As an extension, perhaps allow `host` and `port` to be specified in `site.json` as well (the CLI option will still take precedence in case both are specified)~~ – on second thought, having this might make `site.json` not as deploy-agnostic anymore, but I left it here as food for thought.

_No response_

### Additional context

_No response_

LamJiuFong:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**  
Resolves #2382 
Added `-a` option to the `serve` command

Default host/address is 127.0.0.1. If the custom host specified by users is not working, run server on the default host

**Testing instructions:**
Run `markbind serve -a <address>`, eg. `markbind serve -a 127.0.0.1`

Allow markbind serve to specify custom host

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**


packages/cli/src/lib/live-server/index.js
```diff
@@ -281,7 +281,13 @@ LiveServer.start = function(options) {
       setTimeout(function() {
         server.listen(0, host);
       }, 1000);
-    } else {
+    } else if (e.code === 'EADDRNOTAVAIL') {
+      console.log('address(host) %s is not available. Trying another address'.yellow, host);
+      setTimeout(function() {
+        server.listen(port, '0.0.0.0');
```
yucheng11122017, line 287: I'll prefer to make 127.0.0.1 as the default address if the host is invalid. Because 0.0.0.0 has some special meaning and might accidentally expose it to the rest of the interner (see [link](https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0)).

packages/cli/index.js
```diff
@@ -56,6 +56,7 @@ program
   .option('-p, --port <port>', 'port for server to listen on (Default is 8080)')
   .option('-s, --site-config <file>', 'specify the site config file (default: site.json)')
   .option('-d, --dev', 'development mode, enabling live & hot reload for frontend source files.')
+  .option('-a, --address <address>', 'specify the server address/host (Default is localhost')
```
yucheng11122017, line 59: Forgot a cloding bracket here 😔

yucheng11122017 (commented):
Hi @LamJiuFong thanks for the work! 
One thing is I think it might be better to keep 127.0.0.1 as the default if the customised port isn't valid. What do you think?

packages/cli/index.js
lhw-1, line 59: Instead of "localhost", let's specify the address as `127.0.0.1`.

packages/cli/src/lib/live-server/index.js
```diff
@@ -281,7 +281,13 @@ LiveServer.start = function(options) {
       setTimeout(function() {
         server.listen(0, host);
       }, 1000);
-    } else {
+    } else if (e.code === 'EADDRNOTAVAIL') {
+      console.log('address(host) %s is not available. Trying another address'.yellow, host);
```
lhw-1, line 285:
```suggestion
      console.log('%s is not available. Trying another address.'.yellow, host);
```

lhw-1 (commented):
The functional code should be good to go once you've addressed the comments - remember to update the relevant documentation and tests afterwards.

LamJiuFong:
Hi @yucheng11122017 @lhw-1, thank you for the suggestions. Initially I was thinking of setting 0.0.0.0 as the default host so that the users can run the server once there is any available host. However, I did overlook the safety issues of allowing this to happen. I think it's good to set '127.0.0.1' as the default host.

I will work on the descriptions too.

yucheng11122017:
> Do you think it is useful to add a warning for `0.0.0.0` specifically? I feel this way as our target audience are not only for technical people and could potentially unknowingly open up dangers to our non-technical users

Yeah this sounds good!

packages/cli/src/lib/live-server/index.js
```diff
@@ -158,7 +158,7 @@ function entryPoint(staticHandler, file) {
  */
 LiveServer.start = function(options) {
   options = options || {};
-  var host = options.host || '0.0.0.0';
+  var host = options.host !== undefined ? options.host : '127.0.0.1';
```
yucheng11122017, line 161:
```suggestion
  var host = options.host || '127.0.0.1';
```

yucheng11122017 (commented):
Hi @LamJiuFong thanks for the changes! Last thing is to add a warning as suggested by @itsyme when 0.0.0.0 is used. :)

LamJiuFong:
Would like to seek advice on the [latest commit](https://github.com/MarkBind/markbind/pull/2395/commits/d2a1d04b36397dde3974a71413b2b464c2633050), I display a warning and prompt the user to input "y" in the CLI to proceed with host 0.0.0.0 at their own risk. Is this a good idea to warn the users?

packages/cli/src/cmd/serve.js
```diff
@@ -63,13 +75,26 @@ function serve(userSpecifiedRoot, options) {
...
     mount: [],
   };
 
   site
     .readSiteConfig()
     .then(async (config) => {
+      if (serverConfig.host === '0.0.0.0') {
+        const response = await questionAsync(
+          'WARNING: Using the address \'0.0.0.0\' may pose vulnerabilities and is generally discouraged.'
+          + 'Proceed with caution? [y/N] ');
```
itsyme, line 89: Perhaps the vulnerabilities could be made more explicit, maybe something along the lines of ... could expose your server to the internet, which may pose security risks. .... As I feel that if the user is not aware of what vulnerabilities there are, they may choose to ignore the warning without fully understanding the situation. The rest of the message seems good!

LamJiuFong: Yup, I believe that would be better, thanks!

packages/cli/src/cmd/serve.js
```diff
@@ -111,8 +137,9 @@ function serve(userSpecifiedRoot, options) {
       const server = liveServer.start(serverConfig);
       server.addListener('listening', () => {
         const address = server.address();
-        const serveHost = address.address === '0.0.0.0' ? '127.0.0.1' : address.address;
-        const serveURL = `http://${serveHost}:${address.port}`;
+        const serveHost = address.address;
```
itsyme, line 140: What do you think about adding some validation to the user input host address?

I believe that this could be better for UX as
- it checks the host address before building the pages (which could take some time)
- it better handles the errors that are being thrown. Currently users will be exposed to errors like `Error: getaddrinfo ENOTFOUND 127.0.300.1`

I was thinking some sort of simple validation like checking if the input is an actual IP address and if any number in the address exceeds 255

packages/cli/src/cmd/serve.js
itsyme: Thanks for making the change the latest warning message looks good to me!

packages/cli/src/cmd/serve.js
LamJiuFong: Hi, good idea! I was thinking if we want to do this validation, we can actually just do the validations (including check if port is unavailable, check if host is being used etc) all at once before building the pages. Currently, all these validations are done after building the pages. 

Since this PR is focusing on adding another small feature, I suggest we create a separate issue and pull request to specifically address this validation issue. What do you think?

itsyme: Yep I think a separate issue is fine as well! I believe there is some level of error handling for when the host is being used. From my experience if I specified a used host, the nearest unused hostname will be used, which allows the user to still serve the site. I think we can discuss about what needs to be validated once the issue is opened!

packages/cli/src/lib/live-server/index.js
```diff
@@ -158,7 +158,7 @@ function entryPoint(staticHandler, file) {
...
   var port = options.port !== undefined ? options.port : 8080; // 0 means random
```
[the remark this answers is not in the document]
LamJiuFong: Thanks for the suggestion, I wasn't aware of the difference between `null` and `undefined` in JS. Thank you for reminding.
I will also set line 161 to be 
`var host = options.host ?? '127.0.0.1';`

LamJiuFong:
@yucheng11122017 @itsyme 
I researched on the difference between the null coalescing operator `??` and the OR operator `||` and found out that the main difference between them is during the handling of falsy values like 0, empty string, empty list etc.
eg. ` 0 ?? 100` will return `0` but `0 || 100` will return `100`. 

I was thinking if we should use the `??` operator while handling missing host and port (line 161-162) so that if the user specifically enter `0`, markbind will not default it to a specific host/port defined by us. 
eg. using `0 || 127.0.0.1` will set the host to be 127.0.0.1 but `0 ?? 127.0.0.1` will set the host to be 0.0.0.0
I feel like the latter is more suitable. What do you think?

kaixin-hc:
Sounds good to me :) do make the change, I think this PR is almost there!

itsyme:
> @yucheng11122017 @itsyme I researched on the difference between the null coalescing operator `??` and the OR operator `||` and found out that the main difference between them is during the handling of falsy values like 0, empty string, empty list etc. eg. ` 0 ?? 100` will return `0` but `0 || 100` will return `100`.
> 
> I was thinking if we should use the `??` operator while handling missing host and port (line 161-162) so that if the user specifically enter `0`, markbind will not default it to a specific host/port defined by us. eg. using `0 || 127.0.0.1` will set the host to be 127.0.0.1 but `0 ?? 127.0.0.1` will set the host to be 0.0.0.0 I feel like the latter is more suitable. What do you think?

I agree! My opinion is that we should use the nullish coalescing operator `??` in this case as this line should only be handling the responsibility of checking if any input is present from the user. For the other cases (e.g. 0, empty string, etc.), the responsibility should lie with the validation function instead.

yucheng11122017:
> > @yucheng11122017 @itsyme I researched on the difference between the null coalescing operator `??` and the OR operator `||` and found out that the main difference between them is during the handling of falsy values like 0, empty string, empty list etc. eg. ` 0 ?? 100` will return `0` but `0 || 100` will return `100`.
> > I was thinking if we should use the `??` operator while handling missing host and port (line 161-162) so that if the user specifically enter `0`, markbind will not default it to a specific host/port defined by us. eg. using `0 || 127.0.0.1` will set the host to be 127.0.0.1 but `0 ?? 127.0.0.1` will set the host to be 0.0.0.0 I feel like the latter is more suitable. What do you think?
> 
> I agree! My opinion is that we should use the nullish coalescing operator `??` in this case as this line should only be handling the responsibility of checking if any input is present from the user. For the other cases (e.g. 0, empty string, etc.), the responsibility should lie with the validation function instead. Wdyt @yucheng11122017 ?

Sounds good! Thanks for the research. Remember to write it in your knowledge base to get some credit HAHA

yucheng11122017 (approved):
LGTM! Thanks for the work @LamJiuFong

itsyme (approved):
LGTM! Good work @LamJiuFong!

yucheng11122017:
@all-contributors please add @LamJiuFong

yucheng11122017:
https://github.com/all-contributors please add @LamJiuFong for code

yucheng11122017:
@all-contributors please add @LamJiuFong for code

Issue #2404, opened 2024-02-05
Support soft-wrapping for code blocks


damithc:
For cases like the below, can we provide some sort of soft-wrapping support (similar to Intellij IDEA) so that the hidden part of the text can be made visible, especially for printing?

![image](https://github.com/MarkBind/markbind/assets/1673303/8d91affc-265b-423b-a5d6-217a7703e60d)

Tim-Siu:
Adding Intellij IDEA like indention-aware code wrapping can be quite hard. A pure CSS approach may not be viable due to the nature of how whitespace is handled in HTML and CSS. Current logic in markdown-it does not involve indentation. We might have to work on the parser there.

A naive soft wrapping preserves the original line breaks and spaces but does not add additional indentation to each wrapped line.
<img width="995" alt="Screenshot 2024-02-05 at 13 01 46" src="https://github.com/MarkBind/markbind/assets/61866948/ea22ed2b-4f5b-4284-8c2d-6b52368228a9">

Should we consider having this naive soft-wrapping in printing first?

yucheng11122017:
Hi @damithc, there is some existing functionality already with the plugin [CodeWrap](https://markbind.org/userGuide/usingPlugins.html#plugin-codeblockwrapbuttons).
If the user has already wrapped the text using the plugin, when printing the wrap is maintained (ie the code is wrapped) 
![image](https://github.com/MarkBind/markbind/assets/77204346/282000bc-8335-4969-ad89-69d00076f7db)

Otherwise, it is as shown in your display. 
Would it be sufficient to make wrapping text the default behaviour when this is plugin is enabled so that when users print, it should already be wrapped unless chosen to do so? 
Or should be code will be wrapped as long as there is printing? (regardless of whether this plugin is enabled and whether or not the wrap text button was clicked)

damithc:
[...]

PR #2413, opened 2024-02-11
Automatically add line-numbers when soft-wrapping (#2404)

Tim-Siu:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Currently, code wrapping is supported through the `codeBlockWrapButton` plugin. A problem with wrapping is that the indention is lost when a line of code is wrapped, which may alter the semantic meaning of the code for certain programming languages (e.g., Python). A workaround is that we automatically add line-numbers for the code when wrapping is applied, regardless of whether it is explicitly specified by the site creator.

This PR hasn't addressed the printing part of (#2404) yet.

**Testing instructions:**
Add a fenced code block with a really long line in the test site. Activate the `codeBlockWrapButton` plugin by adding 
```
  "plugins": [
    "codeBlockWrapButtons"
  ]
```
in `site.json` of the test site.

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core/src/plugins/codeBlockWrapButtons.ts
+++ b/packages/core/src/plugins/codeBlockWrapButtons.ts
@@ -42,16 +42,19 @@ function toggleCodeBlockWrap(element) {
   const code = pre.querySelector('code');
   const classList = code.classList; 
 
+  if (code.dataset.originalLineNumbers === undefined) {
+    code.dataset.originalLineNumbers = classList.contains('line-numbers').toString();
+  }
+
   if (classList.contains('wrap')) {
     classList.remove('wrap');
-    if (code.dataset.addedLineNumbers) { 
+    if (code.dataset.originalLineNumbers === 'false') {
       classList.remove('line-numbers');
     }
   } else {
     classList.add('wrap');
-    if (!code.dataset.originalLineNumbers) { 
-      classList.add('line-numbers'); 
-      code.dataset.addedLineNumbers = true;
+    if (code.dataset.originalLineNumbers === 'false') {
+      classList.add('line-numbers');
     }
   }
 }
--- a/packages/core/src/plugins/codeBlockWrapButtons.ts
+++ b/packages/core/src/plugins/codeBlockWrapButtons.ts
@@ -37,16 +37,25 @@ function getButtonHTML() {
 }
 
 const wrapCodeBlockScript = `<script>
-    function toggleCodeBlockWrap(element) {
-      const pre = element.parentElement.parentElement;
-      const classList = pre.querySelector('code').classList; 
-      if (classList.contains('wrap')) {
-          classList.remove('wrap');
-      } else {
-          classList.add('wrap')
-      }
+function toggleCodeBlockWrap(element) {
+  const pre = element.parentElement.parentElement;
+  const code = pre.querySelector('code');
+  const classList = code.classList; 
+
+  if (classList.contains('wrap')) {
+    classList.remove('wrap');
+    if (code.dataset.addedLineNumbers) { 
+      classList.remove('line-numbers');
     }
-    </script>`;
+  } else {
+    classList.add('wrap');
+    if (!code.dataset.originalLineNumbers) { 
+      classList.add('line-numbers'); 
+      code.dataset.addedLineNumbers = true;
+    }
+  }
+}
+</script>`;
 
 export = {
   getScripts: () => [wrapCodeBlockScript],


EltonGohJH:
> Hmmm I see.. this current implementation adds line numbers regardless of user's settings. I'm concerned about whether this would be flexible enough for the user What do @kaixin-hc @EltonGohJH @itsyme think?

I agreed that it will be inflexible and the users may not expect this behaviour. Maybe we can target the printing aspect?

Wdyt?

@yucheng11122017 @kaixin-hc @itsyme

kaixin-hc:
It feels like the main issue is what if it wraps line numbers the user does not expect to wrap (since not all languages consider indents the same way as python). If we force this change for all wrapping when clicked regardless of users settings, I'm not sure yet that there would be no undesired behaviour.

If when printing, not wrapping and wrapping but no line numbers both often distort the meaning unexpectedly ... Then I agree with Elton that a change made only for print makes sense? In other words when printing code blocks will auto wrap and add line numbers, otherwise it will follow existing behaviours.

Another alternative (but breaking) might be to have all code blocks automatically have line numbers that must be manually disabled, which would cause the default behaviour to be "correct" when printing if not otherwise overriden. But it's probably not worth making a breaking change for this.

yucheng11122017:
The current behavior would definitely be a breaking change as well. I would prefer to add only when printing as well 
@damithc is that ok?

damithc:
> If when printing, not wrapping and wrapping but no line numbers both often distort the meaning unexpectedly ... Then I agree with Elton that a change made only for print makes sense? In other words when printing code blocks will auto wrap and add line numbers, otherwise it will follow existing behaviours.

Yes, when printing, should be auto-wrapped (but only if there are lines longer than the container width), and line numbers should be added. However, if wrapping is not needed for a given code block, the current printing behavior should remain (i.e., no line numbers unless already enabled by the user), right?

> Another alternative (but breaking) might be to have all code blocks automatically have line numbers that must be manually disabled, which would cause the default behaviour to be "correct" when printing if not otherwise overriden. But it's probably not worth making a breaking change for this.

This is what we had in an earlier version but we changed it because we found most of the time code blocks don't need line numbers and it was a hassle to disable it in many places. So, we decided the default to be 'no line numbers'. 

> It feels like the main issue is what if it wraps line numbers the user does not expect to wrap (since not all languages consider indents the same way as python). If we force this change for all wrapping when clicked regardless of users settings, I'm not sure yet that there would be no undesired behaviour.

The current behavior of the line-wrapping plugin is already flawed because it doesn't differentiate between real line breaks and soft/fake line breaks. Alternatives:
1. Add a new symbol to indicate soft line breaks, like Intellij does it: This seems hard to do, and requires the reader to know the meaning of the new symbol.
 
image

2. Show line numbers (even if not enabled) when line wrapping: This seems easier to do, and more intuitive for the reader to understand. It may be unexpected for the reader to see line numbers, but line numbers are harmless, so should be fine?

kaixin-hc:
Thanks for the additional context prof! Found #1671 where this is discussed and #1734 .

TLDR: agree about softwrapping with line numbers for overflowing code on print (but no change for non overflowing code) makes sense because it helps to avoid changing/losing the meaning of the code, which is the worst part in thr status quo.

Additional rambles on alternative 2:

Alternative 2, which is what I believe this PR is suggesting is followed on several websites. GitHub has line numbers and softwraps all code on mobile:



But in the markbind context, it's possible this will not make sense... 

Admittedly, in cases like that (an ASCII image that overflows the page size) whether it softwraps on printing or not it will still look weird, and is a rare edge case.

After doing some reading, it seems no line numbers is (frequently) preferred from a UX perspective, but individuals often have the opposite preference...

So I'm hesitant to say all code wrapped blocks should enable line numbers. If you wrap the code and the meaning doesn't change when you think it's a hard line break instead of a soft line break, I can see why a user who doesn't like line numbers would still want line numbers to be disabled even with code wrapped. Having the numbers go back on would be unexpected (and breaking?)

itsyme:
> > Hmmm I see.. this current implementation adds line numbers regardless of user's settings. I'm concerned about whether this would be flexible enough for the user What do @kaixin-hc @EltonGohJH @itsyme think?
> 
> I agreed that it will be inflexible and the users may not expect this behaviour. Maybe we can target the printing aspect?
> 
> Wdyt?
> 
> @yucheng11122017 @kaixin-hc @itsyme



> > Hmmm I see.. this current implementation adds line numbers regardless of user's settings. I'm concerned about whether this would be flexible enough for the user What do @kaixin-hc @EltonGohJH @itsyme think?
> 
> I agreed that it will be inflexible and the users may not expect this behaviour. Maybe we can target the printing aspect?
> 
> Wdyt?
> 
> @yucheng11122017 @kaixin-hc @itsyme

I agree with @EltonGohJH on targeting only the printing aspect as that is the only situation where the user is unable to scroll and the line numbers will improve the clarity in this situation.

> Yes, when printing, should be auto-wrapped (but only if there are lines longer than the container width), and line numbers should be added. However, if wrapping is not needed for a given code block, the current printing behavior should remain (i.e., no line numbers unless already enabled by the user), right?

I agree with this behaviour as well

yucheng11122017:
> > > Hmmm I see.. this current implementation adds line numbers regardless of user's settings. I'm concerned about whether this would be flexible enough for the user What do @kaixin-hc @EltonGohJH @itsyme think?
> > 
> > 
> > I agreed that it will be inflexible and the users may not expect this behaviour. Maybe we can target the printing aspect?
> > Wdyt?
> > @yucheng11122017 @kaixin-hc @itsyme
> 
> I agree with @EltonGohJH on targeting only the printing aspect as that is the only situation where the user is unable to scroll and the line numbers will improve the clarity in this situation.
> 
> > Yes, when printing, should be auto-wrapped (but only if there are lines longer than the container width), and line numbers should be added. However, if wrapping is not needed for a given code block, the current printing behavior should remain (i.e., no line numbers unless already enabled by the user), right?
> 
> I agree with this behaviour as well
Sounds great!

Tim-Siu:
I cannot think of a way to determine if wrapping happens. Currently, line-numbers are added in `markdown-it`. Wrapping seems to be handled by css and whether it actually happens is hard to tell. Can anyone suggest any solution to this?

yucheng11122017:
> I cannot think of a way to determine if wrapping happens. Currently, line-numbers are added in `markdown-it`. Wrapping seems to be handled by css and whether it actually happens is hard to tell. Can anyone suggest any solution to this?

Hi @Tim-Siu if im understanding this correctly, the idea is that line numbers will naturally show if there is line wrapping already? Because if there is wrapping, a new line will be present but the numbers doesn't increase. Please correct me if i'm understanding your question wrongly! :)

yucheng11122017:
> > I cannot think of a way to determine if wrapping happens. Currently, line-numbers are added in `markdown-it`. Wrapping seems to be handled by css and whether it actually happens is hard to tell. Can anyone suggest any solution to this?
> 
> Hi @Tim-Siu if im understanding this correctly, the idea is that line numbers will naturally show if there is line wrapping already? Because if there is wrapping, a new line will be present but the numbers doesn't increase. Please correct me if i'm understanding your question wrongly! :)

Hi @Tim-Siu, is there anything you want to clarify? Is there any problem you are encountering?

Tim-Siu:
> > > I cannot think of a way to determine if wrapping happens. Currently, line-numbers are added in `markdown-it`. Wrapping seems to be handled by css and whether it actually happens is hard to tell. Can anyone suggest any solution to this?
> > 
> > 
> > Hi @Tim-Siu if im understanding this correctly, the idea is that line numbers will naturally show if there is line wrapping already? Because if there is wrapping, a new line will be present but the numbers doesn't increase. Please correct me if i'm understanding your question wrongly! :)
> 
> Hi @Tim-Siu, is there anything you want to clarify? Is there any problem you are encountering?

Yes, you understanding of the current implementation in the PR is correct. I haven't figured out ways to implement the suggestions though. I am still investigating.

yucheng11122017:
Hi @Tim-Siu is this PR still relevant?

Tim-Siu:
> Hi @Tim-Siu is this PR still relevant?

Hi @yucheng11122017 it can be closed if we've decided to only target the printing aspect.

yucheng11122017:
Closed as no longer relevant - replaced by https://github.com/MarkBind/markbind/pull/2431

Issue #2430, opened 2024-02-19
Footnote numbered wrongly when importing footnote from include


yiwen101:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

#2203

### Tell us about your environment

macOS

### MarkBind version

5.2.0

### Describe the bug and the steps to reproduce it

Adding this to "topic1.md"
```
## Footnotes

1 + 1 = 2 ^[Math1]
<div id="examples" class="d-none">

1 + 1 = 2 ^[Math2]
</div>

third test^[last footnote]
```

Add this to index.md:
```

footnote1^[first footnote]

<include src = "./topic1.md"></include>
```

We see:
<img width="238" alt="Screenshot 2024-02-19 at 13 08 55" src="https://github.com/MarkBind/markbind/assets/121547057/c21f495f-4751-436a-9244-a138df4eac6c">

### Expected behavior

1 footnotes that not visible to users are not rendered?
2 footnotes from include are properly indexed and displayed.

### Anything else?

I feel it is reasonable to fix this first before fixing #2203, as otherwise we are forced to add either:
1 weak test case that do not take care of the id of the footnote
2 test case that expect to get footnote with wrong ids

Both ways are not idea.

PR #2441, opened 2024-03-03
Debug cannot import footnote from hash

### LINKED ISSUE #2203: Footnote not being imported correctly in includes when hash is used

### Any related issues?

_No response_

Windows 10

### MarkBind version

V4.1.0

If includes src has a hash, footnotes are not imported over correctly. 

Example:
```
<include src="userGuide/syntax/footnotes.md#examples"> 
</include>
```
will not render the footnote at the bottom. 

[Source code](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/includePanelProcessor.ts#:~:text=wrapped%20include%20element-,if%20(hash)%20%7B,%7D,-if%20\(isTrim) causing the issue: 

```
  if (hash) {
    const $ = cheerio.load(actualContent);
    const actualContentOrNull = $(hash).html();
    actualContent = actualContentOrNull || '';

    if (actualContentOrNull === null && !isOptional) {
      const error = new Error(`No such segment '${hash}' in file: ${actualFilePath}\n`
       + `Missing reference in ${context.cwf}`);
      logger.error(error);

      actualContent = cheerio.html(createErrorNode(node, error) as cheerio.Element);
    }
  }
```

The footnotes are rendered at the end outside of the element with the hash hence it is not imported over. 

This is currently affecting the [UG Reader Facing Features Footnote](https://markbind.org/userGuide/readerFacingFeatures.html#:~:text=details...-,Footnotes,-1%20%2B%201%20%3D%202)

image

Footnote reference is not rendering at all right now because there is no new line above the code. After it is added, the footnote is not created due to this issue
There is a math footnote at the bottom but the example is not referring to that. This is created due to the code in details which also makes a math footnote although the div is hidden.

### Expected behavior

Expect the footnotes to be imported over the includes even if there is a hash supplied.

### Anything else?

_No response_

yiwen101:
  Resolve #2203  (but #2430 still persists) 
- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

**Testing instructions:**

Fix issue of cannot import footnote from hash

---

- [x] No unrelated changes 

currently, when importing from hash, footnote is missing. For example, with 
```
<div id="import">

text^[footnote]
</div>
```
and
```
<include src="contents/topic1.md#import"/>
```
this is rendered
Screenshot 2024-03-04 at 03 30 29
but we expect
Screenshot 2024-03-04 at 03 28 23

This is fixed by using regular pattern to greedily extract and append the missing information.

Please note that #2430 still persists (the index of the imported footnotes may remain be conflicting). This PR only fixes only failure in importing footnote from hash, but did not touch the issue of wrong index.

---

**Reviewer checklist:**


packages/core/test/unit/html/includePanelProcessor.test.ts
```diff
@@ -87,6 +87,44 @@ test('includeFile replaces <include src="doesNotExist.md" optional> with empty <
...
+  const json = {
+    'index.md': index,
+    'include.md': include,
+  };
+
+  fs.vol.fromJSON(json, '');
+
+  const nodeProcessor = getNewDefaultNodeProcessor();
+  const result = await nodeProcessor.process(indexPath, index);
+  // to make lint happy; if write as array.join, some lines more than 100 characters;
```
kaixin-hc, line 115: This is really quite hard to read, especially with the necessary spaces... I also don't love the defining of throwaway constants

Can you explore 1) one long string with string concatenation at reasonable points 2) disabling lint for just that section to see if it is more readable?

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +230,24 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
     const $ = cheerio.load(actualContent);
     const actualContentOrNull = $(hash).html();
     actualContent = actualContentOrNull || '';
+    if (actualContent !== '') {
+      const footnotePattern = /<a aria-describedby="footnote-label" href="#fn-\d+-\d+">\[\d+\]/g;
+      const matchs = actualContent.match(footnotePattern);
+      if (matchs) {
```
kaixin-hc, line 236: nit: the plural of match is matches. can you change the instances of `matchs` to `matches`?

kaixin-hc (commented):
Yay, thanks for the tests!

The issue https://github.com/MarkBind/markbind/issues/2203 also raises the example of the footnote in reader facing features. However, this is not fixed by your PR (is it because the new line isn't there?)

Can you also fix this documentation bug to demonstrate that this feature works?

packages/core/test/unit/html/includePanelProcessor.test.ts
yiwen101: Thanks for the comment; how do you find it after this [commit](https://github.com/MarkBind/markbind/pull/2441/commits/eecebc2f249dcc2afe67c5fe18110a57d6d610cd)?

packages/core/test/unit/html/includePanelProcessor.test.ts
```diff
@@ -112,15 +112,28 @@ test('includeFile import footnote from hash', async () => {
...
-  const a = '<div>\n<p>text<trigger for="pop:footnotefn-1-1"><sup class="footnote-ref">';
-  const b = '<a aria-describedby="footnote-label" href="#fn-1-1">[1]</a></sup></trigger></p>';
-  const c = '</div><hr class="footnotes-sep">\n<section class="footnotes">\n';
-  const d = '<ol class="footnotes-list">\n<popover id="pop:footnotefn-1-1">\n';
-  const e = '            <template #content><div>\n              <p>footnote</p>\n\n';
-  const f = '            </div></template>\n          </popover>\n\n<li id="fn-1-1"';
-  const g = ' class="footnote-item"><p>footnote</p>\n</li>\n\n</ol>\n</section>\n';
-  const expected = a + b + c + d + e + f + g;
+  // to make lint happy
+  let longStr = '<p>text<trigger for="pop:footnotefn-1-1"><sup class="footnote-ref"><a aria-describedby="';
```
kaixin-hc, line 116: nit: can we have a better name as opposed to longStr? perhaps something like "footnoteStr" ? 

def think the array.join() is a lot more readable!

and actually i was thinking something like the following!
```typescript
let footnoteStr = 'some text etc ...' 
     + 'more text etc...'
     + 'even more text etc';
```

packages/core/test/unit/html/includePanelProcessor.test.ts
kaixin-hc: have replied

kaixin-hc:
Behavior for footnotes here is definitely unexpected...

Screenshot 2024-03-04 at 3 23 04 PM
Screenshot 2024-03-04 at 3 23 16 PM

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +230,24 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
...
+    if (actualContent !== '') {
+      const footnotePattern = /<a aria-describedby="footnote-label" href="#fn-\d+-\d+">\[\d+\]/g;
+      const matches = actualContent.match(footnotePattern);
+      if (matches) {
+        const hrefPattern = /href="#(fn-\d+-\d+)"/;
+        let footnoteNumber = 1;
+        let toAppend = '<mb-temp-footnotes>';
+        matches.forEach((match) => {
+          const href = match.match(hrefPattern)![1];
+          const replaceTo = `<a aria-describedby="footnote-label" href="#${href}">[${footnoteNumber}]`;
```
yucheng11122017, line 242: Why is it necessary to do this replace here? Why are you changing the number for the footnote?

yucheng11122017 (commented):
Thanks for the work @yiwen101! 
This is a rather tricky bug to fix so appreciate the work. 

I have some concerns about the ordering of the footnote. Could you explain why you change the footnote number and why the appended footnote item will has the number 1-1? 

Thanks!

packages/core/test/unit/html/includePanelProcessor.test.ts
```diff
@@ -87,6 +87,57 @@ test('includeFile replaces <include src="doesNotExist.md" optional> with empty <
...
+  const json = {
+    'index.md': index,
+    'include.md': include,
+  };
+
+  fs.vol.fromJSON(json, '');
+
+  const nodeProcessor = getNewDefaultNodeProcessor();
+  const result = await nodeProcessor.process(indexPath, index);
+  // to make lint happy
```
yucheng11122017, line 115: Also could you write a better comment here? Why is lint complaining and what is done to fix it

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +230,24 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
     const $ = cheerio.load(actualContent);
     const actualContentOrNull = $(hash).html();
     actualContent = actualContentOrNull || '';
+    if (actualContent !== '') {
+      const footnotePattern = /<a aria-describedby="footnote-label" href="#fn-\d+-\d+">\[\d+\]/g;
```
yucheng11122017, line 234: oh also small thing but i was wondering if using cheerio here makes more sense?

packages/core/test/unit/html/includePanelProcessor.test.ts
yiwen101: Thanks for clarifying; will be working on that.

yiwen101:
> Behavior for footnotes here is definitely unexpected...
> 
> Screenshot 2024-03-04 at 3 23 04 PM Screenshot 2024-03-04 at 3 23 16 PM

emm, thanks for pointing out this (and sorry for not examining it closely);will be working on it.

yiwen101:
> Thanks for the work @yiwen101! This is a rather tricky bug to fix so appreciate the work.
> 
> I have some concerns about the ordering of the footnote. I understand you aren't trying to address the wrong ordering here but I think this might result in overlaps - not just an issue with wrong ordering. Could you explain why you change the footnote number and why the appended footnote item will has the number 1-1?
> 
> Thanks!

Thanks for the careful review. 
The the number 1-1 is an unintended bug, thank you for pointing it out!
The footnoteNumber is just a temporary placeHolder; the thinking is that when I/somebody attempt 2430 in the future and add footnoteNumber as an argument to this method, the rest of the code does not need to be changed (wait, just come to me, how to do passing by reference in javascript? nevermind, there must be a way to do it).

packages/core/test/unit/html/includePanelProcessor.test.ts
yiwen101: Fixed in latest commit
Screenshot 2024-03-14 at 16 31 39

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +230,24 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
...
+      if (matches) {
+        const hrefPattern = /href="#(fn-\d+-\d+)"/;
+        let footnoteNumber = 1;
+        let toAppend = '<mb-temp-footnotes>';
+        matches.forEach((match) => {
+          const href = match.match(hrefPattern)![1];
+          const replaceTo = `<a aria-describedby="footnote-label" href="#${href}">[${footnoteNumber}]`;
+          footnoteNumber += 1;
+          actualContent = actualContent.replace(match, replaceTo);
+          toAppend += `\n<li id="${href}" class="footnote-item">${$('#fn-1-1.footnote-item')!.html()}</li>`;
```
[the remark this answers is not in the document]
yiwen101: Yes, this is a bug. Thank you for pointing this. Fixed in latest commit.

packages/core/test/unit/html/includePanelProcessor.test.ts
yiwen101: Fix in latest commit; no longer has this comment by writing in the style suggested by @kaixin-hc

packages/core/src/html/includePanelProcessor.ts
yiwen101: Emm, but is not to use cheerio, we need the exact id (eg: `#${href}.footnote-item`)? So we still need to pattern matching (and also to replace the index in the original content).

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +230,24 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
     const $ = cheerio.load(actualContent);
     const actualContentOrNull = $(hash).html();
     actualContent = actualContentOrNull || '';
+    if (actualContent !== '') {
+      const footnotePattern = /<a aria-describedby="footnote-label" href="#fn-\d+-\d+">\[\d+\]/g;
+      const matches = actualContent.match(footnotePattern);
+      if (matches) {
+        const hrefPattern = /href="#(fn-\d+-\d+)"/;
+        let footnoteNumber = 1;
```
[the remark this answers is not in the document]
yiwen101: Similar as the next issue; addressed in latest commit

packages/core/src/html/includePanelProcessor.ts
yiwen101: I have improved this with the following changes:
Screenshot 2024-03-14 at 16 54 05

where the default value is: 
Screenshot 2024-03-14 at 16 54 00

last modification: 
Screenshot 2024-03-14 at 16 54 13

yiwen101: Motivation being, as this initially is to fix the error in display footnote in the user doc, I decided to be defensive and make sure that at least the footnote from a single hash import is going to look good (while not go beyond the scope and do not make changes that is outside the` hash` part)

yiwen101:
@kaixin-hc @yucheng11122017 
I have made the requested changes, and is ready for review;
Summary of the changes: 
1 improve the readability of the test
2 improve the code quality issue of hardcoded `footnoteNumer = 1`
3 fix bug. 
Some additional note on to the third point: 
Current behavior of serving user doc: 
Screenshot 2024-03-14 at 16 18 22
Screenshot 2024-03-14 at 16 18 29

The bahaviour of the master branch 

(after adding empty to enable the markdown:)
Screenshot 2024-03-14 at 16 15 47
Screenshot 2024-03-14 at 16 14 54
(no pop up, when mouse is over; no response when clicked)
Screenshot 2024-03-14 at 16 14 42

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -165,6 +165,15 @@ function _deleteIncludeAttributes(node: MbNode) {
   delete node.attribs.omitFrontmatter;
 }
 
+function buildGetNextFootnodeNumber() {
+  let footnoteNumber = 1;
+  function defaultGetFootnoteNumber() {
```
yucheng11122017, line 170:
```suggestion
  function getFootnoteNumber() {
```

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -165,6 +165,15 @@ function _deleteIncludeAttributes(node: MbNode) {
   delete node.attribs.omitFrontmatter;
 }
 
+function buildGetNextFootnodeNumber() {
+  let footnoteNumber = 1;
```
yucheng11122017, line 169: Can we just let this start from 0 in this case? So that we don't have this confusing add and then delete below

packages/core/src/html/includePanelProcessor.ts
yucheng11122017: Yeah I get that cheerio will do the same thing but I think its cleaner to use cheerio to select out and append. In that case, we can use cheerio node to do the modifications below as well. What do you think?

yucheng11122017 (commented):
Hi @yiwen101 thanks for the work! Some more corrections

packages/core/src/html/includePanelProcessor.ts
yiwen101: @yucheng11122017 
Thank you for the clarification; I have explored cheerio apis and made the requested changes. Following is a quick snapshot:
Screenshot 2024-03-14 at 23 47 04

yiwen101: (console.log on last line was removed in later commits)

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +240,26 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
...
+      const footnodeHrefs = hashNode.find('a[aria-describedby="footnote-label"]')
+        .map(function (this: any) {
+          $(this).text(`[${getNextFootnodeNumber()}]`);
+          return $(this).attr('href');
+        })
+        .get();
+      if (footnodeHrefs.length > 0) {
+        const tempFootnotes = $('<mb-temp-footnotes></mb-temp-footnotes>');
+        footnodeHrefs.forEach((href) => {
+          const listItem = $('<li></li>').attr('id', href.substring(1))
```
yucheng11122017, line 254: Could you add a comment to explain why we need to do a substring here? It is just to get rid of the hash in front right?

yucheng11122017 (approved):
Hi @yiwen101 thanks so much for cleaning up the code! I think using cheerio makes it more consistent with the rest of the code base. Just addition of one comment and I think this PR is ready to merge. 

Thanks for the good work!

packages/core/src/html/includePanelProcessor.ts
yiwen101: Thank you for the comment; has added the the latest commit

packages/core/src/html/includePanelProcessor.ts
```diff
@@ -230,6 +240,27 @@ export function processInclude(node: MbNode, context: Context, pageSources: Page
...
+      const footnodeHrefs = hashNode.find('a[aria-describedby="footnote-label"]')
+        .map(function (this: any) {
+          $(this).text(`[${getNextFootnodeNumber()}]`);
+          return $(this).attr('href');
+        })
+        .get();
+      if (footnodeHrefs.length > 0) {
+        const tempFootnotes = $('<mb-temp-footnotes></mb-temp-footnotes>');
+        footnodeHrefs.forEach((href) => {
+          // substring to remove the # from the href
```
yucheng11122017, line 254:
```suggestion
          // substring function called to remove the # from the href
```

PR #2375, opened 2023-10-28
Add horizontal spacing between the icon and text in the custom icon list #2352

Ashleeey1224:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [x] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

This PR resolves issue https://github.com/MarkBind/markbind/issues/2352 by adding i-spacing attribute for modifying horizontal spacing between the icon and text in the custom icon list. 
- This feature enhancement is based on SPWwj's implementation of CustomListIconProcessor.ts.
- Refer issues: #2352, #2316.

**Overview of changes:**
1. By defining "{i-spacing="xem"}" attribute, user can customise the horizontal distance between icon and text. x means numbers.

  - CODE:

  `* **Item 1**: i-spacing is 0.15em {icon="glyphicon-education" i-width="70px" i-spacing="0.15em"}`
  `* **Item 2**: i-spacing is 0.55em{i-width="70px" i-spacing="0.55em"}`
  ` * Item 2.1: no i-spacing attribute, default is 0.35em {icon="fas-question-circle"}`
  `  * Item 2.2: `
  ` * Item 2.3: overwrite i-spacing attribute by 2em {i-spacing="2em"}`
  `* **Item 3**{i-width="70px"}`
  `  * Item 3.1: {icon="glyphicon-book" i-width="70px"}`
  `  * Item 3.2: {icon="/images/AzurLane_roma copy.png" i-size="10px" i-class="rounded"}`
<br>

  - OUTPUT:

<p align="center">
Screenshot 2023-10-29 at 12 02 50 am
</p>

**Anything you'd like to highlight/discuss:**

Once a i-spacing attribute is assigned to an icon, it will continue apply to icons at the same level (e.g., Item2.3, Item3.1, Item3.2), until it is overridden by a different i-spacing attribute (e.g., from Item 2.1 to Item 2.3).

**Testing instructions:**
1. Confirm that you have the correct branch for this pull request checked out on your local environment.
3. Establish a new MarkBind project using the markbind init command.
4. Experiment with the list using the provided syntax from above.

**Proposed commit message: (wrap lines at 72 characters)**
Add i-spacing attribute to custom icon list

- Providing an option to adjust spacing between the icon and text for enhanced user experience.

- Add "i-spacing" attribute to ICON_ATTRIBUTES and interface IconAttributes

[...]

PR #2444, opened 2024-03-05
Add text for icon

### LINKED ISSUE #2351: Use text as icon for lists

### Any related issues?

_No response_

Author Usability, Reader Usability, Syntax

## Issue Description

Referring to the PR discussion [here](https://github.com/MarkBind/markbind/pull/2316#discussion_r1263543059), we need a more intuitive way to represent text as icons for lists.

Some idea:
```
Step 1: blah blah
blah blah

Step 2: blah blah
blah blah
```

This resembles numbered lists, but without the auto-incrementing functionality.

_No response_

### Additional context

_No response_

yucheng11122017:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:
Resolves #2351

**Overview of changes:**
Add support for text as bullet point
Users can now add text to the list 
Users can control the size of the text using `t-size` and the class using `t-class`. This will be seperate from those used by the icon. 

This feature supports both text and icons and there is some overlap between the two. Since markdown processor is used to process the text, the text is able to support icons. Eg. if text = ":+1:" this will be be converted to the emoji. 
However, this is unable to support images hence the original feature is maintained.  

I'm also thinking about splitting IconAttributes into two - one for text and one for icon. If you think this is better, do let me know!

**Testing instructions:**
Example: 
```
* Item 1 { text="aaassatext" t-class="badge rounded-pill my-1 bg-primary text-white"}

* Item 2 { text="as" icon="/images/deer.jpg" t-size="10px" i-width="60px" i-height="44px" }
```

Add text support for icon

---

- [X] No unrelated changes 

---

**Reviewer checklist:**


docs/userGuide/syntax/lists.md
```diff
@@ -27,11 +27,11 @@
 * Item 2
   * Item 2.1 { icon="fas-file-code" }
   * Item 2.2
-* Item 3 { icon="fas-code-branch" }
+* Item 3 { text="3" }
```
EltonGohJH, line 30: Also refer to https://github.com/MarkBind/markbind/pull/2316

I think you are not showcasing what is this supposed to be used for `{ text="Step 2", class="text-warning" }`. (Sort of like using step2 as an icon)

From my understandings, it seem that this is to use text in place of icons (so I think you are not showcasing what this feature is for).

packages/core/src/html/CustomListIconProcessor.ts
```diff
@@ -10,14 +10,17 @@ interface EmojiData {
 
 const emojiData = emojiDictionary as unknown as EmojiData;
 
-const ICON_ATTRIBUTES = ['icon', 'i-width', 'i-height', 'i-size', 'i-class'];
+const ICON_ATTRIBUTES = ['icon', 'i-width', 'i-height', 'i-size', 'i-class', 'text', 't-size', 't-class'];
```
EltonGohJH, line 13: i-size and t-size is not isolated for each icon unless you set each of the manually. (So if you set 1 with t-size, it will affect the whole list.)

I think this is not explained in the documentation. I would lean towards this as a bug rather than a lack of documentation.

Should we fix this or create a new ticket hmmm?

EltonGohJH (changes requested):
The overall implementation looks good on my end.
There are some documentation issues and also a potential (existing) bug that I have found.

EltonGohJH:
@yucheng11122017 
I realised the first t-size and i-size one will affect the next ones on the same level is sort of intended.  The styling is just carried down the levels. I think the documentation mentioned `If you customize any item on a certain level, you must also customize the first item on that level. If not, the list will revert to its uncustomized form.`

I think this is too vague. Do you think we should explain this behaviour in more depth? (I think that this is more of a docs issue rather than bug now)

docs/userGuide/syntax/lists.md
```diff
@@ -27,11 +27,11 @@
...
   * Item 3.1 
-* Item 4 { icon="octicon-git-pull-request" }
+* Item 4 { text="Pull request" icon="octicon-git-pull-request" }
   * Item 4.1 { icon="mif-perm-media" }
-* Item 5 { icon="glyphicon-education" }
+* Item 5 { text="" icon="glyphicon-education" }
```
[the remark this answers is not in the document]
yucheng11122017: I was thinking that it would be good to show how to disable the text because of the cascading effect - if they don't want the rest of the bullets to have text, they would have to do something like this.

yucheng11122017:
> @yucheng11122017 I realised the first t-size and i-size one will affect the next ones on the same level is sort of intended. The styling is just carried down the levels. I think the documentation mentioned `If you customize any item on a certain level, you must also customize the first item on that level. If not, the list will revert to its uncustomized form.`

Yup I think this effect was intended by the original feature - which is why this is kept here. 
 
> I think this is too vague. Do you think we should explain this behaviour in more depth? (I think that this is more of a docs issue rather than bug now)

Ok sure! I'll update it

docs/userGuide/syntax/lists.md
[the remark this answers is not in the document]
kaixin-hc: (can we put a code comment...? hmm)

docs/userGuide/syntax/lists.md
kaixin-hc: Screenshot 2024-03-10 at 4 17 02 PM

Agree that this isn't really super clear that text is being used in place of an icon. It just kinda looks like the icon is removed?

yucheng11122017: This feature is to add the text next to the icon though - it doesn't replace the icon. 
I was thinking that it would be better if users can use both concurrently which is why there is the icon and then the text on the right. 
What do you all think? Is it better to replace the icon instead?

yucheng11122017:
Hi @kaixin-hc and @EltonGohJH, thank you for the comments! 
Realised that I did not push the documentation changes that I made... 😢 

One thing that I realised is that if the user disables both the text and the icon, there will basically be no bullet point. 
Do you think that this should be intended behaviour or should it revert back to the originally uncustomed bullet point?

docs/userGuide/syntax/lists.md
EltonGohJH: @yucheng11122017 
I thought it is to replace the icon.
image

yucheng11122017: Ok fair enough

I would prefer to leave the functionality like this though to allow users to use both concurrently

EltonGohJH (approved):
LGTM!
Let me double check tmr before I merge!

EltonGohJH:
> Hi @kaixin-hc and @EltonGohJH, thank you for the comments! Realised that I did not push the documentation changes that I made... 😢
> 
> One thing that I realised is that if the user disables both the text and the icon, there will basically be no bullet point. Do you think that this should be intended behaviour or should it revert back to the originally uncustomed bullet point?

I think that this behaviour is okay. Uncustomed bullet looks out of place with the icon imo. @kaixin-hc Any thoughts?

docs/userGuide/syntax/lists.md
```diff
@@ -18,31 +18,61 @@
...
+The customised icon appears for Item 2 but not for Item 1.1.
+
+Hence, if you customize any item on a certain level, you must also **customize the first item on that level**. If not, the list will revert to its uncustomized form.
+If you wish to remove the customization from the following levels, you can set `text` and/or `icon` to be an empty string `""`. 
+
+Example:
+<include src="codeAndOutput.md" boilerplate >
+<variable name="highlightStyle">markdown</variable>
+<variable name="code">
+* Item 1 { icon="glyphicon-education" text="Only for this bullet"}
```
EltonGohJH, line 61: I realised a nit.
Can we have space at the start and en

yucheng11122017:
> Adding screenshot for the sake of this discussion. Screenshot 2024-03-11 at 4 35 00 PM
> 
> I don't know how uncustom bullet point will look... But I will say I think it looks really strange to have no bullet point. It doesn't look like a list at all, and if the number of items continue, wouldn't it just look like a paragraph? :// Not a big fan

Yeah I agree. But I guess this is a very specific scenario where the user disables both for some reason. And they could easily just... NOT do that. 
The empty string is more to disable the text or the icon in the event where a user only wants an icon or only wants the text. So I think that this seems to be more of a user issue. 

> 
> In general I like the idea of more customisability and this looks robust! Especially being able to use emoji (maybe you could demonstrate it in the docs?)

Ok sounds good!

> 
> One reservation I have is that the text of the list and the text of the bullet point feel misaligned. But perhaps this can be looked at in another PR...

True. With the #2375, users would be able to adjust the space manually if needed. 

> 
> > "IconAttributes into two - one for text and one for icon"
> 
> --> I'm not sure why you would want to do this!
Ok sounds good!

damithc:
Suggestion: Also have example that looks like a real use case (i.e., some examples should show best-case scenarios, to get user buy-in). 

image

image

tlylt:
> Suggestion: Also have example that looks like a real use case (i.e., some examples should show best-case scenarios, to get user buy-in).

We should try finding a suitable place in our UG to use this feature (besides its documentation).

yucheng11122017:
What do you all think about something like this for using customized icons within the documentation? 
This is the icons page on the UG
image

EltonGohJH:
> What do you all think about something like this for using customized icons within the documentation? This is the icons page on the UG image

@yucheng11122017 
I like the idea but I feel that it might not be related to the scope of this PR right? Unless you mean change to icon and also using text with the icon?

damithc:
> What do you all think about something like this for using customized icons within the documentation?
> This is the icons page on the UG
> image

The reader is unlikely to realize which feature is being used ... 🤔 It needs to be more obvious.

kaixin-hc:
Your example looks legitimate, but isn't really convincing as to why I can't just type it out in text rather than changing the bullet point... 😅 

Maybe you can do something with emojis, or a pro-con list? And you might need to identify clearly what is being used...

damithc:
> Your example looks legitimate, but isn't really convincing as to why I can't just type it out in text rather than changing the bullet point... 😅

Yup. In fact, this flaw applies as long as there is only one line of content for each item, no matter what text/icon you use. The value of this feature only shines through when each item has more than one line of content in which case the list automatically applies an indent to 2nd and subsequent lines of content. See the example I gave for reference.

yucheng11122017:
How about this [part of the UG](https://markbind.org/userGuide/deployingTheSite.html#generic-steps-for-deploying-a-markbind-site)? 

image

I can change it to step 1, step 2, step 3, step 3.1, step 3.2

itsyme:
> How about this [part of the UG](https://markbind.org/userGuide/deployingTheSite.html#generic-steps-for-deploying-a-markbind-site)?
> 
> image
> 
> I can change it to step 1, step 2, step 3, step 3.1, step 3.2
> 
> I can see some areas in the DG [here](https://markbind.org/devdocs/devGuide/bootcamp/fixABug.html) to use this as well
> 
> what do you all think!

I think this may be a good place to use the feature if you adjust the steps into multi-lines, for example the first step could be
'Step 1: Create ...
             Note: Ensure that ..."

yucheng11122017:
Added an example in the UG!

kaixin-hc:
Screenshot 2024-03-23 at 2 41 46 PM
just visually this looks a little inconsistent to me! maybe we can change the list on top to use the same syntax?

Also, do fix merge conflicts so this is mergeable!

damithc:
> just visually this looks a little inconsistent to me! maybe we can change the list on top to use the same syntax?
> Also, do fix merge conflicts so this is mergeable!

The second list is a good example. Yup, we can use the same formatting for both lists.

kaixin-hc (approved):
lgtm! thanks yu cheng. i'll merge this later today if no more comments.

Tests wise i think we can ignore the unit tests failing since its tested with the test site + its mostly failing due to those sections not being tested yet it seems.

damithc:
Feature in action ...
before:
image


after:
image

Issue #2099, opened 2023-01-17
Consider implementing a way to test messages from the logger


lhw-1:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### What is the area that this feature belongs to?

Code Quality, Testing

### Is your feature request related to a problem? Please describe.

This issue was brought up in Pull Request #2053.

Currently, there is no way to implement unit tests for logging outputs from the winston logger; by allowing the current unit testing methods to cover logging outputs (e.g. warnings) as well, we can ensure consistent behavior across different components and increase the robustness of the system as a whole, not just what is visible.

### Describe the solution you'd like

The request is to consider the different approaches to allow the winston logger outputs to be detected by our current testing approaches. 

### Describe alternatives you've considered

Alternatively, as jest is adopted as the testing tool for MarkBind, we can investigate approaches to test winston logger using jest (which seems to already be feasible as can be seen in [this StackOverflow discussion](https://stackoverflow.com/questions/57771616/how-to-use-jest-to-mock-winston-logger-instance-encapsulated-in-service-class).

luminousleek:
Hi, I’d like to work on this issue, probably using Jest to mock Winston

luminousleek:
After having a bit of a think, I'm not sure whether we should be testing the console output itself, or what the Winston logger is called with. Testing the console output itself seems like we're also testing the Winston logger, which seems out of the scope of a unit test. Instead, I think we should just test that the Winston logger has been called with the appropriate logging level and message. So something like
```
import * as logger from '/packages/core/src/utils/logger'

// mock the warn function of the logger here

expect(logger.warn).toHaveBeenCalledWith(args...);
```

What do you think about it?

kaixin-hc:
Seems like a sound approach to me, I think its worth investigating along these lines!

PR #2463, opened 2024-03-15
Test logger calls in tests for NodeProcessor

luminousleek:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [x] Others, please explain: Improve tests

Partially addresses #2099
Related to #2053 

**Overview of changes:**
* Mocked the logger
* Added expectations to test if the logger is called with the correct input
* Added constants for expected logger input

_Mocking vs Spying_

I chose to mock the logger instead of spy on it so as to decouple the logger from this test. If I had spied on the logger, the actual implementation of the logger would have been used, which may also consume (slightly) more resources. Also, errors from `winston` might also interfere with the test (even though that's very unlikely).

One side-effect of mocking the logger is that the logger output will not be printed in the console, as seen in the screenshot below (left is spied, right is mocked).

Screenshot 2024-03-15 at 6 58 16 PM

If the logger output is still needed to be printed in the console, then a mock implementation of the `warn` method could look something like this
```javascript
jest.mock('../../../src/utils/logger', () => ({
  warn: jest.fn(msg => console.log(`warn: ${msg}`)),
}));
```
and would result in this being printed to the console

Screenshot 2024-03-15 at 9 22 20 PM

which does look pretty clunky. Do let me know which approach you prefer. 

_Inconsistent naming in `NodeProcessor.data.ts`_

The naming conventions of the test inputs in `NodeProcessor.data.ts` are inconsistent. Many of the elements there have a test to check if the slots do override the attributes of the elements, but they are named in two separate ways:
1. `PROCESS_[ELEMENT]_ATTRIBUTES_NO_OVERRIDE`
2. `PROCESS_[ELEMENT]_[ATTRIBUTE]_SLOT_TAKES_PRIORITY`

I think that's why the author of #2053 wrote another constant called `PROCESS_PANEL_HEADER_SLOT_TAKES_PRIORITY` when the constant `PROCESS_PANEL_ATTRIBUTES_NO_OVERRIDE` tests for the exact same thing. That's why I have an `expect` statement after each of these two tests, because in both times the logger is called with the same input. 

A PR could be done to make the naming convention consistent, as well as to remove the unnecessary test.

_Logger warnings still inconsistent_

With the merging of #2053, Panels, Dropdowns, Modals and Popovers print warnings to the console if there are slots overriding the element attributes. However, other elements such as Quizzes and QOptions do not print warnings when this happens. This is because in the code of `MdAttributeRenderer.ts`, not every attribute is checked to see if there is an overriding slot, so the logger will not warn if the unchecked attributes are overridden. 

A simple (though untested) fix could be to have a new method as follows
```javascript
processAttribute(node, attribute, isInline, slotName = attribute) {
    if (!this.hasSlotOverridingAttribute(node, attribute, slotName) {
        this.processAttributeWithoutOverride(node, attribute, isInline, slotName);
    }
}
```
and replace all the `processAttributeWIthoutOverride` methods with this.

**Testing instructions:**

```
Test logging calls in NodeProcessor

Previously, calls to the logger were not tested for in testing
NodeProcessor.

Let's mock the logger and test if it is being called correctly.
```
---

- [x] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core/test/unit/html/NodeProcessor.test.ts
+++ b/packages/core/test/unit/html/NodeProcessor.test.ts
@@ -80,10 +80,13 @@ test('processNode processes quiz attributes and inserts into dom as slots correc
 });
 
 test('processNode processes popover attributes and inserts into dom as slots correctly', () => {
+  const warnSpy = jest.spyOn(logger, 'warn');
   processAndVerifyTemplate(testData.PROCESS_POPOVER_ATTRIBUTES,
                            testData.PROCESS_POPOVER_ATTRIBUTES_EXPECTED);
   processAndVerifyTemplate(testData.PROCESS_POPOVER_ATTRIBUTES_NO_OVERRIDE,
                            testData.PROCESS_POPOVER_ATTRIBUTES_NO_OVERRIDE_EXPECTED);
+  expect(warnSpy).toHaveBeenCalledWith(testData.PROCESS_POPOVER_ATTRIBUTES_NO_OVERRIDE_HEADER_WARN_MSG);
+  expect(warnSpy).toHaveBeenCalledWith(testData.PROCESS_POPOVER_ATTRIBUTES_NO_OVERRIDE_CONTENT_WARN_MSG);
 });
 
 test('processNode processes tooltip attributes and inserts into dom as slots correctly', () => {
@@ -126,8 +129,10 @@ test('processNode processes dropdown header attribute and inserts into DOM as he
 });
 
 test('processNode processes dropdown with header slot taking priority over header attribute', () => {
+  const warnSpy = jest.spyOn(logger, 'warn');
   processAndVerifyTemplate(testData.PROCESS_DROPDOWN_HEADER_SLOT_TAKES_PRIORITY,
                            testData.PROCESS_DROPDOWN_HEADER_SLOT_TAKES_PRIORITY_EXPECTED);
+  expect(warnSpy).toHaveBeenCalledWith(testData.PROCESS_DROPDOWN_HEADER_SLOT_TAKES_PRIORITY_WARN_MSG);
 });
 
 test('markdown coverts inline colour syntax correctly', async () => {
--- a/packages/core/test/unit/html/NodeProcessor.test.ts
+++ b/packages/core/test/unit/html/NodeProcessor.test.ts
@@ -11,8 +11,7 @@ import { MbNode, parseHTML } from '../../../src/utils/node';
 jest.mock('../../../src/utils/logger', () => ({
   warn: jest.fn(),
 }));
-
-afterEach(() => {
+beforeEach(() => {
   jest.clearAllMocks();
 });
 
--- a/packages/core/test/unit/html/NodeProcessor.test.ts
+++ b/packages/core/test/unit/html/NodeProcessor.test.ts
@@ -14,7 +14,7 @@ jest.mock('../../../src/utils/logger', () => ({
 
 afterEach(() => {
   jest.clearAllMocks();
-})
+});
 
 /**
  * Runs the processNode or postProcessNode method of NodeProcessor on the provided
--- a/packages/core/test/unit/html/NodeProcessor.test.ts
+++ b/packages/core/test/unit/html/NodeProcessor.test.ts
@@ -2,10 +2,15 @@ import path from 'path';
 import cheerio from 'cheerio';
 import htmlparser from 'htmlparser2';
 import * as testData from './NodeProcessor.data';
+import * as logger from '../../../src/utils/logger';
 import { Context } from '../../../src/html/Context';
 import { shiftSlotNodeDeeper, transformOldSlotSyntax } from '../../../src/html/vueSlotSyntaxProcessor';
 import { getNewDefaultNodeProcessor } from '../utils/utils';
 import { MbNode, parseHTML } from '../../../src/utils/node';
+
+jest.mock('../../../src/utils/logger', () => ({
+  warn: jest.fn(),
+}));
 
 /**
  * Runs the processNode or postProcessNode method of NodeProcessor on the provided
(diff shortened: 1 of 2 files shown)


LamJiuFong:
@luminousleek Thanks for the great work! I have a few suggestions:

1. Instead of printing the msg to the console, can we just return a string and compare it against the expected string?
eg. change `jest.fn(msg => 'console.log(warn: ${msg}))` to `jest.fn(msg => 'warn: ${msg})` and perform `expected(...).toEqual(...)`. In that case, we are able to check the output message while not printing them. 

2. Since we are probably going to use the mock logger and the mock functions in other test files, do you think it is a good idea to create a manual mock instead so that we do not have to reimplement the mock functions again in other files?
For your reference, [jest-manual-mocks](https://jestjs.io/docs/manual-mocks)

3. If I'm not wrong, the mock logger is shared among all the tests in a file, in other words, the mock logger's data is not resetting for every test and is actually accumulated. I am thinking of adding something like :
```
afterEach(() => {
    jest.clearAllMocks();
});
```
 to reset the data after every test. 

Please feel free to add on your comments!

packages/core/test/unit/html/NodeProcessor.test.ts
```diff
@@ -2,11 +2,16 @@ import path from 'path';
...
 import htmlparser from 'htmlparser2';
 import * as testData from './NodeProcessor.data';
+import * as logger from '../../../src/utils/logger';
 import { Context } from '../../../src/html/Context';
 import { shiftSlotNodeDeeper, transformOldSlotSyntax } from '../../../src/html/vueSlotSyntaxProcessor';
 import { getNewDefaultNodeProcessor } from '../utils/utils';
 import { MbNode, parseHTML } from '../../../src/utils/node';
 
+jest.mock('../../../src/utils/logger', () => ({
+  warn: jest.fn(),
```
LamJiuFong, line 12: I am thinking of changing this to `warn: jest.fn(msg => 'warn: ${msg}')` so that we can check the output string as well. What do you think?

luminousleek:
> 2. Since we are probably going to use the mock logger and the mock functions in other test files, do you think it is a good idea to create a manual mock instead so that we do not have to reimplement the mock functions again in other files?
>    For your reference, [jest-manual-mocks](https://jestjs.io/docs/manual-mocks)

Yeah I think this is a good idea, will probably do it in another PR. Thanks for the link.

> 3. If I'm not wrong, the mock logger is shared among all the tests in a file, in other words, the mock logger's data is not resetting for every test and is actually accumulated. I am thinking of adding something like :
> 
> ```
> afterEach(() => {
>     jest.clearAllMocks();
> });
> ```
> 
> to reset the logger to a clean state after every test. Might not be very useful for now but I think it is safer to do so, because issues might arise in the future (eg. checking mock.calls.length)

Yup I agree, the length issue didn't occur to me until you mentioned it. Thanks for commenting on this PR!

itsyme:
> @luminousleek Thanks for the great work! I have a few ideas:
> 
> 1. I think it is fine to not print out the message to the console, using `toHaveBeenCalledWith()` to check the input of the logger is enough to ensure the correctness of the method.
> 2. Since we are probably going to use the mock logger and the mock functions in other test files, do you think it is a good idea to create a manual mock instead so that we do not have to reimplement the mock functions again in other files?
>    For your reference, [jest-manual-mocks](https://jestjs.io/docs/manual-mocks)
> 3. If I'm not wrong, the mock logger is shared among all the tests in a file, in other words, the mock logger's data is not resetting for every test and is actually accumulated. I am thinking of adding something like :
> 
> ```
> afterEach(() => {
>     jest.clearAllMocks();
> });
> ```
> 
> to reset the logger to a clean state after every test. Might not be very useful for now but I think it is safer to do so, because issues might arise in the future (eg. checking mock.calls.length)
> 
> Please feel free to add your comments!

Great point about the clearing mocks! I think that is particularly important to note as bugs that arise from not clearing mocks may be harder to spot when it happens. Great catch!

yucheng11122017:
Hi @luminousleek thank you very much! This looks like excellent and detailed work. 

- With regards to mocking vs spying, I think your reasons make sense 
- With regards, to naming conventions, yeah let's open an issue for this and tackle it in the future. 
- Lastly, with regards to the warnings, I'm actually not too sure why for some we don't have this warning and why for some we do. If one of the other developers can advice, that would be great. But if not, I think what you suggested (implementing warnings for all) would be the best way going forward.

yucheng11122017 (approved):
LGTM!

luminousleek:
> 3. If I'm not wrong, the mock logger is shared among all the tests in a file, in other words, the mock logger's data is not resetting for every test and is actually accumulated. I am thinking of adding something like :
> 
> ```
> afterEach(() => {
>     jest.clearAllMocks();
> });
> ```
> 
> to reset the logger to a clean state after every test. Might not be very useful for now but I think it is safer to do so, because issues might arise in the future (eg. checking mock.calls.length)

Added the `afterEach` teardown hook, @yucheng11122017 do please take a look (and sorry for doing this after your review)

itsyme:
> > 3. If I'm not wrong, the mock logger is shared among all the tests in a file, in other words, the mock logger's data is not resetting for every test and is actually accumulated. I am thinking of adding something like :
> > 
> > ```
> > afterEach(() => {
> >     jest.clearAllMocks();
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to reset the logger to a clean state after every test. Might not be very useful for now but I think it is safer to do so, because issues might arise in the future (eg. checking mock.calls.length)
> 
> Added the `afterEach` teardown hook, @yucheng11122017 do please take a look (and sorry for doing this after your review)
> 
> Edit: after looking at some code on StackOverflow regarding this, it seems that they put the `jest.clearAllMocks()` method in a `beforeEach` hook instead of an `afterEach` hook. I think this makes a bit more sense in the edge case that this test suite runs after another test suite which didn't reset the mocks after the other tests were run.

Good catch! I agree with the use of `beforeEach` as I feel that this would ensure that the tests in the file are guaranteed to have cleared mocks before each test is run rather than depending on another test file to ensure that the mocks are cleared. Here's an example for clarity:

Assume test 1 and test 2 are in two separate files and are run sequentially,

In `beforeEach`:
Test 1 clear mocks -> Test 1 runs -> Test 2 clear mocks -> Test 2 runs

In `afterEach`:
Test 1 runs -> Test 1 clear mocks -> Test 2 runs -> Test 2 clear mocks

In the case of using `afterEach` you can see that the clearing of mocks relies on the previous test 1 to clear it, whereas in `beforeEach` if the Test 2 run fails because mocks were not cleared, it is easier to tell that the mocks are not cleared as the Test 2 file does not have the `beforeEach` teardown hook. This makes each test file responsible for the clearing of mocks for their own tests.

itsyme (approved):
LGTM! Good catch of the use of `beforeEach`!

PR #1382, opened 2020-11-01
Intralink validator

wxwxwxwx9:
**What is the purpose of this pull request?**
- [x] Feature addition or enhancement

This PR resolves issue #824 by introducing a crawler that validates all intra-link during site generation.

This PR is a replacement of the PR #1375 as the branch from that PR would introduce a lot of merge conflict to the master branch. It was deemed that it would be easier to create another branch and make a new PR. 

All the conversations and reviews can be viewed on PR #1375.

**Overview of changes:**
Implemented a method to check for all intra links during parsing.

**Anything you'd like to highlight / discuss:**
Code quality and edge cases. 

**Testing instructions:**
`npm run test`

**Proposed commit message: (wrap lines at 72 characters)**
Implement broken intralink validator

Currently, users have to manually check if their intralinks are valid.
Let's create a crawler to validate intralinks automatically. 

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [x] Added tests for bug fixes or features
- [x] Linked all related issues
- [x] No blatantly unrelated changes
- [x] Pinged someone for a review!

PR #2465, opened 2024-03-16
Efficient validation for intra-link with hash 

yiwen101:

- [x] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [x] DevOps
- [ ] Others, please explain:

working on #1418

**Overview of changes:**
Store all the elements with ids (accessible) in the siteLinkManager
Validate intra-link with hash at the link processor;
Fixes 15 detected invalid intra-link with hash in the current documentation.

-------------------------------------------------------------------------------------------------------------------
This is a draft PR; has encountered issue and stuck when working on this issue, so I post my work so far to seek help.

Current implementation: 
1 for nodes with node.attribs.ids, simply add to the collection in the siteLinkManager

2 for header tags, add to the collection in the siteLinkManager after they have been granted ids 
3 for include nodes, after they have been processed, recursively add their and their children ids to the 
    collection in the siteLinkManager; if they/their children are header tags, grant them ids with the same util method as in 2
Screenshot 2024-03-16 at 10 36 58

Current issue:
1 some header added in step 3 seems to be off: 
Screenshot 2024-03-16 at 10 22 30
Screenshot 2024-03-16 at 10 39 07

2 there are still some hashes missing, not collected: 
Screenshot 2024-03-16 at 10 29 50

**Testing instructions:**

Implement efficient validation for hash intra-link

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**


yiwen101:
It turns out that my code is good. All the 15 mismatches are actually legitimate broken links :(

I will fix all the broken links with hash in this PR as well. For ease of verification of reviewer, I will also post pictures of the positions of these broken hash links here:

 
Screenshot 2024-03-19 at 15 40 52
Screenshot 2024-03-19 at 15 44 06
Screenshot 2024-03-19 at 15 37 08
Screenshot 2024-03-19 at 15 35 39
arkBind/markbind/assets/121547057/c14aeb4a-e60e-42ed-98f5-92b96d25ebab">
Screenshot 2024-03-19 at 15 29 48
Screenshot 2024-03-19 at 15 28 08
Screenshot 2024-03-19 at 15 25 46
Screenshot 2024-03-19 at 15 22 36
Screenshot 2024-03-19 at 15 13 51

docs/devGuide/bootcamp/exploreMarkBind.md
```diff
@@ -255,7 +255,7 @@ Now, let's try editing the `index.md` file to add some content of our own!
 1. Modify or add some content utilizing one or more of the following components:
     - <a tags="environment--combined" href="/userGuide/components/presentation.html#boxes">Box</a><a tags="environment--dg" href="https://markbind.org/userGuide/components/presentation.html#boxes">Box</a>
     - <a tags="environment--combined" href="/userGuide/components/presentation.html#panels">Panel</a><a tags="environment--dg" href="https://markbind.org/userGuide/components/presentation.html#panels">Panel</a>
-    - <a tags="environment--combined" href="/userGuide/components/presentation.html#tooltips">Tooltip</a><a tags="environment--dg" href="https://markbind.org/userGuide/components/presentation.html#tooltips">Tooltip</a>
+    - <a tags="environment--combined" href="/userGuide/components/presentation.html#relevant-tips-and-amp-tricks">Tooltip</a><a tags="environment--dg" href="https://markbind.org/userGuide/components/presentation.html#relevant-tips-and-amp-tricks">Tooltip</a>
```
tlylt, line 258:
```suggestion
    - <a tags="environment--combined" href="/userGuide/components/popups.html#tooltips">Tooltip</a><a tags="environment--dg" href="https://markbind.org/userGuide/components/popups.html#tooltips">Tooltip</a>
```

This should be the intended link

tlylt:
> Efficient 

Could you test with the cs2103T website to check the before & after timing of running `markbind-serve/build` over a large website, to validate whether this solution is in practice efficient?

yiwen101:
> > Efficient
> 
> Could you test with the cs2103T website to check the before & after timing of running `markbind-serve/build` over a large website, to validate whether this solution is in practice efficient?

Thanks for the comment, the following if the result on my end: 
valid-hash branch, intrasiteLinkValidation {enabled: true} 
Screenshot 2024-03-20 at 10 49 51

valid-hash branch, intrasiteLinkValidation {enabled: false} 
Screenshot 2024-03-20 at 10 52 29

master-branch,  intrasiteLinkValidation {enabled: false}
Screenshot 2024-03-20 at 10 55 10

The additional cost should be negligible, and literally invisible (as master branch in theory should run faster; but run time fluctuate)

docs/devGuide/bootcamp/exploreMarkBind.md
yiwen101: I see, thanks

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -14,13 +15,22 @@ const tagsToValidate: Set<string> = new Set([
   'script',
 ]);
 
+function isHeaderTag(node: MbNode): boolean {
```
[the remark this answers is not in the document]
yiwen101: Thank you for the suggestion, applied in the latest commit.

tlylt:
> The additional cost should be negligible, and literally invisible (as master branch in theory should run faster when all other factor hold constant, but it turns out run the slowest, suggesting that even the fluctuation in runtime caused by other factors has significantly higher impact than the runtime contributed by the changes)

👍

For your test run on 2103T website, how many valid intra-link hash errors were detected? Would be useful info for @damithc for follow-up broken link fixes.

kaixin-hc (approved):
Thanks for your hard work! I didn't spot anything else while reading through the code. I agree with Elton that including a bit of comment documentation would be good.

One thing I wonder is if it would be possible and desired to add something that tests this to the test site.

packages/core/src/html/headerProcessor.ts
```diff
@@ -12,22 +12,26 @@ const _ = {
 /*
  * h1 - h6
  */
-export function setHeadingId(node: MbNode, config: NodeProcessorConfig) {
+export function setHeadingId(node: MbNode,
+                             config: NodeProcessorConfig,
+                             callFromSiteLinkManager: boolean = false) {
```
yucheng11122017, line 17:
```suggestion
                             callFromSiteLinkManager: boolean = false): string {
```

packages/core/src/html/headerProcessor.ts
```diff
@@ -12,22 +12,26 @@ const _ = {
...
-  } else {
-    headerIdMap[slugifiedHeading] = 2;
+  if (!callFromSiteLinkManager) {
+    if (headerIdMap[slugifiedHeading]) {
+      headerId = `${slugifiedHeading}-${headerIdMap[slugifiedHeading]}`;
+      headerIdMap[slugifiedHeading] += 1;
+    } else {
+      headerIdMap[slugifiedHeading] = 2;
+    }
+    node.attribs.id = headerId;
```
yucheng11122017, line 32: Why is the headerId not set when called by SiteLinkManager?

packages/core/src/html/headerProcessor.ts
```diff
@@ -12,22 +12,26 @@ const _ = {
 /*
  * h1 - h6
  */
-export function setHeadingId(node: MbNode, config: NodeProcessorConfig) {
+export function setHeadingId(node: MbNode,
```
yucheng11122017, line 15: Hmmm this name is rather misleading because the headingId is only set IF not called from siteLinkManager. And it also returns the value - doesn't just set

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,48 @@ export class SiteLinkManager {
...
+    this.filePathToHashesMap.forEach((hashes, filePath) => {
+      result += `${filePath}:\n`;
+      hashes.forEach((hash) => {
+        result += `  ${hash}\n`;
+      });
+    });
+    return result;
+  }
+
+  maintainInclude(node: MbNode, cwf: string) {
```
yucheng11122017, line 104:
```suggestion
  maintainHashesForInclude(node: MbNode, cwf: string) {
```

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,48 @@ export class SiteLinkManager {
...
+    if (!this.filePathToHashesMap.has(path)) {
+      this.filePathToHashesMap.set(path, new Set());
+    }
+    if (forceWrite !== '') {
+      this.filePathToHashesMap.get(path)!.add(forceWrite);
+    } else if (node.attribs!.id) {
+      this.filePathToHashesMap.get(path)!.add(node.attribs!.id);
+    }
+  }
+
```
yucheng11122017, line 92: Hmmm I see that this is mostly used for testing purposes. Should this be moved into test file instead?

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,48 @@ export class SiteLinkManager {
     this._addToCollection(resourcePath, cwf);
     return 'Intralink collected to be validated later';
   }
+
+  maintainFilePathToHashesMap(node: MbNode, cwf: string, forceWrite: string = '') {
```
yucheng11122017, line 78: Could you add a comment explaining what is forceWrite and when should it be used?

yucheng11122017 (commented):
Hi @yiwen101 thanks for the work! I think this is a bit tricky so please try to explain the code a little. I added some questions and also some nits.

packages/core/src/html/headerProcessor.ts
yiwen101: This is to prevent regression.

Screenshot 2024-03-25 at 01 02 07
Screenshot 2024-03-25 at 01 01 56

The validation is only for check purpose, and should not modify the current state;

packages/core/src/html/SiteLinkManager.ts
yiwen101: I want argue for the current code here;

To test the internal state of the siteLinkManager, we need to either 1 write getter for the filePathToHashesMap, 2 write comparator inside the SiteLinkManager 3 add print method.  I think 3, the current way is the least intrusive (and most helpful when I do debugging with console.log)

yiwen101:
Sorry for getting back to the review late; was overwhelmed by a hackathon due Friday this week

@kaixin-hc 
Thank you for the careful review and suggestions. Could you help elaborate further on how could I add this to test site?
Current implementation of the functional test seems only adequate of recursively comparing the files in the expected folder and the actual files during the build process. I do not know how to "expect" error logs. The [PR](https://github.com/MarkBind/markbind/pull/1382/files) that brings in the intra-link validation feature also did not add functional test. 

@EltonGohJH @yucheng11122017 
Thank you for the careful review; I have made the requested changes and added method document.

The only exception is the "print" method in SitelinkManager. Although it is most for test purpose, I believe that it is a necessary evils and the best resort among all choices, so should leave it as it is.

packages/core/src/html/headerProcessor.ts
yiwen101: Renamed to "processHeadingId" and improved method documentation.

packages/core/src/html/SiteLinkManager.ts
yiwen101: Thank you for the suggested name. Have commited it, and the suggested name is indeed much clearer

packages/core/src/html/SiteLinkManager.ts
yiwen101: Method document added in later commit.

packages/core/src/html/headerProcessor.ts
yiwen101: Regarding the aspect not related to internal state: 

When the "maintainHashesForInclude" is called, the heading names is already processed, and we do not know how many same headings has been processed since then, so we can not just anyhow append the name with the number of times it has occured (or minus something).

The appending with the number of times something occur is because id must by unique. This happens internally and the user does not know, so we should not be expecting user to include the -${counter} in the first place.

A not so important but still valid reason is that, the existence of ${name}-{counter} necessitates the existence of ${name}, so no matter what the current state is, it is never wrong to return the name without counter.

packages/core/src/html/SiteLinkManager.ts
EltonGohJH: @yiwen101 

I think I am okay with this method if you really cannot find a better way but you may need to rename it since it does not really print but return the string of the hashmap.

I think it is possible to create a class that extends from SiteLinkManager in the test file right that has the getter? Do you think this will work and fix the issue?

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,57 @@ export class SiteLinkManager {
     this._addToCollection(resourcePath, cwf);
     return 'Intralink collected to be validated later';
   }
+
+  /**
+   * Add sections that could be reached by intra-link with hash to this node to FilePathToHashesMap,
+   * The reachable sections include nodes with ids and headings.
+   *
+   * ForceWrite should only be called when processing heading node with the maintainHashesForInclude method.
```
yucheng11122017, line 82:
```suggestion
   * forceWrite should only be used when processing heading node.
```
Could you explain why?

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,57 @@ export class SiteLinkManager {
     this._addToCollection(resourcePath, cwf);
     return 'Intralink collected to be validated later';
   }
+
+  /**
+   * Add sections that could be reached by intra-link with hash to this node to FilePathToHashesMap,
```
yucheng11122017, line 79:
```suggestion
   * Add sections that could be reached by intra-link with hash to this node to filePathToHashesMap,
```

packages/core/src/html/SiteLinkManager.ts
yucheng11122017: I think also the name `forceWrite` is not accurate. This also seems like a boolean instead of an override

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,57 @@ export class SiteLinkManager {
...
+      this.filePathToHashesMap.set(path, new Set());
+    }
+    if (forceWrite !== '') {
+      this.filePathToHashesMap.get(path)!.add(forceWrite);
+    } else if (node.attribs!.id) {
+      this.filePathToHashesMap.get(path)!.add(node.attribs!.id);
+    }
+  }
+
+  printFilePathToHashesMap() {
```
yucheng11122017, line 99: Reminder to rename this based on @EltonGohJH's comment!

packages/core/src/html/SiteLinkManager.ts
```diff
@@ -69,4 +74,57 @@ export class SiteLinkManager {
...
+      result += `${filePath}:\n`;
+      hashes.forEach((hash) => {
+        result += `  ${hash}\n`;
+      });
+    });
+    return result;
+  }
+
+  /**
+   * Recursively add reachable sections of the included node to the FilePathToHashesMap for validation later,
```
yucheng11122017, line 111:
```suggestion
   * Recursively add reachable sections of the included node to the filePathToHashesMap for validation.
```

packages/core/src/html/headerProcessor.ts
```diff
@@ -9,25 +9,31 @@ const _ = {
   has,
 };
 
-/*
- * h1 - h6
+/**
+ * Increment the counter for same heading text, and return the id of the heading
+ * if is called from SiteLinkManager, it will directly return the id without adding counter
```
yucheng11122017, line 14:
```suggestion
 * If called from SiteLinkManager, it will directly return the id without adding counter
```

packages/core/src/html/headerProcessor.ts
```diff
@@ -9,25 +9,31 @@ const _ = {
...
+ * heading text refers to the h1 - h6 text content
  */
-export function setHeadingId(node: MbNode, config: NodeProcessorConfig) {
+export function processHeadingId(node: MbNode,
```
yucheng11122017, line 17:
```suggestion
export function processAndReturnHeadingId(node: MbNode,
```
Ok not sure about this name but

yucheng11122017 (commented):
Hi @yiwen101 thanks for the work and sorry for the late review! I think the implementation and so on is good: just some comments regarding the naming.

packages/core/src/html/SiteLinkManager.ts
yiwen101: @EltonGohJH 
Thank you so much for your reply, I was not aware that I can do testing like this in typescript before! This is valuable learning point for me;

I have deleted this method in the latest commit, and created a class extending the siteLinkManager as instructed by you. It works well!

packages/core/src/html/SiteLinkManager.ts
yiwen101: Method deleted.

packages/core/src/html/SiteLinkManager.ts
yiwen101: Hi, I have tried improving the code quality in a later commit; this parameter no longer exists.

yiwen101:
@yucheng11122017 
Thank you for your careful review and various suggestions on improving code quality.

In a later commit, I made following changes to existing methods to improve code quality. They are:
1 rename "processAndReturnHeadingId" back to "setHeadingId", and modify its behaviour accordingly.
2 remove the "forceWrite" from `maintainFilePathToHashesMap`
3 modify the maintainHashesForInclude accordingly.

I hope that the code becomes clearer after the changes.

packages/core/test/unit/utils/utils.ts
```diff
@@ -38,7 +38,25 @@ export function getNewPluginManager(plugins: string[], pluginsContext: PluginCon
   return new PluginManager(fileConfig, plugins, pluginsContext);
 }
 
-export function getNewSiteLinkManager() {
+class SiteLinkManagerMock extends SiteLinkManager {
```
EltonGohJH, line 41: This is not really a Mock class by the definition of Mock right?

EltonGohJH (approved):
LGTM. Other than the naming of the class for testing

packages/core/test/unit/utils/utils.ts
yiwen101: Emm, I see; thanks for pointing this out;
I have come up with a list of potential names for it, but not sure whether "EnhancedSiteLinkManager" is the best choice. "TestableSiteLinkManager", "UtilitySiteLinkManager", "TestHelperSiteLinkManager" and  "AccessibleSiteLinkManager" are some other choices.  What is your idea?

EltonGohJH: TestableSiteLinkManager sounds good to me!

yiwen101: Done in latest commit.

PR #2511, opened 2024-04-10
Implement method to process attributes that can be overridden by slots

### LINKED ISSUE #2476: Add logger warnings when slots override attributes in NodeProcessor.ts

### Any related issues?

#2463, #1715

Testing

There are two issues that arise from #2463 (and I'll be quoting myself):

_Inconsistent naming in `NodeProcessor.data.ts`_

The naming conventions of the test inputs in `NodeProcessor.data.ts` are inconsistent. Many of the elements there have a test to check if the slots do override the attributes of the elements, but they are named in two separate ways:
1. `PROCESS_[ELEMENT]_ATTRIBUTES_NO_OVERRIDE`
2. `PROCESS_[ELEMENT]_[ATTRIBUTE]_SLOT_TAKES_PRIORITY`

I think that's why the author of #2053 wrote another constant called `PROCESS_PANEL_HEADER_SLOT_TAKES_PRIORITY` when the constant `PROCESS_PANEL_ATTRIBUTES_NO_OVERRIDE` tests for the exact same thing. That's why I have an `expect` statement after each of these two tests, because in both times the logger is called with the same input. 

_Logger warnings still inconsistent_

With the merging of #2053, Panels, Dropdowns, Modals and Popovers print warnings to the console if there are slots overriding the element attributes. However, other elements such as Quizzes and QOptions do not print warnings when this happens. This is because in the code of `MdAttributeRenderer.ts`, not every attribute is checked to see if there is an overriding slot, so the logger will not warn if the unchecked attributes are overridden. 

Firstly, standardise the constant names in `NodeProcessor.data.ts` to `PROCESS_[ELEMENT]_[ATTRIBUTE]_SLOT_TAKES_PRIORITY` as it is clearer than the other alternative. 

To fix the inconsistent logger warnings once and for all, implement this new method
```javascript
processAttribute(node, attribute, isInline, slotName = attribute) {
    if (!this.hasSlotOverridingAttribute(node, attribute, slotName) {
        this.processAttributeWithoutOverride(node, attribute, isInline, slotName);
    }
}
```
and use it for every attribute of the node. 

Lastly, create unit tests so for every attribute that can potentially be overriden. 

This issue will require at least 3 PRs, namely:

- [x] #2483
- [x] #2511 
- [ ] #2525
- [ ] #2526

_No response_

### Additional context

_No response_

luminousleek:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Partially resolves #2476 

**Overview of changes:**

Implement `processSlotAttribute` method and use it for attributes that can be overridden by slots.

For reference, these are the attributes that can be overridden by slots:
| Component         | Attributes           |
| ----------------- | -------------------- |
| Panel             | Header, Alt               |
| Box               | Icon, Header         |
| Popover           | Header, Content      |
| Modal             | Header               |
| Dropdown          | Header               |
| Scroll-top-button | Icon                 |
| Question          | Header, Hint, Answer |
| Q-Option          | Reason               |
| Quiz              | Intro                |
| Tooltip          | Content          |
| Tab              | Header            |
| A-Point        | Header, Content, Label |

The docs are not correct regarding this.

**Testing instructions:**

```
Implement processSlotAttribute method

There is some duplicated code to process attributes that can
be overridden by slots. In addition, not every attribute that can be
overridden with slots are processed with the 
hasSlotOverridingAttribute method. As such, there will be no
warning in the logger when these attributes are overridden.

Let's put this duplicated code into a method and call it for every
attribute that can be overridden by a slot.
```

---

- [ ] Added tests for bug fixes or features (will do in the next PR!)
- [x] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core/src/html/MdAttributeRenderer.ts
+++ b/packages/core/src/html/MdAttributeRenderer.ts
@@ -121,7 +121,7 @@ export class MdAttributeRenderer {
    */
 
   processPanelAttributes(node: MbNode) {
-    this.processAttributeWithoutOverride(node, 'alt', false, '_alt');
+    this.processSlotAttribute(node, 'alt', false, '_alt');
     this.processSlotAttribute(node, 'header', false);
   }
 
--- a/packages/core/src/html/MdAttributeRenderer.ts
+++ b/packages/core/src/html/MdAttributeRenderer.ts
@@ -109,7 +109,7 @@ export class MdAttributeRenderer {
   }
 
   processTooltip(node: MbNode) {
-    this.processAttributeWithoutOverride(node, 'content', true);
+    this.processSlotAttribute(node, 'content', true);
   }
 
   processModalAttributes(node: MbNode) {
@@ -148,7 +148,7 @@ export class MdAttributeRenderer {
    */
 
   processTabAttributes(node: MbNode) {
-    this.processAttributeWithoutOverride(node, 'header', true);
+    this.processSlotAttribute(node, 'header', true);
   }
 
   /*
@@ -197,8 +197,8 @@ export class MdAttributeRenderer {
   }
 
   processAnnotationPointAttributes(node: MbNode) {
-    this.processAttributeWithoutOverride(node, 'content', false);
-    this.processAttributeWithoutOverride(node, 'header', false);
-    this.processAttributeWithoutOverride(node, 'label', false);
+    this.processSlotAttribute(node, 'content', false);
+    this.processSlotAttribute(node, 'header', false);
+    this.processSlotAttribute(node, 'label', false);
   }
 }
--- a/packages/core/src/html/MdAttributeRenderer.ts
+++ b/packages/core/src/html/MdAttributeRenderer.ts
@@ -75,10 +75,23 @@ export class MdAttributeRenderer {
     return hasAttribute;
   }
 
-  processPopoverAttributes(node: MbNode) {
-    if (!this.hasSlotOverridingAttribute(node, 'header')) {
-      this.processAttributeWithoutOverride(node, 'header', true);
+  /**
+   * Checks if there is a pre-existing slot for the attribute.
+   * If there is a slot, it deletes the attribute and logs a warning.
+   * If there is no slot, it processes the markdown attribute. 
+   * @param node Element to process
+   * @param attribute Attribute name to process
+   * @param isInline Whether to process the attribute with only inline markdown-it rules
+   * @param slotName Name attribute of the <slot> element to insert, which defaults to the attribute name
+   */
+  processSlotAttribute(node: MbNode, attribute: string, isInline: boolean, slotName = attribute) {
+    if (!this.hasSlotOverridingAttribute(node, attribute, slotName)) {
+      this.processAttributeWithoutOverride(node, attribute, isInline, slotName);
     }
+  }
+
+  processPopoverAttributes(node: MbNode) {
+    this.processSlotAttribute(node, 'header', true);
 
     // Warn if there is a content slot overriding the attributes 'content' or 'src'
     const hasSlotAndContentAttribute = this.hasSlotOverridingAttribute(node, 'content', 'content');
@@ -100,9 +113,7 @@ export class MdAttributeRenderer {
   }
 
   processModalAttributes(node: MbNode) {
-    if (!this.hasSlotOverridingAttribute(node, 'header')) {
-      this.processAttributeWithoutOverride(node, 'header', true);
-    }
+    this.processSlotAttribute(node, 'header', true);
   }
 
   /*
@@ -111,9 +122,7 @@ export class MdAttributeRenderer {
 
   processPanelAttributes(node: MbNode) {
     this.processAttributeWithoutOverride(node, 'alt', false, '_alt');
-    if (!this.hasSlotOverridingAttribute(node, 'header')) {
-      this.processAttributeWithoutOverride(node, 'header', false);
-    }
+    this.processSlotAttribute(node, 'header', false);
   }
 
   /*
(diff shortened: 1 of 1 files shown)


packages/core/src/html/MdAttributeRenderer.ts
```diff
@@ -75,10 +75,23 @@ export class MdAttributeRenderer {
...
+    if (!this.hasSlotOverridingAttribute(node, attribute, slotName)) {
+      this.processAttributeWithoutOverride(node, attribute, isInline, slotName);
     }
+  }
+
+  processPopoverAttributes(node: MbNode) {
+    this.processSlotAttribute(node, 'header', true);
 
     // Warn if there is a content slot overriding the attributes 'content' or 'src'
     const hasSlotAndContentAttribute = this.hasSlotOverridingAttribute(node, 'content', 'content');
```
luminousleek, line 97: This bit was left unchanged because the logic for the `content` attribute is slightly different. According to the [docs](https://markbind.org/userGuide/components/popups.html#popovers), the `src`and `content` attributes and the `content` slot can be used for popovers. If all three are present then MarkBind will prioritise them in the following order:
1. `content` slot
2. `content` attribute
3. `src` attribute

So the logic to process this attribute (and especially the logger warning) is different than the other attributes. Specifically, it needs to check for the case that the `src` and `content` attributes are both used, and show a warning message if that is the case. The `processSlotAttribute` method does not account for this case, so this chunk of code is still here.

packages/core/src/html/MdAttributeRenderer.ts
```diff
@@ -75,10 +75,23 @@ export class MdAttributeRenderer {
...
+  /**
+   * Checks if there is a pre-existing slot for the attribute.
+   * If there is a slot, it deletes the attribute and logs a warning.
+   * If there is no slot, it processes the markdown attribute.
+   * @param node Element to process
+   * @param attribute Attribute name to process
+   * @param isInline Whether to process the attribute with only inline markdown-it rules
+   * @param slotName Name attribute of the <slot> element to insert, which defaults to the attribute name
+   */
+  processSlotAttribute(node: MbNode, attribute: string, isInline: boolean, slotName = attribute) {
```
luminousleek, line 87: I'm not entirely happy with the name of the method but it's the best I can come up with. Some other alternatives I've considered are:
* `processSlottableAttribute`
* `processAttributeCanBeOverridden`
* `processOverridableAttribute`
* `processAttributeWithSlot`

Let me know your thoughts thanks.

luminousleek:
The `check_docs` test is failing because the PR is doing its job - it's giving logger warnings when attributes for a quiz are being overridden:
Screenshot 2024-04-10 at 4 21 56 PM

The warnings arise from the panel in the Header and Hint example with **slots** at [this page](https://markbind.org/userGuide/components/others.html#questions-and-quizzes) (the relevant documentation file is `/docs/userGuide/syntax/questions.md`), where the code for the question in the panel is
```html
<question type="checkbox" header="Which of the following is true?" hint="Think out of the box! :fas-box:">
  <!-- Header slot -->
  <div slot="header">
<p>Which of the following is correct?</p>
<p>Challenge: Try to get all the answers on your first try! ⭐️ ⭐️</p>
  </div>
<!-- other code omitted for brevity -->
  <!-- Hint slot -->
  <div slot="hint">
<p>Think out of the box! <span aria-hidden="true" class="fas fa-box"></span></p>
<p>Need another hint? <tooltip content="Two of the answers are correct!">Hover over me!</tooltip> <span aria-hidden="true" class="fas fa-mouse-pointer"></span></p>
  </div>
</question>
```

As can be seen, the `header` and `hint` attributes are overridden by their respective slots, which then triggers a logger warning. A simple fix would be to delete the `header` and `hint` attributes, so I'm wondering if it's OK to do this in this PR or if I need to create a separate PR for this.

packages/core/src/html/MdAttributeRenderer.ts
yucheng11122017: I think processSlotAttribute is ok!

packages/core/src/html/MdAttributeRenderer.ts
```diff
@@ -186,7 +193,7 @@ export class MdAttributeRenderer {
   }
 
   processScrollTopButtonAttributes(node: MbNode) {
-    this.processAttributeWithoutOverride(node, 'icon', true);
+    this.processSlotAttribute(node, 'icon', true);
   }
 
   processAnnotationPointAttributes(node: MbNode) {
```
yucheng11122017, line 199: Should we use this method here as well?

yucheng11122017 (commented):
Thank you for the detailed PR @luminousleek! 

I realized that our documentation might not be fully up to date - other attributes seem to work with a slot as well (like a-point) but it doesn't seem to be present in our documentation.
I think we might need to double check this behavior and update the documentation

packages/core/src/html/MdAttributeRenderer.ts
luminousleek: Yes

luminousleek:
I tested the rest of the other attributes for the components to see whether they're able to be overridden by slots.

Looks like the `header`, `content` and `label` attributes for `a-point` are able to be overridden by slots
Screenshot 2024-04-12 at 1 02 29 PM

Also, the `header` attribute for `tab` is also able to be overridden by a slot
Screenshot 2024-04-12 at 1 26 55 PM

As well as the `content` attribute for `tooltip`
Screenshot 2024-04-12 at 1 34 01 PM

And the `alt` attribute for `panel`
Screenshot 2024-04-12 at 1 39 29 PM

Basically that's all the attributes that are handled by `MdAttributeRenderer.ts`. The PR is updated to use `processSlotAttributes` for the rest of the attributes.

kaixin-hc (approved):
Great catch. I think we can leave the documentation for another PR or concurrent with adding tests for the ones that aren't tested yet.

I like how this makes the code overall clearer by using a simpler method for everything, thanks for your work!

Issue #2140, opened 2023-02-07
Utilize GitHub Action to aid PR review


tlylt:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### What is the area that this feature belongs to?

Other

### Is your feature request related to a problem? Please describe.

Use GitHub Action, possibly available ones, to automate and remind PR authors of the potential flaws:
- [x] missing proposed commit message
- [x] broken link
- [ ] the need to update test/docs for certain changes (could be harder to achieve)
  - For example, if we know that a piece of code is coupled to some documentation, can we detect that both the code and the documentation are updated?
- etc

Any other helpful points pls feel free to edit or comment below.

### Describe the solution you'd like

As above

tlylt:
utilizing our existing intra-site broken link detection could be a good first step:
- [ ] a step in the CI that serve/builds the docs, check that the console does not contain broken link warning logs.

tlylt:
To add some background info: I was primarily looking to use [danger.js](https://github.com/danger/danger-js) for this issue. However, even though the tool itself is reasonably easy to integrate, it is not straightforward to use with a forking workflow because of permission issues (i.e, when someone is making a PR from a fork, the triggered GitHub Action has limited permission to prevent leaking secrets etc).
- relevant [link](https://github.com/danger/danger-js/issues/1060) if danger.js is to be used
- a [presentation](https://github.com/nus-cs3281/2023/issues/34) I made attempting to integrate dangerjs in markbind, in case it's helpeful

yucheng11122017:
Another thing that I would be interested to do is to check if the PR is tagged with the release tags upon merging! I think this is something we always forget to do (opps) but helps make releases a lot easier.

yucheng11122017:
> Another thing that I would be interested to do is to check if the PR is tagged with the release tags upon merging! I think this is something we always forget to do (opps) but helps make releases a lot easier.
> 
> I was thinking perhaps it can ping the person who merged if they forget to tag it.

@KevinEyo1 will be taking up this issue

KevinEyo1:
Hi, I'm working on the checking of the coupled code portion now. I have thought about how to implement this, some considerations up for discussion. (This also applies to if Danger.js is used, since it also requires write access and also requires checking out of code)

Current implementation idea (or use Danger.js)
A file mapping list of files and their coupled test/docs file, which I think will have to be manually updated to provide better accuracy
A script to handle the logic (can also be part of workflow)
[...]

PR #2523, opened 2024-04-22
Using DangerJS to check changes coupling of implementation files to test or documentation files

KevinEyo1:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [X] DevOps
- [ ] Others, please explain:

**Overview of changes:**
#2140
Added a workflow that utilizes DangerJS to check that certain implementation files having changes, has changes in its corresponding test or documentation files.

Currently run on `pull_request` and logs the error in CI logs. Does not include implementation of manual/automatic coupling of files. Once that is implemented with good accuracy, workflow can change to failing the check instead of just logging the error.

**Testing instructions:**
Based on test couplings, proper error logs should be found in CI logs

It can be easy to miss out on updating test or documentation files
when implementing or changing features.

Let's add a workflow utilizing DangerJS to aid in checking and 
alerting developers if they forget to do so.

---

- [X] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/.github/workflows/check-file-diff.yml
+++ b/.github/workflows/check-file-diff.yml
@@ -1,0 +1,23 @@
+name: "Check File Differences"
+on:
+    pull_request:
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+  check-couplings:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+      - run: npm run setup
+      - name: Run Danger
+        run: npx danger ci
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -11,12 +11,6 @@ const couplings: Couplings = {
     'packages/cli/test/functional/test_site/testList.md',
   ],
   'package.json': ['package-lock.json'],
-  'packages/core/template/project/developerGuide/Configuration.md': [
-    'packages/cli/test/functional/test_site_templates/test_project/'
-    + 'expected/developerGuide/Configuration.html',
-    'packages/cli/test/functional/test_site_templates/test_project/'
-    + 'expected/developerGuide/Configuration.page-vue-render.js',
-  ],
 };
 
 const { git } = danger;
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -1,0 +1,40 @@
+import { danger } from 'danger';
+import * as logger from '@markbind/core/src/utils/logger';
+
+interface Couplings {
+  [key: string]: string[];
+}
+
+const couplings: Couplings = {
+  '.github/workflows/check-file-diff.yml': [
+    'package.json',
+    'babel.config.js',
+  ],
+  'babel.config.js': [
+    'package.json',
+    'lerna.json',
+  ],
+};
+
+const { git } = danger;
+
+Promise.resolve().then(() => {
+  const allModifiedFiles = [...git.modified_files, ...git.created_files];
+  const messages: string[] = [];
+
+  Object.entries(couplings).forEach(([implementationFile, dependentFiles]) => {
+    dependentFiles.forEach((dependentFile) => {
+      if (allModifiedFiles.includes(implementationFile) && !allModifiedFiles.includes(dependentFile)) {
+        messages.push(`Changes to ${implementationFile} should include changes to ${dependentFile}`);
+      }
+    });
+  });
+
+  if (messages.length > 0) {
+    logger.info(`Detected issues with file couplings:\n${messages.join('\n')}`);
+    logger.info('Please ensure implementation changes are accompanied '
+     + 'by corresponding test or documentation updates.');
+  } else {
+    logger.info('All file couplings are correctly updated.');
+  }
+}).catch((err: Error) => logger.error(err));
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -36,8 +36,8 @@ Promise.resolve().then(() => {
   });
 
   if (messages.length > 0) {
-    logger.info(`Detected issues with file couplings:\n${messages.join('\n')}`);
-    logger.info('Please ensure implementation changes are accompanied '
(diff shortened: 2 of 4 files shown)


Dangerfile.js
yucheng11122017, line 1: Can we move this into the scripts folder instead!

Dangerfile.js
```diff
@@ -0,0 +1,46 @@
...
+    fs.appendFileSync(filePath, `RESULT<<EOF\n${messageString}\nEOF\n`);
+  } else {
+    fs.appendFileSync(filePath, 'RESULT<<EOF\nAll file couplings are correctly updated.\nEOF\n');
+  }
+}).catch((error) => {
+  const errorPath = path.join(process.env.GITHUB_ENV || '.', 'action_output');
+  if (error.message) {
+    fs.appendFileSync(errorPath, `ERROR_LOGS<<EOF\n${error.message}\nEOF\n`);
+  } else {
+    fs.appendFileSync(errorPath, 'ERROR_LOGS<<EOF\nNo explicit errors caught.\nEOF\n');
```
yucheng11122017, line 44: Is this actually true? It looks like there are errors but just no error messages right

Dangerfile.js
```diff
@@ -0,0 +1,46 @@
...
+      if (allModifiedFiles.includes(implFile) && !allModifiedFiles.includes(testOrDocFile)) {
+        messages.push(`Changes to ${implFile} should include changes to ${testOrDocFile}`);
+      }
+    });
+  });
+
+  const filePath = path.join(process.env.GITHUB_ENV || '.', 'action_output');
+  if (messages.length > 0) {
+    const messageString = messages.join('\n');
+    fs.appendFileSync(filePath, `RESULT<<EOF\n${messageString}\nEOF\n`);
```
yucheng11122017, line 35: hmmm should this be appending? or should a new file be created at the start? Will the old error logs also be included?

.github/workflows/check-file-diff.yml
```diff
@@ -0,0 +1,31 @@
...
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Print Output
+        run: |
+          echo "Results:"
+          echo "$RESULT"
+          echo "Errors:"
+          echo "$ERROR_LOGS"
+        env:
+          RESULT: ${{ env.RESULT }}
+          ERROR_LOGS: ${{env.ERROR_LOGS}}
```
yucheng11122017, line 31: Should we add permissions for this workflow?

.github/workflows/check-file-diff.yml
```diff
@@ -0,0 +1,31 @@
...
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+  check-couplings:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
yucheng11122017, line 15: Can we use v3 first to be consistent with other workflows?

yucheng11122017 (commented):
Some initial comments!

package.json
```diff
@@ -40,7 +40,9 @@
     "walk-sync": "^2.0.2"
   },
   "dependencies": {
+    "@markbind/core": "^5.4.0",
```
yucheng11122017, line 43: ? why is there this change?

Dangerfile.js
yucheng11122017, line 1: Can we change this to TS!

yucheng11122017 (commented):
Some comments!

package.json
KevinEyo1: Hi, this is due to importing logger as per the eslint

Dangerfile.js
KevinEyo1: Good point, I will work on it.

Dangerfile.js
```diff
@@ -0,0 +1,37 @@
...
+  ],
+};
+
+const { git } = danger;
+
+Promise.resolve().then(() => {
+  const allModifiedFiles = [...git.modified_files, ...git.created_files];
+  const messages = [];
+
+  Object.entries(couplings).forEach(([implFile, testOrDocFiles]) => {
```
kaixin-hc, line 22: This makes sense to me in theory, but given that the sample couplings you have here are package.json and either lerna or babel, which is not a TestOrDoc file, I think reading this code is confusing. Maybe we should name it as "dependent files" or something?

Or maybe this isn't such a good example of a coupling? You could put site and site test maybe 🤔

dangerfile.ts
```diff
@@ -0,0 +1,40 @@
...
+const couplings: Couplings = {
+  '.github/workflows/check-file-diff.yml': [
+    'package.json',
+    'babel.config.js',
+  ],
+  'babel.config.js': [
+    'package.json',
+    'lerna.json',
+  ],
+};
```
yucheng11122017, line 17: Could we add more legitimate examples here!

yucheng11122017 (commented):
Last nit and I think we should be good to go!

dangerfile.ts
KevinEyo1: Alright, then I will remove the testing couplings completely.

kaixin-hc (approved):
changes look good - added a nit about how it should be logged.

i am concerned about people not checking console logs for CI, its a lot more troublesome and honestly wouldn't occur to me. But since this is a piece by piece implementation and you've mentioned you'd like to look at generating an automated list of reliant files, I think it is OK for now

EDIT: Also tests are failing - do fix thanks

dangerfile.ts
```diff
@@ -0,0 +1,45 @@
...
+
+interface Couplings {
+  [key: string]: string[];
+}
+
+const couplings: Couplings = {
+  'packages/core/src/html/CustomListIconProcessor.ts': [
+    'docs/userGuide/syntax/lists.md',
+    'packages/cli/test/functional/test_site/testList.md',
+    'packages/cli/test/functional/test_site/expected/testList.html',
```
yucheng11122017, line 12: Hmmm tbh i feel like there isn't a need to add the .html and .page-vue-render.js files here. Because they will have to updated otherwise the tests will fail automatically.

yucheng11122017 (approved):
LGTM! Thanks for the work

package.json
tlylt: Why not just use console.log ? There's no need to have this dependency introduced for the sake of a log to the console

PR #2557, opened 2024-06-14
Fix Tab-Group Header not displayed 

### LINKED ISSUE #2524: Tab-Group Header not displayed

### Any related issues?

_No response_

macOS Ventura

### MarkBind version

5.4.0

In [this part](https://markbind.org/userGuide/components/presentation.html#tabs) of User Guide, there is a code snippet that shows how to specify a header for a Tab-group. However, the header is not there, as seen in the screenshot below.

Screenshot 2024-04-23 at 11 26 24 AM

### Expected behavior

The tab group should have a header called "Third tab group 🌌 " for the first tab group and "Disabled fourth tab group" for the second tab group

### Anything else?

I reverted the codebase to v3.0.0 to test it myself and it still doesn't work in that version :( So I don't think it's caused by the Bootstrap migration from v4 to v5.

gerteck:

- [ ] Documentation update
- [x] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Fixes #2524

**Overview of changes:**

Header does not show up for the `TabGroup` component. This was likely due to the header being passed as a prop instead of a named slot into `Dropdown`. ~To fix this, pass header as named slot.~ To fix this, the header is now treated as a prop.

From what I understand:
* `TabGroup` internally consists of the dropdown tab component. However, internally, TabGroup does not render this dropdown, but instead adds itself to the parent `TabSet`, which is then in charge of rendering children `Tab` or `TabGroup`. 
* In `TabSet`, the header was originally passed into Dropdown from Tabset as an attribute / vue prop of `Dropdown`. `Dropdown` then uses this prop to display the header:
  * [commit history](https://github.com/MarkBind/vue-strap/commit/96adcfb15cc90d7ccba796ab5771267eccfdef9a?diff=split&w=0)
* After that, to add markdown support to the header, the header was passed using slots. I would like to see if this is the source of the bug, but at this point, I can't seem to test this version of MarkBind, due to depreciated dependencies/errors faced.
  * [commit history](https://github.com/MarkBind/vue-strap/commit/04dd7e210f6cabea02c0e6e766869a1b1ea155ab) 

**[ Requesting Comments ]**: I don't quite understand how the Vue component's prop is being passed as a named slot. Referring to the [developer guide](https://markbind.org/devdocs/devGuide/development/writingComponents.html), I tried to trace the processing but I think I might be facing a skill issue. I'm confused as it says that "MarkBind attributes are passed to the Vue component as props. The type of the prop will be a String." However, the prop is able to be processed as a named slot.

This seems to be the core of the issue behind the bug. In `Tabset`, the header is being passed into `dropdown` similarly as a prop, which is not registered as a namedSlot, and ends up as an attribute when rendered. ~To fix this, I directly passed the header as a named slot. However the discrepancy and origin of the bug leaves me uneasy.~

**Illustration**

Sample Code in MarkBind Syntax (where header is passed as an attribute):
```
<!-- header is being passed as a prop -->
<dropdown header="Prop Header" class="w-100"> 
    <ul slot="dropdown-menu" class="dropdown-menu">
      <li><a href="#dropdown" class="dropdown-item">Action</a></li>
      <li><a href="#dropdown" class="dropdown-item">Separated link</a></li>
    </ul>
  </dropdown>
```

Dropdown Vue Component Template (where header is accepted as a named slot):

https://github.com/MarkBind/markbind/blob/2c8276468592f964663ff3193e57345b63fa7f70/packages/vue-components/src/Dropdown.vue#L28-L51

Output:

While the header was passed in as a prop, it is parsed as a named slot.

image

Some references:
* [Legacy VueStrap Dropdown](https://github.com/yuche/vue-strap/blob/master/src/Dropdown.vue)
* [Add Embedded HTML in dropdown Commit](https://github.com/MarkBind/vue-strap/commit/e9c4ec9b8966ba1ea7032b9294bb68651527e804)
* [Update of header slots for Vue Components Commit](https://github.com/MarkBind/markbind/commit/640d18be106b5ca8b0b6ab0a60a21b3d19139b67)

**Testing instructions:**

Refer to user guide: https://markbind.org/userGuide/components/presentation.html#tabs
Refer to PR's localhost user guide: http://127.0.0.1:8080/userGuide/components/presentation.html#tabs

Before:
image

After:
image

Fix TabGroup Header Rendering Issue

The header for the TabGroup component was not displaying 
correctly, due to the header being passed as a prop but 
treated as a named slot in the Dropdown component. 
To fix this, the header is now treated as a prop.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**


lhw-1:
> I don't quite understand how the Vue component's prop is being passed as a named slot. Referring to the [developer guide](https://markbind.org/devdocs/devGuide/development/writingComponents.html), I tried to trace the processing but I think I might be facing a skill issue. I'm confused as it says that "MarkBind attributes are passed to the Vue component as props. The type of the prop will be a String." However, the prop is able to be processed as a named slot.

The [`createSlotTemplateNode`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/elements.ts#L11-L13) method is used to convert slots into template nodes directly; this is called by the [`processAttributeWithoutOverride`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.ts#L30-L50) method, which is in turned called for specified named slots that we are looking out for. This is how we process named slots - hopefully this addresses your question (and if it doesn't, do let me know).

Regardless, this issue might need further investigation 😓

packages/vue-components/src/Tabset.vue
```diff
@@ -24,10 +24,12 @@
           v-else
           :key="index"
           class="nav-item"
-          :header="t.headerRendered"
           :class="{active:t.active}"
           :disabled="t.disabled"
         >
+          <template #header>
+            <span v-html="t.headerRendered"></span>
+          </template>
```
lhw-1, line 32: Instead of making this change (since we rely on the conversions from slots to templates, and do not use header templates directly in our `.vue` files), a suggestion I have is to treat the `header` as a prop instead?

In `Dropdown.vue`, we can first include the `header` as a prop:

```
props: {
  ...
  header: {
    type: String,
    default: '',
  },
  ...
```

And then we can replace the `<slot name="header"></slot>` under the `v-if`, with `{{ header }}`. We could also rename `header` to something like `tabgroup-header` to differentiate it as well.

This might be a less hacky solution, but if we can investigate the issue deeper and find a better solution, that would be best 👍

gerteck:
>
> The [`createSlotTemplateNode`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/elements.ts#L11-L13) method is used to convert slots into template nodes directly; this is called by the [`processAttributeWithoutOverride`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.ts#L30-L50) method, which is in turned called for specified named slots that we are looking out for. This is how we process named slots - hopefully this addresses your question (and if it doesn't, do let me know).
> 
> Regardless, this issue might need further investigation 😓

It does address my question, thank you! Running it in the debugger with the appropriate breakpoint, I can see it too:

image

Could the reason be because `dropdown` is directly being imported and referenced by `Tabset.vue`, and consequently is not subjected to the node processing flow, hence it would be wrong to pass the slot as an attribute within the `Tabset.vue` file?

If this is the case, I agree with your suggestion to treat header as a prop which was done previously as well. It definitely does seem like a better alternative that fits the bigger picture. 😄

Tim-Siu:
> > The [`createSlotTemplateNode`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/elements.ts#L11-L13) method is used to convert slots into template nodes directly; this is called by the [`processAttributeWithoutOverride`](https://github.com/MarkBind/markbind/blob/master/packages/core/src/html/MdAttributeRenderer.ts#L30-L50) method, which is in turned called for specified named slots that we are looking out for. This is how we process named slots - hopefully this addresses your question (and if it doesn't, do let me know).
> > Regardless, this issue might need further investigation 😓
> 
> It does address my question, thank you! Running it in the debugger with the appropriate breakpoint, I can see it too:
> 
> image
> 
> Am I right to say that because `dropdown` is directly being imported and referenced by `Tabset.vue`, and consequently is not subjected to the node processing flow, hence it would be wrong to pass the slot as an attribute within the `Tabset.vue` file? (As you mentioned, that we do not use header templates directly in our `.vue` files)
> 
> If this is the case, I agree with your suggestion to treat header as a prop which was done previously as well. It definitely does seem like a better alternative that fits the bigger picture. 😄

Hi @gerteck ,

Thank you for your great research!

I concur with your analysis regarding the root cause of this issue. It appears that the nodes are not subject to recursive processing, and the `<dropdown>` within `Tabset.vue` does not undergo transformation by the `NodeProcessor` to encapsulate the attribute within a `<template>` wrapper for slot content.

For the immediate solution, I support the two approaches you've outlined:
- Manually setting the slot content when importing `<dropdown>`
- Passing the header as a prop

@lhw-1's suggestions appear reasonable, and I recommend adhering to the second approach.

For a more comprehensive solution that might involve modifications to the `NodeProcessor` (such as recursively processing nodes), we can create a separate issue to address this at a later time when it becomes a priority.

packages/vue-components/src/Dropdown.vue
```diff
@@ -11,7 +11,7 @@
         :class="{'disabled': disabledBool}"
         data-bs-toggle="dropdown"
       >
-        <slot name="header"></slot>
+        <span v-html="tabgroupHeader"></span>
```
yucheng11122017, line 14: Hmm I'm not keen on this naming since this is only true for tab groups but not for other headers. 
Could we keep the header name?

yucheng11122017 (commented):
Hi @gerteck thanks for the work! 
I realised that tabs doesnt have any test cases.. which is why we probably missed out on this regression in behaviour. If you have the time and capacity, you could try writing test cases for tabs. If you don't though, it's ok. Let us know and we will open a issue for this.

packages/vue-components/src/Dropdown.vue
gerteck: Sure! The rationale was give greater differentiation in the way the header is passed as a prop instead of relying on slot conversion, but it is evident from the code directly. 

I would be down to try writing some test cases for the tabs! Test cases would also prevent any further regressions until a better way to pass down the header is found.

Tim-Siu:
> Hi @gerteck thanks for the work! I realised that tabs doesnt have any test cases.. which is why we probably missed out on this regression in behaviour. If you have the time and capacity, you could try writing test cases for tabs. If you don't though, it's ok. Let us know and we will open a issue for this.

I think we do have some unit test regrading tab-group in `NodeProcessor.test.ts` but more advanced unit tests or functional test may be necessary to test the intended behavior for the header.

packages/vue-components/src/Dropdown.vue
gerteck: I just realised that setting it as the same name header causes two of the current tests to fail under `dropdown` as the header no longer shows up as an attribute in the html rendered. This is probably because of the slot name and the attribute prop name having the same name.

I could edit the testcases to remove the header, or simply revert to using a separate name again to prevent this issue

gerteck:
* Added some snapshot tests for Tabset (Tabs), Tabgroup, Tabs to catch visual regressions during future changes.  
* Renamed the `header` prop back into `tabGroupHeader`, as currently, it is a workaround used only to generate tabgroups in `Tabset`, and I think it would be confusing to have both a `header` named slot and prop.
  * It also seemed to cause some regression which I still don't understand the cause of. Having an added "header" prop seems to modify the DOM of nested `dropdown`s (as in `Dropdown.spec.js`), although I don't see any visual change.


* Readded the named slot `header` 
  * I missed out a regression when I removed the named slot `header`. (The impacted use case was `navbar` with `dropdown` children in markbind syntax where it was passed as a attribute and converted into a named slot). 
  * It also looks like there are no test cases for this so I will add some as well.

Tim-Siu (approved):
Thank you for delving into adding snapshot tests for the Vue plugins. I think they are well-written and sufficient to cover the concern of the issue, i.e., Tab-Group Header not displayed in when Dropdown is nested in Tabset.

tlylt:
@Tim-Siu please replace the commit message with the proposed message whenever you merge the PR in the future.

Tim-Siu:
> @Tim-Siu please replace the commit message with the proposed message whenever you merge the PR in the future.

Sorry for the mistake. Should I update the commit message and force push?

tlylt:
> > @Tim-Siu please replace the commit message with the proposed message whenever you merge the PR in the future.
> 
> Sorry for the mistake. Should I update the commit message and force push?

Think we can leave it, just take note next time

PR #2477, opened 2024-03-27
Add pagefind

jingting1412:
**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [x] Feature addition or enhancement
- [ ] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

**Overview of changes:**
To track the addition of pagefind for full-text search. Welcome anyone to help look through. 

**Anything you'd like to highlight/discuss:**

**Testing instructions:**

**Proposed commit message: (wrap lines at 72 characters)**

---

**Checklist:** :ballot_box_with_check:

- [ ] Updated the documentation for feature additions and enhancements
- [ ] Added tests for bug fixes or features
- [ ] Linked all related issues
- [ ] No unrelated changes 

---

**Reviewer checklist:**

Indicate the [SEMVER](https://semver.org/) impact of the PR:
- [ ] Major (when you make incompatible API changes)
- [ ] Minor (when you add functionality in a backward compatible manner)
- [ ] Patch (when you make backward compatible bug fixes)

At the end of the review, please label the PR with the appropriate label: `r.Major`, `r.Minor`, `r.Patch`.

Breaking change release note preparation (if applicable):
- To be included in the release note for any feature that is made obsolete/breaking

> Give a brief explanation note about:
>   - what was the old feature that was made obsolete
>   - any replacement feature (if any), and
>   - how the author should modify his website to migrate from the old feature to the replacement feature (if possible).

Issue #2084, opened 2023-01-15
Investigate and list what needs to be done for Vue V3 migrartion


ang-zeyu:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Is your feature request related to a problem? Please describe.

This is a hard ticket as it requires fairly comprehensive understanding of the entire codebase, but would be nice to have soon.
With SSR, things are a tad bit more complex as well.

### Describe the solution you'd like

The request here is to comprehensively investigate and document what needs to be done for vue v3 migration into a separate issue as a checklist of items.

If any, link said items to other tickets in the repository as well.

ang-zeyu:
@MarkBind/active-devs, if you're interested as well this is very nice to have.

EltonGohJH:
@MarkBind/cs3282-developers 
Do give any thoughts on this plan.

- [ ] Refactor usage of vue-server-renderer, vue-template-compiler in PageVueServerRenderer.ts. I seriously think it is simply replacing most of the stuff using createSSRApp function introduced in vue3.
- [ ] Remove vue2 plugins. I will update the list here. While removing we need to refactor too (portal-vue is no longer used and replace with teleport which is native to vue)
- [ ] Refactor all components to vue 3. (A lot of stuff is deprecated and we would need to refactored every vue components. Github co-pilot helps so if anyone is helping in the future can try that.
- [ ] Check that hot reload is working. I think with the update of functions in PageVueServerRenderer. I am not sure whether that is working.


I dun think there is any need to change anything with regards to webpack. I updated some components to vue3 on my end. The page is loading but I think I am ironing out some bugs. (I am not 100% sure whether my idea is working too and I am just trying)


Q1. It is not possible to tell whether there is FOUC. I throttle the network locally but I still dun see the FOUC.  (Do I need to upload to netlify and then I can find out. ) 
Q1b. Also do anyone know why we are so particular/delibrate in fixing in aleviating FOUC.
Q2. Are there any invariants that we need to keep in mind? (I feel like with vue3 testcases will have to be updated and I am not sure how much regression testing there will be.)

lhw-1:
Just as a side note, #1536 should be important to keep in mind :)

EltonGohJH:
@jovyntls 
SFLR. I guess refactor might be the wrong word to use here. Maybe update and fix warning would be more apt.
I think that library wise only portal-vue is affected since it is now supported by vue 3 natively. 
[...]

PR #2568, opened 2024-08-12
Add PageFind Plugin (Re-PR) 

gerteck:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

This PR introduces the PageFind plugin to the project, enabling full-text search capabilities for static sites.

This PR builds on #2477

See [PageFind](https://pagefind.app/)

In brief, PageFind plugin indexes all files during the build process and generates the necessary search data. This data is then bundled into static assets that are served alongside the site. The assets are saved directly into the output directory (/_site/pagefind), and the plugin script is designed to load these assets dynamically.

* Currently facing difficulties trying to access the generated pagefind assets within the plugin script, not sure how to do so,  investigating. Current workaround to showcase an MVP is to directly save and load these assets into MarkBind first, but not a practical solution.
* Using PageFind's default UI, which doesn't seem to fit into the theme of MarkBind.
* Am currently conflicted on how to proceed, whether to create a new custom Vue component.

Additional Tasks:
- Update Documentation
- Update Testing

**Testing instructions:**

1. Add pagefind plugin into `site.json`
```
...
"plugins": [
  "pagefind"
]
```
2. In any page, add the searchbar component: `<searchbar pagefind></searchbar>`

Add PageFind Plugin

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**


damithc:
Thanks for picking this @gerteck 
It is indeed a good addition to MarkBind if we can have a decent built-in full-text search, instead of having to rely on external search services as we do now.

lhw-1:
As discussed previously, this is a fantastic addition to MarkBind!

Something to consider after this PR would be the usability aspect. Given that the existing search function (by header) differs from this one (by text), with neither being replacements for the other (and with valid use cases for both), what we should consider is:

How do we increase the visibility of this plugin? Is documentation sufficient for users to be able to find and use this plugin, or should we actively include it in our templates? (And if so, how do we arrange / prioritise the two search functions in the UI?)

As mentioned during the group discussion, [SICP JS](https://sourceacademy.nus.edu.sg/sicpjs/3.2.1) in Source Academy handles this problem by putting both search functions side-by-side. By this train of thought, it might be worthwhile to make both of the search functions into modular components.

gerteck:
> How do we increase the visibility of this plugin? Is documentation sufficient for users to be able to find and use this plugin, or should we actively include it in our templates? (And if so, how do we arrange / prioritise the two search functions in the UI?)
> 
> As mentioned during the group discussion, [SICP JS](https://sourceacademy.nus.edu.sg/sicpjs/3.2.1) in Source Academy handles this problem by putting both search functions side-by-side. By this train of thought, it might be worthwhile to make both of the search functions into modular components.

Thank you for the insights! Completely agree with the further considerations for the feature addition. I think it relates with @2529 and there will be a need for further analysis of the cost and benefit of making certain plugins default and increasing the usability/visibility of the features of each plugin. 

The integration of both search functions as modular components could also be new issues to investigate and discuss!

Will work on adding documentation for the plugin after further discussion and confirmation on the feature 👍

damithc:
To add, the ideal outcome is that this plugin replaces both the current search (i.e., headers only) feature and the need to use third party search. The ideal is if this plugin can do what Algolia currently does for us (in terms of what user sees -- example https://nus-cs2103-ay2425s2.github.io/website/ ), or at least come close.

lhw-1:
> To add, the ideal outcome is that this plugin replaces both the current search (i.e., headers only) feature and the need to use third party search.

During the discussion with @gerteck, the point made was that searching by header and searching by text have somewhat different use cases and each of them might be necessary for different situations (especially since MarkBind tends to be header-heavy given its use case). If the initial goal was to replace the current search with this plugin, then I think we should consider one of these:

1. Either make this new plugin into a component & make it the default search, or:
2. Find a way to integrate the two components into one (maybe there can be an overarching search component with a toggle between header-search and text-search via a prop?)

Of course, keeping in mind that this is out of scope for this PR & we may have to open a new issue to continue discussing this (@gerteck feel free to start this discussion issue), also given this comment:

> there will be a need for further analysis of the cost and benefit of making certain plugins default and increasing the usability/visibility of the features of each plugin.

Tim-Siu:
Since Vue2 has reached the end of its life and affects our ability to utilize the latest packages, I also agree it is time to seriously consider migrating to Vue3.

damithc:
Yup, no objections to migrating to Vue 3. Als see https://github.com/MarkBind/markbind/issues/2084

gerteck:
Closed in favor of #2857 , which will hopefully be the final implementation for the first version of the pagefind plugin in MarkBind

Thanks to the migration from Vue 2 -> Vue 3 previously, we can implement a more robust Vue 3 searchbar component.

A summary of the discussion in this thread includes how
* the ideal outcome is that this plugin replaces both the current search (i.e., headers only) feature and the need to use third party search. 
* The ideal is if this plugin can do what Algolia currently does for us or at least come close.
* Given a robust implementation, we can deprecate the current search bar and not worry too much about backward compat.

A more recent reference for beta PageFind implementation can be found at #2771

Closing this PR!

PR #2584, opened 2025-01-17
Add code highlighting based on absolute positions

### LINKED ISSUE #2573: Highlighting code blocks: leading spaces should be counted
Context: I was  trying to highlight these leading spaces, but was not able to.
image

Consider the following code:
``````
```{highlight-lines="[:4]"}
  abcd
````
```````

Expected: highlight `ab` and the two spaces in front
Actual: highlights `abcd`

So, it seems the current counting doesn't include leading spaces.

### PR DESCRIPTION
**Pull request by AgentHagu**

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
This PR closes #2573. In summary, it allows users to now optionally highlight leading whitespace, giving them more options when using MarkBind.

Users can now add an optional `+` before their char bound specifier to indicate that the index should be considered as its absolute value, i.e. inclusive of leading whitespace/indentation. For example, previously `1[:4]` would highlight the `abcd` in `  abcd`. But now, by using `+1[:4]`, MarkBind will highlight `  ab` in `  abcd` (i.e. it will also highlight the first 2 spaces in the string since it considers the end bound of 4 as its absolute value, rather than relative to the indentation level).

This PR also adds automatic conversion of tabs to 4 spaces by default. The User Guide has been updated to reflect the changes.

**Testing instructions:**
A test similar to the example brought up in #2573 can be used to test the new functionality. For example:

````
```{highlight-lines="1[:4]"}
  abcd
```
````
`abcd` should be highlighted, i.e. exclusive of the 2 whitespaces in the beginning.

````
```{highlight-lines="+1[:4]"}
  abcd
```
````
Now, `  ab` should be highlighted, since the "+" symbol was added, so the absolute value of the bounds is considered in the bounds calculation.

Add optional absolute char indexing for highlighting

Currently, there is no way to include highlighting of leading
whitespace when using the `highlight-lines` attribute. Providing such
an option may be beneficial to users, giving them more freedom
when using MarkBind.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**

### CODE REVIEW

**Review by gerteck — COMMENTED** [review]
Looking at the use case of highlighting, it seems that the reason why this feature was developed this way was because of the usecase of highlighting code blocks, which are appropriately indented. (Given that the documentation is under [code](https://markbind.org/userGuide/formattingContents.html#line-highlighting)

Additionally, to capture whitespace from front of line, it could be better to add symbol in front rather than the back?

File: packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts [hunk 1]
```diff
@@ -95,38 +96,50 @@ export class HighlightRuleComponent {
...
+   * on the user's preference.
    *
    * @param bound The user-defined bound
    * @param line The given line
+   * @param highlightSpaces Whether to highlight leading whitespaces
    * @returns {[number, number]} The actual bound computed
    */
-  static computeCharBounds(bound: [number, number], line: string): [number, number] {
+  static computeCharBounds(bound: [number, number], line: string,
+                           highlightSpaces: boolean): [number, number] {
```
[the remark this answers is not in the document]
AgentHagu: Good catch! Thanks

AgentHagu:
> Additionally, to capture whitespace from front of line, it could be better to add symbol in front rather than the back?

Good idea, I'll try moving the symbol to the front, such as an underscore like "_1[2:4]" to make it more intuitive

gerteck:
> > Additionally, to capture whitespace from front of line, it could be better to add symbol in front rather than the back?
> 
> Good idea, I'll try moving the symbol to the front, such as an underscore like "_1[2:4]" to make it more intuitive

Sorry, would something like

`1[_:4]` where `_` replaces the digit if we want to include the whitespace? Not sure if it will complicate things, but it seems most intuitive to me, since it relates to the start of the highlight / line 


Does anyone else have any suggestions?

damithc:
Thanks @AgentHagu for this PR, and others for comments.
In terms of feature design, what if we treat spaces as regular characters, without using any special syntax? Would that be cleaner?
Don't worry about backward compatibility. It is unlikely to affect any existing MarkBind sites, as this is a very much an edge case of usage.

AgentHagu:
> Thanks @AgentHagu for this PR, and others for comments. In terms of feature design, what if we treat spaces as regular characters, without using any special syntax? Would that be cleaner? Don't worry about backward compatibility. It is unlikely to affect any existing MarkBind sites, as this is a very much an edge case of usage.

Thanks for the feedback! Just to clarify, do you mean having char-bound highlighting ignore the indentation level by default? So the spaces within the indents are now automatically considered in the char bounds?

damithc:
> Thanks for the feedback! Just to clarify, do you mean having char-bound highlighting ignore the indentation level by default? So the spaces within the indents are now automatically considered in the char bounds?

Oh, so that's why spaces were not counted before! I didn't realise. So, the current counting is relative to the first non-space char, by default?
Perhaps the solution is to introduce a notation to indicate absolute char counting instead?

AgentHagu:
> Oh, so that's why spaces were not counted before! I didn't realise. So, the current counting is relative to the first non-space char, by default?

Yep, by default the indentation level is checked and accounted for and the index counting is relative to the first non-whitespace char.

> Perhaps the solution is to introduce a notation to indicate absolute char counting instead?

To clarify, do you mean introducing something like `1[0.3]` instead of `1[0:3]` for whitespace-inclusive char-bound highlighting? If so, I think this could be a good solution that doesn't affect existing MarkBind sites and also provides new options to users. It also means we wouldn't need the special syntax anymore.

damithc:
> To clarify, do you mean introducing something like `1[0.3]` instead of `1[0:3]` for whitespace-inclusive char-bound highlighting? If so, I think this could be a good solution that doesn't affect existing MarkBind sites and also provides new options to users. It also means we wouldn't need the special syntax anymore.

@AgentHagu I was thinking of something like `+1[0:3]` where `+` indicates numbers are absolute i.e., similar to your current syntax but the meaning is different.

AgentHagu:
> @AgentHagu I was thinking of something like `+1[0:3]` where `+` indicates numbers are absolute i.e., similar to your current syntax but the meaning is different.

I see. In that case, I believe the implementation should already support absolute char counting, so only the syntax needs to be updated

damithc:
Just so we don't forget later, this is a user-visible change, which means UG needs to be updated as well. But you can wait till the feature behaviour is finalised before adding the UG update to the PR.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts
```diff
@@ -43,22 +43,29 @@ export class HighlightRuleComponent {
...
+      if (sliceMatch === linesliceCharMatch) {
+        isHighlightSpaces = groups.shift() === '_';
+      }
+
       const lineNumber = HighlightRuleComponent
         .isValidLineNumber(groups.shift() ?? '', 1, lines.length, lineNumberOffset);
       if (!lineNumber) return null;
 
       const isUnbounded = groups.every(x => x === '');
+
```
lhw-1, line 60:
```suggestion
```

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts
```diff
@@ -95,38 +102,50 @@ export class HighlightRuleComponent {
...
       start += indents.length;
       // Clamp values
       if (start < indents.length) {
         start = indents.length;
       } else if (start > line.length) {
         start = line.length;
       }
+    } else if (start > line.length) {
+      start = line.length;
     }
```
lhw-1, line 134: The conditional logic is slightly convoluted for me here. If `start !== UNBOUNDED` is true, and `isHighlightSpaces` is true, and `start > line.length` is false, what happens? (Will this even happen in theory?)

Also, given that Lines 129-130 and 132-133 are duplicated, I wonder if there is a way to simplify this chunk of code. We don't need to shy away from ternary conditional operators if they make the code easier to understand.

packages/core/test/unit/lib/markdown-it/highlight/HighlightRuleComponent.test.ts
```diff
@@ -44,28 +44,55 @@ describe('parseRuleComponent', () => {
...
+    expect(bounds).toEqual([3, '  some text'.length]);
+  });
+
+  test('handles out-of-range bounds correctly', () => {
+    const bounds = HighlightRuleComponent.computeCharBounds([30, 40], '  some text', true);
+    expect(bounds).toEqual(['  some text'.length, '  some text'.length]);
+  });
+
+  test('handles line-length end correctly', () => {
+    const bounds = HighlightRuleComponent.computeCharBounds([0, 4], '  abcd', true);
```
lhw-1, line 91: Why `abcd` here in particular? (Let me know if it's a non-issue)

lhw-1 (commented):
Thanks for the PR, @AgentHagu!

Revisiting a discussion point that we brought up earlier - in the cases of other kinds of whitespaces being present (e.g. tab or newline), how do we handle it? E.g. I can see a situation where the user copy-pastes in an existing chunk of code with whitespace, and non-space whitespace characters are added in directly. This may just be a documentation problem though given that this is not highly likely & usually these are converted to whitespaces in many IDEs etc, what do you think? Or is there a tweak we can do that will efficiently cover these cases?

packages/core/test/unit/lib/markdown-it/highlight/HighlightRuleComponent.test.ts
AgentHagu: There's no underlying reason for `abcd`, I can update the test-case to follow the other test-cases' string input of `'  some text'`.

Looking back, I'm also not sure if this test-case is needed, since it seems to overlap with the first test case's input. So maybe I'll remove it entirely instead.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts
AgentHagu: > If start !== UNBOUNDED is true, and isHighlightSpaces is true, and start > line.length is false, what happens? (Will this even happen in theory?)

In theory, it's possible for that to occur. For example, for `  some text`, start = 2 would be bounded and <= line.length and if `isHighlightSpaces` is true, then that means we should leave the start value as is, since we wish to use the absolute value of those bounds.

> Also, given that Lines 129-130 and 132-133 are duplicated, I wonder if there is a way to simplify this chunk of code. We don't need to shy away from ternary conditional operators if they make the code easier to understand.

Thanks for the feedback! I'll see if there's a way to simplify it more to improve readability.

AgentHagu:
> Thanks for the PR, @AgentHagu!
> 
> Revisiting a discussion point that we brought up earlier - in the cases of other kinds of whitespaces being present (e.g. tab or newline), how do we handle it? E.g. I can see a situation where the user copy-pastes in an existing chunk of code with whitespace, and non-space whitespace characters are added in directly. This may just be a documentation problem though given that this is not highly likely & usually these are converted to whitespaces in many IDEs etc, what do you think? Or is there a tweak we can do that will efficiently cover these cases?

Thanks for the feedback! I'll investigate this further to see if/how this problem may occur and how we could handle it.

lhw-1:
> My current idea is to see whether it's possible to convert such tabs to the corresponding number of whitespace, which should remedy the issue automatically. What do you think of this solution?

That would make sense - we'll just need to specify this behavior in the documentation :)

damithc:
> My current idea is to see whether it's possible to convert such tabs to the corresponding number of whitespace, which should remedy the issue automatically. What do you think of this solution?

@AgentHagu Note that a tab may be considered a different number of spaces by different tools: some use 2 spaces, some use 4, and some use 8 (e.g., GitHub). But yes, we can pick one (e.g., 4) and keep the users informed about it.

lhw-1:
@AgentHagu What is the current status of this PR?

AgentHagu:
@lhw-1 Sorry for the lack of updates, currently fixing some issues with the tab conversion. The user guide has been updated to reflect the new behavior as well.

AgentHagu:
Upon initial investigations, it seems that as of the current `master` branch, tab characters are 8 spaces wide by default. I've confirmed this behavior using a simple test with @IanCheah but if anyone would like to, they can also confirm it on their own.

**Code:**
image

**Code with highlighting to indicate spaces or tabs**
image

**Actual output:**
image

If the consensus is that tabs should instead be 4 spaces wide instead of 8, this PR will also target this issue with its auto-conversion of tabs to 4 spaces wide. The UG has also been updated to indicate this change. However, it is possible that existing MarkBind sites that use tabs in their fenced code will be affected by this change, since visually the tabs will be cut in half.

To avoid breaking existing sites, we could also choose to retain the 8 space wide tabs for backward compatibility. What does everyone think?

lhw-1:
@damithc in terms of the tab conversion to 4 instead of 8, would this still hold?

> Don't worry about backward compatibility. It is unlikely to affect any existing MarkBind sites, as this is a very much an edge case of usage.

Given that this is a visual change, and the only functionality (that I can think of at the moment) that might be affected by tab size is the highlight feature being merged in with this PR, I think it should be fine to make this change? I don't think the 8 spaces was an intentional implementation (given that 2 or 4 is the usual standard).

damithc:
> Given that this is a visual change, and the only functionality (that I can think of at the moment) that might be affected by tab size is the highlight feature being merged in with this PR, I think it should be fine to make this change? I don't think the 8 spaces was an intentional implementation (given that 2 or 4 is the usual standard).

@lhw-1 Yes, it should be fine. I don't use tabs in my code anyway. So, none of my MarkBind sites will be affected.

lhw-1:
@AgentHagu What is the status of this PR?

AgentHagu:
@lhw-1 The PR should be finished, accomplishing the original task and adding auto conversion of tabs to 4 spaces.

lhw-1:
Once the PR is tested and good to go, we can merge it in. Thanks @AgentHagu!

lhw-1:
@all-contributors please add @AgentHagu for code

lhw-1:
https://github.com/all-contributors please add @AgentHagu for code

PR #2586, opened 2025-01-23
Fix broken annotation label

### LINKED ISSUE #2563: Image annotation: example output is broken
This doesn't seem right:
image

Location https://markbind.org/userGuide/components/imagesAndDiagrams.html#annotations

Not sure if the feature itself is broken, or it's the example that is broken (due to how it is generated in the UG).

Incogdino:

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Closes #2563 

**Overview of changes:**
After investigating #2563, the image annotation started breaking after converting to use the slots API.

According to Vue's documentation, content passed by the slots API are considered to be owned by the parent component that passes them in and so styles do not apply to them. 

A separate CSS rule to target all elements within the div that holds the slot is added to override the parent's styles. This fixed the issue and the problem no longer persists.

Some screenshots: 
image
image

**Testing instructions:**
Testing can be done locally by navigating to the `docs` folder and running `npm markbind serve -d -o userGuide/components/imagesAndDiagrams.md`

Fix broken annotation label

---

- [X] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/vue-components/src/annotations/AnnotatePoint.vue
+++ b/packages/vue-components/src/annotations/AnnotatePoint.vue
@@ -198,11 +198,12 @@ export default {
       return (this.hasContent || this.hasHeader) && (this.legend === 'popover' || this.legend === 'both');
     },
     computedBottomHeader() {
-      const labelSlotContent = this.$scopedSlots.label == undefined ? undefined : this.$scopedSlots.label();
-      const headerSlotContent = this.$scopedSlots.header == undefined ? undefined : this.$scopedSlots.header();
+      const labelSlotContent = this.$scopedSlots.label === undefined ? undefined : this.$scopedSlots.label();
+      const headerSlotContent = this.$scopedSlots.header === undefined
+        ? undefined : this.$scopedSlots.header();
 
-      const labelText = labelSlotContent == undefined ? undefined : labelSlotContent[0].children[0].text
-      const labelHeader = headerSlotContent == undefined ? undefined :headerSlotContent[0].children[0].text
+      const labelText = labelSlotContent === undefined ? undefined : labelSlotContent[0].children[0].text;
+      const labelHeader = headerSlotContent === undefined ? undefined : headerSlotContent[0].children[0].text;
 
       if (labelText === undefined && labelHeader !== undefined) {
         return labelHeader;
--- a/packages/vue-components/src/annotations/AnnotatePoint.vue
+++ b/packages/vue-components/src/annotations/AnnotatePoint.vue
@@ -198,13 +198,20 @@ export default {
       return (this.hasContent || this.hasHeader) && (this.legend === 'popover' || this.legend === 'both');
     },
     computedBottomHeader() {
-      if (this.label !== '' && this.header === '') {
-        return this.label;
+      const labelSlotContent = this.$scopedSlots.label == undefined ? undefined : this.$scopedSlots.label();
+      const headerSlotContent = this.$scopedSlots.header == undefined ? undefined : this.$scopedSlots.header();
+
+      const labelText = labelSlotContent == undefined ? undefined : labelSlotContent[0].children[0].text
+      const labelHeader = headerSlotContent == undefined ? undefined :headerSlotContent[0].children[0].text
+
+      if (labelText === undefined && labelHeader !== undefined) {
+        return labelHeader;
       }
-      if (this.label === '' && this.header !== '') {
-        return this.header;
+      if (labelText !== undefined && labelHeader === undefined) {
+        return labelText;
       }
-      return `${this.label}: ${this.header}`;
+
+      return `${labelText}: ${labelHeader}`;
     },
   },
   methods: {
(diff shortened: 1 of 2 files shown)


Tim-Siu:
Screenshot 2025-02-03 at 11 50 41
Thanks @Incogdino for investigating this issue. The investigation is sound. It is also mentioned in [the documentation](https://vuejs.org/guide/components/slots) that "The slot mechanism is declarative: the component cannot directly read or transform the slot content". The fix works for me with both safari and chrome on macOS.

Tim-Siu (approved):
LGTM. This modification targets only the hover-label and should be a safe fix.

lhw-1:
On the [deploy preview](https://deploy-preview-2586--markbind-master.netlify.app/), this is what it looks like:

image

Additionally, when tested on both Windows (Chrome, Edge, Firefox) and Ubuntu (Chrome, Firefox), the issue still seems to persist @Incogdino

lhw-1:
The issue linked in #2563 is targeted at fixing this part in particular:

image

Incogdino:
> On the [deploy preview](https://deploy-preview-2586--markbind-master.netlify.app/), this is what it looks like:
> 
> image
> 
> Additionally, when tested on both Windows (Chrome, Edge, Firefox) and Ubuntu (Chrome, Firefox), the issue still seems to persist @Incogdino

Sorry for the overlook! Totally missed out on the arrows in the image!

It should be fixed now. The same issue is with changing from props to slots in the vue component.

image

Here I tested label 2 with no header and label 3 with both label and header

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -198,13 +198,21 @@ export default {
...
     },
     computedBottomHeader() {
-      if (this.label !== '' && this.header === '') {
-        return this.label;
+      const labelSlotContent = this.$scopedSlots.label === undefined ? undefined : this.$scopedSlots.label();
+      const headerSlotContent = this.$scopedSlots.header === undefined
+        ? undefined : this.$scopedSlots.header();
+
+      const labelText = labelSlotContent === undefined ? undefined : labelSlotContent[0].children[0].text;
+      const labelHeader = headerSlotContent === undefined ? undefined : headerSlotContent[0].children[0].text;
```
lhw-1, line 206: We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here, i.e.

```suggestion
      const labelSlotContent = this.$scopedSlots.label?.();
      const headerSlotContent = this.$scopedSlots.header?.();

      const labelText = labelSlotContent?.[0]?.children?.[0]?.text;
      const labelHeader = headerSlotContent?.[0]?.children?.[0]?.text;
```

You can probably combine these to compact them even further but I think this is fine enough considering readability.

packages/vue-components/src/annotations/AnnotatePoint.vue
```diff
@@ -198,13 +198,21 @@ export default {
...
+
+      if (labelText === undefined && labelHeader !== undefined) {
+        return labelHeader;
       }
-      if (this.label === '' && this.header !== '') {
-        return this.header;
+      if (labelText !== undefined && labelHeader === undefined) {
+        return labelText;
```
lhw-1, line 212: Of the two missing cases here: if both are not undefined I think there shouldn't be an issue, but if both are undefined, should the case be handled? Can such a case arise?

lhw-1 (changes requested):
Some nits on the new changes @Incogdino

packages/vue-components/src/annotations/AnnotatePoint.vue
lhw-1: If you have further thoughts on making this more readable, feel free to suggest!

packages/vue-components/src/annotations/AnnotatePoint.vue
Incogdino: Such a case would not arise since the function will only be called given that the legend option specifies a legend to be created and a label option is specified in the `v-if` directive 

```
    <div v-if="hasBottomText && hasLabel" class="legend-wrapper">
      <div class="text-header">
        {{ computedBottomHeader }}
      </div>
      <div v-if="hasContent">
        <slot name="content"></slot>
      </div>
    </div>
```

Incogdino: To add on, the missing label case is likely unnecessary as well since the block would not be rendered if label is not specified.

packages/vue-components/src/annotations/AnnotatePoint.vue
Incogdino: I agree this makes it much more readable!

lhw-1 (approved):
Feature seems to work as intended, and the documentation update is a nice touch. Thanks @Incogdino!

PR #2599, opened 2025-02-10
Add custom host validation

### LINKED ISSUE #2417: Validation when specifying custom host

### Any related issues?

#2382

_No response_

Specifying a custom host in #2395 could cause some unhandled errors to be exposed to the user after the building of pages when the specified host is not valid. This could lead to 

1. Users being confused by the unhandled error
2. Users having to wait till the pages are finished building to realise there is an error

which could create a not so good UX.

Validate the host before the building of pages. Should discuss what should be validated and what should not as well.

_No response_

### Additional context

_No response_

AgentHagu:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Fixes #2417

This PR adds early custom host validation to ensure the host specified is a valid one. 

Currently, there is no validation, which results in an error. For example, the command `markbind serve -a 999.999.999.999` would lead to something like this:
image

Ideally, the validation occurs before the site is built since a user might give an invalid address and try to serve a large site with many pages. The user will have to wait for the site to finish building just to find out that the custom host they gave is invalid and to try again. For example:
image

Currently, there is a zero IP address check to warn users of security risks. However, this zero IP regex check also flags invalid IPs like `0`. By moving the validity check earlier before this zero IP address check, it will ensure that the regex will only check valid IPs. 

Otherwise, the following scenario may occur for users:
1. User enters `0` as the IP
2. User gets a warning that the address `0.0.0.0`
has security risks
3. User proceeds by entering 'y'
4. User gets an error that the IP is invalid

This may be confusing to users, who got a warning for their address, acknowledged that warning, only to find out later that it was invalid. While not a huge issue, it's a minor inconvenience we can avoid.

The User Guide will be updated about `serve` accepting IPv6 issues in PR #2621, specifically this [commit](https://github.com/MarkBind/markbind/pull/2621/commits/18e85c239511de94660e1e6c57afcafcb53d3170)

**Testing instructions:**
Use the `serve` command with a valid and an invalid host to test this fix.

For example:
**Valid addresses**
`markbind serve -a 127.0.0.1` 
`markbind serve -a 0.0.0.0`
`markbind serve -a ::1`
`markbind serve -a localhost`

**Invalid addresses**
`markbind serve -a 999.999.999.999`
`markbind serve -a 127`
`markbind serve -a abcdefg`

Add custom host validation

Currently, there is no deliberate host validation. If an invalid host
is provided by the user, the error will be caught but only after the
site is generated.

By adding a deliberate and earlier host validation, this will prevent
the user from waiting for the site to generate before realising that
the host is invalid.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**


AgentHagu:
Besides just IP addresses, should we also allow for hostnames like "localhost"?

In general, what should we count as a valid host for MarkBind?

Incogdino:
The feature works as expected! Tested on windows.  

This feature can greatly reduce the time needed when testing the site!

A little shortcut I use during testing to shorten the time needed to load the webpage would be to specify the `-o` flag if the goal is to just test the output of a single page! This serves up a single page and loads the other pages only when they are navigated to.

Incogdino:
> Besides just IP addresses, should we also allow for hostnames like "localhost"?
> 
> In general, what should we count as a valid host for MarkBind?

I have the following ideas:
1. Allow for IPV6 addresses to be specified (though I dont see much reason a user might want to use it, except maybe for the shorthand for localhost `::1`)
2. Allow users to specify port number along with the IP address (the current implementation automatically redirects to a different port number if the default port is being used), though again, I don't quite see why users might want to specify separate ports

lhw-1:
> Besides just IP addresses, should we also allow for hostnames like "localhost"?
> 
> In general, what should we count as a valid host for MarkBind?

Yes, we should allow for `localhost`. Beyond that, custom hostnames would be difficult to account for without allowing a wide range of allowed hostnames that are potentially invalid as well. But given that it would have worked fine until now, we should certainly look into how we can accommodate for this.

> Allow for IPV6 addresses to be specified (though I dont see much reason a user might want to use it, except maybe for the shorthand for localhost ::1)

We certainly should, by the same logic: if it is possible now for a user to use it (and it is valid), then it should still be possible. Anyway, ipv6 addresses are definitely valid use cases.

> Allow users to specify port number along with the IP address (the current implementation automatically redirects to a different port number if the default port is being used), though again, I don't quite see why users might want to specify separate ports

The [`-p` flag](https://markbind.org/userGuide/cliCommands.html#serve-command) is meant for specifying port numbers, so this wouldn't be necessary.

lhw-1:
@AgentHagu What is the status of this PR?

packages/cli/src/util/cliUtil.js
tlylt, line 1: did i miss anything? what is the change needed in this file?

AgentHagu: Oh my bad, I initially refactored my IP checks to cliUtil.js and reverted the changes but didn't completely revert it. There shouldn't be any changes need in this file, I'll amend this now

AgentHagu:
Good idea @tlylt @lhw-1, I'll add the test cases, thanks for the suggestion!

tlylt (approved):

lhw-1 (approved):
LGTM. Thanks @AgentHagu!

PR #2621, opened 2025-03-05
Fix issues with MarkBind's patch of live-server

### LINKED ISSUE #2617: Invalid URL of IPv6 address used for `serve` command's live preview

### Any related issues?

_No response_

Windows 10

### MarkBind version

5.5.3

When IPv6 addresses are provided for the `serve` command, the URL opened for live preview is an invalid one. This often leads to an error pop-up, which may mislead users to believe that the address they provided was invalid. In reality, it was the live preview URL that was invalid.

For example, `markbind serve -a ::1` will attempt to launch the URL "http://::1:8080". The proper URL for this address is actually "http://[::1]:8080/", i.e. the IPv6 address is enclosed in square brackets.

Image

Image

There's currently a workaround which is to add the square brackets within the `serve` command, however this is unintuitive for users.

### Expected behavior

When an IPv6 address is specified, the URL launched for live preview automatically appends the square brackets, leading to a valid link.

For example, the command `markbind serve -a ::1` should lead to the following page:

Image

### Anything else?

_No response_

### LINKED ISSUE #2620: Unavailable address opening the wrong URL during `serve` command

### Any related issues?

_No response_

Windows 10

### MarkBind version

5.5.3

When an unavailable address is given for the `serve` command, the user is informed and is told that MarkBind will try another address. However, the URL opened still uses the original unavailable address, leading the user to the wrong URL for their MarkBind site. This could lead to confusion for the user who might suspect their site has failed to load.

For example:
`markbind serve -a 1.1.1.1` will lead to the following messages from MarkBind
Image

However, despite serving on "http://127.0.0.1:8080", the URL opened is still "http://1.1.1.1:8080". Also, going to the actual URL will load the MarkBind site, so the actual site is properly loaded, it's just that the URL opened is wrong.
Image

### Expected behavior

When an unavailable address is given, the user is informed and the URL opened correctly matches the one in which the MarkBind site is served. 

Using the previous example, the URL opened should be http://127.0.0.1:8080, leading to the user's MarkBind site.

### Anything else?

_No response_

### LINKED ISSUE #2651: Missing warning for IPv6 zero address

### Any related issues?

_No response_

Windows 11

### MarkBind version

5.6.0

Currently, there's a warning for the user when they try to serve on address `0.0.0.0`. However, when the IPv6 zero address (`::`) is used there is no such warning, even though it would have the same security risks.

For e.g.:
- `markbind s -a 0.0.0.0` would result in the following warning:
   `WARNING: Using the address '0.0.0.0' could potentially expose your server to the internet, which may pose security risks. Proceed with caution? [y/N]`
- `markbind s -a ::` would not result in any warning at all, even though it also exposes the server to the internet.

The issue lies with the following code:
```js
function isIPAddressZero(address) {
  const patternForZero = /^0(\.0)*$/;

  return patternForZero.test(address);
}
```

There is currently no regex/check for IPv6's zero address, that's why the warning doesn't appear.

### Expected behavior

When users try to serve on the IPv6 zero address, they should be appropriately warned about the security risks involved, similar to how we do it for `0.0.0.0`.

### Anything else?

_No response_

AgentHagu:

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
- Fixes #2617 
- Fixes #2620 
- Adds missing CHANGED comments to `live-server` patch
- Fixes #2651 
- Updates logging code for zero address (`0.0.0.0` or `::`)
   `live-server`'s original code has some code to help log all the IPv4 addresses opened when serving on `0.0.0.0`.
    To accommodate IPv6 addresses, we should also update this code to log all the IPv6 addressed opened when serving on `::`.

The logging code changes can only be observed/tested if the logLevel in `packages\cli\src\cmd\serve.js` is updated to a higher level than 2.

The current IPv6 zero regex flags some invalid IPv6 strings, like `0::0::0` or `:0:`. This issue of invalid addresses should be fixed by PR #2599. At most, users will be warned about using the zero address, only to find out later that it's invalid, which may be a bit inconvenient or unintuitive. This could be fixed by updating #2599 to move the IP validity check earlier before the zero IP address check.

**Testing instructions:**
To test fixes for #2617:
1.  Serve a markbind site using an IPv6 address, such as `markbind serve -a ::1`. 
2. The URL opened should be something like `http://[::1]:8080/`, i.e. the IPv6 address is wrapped in square brackets to form a valid URL. IPv4 addresses should not be affected/adjusted.

To test fixes for #2620:
1. Serve a MarkBind site using an unavailable address, such as:
   - `markbind serve -a 1.1.1.1`
   - `markbind serve -a 192.168.1.0`
   
    Alternatively, send the same command with an originally available address twice, effectively making that address unavailable for the second `serve` command.

3. The URL opened after serving the site should be the one the MarkBind site is actually hosted on, not the unavailable address provided in the command.

To test fixes for #2651:
1. Try serving a MarkBind site with addresses like `::`, `0::`, `0::0` or any equivalent address.
2. Verify that the following warning appears
     `WARNING: Using the address '0.0.0.0' or '::' could potentially expose your server to the internet, which may pose security risks.`

To test IPv6 logging code update:
1. Go to `packages\cli\src\cmd\serve.js` and update line 88 `logLevel: 0` to any level higher than 2.
2. Serve a MarkBind site using the IPv6 zero address `::`.
3. Proceed when warned about the zero address
4. Verify that the available IPv6 addresses are logged, for example:
    image

Fix issues with MarkBind's patch of live-server

Currently, there are a few issues regarding MarkBind's patch of
`live-server` in `packages\cli\src\lib\live-server\index.js`. The issues
are:
1. Opening an invalid URL for IPv6 addresses
2. Opening the wrong URL when attempting to serve on an
    unavailable address.
3. Missing `CHANGED` comments in our patch
4. Missing warning when user serves on `::`
5. Missing patch for `live-server` code to log available IPv6 addresses

Fixes include:
- Wrapping IPv6 addresses in square brackets `[]` to form valid URLs.
- Explicitly reassigning `host` when the provided address is
   unavailable.
- Adding the missing `CHANGED` comments.
- Updating the zero address check to include IPv6's zero address.
- Updating logging to include available IPv6 addresses for `::`.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**

- [x] Minor (when you add functionality in a backward compatible manner)


packages/cli/src/lib/live-server/index.js
```diff
@@ -284,7 +284,8 @@ LiveServer.start = function(options) {
     } else if (e.code === 'EADDRNOTAVAIL') {
       console.log('%s is not available. Trying another address'.yellow, host);
       setTimeout(function() {
-        server.listen(port, '127.0.0.1');
+        host = "127.0.0.1";
```
tlylt, line 287: so the issue is that the new address that it is now trying, which is `127.0.0.1` is not re-assigned to reflect that...I think we need to do the same for the port in the `EADDRNOTAVAIL` clause, which has the same problem?

Looking at this code...we probably should improve by iterating through a list of host and port, since the address and port we tried could still be in use/not available, probably a separate ticket..

tlylt (commented):
Interesting PR, i didn't know we support ipv6. i guess this pr fixes a bug and adds a new feature.

You might want to consider adding a tip about that ipv6 support in our user guide, somewhere in https://markbind-master.netlify.app/userguide/clicommands#serve-command

AgentHagu:
> You might want to consider adding a tip about that ipv6 support in our user guide, somewhere in https://markbind-master.netlify.app/userguide/clicommands#serve-command

Hi @tlylt, thanks for the suggestion! I'll update the user guide accordingly, perhaps as a note or something about our serve command's address `-a` option

packages/cli/src/lib/live-server/index.js
AgentHagu: > so the issue is that the new address that it is now trying, which is 127.0.0.1 is not re-assigned to reflect that...I think we need to do the same for the port in the EADDRNOTAVAIL clause, which has the same problem?

Yep, that caused `openHost` for `openURL` to use the old value rather than using the default value of `127.0.0.1`. 

About the port, I think it shouldn't have the same problem (but I agree it would still be good to reassign just in case). It looks like for port, the code dynamically gets the server's port value from checking the server `address()`'s port, so the port value used will always be up to date even without reassigning:
```
    var address = server.address();
    var serveHost = address.address;
    var openHost = host;

    var serveURL = protocol + '://' + serveHost + ':' + address.port;
    var openURL = protocol + '://' + openHost + ':' + address.port;
``` 

But I'll still add in the reassignment of port just to ensure consistency and in case this isn't followed in the future

tlylt: I see. I took another look at the file that you are changing because if we could get the port from the server, we should be able to get the address too.

It looks like this code is a patch of live-server, so we might need to follow the `CHANGED` markers to indicate the changes as well.

The code looks more complicated than what we need for our use-case and i'm not sure if there's some restrictions such that this comment implies we only serve for ipv4?
- line 316
```
// flatten address data, use only IPv4
```

AgentHagu: Good point about the `CAUGHT` markers! I'll add those in the next commit for changes relative to the original `live-server` file. It seems there are a few missing `CAUGHT` markers in our current patch, such as our addition of the `EADDRNOTAVAIL` check.

----

> The code looks more complicated than what we need for our use-case and i'm not sure if there's some restrictions such that this comment implies we only serve for ipv4?

I took a closer look at this section, which starts from line 308:
```js
if (LiveServer.logLevel > 2 && address.address === "0.0.0.0") {
  var ifaces = os.networkInterfaces();
  serveURLs = Object.keys(ifaces)
    .map(function(iface) {
      return ifaces[iface];
    })
    // flatten address data, use only IPv4
    .reduce(function(data, addresses) {
      addresses.filter(function(addr) {
        return addr.family === "IPv4";
      }).forEach(function(addr) {
        data.push(addr);
      });
      return data;
    }, [])
    .map(function(addr) {
      return protocol + "://" + addr.address + ":" + address.port;
    });
}
```

This block is some helper code for logging all available addresses when serving on `0.0.0.0`, explicitly listing the URLs users can use to access their MarkBind site. The "IPv4" comment exists because `0.0.0.0` typically only binds to IPv4 addresses, so the logging code filters out IPv6 addresses accordingly. It shouldn't have any impact on the `serve` command's ability to handle IPv6 connections.

That said, there is a minor issue:
Since we're now supporting IPv6 addresses, we should also check for the IPv6 equivalent of `0.0.0.0`, which is `::`. Currently, when users serve on `0.0.0.0`, they receive a warning:
```
WARNING: Using the address '0.0.0.0' could potentially expose your server to the internet, which may pose security risks.
```
However, no such warning exists for `::`, even though it can have similar implications.

I can add that additional warning for `::` in this PR and update the logging code to also log available IPv6 addresses when users use `::`. Let me know if that sounds good!

tlylt: > I can add that additional warning for :: in this PR and update the logging code to also log available IPv6 addresses when users use ::. Let me know if that sounds good!

Sure

AgentHagu:
As of now, the PR is mostly complete. The only suggestion missing is the reassignment of port. Currently, I'm having issues with figuring out how to do it.

Unlike the reassignment of `host` where we give a default value to use, we are currently relying on `.listen(0)` to give us a random available port. This port value is only available after the server completes the `.listen()` execution. One possible way to reassign right after is to use a callback during `listen()`, like:

```js
server.listen(0, host, () => {
  port = server.address().port;
});
```

However, from my own testing, this reassignment tends to come _after_ the `listening` event is emitted. This means the listener for `listening` at line 297 will still use the previous `port` value, rather than the reassigned one, defeating the purpose of this reassignment change.

If anyone is more familiar with the `http` module, `listen()` or the scheduling between the callback and the event emitting, feel free to fix this!

tlylt:
> As of now, the PR is mostly complete. The only suggestion missing is the reassignment of port. Currently, I'm having issues with figuring out how to do it.

We can leave it as is. My question was originally on how come we needed to do this for address but not for port. Looking at the code now, i suppose using 0 gives a random port, which is fine, but using 0.0.0.0 to give any ("random") address is unsafe, so we can't just do the same thing.

docs/userGuide/cliCommands.md
```diff
@@ -125,7 +125,9 @@ The caveat is that not building all pages during the initial process, or not reb
    Force live reload to process all files in the site, instead of just the relevant files. This option is useful when you are modifying a file that is not a file type monitored by the <trigger trigger="click" for="modal:cliCommands-livePreview">live preview</trigger> feature.
 
 * `-a <address>`, `--address <address>`<br>
-    Specify the server address/host (Default is 127.0.0.1).
+    Specify the server address/host (Default is 127.0.0.1). <br>
+    Supports both IPv4 and IPv6 addresses. <br>
+    {{ icon_examples }} `-a 192.168.0.1`, `-a ::1`
```
tlylt, line 130: Let's give examples that are not usually unavailable...but TIL `192.168.0.1` is a common default IP address assigned to many home and small business routers :0

packages/cli/src/lib/live-server/index.js
```diff
@@ -298,43 +302,51 @@ LiveServer.start = function(options) {
...
           return data;
         }, [])
         .map(function(addr) {
-          return protocol + "://" + addr.address + ":" + address.port;
+          return protocol + "://" + (isIpv6 ? `[${addr.address}]` : addr.address) + ":" + address.port;
         });
     }
 
     // Output
+    // CHANGED: Refactor logging code
```
tlylt, line 337: let's retain the original code? I don't see obvious benefit of the updated log

tlylt (changes requested):

packages/cli/src/lib/live-server/index.js
AgentHagu: I updated the code here because there was an issue with IPv6 addresses. 

The original code only outputs 1 serveURL message when `serveURL !== openURL`, even if `serveURLs.length > 1` . If a developer tried logging and used an IPv6 zero address like `::0`, the serveURL would be `::0` but the openURL would be `::`. This means that even though there might be multiple addresses used, only 1 will be logged which could be misleading.

For example, before the refactor commit, if a developer set `logLevel > 2` and did `markbind s -a ::`, they would see something like:
image
All available addresses opened by the zero address are listed out.

But if a developer does a different command like `markbind s -a ::0`, they would instead see:
image
I.e. only the address given would be listed, the other addresses opened would not be logged, even though if we check the previous example's URLs, they would work as per normal.

The refactored code just ensures the available addresses are all logged, regardless of whether `serveURL === openURL`. For example, the previous command `markbind s -a ::0` now leads to:
image

Let me know if you'd prefer to keep the original code. If this change seems unnecessary, I'm happy to revert it.

docs/userGuide/cliCommands.md
AgentHagu: Ok! Should I stick with the usual 127.0.0.1 example or give a different example from our default address?

tlylt: The usual is fine, one for ipv4 one for ipv6

packages/cli/src/lib/live-server/index.js
tlylt: How does the user configure the log level? I had the assumption that it's not something we expose to the user.

I'm ok to adjust the handling logic, but would like to keep the format of how it was logged, if possible

AgentHagu: > How does the user configure the log level? I had the assumption that it's not something we expose to the user.

To my knowledge, the user shouldn't be able to configure the log level, so you're right that this change isn't something that's visible to the user.

> I'm ok to adjust the handling logic, but would like to keep the format of how it was logged, if possible

Ok, I'll keep the log message format to the original `Serving "{root}" at {openURL} ({serveURL})`!

AgentHagu: Hi, I think it should still show. Did you update the logLevel in `serve.js` to something greater than 2, like 3?
image

Example output from `markbind s -a ::0`:
image

Let me know if it still doesn't work, I'll look into it

tlylt: Thanks, I see it now. Given that the log level is not configurable, i think it's fine.

Issue #2615, opened 2025-02-27
Possible outdated typescript module settings to be updated?


gerteck:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

#1877

### Tell us about your environment

Win 11

### MarkBind version

5.3.3 

### Describe the bug and the steps to reproduce it

The project's current typescript configuration settings are:

https://github.com/MarkBind/markbind/blob/2cadb784fb53f6135ddd070056190142b50dc6d4/tsconfig_base.json#L1-L14

While reading typescript documentation, I noticed perhaps the `module` option should be updated moving forward (particularly, set to `node16 / nodenext`), especially as we continue to update Node.js versions (See tip-box in picture):

![Image](https://github.com/user-attachments/assets/0dac53fc-b5aa-4f3f-a1e7-83237ca1c85b)

![Image](https://github.com/user-attachments/assets/68059b63-0b2e-4982-9e1c-10bba1338e1f)

* [Image-Source](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format)

* [ESM support in Node.js introduced in TypeScript 4.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html)
* [Module Option](https://www.typescriptlang.org/tsconfig/#module)
* [Module Resoultion Option](https://www.typescriptlang.org/tsconfig/#moduleResolution)

I haven't fully processed what this means and there is any possiblity for regressions etc. However, I note that previous efforts in integrating new node packages was impeded by this issue, as in #2477 , which was resolved by updating this module option.

Any further investigation would be helpful.

gerteck:
thanks for the comprehensive summary @Harjun751 .

Yes, option 2 sounds good to unblock the ESM migration. For ease and manageability of migration, let's prioritize minimizing the immediate changes needed to unblock ESM support.
* Even with an increase in node_modules size, it will only affect the CLI tool (during site generation). Since packages/core-web does not use lodash, there will be no bundle size impact for end-users visiting the generated websites.



After this migration is complete, we can work on incrementally refactoring these to native code (e.g., using `?.` for `get`, `Object.hasOwn` for `has`, `util.isDeepStrictEqual` for `isEqual`) to eventually remove the dependency entirely.
* It also seems that most of our lodash usage (like isBoolean, has, get, flatMap) are those with native equivalents in modern Node.js versions, so it should be manageable.

[...]

PR #2614, opened 2025-02-25
Update Mermaid script for efficiency and decouple from Vue

gerteck:
Dynamically load mermaid only if there are mermaid diagrams present on the page

**What is the purpose of this pull request?**

- [ ] Documentation update
- [ ] Bug fix
- [x] Feature addition or enhancement
- [x] Code maintenance
- [ ] DevOps
- [ ] Improve developer experience
- [ ] Others, please explain:

**Overview of changes:**

Update mermaid plugin script. This PR dynamically loads the Mermaid script only if there are Mermaid diagrams present on the page, improving efficiency and reducing unnecessary script loading.

In PR, Mermaid script only loaded when there are Mermaid diagrams (.mermaid elements) present on the page, reducing page load time and script size when the diagrams are not used.

* Added logic to check for the presence of Mermaid diagrams before loading the script.
* If diagrams found, Mermaid library is imported dynamically using import() to avoid unnecessary loading.
* ~~Vue directive for Mermaid rendering remains the same, now only initialized after the script is loaded.~~

**Anything you'd like to highlight/discuss:**

Other todo:
* Add in mermaid DOCS that standard `<pre>` formatting as per mermaid documentation can now be used too! 🥳 🥳. For example, with this, we can now use this [example mermaid diagram format](https://mermaid.js.org/config/usage.html#simple-full-example), and it should render correctly:

```
<pre class="mermaid">
  graph LR
      A --- B
      B-->C[fa:fa-ban forbidden]
      B-->D(fa:fa-spinner);
</pre>
```

This better aligns with how users can canonically use HTML in markdown files in markbind!

**Testing instructions:**

* Add Mermaid diagrams to a page.
* Ensure mermaid js loaded from CDN only when the Mermaid diagrams are present on page.
* Check that diagrams render correctly.
* On pages without Mermaid diagrams, verify that the Mermaid script is not loaded.

Currently, tests fail because snapshots of scripts differ. Not updating the scripts till further review is made to avoid polluting files changes for code review.

Otherwise:
* `cd .\packages\cli\test\functional\test_site\`
* `markbind serve -d`
* go to `http://127.0.0.1:8080/test_site/testMermaid.html` (or any other local host port used)
* Verify mermaid diagrams shown are ok

**Proposed commit message: (wrap lines at 72 characters)**

Update Mermaid Plugin

Dynamically load Mermaid js asset only if diagrams present.
Decouple rendering of mermaid diagrams from Vue directives.
Instead, use MutationObservers to load dynamically added 
[...]

Issue #2613, opened 2025-02-25
Package.json / Package-lock.json dependency Issue


gerteck:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Any related issues?

#2615 

### Tell us about your environment

Win 11

### MarkBind version

5.3.3

### Describe the bug and the steps to reproduce it

If you delete `package-lock.json`, `node_modules` folder, and run `npm install`, you will encounter errors (specifically, installing dependencies in the core package). 

![Image](https://github.com/user-attachments/assets/04953058-0a62-472d-b806-30173e5d73ef)

This should not be the case. Buggy `package-lock.json` has been impeding Vue Migration efforts (causing different vue versions, etc, issues installing new dependencies etc. and can't refresh package-lock due to this issue) . Currently, have implemented a quick fix in my own branch, but I think this needs serious investigation.

tlylt:
@gerteck did you try using some related lerna commands? I think we are not supposed to just delete and rerun npm install since we are using lerna

gerteck:
Probably due to bypassing lerna and it being a monorepo, closing this issue

Issue #2612, opened 2025-02-25
Error message on `cli`, `core` tests


gerteck:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Tell us about your environment

Win 11, Ubuntu GH Actions

### MarkBind version

5.3.3

## CLI Tests

There are error messages in the MarkBind `packages/cli` tests. See:
![Image](https://github.com/user-attachments/assets/49359d40-2b8c-4656-b3c3-58a47faad79d)

- I am running this on master branch, individually through `cd packages/ cli; npm run test`. Due to the tests running in parallel, if you run `npm run test` from root, lerna will run tests in parallel, so test messages are warped, makes debugging hard. So best to run each test locally individually.

Additionally, in the latest PR check, this 'error' also shows.
https://github.com/MarkBind/markbind/actions/runs/13458086184/job/37606471167#step:8:233

## Core Tests

There are also error messages when running `core` tests. 

![Image](https://github.com/user-attachments/assets/6bac0268-c4c8-4368-a0b5-4a173face017)

i.e. cyclic reference, does not exist. 

## Issue

This error is misleading. Is this supposed to be expected behaviour? If it is, could the error be caught and not shown? 
* I suspect this is the case for core tests.

Otherwise, is the test actually failing in some segment, but reported as pass?
* I suspect this might be the case for cli tests, hopefully not.

AgentHagu:
> Otherwise, is the test actually failing in some segment, but reported as pass?
    - I suspect this might be the case for cli tests, hopefully not.

After some investigation, the error during testing appears to be intentional. Firstly, the error message comes from the implementation of the `src` attribute for popover elements. This behavior was introduced in PR #1780. The purpose of the error was to prevent URLs like "https://www.google.com/" from being used in the `src` attribute.

The specific test code triggering the error is as follows:
```
<popover header="URLs are not valid" src="http://www.shouldnotwork.com">
  URLs should not be valid
</popover>
```

This popover deliberately triggers the error and the snapshot test expects the popover's HTML content to be replaced with an error message like this:

![Image](https://github.com/user-attachments/assets/6c99dd13-2161-4940-9786-88ffa5d70875)


I do agree that the error is misleading though. The CLI tests do not verify if the error is properly logged, so perhaps we can switch from logging this as an error to logging it as a warning instead. 

Users will still be notified and developers can ignore the warnings during testing. A PR could also be created to improve the error message since the current one is not user-friendly anyways.

[...]

PR #2622, opened 2025-03-06
Migrate from Vue 2 to Vue 3

### LINKED ISSUE #1536: Change $scopedSlots back to $slots when Vue is upgraded to v3.0
To fix the issue **"Modal content disappearing after reopening"** raised in #1519, we have changed all the `$slots` syntax to `$scopedSlots` (#1535).

However, as highlighted by @ang-zeyu [comment (#1519)](https://github.com/MarkBind/markbind/issues/1519#issuecomment-812896067), this change will have to be reverted when we upgrade Vue to v3.0 and above, since `$scopedSlots` will no longer be supported. 

We will have to find another way to resolve the issue in #1519 then should the issue of using `$slots` still persist.

### LINKED ISSUE #2084: Investigate and list what needs to be done for Vue V3 migrartion

### LINKED ISSUE #2171: Fix or simplify live reloading logic
Referring to [this line](https://github.com/jovyntls/markbind/blob/c2fc1725cd73c578342585ce82733aa5ad77715d/packages/core/src/Page/index.js#L529:~:text=.pageConfig.-,soucePath,-%5D%20%3D%20builtPage) in Page/index.js:
```javascript
    // Each source path will only contain 1 copy of build/re-build page (the latest one)
    pageVueServerRenderer.pageEntries[this.pageConfig.soucePath] = builtPage;
```

Because of a typo (sourcePath->soucePath), `pageEntries` in PageVueServerRenderer always only has one entry (with the key `undefined`). However, live reloading still seems to work: making changes to one or multiple Vue source files triggers live reload as expected. Same for md files. I haven't been able to find any unexpected behaviour caused by this line.

Instead of fixing the typo, I'm wondering if it's possible to remove this line entirely, or simplify the logic related to it?

### LINKED ISSUE #2597: Steps Needed for Vue Migration (Vue 2 -> Vue 3)
### Any related issues?

#2084 Vue Migration Investigation

Others: 
* #1702 Bootstrap 5
* #1536 scopedSlots depreciation

## Vue Migration Vue 2 to Vue 3

MarkBind (v5.5.3) is currently using Vue 2. Vue2 has reached EOL, (end-2023) and will not receive any type of support. Packages supporting Vue2 used may not be maintained by extension. This limits the extensibility and maintainability of MarkBind, especially the vue-components package. (UI Library Package).

Vue3 has TypeScript support, noticeable performance improvements and more responsive application. Also has smaller bundle size. Do note that  Vue 3 does not support IE13 (internet explorer).  Note that the migration is a time intensive process with codebase understanding required. Currently, MarkBind Vue components are authored in the Vue Options API style. While migrating to Vue 3, we should continue to use this API style. Future work can consider cost benefit of switching to using Composition API.

[Breaking changes](https://v3-migration.vuejs.org/breaking-changes/) to expect include: breaking Vue framework API changes, different dependencies used for Vue 2 vs Vue 3, SSR.

Intricate understanding of both Vue 2, Vue 3, low-level dependency APIs and MarkBind (in particular with regards to SSR, Hydrating, Bundling) is needed to conduct this. Some relevant links to understanding / reviewing the PR:
* [Vue 3 API Reference](https://vuejs.org/api/application.html#createapp)
* [Vue SFC](https://vuejs.org/guide/scaling-up/sfc.html)
* [Vue Rendering Mechanism](https://vuejs.org/guide/extras/rendering-mechanism)
* [Vue Migration Build](https://v3-migration.vuejs.org/migration-build)
* [@vue/compiler-sfc](https://www.npmjs.com/package/@vue/compiler-sfc)
* [@vue/server-renderer](https://www.npmjs.com/package/@vue/server-renderer)

Additionally, here are some relevant documentation of the codebase I have written and compiled to speed up the understanding process for purpose of peer development/review.
* [pageVueServerRenderer](https://gerteck.github.io/mb-architecture/contents/core/pageVueServerRenderer.html)
* [Core-Web Package](https://gerteck.github.io/mb-architecture/contents/core-web/core-web.html)
* [Vue-Components](https://gerteck.github.io/mb-architecture/contents/vue/vueComponents.html)

### **Compilation and SSR of Vue (Core):**

`Page/PageVueServerRenderer.ts` needs to be updated
1. [x] 'vue-template-compiler' has been replaced by '@vue/compiler-sfc', hence`compileVuePageAndCreateScript` method which compiles page into Vue Application to get the render function must be updated. For Vue 3, the way the render function is created and consumed is different. Do this without disabling SSR.
2. [x] `renderVuePage`: Method that Renders Vue page app into html string (Vue SSR). Ensure that it works during migration process. Updated to use `@vue/server-render` package for Vue3
3. [x] Ensure no regressions, do testing etc.

### **Webpack bundling (Core-Web)**
`Core-Web` library needs to be updated
1. [x] Update `core-web/src/index.js` setup() process. Ensure setup functions for mounting Vue apps are not affected. Updated to use new Vue `app instance` setup API.
2. [x] Update `vue-loader` package
3. [x] Update webpack configurations if relevant.
4. [x] Ensure bundling no regressions, (to do after refactoring vue components) 
5. [x] `VueCommonAppFactory.js`.The appFactory helps create the Vue app instances needed for each page, to set up the necessary configurations. Configure it if necessary for Vue 3, and adapt if needed for compatibility with Vue3 components

### **Vue Components SFCs**
Vue-Components Library needs to be updated
- [x] Update vue-components SFCs.
- [x] Update dependencies and librarie used
Note: Vue 2 Options API -> Vue 3 Options API should not have too many changes. Should be the most straightforward process. However, a lot of dependencies have to be updated such as `floating vue`, etc. 
- [x] Ensure no regressions as in https://github.com/MarkBind/markbind/issues/1519

### **Vue Testing**
Update all testing packages and dependencies
- [x] Update testing dependencies for Vue 3
- [x] Update tests (to be done after vue components refactoring)

### **Plugins:**
Update necessary plugins (change in Vue directive API)
1.  [x] Ensure usability and non-regression of `dataTable` plugin as Vue Directive used.
2.  [x] Ensure usability and non-regression of `Mermaid` plugins as Vue Directive used.

| Feature              | Vue 2                          | Vue 3                          |
|----------------------|--------------------------------|--------------------------------|
| Instance Creation    | `new Vue()`                    | `createApp()`                  |
| Global Configuration | On Vue constructor             | On app instance                |
| Global APIs          | Attached to Vue                 | Instance methods on app        |
| Reactivity System    | `Object.defineProperty`        | `Proxy`                        |
| Composition API      | Not available                  | Available                      |
| Fragments            | Single root element required   | Multiple root elements allowed |
| Teleport             | Not available                  | Built-in `<teleport>` support  |

* In Vue 2, plugins add global level functionality. In Vue 3, each `createApp` call creates completely isolated app instance with its own config and global properties. Due to differences, the Vue instance is setup differently.
* Migrating from `new Vue` to createSSRApp(): case study [here](https://v3.ru.vuejs.org/guide/render-function.html) on using render function
* In Vue2, static render functions explicitly separated from main render function, but in Vue3, it is not. Hence, only one render function. (`compileVuePageAndCreateScript`)
* package `vue-server-renderer` is for Vue 2.0. Corresponding functionality for Vue 3 is in `@vue/server-renderer`, which is actually a dependency of the main `vue` package.

>  Custom Directives: Vue.js 2: Custom directives are registered globally using Vue.directive. Vue.js 3: Custom directives are registered using the app.directive method, making it more modular and scoped to the app instance.

## FAQ
- What about considering letting vue2 and vue3 co-exist? 
We can use a migration build (Vue 2.7) that allows both Vue 2 & 3 code in a single codebase. Once done, migrate over to Vue 3.  However, I faced some issues with getting `@vue/compat` to work.

### **Some other sources:** 
- [Migration QA](https://www.monterail.com/blog/vue-3-migration-qa)
- [Vue2-Vue3 Differences](https://blog.shiftasia.com/difference-between-vuejs-2-and-3-when-to-use-them/)
- [Official Migration Guide](https://v3-migration.vuejs.org/migration-build.html)

_No response_

### Additional context

_No response_

### LINKED ISSUE #2610: Some Vue components tests are erroneous

### Any related issues? 

No

Environment: Win 11
MarkBind version: 5.3.3

Several vue-component tests are erroneous.

See:
https://github.com/MarkBind/markbind/blob/008aa1d36fe193e0b1fddc784b907794b1d6fa15/packages/vue-components/src/__tests__/Box.spec.js#L227-L240

Actually, only the wrapper is being compared. So the test is not really testing anything. For instance, you could even mount a different component and the test would pass. 

Image

Other things:
* If facing issues getting the jest to compile, try removing your jest cache. `npx jest --clearCache`. Faced this issue, but could be because I have been swapping configurations between Vue 2 and Vue 3.
* To run individual tests, the path (e.g. `npm test ./src/__tests__/Box.spec.js` needs to be delimited by `/` not `\`. For case of autocomplete path on windows. (Bc. of default test regex).

#### Breadcrumbs

https://github.com/MarkBind/markbind/blob/aa1fdbbc852b4b0e7f05ef1f7d228d026a5d7adb/packages/vue-components/src/__tests__/Breadcrumb.spec.js#L5-L10

On line 7 the class is actually not closed. 

#### Popovers

Some of the snapshot tests for popovers don't really seem to test functionality rather than just testing on the surface given the stubs.

gerteck:

- [x] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- Others, please explain:

### **Overview of PR:**

Fixes #2597 - Steps Needed for Vue Migration. 

MarkBind (v5.5.3) is currently using Vue 2. However, Vue 2 has reached EOL in 2023 and limits the extensibility and maintainability of MarkBind, especially the vue-components package, and external dependencies for Vue 2 components that are no longer actively maintained. (UI Library Package). This PR migrates MarkBind from Vue 2 to Vue 3. This builds on extensive efforts from #2224 - Vue Migration attempt 2 years ago by @EltonGohJH, which has helped immensely to set the foundation for this PR. 

There are no functional changes to how MarkBind behaves:
* SSR is preserved and (fully) functional
* Hydration is preserved and (fully) functional
* Vue components preserved using Options API (vs. Composition API)

### **Overview of changes:**

Some basic understanding of both Vue 2, Vue 3 and a good understanding of MarkBind (in particular with regards to SSR, Hydrating, Bundling) is needed to review? the PR. I have previously linked some hopefully helpful resources in issue #2597 to help in understanding this PR.

The Vue migration involves major changes in 3 out of 4 packages of Markbind, namely the `core`, `core-web`, and `vue-components`. Changes in `cli` are only functional snapshot tests. Changes have been separated into commits that are aim to be modular and digestable.

**Commit History Overview**
1. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/9252c201840c422659c047e88586f00fa088781e): `Adapt processing logic to use Vue 3.3.11` - vue is updated for the core library. 
    * This entails changes in API to both the [renderFunction](https://www.npmjs.com/package/@vue/compiler-sfc) generation function, and the [renderToString](https://www.npmjs.com/package/@vue/server-renderer) function. There is no documentation for the renderFunction generation function, you need to directly reference the type definitions from [vuejs source code](https://github.com/vuejs/core/blob/main/packages/compiler-core/src/options.ts) as specified by npm package description. 
    * I also update the bundled vue binary to V3.3.11 here.

2. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/2ec3b288ad0e0649e2fcb13b2775b2d6567f6897): `Migrate Vue Components` - I update the configurations for Vue Components to Vue3 (test, lint). 
    * I also update configuration dependencies in the `vue-components` package (`vue-style-loader`, `jest-env-jsdom`). 
    * Next, I update minimal syntax changes in all components (and directives). Reviewers can refer to the [vue migration guide](https://v3-migration.vuejs.org/) to look at syntax changes or API changes that are mentioned there. 
    * Naturally, vue component dependencies are also updated (`floating-vue`, `vue-final-modal`, `portal-vue`). Every component or configuration dependency updated probably has some separate form of migration guide online. 
    * Changes are largely minimal.

3. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/6bcf6dbd71e2386267458283a56509416cb6680f): `Update core-web package 1` - Update dependencies for core-web (bundling). Update configuration for webpack. 

4. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/450d78b35627604b3d4eb60652bd4956a52c97f4): `Update core-web package 2` - Update the essential webpage js logic `index.js` in core-web. 
    * Simply put, this entails using the new renderFunction that we previously generated in commit 1 and registering the vue components in commit 2. 
    * Additionally, I fixed a bug in the webpack config that I did not detect previously. (see commit message for details). 
    * Binaries for Vue components and markbind.js are regenerated.

5. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/0486230ee73c210f5b830c46c3634c3b9d64ab15): `Fix Hydration Issue for Question component` - Fixed issue that I did not detect previously, related to commit 2.

6. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/a7af11a5e604dc422c21326b6d21251325ddfc56): `Update Vue Testcases and Snapshots` - As per title. I also realized ( #2610 ), so I amended the erroneous tests. No point keeping useless tests. 
    * There are some API changes to the testing so they were updated.

7. [Commit](https://github.com/MarkBind/markbind/commit/0b93db3b0f82288d41463729243391eb4bc75c27): `Merge master`: PR updated with necessary changes from pre-requisite PRs. 
    * Basically, the PRs decouples plugins from using Vue. (They were using a workaround that works with Vue 2, but can't be done with Vue 3.) Thanks to these PRs, we don't need to worry about vue migration affecting plugins. Coincidentally, the PRs also enhance efficiency and fix other bugs identified. 

8. Update license txt file as per [MIT License](https://opensource.org/license/mit)

9. [Commit](https://github.com/MarkBind/markbind/pull/2622/commits/7ea3d6d544e7b0d3b69333f2d88279d793d67da5): `Update CLI Snapshots` - Snapshots for test generated sites are regenerated and commited. 
    * Don't be intimidated - actually there's no real change in most if not all of the html files. It's only because I updated the vue.js file name to better reflect the vue version. (which is awesome - means no change in HTML rendered!)
    *  Ignore the changes to the renderfunction binary (`*.renderjs`), it is not meaningful. I realized that the render function is unnecessarily compared. This causes a lot of issues like pollution, unnecessary increased test load. I will see if I can make a separate PR to address this issue to increase developers quality of life. Related to #1637.

10. Congratulations for getting this far. There's no other unrelated initial changes. (Further changes and bug fixes are detailed in comment history below, or resolved comments in code review)

Changelog in comment history:
> **Changelog** (Beyond initial PR changes):
> * Built the CS2103 website on gh-pages (serves updated version)
> * Update modal opening animation (update breaking syntax)
> * Update dropdown functionality (hide submenus - update breaking syntax with new logic)
> * Fix tree diagram functionality (whitespace issue - vue breaking change)
> * Fix spacing between elements (whitespace issue - vue breaking change)
> * Investigate render warning (invalid HTML in docs)
> * Fix to images not loaded, panel somehow missing (register plugins in new Vue instance, Retriever.vue)
> * Updated fix to navbar content not properly handled (update breaking vue CSS scoped syntax)
> * Update Question hydration error - (disable animations for this PR for now)
> * Add docs for how MarkBind uses Vue & Update outdated docs on vue 
> * Fix whitespace issue in site-nav / page-nav (update portal-vue breaking syntax)
> * Closeable Directive fixed (update hook)
> * Fixed Pic component on load computation of height and width
> * Fixed Thumbnail and Annotate components `@load` handling case where it does not fire.

---

### Issues Fixed

Fixes #2597 - Steps Needed for Vue Migration
Fixes #2084 - Investigate Vue Migration

Fixes #2610 - removed irrelevant ineffective tests that don't work.

Fixes #2171 - reloading logic abstracted into `savePageRenderFnForHotReload` method in `PageVueServerRenderer`, to reduce coupling from `Page/index.js`.

Fixes #1536 - scopedSlots have been replaced to slots as per Vue 3 migration

---

### Others

**Related PRs for Reference**:
https://github.com/MarkBind/markbind/pull/2596 - Unformatted Vue Migration PR (Closed)
https://github.com/MarkBind/markbind/pull/1534 - Vue 2 SSR to Resolve FOUC

**Related PRs (not needed for Reference)**:
Additionally, I have tried as best as possible to isolate any separable changes into separate PRs from this PR for ease of code review and understanding. It is not necessary to look through those changes. These PRs, which include both enhancement and code maintanence, include:
* https://github.com/MarkBind/markbind/pull/2625 - Refactor codeBlockButtonPlugins
* https://github.com/MarkBind/markbind/pull/2626 - Update dataTable
* https://github.com/MarkBind/markbind/pull/2614 - Update Mermaid

I don't think it is feasible to further break down this PR (tests would not work).

Additionally, during the migration process, I have identified some issues, such as #2610, #2612, #2613, #2615, #2623, #2627, #2629, #2630, #2631, #2624. Most of these issues are not addressed in the scope of this PR, but a review of this PR might provide context to tackle these issues separately. Also, there are definitely regressions or changes due to change in dependencies, especially for the vue components. These can also be tackled separately in the future. Appreciate the help to identify any regressions if spotted and raising appropriate issues after the Vue Migration has been merged. 

To-do:
* Update documentation on Vue related sections. From what I can see, the writing components section. However, actually I think the section is general enough that it doesn't really need much changes for correctness. Maybe I will raise a separate issue and add on to the documentation and make changes separately. idk. thoughts?

---

**Testing instructions:**

Some general tests I could think of:

1. **Run Tests**
   - `npm run test` for `core`, `cli`, and `vue-components`.

2. **Build & Serve Sample Site**
   - `markbind init`, `markbind build`, `markbind serve`.
   - Verify components render correctly and interact as expected.

3. **Check SSR & Hydration**
   - Confirm no hydration mismatch or console warnings.
   - Verify correct server-rendered HTML and client hydration.

4. **Plugin Compatibility**
   - Test various plugins to ensure migration has no side effects.

5. **Cross-browser Check**
   - Open site in Chrome, Firefox, Safari, and Edge.

---

Migrate from Vue 2 to Vue 3

Vue 2 has reached end-of-life in 2023, limiting maintainability,
extensibility of MarkBind. This commit migrates the codebase 
(core, CLI, and vue-components) to Vue 3.3.11.

The migration includes major updates to server-side rendering
logic, hydration, component syntax, bundling setup (Webpack), 
and test  configurations. Vue 2 dependencies and deprecated 
APIs have been replaced accordingly. Snapshot tests and test 
site output have also been updated to reflect new Vue version.

This migration improves future extensibility and compatibility
with modern tooling and Vue ecosystem packages.

---

- [x] No unrelated changes (Best effort) 

---

**Reviewer checklist:**

---

This release deprecates support for Vue 2 and introduces full support for Vue 3.

Summary of Breaking Changes:
Vue 2 is no longer supported. All components and internal rendering logic now use Vue 3.3.11. Vue 2-based plugins and custom components must be updated to comply with Vue 3 syntax and API. Deprecated HTML5 syntax support is removed, such as `<font>`, `<big>`, and similar tags that are no longer rendered or styled reliably.

Scoped slots API (scopedSlots) has been replaced by the new slots API as per Vue 3 standards. Internal test and SSR setup has been restructured to reflect modern Vue 3 testing libraries and bundling practices.

What Site Authors Need to Do:
If you are using custom Vue components and directives, please migrate them using the [Vue 3 Migration Guide](https://v3-migration.vuejs.org/). If your site uses deprecated HTML tags, replace them with semantic HTML5 alternatives or appropriate CSS styling. If you have extended MarkBind through plugins that rely on Vue APIs, refer to updated documentation for best practices post-migration.


gerteck:
Site is live and running on Vue 3!

Preview at: [https://deploy-preview-2622--markbind-master.netlify.app/](https://deploy-preview-2622--markbind-master.netlify.app/)

Go to the console and type: `Vue.version`. We're running on `Vue version 3.3.11`. 

Ready for review - in the meantime, will try to author a separate PR to aid with PR-review.

EltonGohJH:
Hi hi @gerteck,
As mentioned, I think we could deploy a Netlify version of CS2103T for @damithc to test.

Prof @damithc can do that as a first layer of sanity test IMO.

EltonGohJH:
Anyways, thanks @gerteck for continuing this feature. Some of the solutions to the problem I had were ingenious, to say the least.

I will try to review and comment on the core SSR flow but likely need someone to help with a in depth check on the other parts like whether all the vue components are functioning.

CC @lhw-1

tlylt:
@gerteck great work 🚀 
Just started playing around with the deployed site to check for any defects. One minor thing I noticed when comparing against https://markbind-master.netlify.app/
- The quiz's next icon animates from the top left to bottom right when clicked.

Please see video comparing the deployed preview and current master

https://github.com/user-attachments/assets/78e9d1c8-fff9-46c7-ab97-f82d75cbed55

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -3,49 +3,52 @@
...
+    filename: pageConfig.sourcePath,
+    id: pageConfig.sourcePath,
+    compilerOptions,
+  };
+  const compiled = compileTemplate(templateOptions);
+  const outputContent = compiled.code;
+
+  const scriptContent = `
+    const renderFn = new Function(${JSON.stringify(outputContent)});
+    var render = renderFn();
```
EltonGohJH, line 51: This part looks good to me. I think is a very good implementation of what we really have.

packages/vue-components/src/Retriever.vue
```diff
@@ -92,13 +93,14 @@ export default {
...
+          const TempComponent = {
             template: `<div>\n${result}\n</div>`,
             data() {
               return rootData;
             },
-          });
-          new TempComponent().$mount(this.$el);
+          };
+          const app = createApp(TempComponent);
+          app.mount(this.$el);
```
EltonGohJH, line 103: This part looks good to me.
This is needed because we only render dynamically for this component.

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -61,45 +64,60 @@ async function compileVuePageAndCreateScript(
...
+  const render = new Function('Vue', renderFn)(Vue);
+  const app = createSSRApp({
+    render,
     ...appFactory(),
   });
+  app.use(plugin);
+  const html = await renderToString(app);
+  const wrappedHtml = `<div id="app">${html}</div>`;
+  return wrappedHtml;
+}
```
EltonGohJH, line 103: LGTM  seems quite straight forward of using SSR to render

packages/core-web/src/index.js
```diff
@@ -170,44 +173,32 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
 
 function setupWithSearch() {
-  const vm = new Vue({
-    render(createElement) {
-      return pageVueRenderFn.call(this, createElement);
-    },
-    staticRenderFns: pageVueStaticRenderFns,
+  // render exist in global window object
+  // see PageVueServerRenderer.ts
+  const app = createSSRApp({
```
EltonGohJH, line 192: @gerteck 
Same as above

packages/core-web/src/index.js
```diff
@@ -170,44 +173,32 @@ window.handleSiteNavClick = function (elem, useAnchor = true) {
...
   });
-  /*
-   * For SSR, if we mount onto the wrong element (without data-server-rendered attribute) in our SSR setup,
-   * hydration will fail silently and turn into client-side rendering, which is not what we want.
-   * Thus, we will always force hydration so that we always know when hydration has failed, so that we can
-   * address the hydration issue accordingly.
-   */
-  vm.$mount('#app', true); // second parameter, 'true', enables force hydration
+  app.use(plugin);
+  app.mount('#app');
```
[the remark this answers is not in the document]
gerteck: To set some context - from what I understand in differences between `createApp()` and `createSSRApp()` is whether the app runs in `hydration mode`. Reference: [Vue docs](https://vuejs.org/guide/scaling-up/ssr.html#rendering-an-app)
> To make the client-side app interactive, Vue needs to perform the hydration step. During hydration, it creates the same Vue application that was run on the server, matches each component to the DOM nodes it should control, and attaches DOM event listeners.
> To mount an app in hydration mode, we need to use [createSSRApp()](https://vuejs.org/api/application#createssrapp) instead of createApp():

This should be called client side - `index.js` is the entry point for the markbind binary that is bundled in `core-web`. Since it is clientSide - definitely should be using `CreateSSRApp()` - as the client is also served the `*.html` that was pre-rendered, and we want the client to do hydration and check differences betweeen served `html` and DOM structure given by the render function. 

--- 

A related question is whether we should be calling `createApp()` or `createSSRApp()` when we render the `html` in the server. (that is to be sent to client). Referring to this portion in `PageVueServerRenderer`:

https://github.com/gerteck/markbind/blob/7ea3d6d544e7b0d3b69333f2d88279d793d67da5/packages/core/src/Page/PageVueServerRenderer.ts#L88-L103

Theoretically, it should be fine to use `createApp()` on server to render HTML since there's no hydration involved, but I couldn't find any information on this. Following the [serverRenderer (renderToString) npm usage docs](https://www.npmjs.com/package/@vue/server-renderer), it seems that it is supposed to be used with `createSSRApp()` so there is that.

gerteck:
> @gerteck great work 🚀 Just started playing around with the deployed site to check for any defects. One minor thing I noticed when comparing against https://markbind-master.netlify.app/
> 
> * The quiz's next icon animates from the top left to bottom right when clicked.


Thank you for your help on this PR! I also noted that [popovers](https://deploy-preview-2622--markbind-master.netlify.app/userguide/components/popups#popovers) also lost their spacing:

Picture1

It's could be because the vue component libraries were updated or some differences in the configuration syntax, since I made minimal changes enough just to satisfy any errors and warnings.

gerteck:
> Hi hi @gerteck, As mentioned, I think we could deploy a Netlify version of CS2103T for @damithc to test.

Thank you @EltonGohJH for making time to look through the PR! Would not have been able to work on this without your previous work on it. Agreed on the suggestion.

gerteck:
Currently I forked the CS2103 website and deployed it on my own gh-pages as like a quick draft to view it running on Vue 3 for now, to check for any major regressions first.

Can access it here: [https://gerteck.github.io/cs2103-website/index.html](https://gerteck.github.io/cs2103-website/index.html)
Can always verify the Vue version by running `Vue.version` in the console. Since the site build is running locally I used this MarkBind's vueMigration branch to generate the served pages.

I didn't update the Agolia plugin API Keys and stuff so it will redirect to the original site.

packages/vue-components/src/SearchbarPageItem.vue
```diff
@@ -102,7 +102,7 @@ export default {
         width: 92%;
         white-space: normal;
         vertical-align: top;
-        word-break: break-word;
+        word-break: break-all;
```
tlylt, line 105: why do we need to change this?

packages/vue-components/src/__tests__/Breadcrumb.spec.js
```diff
@@ -4,7 +4,7 @@ import Breadcrumb from '../Breadcrumb.vue';
 
 const SITENAV_START = `
 <nav id="site-nav">
-<div class="nav-component slim-scroll>
+<div class="nav-component" slim-scroll>
```
tlylt, line 7:
```suggestion
<div class="nav-component slim-scroll">
```
i think slim-scroll is a classname

packages/core/src/Page/PageVueServerRenderer.ts
```diff
@@ -3,49 +3,52 @@
...
+import type { SFCTemplateCompileOptions, CompilerOptions } from 'vue/compiler-sfc';
+
 import path from 'path';
 import fs from 'fs-extra';
-import * as VueCompiler from 'vue-template-compiler';
-import { createRenderer } from 'vue-server-renderer';
 import * as logger from '../utils/logger';
 import type { PageConfig, PageAssets } from './PageConfig';
 import type { Page } from '.';
-/* eslint-enable import/no-import-module-exports */
```
tlylt, line 14: let's keep this line? (though the explanation at line 3 is no longer up-to-date

tlylt (commented):
Some more issues I noticed on the rendered site:
- The tree plugin is no longer working:
  - Screenshot 2025-03-13 at 9 19 05 PM
- Modal opening animation is gone (before: a smooth transition to the opened modal, now: immediate opening of modal window)

damithc:
One issue I noticed: 2nd level drop-down menus do not disappear as they should

image

tlylt:
Tried serving the markbind docs locally with this branch and got this warning:
```
 v5.5.3
info: Website generation started at 6:45:41 AM
info: Building assets...
info: Assets built
info: Generating pages...
[------------------------------------------------------------] 0 / 60 pages builtunsafe attribute name: tim"
[@vue/server-renderer] Skipped rendering unsafe attribute name: tim"
unsafe attribute name: xu"
[@vue/server-renderer] Skipped rendering unsafe attribute name: xu"
[============================================================] 60 / 60 pages built
```
due to 
```
      <td align="center" valign="top" width="14.28%"><a href="http://tim-siu.github.io"><img src="https://avatars.githubusercontent.com/u/61866948?v=4?s=100" width="100px;" alt="Shuyao "Tim" Xu"/><br /><sub><b>Shuyao "Tim" Xu</b></sub></a><br /><a href="https://github.com/MarkBind/markbind/commits?author=Tim-Siu" title="Code">💻</a></td>
```

this is generated code so probably out of our control, but am wondering if we have a way to disable check for vue server-renderer just for this part..

packages/vue-components/src/SearchbarPageItem.vue
gerteck: Somewhere during the migration process I think there was a linting or style error saying that `break-word` was depreciated, so I changed it to fix it then:

I believe it was this error:
https://rocketvalidator.com/html-validation/css-the-value-break-word-is-deprecated
https://stackoverflow.com/questions/47413677/alternative-for-word-break-break-word

However I don't see it anymore if i change it back. Not sure why.

gerteck:
> One issue I noticed: 2nd level drop-down menus do not disappear as they should

Identified that the issue was due to a [breaking change in Vue 3](https://v3-migration.vuejs.org/breaking-changes/children)

I have made a non-trivial change to the logic to hide other submenu instances at the same level, shifting from the use of `$children` to maintaining a list of submenus at each level using `provide/inject` to resolve this issue.

packages/core/src/Page/PageVueServerRenderer.ts
gerteck: okay! I must have accidentally deleted it. Thanks!!

gerteck:
> Some more issues I noticed on the rendered site:
> 
> * The tree diagram functionality is no longer working:
>   
> * Modal opening animation is gone (before: a smooth transition to the opened modal, now: immediate opening of modal window)

I have looked into the tree diagram functionality: The root cause is due to changing of default handling of whitespaces from Vue 2 to Vue 3, specifically: [https://vuejs.org/api/application.html#app-config-compileroptions-whitespace](https://vuejs.org/api/application.html#app-config-compileroptions-whitespace)
* default in Vue 2 is to preserve, changed to condense in Vue 3.

A few ways we can approach this
* set default option to `preserve` for the Vue app in render function generation, rendering and mounting
* transform tree element to `<pre class="tree"/>`, as `<pre/>` use case is to preserve all whitespaces
* do both.

Will investigate modal opening animation soon alongside that of quiz button animation

packages/vue-components/src/questions/Question.vue
```diff
@@ -80,7 +80,7 @@
             Check
           </button>
           <button
-            v-if="qState.state !== 0 && questions"
+            v-show="qState.state !== 0 && questions"
```
tlylt, line 83: Just found out why the quiz is not working well...probably due to transition groups that no longer work as expected due to this change such that the elements appearance sequence is different.

You mentioned that this is to fix a hydration issue and when I tested locally by changing v-show back to v-if, it indeed triggered hydration warnings in the console. I looked into this further as we don't fetch this question from a server, so I am curious why it is not able to have a consistent state between server and client.

After a few tries, the following changes seems to be the fix:
- in QuizQuestionMixin.js
```
  inject: {
    questions: {
      default: [], // or () => [], need to check the type again
    },
```

tlylt:
> * default in Vue 2 is to preserve, changed to condense in Vue 3.

Great finding, yup i just searched the web and it seems like a known issue https://github.com/vuejs/core/issues/2431

I think if there are more stuff broken because of this, we can try adjusting the compiler option, if not we can probably fix the tree functionality, which i suppose should be doable.

gerteck:
> > * default in Vue 2 is to preserve, changed to condense in Vue 3.
> 
> Great finding, yup i just searched the web and it seems like a known issue [vuejs/core#2431](https://github.com/vuejs/core/issues/2431)
> 
> I think if there are more stuff broken because of this, we can try adjusting the compiler option, if not we can probably fix the tree functionality, which i suppose should be doable.

I just tested adjusting the compiler option, it works with just a simple tweak:
image

Result:


Though the documentation for the compilerOptions is non-existent, so would need to refer to the codebase for the options: https://github.com/vuejs/core/blob/main/packages/compiler-core/src/options.ts

tlylt:
> I just tested adjusting the compiler option, it works with just a simple tweak:

I'm ok with this change since we preserve existing behavior, we can create a separate issue to track this to explore changing it to condense

tlylt:
> transform tree element to `<pre class="tree"/>, as <pre/>` use case is to preserve all whitespaces

just a note, i now remembered that i probably intentionally not use `pre` because we allow markdown syntax, so there could be elements within a parent div, but not ideal if the parent is `pre`

packages/vue-components/src/questions/Question.vue
gerteck: Currently investigating this issue, and I can't seem to figure out the hydration error exists yet as well.

Because of how `Quiz` and `Question` both check if `questions` is defined or undefined to check if the question is in a quiz, the change to default `questions` to `[]` breaks some logic in the question component.

I will do some more investigation to either try to uncouple it or provide some workaround.

tlylt: > Because of how Quiz and Question both check if questions is defined or undefined to check if the question is in a quiz, the change to default questions to [] breaks some logic in the question component.

i see, how about changing those logic to check if the question array is of length 0? (i didn't check, just thinking out loud)

gerteck:
> The quiz's next icon animates from the top left to bottom right when clicked.
> Just found out why the quiz is not working well...probably due to transition groups that no longer work as expected due to this change such that the elements appearance sequence is different.
> You mentioned that this is to fix a hydration issue and when I tested locally by changing v-show back to v-if, it indeed triggered hydration warnings in the console. 

I think regarding the Question hydration error - I can revert the changes and let the hydration warning lapse for now.

An alternative could be to find a workaround to avoid this, though not sure how feasible nor if worth it.

Upon investigation, it could be a internal vue issue during rendering - See vue/6715 (https://github.com/vuejs/core/issues/ 6715). This seems to be patched in a later vue version (after v3.3.11). 

So we can choose to update Vue to latest version. (Currently using v3.3.11). However, I faced issues doing so, seemingly related to typescript:
image

It could be due to outdated typescript version, some typing library issue or something else. I previously tried to resolve this issue but I couldn't make enough headway as the scope easily expanded beyond what was comprehensible in a single PR. I also raised some issues which don't seem very trivial to debug. 

Hence -  I propose I revert the changes to questions, and let the hydration warning lapse for now. (The hydration warning is quite minor, maybe only causes slight rendering inefficiency in this case as it does not cause it to rerender, and everything still works.)
Then, we raise an issue to update the vue version as well as typescript errors and set it to higher priority.

gerteck:
> Tried serving the markbind docs locally with this branch and got this warning (Skipped rendering unsafe attribute name)
> this is generated code so probably out of our control, but am wondering if we have a way to disable check for vue server-renderer just for this part..

I couldn't find a way to disable the warnings for server-renderer meaningfully. However, the warning does point out a legitimate issue though — the HTML generated by the All Contributors bot has unescaped quotes in the alt attribute:
```
<img src="URL" width="100px;" alt="Shuyao "Tim" Xu"/>
```
https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/docs/about.md?plain=1#L133
The issue is occuring because the inner quotes around are not escaped (i.e., should be &quot;). Causing the html to be recognised as `alt = "Shuyao "` and `Tim` and `Xu"`. The error in the code isn't generated by vue renderer, but rather the all-contributors bot. 

Since this is an edge case, maybe instead we can open a small PR to manually escape the quotes in about.md (i.e., alt="Shuyao &quot;Tim&quot; Xu"). Also, maybe file a PR or issue upstream with the All Contributors project to raise concern on handling alt attributes with quotes.

Tim-Siu:
> > Tried serving the markbind docs locally with this branch and got this warning (Skipped rendering unsafe attribute name)
> > this is generated code so probably out of our control, but am wondering if we have a way to disable check for vue server-renderer just for this part..
> 
> I couldn't find a way to disable the warnings for server-renderer meaningfully. However, the warning does point out a legitimate issue though — the HTML generated by the All Contributors bot has unescaped quotes in the alt attribute:
> 
> ```
> Shuyao
> ```
> 
> The issue is occuring because the inner quotes around are not escaped (i.e., should be & quot;). Causing the html to be recognised as `alt = "Shuyao "` and `Tim` and `Xu"`. The error in the code isn't generated by vue renderer, but rather the all-contributors bot.
> 
> https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/docs/about.md?plain=1#L133
> 
> Since this is an edge case, maybe instead we can open a small PR to manually escape the quotes in about.md (i.e., alt="Shuyao "Tim" Xu"). Also, maybe file a PR or issue upstream with the All Contributors project to raise concern on handling alt attributes with quotes.

I will file a PR to fix this 😭

packages/vue-components/src/questions/Question.vue
gerteck: I tried that too but unfortunately I realized its quite coupled (q a bit of logic depends on question being defined or undefined)

https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/packages/vue-components/src/questions/QuizQuestionMixin.js#L44-L49

(^ here, the question immediately gets pushed into the array if the array is defined so the length would not be 0)

Also I don't think it hits at the core of the issue of the hydration error, so refactoring might not fix the hydration error - though I think refactoring would definitely not be bad, but maybe expand the PR scope too much

Tim-Siu:
> > Tried serving the markbind docs locally with this branch and got this warning (Skipped rendering unsafe attribute name)
> > this is generated code so probably out of our control, but am wondering if we have a way to disable check for vue server-renderer just for this part..
> 
> I couldn't find a way to disable the warnings for server-renderer meaningfully. However, the warning does point out a legitimate issue though — the HTML generated by the All Contributors bot has unescaped quotes in the alt attribute:
> 
> ```
> Shuyao
> ```
> 
> The issue is occuring because the inner quotes around are not escaped (i.e., should be `&quot;`). Causing the html to be recognised as `alt = "Shuyao "` and `Tim` and `Xu"`. The error in the code isn't generated by vue renderer, but rather the all-contributors bot.
> 
> https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/docs/about.md?plain=1#L133
> 
> Since this is an edge case, maybe instead we can open a small PR to manually escape the quotes in about.md (i.e., alt="Shuyao `&quot;`Tim`&quot;` Xu"). Also, maybe file a PR or issue upstream with the All Contributors project to raise concern on handling alt attributes with quotes.

Thanks for the investigation. Indeed this seems to be All Contributors bot not escaping the quote. I am not sure if we can manually modify it though. The header mentions `<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->`

gerteck:
> * Modal opening animation is gone (before: a smooth transition to the opened modal, now: immediate opening of modal window)

Thanks for spotting this! Was due to some naming changes transition in Vue and the package.
* https://v3.vue-final-modal.org/guide/properties/#transition
* https://v3-migration.vuejs.org/breaking-changes/transition

Have pushed changes

packages/vue-components/src/questions/Question.vue
tlylt: > Because of how Quiz and Question both check if questions is defined or undefined to check if the question is in a quiz, the change to default questions to [] breaks some logic in the question component, so it can't be used.

could you share what logic was broken?

tlylt:
> I think regarding the Question hydration error - I can revert the changes and let the hydration warning lapse for now.
> An alternative could be to find a workaround to avoid this, though not sure how feasible nor if worth it.

What about removing the animation altogether and create a separate issue for re-implementing that instead?

 > Thanks for the investigation. Indeed this seems to be All Contributors bot not escaping the quote. I am not sure if we can manually modify it though. The header mentions  Maybe we can have a separate issue addressing this if deemed important.

Yes given the state of the all contributor project it doesn't seem likely that they will merge a fix from us. Manual fix might be overridden when the page is auto-updated. Might need to test it out because I wonder if they really update the username on every request...If they don't, then we can manually update this html. If they do...unless we request you to change your username...we can live with this warning for now.

gerteck:
Updated Links for convenience:

* https://deploy-preview-2622--markbind-master.netlify.app/
* https://gerteck.github.io/cs2103-website/se-book-adapted/index.html
  * (https://github.com/gerteck/cs2103-website)

Aggregated Changes since previous:
* Modals animation updated
* Nested Dropdowns hide other submenus
* Trees work
* Spaces between popovers updated

tlylt:
A few more issues on mobile
- the navbar content is not properly handled
- weird white space when opening the site nav or page nav

image
image
image

tlylt:
One more issue on the 2103T website:
- Left is vue3: https://gerteck.github.io/cs2103-website/schedule/week9/topics.html
- Some images are not loaded
- The panel outer layout is somehow missing
image

packages/vue-components/src/Retriever.vue
gerteck: I realized I forgot to register the custom components and directives in this new Vue instance, causing bugs 😢

packages/vue-components/src/questions/Question.vue
gerteck: To set the context, In the current logic, the only way for the `questions` array to be defined is if the quiz was a parent component and initialized the questions array. So the quiz being undefined is used in the logic to check whether it was in a quiz or not.

image

To elaborate specifically why, the quiz would `provide` a defined array [], so if our `injected` question was defined, it must mean that the question is in a parent quiz component.

https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/packages/vue-components/src/questions/Quiz.vue#L116-L130

This affects the logic in the snippet in previous comment. If we define the default value of the questions array to be a defined, but empty array. The question is pushed into the `questions` array (through [provide/inject](https://vuejs.org/guide/components/provide-inject) by Quiz), because it thinks it is a parent Quiz component.  At the same time, it hides the question** as well. (To be shown by the quiz component as per quiz logic later)

https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/packages/vue-components/src/questions/QuizQuestionMixin.js#L44-L49

```js
 created() { 
   if (this.questions) {               // thinks it is in a quiz
     this.active = false;              // hides the question
     this.questions.push(this);   // adds this question in the array
   } 
 }, 
```

This raises some issues because some methods in `question.vue` uses the property of whether `questions` is defined to determine if it is in the quiz. I think it could be refactored to make it more clear and have proper methods as this is logical dependency is not so clear at first (or second) glance.

https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/packages/vue-components/src/questions/Question.vue#L150-L163

However, a more relevant issue is that I don't think it resolves the hydration issue. I noted that the reason why the hydration error doesn't show up is because if we set `questions:` `default: []`, as per the logic in `created()` above is because it hides the question. `active=false`. Specifically this causes the whole question to not render (`v-if=active`):

https://github.com/MarkBind/markbind/blob/ea7cf57e067f0880d470265fad2fae8a0d3eaa24/packages/vue-components/src/questions/Question.vue#L1-L3

 But actually if we set it back to `active=true` actually the hydration error still persists even if `questions` array is defined..

gerteck:
> One more issue on the 2103T website:
> 
> * Left is vue3: https://gerteck.github.io/cs2103-website/schedule/week9/topics.html
> * Some images are not loaded
> * The panel outer layout is somehow missing
> * (there is a hydration warning in the console, not sure if it's related)

Noted on the images and panels not loading/missing. I realized this was because I did not register MarkBind's custom Vue components in new Vue instances that were mounted when we dynamically load content, pertinently in `<includes>`. This caused it to not recognise custom vue components such as `<pic>`, `<panel>` in rendering, causing these issues.

To resolve this, I have globally exposed the MarkBindVuePlugin on the window object such that we can grab it and register it in `Retriever.vue` when we initialize a new Vue instance to dynamically load content, such as when we expand panels, or load through `<includes>`.

Have yet to investigate the mobile accessibility issues, will get to it when I can

gerteck:
I am investigating the changes in mobile accessibility styling - it looks to be caused by `markbind.min.css` not loading in, either not being bundled or not being accessed properly. Trying to find out what has caused this regression and how to revert it.

gerteck:
@damithc I've updated the site with the latest version of the branch: https://gerteck.github.io/cs2103-website/index.html
* I will update the site as issues are fixed https://github.com/gerteck/cs2103-website/deployments

tlylt:
@gerteck great work so far!

Could you also include some details of how we integrate Vue in our Dev Guide (under Design, perhaps a new page)? Need not explain everything, but keep at high level and assume that reader understands Vue and MarkBind. I'm looking for how we import/setup the vue app/create render function/some information about CSR SSR, so that this can help future Vue migration work and point out important places to take note of (such as "we initialize a new Vue instance to dynamically load content")
- @EltonGohJH can probably help with reviewing the docs 🙏 

I won't be available until April so I'm going to give my blessing now and support for this PR to be merged once we resolved all critical issues 🚀 🚀

damithc:
> I won't be available until April so I'm going to give my blessing now and support for this PR to be merged once we resolved all critical issues 🚀 🚀

Thanks for your inputs so far, @tlylt 💯

gerteck:
I have added a new page detailing how MarkBind integrates Vue (SSR, CSR, Hydration, Execution Flow).

* https://deploy-preview-2622--markbind-master.netlify.app/devguide/design/vueintegration

I think it would be useful for reviewers as well, so commenting it here to highlight it.

gerteck:
I have investigated and identified the cause of regression for site-nav and page-nav whitespace. 

This was due to a change in `portal-vue` rendering mechanism, specifically [portal-target no longer rendering a root element](https://portal-vue.linusb.org/guide/migration.html#portaltarget-component)
* A high-level explanation - Due to this expectation, the style attribute accounting for the `--sticky-header-height` was applied twice, causing this whitespace.
* Fix: applied a parent div to `portal-target` to 'absorb' this styling to maintain the expectation as per `vue-portal` v2. We can consider refactoring this in the future.  However, note that the change to refactor is not too trivial and requires some significant understanding of current styling in `main.css`, with regards to `#page-nav` and `#site-nav`, and how it interacts with `Overlay`, `OverlaySource`, and `portal-vue`.

damithc:
> I have added a new page detailing how MarkBind integrates Vue (SSR, CSR, Hydration, Execution Flow).
> 
> * https://deploy-preview-2622--markbind-master.netlify.app/devguide/design/vueintegration
> 
> I think it would be useful for reviewers as well, so commenting it here to highlight it.

@gerteck Thanks for adding to dev docs. Better dev docs will help future devs, for sure.

Given below are the two versions of the textbook PDF files. This is not exactly the result of 'save as PDF' but a result of a Python script that I used to generate the PDF and add a ToC and an index.

[textbook-vue2.pdf](https://github.com/user-attachments/files/19401955/textbook-vue2.pdf)
[textbook-vue3.pdf](https://github.com/user-attachments/files/19401956/textbook-vue3.pdf)

By the way, your staged CS2103 site says `2020 Aug-Dec`. Perhaps you are using the `master` branch rather than the `cs2103` branch of the website repo? But it doesn't matter much as the differences are minor.
image

gerteck:
I initially updated the directive's `bind()` hook to `beforeMount()` based on the [official Vue migration guide](https://v3-migration.vuejs.org/breaking-changes/custom-directives.html). However, this change caused the directive to stop working correctly — since the DOM mutations need to happen after the element is mounted.

I've now fixed this by updating the hook from beforeMount() to mounted(), ensuring all DOM manipulations and event bindings happen after the element is fully available in the DOM.

The `closeable` directive works now: https://deploy-preview-2622--markbind-master.netlify.app/userguide/reusingcontents#allowing-users-to-remove-some-contents

gerteck:
---
Links for convenience:

* Vue 2 MarkBind Docs: https://markbind-master.netlify.app/
* Vue 2 CS2103 Site https://nus-cs2103-ay2425s2.github.io/website/index.html

* Vue 3 MarkBind Docs:  https://deploy-preview-2622--markbind-master.netlify.app/
* Vue 3 CS2103 Site: https://gerteck.github.io/cs2103-website/se-book-adapted/index.html
  * (https://github.com/gerteck/cs2103-website)

---
Edit: I am including this for easier reference in latest comment

**Changelog** (Beyond initial PR changes):
* Built the CS2103 website on gh-pages (serves updated version)
* Update modal opening animation (update breaking syntax)
* Update dropdown functionality (hide submenus - update breaking syntax with new logic)
* Fix tree diagram functionality (whitespace issue - vue breaking change)
* Fix spacing between elements (whitespace issue - vue breaking change)
* Investigate render warning (invalid HTML in docs)
* Fix to images not loaded, panel somehow missing (register plugins in new Vue instance, Retriever.vue)
* Updated fix to navbar content not properly handled (update breaking vue CSS scoped syntax)
* Update Question hydration error - (disable animations for this PR for now)
* Add docs for how MarkBind uses Vue & Update outdated docs on vue 
* Fix whitespace issue in site-nav / page-nav (update portal-vue breaking syntax)
* Closeable Directive fixed (update hook)

**To Do**:
* Investigate hydration errors in served CS2103 site.

**Issues outside scope of this PR**:
* Re-enable animations for Question
* Refactor styling for site-nav, page-nav, overlaysource
* Refactor closable directive (issue has been made)

gerteck:
I have found some panels not loading:

image

However, looking into the cs2103 website code, I am not too sure how the panel is including the src:

Picture1

Could I be setting up the website wrongly?

damithc:
> I don't think i understand how `unit-inElsewhere` works, so I can't really debug it

@gerteck This is a case of using a feature called `boilerplate` (that feature saves us from having to duplicate a boilerplate file in multiple folders). The `unit-inElsewhere-asFlat.md` is actually in the `_markbind\boilerplates` folder but MarkBind is supposed to act as if the file is in the the folder specified in front of it (in this case, `../../../uml/sequenceDiagrams/parallelPaths`).

Incogdino:
image

The images appear to be loading at max width on the vue 3 site as compared to the 2103 site

lhw-1:
Ah, same observation as @Incogdino. Left is on vue 2, right is on vue 3.

image

lhw-1:
I've taken a closer look at both of the CS2103T site deployments, and apart from the image size issue and the previously mentioned site-nav and page-nav whitespaces (which we can tackle outside of this PR), there does not seem to be major issues.

As discussed with @damithc, we're pushing for this to be merged in within the next 2 days, after which we will release MarkBind v6.0.0. @MarkBind/active-devs let's proceed with the final round of checks and testing, and @gerteck do let us know your plans for this PR if there are any remaining.

gerteck:
> I've taken a closer look at both of the CS2103T site deployments, and apart from the image size issue and the previously mentioned site-nav and page-nav whitespaces (which we can tackle outside of this PR, as per [comment](https://github.com/MarkBind/markbind/pull/2622#issuecomment-2744983002)), there does not seem to be major issues.
> 
> As discussed with @damithc, we're pushing for this to be merged in within the next 2 days, after which we will release MarkBind v6.0.0. @MarkBind/active-devs let's proceed with the final round of checks and testing, and @gerteck do let us know your plans for this PR if there are any remaining.

Thank you for your help!

I am looking into the image size issue, I have isolated and identified the regression, and can push a fix in the next hour or so.

Otherwise, my plan would be that any other changes. if not major or functionality breaking can be left to be fixed after the PR has been merged.

gerteck:
pic

The pictures did not resize and defaulted to max size was because the method to set height and width was not firing, which were set to fire on `load` event. It did not fire on Vue 3 hydration, but did so for Vue 2.

The reason the load event fired in Vue 2 but not in Vue 3 is likely due to a difference in hydration behavior between Vue 2 and Vue 3 during SSR (Server-Side Rendering). In short, it seems like:
* Vue 2: load event fired after hydration of DOM.
* Vue 3: likely the onload event is fired before Vue runs and hydrates the DOM.

This seems to be a somewhat known (?) issue through search.

It could be that that when the component is hydrated in Vue 3, the image is already in the DOM and is marked as loaded and not fire `load` event for performance reasons. Or that Vue 2 retriggers the load event etc. I don’t know the exact reason.

[Load Event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)

What I’ve done is to add a `onMount` routine to run the height and width calculation and setting if the image is already loaded, to be sure that it runs in any case.

It also seems to be used in `Annotate`, `Thumbnail`, let me check it and push changes if needed as well

IanCheah:
Hi @gerteck, I have noticed a little difference under Admin Info/Ip/Week2.
Vue 3: [https://gerteck.github.io/cs2103-website/admin/ip-w2.html](https://gerteck.github.io/cs2103-website/admin/ip-w2.html)
Vue 2: [https://nus-cs2103-ay2425s2.github.io/website/admin/ip-w2.html](https://nus-cs2103-ay2425s2.github.io/website/admin/ip-w2.html)
image
On the left is Vue3, right is Vue2. There seem to be a little issue with the rendering of the mermaid diagram.
image

gerteck:
> Hi @gerteck, I have noticed a little differences under Admin Info/Ip/Week2.

Thanks @IanCheah for your help! I will have to investigate the tabs issue. 

Regarding the mermaid rendering, it seems related to CORS and seems to be quite flaky, it could have been a regression from my previous PR shifting mermaid to load on demand. I can only replicate the issue sometimes. It seems that some investigate on that is due as well, though I suppose if we support offline loading and bundle it together it should totally resolve the issue, which is related to #2548

Should I work on it now, or should we postpone it till after the migration, and set up a separate issue to tackle this?

IanCheah:
> I think we should postpone this issue till after the migration, and set up a separate issue to tackle this? Otherwise, any more commits will continue to balloon this PR. What do you think?

Yep I agree! Adding more commits may balloon the pr and even potentially introduce unforseen errors. We can set up a separate issue once the pr is merged.

lhw-1 (approved):
Thanks for all your work, @gerteck! 

This has no doubt been a massive undertaking on your part, and your efforts in documenting all the steps are greatly appreciated - I'm sure future devs will benefit from your efforts. Thanks to @EltonGohJH who started this effort 2 years ago, as well as to all the devs who helped to review and test this PR as well (@tlylt, @Tim-Siu, @Incogdino, @IanCheah, and of course @damithc)!

LGTM, and if there are no other updates, we can proceed with the planned release tonight :+1:

damithc:
> I think we should postpone this issue till after the migration, and set up a separate issue to tackle this? Otherwise, any more commits will continue to balloon this PR. What do you think?

Yup, that should be fine. But it needs to be fixed before releasing V6, as it breaks current usages.

damithc:
> This has no doubt been a massive undertaking on your part, and your efforts in documenting all the steps are greatly appreciated - I'm sure future devs will benefit from your efforts. Thanks to @EltonGohJH who started this effort 2 years ago, as well as to all the devs who helped to review and test this PR as well (@tlylt, @Tim-Siu, @Incogdino, @IanCheah,

I second that 💯 👍

PR #2625, opened 2025-03-08
Refactor codeBlockButton Plugins

### LINKED ISSUE #2624: codeBlockCopyButton and codeBlockWrapButtons Plugins Side Effects

### Any related issues?

Previous PRs on these Plugins for reference: #1043 #1668

Win 11

### MarkBind version

5.5.3

The codeBlockCopyButtons.ts and codeBlockWrapButtons.ts plugins introduce unwanted side effects.

See:

https://github.com/MarkBind/markbind/blob/2cadb784fb53f6135ddd070056190142b50dc6d4/packages/core/src/plugins/codeBlockCopyButtons.ts#L67-L76

https://github.com/MarkBind/markbind/blob/2cadb784fb53f6135ddd070056190142b50dc6d4/packages/core/src/plugins/codeBlockWrapButtons.ts#L51-L60

Both plugins process all `<pre>` nodes and insert copy and wrap buttons to `<pre>` elements if enabled, no matter if it contains `<code>` or not. 

For some context, note that `markdown-it` follows the [CommonMark spec](https://spec.commonmark.org/0.31.2/#fenced-code-blocks), and processes triple backticks into (''') into `<pre/>` and `<code>`.  For example, in the first code block below, this code snippet that we might  write in markdown file, gets converted to the following in html on the web page (second code block):

```md
'''js
console.log("Hello, world!");
'''
```
⬇ 
```html
<pre><code class="language-js">console.log("Hello, world!");</code></pre>
```
We should minimize the side effect by only appending/inserting these code-copy or code-wrap buttons if we find `<code>` in the `<pre>` element. 

Why is this not just a trivial issue? The [`<pre>` (preformatted text)](https://www.w3schools.com/TAgs/tag_pre.asp) element is used to display text exactly as is written in HTML code, preserving spaces, line breaks and indentation. If any user wants to use this element, as per `HTML` specifications, they will incur this side effect if they have the plugin enabled. (- MarkBind should support `HTML`)

This also may have other unexpected effects on other components that use `<pre>` to format their components. This includes - `mermaid` [example](https://mermaid.js.org/config/usage.html#simple-full-example), where the mermaid diagrams are formatted in `<pre/>` elements for post-processing by mermaid js. Having the plugin enabled causes the buttons to be inserted in the element which should only contain the diagram code, causing the mermaid diagrams to fail to render because of the added syntax inside the block.

### Steps to Reproduce
Add to your site.json
```
  "plugins": [
    "codeBlockWrapButtons",
    "codeBlockCopyButtons"
  ]
```

We can test a negative impact of this side effect, try adding some ASCII Art as so:

```
<pre>
   /\_/\
  ( o.o )
   > ^ <
</pre>
```

Image

Both copy and wrap buttons will show up beside the cat. (They should not show up since the cat is not code, the buttons should only show up for code blocks)

### Others

https://github.com/MarkBind/markbind/blob/2cadb784fb53f6135ddd070056190142b50dc6d4/packages/core-web/src/styles/markbind.css#L288-L307

PRs #1043 and #1668 also add direct changes to `markbind.css`, which also have similar negative side effects (on mermaid, and possibly for future implementation). I propose to 1. shift the css changes into the plugins instead of adding it globally, and 2. better scope the css rules to minimize side effects:

This was also mentioned in the PR, I think it was a good catch which would have prevented this portion of the issue:

https://github.com/MarkBind/markbind/pull/1043#pullrequestreview-364223292

> Just a follow-up thought: It feels kinda odd that we add plugin styles to the main markbind.css file. Perhaps we should come up with a better way for plugin writers to keep assets contained within the plugin folder itself.

We can do this by using cheerio to add the style changes through head, or link it as an additional css file, instead of directly adding to the markbind css. This would be better for maintenance in the future as the css would be located together with the plugin.

Some overall refactoring would be good.
 

### Expected behavior

-

### Anything else?

- [HTML `<pre>` Tag](https://www.w3schools.com/TAgs/tag_pre.asp)
- [CommonMark spec for ``` (backticks fenced code blocks))](https://spec.commonmark.org/0.31.2/#fenced-code-blocks)

gerteck:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Refactored logic to insert buttons only if pre element contains the code element. Scoped CSS rules (specifically, for `<pre/>` element to minimize side effects. Modularized CSS rules to be included only if plugin is enabled.

Refer to https://github.com/MarkBind/markbind/pull/1043 https://github.com/MarkBind/markbind/pull/1668 to check CSS rules that were added for the code block buttons.

Fixes #2624

**Testing instructions:**

See #2624 

Enable relevant plugins, and put some sample code into index.md:

```
## Test

[`][`][`]js 
console.log('hello world');
[`][`][`]

<pre>
   /\_/\
  ( o.o )
   > ^ <
</pre>
```

Buttons should not show up beside the cat. It should only show up for the code block above it.

Refactor codeBlockButton Plugins

Refactored logic to insert plugin buttons only if
pre element contains the code element as per CommonMark
spec for fenced code blocks. Scoped CSS rules for pre element 
to minimize side effects. Modularized previously global button 
CSS rules to be included only if plugin is enabled.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core/src/plugins/codeBlockCopyButtons.ts
+++ b/packages/core/src/plugins/codeBlockCopyButtons.ts
@@ -7,6 +7,8 @@ import {
 } from './codeBlockButtonsAssets/codeBlockButtonsContainer';
 import type { PluginContext } from './Plugin';
 
+const CSS_FILE_NAME = 'codeBlockButtonsAssets/codeBlockButtons.css';
+
 const COPY_ICON = `
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
   width="18" height="18" viewBox="0 0 18 18" version="1.1">
@@ -65,10 +67,13 @@ const copyCodeBlockScript = `<script>
     </script>`;
 
 export = {
+  getLinks: () => [`<link rel="stylesheet" href="${CSS_FILE_NAME}">`],
   getScripts: () => [copyCodeBlockScript],
   processNode: (_: PluginContext, node: MbNode) => {
-    if (node.name === 'pre' && !doesFunctionBtnContainerExistInNode(node)) {
-      cheerio(node).append(CONTAINER_HTML);
+    if (node.name === 'pre' && node.children?.some(child => child.name === 'code')) {
+      if (!doesFunctionBtnContainerExistInNode(node)) {
+        cheerio(node).append(CONTAINER_HTML);
+      }
     } else if (isFunctionBtnContainer(node)) {
       cheerio(node).append(getButtonHTML());
     }
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -284,57 +284,6 @@ li.footnote-item:target {
     outline: none;
 }
 
-/* "Copy" and "wrap text" code block buttons */
-pre {
-    position: relative;
-    display: flex;
-}
-
-button.function-btn {
-    border: 1px solid #d7dadc;
-    border-radius: 5px;
-    color: darkgray;
-    cursor: pointer;
-    display: none;
-    height: 30px;
-    margin: 0.17rem;
-    padding: 0.35rem;
-    position: absolute;
-    right: 0;
-    width: 30px;
-    text-align: center;
-    white-space: nowrap;
-}
-
-button.function-btn + button.function-btn {
-    right: 36px;
-}
-
-pre:hover > .function-btn-container > button.function-btn {
-    display: block;
-}
-
-.function-btn:hover {
-    transition: all 0.5s ease;
-    color: gray;
-}
-
-.function-btn-body {
-    align-items: center;
-    display: flex;
-}
-
-.function-btn svg {
-    fill: currentcolor;
-}
-
-/* Wrap class used for "wrap text" button. */
-code.wrap {
-    white-space: pre-wrap;
-    word-wrap: break-word;
-    word-break: break-all;
-}
-
 /* Octicon sizing */
 .octicon {
     height: 1em;
--- a/packages/core/src/plugins/codeBlockButtonsAssets/codeBlockButtons.css
+++ b/packages/core/src/plugins/codeBlockButtonsAssets/codeBlockButtons.css
@@ -1,0 +1,50 @@
+/* "Copy" and "wrap text" code block buttons */
+pre:has(.function-btn-container) {
+    position: relative;
+    display: flex;
+}
+
+button.function-btn {
+    border: 1px solid #d7dadc;
+    border-radius: 5px;
+    color: darkgray;
+    cursor: pointer;
+    display: none;
+    height: 30px;
+    margin: 0.17rem;
+    padding: 0.35rem;
+    position: absolute;
+    right: 0;
+    width: 30px;
+    text-align: center;
+    white-space: nowrap;
+}
+
+button.function-btn + button.function-btn {
+    right: 36px;
+}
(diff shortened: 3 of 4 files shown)


packages/core/src/plugins/codeBlockCopyButtons.ts
```diff
@@ -65,10 +67,13 @@ const copyCodeBlockScript = `<script>
     </script>`;
 
 export = {
+  getLinks: () => [`<link rel="stylesheet" href="${CSS_FILE_NAME}">`],
   getScripts: () => [copyCodeBlockScript],
   processNode: (_: PluginContext, node: MbNode) => {
-    if (node.name === 'pre' && !doesFunctionBtnContainerExistInNode(node)) {
-      cheerio(node).append(CONTAINER_HTML);
+    if (node.name === 'pre' && node.children?.some(child => child.name === 'code')) {
```
AgentHagu, line 73: I'm not familiar with this part of the codebase but is it possible for the following case to occur for a node:
- `node.name === 'pre'` is True
- the node has children with `name === 'code'`
- `!doesFunctionBtnContainerExistInNode(node)` is False
- `isFunctionBtnContainer(node)` is True

If it's possible, there might be a slight issue in the logic. In the old code, this scenario will result in:
1. The `if` condition is False, so it moves on to the `else-if`
2. The `else-if` condition is True, so it continues with `cheerio(node).append(getButtonHTML());`

However, in the new code, the same scenario will instead:
1. The `if` condition is True, so it moves on into the inner code
2. The inner `if` condition is False, so it doesn't continue
3. The `else-if` condition is never checked since the `if` condition already passed.

The new code would not result in `cheerio(node).append(getButtonHTML());` being executed. Will this be a possible issue? Or is such a scenario impossible to begin with?

gerteck: Thank you for raising this up!

> node.name === 'pre' is True
the node has children with name === 'code'
!doesFunctionBtnContainerExistInNode(node) is False
isFunctionBtnContainer(node) is True

This scenario wouldn't occur because if the node is a `<pre/>`  element, it shouldn't be a functionButtonContainer. 

This is because

```
export const CONTAINER_HTML = '<div class="function-btn-container"></div>';
```

The function button container is a `div` element.

AgentHagu: I see, thanks for the clarification!

Tim-Siu (approved):
This PR successfully addresses the problems spotted in #2624. It also refactors the code block button CSS by moving it from the general stylesheet (`markbind.css`) to a dedicated CSS file (`codeBlockButtons.css`). 

### Improvements:
1. Addressed #2624 through Improved selector: Using `:has(.function-btn-container)` is more specific than the previous selector. This will make sure code block plugin does not affect other plugins that uses `<pre>`.
2. Better organization: Moving CSS to a dedicated file improves modularity (and potentially performance)

### Suggestions:
Actually don't have any suggestion. 

The improved conditional check `node.children?.some(child => child.name === 'code')` is good. I can't think of any corner case.

Thanks @AgentHagu for carefully reviewing the code and @gerteck for detailed explanation.

The changes look good!

lhw-1 (approved):
LGTM. The changes make sense as per #2624 and possible concerns were addressed by @AgentHagu and @gerteck. Thanks @gerteck for the PR!

PR #2626, opened 2025-03-08
Refactor dataTable plugin to decouple from Vue

gerteck:

- [ ] Documentation update
- [ ] Bug fix
- [x] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**

Decouple dataTable from Vue by using MutationObserver instead of relying on Vue Directives to process dynamically loaded content.

Previously, dataTable is processed through Vue by adding it as a vue directive. However, this approach was not entirely correct, as the Vue directive could only be registered after mounting (as it is added as a script). This approach was used to account for dynamically added content such as includes, and panels (which are dynamically added through defined Vue components). This would also result in a Vue warning as so:

image

This is because when the content is hydrated, the vue directive would not have been registered yet, and Vue would not recognise `v-datatable` as a registered directive. (note that this Vue warning only shows in developmental mode, i.e. `markbind serve -d`, and does not show up on production vue, i.e. markbind website, or `markbind serve`.) 

This becomes a further issue when migrating Vue, as Vue is no longer a global object where we can register the vue directive. Instead, this PR proposes the use of [MutationObservers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) which observes changes to the DOM (e.g., added/removed nodes, attribute changes) and executes a callback when a change is detected (i.e. run datatable or any other external js plugin file).

My only concern is that shifting from Vue Directives to MutationObserver could introduce some overhead or inefficiency as Vue directives could possibly be more optimized for vue components, as it has access to Vue's reactivity system and can trigger updates directly when data changes. Not sure. Perhap MutationObserver might be more efficient as well.

However, I think that using MutationObserver will make it easier to reason about the logic of how the plugins are applied.

**Other changes**:
Also updated tests as `<span>` was wrapping `<table>` causing hydration issues.

Snapshots tests updated. Additionally, unit test testing datatable has been updated (conditional checking for  vue directive attribute removed) .

Partially addresses #2627

Note: this approach will also be adopted by the mermaid plugin, and I will make a separate PR utilizing a similar approach for the plugin.

**Testing instructions:**

- `cd .\packages\cli\test\functional\test_site_table_plugin\ `
- `markbind serve -d`

Refactor dataTable plugin

Decouple dataTable from Vue by using MutationObserver instead 
of relying on Vue Directives to process dynamically loaded content.

---

- [x] No unrelated changes 

---

**Reviewer checklist:**


Tim-Siu:
Thank you for spotting the vue hydration issue. It indeed resolved the "[Vue warn]"

I can't really remember why I put there a <span> haha. I now see that this creates a mismatch between server and client templates?

packages/cli/test/functional/test_site_table_plugin/index.md
```diff
@@ -48,14 +48,14 @@ Test sortable with searchable
 ---
 Test table can be included
 
-<span><include src="others/tableToBeIncluded.md" /></span>
+<include src="others/tableToBeIncluded.md" />
```
Tim-Siu, line 51: Great! This fixed the warning

packages/core/src/plugins/dataTable.ts
```diff
@@ -27,20 +27,72 @@ const initScript = `
...
+                   $(node).hasClass('sortable-searchable-table'))) {
+                newTables.push(node);
+              }
+              
+              if (node.nodeType === 1 && node.querySelectorAll) {
+                const tablesInNode = node.querySelectorAll(
+                'table.sortable-table:not(.dt-processed), ' +
+                'table.searchable-table:not(.dt-processed), ' +
+                'table.sortable-searchable-table:not(.dt-processed)');
+                if (tablesInNode.length) {
```
Tim-Siu, line 60: Smart use of :not(.dt-processed) pseudo-class prevents redundant initialization attempts. This avoids potential memory leaks from multiple DataTable instances on the same DOM element.

Tim-Siu (approved):
This PR decouples the dataTable plugin from Vue by replacing Vue directives with MutationObserver to detect and process dynamically loaded tables. 

The change addresses:
- Vue warnings caused by directives not being registered before content hydration
- Future compatibility issues with Vue migration (where Vue is no longer global)
- Proper initialization of tables in dynamically added content (includes, panels, modals)

The implementation uses MutationObserver to watch DOM changes and initialize tables as they're added, applying appropriate DataTable processing and marking them with a `.dt-processed` class.

Test fixes correctly remove unnecessary `<span>` wrappers around `<include>` tags.

This looks like a solid improvement that will make the plugin more maintainable and fix the current hydration warnings.

LGTM 👍

gerteck:
> Thank you for spotting the vue hydration issue. It indeed resolved the "[Vue warn]"
> 
> I can't really remember why I put there a haha. I now see that this creates a mismatch between server and client templates?

yep! I think the issue is that the [`<span>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) can only contain permitted [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) (text and markup only), so when the web browser receives the rendered HTML it does some reordering of some sort to reorder the invalid HTML into something valid which in turn causes the mismatch between rendered DOM and received HTML. 

I also just pushed the update to the snapshot files (to account for `<span>` removal) and the unit test (removed check for v-dataTable Vue directive) for the DataTables. Testing for `dt-processed` in the unit tests and rendering of dataTables might be more complex as that requires integration testing with the dataTable js file.

Thanks @Tim-Siu for your speedy reply!!

Tim-Siu:
I'm comfortable omitting the dataTables rendering checks from unit tests since they weren't previously included. For testing the dt-processed class, if it is too complex to implement, we can skip it. The integration tests in the PR (might) already validate this functionality.

lhw-1 (approved):
I've read through the review history and comments; LGTM. Thanks @gerteck!

PR #2648, opened 2025-03-16
Add card stack component

### LINKED ISSUE #2553: Provide a search feature for icons
MarkBind supports including icons from a variety of sources (Fontawesome, Octicons, Bootstrap, Material, Glypyicons). It would be good to provide a unified search feature that allows users to search for icons based on keywords, so that they can easily locate suitable icons irrespective of which source they come from.

Incogdino:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Closes #2553

**Overview of changes:**
Add CardStack Component

**Testing instructions:**

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core/src/html/MdAttributeRenderer.ts
+++ b/packages/core/src/html/MdAttributeRenderer.ts
@@ -161,6 +161,13 @@ export class MdAttributeRenderer {
   }
 
   /*
+   * Card Stack
+   */
+  processCardAttributes(node: MbNode) {
+    this.processSlotAttribute(node, 'header', false);
+  }
+
+  /*
    * Dropdowns
    */
 
--- a/packages/core/src/html/NodeProcessor.ts
+++ b/packages/core/src/html/NodeProcessor.ts
@@ -212,6 +212,9 @@ export class NodeProcessor {
       case 'tooltip':
         this.mdAttributeRenderer.processTooltip(node);
         break;
+      case 'card':
+        this.mdAttributeRenderer.processCardAttributes(node);
+        break;
       case 'modal':
         this.processModal(node);
         break;
--- a/packages/vue-components/src/__tests__/CardStack.spec.js
+++ b/packages/vue-components/src/__tests__/CardStack.spec.js
@@ -1,0 +1,90 @@
+import { mount } from '@vue/test-utils';
+import CardStack from '../cardstack/CardStack.vue';
+import Card from '../cardstack/Card.vue';
+
+const DEFAULT_STUBS = {
+  card: Card,
+};
+
+const CARDS_CUSTOMISATION = `
+  <card header="Normal Body" tag="Normal" keywords="Body" disabled>
+      Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor 
+      incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud 
+      exercitation ullamco laboris nisi ut aliquip ex ea commodo
+  </card>
+  <card header="Short Body" tag="Short" keywords="Body">
+      Lorem ipsum dolor sit amet, consectetur adipisicing
+  </card>
+`;
+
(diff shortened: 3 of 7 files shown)


gerteck:
Play around with it here:

* https://deploy-preview-2648--markbind-master.netlify.app/userguide/components/presentation#cardstacks

AgentHagu:
There is some weird behavior with the searchbar of the card stack. It seems to only search using the last word in the search bar. For example, consider this sequence of searches:
1. Basic
2. Basic card
3. Basic card question

The expected behavior depends on whether the searchbar should be an OR or an AND search but either way, the actual behavior doesn't seem to follow either. The actual behavior seems to be that the searchbar only uses the last word, so for 1, 2 and 3, it searches "basic", "card" and "question" respectively. Is this intended behavior? Screenshots below:

image
image
image

lhw-1:
In terms of the tags, the current implementation seems to be an OR implementation; I am thinking if AND is more intuitive. Perhaps we could also provide a way to choose between the two behaviors? And as a nit, in terms of the design, adding some space between the "Tags:" and the actual tags would be nice.

That aside, as @AgentHagu pointed out, the behavior of the search bar is rather unintuitive. Filtering based on the last keyword seems to be the current behavior, which is rather unexpected. 

E.g. this would not be the expected result for users, I would imagine:

image

@Incogdino What are your thoughts?

Incogdino:
> In terms of the tags, the current implementation seems to be an OR implementation; I am thinking if AND is more intuitive. Perhaps we could also provide a way to choose between the two behaviors?
> 
I discussed this with prof @damithc the other day during lecture and the consensus was that an OR search would better fit the purpose here since users are most likely going to filter according to what they want to appear in the card stack. As for whether we should allow users to choose between AND/ OR tag search, it is a good idea and I think we should have it too but it might lead to some inconsistencies in the final site when multiple card stacks are involved in the application and users toggle between them. 

>  And as a nit, in terms of the design, adding some space between the "Tags:" and the actual tags would be nice.

Will work on this!

> That aside, as @AgentHagu pointed out, the behavior of the search bar is rather unintuitive. Filtering based on the last keyword seems to be the current behavior, which is rather unexpected.

I have fixed this issue in the latest commit.

damithc:
> As for whether we should allow users to choose between AND/ OR tag search, it is a good idea and I think we should have it too but it might lead to some inconsistencies in the final site when multiple card stacks are involved in the application and users toggle between them.

This can be added as a nice-to-have feature, at a later time. The option to toggle between the two modes can be provided near the tick boxes (see below), so that it applies to the specific stack, not site-wide?
image

Incogdino:
> I see that the latest commit is updating this feature to Vue 3 - is the PR ready for review?

Let me update the docs to include the icons that markbind supports (for #2553)

Incogdino:
> I see that the latest commit is updating this feature to Vue 3 - is the PR ready for review?

Could we review this pr for now and leave #2553 on hold in the meantime. I am trying to find a way to retireve all the icon names without having to manually copy and paste them from the source itself.

If anyone has any ideas, please do let me know!

lhw-1:
I see, so we possibly merge the feature in first and leave the actual implementation on the MarkBind documentation to another PR later on. Sure, let's do that and keep this PR to introducing the new feature only.

lhw-1 (approved):
LGTM, the feature works well; I couldn't find any regressions. Features like selecting between AND / OR, as well as using this for icons, can be followed-up in later PRs. Thanks @Incogdino!

gerteck (approved):
Looks good to me too!

damithc:
Finally managed to use this feature in a production site https://nus-cs2103-ay2526-s2.github.io/website/admin/appendixH-tipsFromStudents.html

Thanks again for adding the initial version of the Card Stack feature @Incogdino 
Thanks to all others who pitched in too.

PR #2649, opened 2025-03-16
Add color for highlighting fenced code blocks

### LINKED ISSUE #2594: Allow users to specify highlight colour in inline code

### Any related issues?

#2051

Reader Usability

I noticed that when I highlight text within a fenced block, the highlight color is black or grey depending on whether we are in light or dark mode. It can be frustrating because the highlight doesn't stand out very well.

For example in light mode, I think the highlight in orange stands out better as compared to the default highlight color

Image

Image

I would like to propose allowing users to specify an additional field when typing out inline code. In the example below, the color field newly added.

```

java'''{.line-numbers highlight-lines="2" color="Orange"}
public class RandomNumberGenerator {
    private int min;
    private int max;

    public RandomNumberGenerator(int min, int max) {
        this.min = min;
        this.max = max;
    }

    public int generate() {
        return min + (int) (Math.random() * (max - min + 1));
    }

    public static void main(String[] args) {
        RandomNumberGenerator rng = new RandomNumberGenerator(1, 100);
        System.out.println("Random Number: " + rng.generate());
    }
}
'''
```

N.A.

### Additional context

_No response_

IanCheah:

- [ ] Documentation update
- [ ] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Fixes #2594 

**Overview of changes:**
#### The new highlight feature
Now users are able to specify the color at each section they want to highlight by adding a `@` followed by a color.
Previously, users will specify an inline code highlight as shown below. Depending on the code theme, the highlight will be default to `#e6e6fa` for the light theme and `#000000` for the dark theme.
`java {.line-numbers highlight-lines="1[:],3['Inventory'],3[4::6],4['It\'s designed'],5,6[8:15],6[18:],8[0::2],12[:]-14,16-18,20[12:]-22,24[1::]-26"}`

With the new update, users can specify a color they want at each section. For example,
`java {.line-numbers highlight-lines="1[:]@pink,3['Inventory']@blue,3[4::6]@orange,4['It\'s designed']@#000000,5@rgb(120, 130, 250),6[8:15]@purple,6[18:]@yellow,8[0::2],12[:]-14,16-18,20[12:]-22,24[1::]-26"}`
Those sections of code with no color specified will fallback on the default highlight color in the previous implementation.

Running the above code will result in a inlinecode block that looks like this:
image

#### Changes to the documentation
I have added a relevant section under the User Guide. Here is the link to it [https://inline-highlight.netlify.app/userguide/formattingcontents#code](https://inline-highlight.netlify.app/userguide/formattingcontents#code). 

#### Possible Improvements
Currently, users will be able to insert color in the format of either strings like `blue`, `red`, etc or hexadecimal representations of colors like `#000000`, `#e6e6fa`, etc. I intended for formats like `rgb(230,230,230)` or `rgba(255, 87, 51, 0.5)` to be recognized as well, which would be beneficial for people who are proficient with CSS.

#### The use of @
I used `@` to separate the code lines as well as the color. The reason why I chose `@` is because I wanted to use the `string::split` method and I had to choose a character that was not already present. Feel free to let me know if there are better/more intuitive way rather than `@`

**Testing instructions:**
1. Fetch the PR
2. Try out highlight feature as specified in the [user guide](https://markbind.org/userGuide/formattingContents.html).
3. Add a `@color` field behind every comma-separated highlight section.

Add flexibility to highlighting fenced code blocks

---

- [x] No unrelated changes 

---

**Reviewer checklist:**

- [x] Minor (when you add functionality in a backward compatible manner)


packages/core/src/lib/markdown-it/highlight/helper.ts
```diff
@@ -1,4 +1,22 @@
-// Common helper functions to be used in HighlightRule or HighlightRuleComponent
+import * as fs from 'fs';
+import * as path from 'path';
+
+const siteJsonPath = path.join(__dirname, '../../../../../../docs/site.json');
```
Tim-Siu, line 4: I'm not sure if I understand this correctly. Would this always read the site.json from the docs? It won't read the site.json from the custom site, right?

IanCheah: Yep, it just reads the site.json from the docs, which is wrong. I'm not sure how to extract the codeTheme field from the site.json from the custom site though.

AgentHagu: It seems that the Site class stores the `rootPath` of the user's MarkBind site. Perhaps there's a way to pass that information to this file to get an absolute path to the `site.json`.

IanCheah: Thanks for the inputs. I reused the initial logic to handle for the case where I want the default coloring. Initially, highlights are dealt with by the highlighted class. So what I did is to have a conditional check at each highlight, if there is no color specified, I reused the logic of the highlighted class, else, I simply added a custom style depending on the color input.

packages/core/src/html/codeblockProcessor.ts
```diff
@@ -142,9 +166,14 @@ export function highlightCodeBlock(node: MbNode) {
       return;
     }
 
-    const bounds = lineNode.attribs['hl-data'].split(',').map(boundStr => boundStr.split('-').map(Number));
-    bounds.forEach(([start, end]) => traverseLinePart(lineNode, start, end));
-
+    const bounds = lineNode.attribs['hl-data']
```
gerteck, line 169: This organization makes this section clearer 👍

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts
```diff
@@ -74,7 +77,8 @@ export class HighlightRuleComponent {
     const lineNumber = HighlightRuleComponent
       .isValidLineNumber(compString, 1, lines.length, lineNumberOffset);
     if (lineNumber) {
-      return new HighlightRuleComponent(lineNumber);
+      // return new HighlightRuleComponent(lineNumber, undefined, undefined, color);
```
gerteck, line 80: is this dead code? Can it be removed? It was slightly confusing.

packages/core/src/lib/markdown-it/highlight/HighlightRuleComponent.ts
```diff
@@ -74,7 +77,8 @@ export class HighlightRuleComponent {
...
+      return new HighlightRuleComponent(lineNumber, undefined, undefined);
```
gerteck, line 81: Could we keep the original 
`HighlightRuleComponent(lineNumber);`? Would it be clearer?

packages/core/src/lib/markdown-it/index.ts
```diff
@@ -9,6 +9,7 @@ import * as logger from '../../utils/logger';
 
 import { HighlightRule, HIGHLIGHT_TYPES } from './highlight/HighlightRule';
 import { Highlighter } from './highlight/Highlighter';
+// import { defaultColor, getCurrentTheme } from './highlight/helper';
```
gerteck, line 12: Same here

packages/core/src/lib/markdown-it/index.ts
```diff
@@ -146,38 +147,40 @@ markdownIt.renderer.rules.fence = (tokens: Token[],
   // wrap all lines with <span> so we can number them
   str = lines.map((line, index) => {
```
gerteck, line 148: I think the logic here could be a bit clearer. I think it could be optimized to be easier to follow. 

 1. Perhaps consider storing `immediateHighlight` in a separate variable instead of mutating `str` and returning it? This could make the logic much clearer to understand, e.g. something like
```
// FOR REFERENCE ONLY idk if it directly works
const immediateHighlight = highlightRules.find((rule) => {
  return rule.getHighlightType(currentLineNumber).some(({ highlightType, color }) => {
    if (highlightType === HIGHLIGHT_TYPES.WholeLine) {
      return Highlighter.highlightWholeLine(line, color);
    }
    if (highlightType === HIGHLIGHT_TYPES.WholeText) {
      return Highlighter.highlightWholeText(line, color);
    }
    return false;
  });
});

if (immediateHighlight) {
  return immediateHighlight;
}
```

2. Maybe pull out the `boundsWithColors` earlier, since it seems to have different logic (returns `false`, handled outside). Maybe it would be better to handle outside of `hasImmediateReturn`. Can also consider using `map`, `flatmap` as well to directly define `boundsWithColors`. But mostly to separate out the logic.

```
const boundsWithColors = highlightRules.flatMap((rule) =>
  rule.getHighlightType(currentLineNumber)
    .filter(({ highlightType, bounds }) => highlightType === HIGHLIGHT_TYPES.PartialText && bounds)
    .flatMap(({ bounds, color }) => bounds.map(bound => ({ bounds: bound, color: color || '' })))
);
```

packages/core/src/lib/markdown-it/index.ts
```diff
@@ -146,38 +147,40 @@ markdownIt.renderer.rules.fence = (tokens: Token[],
...
+
+    // Collect all bounds and colors for partial text highlights
+    const boundsWithColors: Array<{ bounds: [number, number], color: string }> = [];
+    const hasImmediateReturn = highlightRules.some((rule) => {
+      const results = rule.getHighlightType(currentLineNumber);
+      return results.some((result) => {
+        const { highlightType, bounds, color } = result;
+        const highlightColor = color || '';
+        if (highlightType === HIGHLIGHT_TYPES.WholeLine) {
+          str = Highlighter.highlightWholeLine(highlightedLine, highlightColor);
```
gerteck, line 160: Why is the parent `str` being overwritten in the loop?

If the parent str is the final concatenated result why is it being overwritten inside the loop?

Let me know if I am misunderstanding something. 

```
  str = lines.map((line, index) => {
    const currentLineNumber = index + 1;
    const highlightedLine = line; // Start with the original line

    // Collect all bounds and colors for partial text highlights
    const boundsWithColors: Array<{ bounds: [number, number], color: string }> = [];

    const hasImmediateReturn = highlightRules.some((rule) => {
      const results = rule.getHighlightType(currentLineNumber);
      return results.some((result) => {
        const { highlightType, bounds, color } = result;
        const highlightColor = color || '';

        if (highlightType === HIGHLIGHT_TYPES.WholeLine) {
          str = Highlighter.highlightWholeLine(highlightedLine, highlightColor);
          return true;

        } else if (highlightType === HIGHLIGHT_TYPES.WholeText) {
          str = Highlighter.highlightWholeText(highlightedLine, highlightColor);
          return true;

        } else if (highlightType === HIGHLIGHT_TYPES.PartialText && bounds) {
          bounds.forEach((bound) => {
            boundsWithColors.push({ bounds: bound, color: highlightColor });
          });

          return false;
        }
        return false;
      });
    });

    if (hasImmediateReturn) {
      return str;
    }

    if (boundsWithColors.length > 0) {
      return Highlighter.highlightPartOfText(line, boundsWithColors);
    }

    return `<span>${highlightedLine}\n</span>`;
  }).join('');
```

gerteck (changes requested):
Thank you for your PR! Have some changes and comments that I think need to be addressed, do let me know if I have misunderstood the logic.

packages/core/src/lib/markdown-it/index.ts
IanCheah: I totally agree with you! The current code I have is really difficult to follow. I have made the changes you proposed to improve the readability.

packages/core/src/html/codeblockProcessor.ts
```diff
@@ -102,19 +106,39 @@ function traverseLinePart(node: NodeOrText, hlStart: number, hlEnd: number): Tra
...
     } else {
       const [start, end] = data.highlightRange;
       const cleaned = util.unescapeHtml(child.data);
       const split = [cleaned.substring(0, start), cleaned.substring(start, end), cleaned.substring(end)];
       const [pre, highlighted, post] = split.map(md.utils.escapeHtml);
-      const newElement = cheerio(`<span>${pre}<span class="highlighted">${highlighted}</span>${post}</span>`);
-      cheerio(child).replaceWith(newElement);
+
+      if (color) {
+        const s = `<span>${pre}<span style='background-color: ${color};'>${highlighted}</span>${post}</span>`;
```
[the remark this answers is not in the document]
tlylt: I don't think validation is required.

I'll leave it to the main reviewers to decide.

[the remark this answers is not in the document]
IanCheah: I have discussed with @damithc. Supporting bootstrap is a good nice to have feature, which I can do after this PR is merged.

packages/core/src/lib/markdown-it/highlight/HighlightRule.ts
```diff
@@ -69,35 +86,84 @@ export class HighlightRule {
...
+            highlightType: HIGHLIGHT_TYPES.WholeText,
+            bounds: null,
+            color: this.color,
+          });
+        }
+      }
+    });
+
+    return results;
   }
```
yyccbb, line 167: I think this method might be a bit too long. Maybe it would be good to split into multiple smaller methods for better maintainability.

packages/core/src/lib/markdown-it/highlight/HighlightRule.ts
```diff
@@ -43,16 +57,19 @@ export class HighlightRule {
...
+      .map(compString => HighlightRuleComponent.parseRuleComponent(compString, lineOffset, lines));
     if (components.some(c => !c)) {
       // Not all components are properly parsed, which means
       // the rule itself is not proper
       return null;
     }
 
-    return new HighlightRule(components as HighlightRuleComponent[]);
+    const color: string = inputColor && COLOR_MAPPING[inputColor] ? COLOR_MAPPING[inputColor] : inputColor;
+    return new HighlightRule(components as HighlightRuleComponent[], color);
```
yyccbb, line 72: Would it be good if exeptions are raised when invalid rules are supplied?

packages/core/src/lib/markdown-it/highlight/helper.ts
yyccbb, line 1: I think more documentation is needed for collateAllIntervalsWithColors().
In addition, what is the rationale for using the color of the last overlapping interval instead of, for example, the longest?

yyccbb (commented):
Suggested some possible improvements. LGTM!

packages/core/src/lib/markdown-it/highlight/HighlightRule.ts
IanCheah: Invalid rules will simply be ignored and no highlight will appear

IanCheah:
> As for using "@" to indicate color, I am thinking would it be possible to just use "#" will a more robust parsing mechanism?

Users can specify a color using hex, for e.g. `#000000`. Using `#` to indicate color may lead to colors that are specified using hex to be incorrectly parsed. For e.g `1[:]##000000` will result in the color being wrongly parsed.

packages/core/src/lib/markdown-it/highlight/helper.ts
IanCheah: Actually, overlapped intervals should not happen in the first place. I have discussed with @damithc  regarding this. We came to a decision to use the last color because it felt more intuitive. Perhaps the user was trying to highlight something with color A but decided to use another color at the end, he or she can simply do that by overlapping the current highlighted region with another color.

damithc: Using the last colour seems like the more intuitive choice, as that's how painting works.

yyccbb: Noted. Thanks for the explanation!

yyccbb:
> > As for using "@" to indicate color, I am thinking would it be possible to just use "#" will a more robust parsing mechanism?
> 
> Users can specify a color using hex, for e.g. `#000000`. Using `#` to indicate color may lead to colors that are specified using hex to be incorrectly parsed. For e.g `1[:]##000000` will result in the color being wrongly parsed.

I think there was a typo. What I meant was something like `1[:]#00ff00` or `1[:]#y`. I think in general putting both hex color codes and color names after the same `#` will not lead to confusion? Correct me if I am wrong.
I just thought it would be good to avoid introducing new symbols as far as possible to make it easier on the users.

packages/core/src/lib/markdown-it/highlight/HighlightRule.ts
IanCheah: I agree! I have done the necessary changes.

IanCheah:
> > > As for using "@" to indicate color, I am thinking would it be possible to just use "#" will a more robust parsing mechanism?
> > 
> > 
> > Users can specify a color using hex, for e.g. `#000000`. Using `#` to indicate color may lead to colors that are specified using hex to be incorrectly parsed. For e.g `1[:]##000000` will result in the color being wrongly parsed.
> 
> I think there was a typo. What I meant was something like `1[:]#00ff00` or `1[:]#y`. I think in general putting both hex color codes and color names after the same `#` will not lead to confusion? Correct me if I am wrong. I just thought it would be good to avoid introducing new symbols as far as possible to make it easier on the users.

Currently what the code does is it will split by `@` to obtain the color specified on the right of `@`, the color will then be added to the HTML like this `style="background-color:${color};"`.

If split by `#` to obtain the resulting hex, say `000000`, there will need to be additional checks before adding it into the HTML. This is because the current approach will result in something like `style="background-color:000000;"`

What about something that is not already present?

lhw-1:
> I think there was a typo. What I meant was something like `1[:]#00ff00` or `1[:]#y`. I think in general putting both hex color codes and color names after the same # will not lead to confusion? Correct me if I am wrong.

On the contrary, given that the convention is to have `#` for color hex codes and not for names (e.g. "red" or "blue"), having `#` would not be consistent with standard color notation conventions. Moreover, I don't think I would like the additional logic we would have to implement (e.g. check if string is 6 or 3 letters containing hexadecimal to decide if it is hex or not) - seems unnecessary. I think using `@` is perfectly fine. 

I agree that the usage of the color highlighting makes the string quite complex though. The example given: 

```
java {.line-numbers highlight-lines="1[:]@pink,3['Inventory']@blue,3[4::6]@orange,4['It\'s designed']@#000000,5@rgb(120, 130, 250),6[8:15]@purple,6[18:]@yellow,8[0::2],12[:]-14,16-18,20[12:]-22,24[1::]-26"}
```

This is quite unreadable to me at this point, and if the person wants more organized color highlighting for their code (e.g. highlight whenever certain keywords appear), this will be quite annoying to modify one by one. I think it's certainly going to be worth discussing how to improve this feature. 

But as a first stage implementation, I think this is a good first step.

lhw-1 (approved):
Thanks for the PR @IanCheah! A few comments mostly on a potential refactor PR after this one (and likely after v6). 

Another idea is expanding the colored highlight idea for specific keywords; it might be nice if the user can e.g. pass in a key-value mapping for keywords to colors, and we apply that mapping over a given code block. Maybe we can even expand this to beyond code blocks, to the page (or pages) in general. I think this can be further discussed in a separate issue!

Issue #2548, opened 2024-05-07
Support Mermaid diagrams in  offline mode


damithc:
In some cases, a MarkBind site might need to run in locally, without Internet (e.g., running a course website during an exam). 

Suggestion: support Mermaid plugin in offline mode

damithc:
@Tim-Siu any thoughts?

Tim-Siu:
> @Tim-Siu any thoughts?

Hi @damithc ,

This is possible if we bundle the mermaid js code with Markbind. Do we need to have a separate mode for this or shall we just make it the default and only option?

damithc:
> This is possible if we bundle the mermaid js code with Markbind. Do we need to have a separate mode for this or shall we just make it the default and only option?

@Tim-Siu With other similar features/plugins, we bundle the required files and that is the only option. The downside is that we need to update the bundled file frequently, but I guess we have to live with that.

PR #2677, opened 2025-04-11
Bundle mermaid locally

### LINKED ISSUE #2667: Mermaid Script CORS issue

### Any related issues?

_No response_

Win 11

### MarkBind version

5.6.0

On the left is Vue3, right is Vue2. There seem to be a little issue with the rendering of the mermaid diagram.

_Originally posted by @IanCheah in https://github.com/MarkBind/markbind/issues/2622#issuecomment-2764379082_

Here is a screenshot:

Sometimes there is the CORS issue (left), sometimes there is no CORS issue (right). Both screenshots were taken of the same page at different times. Seems to be flaky.

Image

https://markbind-master.netlify.app/userguide/components/imagesanddiagrams#mermaid-diagrams

### Expected behavior

Should work.

### Anything else?

Seems to be quite a concerning regression actually, not sure why this is the case and why it was only spotted so late on by @IanCheah. 

However, since it is a CORS issue, if we directly support offline loading (#2548) and bundle it together it should totally resolve the issue.

Incogdino:
- Updated mermaid to use local version (V10.6.0) if a specific version is not specified.
- Updated tests.

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Closes #2667 
Part of #2548

**Overview of changes:**
- Updated mermaid to use local version (V10.6.0) if a specific version
is not specified.
- Updated tests.

We could explore upgrading mermaid to a later version in a future issue.

**Testing instructions:**

Bundle mermaid locally

- Updated mermaid to use local version (V10.6.0) if a specific version
is not specified.
- Updated tests.

---

- [X] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/vue-components/src/Tab.vue
+++ b/packages/vue-components/src/Tab.vue
@@ -29,9 +29,6 @@ export default {
     },
   },
   computed: {
-    headerRendered() {
-      return this.$refs.header.innerHTML;
-    },
     active() {
       return this._tabset.show === this;
     },
@@ -44,6 +41,9 @@ export default {
     disabledBool() {
       return toBoolean(this.disabled);
     },
+  },
+  mounted() {
+    this.headerRendered = this.$refs.header?.innerHTML;
   },
   created() {
     this._ingroup = this.$parent && '_tabgroup' in this.$parent;
--- a/packages/vue-components/src/TabGroup.vue
+++ b/packages/vue-components/src/TabGroup.vue
@@ -32,12 +32,12 @@ export default {
       // eslint-disable-next-line no-bitwise
       return ~this.tabs.indexOf(this._tabset.show);
     },
-    headerRendered() {
-      return this.$refs.header.textContent.trim();
-    },
     disabledBool() {
       return coerce.boolean(this.disabled);
     },
+  },
+  mounted() {
+    this.headerRendered = this.$refs.header.textContent.trim();
   },
   created() {
     this._tabgroup = true;


packages/core/src/plugins/mermaid.ts
```diff
@@ -1,15 +1,15 @@
 import cheerio from 'cheerio';
 import { PluginContext, FrontMatter } from './Plugin';
 
-const DEFAULT_CDN_ADDRESS = 'https://unpkg.com/mermaid@10/dist/mermaid.esm.min.mjs';
+const DEFAULT_MERMAID_PATH = './mermaidAssets/mermaid.min.js';
```
IanCheah, line 5: Since we are specifying `DEFAULT_MERMAID_PATH` here. How about exporting this path and all other places reference this specific one.

So in other places instead of doing `./mermaidAssets/mermaid.min.js` we replace it with `DEFAULT_MERMAID_PATH`

packages/core/src/plugins/mermaid.ts
```diff
@@ -1,15 +1,15 @@
...
```
Tim-Siu, line 4: Can be simplified `const DEFAULT_MERMAID_PATH = 'mermaidAssets/mermaid.min.js';` as well. I used this in other plugins, but not sure which is the best practice.

packages/core/src/plugins/mermaid.ts
Tim-Siu: Seems like we don't need mermaid path at other files at the moment? It could be future-proving though.

Tim-Siu (approved):
An effective and straightforward fix. LGTM.

packages/core/src/plugins/mermaid.ts
IanCheah: I was suggesting that because I saw `./mermaidAssets/mermaid.min.js` being used in quite a few files under `test_site/expected`.

lhw-1 (approved):
Thanks for the fix @Incogdino! Apart from the issues brought up by other reviewers, LGTM

packages/core/src/plugins/mermaid.ts
Incogdino: Hmm I think those are auto generated when I run `npm run updatetest`

packages/core/src/plugins/mermaid.ts
Incogdino: Ill update it so that it stays consistent with the convention that we are already using. I don't think there is much difference either, except maybe using `./` is a bit more explicit in defining the path.

gerteck:
Regressions Found:

image

Mermaid rendering does not work in panels (dynamic rendered content)

image

Some further work to refactor the logic would be good

PR #2685, opened 2025-04-13
Revert Local Mermaid Bundling

lhw-1:
This PR reverts #2677.

There are regressions caused by #2677; given that the errors are likely unrelated to the Vue 3 Migration as pointed out by @Tim-Siu [in a comment](https://github.com/MarkBind/markbind/issues/2667#issuecomment-2791957357) on #2667, we can revert this change for now, and take another look at the core issues after the v6 release.


Tim-Siu (approved):
Since the commit tree is clean, this auto-reversion is correct and passes the test. LGTM!

PR #2689, opened 2025-04-14
Fix hidden list numbering

### LINKED ISSUE #2628: Numbered lists not showing more than one digit
Problem noticed in https://nus-cs2103-ay2425s2.github.io/website/schedule/week8/project.html

Image

To replicate, you can preview the course website locally using instructions in https://github.com/nus-cs2103/website-base  (remember to switch to `cs2103` branch first)

Incogdino:
1. List are now left aligned. This ensures that the list numbering is not hidden by custom css style rules and also preserves the indentation of the list elements so that they are reasonably small enough for mobile users.

- [ ] Documentation update
- [X] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

Closes #2628 
**Overview of changes:**
Updated markbind default list css property `list-style-position` from `outside` to `inside`. This causes the list numbering to be inside the block box rather than outside the block box. More information [here](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position)

The downside(maybe) to this would be that certain list numbering will be slightly misaligned as the number of elements in the list increases. For instance, from number 9 to number 10, the start of content isn't aligned vertically.

image

Screenshot of the CS2103 website on my macOS:
Screenshot 2025-04-14 at 5 58 11 PM

**Testing instructions:**

Update list css

1. List are now left aligned. This ensures that the list numbering is
not hidden by custom css style rules and also preserves the
indentation of the list elements so that they are reasonably small
enough for mobile users.

---

- [X] No unrelated changes 

---

**Reviewer checklist:**

### DIFF
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -328,10 +328,23 @@ hr.thick-3 {
     border-top-width: 0.75rem;
 }
 
-ol{
-    padding-left: 1.5rem;
-    list-style-position: inside;
+ol {
+    counter-reset: foo;
+    display: table;
+    padding-left: 0.5rem;
 }
+
+ol > li {
+    counter-increment: foo;
+    display: table-row;
+}
+
+ol > li::before {
+    content: counter(foo) ".\00a0";
+    display: table-cell; /* aha! */
+    text-align: right;
+}
+
 
 ul {
     padding-left: 1.5rem;
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -331,6 +331,7 @@ hr.thick-3 {
 ol,
 ul {
     padding-left: 1.5rem;
+    list-style-position: inside;
 }
 
 h1,
--- a/packages/core-web/src/styles/markbind.css
+++ b/packages/core-web/src/styles/markbind.css
@@ -328,10 +328,13 @@ hr.thick-3 {
     border-top-width: 0.75rem;
 }
 
-ol,
-ul {
+ol{
     padding-left: 1.5rem;
     list-style-position: inside;
+}
+
+ul {
+    padding-left: 1.5rem;
 }
 
 h1,


gerteck:
Here's a side by side comparison of the effect of changing this style:
image

While this fix does indeed solve the issue, I wonder if there is a more stylish way to go about it...

Incogdino:
> Here's a side by side comparison of the effect of changing this style:
> 
> * The indentation of bullet points gets pushed "inside" the bullet content:
>   image
>   Left - New, Right - Old
> 
> While this fix does indeed solve the issue, I wonder if there is a more stylish way to go about it...

This is likely due to the `list-style-position` property applying to both ordered and unordered lists.

gerteck:
Thanks for the quick change! Does anyone else have any opinions on the increased indent on the numbered lists?

Incogdino:
> Why do we need to remove the padding defined in main.css?

I was thinking maybe in a case where authors define their own layouts but did not add any padding. With `main.css`, the numbered lists only gets hidden when the number of items in the list >10k

tlylt:
> I was thinking maybe in a case where authors define their own layouts but did not add any padding. With main.css, the numbered lists only gets hidden when the number of items in the list >10k

In my opinion it's a known risk that styles might break if users choose to update the default css, any other styling might also break if done incorrectly. Not that there's no need to improve the feature itself, but the approach this PR is going for doesn't resolve the problem nicely without introducing "new problems":
- `list-style-position: inside;` styling wise I would prefer to keep it outside, so there's no "misaligned" issue

Incogdino:
> In my opinion it's a known risk that styles might break if users choose to update the default css, any other styling might also break if done incorrectly. Not that there's no need to improve the feature itself, but the approach this PR is going for doesn't resolve the problem nicely without introducing "new problems":
> 
> * `list-style-position: inside;` styling wise I would prefer to keep it outside, so there's no "misaligned" issue

I agree that keeping styling outside should be preferred as it keeps the alignment of the list elements. Perhaps we can add it as a known issue/ warning to users who are trying to insert custom css stylesheets into their documents?

I also came across this [thread ](https://stackoverflow.com/questions/10428720/how-to-keep-indent-for-second-line-in-ordered-lists-via-css)on stack overflow where the accepted answer suggested using the browser's table layout algorithm to achieve the same effect. I've tried it out and it does preserve the outside property and keeps the indentation while avoiding the hidden list numbers.

image

I have not found any regressions yet. @tlylt what do you thhink? I have pushed the commit if you want to try it out.

tlylt:
> Perhaps we can add it as a known issue/ warning to users who are trying to insert custom css stylesheets into their documents?

i don't see any mention of how custom css can be used in our UG, perhaps we can add some details of that (and the potential pitfalls) in https://markbind-master.netlify.app/userguide/usinghtmljavascriptcss

Regarding the new proposed changes, I have a preference to keep the current simple version as is since the existing layout css works well enough, but thanks @Incogdino for the investigation and finding out the root cause🚀

Incogdino:
> i don't see any mention of how custom css can be used in our UG, perhaps we can add some details of that (and the potential pitfalls) in https://markbind-master.netlify.app/userguide/usinghtmljavascriptcss
> 
> Regarding the new proposed changes, I have a preference to keep the current simple version as is since the existing layout css works well enough, but thanks @Incogdino for the investigation and finding out the root cause🚀

Okay in that case I shall close this PR and raise an issue to update the documentation

PR #2836, opened 2026-02-16
Migrate CLI output from CJS to ESM

Harjun751:

- [ ] Documentation update
- [ ] Bug fix
- [X] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
At a high level,
* Introduce configuration required for ESM output
* Update import syntax where required
* Workaround packages broken by ESM imports
* Change jest test runner from ts-jest to use babel transpilation

(I have additional details written down in the commit bodies)

Happy CNY folks!  Working on this PR was a lot of trouble due to the very many subtle ways that the packages broke when using ESM. There were about 4 major breakages:

### Tests in `cli/tests/functional/` (POSSIBLE BREAKING CHANGE....?!?)
A significant thing that broke was the tests in the functional/ folder. Due to the added ESM configuration, it parsed the plugins (which are javascript files) within the test sites as ESM, so it broke when building the site with those files since they used CJS syntax (namely, `require`).

The fix was this was to add a package.json at the scope of the test sites to let those plugin files be interpreted as CJS instead of ESM.

I don't think this will be an issue for users since I don't think there's a use case where there will be a package.json in the scope of the custom plugins. Either way, if there is, the user should already be using ESM syntax then, so they shouldn't encounter a breakage. But, if in the case that

1. The user has a package.json  in the root of their markbind site
2. That package.json  is set up with `type: "module"` (ESM syntax)
3. The user is using custom plugins
4. The custom plugins are written in CJS syntax

Then, the site will not be able to be built.

### Lodash
We discussed this in telegram - pasting my findings here for any poor soul in the future who somehow ends up here and needs this information for some reason:

<details>
Yea okay I somewhat figured it out, but not really. Really insidious issue.

I was working on cjs->esm migration, and due to that some of the imports and types even had to be updated, and one of the ones I needed to update was @types/lodash. I needed to bump that package up, it was currently set to ^4.14.181. So that I did, I bumped it up to ^4.17.0 and then when I ran npm i it failed completely.

It was raising an issue that the core package failed to build due to TS2790: The operand of a 'delete' operator must be optional. This was really weird to me because I did not touch the core package at all, and I don't see why this issue was being raised up only NOW.

For reference, it was erroring out on the delete call there at the bottom.
```javascript
    const bounds = lineNode.attribs['hl-data']
      .split(',')
      .map((boundStr) => {
        const [range, color] = boundStr.split(':');
        const [start, end] = range.split('-');
        return [start, end, color];
      });
    bounds.forEach(([start, end, color]) => traverseLinePart(lineNode, Number(start), Number(end), color));
    delete lineNode.attribs['hl-data'];
```

This code didn't even seem to be related to lodash, so why was it even erroring?

My first suspicion was that somehow running npm i bumped the typescript package version, causing stricter checks to be enacted. I looked at the package-lock.json diff and this was not the case - the ONLY changes were related to lodash.

Next line of thinking was did the htmlparser2 package change in some way? The lineNode object was of type DOMElement, and I was wondering if somehow in someway in hell updating lodash types changed the object structure of DOMElement. 

Probably not the case, so I looked on. Checking the core package-json, I saw that types/lodash was actually defined in there too.

In monorepos/npm workspaces, when npm sees that more than 1 subpackage has a similar dependency, it ["hoists"](https://www.jonathancreamer.com/inside-the-pain-of-monorepos-and-hoisting/) dependencies. npm will select a compatible version for the 2 subpackages using semver to evaluate. The dependencies will all be in the root node_modules folder and symlinked to the specific subpackage that requires them.

So what was happening was that by updating the cli's types/lodash, npm picked another "theoretically compatible" version of types/lodash to use in core. This updated version, for some reason, was NOT, in fact, compatible?!11?? EVEN though that 1) its just a types/ module and 2) it was a minor version bump, that broke core.

The funny thing is that I still don't even know how it broke. All i know is that THIS check occured before the delete call:
```javascript
    if ((!lineNode.attribs) || !_.has(lineNode.attribs, 'hl-data')) {
      return;
    }
```

You can see that lodash *was* used here. But that doesn't explain how types/lodash somehow modifies the inferred type of DOMElement, from having an optional "attribs" element to it being strictly required, hence causing typescript to yell out.

Compouding on my confusion on this issue, the has function in lodash wasn't even touched according to the changelog of lodash itself. I also can't find changelogs for types/lodash because they don't seem to track that in the DefinitelyTyped repo.

The ""fix"" was to replace "^4.14.blahblahblah" in the core/ package to remove the "^" symbol so that that version was forced to be used for only the core package.
</details>

### Winston
Winston v2 doesn't work with ESM. v3 fixes this, but migration is not trivial so we should put that off for a later time.
To fix this, I created a barrel file to import it using CJS syntax.

### jest/ts-jest
This was the worst one to debug. It broke for specifically manual mocks - I tried a lot of things fixing it and it's ugly. Wish I had the resources to compile here so I could show yall, but I didn't save them and I don't want to go through the trauma of searching for it. Most of my time was on this and I was really going crazy lol. In the end, changed to using babel to transpile ESM -> CJS.

**Testing instructions:**

Migrate CLI from CommonJS to ES Modules

- Convert CLI package to ESM with "type": "module"
- Switch from ts-jest to babel-jest for testing
- Update all imports to ESM syntax
---

- [ ] No unrelated changes 

---

**Reviewer checklist:**

- [x] Major (when you make incompatible API changes)


packages/cli/.eslintrc.cjs
```diff
@@ -0,0 +1,38 @@
+module.exports = {
+  extends: ['../../.eslintrc.js'],
+  overrides: [
+    {
+      files: ['./**/*.ts'],
```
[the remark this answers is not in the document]
Harjun751: We need to update eslint to a higher version to allow use to use the latest ecmascript features. Currently, it's right that it fails on `import.meta`, but we should tackle this in another PR intended to upgrade the eslint dependency I think.

packages/cli/tsconfig.json
```diff
@@ -1,11 +1,13 @@
 {
   "extends": "../../tsconfig_base.json",
-  "exclude": ["node_modules", "**/*.test.ts", "dist", "**/*.test.js", "test/**/*.js", "**/__mocks__", "coverage"],
+  "exclude": ["node_modules", "dist", "**/*.test.js", "test/**/*.js", "**/__mocks__", "coverage"],
```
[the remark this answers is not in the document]
Harjun751: For now this is required since the functional tests are built and then ran. Maybe we can consider moving the test files to be ran with tsx, but I think for now this works?

Harjun751:
@gerteck alright, it's ready now! With #2838 merged I'm now using some node-20 specific syntax which works with the CI.

packages/cli/src/lib/live-server/index.js
```diff
@@ -15,25 +15,27 @@
...
+import os from "os";
+import chokidar from "chokidar";
+import fsUtil from "@markbind/core/src/utils/fsUtil";
+import "colors";
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
 
 // CHANGED: added absolute path that directs to the live-server directory
-const pathToLiveServerDir = path.dirname(require.resolve('live-server'));
+// Use createRequire for Node 18 compatibility (import.meta.resolve is synchronous only in Node 20.6.0+)
```
gerteck: outdated comment?

packages/cli/test/functional/test_site/_markbind/plugins/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
gerteck: Understand the situation with the tests in `cli/tests/functional/ `
The fix to 
> add a package.json at the scope of the test sites to let those plugin files be interpreted as CJS instead of ESM.

As you mentioned,
> Either way, if there is, the user should already be using ESM syntax then, so they shouldn't encounter a breakage.

Hence, the `package.json` here is intended to be a short term fix right? We should also be hopefully migrating these test custom plugins in to ESM formats, since we would expect custom plugins to be written in ESM moving forward.

Maybe we should open an issue to track this chore as well? I can offer to tackle this hopefully small task if needed.

Or we can directly migrate them in this PR as well.

packages/cli/tsconfig.json
gerteck: Actionables (add as separate comment so easier for you to see):

* Add `**/*.test.ts` to the exclude array to prevent unit tests from being compiled into dist/. (The two unit tests are run directly through jest), as a short term fix.

* To prevent tests from being shipped, add to `.npmignore` the line: `dist/test`. So it does not include it in the npm release.

Additionally, can consider opening an issue as per your suggestion on running with tsx to make sure it is documented and potentially worke on in the future

gerteck (approved):
LGTM, awesome first step towards ESM

Issue #2849, opened 2026-02-28
Misleading deploy command output


yihao03:
### Please confirm that you have searched existing issues in the repo

Yes, I have searched the existing issues

### Tell us about your environment

WSL Ubuntu 24.04 LTS

### MarkBind version

6.3.1

### Describe the bug and the steps to reproduce it

When doing `markbind d`, the console always shows:
```sh
info: The website has been deployed at: https://<author>.github.io/<repo>
```
regardless of the actual deployment status which could be misleading. e.g. the ci/cd pipeline could fail and the website is not deployed

### Expected behavior

change the wording to `deployment has been initiated, view status at <workflow url>` or similar

gerteck:
Maybe can tackle #1575 at the same time in this issue

gerteck:
Upon further investigation, the reason why the errors don't log on wrong publish is because of `gh-pages` dependency. The current latest `npm` version (6.3.0), in the `publish` function:

> When an error occurs, it calls done(error) which calls the callback — but it never calls return Promise.reject(error).

This means that the catch error will never trigger, so it will always think the publish is successful as it always resolves.

A note is that in the latest master branch, they have fixed this issue, but they have not made a release since jan 2025 for this fix. 

So, unfortuntately, adding process.exit does not fix it because it never hits that part... :( 


The only proper way to fix it is when `gh-pages` makes a new release and we update the dependency ver,,,

PR #2852, opened 2026-03-02
Update markbind deploy success log message

yihao03:
- **Update deploy log message**
- **Refactor helper function into util directory**
- **Lint and update tests**

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
Addresses #2849.
Now update the log message to point to URL of the GitHub Actions **and** the deployed site URL

**Testing instructions:**
Deploy a website using markbind and observe the log message.

Update deploy log message to include GitHub Actions URL and deployed site URL

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**

>
> - what was the old feature that was made obsolete
> - any replacement feature (if any), and
> - how the author should modify his website to migrate from the old feature to the replacement feature (if possible).

### DIFF
--- a/packages/core/index.ts
+++ b/packages/core/index.ts
@@ -2,4 +2,4 @@ import { Site } from './src/Site/index.js';
 import { Template } from './src/Site/template.js';
 
 export { Site, Template };
-export type { DeployResult } from './src/Site/SiteDeployManager';
+export type { DeployResult } from './src/Site/SiteDeployManager.js';


gerteck (commented):
Wondering, does the changes account for enough flexibility in supporting deployment through other CI platforms (e.g. Travis, AppVeyor CI, Circle CI), and also deploying to 
netlify? 

https://markbind.org/userGuide/deployingTheSite.html#deploying-to-github-pages

Changing message from "deployed at" to "will be deployed at" and adding the Actions URL is a good fix to avoid misleading, as `ghpages.publish` only resolves on a successful git push, not when the site is actually live.

However, Ithe exported DeployResult type shape:
```ts
export type DeployResult = {
  ghPagesUrl: string | null,
  ghActionsUrl: string | null,
};
```

The field ghActionsUrl couples to GitHub Actions specifically. While currently the deploy command only handles GitHub URLs (not technically a regression), exporting this type from `core/index.ts` can make it reduce flexibility and make it hard to support the logging of other CI platforms like Travis, AppVeyor, etc and or deployment targets like Netlify.

Is it possible to consider a more platform-neutral field name like `ciDashboardUrl` and also make it flexible to these other types of CI and deployment? Perhaps there is room to ehance the logging for deploy much more in a single shot.

gerteck:
Todos:

* Address copilot comments regarding some edge cases (either implement, or comment why not applicable and resolve them)
* Directly put the function in the same file `cmd/deploy.ts` where it is called, don't need it in a separate util file  as it's 15 lines and not used elsewhere

Otherwise great work on this 👍

gerteck (changes requested):

packages/core/src/Site/SiteDeployManager.ts
```diff
@@ -159,57 +169,80 @@ export class SiteDeployManager {
...
+      // https://github.com/<name|org>/<repo>.git (HTTPS)
+      const repoNameWithExt = parts[parts.length - 1];
+      const repoName = repoNameWithExt.substring(0, repoNameWithExt.lastIndexOf('.'));
+      const name = parts[parts.length - 2].toLowerCase();
+      return { name, repoName };
+    } else if (remoteUrl.startsWith(SSH_PREAMBLE)) {
+      // git@github.com:<name|org>/<repo>.git (SSH)
+      const repoNameWithExt = parts[parts.length - 1];
+      const repoName = repoNameWithExt.substring(0, repoNameWithExt.lastIndexOf('.'));
+      const name = parts[0].substring(SSH_PREAMBLE.length);
```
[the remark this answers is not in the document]
yihao03: Notably, in the same file we have `extractRepoSlug()` that forces `.git` on the remotes. To avoid code duplication and ensure consistent logic, we shall just relax the `.git` restriction and call `parseGitHubRemoteUrl()` in `extractRepoSlug()`

gerteck (approved):
lgtm, the export of `DeployResult` seems a bit weird but I suppose this is the best way to do it for now

packages/core/src/Site/SiteDeployManager.ts
```diff
@@ -159,57 +169,82 @@ export class SiteDeployManager {
...
+    const parts = remoteUrl.split('/');
+
+    // get repo name
+    const repoNameWithExt = parts[parts.length - 1];
+    const dotIndex = repoNameWithExt.lastIndexOf('.');
+    const repoName = dotIndex === -1 ? repoNameWithExt : repoNameWithExt.substring(0, dotIndex);
+
+    if (remoteUrl.startsWith(HTTPS_PREAMBLE)) {
+      // https://github.com/<name|org>/<repo>.git (HTTPS)
+      const owner = parts[parts.length - 2];
```
MoshiMoshiMochi: Just checking here, we don't support enterprise urls right? (since the `constructGhPagesUrl()` builds a site defaulting to the `github.io`.  Plus the docs don't mention anything about deploying to enterprise urls.

But if we do then, if I'm not wrong, their urls start with something like `https://github.enterprise.com`, so the current `HTTPS_PREAMBLE` check would fail to catch it.

yihao03: Upon closer inspection, github enterprise doesn't use a customized subdomain, rather they have a slug for their organisation. the only instances where there could be customized urls would be self hosted instances. see https://docs.github.com/en/enterprise-cloud@latest/admin/managing-your-enterprise-account/changing-the-url-for-your-enterprise 
therefore i think this is out of scope

Issue #2882, opened 2026-04-06
Warning when running for a fresh site


damithc:
version: v7.0.0

Commands:
```
markbind init
markbind serve -o
```

Output:

<img width="1799" height="795" alt="Image" src="https://github.com/user-attachments/assets/3acffd23-560e-4811-b208-d864c194b9da" />

Are these warnings/errors expected?

damithc:
I tried v7.0.0 with the CS2103 website. It seems PlantUML and Mermaid diagrams are not working.

<img width="990" height="264" alt="Image" src="https://github.com/user-attachments/assets/5b859b1d-f975-4966-aad7-30af7a5fc286" />

<img width="989" height="361" alt="Image" src="https://github.com/user-attachments/assets/455036bc-cc36-44f4-9927-b5eee55db9dc" />

Do I need to update the node version?

gerteck:
Yes, maybe we can add a node version guard check to exit early below node v22.22 in a minor release, advice to use v6 if using a lower node version

damithc:
With @gerteck's help, we traced the warnings to an outdated node version I had. 

The errors seems to be related to the PageFind feature, and is persistent even after updating the node version. @MoshiMoshiMochi have a look when you have time.

damithc:
> The errors seems to be related to the PageFind feature, and is persistent even after updating the node version. [@MoshiMoshiMochi](https://github.com/MoshiMoshiMochi) have a look when you have time.

In fact it is because these pages don't exist when using `-o` flag. I think we should not show those as errors in the console, as missing pages is expected when using `-o` mode.

MoshiMoshiMochi:
Noted. Will investigate further on this issue to fix it.

damithc:
@MoshiMoshiMochi If the site is not using PageFind search, do we still build the PageFind index?

MoshiMoshiMochi:
Hi Prof @damithc, yes! Not using Pagefind search will skip its indexing during the build. 

As PageFind search is tied to `enableSearch` option in `site.json` (which is `true` by default), the only way to not use PageFind search would be to set it to `false`. This would skip all Pagefind related things during the build.

However, as `enableSearch` is also tied to our current built-in search, setting it to `false` would inherently disable it as well. Since Pagefind is still an incomplete feature, let me know if you would prefer to create a temporary option to toggle Pagefind search instead, or if you feel the current `enableSearch` option is fine.

PR #2885, opened 2026-04-06
Replace error logging with silent skip when page DNE

MoshiMoshiMochi:
Pages that exist get indexed normally while pages that don't are skipped silently without returning an error message for missing pages.

- [ ] Documentation update
- [x] Bug fix
- [ ] Code maintenance
- [ ] DevOps
- [ ] Others, please explain:

**Overview of changes:**
 #2882
Replace error logging with silent skip if page does not exist when serving one-page.

There should no longer be any pagefind related indexing errors when serving one-page.
image

This issue arose from #2880, where I began switching the searchable page logic to respect page's option. It was an oversight an oversight by me during development as I had only tested the indexing with a full build. However, using `markbind serve -o` only the landing page is built at startup. The `sitePages.pages` array contains all pages defined in `site.json`, but most of their HTML files haven't been generated yet as they are built on-demand when u navigate to them.

As prof @damithc mentioned,
> In fact it is because these pages don't exist when using -o flag. I think we should not show those as errors in the console, as missing pages is expected when using -o mode.

Hence, I added a check if a file exist first. If it doesn't & is serving one-page,  silently skip to the next file to index. This ensures that the failed to index error doesn't occur because the file did not exist.

This is 100% my bad and on me, my apologies for not realising this earlier. Will be sure to be more thorough when testing out newly added/updated features in the future.

**Testing instructions:**

Add silent skip when indexing a page that does not exist

---

- [ ] No unrelated changes 

---

**Reviewer checklist:**


gerteck:
no worries @MoshiMoshiMochi , it is but a edge case that is not breaking  😄 

Speedy quick fix!

if anything, these nits should be caught by the PR reviewer as well 😅 

It could also mean our testcases are not comprehensive enough, to which we can improve on!

MoshiMoshiMochi:
As discussed, I've update enabling pagefind to its' own option where it is disabled by default. Users will now what to enable it within by declaring the pagefind configuration and setting `enablePagefind` to `true`

```json
{
  "pagefind": {
    "enablePagefind": true
  }
}
```

Hence, as it is no longer tied to `enableSearch` which is `true` by default, all functional tests no longer captures pagefind.

packages/core/src/Site/SiteGenerationManager.ts
yihao03, line 1: is it truly the best way to do this? can we just skip indexing the site when serving one page?
We can add `this.onePagePath` to the if clause on line 319 to skip indexing altogether. 

on another note, unrelated but curious, why is the pagefind import done this way? 
```typescript
import * as pagefind from 'pagefind';
```

```typescript
const { createIndex, close } = pagefind;
```

instead of just doing 
```typescript
import { createIndex, close } from 'pagefind';

Comment (gerteck) [reply packages/core/src/Site/SiteGenerationManager.ts:1]: Input just for the curiosity part

For some reason, the official docs' code snippet (which was probably used as the first draft) as on https://pagefind.app/docs/node-api/ also does it that way

image

Comment (MoshiMoshiMochi) [reply packages/core/src/Site/SiteGenerationManager.ts:1]: > is it truly the best way to do this? can we just skip indexing the site when serving one page?
We can add this.onePagePath to the if clause on line 319 to skip indexing altogether.

I see where you're coming from and technically we could since this is kind of a band-aid fix. But ideally pagefind will support incremental page indexing like the current built-in search does (which I am working on it in this [PR](#2890). And I guess logically speaking if the user were to serve with just one-page, they might expect it to index that page (though it is kinda useless without the [incremental indexing](#2890)).

As such, right now there is an overall incompatibility between `indexSiteWithPagefind` when serving one-page as the errors raised were because.
> The sitePages.pages array contains all pages defined in site.json, but most of their HTML files haven't been generated yet as they are built on-demand when u navigate to them.

And the current fix will be more compatible with the changes made in #2890 

--------

Regarding the curiosity, honestly I don't think there's a difference. As @gerteck mentioned, the official docs does it that way so I kinda just followed suit. 

But my guess would be that by only referencing `const { createIndex, close } = pagefind; ` within `indexSiteWithPagefind` directly, we kinda keep the global scope of `SiteGenerationManager` cleaner. `createIndex()` & `close()` are kinda generic function names (well maybe less so `createIndex`). So if for some reason we needed to use another `close()` method from a different interface we could potentially face naming collision issues.

But hey this is just my speculation and justification for keeping the imports this way.

Also I've kinda mocked my testcases to these namespace imports so if we do decide to change it, maybe we could do it in another PR lol.

Comment (yihao03) [reply packages/core/src/Site/SiteGenerationManager.ts:1]: on @gerteck 's screenshot from the official site, I think it's not exactly what's happening here? The example invokes the `createIndex()` method on the `pagefind` object while here, we are destructuring the `createIndex()` function from the `pageFind` object.

>  we kinda keep the global scope of SiteGenerationManager cleaner. createIndex() & close() are kinda generic function names

right right i never thought of that! 

> And I guess logically speaking if the user were to serve with just one-page, they might expect it to index that page

Doesn't this implementation indiscriminately skip all pages?

Comment (MoshiMoshiMochi) [reply packages/core/src/Site/SiteGenerationManager.ts:1]: > Doesn't this implementation indiscriminately skip all pages?

Not quite. Based on these fixes, pagefind will still index the respective page when serving just one-page. It just doesn't log the file DNE exist error when serving one-page as it isn't applicable in this case.

In the case of serving one-page, only the first page will be indexed (assuming of course that the first page is searchable). 

```ts
const indexingResults = await Promise.all(
            searchablePages.map(async (page) => {
              const fileExists = await fs.pathExists(page.pageConfig.resultPath);
              if (!fileExists && this.onePagePath) {
                return null;
              }
              try {
                const content = await fs.readFile(page.pageConfig.resultPath, 'utf8');
                const relativePath = path.relative(this.outputPath, page.pageConfig.resultPath);

                return index.addHTMLFile({
                  sourcePath: relativePath,
                  content,
                });
              } catch (error) {
                logger.error(`Failed to index ${page.pageConfig.resultPath}: ${error}`);
                return null;
              }
            }),
          );
```

Of course this isn't the best fix as it means Pagefind will still iterate through all searchable pages only to skip most of them. However, this will be more explicitly addressed in the #2890 where there is a way of just indexing specific pages when needed. 

Either ways, I still feel that this current "band-aid" fix is valuable as it serves as an additional layer of protection (from inaccurate logs) in case `indexSiteWithPagefind` is called when serving one-page.

gerteck (approved):
LGTM

